Issue #5751, opened 2019-10-28
Coprcessor Batch Framework all RPN functions list (both implemented and not implemented).


lonng:
The following list of functions is already implemented by the TiDB, but has not been ported to the Coprocessor Batch framework. **All unchecked items can be claimed**

- [ ] ASCII
- [x] AbsDecimal
- [x] AbsInt
- [x] AbsReal
- [x] AbsUInt
- [ ] Acos
- [ ] AddDateAndDuration
- [ ] AddDateAndString
- [ ] AddDateDatetimeInt
- [ ] AddDateDatetimeString
- [ ] AddDateIntInt
- [ ] AddDateIntString
- [ ] AddDateStringDecimal
- [ ] AddDateStringInt
- [ ] AddDateStringString
- [ ] AddDatetimeAndDuration
- [ ] AddDatetimeAndString
- [ ] AddDurationAndDuration
- [ ] AddDurationAndString
- [ ] AddStringAndDuration
- [ ] AddStringAndString
- [ ] AddTimeDateTimeNull
- [ ] AddTimeDurationNull
- [ ] AddTimeStringNull
- [ ] AesDecrypt
- [ ] AesEncrypt
- [ ] Asin
- [ ] Atan1Arg
- [ ] Atan2Args
- [ ] Bin
- [ ] BitAndSig
- [ ] BitCount
- [ ] BitLength
- [ ] BitNegSig
- [ ] BitOrSig
- [ ] BitXorSig
- [ ] CRC32
- [x] CaseWhenDecimal
- [x] CaseWhenDuration
- [x] CaseWhenInt
- [x] CaseWhenJson
- [x] CaseWhenReal
- [x] CaseWhenString
- [x] CaseWhenTime
- [x] CastDecimalAsDecimal
- [x] CastDecimalAsDuration
- [x] CastDecimalAsInt
- [x] CastDecimalAsJson
- [x] CastDecimalAsReal
- [x] CastDecimalAsString
- [x] CastDecimalAsTime
- [x] CastDurationAsDecimal
- [x] CastDurationAsDuration
- [x] CastDurationAsInt
- [x] CastDurationAsJson
- [x] CastDurationAsReal
- [x] CastDurationAsString
- [x] CastDurationAsTime
- [x] CastIntAsDecimal
- [x] CastIntAsDuration
- [x] CastIntAsInt
- [x] CastIntAsJson
- [x] CastIntAsReal
- [x] CastIntAsString
- [x] CastIntAsTime
[...]

PR #9088, opened 2020-11-22
copr:add make_time

Saigyouji-Yuyuko:
Signed-off-by: yuyuko1030@gmail.com

Issue Number: #5751 

### Check List 

Tests 

- Unit test

### Release note 
No release note.

### DIFF
--- a/components/tidb_query_expr/src/impl_time.rs
+++ b/components/tidb_query_expr/src/impl_time.rs
@@ -5,6 +5,7 @@ use tidb_query_codegen::rpn_fn;
 use tidb_query_datatype::expr::EvalContext;
 
 use crate::RpnFnCallExtra;
+use std::convert::TryInto;
 use tidb_query_common::Result;
 use tidb_query_datatype::codec::data_type::*;
 use tidb_query_datatype::codec::mysql::duration::{
@@ -17,7 +18,6 @@ use tidb_query_datatype::codec::mysql::{Duration, TimeType};
 use tidb_query_datatype::codec::Error;
 use tidb_query_datatype::expr::SqlMode;
 use tidb_query_datatype::FieldTypeAccessor;
-use tidb_query_datatype::FieldTypeFlag;
 
 #[rpn_fn(nullable, capture = [ctx])]
 #[inline]
@@ -224,7 +224,7 @@ pub fn make_date(ctx: &mut EvalContext, year: &Int, day: &Int) -> Result<Option<
     Ok(Some(ret))
 }
 
-#[rpn_fn(capture = [extra,args])]
+#[rpn_fn(capture = [extra, args])]
 #[inline]
 pub fn make_time(
     extra: &RpnFnCallExtra,
@@ -233,59 +233,42 @@ pub fn make_time(
     minute: &Int,
     second: &Real,
 ) -> Result<Option<Duration>> {
-    if second.is_nan() {
-        return Ok(None);
-    }
-    let minutei64 = *minute;
-    let secondi64 = second.into_inner().trunc() as i64;
-    if minutei64 < 0
-        || secondi64 < 0
-        || minutei64 > (MAX_MINUTE_PART as i64)
-        || secondi64 > (MAX_SECOND_PART as i64)
+    let (is_negative, mut hour) = if args[0].field_type().is_unsigned() {
+        (false, *hour as u64)
+    } else {
+        (hour.is_negative(), hour.wrapping_abs() as u64)
+    };
+
+    // Filter out the number that is negative or greater than MAX_MINUTE_PART.
+    let mut minute: u32 = match (*minute).try_into().ok().filter(|m| *m <= MAX_MINUTE_PART) {
+        Some(minute) => minute,
+        None => return Ok(None),
+    };
+
+    let mut nanosecond = (second.fract().abs() * NANOS_PER_SEC as f64) as u32;
+
+    // Filter out the number that is negative or greater than MAX_SECOND_PART.
+    let mut second: u32 = match (second.trunc() as i64)
+        .try_into()
+        .ok()
+        .filter(|s| *s <= MAX_SECOND_PART)
     {
-        return Ok(None);
-    }
-    let mut minute = minutei64 as u32;
-    let mut nanosecond = (second.into_inner().fract() * (NANOS_PER_SEC as f64)) as u32;
-    let mut second = second.into_inner().trunc() as u32;
+        Some(second) => second,
+        None => return Ok(None),
+    };
 
-    let is_unsigned = match args[0] {
-        crate::RpnStackNode::Scalar { field_type, .. } => field_type,
-        crate::RpnStackNode::Vector { field_type, .. } => field_type,
-    }
-    .as_accessor()
-    .flag()
-    .contains(FieldTypeFlag::UNSIGNED);
-    let mut overflow = false;
-    let mut houri64 = *hour;
-    let mut neg = houri64 < 0;
-    if neg && is_unsigned {
-        houri64 = MAX_HOUR_PART as i64;
-        overflow = true;
-        neg = false;
-    }
-    if houri64 > MAX_HOUR_PART as i64 || houri64 < -(MAX_HOUR_PART as i64) {
-        houri64 = MAX_HOUR_PART as i64;
-        overflow = true;
-    }
-    let hour = houri64.abs() as u32;
-
-    if hour == MAX_HOUR_PART
-        && minute == MAX_MINUTE_PART
-        && second == MAX_SECOND_PART
-        && nanosecond > 0
-    {
-        nanosecond = 0;
-    }
-
-    if overflow {
+    let is_overflow = (hour, minute, second, nanosecond)
+        > (MAX_HOUR_PART as _, MAX_MINUTE_PART, MAX_SECOND_PART, 0);
+    if is_overflow {
+        hour = MAX_HOUR_PART as _;
         minute = MAX_MINUTE_PART;
         second = MAX_SECOND_PART;
         nanosecond = 0;
     }
+
     match Duration::new_from_parts(
-        neg,
-        hour,
+        is_negative,
+        hour as _,
         minute,
         second,
         nanosecond,
--- a/components/tidb_query_expr/src/impl_time.rs
+++ b/components/tidb_query_expr/src/impl_time.rs
@@ -4,6 +4,7 @@ use tidb_query_codegen::rpn_fn;
 
 use tidb_query_datatype::expr::EvalContext;
 
+use crate::RpnFnCallExtra;
 use tidb_query_common::Result;
 use tidb_query_datatype::codec::data_type::*;
 use tidb_query_datatype::codec::mysql::time::extension::DateTimeExtension;
@@ -218,6 +219,94 @@ pub fn make_date(ctx: &mut EvalContext, year: &Int, day: &Int) -> Result<Option<
     Ok(Some(ret))
 }
 
+pub fn make_time(
+    ctx: &mut EvalContext,
+    hour: &Int,
+    minute: &Int,
+    second: &Real,
+    fsp: i8,
+    isunsinged: bool,
+) -> Result<Option<Duration>> {
+    let mut hour = *hour;
+    let mut minute = *minute;
+    if second.is_nan() {
+        return Ok(None);
+    }
+    let mut second = second.into_inner();
+
+    if minute < 0 || minute >= 60 || second < 0.0 || second >= 60.0 {
+        return Ok(None);
+    }
+    let mut overflow = false;
+    if hour < 0 && isunsinged {
+        hour = 838;
+        overflow = true;
+    }
+    if hour < -838 {
+        hour = -838;
+        overflow = true;
+    } else if hour > 838 {
+        hour = 838;
+        overflow = true;
+    }
+
+    if (hour == -838 || hour == 838) && minute == 59 && second > 59.0 {
+        second = 59.0;
+    }
+
+    if overflow {
+        minute = 59;
+        second = 59.0;
+    }
+
+    match Duration::parse(
+        ctx,
+        format!("{:02}:{:02}:{}", hour, minute, second).as_bytes(),
+        fsp,
+    ) {
+        Ok(t) => Ok(Some(t)),
+        Err(e) => Err(e.into()),
+    }
+}
+
+#[rpn_fn(capture = [ctx,extra])]
+#[inline]
+pub fn make_time_signed(
+    ctx: &mut EvalContext,
+    extra: &RpnFnCallExtra,
+    hour: &Int,
+    minute: &Int,
+    second: &Real,
+) -> Result<Option<Duration>> {
+    make_time(
+        ctx,
+        hour,
+        minute,
+        second,
+        extra.ret_field_type.get_decimal() as i8,
+        false,
+    )
+}
+
+#[rpn_fn(capture = [ctx,extra])]
+#[inline]
+pub fn make_time_unsigned(
+    ctx: &mut EvalContext,
+    extra: &RpnFnCallExtra,
+    hour: &Int,
+    minute: &Int,
+    second: &Real,
+) -> Result<Option<Duration>> {
+    make_time(
+        ctx,
+        hour,
+        minute,
+        second,
+        extra.ret_field_type.get_decimal() as i8,
+        true,
+    )
+}
+
 #[rpn_fn(nullable)]
 #[inline]
 pub fn month(t: Option<&DateTime>) -> Result<Option<Int>> {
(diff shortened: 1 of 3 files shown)


components/tidb_query_expr/src/impl_time.rs
```diff
@@ -218,6 +219,94 @@ pub fn make_date(ctx: &mut EvalContext, year: &Int, day: &Int) -> Result<Option<
...
 
+pub fn make_time(
+    ctx: &mut EvalContext,
+    hour: &Int,
+    minute: &Int,
+    second: &Real,
+    fsp: i8,
+    isunsinged: bool,
+) -> Result<Option<Duration>> {
+    let mut hour = *hour;
```
andylokandy, line 230: You can reuse [`Duration::new_from_parts()`](https://tikv.github.io/doc/tidb_query_datatype/codec/mysql/duration/struct.Duration.html#method.new_from_parts)

components/tidb_query_expr/src/lib.rs
```diff
@@ -599,6 +619,7 @@ fn map_expr_node_to_rpn_func(expr: &Expr) -> Result<RpnFnMeta> {
         ScalarFuncSig::PeriodDiff => period_diff_fn_meta(),
         ScalarFuncSig::LastDay => last_day_fn_meta(),
         ScalarFuncSig::AddDurationAndDuration => add_duration_and_duration_fn_meta(),
+        ScalarFuncSig::MakeTime => map_make_time_func(value, children)?,
```
andylokandy, line 622: I think the first argument is always signed

components/tidb_query_expr/src/impl_time.rs
```diff
@@ -218,6 +219,94 @@ pub fn make_date(ctx: &mut EvalContext, year: &Int, day: &Int) -> Result<Option<
     Ok(Some(ret))
 }
 
+pub fn make_time(
```
andylokandy, line 222: You can reuse [`Duration::new_from_parts` ](https://tikv.github.io/doc/tidb_query_datatype/codec/mysql/duration/struct.Duration.html#method.new_from_parts)

components/tidb_query_expr/src/lib.rs
Saigyouji-Yuyuko: OK，but this func in TiDB has judged it

Saigyouji-Yuyuko: > I think the first argument is always signed

OK，but this func in TiDB has judged it

andylokandy: Could you tell me which line?

Saigyouji-Yuyuko: > Could you tell me which line?

[https://github.com/pingcap/tidb/blob/master/expression/builtin_time.go#L5838](url)
`	dur, err = b.makeTime(hour, minute, second, mysql.HasUnsignedFlag(b.args[0].GetType().Flag))
`

andylokandy: Interesting~ Anyway, you can capture `args`, which has type  `&[crate::RpnStackNode<'_>]`, so only one rpn fn is sufficient.

components/tidb_query_expr/src/impl_time.rs
andylokandy: This can be used in the last line to replace the parse().

components/tidb_query_expr/src/lib.rs
Saigyouji-Yuyuko: > Interesting~ Anyway, you can capture `args`, which has type `&[crate::RpnStackNode<'_>]`, so only one rpn fn is sufficient.

I see...maybe I will push new code after get off the work.

components/tidb_query_expr/src/impl_time.rs
Saigyouji-Yuyuko: > This can be used in the last line to replace the parse().

I see

andylokandy: `MAX_HOUR_PART` and `MAX_MININTE_PART` are also useful. (https://tikv.github.io/doc/tidb_query_datatype/codec/mysql/duration/index.html)

Saigyouji-Yuyuko: > You can reuse [`Duration::new_from_parts` ](https://tikv.github.io/doc/tidb_query_datatype/codec/mysql/duration/struct.Duration.html#method.new_from_parts)

I see ,and we should handle error without ctx? or by self?

andylokandy: If I'm right(probably not), this function is infallible? Since all overflows are caught.

andylokandy: I mean, you may keep the overflow checked, and call new_from_parts().unwrap().

andylokandy:
I wonder if we can shorten the function into this:

```rust
#[rpn_fn(capture = [extra,args])]
#[inline]
pub fn make_time(
    extra: &RpnFnCallExtra,
    args: &[crate::RpnStackNode<'_>],
    hour: &Int,
    minute: &Int,
    second: &Real,
) -> Result<Option<Duration>> {
    if second.is_nan() {
        return Ok(None);
    }

    let is_unsigned = match args[0].field_type()
        .as_accessor()
        .flag()
        .contains(FieldTypeFlag::UNSIGNED); 
    let (is_neg, hour): (bool, u32) = if is_unsigned {
        (false, hour.try_into()?)
    } else {
        (hour < 0, hour.abs().try_into()?)
    };

    let mut minute: u32 = minute.try_into()?;
    let mut second: u32 = second.into_inner().trunc().try_into()?;
    let mut nano: u32 = (second.into_inner().fract() * (NANOS_PER_SEC as f64)),try_into()?;
    
    if hour > MAX_HOUR_PART {
        minute = MAX_MINUTE_PART;
        second = MAX_SECOND_PART;
        nano = 0;
    }

    match Duration::new_from_parts(
        neg,
        hour,
        minute,
        second,
        nanosecond,
        extra.ret_field_type.get_decimal() as i8,
    ) {
        Ok(duration) => Ok(Some(duration)),
        Err(_) => Ok(None)),
    }
}
```

Saigyouji-Yuyuko:
> I wonder if we can shorten the function into this:
> 
> ```rust
> #[rpn_fn(capture = [extra,args])]
> #[inline]
> pub fn make_time(
>     extra: &RpnFnCallExtra,
>     args: &[crate::RpnStackNode<'_>],
>     hour: &Int,
>     minute: &Int,
>     second: &Real,
> ) -> Result<Option<Duration>> {
>     if second.is_nan() {
>         return Ok(None);
>     }
> 
>     let is_unsigned = match args[0].field_type()
>         .as_accessor()
>         .flag()
>         .contains(FieldTypeFlag::UNSIGNED); 
>     let (is_neg, mut hour): (bool, u32) = if is_unsigned {
>         (false, hour.try_into()?)
>     } else {
>         (hour < 0, hour.abs().try_into()?)
>     };
> 
>     let mut minute: u32 = minute.try_into()?;
>     let mut second: u32 = second.into_inner().trunc().try_into()?;
>     let mut nano: u32 = (second.into_inner().fract() * (NANOS_PER_SEC as f64)).try_into()?;
>     
>     if hour > MAX_HOUR_PART {
>         hour = MAX_HOUR_PART;
>         minute = MAX_MINUTE_PART;
>         second = MAX_SECOND_PART;
>         nano = 0;
>     }
> 
>     match Duration::new_from_parts(
>         neg,
>         hour,
>         minute,
>         second,
>         nano,
>         extra.ret_field_type.get_decimal() as i8,
>     ) {
>         Ok(duration) => Ok(Some(duration)),
>         Err(_) => Ok(None)),
>     }
> }
> ```

maybe something I don't think so，second,minute may be negative,it should be Ok(None), and hour.abs() will be error if hour is i64::MIN

Saigyouji-Yuyuko:
and limition is 838:59:59.000000 ,if hour is 858,minute is 59 ,second is 59 ,that nano should be 0.

Saigyouji-Yuyuko:
And if hour is 1000，second is 70，answer should be Ok(None)，and in you code may be wrong

andylokandy:
> second,minute may be negative
try_into can catch this

> hour.abs() will be error if hour is i64::MIN
good catch! What about `hour.try_into::<u32>()?.checked_abs()?`

> and limition is 838:59:59.000000 ,if hour is 858,minute is 59 ,second is 59 ,that nano should be 0.
I think this check should be in `Duration::from_new_parts`

> And if hour is 1000，second is 70，answer should be Ok(None)，and in you code may be wrong
`Duration::from_new_parts` will return `Err(_)` in this case

Instead of error checking everywhere, I think we can wrap all of this into a `try` block. What do you think?

Saigyouji-Yuyuko:
> > And if the hour is 1000，second is 70，answer should be Ok(None)，and in you code may be wrong
> 
> 
> 
> `Duration::new_from_parts` will return `Err(_)` in this case





No，you set second to maxsecond because hour is bigger than max hour.So I think check should before call new_from_part

andylokandy:
> > > And if the hour is 1000，second is 70，answer should be Ok(None)，and in you code may be wrong?
> > 
> > 
> > 
> > `Duration::new_from_parts` will return `Err(_)` in this case
>
>
>
>
>
> No，you set second to maxsecond because hour is bigger than max hour.So I think check should before call > new_from_part

Thank you! I didn't realize that so many corner cases there could be. So what about add `check_minute_part()` and `check_second_part` before `if hour > MAX_HOUR_PART`?

ps: I'm not mean to force you to stick to my code style. I just think the state machine style, like the `overflow` flag and lots of ifs, is pretty hard for others to understand the real purpose, and is hard to see all special treatments to corner cases.

andylokandy:
> > and limition is 838:59:59.000000 ,if hour is 858,minute is 59 ,second is 59 ,that nano should be 0.

> I think this check should be added to Duration::from_new_parts

I got your idea. This is still necessary:

```rust
    if hour == MAX_HOUR_PART
        && minute == MAX_MINUTE_PART
        && second == MAX_SECOND_PART
    {
        nano = 0;
    }
```

components/tidb_query_expr/src/impl_time.rs
```diff
@@ -218,6 +224,78 @@ pub fn make_date(ctx: &mut EvalContext, year: &Int, day: &Int) -> Result<Option<
...
 }
 
+#[rpn_fn(capture = [extra,args])]
+#[inline]
+pub fn make_time(
+    extra: &RpnFnCallExtra,
+    args: &[crate::RpnStackNode<'_>],
+    hour: &Int,
+    minute: &Int,
+    second: &Real,
```
zhongzc, line 234: `Real` is alias to `NotNan`. It cannot be nan.

Saigyouji-Yuyuko:
fsp 

> ps: I'm not mean to force you to stick to my code style. I just think the state machine style, like the `overflow` flag and lots of ifs, is pretty hard for others to understand the real purpose, and is hard to see all special treatments to corner cases.

to be honest.I agree most of your advices about shorten and cleaning the code.I 'm a fresh man in Rust now and learn much through here.I think we may need some help function to help the code look good.
ps. Duration::new_from_parts will return error if fsp not in range 0..MAX_FSP,I don't think it need return Ok(None)

components/tidb_query_expr/src/impl_time.rs
Saigyouji-Yuyuko: > `Real` is alias to `NotNan`. It cannot be nan.

I see,I call it because vscode notify me it has this function😂

zhongzc:
`is_unsigned` can be one of the helper functions:

```rust
let is_hour_signed = !args[0].field_type().is_unsigned();
let is_minute_signed = !args[1].field_type().is_unsigned();
let is_second_signed = !args[1].field_type().is_unsigned();
```

andylokandy:
> ```rust
> is_minute_signed
> ```

I think the negative minute and second are invalid, so no need to check them.

zhongzc:
```rust
#[rpn_fn(capture = [extra, args])]
#[inline]
pub fn make_time(
    extra: &RpnFnCallExtra,
    args: &[crate::RpnStackNode<'_>],
    hour: &Int,
    minute: &Int,
    second: &Real,
) -> Result<Option<Duration>> {
    let (is_negative, mut hour) = if args[0].field_type().is_unsigned() {
        (false, *hour as u64)
    } else {
        (hour.is_negative(), hour.unsigned_abs())
    };

    // Filter out the number that is negative or greater than MAX_MINUTE_PART.
    let mut minute: u32 = match (*minute).try_into().ok().filter(|m| *m <= MAX_MINUTE_PART) {
        Some(minute) => minute,
        None => return Ok(None),
    };

    let mut nanosecond = (second.fract().abs() * NANOS_PER_SEC as f64) as u32;

    // Filter out the number that is negative or greater than MAX_SECOND_PART.
    let mut second: u32 = match (second.trunc() as i64)
        .try_into()
        .ok()
        .filter(|s| *s <= MAX_SECOND_PART)
    {
        Some(second) => second,
        None => return Ok(None),
    };

    let is_overflow = (hour, minute, second, nanosecond)
        > (MAX_HOUR_PART as _, MAX_MINUTE_PART, MAX_SECOND_PART, 0);
    if is_overflow {
        hour = MAX_HOUR_PART as _;
        minute = MAX_MINUTE_PART;
        second = MAX_SECOND_PART;
        nanosecond = 0;
    }

    match Duration::new_from_parts(
        is_negative,
        hour as _,
        minute,
        second,
        nanosecond,
        extra.ret_field_type.get_decimal() as i8,
    ) {
        Ok(duration) => Ok(Some(duration)),
        Err(err) => Err(err.into()),
    }
}
```
maybe more clear?

Saigyouji-Yuyuko:
@zhongzc I can't agree any more,but unsigned_abs() looks like not in rust,I will modify code by yours.thx

zhongzc:
> @zhongzc I can't agree any more,but unsigned_abs() looks like not in rust,I will modify code by yours.thx

The feature flag `#![feature(unsigned_abs)]` is needed.

Saigyouji-Yuyuko:
> > @zhongzc I can't agree any more,but unsigned_abs() looks like not in rust,I will modify code by yours.thx
> 
> The feature flag `#![feature(unsigned_abs)]` is needed.

maybe we could use saturating_abs()?

zhongzc:
> maybe we could use saturating_abs()?

In fact, `unsigned_abs` is a wrapper of `iX.wrapping_abs() as uX`. See: rust-lang/rust#74759

Saigyouji-Yuyuko:
> > maybe we could use saturating_abs()?
> 
> In fact, `unsigned_abs` is a wrapper of `iX.wrapping_abs() as uX`. See: rust-lang/rust#74759
> 

Maybe I don't like using unstable feature...

zhongzc:
> Maybe I don't like using unstable feature...
> Sorry it merged....，emmmm but my complier do not support it ...
> now support it

No problem, `hour.wrapping_abs() as u64` is ok.

zhongzc:
I also found that I wrote an ugly code like `x.tryinto.ok.filter`. If condition as your origin choice is much better. :p

Saigyouji-Yuyuko:
> I also found that I wrote an ugly code like `x.tryinto.ok.filter`. If condition as your origin choice is much better. :p

for me,A fresh man in rust,I like this style,it looks like too rust ,not C or C++ ，I copy your code to test and push.

zhongzc:
> it looks like too rust

So sorry to make you think so. That uses a very hard and inefficient way to do a simple thing. Everyone can understand
```rust
let mut minute = if 0 <= *minute && *minute <= MAX_MINUTE_PART {
    *minute as u32
} else {
    return Ok(None);
}
```
without comments, right?

BTW, the unit tests you added are awesome! [`make_time` implementation](https://github.com/pingcap/tidb/blob/7a19c709d07d4c33009bb85bba5122b2a658505b/expression/builtin_time.go#L5787-L5813) at TiDB can't even pass the `(838, 58, 59.1, "838:58:59.1")` case. There also need help...

Saigyouji-Yuyuko:
> > it looks like too rust
> 
> So sorry to make you think so. That uses a very hard and inefficient way to do a simple thing. Everyone can understand
> 
> ```rust
> let mut minute = if 0 <= *minute && *minute <= MAX_MINUTE_PART {
>     *minute as u32
> } else {
>     return Ok(None);
> };
> ```
> 
> without comments, right?
> 
> BTW, the unit tests you added are awesome! [`make_time` implementation](https://github.com/pingcap/tidb/blob/7a19c709d07d4c33009bb85bba5122b2a658505b/expression/builtin_time.go#L5787-L5813) at TiDB can't even pass the `(838, 58, 59.1, "838:58:59.1")` case. There also need help...

maybe to a fresh man，this sytle is very interesting.
I also saw the code in TiDB, I think it only check hour ,not check minute ,than modify second is not correct.I was going to pr after this pr merged.

zhongzc (approved):
LGTM

andylokandy:
LGTM

ti-srebot (approved):

Saigyouji-Yuyuko:
wow ,something wrong I guess

PR #8513, opened 2020-08-26
txn: Move MvccTxn::cleanup & MvccTxn::rollback to command

longfangsong:
Signed-off-by: longfangsong <longfangsong@icloud.com>

### What problem does this PR solve?

Problem Summary: 

Currently MvccTxn contains several functions highly related with certain Command, this breaks the SRP of `MvccTxn` and made the struct's code very long and hard to read.

(This is part of #8432.)

### What is changed and how it works?

What's Changed:

Move `cleanup` and `rollback` from `MvccTxn` to its own Command, and also move the tests.

### Check List 

Tests 

- Unit test

PR #9089, opened 2020-11-23
txn: Move check_data_constraint to action module

longfangsong:

Problem Summary:
As I proposed in #8513, we can move these "action" function and related tests to txn/actions.rs, and make the code cleaner.

Though `check_data_constraint` is not used by commands, its also a "high level" operation which is made up with multiple basic operations in MvccTxn, so it should be also an action

What's Changed:
Move `MvccTxn::check_data_constraint` and related tests and test utils to `action` mod.

### Check List 

Tests 

- Unit test (I added a real unit test for it)
- Integration test
Side effects

### DIFF
--- a/src/storage/txn/actions/check_data_constraint.rs
+++ b/src/storage/txn/actions/check_data_constraint.rs
@@ -23,3 +23,98 @@ pub(crate) fn check_data_constraint<S: Snapshot>(
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::storage::mvcc::tests::write;
+    use crate::storage::mvcc::{ErrorInner, MvccTxn, Result as MvccResult};
+    use crate::storage::txn::actions::check_data_constraint::check_data_constraint;
+    use crate::storage::{Engine, TestEngineBuilder};
+    use concurrency_manager::ConcurrencyManager;
+    use kvproto::kvrpcpb::Context;
+    use txn_types::{Key, TimeStamp, Write, WriteType};
+
+    #[test]
+    fn test_check_data_constraint() {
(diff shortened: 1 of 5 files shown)


longfangsong:
@nrc PTAL

src/storage/txn/actions/check_data_constraint.rs
```diff
@@ -0,0 +1,91 @@
...
+                write: Write::new(WriteType::Put, TimeStamp::new(10), None),
+                write_commit_ts: TimeStamp::new(15),
+                key: Key::from_raw(b"a"),
+            },
+        ];
+
+        for case in cases {
+            let Case { expected, should_not_exist, write, write_commit_ts, key } = case;
+            let result = check_data_constraint(&mut txn, should_not_exist, &write, write_commit_ts, &key);
+            assert_eq!(format!("{:?}", expected), format!("{:?}", result));
```
longfangsong, line 88: Should all error types derive `Eq` so we can `assert_eq!(expected, result)` directly here?

nrc: We could try, maybe it is difficult?

nrc (approved):
lgtm

nrc:
PTAL @youjiali1995

youjiali1995 (commented):
lgtm

ti-srebot (approved):

src/storage/txn/actions/check_data_constraint.rs
longfangsong: There are many Error types in the std library which don't impl `Eq`, which made `#[derive(Eq, PartialEq)]` directly impossible.
I'll have a try to `impl Eq` on `mvcc::ErrorInner` we used here manually.

PR #9105, opened 2020-11-25
sst_importer: fix bug that cause tikv panic when ingest sst file

glorv:

Problem Summary:
Fix the bug that in some case, TiKV can't restart successfully due to ingesting sst file error

See https://github.com/pingcap/tidb-lightning/issues/488 for more details.

What's Changed:
- Check file on disk and in key manager separately because in some case, there are remain key info in key manager
- Sync the file and directory after copy sst file to the clone directory

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
  - The new added fail-point tests will fail stably before this pr.
- No code

Side effects

### Release note 
- No Release Note


kennytm:
LGTM

components/sst_importer/src/util.rs
```diff
@@ -50,11 +52,14 @@ pub fn prepare_sst_for_ingestion<P: AsRef<Path>, Q: AsRef<Path>>(
         // RocksDB must not have this file, we can make a hard link.
         fs::hard_link(path, clone)
             .map_err(|e| format!("link from {} to {}: {:?}", path, clone, e))?;
+        File::open(clone)?.sync_all()?;
```
yiwu-arbug, line 55: I guess this is not needed.

components/sst_importer/src/util.rs
```diff
@@ -50,11 +52,14 @@ pub fn prepare_sst_for_ingestion<P: AsRef<Path>, Q: AsRef<Path>>(
...
     } else {
         // RocksDB may have this file, we should make a copy.
         copy_and_sync(path, clone)
             .map_err(|e| format!("copy from {} to {}: {:?}", path, clone, e))?;
     }
+    // sync clone dir
+    File::open(Path::new(clone).parent().unwrap())?.sync_all()?;
```
yiwu-arbug, line 62: return error instead of unwrap?

glorv: Since clone is a valid file path in previous options, the parent() call to parse its directory should not fail here

components/sst_importer/src/util.rs
glorv: @Little-Wallace PTAL

components/sst_importer/src/util.rs
yiwu-arbug: yeah, but I think we can avoid use of unwrap after all.

yiwu-arbug (approved):
LGTM

ti-srebot (approved):

components/sst_importer/src/util.rs
Little-Wallace: Why? Link may be also sync once?

Little-Wallace (approved):
LGTM

PR #9106, opened 2020-11-25
copr: Add enum support for codegen

Xuanwo:

Add enum support in codegen, after this PR merged, we can implement copr functions like:

```rust
#[rpn_fn]
fn cast_enum_as_int(val: EnumRef) -> Result<Option<Int>> {
    Ok(Some(val.value() as i64))
}
```

### Check List

Tests

- Unit test
- Integration test


skyzh (approved):
LGTM

skyzh:
Would you please take a look? @andylokandy

Xuanwo:
merged into #9133

PR #8855, opened 2020-10-20
raftstore: support fsync control

gengliqi:
Signed-off-by: gengliqi <gengliqiii@gmail.com>

### What problem does this PR solve?
Improve TiKV performance on non-NVMe disk.

Problem Summary:

- The 'sync' performance is varied on different types of disk, it could from several hundred to 50K times per sec, we could use fio -fdatasync or pg_test_sync to profile the 'sync' performance of a disk
- Typically, a non-NVMe disk could perform 2K~5K sync per sec
- When TiKV runs under proper stress (lots of write, disk latency is good), the 'sync' frequency could reach 20K as we discovered on grafana panel
So, the 'sync' performance will be one of the bottlenecks of TiKV performance, especially under some types of disk.
This PR provide a config entry to control the raftstore's 'sync' frequency.

### What is changed and how it works?

What's Changed:
Use async ready https://github.com/tikv/raft-rs/pull/403 to support this feature.
The sync policy comes from https://github.com/innerr/tikv/pull/4.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Integration test
- Manual test (add detailed scripts or steps below)

Side effects

- These changes could delay the committed logs' applying, and cause greater latency, so we provide a config entry to enable/disable it when running TiKV on different types of disk.

### Release note 
- Control sync frequency using config entry: raftstore.delay-sync-ns, performance on some types of disk could be better with this.

PR #8847, opened 2020-10-18
raftstore: remove the overhead of traversing new committed entries

gengliqi:
Signed-off-by: gengliqi <gengliqiii@gmail.com>

### What problem does this PR solve?

Problem Summary:
https://github.com/tikv/tikv/pull/8056 fixed a stale read bug, which has an overhead of traversing the new committed entries.
Now we can remove this overhead via epoch checker which was introduced by https://github.com/tikv/tikv/pull/8441.

### What is changed and how it works?

What's Changed:
Removing the traversing overhead.

### Related changes

- Need to cherry-pick to the release branch

Tests 

- Unit test

Side effects

- No

### Release note 
- No Release Note

PR #8662, opened 2020-09-14
raftstore: make apply log and read process more early

gengliqi:
Signed-off-by: gengliqi <gengliqiii@gmail.com>

### What problem does this PR solve?

Make apply log and read process more early.

Problem Summary:

The time of waiting for batching is useless for applying existing logs and handling read.
So move these two processes to the place after getting `Ready` from raft-rs.

### What is changed and how it works?

What's Changed:
Described above.

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- No code

Side effects

- No

### Release note 
* No Release Note

PR #8631, opened 2020-09-09
Remove sync-log config option

hunterlxt:
Signed-off-by: Xintao <hunterlxt@live.com>

### What problem does this PR solve?

If `sync-log` is set to `false`, there is chance that data will be loss.

### What is changed and how it works?

Make this option true always.

### Related changes

- Need to cherry-pick to the release branch

### Release note 
- Set sync-log true always.

PR #9112, opened 2020-11-25
raftstore: support fsync control in release-4.0

gengliqi:

Note that we decide to merge this feature to release-4.0 first, so the https://github.com/tikv/tikv/pull/8855 will be merged later.
Besides fsync control, this PR also cherry-picks https://github.com/tikv/tikv/pull/8662, https://github.com/tikv/tikv/pull/8847, https://github.com/tikv/tikv/pull/8631, and changes some codes.

Improve TiKV performance on non-NVMe disk.

Problem Summary:

- The 'sync' performance is varied on different types of disk, it could from several hundred to 50K times per sec, we could use fio -fdatasync or pg_test_sync to profile the 'sync' performance of a disk
- Typically, a non-NVMe disk could perform 2K~5K sync per sec
- When TiKV runs under proper stress (lots of write, disk latency is good), the 'sync' frequency could reach 20K as we discovered on grafana panel
So, the 'sync' performance will be one of the bottlenecks of TiKV performance, especially under some types of disk.
This PR provide a config entry to control the raftstore's 'sync' frequency.

What's Changed:
Use async ready https://github.com/tikv/raft-rs/pull/403 to support this feature. (This PR uses the https://github.com/tikv/raft-rs/pull/406)
The sync policy comes from https://github.com/innerr/tikv/pull/4.

### Related changes

### Check List 

Tests 

- Integration test

Side effects

- These changes could delay the committed logs' applying, and cause greater latency, so we provide a config entry to enable/disable it when running TiKV on different types of disk.

### Release note 
- Control sync frequency using config entry: raftstore.delay-sync-ns, performance on some types of disk could be better with this.


tests/failpoints/cases/test_early_apply.rs
```diff
@@ -1,92 +0,0 @@
...
-use raft::eraftpb::MessageType;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
-use std::thread;
-use std::time::*;
-use test_raftstore::*;
-
-/// Tests early apply is disabled for singleton.
-#[test]
-fn test_singleton_early_apply() {
```
BusyJay, line 12: I think this test is still necessary.

src/server/status_server/region_meta.rs
```diff
@@ -208,7 +207,6 @@ impl RegionMeta {
             raft_status: raft_group.status().into(),
             raft_apply: RaftApplyState {
                 applied_index: apply_state.get_applied_index(),
-                last_commit_index: apply_state.get_last_commit_index(),
```
BusyJay, line 211: Like I commented in the doc, I think we can set it to `min(commit_index, persisted_index)` to do the check.

components/raftstore/src/store/peer.rs
```diff
@@ -1172,8 +1256,6 @@ impl Peer {
         // could get a newer value, and after that, the leader may read a stale value,
         // which violates linearizability.
         self.get_store().applied_index() >= read_index
-            && !self.is_splitting()
-            && !self.is_merging()
```
NingLin-P, line 1176: This should not be removed!

gengliqi: They are useless. `is_splitting` and `is_merging` should only be used on leader for its lease.
Actually, the `last_committed_split_idx` and `last_committed_prepare_merge_idx` are maintained only on leader all the time.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2613,27 +2589,6 @@ impl ApplyFsm {
 
         self.delegate.metrics = ApplyMetrics::default();
         self.delegate.term = apply.term;
-        let prev_state = (
```
BusyJay, line 2616: The check can be reserved. And you only need to set commit index of `apply_state` using `entries.last().unwrap().get_index()`.

components/raftstore/src/store/peer.rs
gengliqi: The `pending_merge_state` in `is_merging` is important for follower read. I recover this code.

components/raftstore/src/store/fsm/apply.rs
gengliqi: Done.

Cargo.lock
```diff
@@ -3073,7 +3073,7 @@ dependencies = [
 [[package]]
 name = "raft"
 version = "0.6.0-alpha"
-source = "git+https://github.com/pingcap/raft-rs?branch=0.6.x#b5f5830f889f31453590e85613a9bf23d0e2fea9"
+source = "git+https://github.com/gengliqi/raft-rs?branch=fix-clippy#f08f3f537bd55546ee04f972958d6dced59cd552"
```
BusyJay, line 3076: You can update it now.

components/batch-system/src/batch.rs
```diff
@@ -256,9 +260,14 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
...
         if batch.is_empty() {
-            self.handler.pause();
-            if let Ok(fsm) = self.fsm_receiver.recv() {
-                return batch.push(fsm);
+            loop {
+                if self.handler.pause() {
+                    if let Ok(fsm) = self.fsm_receiver.recv() {
+                        return batch.push(fsm);
+                    }
+                } else if let Ok(fsm) = self.fsm_receiver.recv_timeout(Duration::from_micros(100)) {
```
BusyJay, line 268: This looks like magic. Why should it wait for 100 micro seconds if handler is unable to pause? I think this is more related to raftstore instead of batch system itself.

components/raftstore/src/store/config.rs
```diff
@@ -29,8 +29,9 @@ with_prefix!(prefix_store "store-");
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct Config {
-    // true for high reliability, prevent data loss when power failure.
-    pub sync_log: bool,
+    // delay time of raft db sync (us).
+    #[config(skip)]
```
BusyJay, line 33: Why skip?

components/raftstore/src/store/config.rs
```diff
@@ -29,8 +29,9 @@ with_prefix!(prefix_store "store-");
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct Config {
-    // true for high reliability, prevent data loss when power failure.
-    pub sync_log: bool,
```
BusyJay, line 33: This breaks backward compatibility. Cluster that disabled `sync-log` will enable it again and cause high latency after upgrading.

components/raftstore/src/store/config.rs
gengliqi: Because it can not be dynamically updated.

components/batch-system/src/batch.rs
gengliqi: Because some work has not finished yet. Maybe the batch system needs a more general function to meet this need?

components/raftstore/src/store/config.rs
gengliqi: Yes, it's on purpose. Do you have a better idea? Keep this config?

BusyJay: I suggest to keep it on release-4.0 and remove it on master. After we think the feature is stable, we can enable fsync control when sync-log is set to false.

/cc @innerr @zhangjinpeng1987

components/batch-system/src/batch.rs
BusyJay: If work has not finished, it should handle it immediately, no reason to wait.

components/raftstore/src/store/config.rs
gengliqi: 🤔Maybe it can. If users enable fsync control(set delay-sync-us not 0), it can be dynamically updated to other values except 0. If users want to change it to 0 dynamically, we should do more work to make it correct.
For example, https://github.com/gengliqi/tikv/blob/fsync-control-4.0/components/raftstore/src/store/fsm/store.rs#L605-L614, we can consider it is synced when the write batch is empty and fsync control is disabled.

components/batch-system/src/batch.rs
innerr: This (need) comes from a special type of task:
1. It needs time to finish.
2. It doesn't need (a full on core of) CPU resource to run the job.

In here, the special task is `waiting for deadline, then do fsync`.

If it's a common task, we will spend the whole thread to do that job, but since the `waiting for deadline` part no need to use CPU resource, so we put the thread into sleep (by wait_with_timeout) to reserve CPU resource as an optimization, then allow the thread wake up periodly to continue its job.

I think this need and the code here is genetic, not raftstore-specified.

hi-rustin:
This cherry pick PR is for a release branch and has not yet been approved by release team.
Adding the do-not-merge/cherry-pick-not-approved label.

To merge this cherry pick, it must first be approved (/lgtm + /merge) by the collaborators.

AFTER it has been approved by collaborators, please ping the release team in a comment to request a cherry pick review

PR #9117, opened 2020-11-26
sst_importer: fix bug that cause tikv panic when ingest sst file (#9105)

ti-srebot:
cherry-pick #9105 to release-4.0

---

Problem Summary:
Fix the bug that in some case, TiKV can't restart successfully due to ingesting sst file error

See https://github.com/pingcap/tidb-lightning/issues/488 for more details.

What's Changed:
- Check file on disk and in key manager separately because in some case, there are remain key info in key manager
- Sync the file and directory after copy sst file to the clone directory

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
  - The new added fail-point tests will fail stably before this pr.
- No code

Side effects

### Release note 
- No Release Note


overvenus:
LGTM

kennytm (approved):

BusyJay (approved):

PR #9077, opened 2020-11-20
storage: return memory locks as BatchGet response level error

sticnarf:
### What problem does this PR solve?

Issue Number: #9069

### What is changed and how it works?

Set memory lock to the newly added `KeyError` field in `BatchGetResponse`. The client will skip using `pairs` if the error is used.

### Related changes

- https://github.com/pingcap/kvproto/pull/691
- https://github.com/pingcap/tidb/pull/21157

### Check List 

Tests 

- Integration test

### Release note 

- Part of the async commit feature

PR #9120, opened 2020-11-26
storage: return memory locks as ScanResponse level error

sticnarf:
- [x] Depends on the unmerged https://github.com/pingcap/kvproto/pull/697

Problem Summary:

`KvScan` interface has a similar problem with https://github.com/tikv/tikv/pull/9077. TiDB requires all the KV pairs in the range to be returned (if the limit is not reached).

Although queries from TiDB users are transformed into coprocessor queries, the `KvScan` interface is still used in multiple cases like backfilling in DDLs.

Set memory lock to the newly added KeyError field in `ScanResponse`. The client will skip using pairs if the error is used.

### Check List 

Tests 

- Integration test

### Release note 

- Part of the async commit feature

### DIFF
--- a/tests/integrations/server/kv_service.rs
+++ b/tests/integrations/server/kv_service.rs
@@ -1147,8 +1147,7 @@ fn test_txn_heart_beat() {
     );
 }
 
-#[test]
-fn test_batch_get_memory_lock() {
+fn test_with_memory_lock_cluster(f: impl FnOnce(TikvClient, Context, /* raw_key */ Vec<u8>, Lock)) {
     let (cluster, client, ctx) = must_new_cluster_and_kv_client();
     let cm = cluster.sim.read().unwrap().get_concurrency_manager(1);
     let raw_key = b"key".to_vec();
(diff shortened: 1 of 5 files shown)


tests/integrations/server/kv_service.rs
```diff
@@ -1168,11 +1167,34 @@ fn test_batch_get_memory_lock() {
...
+        req.set_context(ctx);
+        req.set_start_key(b"a".to_vec());
+        req.version = 50;
+        let resp = client.kv_scan(&req).unwrap();
+        let lock_info = lock.into_lock_info(raw_key);
+        assert_eq!(resp.pairs[0].get_error().get_locked(), &lock_info);
+        assert_eq!(resp.get_error().get_locked(), &lock_info);
+    });
 }
+
```
youjiali1995, line 1200:
```suggestion
```

youjiali1995 (commented):
lgtm

MyonKeminta (approved):

ti-srebot (approved):

PR #9133, opened 2020-11-27
copr: Add enum/set support for codegen

Xuanwo:

Add enum support in codegen, after this PR merged, we can implement copr functions like:

```rust
#[rpn_fn]
fn cast_set_as_int(val: SetRef) -> Result<Option<Int>> {
    Ok(Some(val.value() as i64))
}
```

### Check List

Tests

- Unit test
- Integration test


Xuanwo:
Let's merge this PR into previous one #9106

Xuanwo:
PTAL @skyzh @zhongzc @andylokandy

skyzh (approved):
LGTM. What about merging logic of `xxxRef`? (e.g. pack them into an enum), so as to reduce duplicated code?

Xuanwo:
> What about merging logic of xxxRef? (e.g. pack them into an enum), so as to reduce duplicated code?

Worth a try, maybe we can implement it in another PR.

Xuanwo:
ping @skyzh to take a look

skyzh (approved):
Great!

Xuanwo:
ping @andylokandy , PTAL

andylokandy:
Generally LGTM, please add a test for the macro expansion result, 'cause i think it can help new comer to figure out what exactly the macro does.

Xuanwo:
Maybe I can add a macro expansion test for `rpn_fn` in another PR? (I think this test must be very verbose).

andylokandy:
However, I think tests are tightly related to the pr itself unlike some additional improvement.

Xuanwo:
OK, I will add a test here.

Xuanwo:
@andylokandy PTAL ~

andylokandy:
Cool! It's the first codegen template for chunk.

andylokandy:
Please fix the test, and rest LGTM.

components/tidb_query_codegen/src/rpn_function.rs
```diff
@@ -2098,6 +2126,117 @@ mod tests_normal {
...
+                        result.push(foo(arg0)?);
+                    }
+                    Ok(Decimal::into_vector_value(result))
+                }
+            }
+        };
+
+        // NOTE: Test via trim all space between code
+        assert_eq!(
+            expected.to_string().replace(" ", ""),
```
andylokandy, line 2233: In this case let mut a = 1; will equal to let muta = 1;. You can simply print the result by dbg!() and paste it here.

Xuanwo: But it's hard to read - -

Xuanwo: Can I call `rustfmt-nightly` here?

andylokandy: Oh, I'd thought it would pretty print... I may mess up with other languages. Seems you could only tweak the spaces manually.. https://github.com/tikv/tikv/blob/9330514b93eea30fa3afb167a68251ffb520cb7b/components/tidb_query_codegen/src/rpn_function.rs#L2094-L2096

Xuanwo: I worked out a new way to compare token tree directly, PTAL.

components/tidb_query_codegen/src/rpn_function.rs
```diff
@@ -1691,6 +1718,49 @@ impl NormalRpnFn {
...
+                        x.to_string() == y.to_string()
+                    } else {
+                        false
+                    }
+                }
+            });
+
+        assert!(result, "expect: {}, actual: {}", &l, &r);
+
+        result
```
andylokandy, line 1761: This will always be `true`

components/tidb_query_codegen/src/rpn_function.rs
andylokandy: Awesome! That's a good approach.

components/tidb_query_codegen/src/rpn_function.rs
```diff
@@ -1758,7 +1758,7 @@ mod tests_normal {
 
         assert!(result, "expect: {}, actual: {}", &l, &r);
 
-        result
+        true
```
andylokandy, line 1761: So no need to return it, I think.

Xuanwo: OK, I got it.

andylokandy:
If an ident in a big token tree goes wrong, it'll be hard to locate. I suggest renaming to `token_tree_equal`, to return the bool (sorry for the last comment) and assert it in the test.

Xuanwo:
> If an ident in a big token tree goes wrong, it'll be hard to locate. I suggest renaming to `token_tree_equal`, to return the bool (sorry for the last comment) and assert it in the test.

I will look into this to find a solution for macro debuging.

Xuanwo:
Debug message looks like following: 

```
expect: TokenStream [Ident { sym: row_index }, Punct { char: ',', spacing: Alone }, Ident { sym: vaxl }], actual: TokenStream [Ident { sym: row_index }, Punct { char: ',', spacing: Alone }, Ident { sym: val }]
thread 'rpn_function::tests_normal::test_enum_fn_generate_real_fn_trait_impl' panicked at 'expect: TokenStream [Ident { sym: row_index }, Punct { char: ',', spacing: Alone }, Ident { sym: vaxl }], actual: TokenStream [Ident { sym: row_index }, Punct { char: ',', spacing: Alone }, Ident { sym: val }]', components/tidb_query_codegen/src/rpn_function.rs:1747:9
```

Maybe this is enough? I found it's hard to get the token's location via public APIs.

andylokandy (approved):
LGTM, well done.

ti-srebot (approved):

PR #9136, opened 2020-11-27
raftstore: notify stale callbacks before applying

### LINKED ISSUE #9070: committed_cb is always invoked which affects the correctness of async apply

master

not releated

### Steps to reproduce

1. Partition the leader and followers.
2. Propose a write request on the leader.
3. A new leader is elected and commits a new entry.
4. Recover the network partition.
5. The old proposal should be dropped and the committed_cb shouldn't be invoked.

### Root cause
All proposals are sent to the apply system even if some of them aren't committed. We can filter them.
image

NingLin-P:

Issue Number: close #9070 

Problem Summary:

Async commit need to invoke the callbacks of committed entries in the store thread, but some callbacks may belong to entries that can't be committed

Filter and notify these stale callbacks

Tests 

- Unit test

### Release note 
- No release note


components/raftstore/src/store/peer.rs
```diff
@@ -79,46 +79,64 @@ struct ProposalQueue<S>
...
+                if p.index == index {
+                    return if p.cb.is_none() { None } else { Some(p) };
+                } else {
+                    panic!(
+                        "{} unexpected callback at term {}, found index {}, expected {}",
+                        self.tag, term, p.index, index
+                    );
+                }
+            } else {
+                apply::notify_stale_req(current_term, p.cb);
```
BusyJay, line 132: I don't think it's a good idea. It can cause requests failure during transferring leader.

components/raftstore/src/store/peer.rs
```diff
@@ -1721,24 +1739,32 @@ where
                     self.raft_group.skip_bcast_commit(true);
                     self.last_urgent_proposal_idx = u64::MAX;
                 }
+                let current_term = self.term();
+                let cbs = committed_entries
+                    .iter()
+                    .filter_map(|e| {
```
BusyJay, line 1745: How about using `retain`?

components/raftstore/src/store/peer.rs
NingLin-P: `find_proposal` is called for each committed entries one by one in order, since `term` is increasing among log entries, proposals enter this branch is impossible to commit (same behavior as before).

components/raftstore/src/store/peer.rs
NingLin-P: The purpose is to find `Proposal` for each committed entry, not changing `committed_entries`

components/raftstore/src/store/peer.rs
```diff
@@ -1721,24 +1739,32 @@ where
                     self.raft_group.skip_bcast_commit(true);
                     self.last_urgent_proposal_idx = u64::MAX;
                 }
+                let current_term = self.term();
+                let cbs = committed_entries
```
BusyJay, line 1743: How about checking if `self.proposals` is empty first?

components/raftstore/src/store/peer.rs
```diff
@@ -2103,7 +2129,8 @@ where
                 }
                 self.should_wake_up = true;
                 let p = Proposal {
-                    is_conf_change: req_admin_cmd_type == Some(AdminCmdType::ChangePeer),
+                    is_conf_change: req_admin_cmd_type == Some(AdminCmdType::ChangePeer)
+                        || req_admin_cmd_type == Some(AdminCmdType::ChangePeerV2),
```
BusyJay, line 2133: Is it a bug?

BusyJay (commented):
Any test case?

components/raftstore/src/store/peer.rs
NingLin-P: Not, it is a mistake but not a bug, because we use `Callback::None` as the callbacks for `ChangePeer` cmds and we will not send these callbacks to the apply thread. Also, if these callbacks did send to the apply thread, some test cases will not pass.

https://github.com/tikv/tikv/blob/9b8785a42f6e03621c0d0ce66799761b3e00f9cc/components/raftstore/src/store/worker/pd.rs#L858

https://github.com/tikv/tikv/blob/9b8785a42f6e03621c0d0ce66799761b3e00f9cc/components/raftstore/src/store/peer.rs#L113-L115

components/raftstore/src/store/peer.rs
NingLin-P: It is checked at

https://github.com/tikv/tikv/blob/9b8785a42f6e03621c0d0ce66799761b3e00f9cc/components/raftstore/src/store/peer.rs#L1717

BusyJay:
I think you need a case to verify the problem mentioned in #9070 is solved.

NingLin-P:
> I think you need a case to verify the problem mentioned in #9070 is solved.

Setting the callback at integration test is quite complicated, I add a unit test instead, PTAL

BusyJay (commented):
LGTM

NingLin-P:
PTAL @MyonKeminta @gengliqi

youjiali1995 (commented):
/LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -1794,7 +1827,6 @@ where
         // needs to be sent to the apply system.
         // Always sending snapshot task behind apply task, so it gets latest
         // snapshot.
-        // TODO: maybe we should move this code to other place to make the logic more clear.
```
gengliqi, line 1797: This TODO work has not been done yet. BTW, I find a bug that the snapshot may not be generated if there is not new committed entry. So the code below should be moved outside. Furthermore, it is unnecessary to send the snapshot task to apply fsm. I will fix them later.

gengliqi (commented):
Rest LGTM

components/raftstore/src/store/peer.rs
NingLin-P: I revert this change, PTAL again

gengliqi (approved):

PR #9146, opened 2020-11-30
copr: Add ENUM support for aggr MAX

Xuanwo:

Add ENUM support for aggr MAX

- Implement a new aggr for ENUM: AggFnExtremumForEnum

### Check List

Tests
- Unit test
- Integration test


skyzh (approved):

Xuanwo:
cc @andylokandy , can you take a look on this PR?

components/tidb_query_datatype/src/codec/mysql/enums.rs
```diff
@@ -91,6 +84,19 @@ impl<'a> EnumRef<'a> {
     }
 }
 
+impl<'a> ToString for EnumRef<'a> {
+    fn to_string(&self) -> String {
+        if self.value == 0 {
+            return String::new();
+        }
+
+        let buf = &self.data[self.value - 1];
```
zhongzc, line 93: Is the `value` always valid? Or allow panic here?

components/tidb_query_aggr/src/impl_max_min.rs
```diff
@@ -186,6 +186,94 @@ where
...
+        let extreme_ref = self
+            .extremum
+            .as_ref()
+            .map(|x| EnumRef::from_owned_value(unsafe { std::mem::transmute(x) }));
+
+        if value.is_some()
+            && (self.extremum.is_none()
+                || extreme_ref
+                    .unwrap()
+                    .to_string()
```
zhongzc, line 251: Is it necessary to allocate string for every comparison?

components/tidb_query_datatype/src/codec/mysql/enums.rs
Xuanwo: Yes, value always valid

components/tidb_query_aggr/src/impl_max_min.rs
skyzh: This is current limitation of copr framework. We must store owned value inside aggregator. But here I think we could store `Enum` instead of `String`, to improve efficiency.

Xuanwo: Maybe we can compare the underlying bytes directly.

Should we need to care about string collections here?

skyzh: I agree. This is not necessary. Can we add some interface like `to_str()` on `EnumRef`?

Xuanwo: Let's give it a try.

skyzh: By the way, I think it's necessary to take collation into consideration here. And I suggest to implement it in separate PR.

Xuanwo: as_str() maybe better?

skyzh: Agree. We may use `as_str()` to get `&str`.

Xuanwo: > By the way, I think it's necessary to take collation into consideration here. And I suggest to implement it in separate PR.

OK, I wiil add a TODO here.

Xuanwo: One more thing, build str from `u8` without convertion requires utf-8 check, should we just panic here or use ` String::from_utf8_lossy` to convert invliad bytes to `U+FFFD`?

skyzh: I think we could return `Err` here.

Xuanwo:
@zhongzc PTAL

zhongzc (approved):
LGTM

ti-srebot (approved):

PR #9147, opened 2020-11-30
copr: Get char weight from collator

andylokandy:
Signed-off-by: Andy Lok <andylokandy@hotmail.com>

Add `fn char_weight(char: <Self::Charset as Charset>::Char) -> Self::Weight` to `Collator`

Tests 

- Unit test

### Release note 

- None


andylokandy:
cc @zhongzc @wangggong

zhongzc (changes requested):
Could you provide utests for the new method?

andylokandy:
> Could you provide utests for the new method?

I think it's covered by the test of other methods.

andylokandy:
It's a rename of original `convert_xxxx()` functions.

components/tidb_query_datatype/src/codec/collation/collator/utf8mb4_unicode_ci.rs
```diff
@@ -8,41 +8,49 @@ pub struct CollatorUtf8Mb4UnicodeCi;
 
 impl Collator for CollatorUtf8Mb4UnicodeCi {
     type Charset = CharsetUtf8mb4;
+    type Weight = u128;
```
Xuanwo, line 11: Why `CollatorUtf8Mb4Bin`, `CollatorUtf8Mb4GeneralCi` and `CollatorUtf8Mb4UnicodeCi` share the same charset but have different weight?

zhongzc: Weight is a sort of "sort key" in terms of collation. Do you have a better alternative name?

Xuanwo: After reading http://www.unicode.org/reports/tr10/#Weight_Level_Defn , I think `weight` is a good name. 

Maybe we can add a reference here?

Xuanwo (commented):
LGTM

zhongzc (approved):

wjhuang2016 (commented):
Why do we need to add weight() to collator?

andylokandy:
> Why do we need to add weight() to collator?

Some collation-aware function may want to use it.

wjhuang2016:
> > Why do we need to add weight() to collator?
> 
> Some collation-aware function may want to use it.

Any example? From the perspective of colation, the weight of a character does not always have a meaning.

andylokandy:
> Any example? From the perspective of colation, the weight of a character does not always have a meaning.

For example, we can use it to refactor `like` in

https://github.com/tikv/tikv/blob/9b8785a42f6e03621c0d0ce66799761b3e00f9cc/components/tidb_query_expr/src/impl_like.rs#L48

wjhuang2016:
> > Any example? From the perspective of colation, the weight of a character does not always have a meaning.
> 
> For example, we can use it to refactor `like`:
> 
> https://github.com/tikv/tikv/blob/9b8785a42f6e03621c0d0ce66799761b3e00f9cc/components/tidb_query_expr/src/impl_like.rs#L48
> 
> It's currently very inefficient and may contain bugs because `sort_compare()` trims space paddings.

Okey.

skyzh (approved):

ti-srebot (approved):

xiongjiwei:
> For example, we can use it to refactor `like`:
> 
> https://github.com/tikv/tikv/blob/9b8785a42f6e03621c0d0ce66799761b3e00f9cc/components/tidb_query_expr/src/impl_like.rs#L48
> 
> It's currently very inefficient and may contain bugs because `sort_compare()` trims space paddings.

`like` function cannot use `char_weight` since `=` is different from `like`. 
see also 
https://github.com/pingcap/tidb/pull/21892
https://github.com/pingcap/tidb/pull/21893

PR #9149, opened 2020-11-30
Introduce IO rate limiter and rate limited file abstraction

tabokie:

Unable to prioritized IO requests produced from different workflows. Unable to explicitly limit IO usage.

What's Changed:

Rate limiter works as a central control to manage IO flow. IO requests issued to `InstrumentedFile` will be regulated by a global IO limiter. This PR introduces the related module interfaces.

Followup works include implementing rate limit algorithm, tagging IO issuer, and changing all existing uses of `std::fs::File` to rate limited file. They are expected to complete within the next month.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

- No release note


components/tikv_util/src/file.rs
```diff
@@ -134,6 +137,232 @@ impl<R: Read> Read for Sha256Reader<R> {
...
+    }
+}
+
+/// A wrapper around `std::fs::File` with capability to track and regulate IO flow.
+pub struct InstrumentedFile {
+    inner: fs::File,
+    limiter: Option<Arc<IORateLimiter>>,
+}
+
+impl Deref for InstrumentedFile {
```
yiwu-arbug, line 232: When will deref be useful? I want to avoid exposing `inner`. Instead we can provide wrapper of file methods like `sync_all`.

components/tikv_util/src/file.rs
```diff
@@ -134,6 +137,232 @@ impl<R: Read> Read for Sha256Reader<R> {
...
+    }
+}
+
+impl Read for InstrumentedFile {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if let Some(limiter) = &mut self.limiter {
+            let mut remains = buf.len();
+            let mut pos = 0;
+            while remains > 0 {
+                let allowed = limiter.request(get_io_type(), IOOp::Read, remains);
```
yiwu-arbug, line 278: shall we cap the size of max request size to, say, 1M? same for `write()`.

components/tikv_util/src/file.rs
```diff
@@ -134,6 +137,232 @@ impl<R: Read> Read for Sha256Reader<R> {
...
+            ErrorKind::InvalidInput,
+            "the source path is not an existing regular file",
+        ));
+    }
+
+    let mut reader = InstrumentedFile::open(from.as_ref())?;
+    let mut writer = InstrumentedFile::create(to.as_ref())?;
+    let perm = reader.metadata()?.permissions();
+
+    let ret = io::copy(&mut reader, &mut writer)?;
```
yiwu-arbug, line 361: fsync new file and directory after copy.

yiwu-arbug (commented):
How about create a new `file_system` component and put everything there? Also the rate limiter and the file abstraction can be in different file under this new component.

components/tikv_util/src/file.rs
tabokie: I think that can be for limiter to decide?

components/tikv_util/src/file.rs
tabokie: Actually this function is copied from standard implementation. Maybe instead of changing the behaviour we can add a new `copy_and_sync`?

components/tikv_util/src/file.rs
yiwu-arbug: but `request` is supposed to be a blocking call. say we request 1G write with one `request` call, we wont get a smooth rate limited IO flow, but instead a spike of write after `request` returns.

components/tikv_util/src/file_system/mod.rs
```diff
@@ -1,13 +1,139 @@
...
+    let mut writer = File::create(to.as_ref())?;
+    let perm = reader.metadata()?.permissions();
+
+    let ret = io::copy(&mut reader, &mut writer)?;
+    fs::set_permissions(to.as_ref(), perm)?;
+    Ok(ret)
+}
+
+/// Copies the contents of one file to another. Synchronize afterwards.
+pub fn copy_and_sync<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> io::Result<u64> {
```
yiwu-arbug, line 119: `copy_and_sync` can call `copy` to save some lines.

components/tikv_util/src/file_system/mod.rs
```diff
@@ -1,13 +1,139 @@
...
+        ));
+    }
+
+    let mut reader = File::open(from.as_ref())?;
+    let mut writer = File::create(to.as_ref())?;
+    let perm = reader.metadata()?.permissions();
+
+    let ret = io::copy(&mut reader, &mut writer)?;
+    fs::set_permissions(to.as_ref(), perm)?;
+    writer.sync_all()?;
```
yiwu-arbug, line 133: let's sync writer's directory as well.

yiwu-arbug (commented):
Shall `file_system` to have its own crate, under tikv/components ?

components/tikv_util/src/file.rs
tabokie: That's not gonna happen, notice that limiter now returns actual granted bytes which can be less than requested bytes, 1G request can be truncated by limiter depending on its strategy, just like RocksDB's design.

yiwu-arbug: I see.

components/sst_importer/src/util.rs
```diff
@@ -97,7 +97,7 @@ mod tests {
     use std::{fs, path::Path, sync::Arc};
```
Connor1996, line 97: all std::fs should be replaced with `tikv_util::file_system`?

components/tikv_util/src/file_system/mod.rs
```diff
@@ -1,13 +1,139 @@
...
+/// Indicates how large a buffer to pre-allocate before reading the entire file.
+fn initial_buffer_size(file: &File) -> usize {
+    // Allocate one extra byte so the buffer doesn't need to grow before the
+    // final `read` call at the end of the file.  Don't worry about `usize`
+    // overflow because reading will fail regardless in that case.
+    file.metadata().map(|m| m.len() as usize + 1).unwrap_or(0)
+}
+
+/// Write a slice as the entire contents of a file.
+pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> io::Result<()> {
```
Connor1996: where are these helper functions used?

components/sst_importer/src/util.rs
tabokie: that'll have to wait on followup PRs

components/tikv_util/src/file_system/mod.rs
tabokie: They are equivalents to std::fs's module functions.

components/file_system/src/lib.rs
```diff
@@ -1,15 +1,147 @@
...
+            sync_dir(parent)?;
+        }
+    }
+    Ok(ret)
+}
+
+/// Copies the contents of one file to another. This function will also
+/// copy the permission bits of the original file to the destination file.
+pub fn copy<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> io::Result<u64> {
+    copy_imp(from.as_ref(), to.as_ref(), false)
```
yiwu-arbug, line 135: nit: comment what the bool is, or even better, use an enum
```suggestion
    copy_imp(from.as_ref(), to.as_ref(), false/*sync*/)
```

yiwu-arbug (approved):
LGTM

components/file_system/src/lib.rs
```diff
@@ -1,15 +1,147 @@
...
+        set_io_type(self.previous_io_type);
+    }
+}
+
+/// Indicates how large a buffer to pre-allocate before reading the entire file.
+fn initial_buffer_size(file: &File) -> usize {
+    // Allocate one extra byte so the buffer doesn't need to grow before the
+    // final `read` call at the end of the file.  Don't worry about `usize`
+    // overflow because reading will fail regardless in that case.
+    file.metadata().map(|m| m.len() as usize + 1).unwrap_or(0)
```
yiwu-arbug, line 85: return result instead of unwrap?

Connor1996 (approved):
LGTM

ti-srebot (approved):

PR #9153, opened 2020-11-30
util: Add IO snooper supported by BPF

Connor1996:

Issue Number: close #xxx 

Problem Summary: 
Add io snooper to get the actual read flow categorized by threads which breaks through the page cache. With that, we can introduce a more accurate io limitation on the whole TiKV.

What's Changed: 
Add io snooper with the support of https://github.com/rust-bpf/rust-bcc

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
 
Metrics
image
 
Bench
image

### Release note 
- Add io snopper supported by BPF


components/iosnoop/src/biosnoop.c
```diff
@@ -0,0 +1,62 @@
...
+ */
+#ifdef REQ_WRITE
+    data.rwflag = !!(req->cmd_flags & REQ_WRITE);
+#elif defined(REQ_OP_SHIFT)
+    data.rwflag = !!((req->cmd_flags >> REQ_OP_SHIFT) == REQ_OP_WRITE);
+#else
+    data.rwflag = !!((req->cmd_flags & REQ_OP_MASK) == REQ_OP_WRITE);
+#endif
+
+    events.perf_submit(ctx, &data, sizeof(data));
```
tabokie, line 59: Definitely a kernel-to-user copy here. I'm not sure the actual overhead when read frequency is normally below 10K/s, but I'd prefer we periodically read out a trace summary every 10 ms or so. I've seen use case that accesses `BPF_HASH` from outside the bpf, maybe we can do it here as well?

Connor1996: Let's see how large the overhead is. And if reading out the summary periodically, it can hardly get an accurate byte for one thread like this:
```
let ctx = IOContext::new();
do_some_read();
let read_bytes = ctx.delta();
```

tabokie: I haven't read the code thoroughly, but I believe the update to global map might lag behind the actual IO operation (it must be async as to not affect user bio? not sure), which means the `ctx.delta()` can be less useful here. Or if by any chance it's synchronously updated, the overhead of acquiring one global mutex can't be easy to swallow.

Connor1996: Yes, the global map is updated asynchronously. I'm trying to make it synchronously.

tabokie: I rethink about it and maybe the precise tracing for one request isn't necessary? To throttle a read, all we need to know is *approximately* how many bytes has been read during some period.

yiwu-arbug: I agree we don't need to trace per query stat. Instead we can pass the IOType into bcc code, and store a BPF map of (IOType, read|write) => bytes. We could collect the map every 10ms maybe and use it for the rate limiter.

We can also have a BPF_HISTOGRAM per IO type to export query latency.

components/iosnoop/src/biosnoop.c
```diff
@@ -0,0 +1,62 @@
...
+
+int trace_pid_start(struct pt_regs *ctx, struct request *req)
+{
+    struct val_t val = {};
+    u64 ts;
+
+    u64 id = bpf_get_current_pid_tgid();
+    u32 pid = id & (((u64)1 << 32) - 1);
+    u32 tgid = id >> 32;
+    if (tgid != ##TGID##) {
```
yiwu-arbug, line 25: what's `##TGID##` and what this check does?

components/iosnoop/src/biosnoop.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn new() -> Self {
+        IOSnooper {
+            handle: None,
+            runnable: Arc::new(AtomicBool::new(true)),
+        }
+    }
+
+    pub fn start(&mut self) -> Result<(), String> {
+        let code = include_str!("biosnoop.c");
+        info!("My pid is {}", nix::unistd::getpid());
```
yiwu-arbug, line 66: Can we include stats only if the PID is the same? In case multiple TiKV processes is co-located on a single host.

components/iosnoop/src/biosnoop.c
Connor1996: to filter the io request from tikv process.

components/iosnoop/src/biosnoop.rs
Connor1996: yes, that's what `##TGID##` does. In kernel level, tgid -> pid, pid -> tid.

Cargo.toml
```diff
@@ -63,6 +63,9 @@ test-engines-panic = [
   "engine_test/test-engine-kv-panic",
   "engine_test/test-engine-raft-panic",
 ]
+bcc-iosnoop = [
```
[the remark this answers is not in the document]
Connor1996: It does have a platform-specific-dependencies, see https://github.com/tikv/tikv/pull/9153/files/42aac47c4a3d0edf7720d4c6057233f892cc8eb0#diff-5a645a821a65175c939f78c4cf3491a3ef5719dedd9b802f7c0fec00cc95a711R20

But for Linux version which doesn't support bcc or bcc is not installed, the tikv build will fail. So I provide a feature to enable/disable so that older Linux can build tikv normally.

components/iosnoop/src/biosnoop.c
Connor1996: okay, it would be great not to trace per query. 
For the latest changes(not push yet), it collects the bytes in bpf hashmap and there is no bfp poll in tikv thread more. Instead, it gets the bytes stats from bpf hashmap(syscall) everytime calling IOContext. So rate limiter just need to call `IOContext::delta()` for any period.

Connor1996: Have no idea about how to passing `IOType` into bcc code.
I think we can let every thread to collect read|write from IOContext periodically and the threads know the IOType of their own. Then the threads push the stats to rate limiter?

components/iosnoop/src/biosnoop.rs
Connor1996: Yes, that's what `##TGID##` does. In kernel level: tgid -> pid, pid -> tid.

components/iosnoop/src/biosnoop.c
tabokie: @Connor1996 I found a `bpf_probe_read_user` function, not sure if it's helpful.

yiwu-arbug: Can the IO type be pass through a BPF_HASH map?

Cargo.toml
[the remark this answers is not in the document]
yiwu-arbug: make sense. Do we need to enable the feature by default in the Makefile?

components/iosnoop/src/biosnoop.c
yiwu-arbug: got it.

Connor1996 (commented):
Now it's refined and there is no poll and global mutex anymore. PTAL again @yiwu-arbug @tabokie

components/iosnoop/src/biosnoop.c
tabokie: No, BPF_HASH only supports kernel to user copy, with bpf_probe_read_user maybe we can pass a pointer of thread_local IOType into bpf.

Cargo.toml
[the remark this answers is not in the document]
Connor1996: let me ask the linux version of our CI. If that, `dist_release` can be built with bcc-iosnoop by default. But for `release`, I still suggest to disable by default, otherwise, contributors may suffer from build problems.

components/iosnoop/src/biosnoop.c
Connor1996: BPF_HASH supports user to kernel. We can pass the tls of pid to that BPF_HASH and check io type in bcc.
See https://github.com/iovisor/bcc/blob/0b2f4d0b981803ea8d8393ad2eec93c699e59df1/examples/cpp/pyperf/PyPerfUtil.cc#L212

tabokie: That's great! I was reading this issue https://github.com/iovisor/bcc/issues/2015, if it does work out maybe we can close it.

Connor1996: seems it is supported at initial. https://github.com/iovisor/bcc/blob/a1a1775bef3cc9739566110093b3d85d82fa2f95/src/cc/libbpf.h#L42
I'm not sure if I'm thinking wrong. But rust-bcc does support `table.set()` too. https://docs.rs/bcc/0.0.30/bcc/table/struct.Table.html#impl

Cargo.toml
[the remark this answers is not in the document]
Connor1996: Sadly, the CI is still bounded on CentOS7 of 3.10 kernel. But we can apply for upgrading the kernel version of a small group of machines.

components/iosnoop/src/biosnoop.rs
```diff
@@ -0,0 +1,188 @@
...
+    // the "events" table is where the "open file" events get sent
+    let stats_table = bpf.table("statsbytype").map_err(|e| e.to_string())?;
+    let type_table = bpf.table("typebypid").map_err(|e| e.to_string())?;
+    unsafe {
+        BPF_TABLE = Some((bpf, stats_table, type_table));
+    }
+    // info!("init io snooper"; "pid" => nix::unistd::getpid().to_string());
+    Ok(())
+}
+
```
tabokie, line 142: I remember there's an event loop in the first edition. Could we still keep that and manage a global IOContext in this crate protected by RWLock and updated periodically? In this way metrics flusher and limiter can share access to it.

components/iosnoop/src/biosnoop.c
```diff
@@ -0,0 +1,81 @@
...
+    Compaction,
+    Replication,
+    LoadBalance,
+    Import,
+    Export,
+} io_type;
+
+BPF_HASH(pidbyreq, struct request *, u32);
+BPF_HASH(typebypid, u32, io_type*);
+BPF_HASH(statsbytype, io_type, struct val_t);
```
tabokie, line 26: Since we don't really need to track writes, separate it to R and W to reduce a little contention? Same to IOContext below.

components/iosnoop/src/biosnoop.rs
```diff
@@ -0,0 +1,188 @@
...
+        if let Some((_, _, t)) = BPF_TABLE.as_mut() {
+            let tid = nix::unistd::gettid().as_raw() as u32;
+            let ptr : *const *const _ = &IO_TYPE_ARRAY.as_ptr().add(idx);
+            println!("tid {} ptr is {:?}", tid, IO_TYPE_ARRAY.as_ptr().add(idx));
+            *IO_TYPE_ARRAY[idx] = IOType::Compaction;
+            println!("{:?}", ptr::read(IO_TYPE_ARRAY.as_ptr().add(idx) as *const IOType));
+            let io_type_ptr =
+                std::slice::from_raw_parts_mut(ptr as *mut u8, std::mem::size_of::<*const IOType>());
+            t.set(&mut tid.to_ne_bytes(), io_type_ptr).unwrap();
+        }
```
tabokie, line 33: Should we print out current thread isn't tracked?

Connor1996: isn't tracked?

components/iosnoop/src/biosnoop.c
Connor1996: I only tracked read bytes before. But seems @yiwu-arbug wants to track both read and write.

components/iosnoop/src/biosnoop.rs
Connor1996: metrics flush and limiter can get the stats from IOContext.

components/iosnoop/src/biosnoop.rs
Connor1996: you mean when BPF_TABLE is none? An error message will be printed when calling `init_io_snooper`.

components/iosnoop/src/biosnoop.c
tabokie: Yes, I mean we can use ReadIOContext and WriteIOContext to separately retrieve the two.

siddontang:
seem we want to try eBPF instead of BCC, see https://en.pingcap.com/blog/why-we-switched-from-bcc-tools-to-libbpf-tools-for-bpf-performance-analysis

ethercflow:
> seem we want to try eBPF instead of BCC, see https://en.pingcap.com/blog/why-we-switched-from-bcc-tools-to-libbpf-tools-for-bpf-performance-analysis

Unfortunate, currently only ubuntu 20.10/kernel 5.8 enable BTF by default.  Without BTF It makes little sense for us to switch.

components/iosnoop/src/biosnoop.c
Connor1996: seems unnecessary, we don't need to get them separately

components/iosnoop/src/biosnoop.rs
Connor1996: we can update it in metrics flush periodically and manage a global IOContext. No need to add an event loop.

components/iosnoop/src/biosnoop.c
tabokie: Why not? Rate limiter only uses read stats, I thought write stats are only picked up for external metrics.

Connor1996: rate limter can get them but not use.

components/iosnoop/src/biosnoop.rs
Connor1996: emmmm. We can't ensure the metrics flush and limiter consume the delta within the same period. Seems sharing access is meaningless.

yiwu-arbug:
Can we send another PR to fix lint/formating? This PR is changing 100+ files.

Connor1996:
> Can we send another PR to fix lint/formating? This PR is changing 100+ files.

Not lint, It's due to extracting collections from `tikv_util` to a separate component to avoid circular dependencies.

yiwu-arbug:
> > Can we send another PR to fix lint/formating? This PR is changing 100+ files.
> 
> Not lint, It's due to extracting collections from `tikv_util` to a separate component to avoid circular dependencies.

How about we do that as a separate PR? Make the code easier to review.

Connor1996:
> > > Can we send another PR to fix lint/formating? This PR is changing 100+ files.
> > 
> > 
> > Not lint, It's due to extracting collections from `tikv_util` to a separate component to avoid circular dependencies.
> 
> How about we do that as a separate PR? Make the code easier to review.

okay

Connor1996:
The io-snoop is working normally now and metrics would be shown like this:
image
ready for review @yiwu-arbug @tabokie

components/file_system/src/lib.rs
```diff
@@ -17,66 +18,17 @@ pub use std::fs::{
...
 
-use std::cell::Cell;
 use std::io::{self, ErrorKind, Read, Write};
 use std::path::Path;
 use std::sync::{Arc, Mutex};
 
+pub use iosnoop::{
+    flush_io_metrics, get_io_type, init_io_snooper, set_io_type, IOContext, IOOp, IOType,
+    WithIOType,
+};
```
tabokie, line 28: I'm not familiar with organizing rust module, but I don't like defining these in a subcrate, could we at least get `IOOp, IOType, WithIOType` out?

tabokie (commented):
If a large component in our system is aiming at DBaaS environment, we must find a way to properly emulate and test it in our CI. Any updates on this? @Connor1996

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,268 @@
...
+use bcc::{table::Table, Kprobe, BPF};
+use crossbeam_utils::CachePadded;
+
+static mut BPF_TABLE: Option<(BPF, Table, Table)> = None;
+static IDX_COUNTER: AtomicUsize = AtomicUsize::new(0);
+// For simplicity, just open large enough array. TODO: make it to be Vec
+static mut IO_TYPE_ARRAY: [CachePadded<IOType>; 100] = [CachePadded::new(IOType::Other); 100];
+
+thread_local! {
+    static IDX: usize = unsafe {
```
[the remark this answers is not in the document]
Connor1996: seems drop() won't be called for thread local, see https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&code=%0Astruct%20Foo%3B%0A%0Aimpl%20Foo%7B%0A%20%20%20%20fn%20print(%26self)%7B%0A%20%20%20%20%20%20%20%20println!(%22print%20foo%22)%3B%0A%20%20%20%20%7D%0A%20%20%20%20%0A%20%20%20%20fn%20drop(%26self)%7B%0A%20%20%20%20%20%20%20%20println!(%22drop%20foo%22)%3B%0A%20%20%20%20%7D%0A%7D%0A%0Athread_local!(static%20F%3AFoo%20%3D%20Foo)%3B%0A%0A%0Afn%20main()%7B%0A%20%20%20%20std%3A%3Athread%3A%3Aspawn(%7C%7C%20%7B%0A%20%20%20%20%20%20%20%20F.with(move%7Cf%3A%26Foo%7C%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20f.print()%3B%0A%20%20%20%20%20%20%20%20%7D)%3B%0A%20%20%20%20%7D).join().unwrap()%3B%0A%20%20%20%0A%20%20%20%20println!(%22exit%22)%3B%0A%7D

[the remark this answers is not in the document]
Connor1996: but I agree, we need some mechanism to recycle it.

components/file_system/src/iosnoop/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+make_static_metric! {
+    pub label_enum IOType {
+        other,
+        read,
+        write,
+        coprocessor,
+        flush,
+        compaction,
+        replication,
+        loadbalance,
```
yiwu-arbug, line 15:
```suggestion
        load-balance,
```

components/file_system/src/iosnoop/null.rs
```diff
@@ -0,0 +1,42 @@
...
+        IOStats::default()
+    }
+
+    #[allow(dead_code)]
+    fn delta_and_refresh(&mut self) -> IOStats {
+        IOStats::default()
+    }
+}
+
+pub fn init_io_snooper() -> Result<(), String> {
```
yiwu-arbug, line 24: do we need to define a error type for the `file_sytem` crate and integrate with the `error_code` crate? (we can do that in another PR)

components/file_system/src/iosnoop/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    pub struct IOBytesVec : IntCounter {
+        "type" => IOType,
+        "op" => IOOp,
+    }
+}
+
+lazy_static! {
+    pub static ref IO_BYTES_VEC: IOBytesVec = register_static_int_counter_vec!(
```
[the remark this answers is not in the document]
yiwu-arbug: where is these metrics get updated?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,325 @@
...
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Mutex;
+
+use bcc::{table::Table, Kprobe, BPF};
+use crossbeam_utils::CachePadded;
+
+const MAX_THREAD_IDX: usize = 128;
+
+static mut BPF_TABLE: Option<(BPF, Table, Table)> = None;
+static mut IO_TYPE_ARRAY: [CachePadded<IOType>; MAX_THREAD_IDX] =
```
yiwu-arbug, line 20: Add comment for why it needs to be cache-line padded?

components/file_system/src/iosnoop/metrics.rs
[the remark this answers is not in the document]
Connor1996: updated in metrics flusher which is a single thread worker https://github.com/tikv/tikv/pull/9153/files#diff-9bc4605522a726d5fd1b56c14d7e08d42842f16097742842bada675ff5147d65R229
So no need to make it thread-local.

components/file_system/src/iosnoop/null.rs
Connor1996: Seems no need, only need to print the error message anyway when bootstrapping TiKV.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+use std::collections::VecDeque;
+use std::ffi::CString;
+use std::ptr;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Mutex;
+
+use bcc::{table::Table, Kprobe, BPF};
+use crossbeam_utils::CachePadded;
+
+const MAX_THREAD_IDX: usize = 128;
```
yiwu-arbug, line 17: Is this enough? How many threads TiKV spawns normally?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+
+    fn allocate(&self) -> IdxWrapper {
+        let idx = self.counter.fetch_add(1, Ordering::SeqCst);
+        // TODO: add thread count metrics
+        IdxWrapper(if idx >= MAX_THREAD_IDX {
+            self.free_list
+                .lock()
+                .unwrap()
```
yiwu-arbug, line 75: handle the error? better count the IO flow as "other" type than panic.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+use std::ptr;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Mutex;
+
+use bcc::{table::Table, Kprobe, BPF};
+use crossbeam_utils::CachePadded;
+
+const MAX_THREAD_IDX: usize = 128;
+
+static mut BPF_TABLE: Option<(BPF, Table, Table)> = None;
```
yiwu-arbug, line 19: comment what these two tables are.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+
+pub fn get_io_type() -> IOType {
+    unsafe { *IDX.with(|idx| IO_TYPE_ARRAY[idx.0]) }
+}
+
+unsafe fn get_io_stats() -> Option<HashMap<IOType, IOStats>> {
+    if let Some((_, t, _)) = BPF_TABLE.as_mut() {
+        let mut map = HashMap::default();
+        for e in t.iter() {
+            let typ = ptr::read(e.key.as_ptr() as *const IOType);
```
yiwu-arbug, line 104:
```suggestion
            let io_type = ptr::read(e.key.as_ptr() as *const IOType);
```

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+
+                return self.io_stats_map.replace(map).unwrap();
+            }
+        }
+        HashMap::default()
+    }
+}
+
+pub fn init_io_snooper() -> Result<(), String> {
+    let stat = unsafe {
```
yiwu-arbug, line 164: comment about why we need to get the namespace id?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+    [CachePadded::new(IOType::Other); MAX_THREAD_IDX];
+
+// The index of the element of IO_TYPE_ARRAY for this thread to access.
+thread_local! {
+    static IDX: IdxWrapper = unsafe {
+        let idx = IDX_ALLOCATOR.allocate();
+        if let Some((_, _, t)) = BPF_TABLE.as_mut() {
+            let tid = nix::unistd::gettid().as_raw() as u32;
+            let ptr : *const *const _ = &IO_TYPE_ARRAY.as_ptr().add(idx.0);
+            t.set(&mut tid.to_ne_bytes(), std::slice::from_raw_parts_mut(ptr as *mut u8, std::mem::size_of::<*const IOType>())).unwrap();
```
yiwu-arbug, line 37: Is it only called when thread initialize? We need to set mixed IO type for, e.g. the unified thread pool.

components/engine_traits/src/metrics_flusher.rs
```diff
@@ -47,6 +48,7 @@ impl<K: KvEngine, R: RaftEngine> MetricsFlusher<K, R> {
                 while let Err(mpsc::RecvTimeoutError::Timeout) = rx.recv_timeout(interval) {
                     kv_db.flush_metrics("kv");
                     raft_db.flush_metrics("raft");
+                    flush_io_metrics();
```
yiwu-arbug, line 51: is there a better place to flush io metrics? this logic is not engine specific.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+        .handler("trace_pid_start")
+        .function("blk_account_io_start")
+        .attach(&mut bpf)
+        .map_err(|e| e.to_string())?;
+    Kprobe::new()
+        .handler("trace_req_completion")
+        .function("blk_account_io_completion")
+        .attach(&mut bpf)
+        .map_err(|e| e.to_string())?;
+    // the "events" table is where the "open file" events get sent
```
yiwu-arbug, line 192: this comment is irrelevant?

components/file_system/src/iosnoop/biosnoop.rs
yiwu-arbug: is `Table` in fact thread-safe?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,332 @@
...
+
+static mut BPF_TABLE: Option<(BPF, Table, Table)> = None;
+// This array records the io type for every thread. The address of this array
+// will be passed into BPF, so BPF code can get io type for specific thread
+// without an extra syscall.
+// It should be a thread local variable, but the address of thread local is not
+// reliable. So define a global array and let each thread writes on a specific
+// element. Thus there is no contention for the element and use padding to avoid
+// false sharing.
+static mut IO_TYPE_ARRAY: [CachePadded<IOType>; MAX_THREAD_IDX] =
```
yiwu-arbug, line 27: Does the inner IOType needs to be atomic?

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: it's lazy initlialized when `get_io_type()` or `set_io_type()` is called for the first time. For mixed IO, we need to call `set_io_type()` explicitly everytime switch the task.

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: no, every thread only accesses the element of its own index, so no contention between threads.

components/engine_traits/src/metrics_flusher.rs
Connor1996: seems it's better to flush io metrics in io limiter. I'll leave a todo here, let's keep it now.

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: unified pool(0.8*LOGIC_CPU_NUM as most), grpc(6), scheduler-worker-pool(4), raftstore(3), apply pool(3), rocksdb(10) with some task worker which would not exceed 10.

Connor1996: to be on the safe side, let's make it to 192

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: Table is not thread-safe. In BPF use case, little data race is acceptable.

components/file_system/src/iosnoop/biosnoop.c
```diff
@@ -0,0 +1,200 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
yiwu-arbug, line 1: clang-format the file once?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,406 @@
...
+// This array records the io type for every thread. The address of this array
+// will be passed into BPF, so BPF code can get io type for specific thread
+// without an extra syscall.
+// It should be a thread local variable, but the address of thread local is not
+// reliable. So define a global array and let each thread writes on a specific
+// element. Thus there is no contention for the element and use padding to avoid
+// false sharing.
+// Leave the last element as reserved, when there is no available index, all
+// other threads will be allocated to that index with IOType::Other always.
+static mut IO_TYPE_ARRAY: [CachePadded<IOType>; MAX_THREAD_IDX + 1] =
```
yiwu-arbug, line 48: WeChatWorkScreenshot_5c4b912b-ca66-411f-aee4-0e3c43aaa71a

but it is also being access in bpf, supposed in kernel thread?

components/file_system/src/iosnoop/biosnoop.rs
yiwu-arbug: Any reference to thread-unsafeness of BPF_TABLE?

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: In BCC code, the io type is read when `blk_account_io_start` is called which is fired in the process context. That is to say, it's called in kernel space of user thread rather than kernel thread. So there is no contention between user and kernel.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,406 @@
...
+            counter: AtomicUsize::new(0),
+            free_list: Mutex::new(VecDeque::new()),
+        }
+    }
+
+    fn allocate(&self) -> IdxWrapper {
+        let idx = if let Some(idx) = self.free_list.lock().unwrap().pop_front() {
+            idx
+        } else {
+            self.counter.fetch_add(1, Ordering::SeqCst)
```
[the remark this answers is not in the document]
Connor1996: good idea.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,406 @@
...
+
+impl IOContext {
+    pub fn new() -> Self {
+        IOContext {
+            io_stats_map: unsafe { get_io_stats() },
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn delta(self) -> HashMap<IOType, IOStats> {
```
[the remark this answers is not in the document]
Connor1996: no sure if there is a case for io limiter

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: https://lwn.net/Articles/779120/

Connor1996: > We are not accessing this struct a lot (only when 1. creating thread, 2. flushing metrics, by default every 30s), let's use a mutex?

okay, will do. Though it doesn't make any effect for relieving the data race

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,406 @@
...
+    unsafe { *IDX.with(|idx| IO_TYPE_ARRAY[idx.0]) }
+}
+
+unsafe fn get_io_stats() -> Option<HashMap<IOType, IOStats>> {
+    if let Some((_, t, _)) = BPF_TABLE.as_mut() {
+        let mut map = HashMap::default();
+        for e in t.iter() {
+            let io_type = ptr::read(e.key.as_ptr() as *const IOType);
+            let stats = ptr::read(e.value.as_ptr() as *const IOStats);
+            map.insert(io_type, stats);
```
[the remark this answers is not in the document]
Connor1996: It will introduce a write-write data race which may make the stats not accurate. And the maximum of u64 is about 18466PB, I think it's large enough.

components/file_system/src/iosnoop/biosnoop.rs
yiwu-arbug: yeah, just avoid using `unsafe` as much as possible and make the compiler happy

yiwu-arbug: btw it seems the ebpf maps are in fact thread-safe (at least for the lookup() and insert() methods), but not the inner values.
WeChatWorkScreenshot_0c661312-6fba-4dcb-a2a6-70448bd2ac4e

yiwu-arbug: cited from https://man7.org/linux/man-pages/man2/bpf.2.html

components/file_system/src/iosnoop/biosnoop.rs
yiwu-arbug: got it. so much to learn here. Can you add comment to explain why it doesn't have to be atomic?

components/file_system/src/iosnoop/biosnoop.rs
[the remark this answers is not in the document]
yiwu-arbug: ok

components/file_system/src/iosnoop/biosnoop.rs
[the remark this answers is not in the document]
yiwu-arbug: If we don't need it now then just remove it.

yiwu-arbug:
rest looking good.

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: 截屏2020-12-22 下午1 44 07
yes, seems the update of the hash map is atomical. good news 😄

Connor1996: And find `__sync_fetch_and_add` can be used to update for integar inner values

yiwu-arbug: `__sync_fetch_and_add` incur full barrier. shall we rerun perf test?

btw how do we test perf overhead now? Can we include a bench program in the code? Can be in another PR.

Connor1996: > yeah, just avoid using unsafe as much as possible and make the compiler happy

After some tries, we can't make it safe due to `Table` is `!Send + !Sync`

Connor1996: Haven't tested the overhead yet. I'm writing a micro bench.

components/file_system/src/iosnoop/biosnoop.c
```diff
@@ -0,0 +1,204 @@
...
+#elif defined(REQ_OP_SHIFT)
+  rwflag = !!((req->cmd_flags >> REQ_OP_SHIFT) == REQ_OP_WRITE);
+#else
+  rwflag = !!((req->cmd_flags & REQ_OP_MASK) == REQ_OP_WRITE);
+#endif
+  io_type type = info->type;
+  struct stats_t zero = {}, *val;
+  val = stats_by_type.lookup_or_init(&type, &zero);
+  if (rwflag == 1) {
+    __sync_fetch_and_add(&val->write, req->__data_len);
```
yiwu-arbug, line 123: shall also use atomic when flushing metrics.

Connor1996:
截屏2020-12-22 下午5 57 55
The result shows no much difference(run multiple times)

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,508 @@
...
+    //         return Err(String::from("Can't get namespace stats"));
+    //     }
+    //     stat
+    // };
+    // let code = code.replace("##DEV##", &stat.st_dev.to_string())
+    //   .replace("##INO##", &stat.st_ino.to_string());
+
+    // compile the above BPF code!
+    let mut bpf = BPF::new(&code).map_err(|e| e.to_string())?;
+    // attach kprobes
```
tabokie, line 203: I am curious about when the probe is detached? And can it be manually done?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,508 @@
...
+
+    #[bench]
+    fn bench_write_enable_io_snoop(b: &mut Bencher) {
+        init_io_snooper().unwrap();
+        bench_write(b);
+    }
+
+    #[bench]
+    fn bench_write_disable_io_snoop(b: &mut Bencher) {
+        unsafe { BPF_TABLE = None };
```
tabokie, line 417: Can the hook still be attached?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,508 @@
...
+        let tmp = TempDir::new().unwrap();
+        let file_path = tmp.path().join("bench_write_io_snoop");
+        let mut f = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .custom_flags(O_DIRECT)
+            .open(&file_path)
+            .unwrap();
+
+        let mut w = vec![A512::default(); 1];
```
tabokie, line 469: I think we can try 1MB write without fsync, and include buffered write too.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,508 @@
...
+
+        let mut f = OpenOptions::new()
+            .read(true)
+            .custom_flags(O_DIRECT)
+            .open(&file_path)
+            .unwrap();
+        let mut r = vec![A512::default(); 1];
+        b.iter(|| {
+            set_io_type(IOType::Read);
+            f.seek(SeekFrom::Start(0)).unwrap();
```
tabokie, line 504: 1M and seek a random offset?

components/file_system/src/iosnoop/biosnoop.c
Connor1996: when flushing metrics, it gets the map value as a whole, so it can't use atomic to get it.

components/file_system/src/iosnoop/biosnoop.rs
tabokie: And btw we can start some background r/w in the meantime.

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: the hook is detached when BPF is dropped

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,508 @@
...
+        })
+    };
+}
+
+pub fn get_io_type() -> IOType {
+    unsafe { *IDX.with(|idx| IO_TYPE_ARRAY[idx.0]) }
+}
+
+unsafe fn get_io_stats() -> Option<HashMap<IOType, IOStats>> {
+    if let Some((_, t, _)) = BPF_TABLE.as_mut() {
```
yiwu-arbug, line 129:
```suggestion
    if let Some((_, stats_table, _)) = BPF_TABLE.as_mut() {
```

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,508 @@
...
+// Leave the last element as reserved, when there is no available index, all
+// other threads will be allocated to that index with IOType::Other always.
+static mut IO_TYPE_ARRAY: [CachePadded<IOType>; MAX_THREAD_IDX + 1] =
+    [CachePadded::new(IOType::Other); MAX_THREAD_IDX + 1];
+
+// The index of the element of IO_TYPE_ARRAY for this thread to access.
+thread_local! {
+    static IDX: IdxWrapper = unsafe {
+        let idx = IDX_ALLOCATOR.allocate();
+        if let Some((_, _, t)) = BPF_TABLE.as_mut() {
```
yiwu-arbug, line 57:
```suggestion
        if let Some((_, _, type_table)) = BPF_TABLE.as_mut() {
```

components/file_system/src/iosnoop/biosnoop.c
yiwu-arbug: IIUC the map values are used as raw memory buffer, and we can cast each cell as atomic int and operate on them. Not sure how to write it in rust though.

components/file_system/src/iosnoop/biosnoop.rs
tabokie: Then it means we can make `enable-io-snoop` dynamically configurable.

Connor1996: good idea, it can be used to debug

Connor1996:
截屏2020-12-23 下午7 54 16
(left:enable, right: disable)
Test on a cluster of 100GB+ data and run sysbench insert workload with 1500 threads. No remarkable overhead.

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: The overhead is related to the frequency of disk io, I don't see 1MB and buffered write make any sense.

Connor1996: background r/w is a little random, let's leave micro bench reproducible. And background r/w scene is covered in the real cluster bench.

yiwu-arbug (approved):
LGTM

components/file_system/src/iosnoop/biosnoop.rs
tabokie: The idea is to make disk IO variance relatively small compared to total time elapsed, and also test different code paths involved in realworld IO. But I guess they aren't that import compared to cluster bench.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -0,0 +1,509 @@
...
+/// corresponding type's map in BCC.
+///
+/// With that information, every time calling `IOContext` it get the stored stats
+/// from corresponding type's map in BCC. Thus it enables TiKV to get the latency and
+/// bytes of read/write request per IO-type.
+
+const MAX_THREAD_IDX: usize = 192;
+
+// Hold the BPF to keep it not dropped.
+// The two tables are `stats_by_type` and `type_by_pid` respectively.
```
tabokie, line 35: Better make this a struct? Repeated tuple unpacking looks tiresome.

tabokie (approved):

ti-srebot (approved):

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: Some threads call `set_io_type()` only on init, and seems no demand on dynamically config now. So won't support that now.

PR #9159, opened 2020-12-01
Optimize scheduler latches to avoid latch-wait (#6094)

youjiali1995:

Issue Number: close #xxx 

Problem Summary:
The latches scheduler would prevent key writing ,if there has been one key in waiting queue. I changed this behavior to check the waiting if there has been a hash value repeat with it .

I set scheduler-concurrency to 512 * 1024 by default. If there are 1000K keys written at the same time, every key would be compared in the deque only twice. I think it is efficient enough.

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
* No release note.

### DIFF
--- a/src/storage/txn/latch.rs
+++ b/src/storage/txn/latch.rs
@@ -3,20 +3,23 @@
 use std::collections::hash_map::DefaultHasher;
 use std::collections::VecDeque;
 use std::hash::{Hash, Hasher};
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
 use std::usize;
+
+const WAITING_LIST_SHRINK_SIZE: usize = 8;
+const WAITING_LIST_MAX_CAPACITY: usize = 16;
 
 /// Latch which is used to serialize accesses to resources hashed to the same slot.
 ///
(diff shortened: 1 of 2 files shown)


src/storage/txn/latch.rs
```diff
@@ -123,25 +182,34 @@ impl Latches {
...
         let mut s = DefaultHasher::new();
         key.hash(&mut s);
-        (s.finish() as usize) & (self.size - 1)
+
+        s.finish()
+    }
+
+    #[inline]
+    fn lock_latch(&self, hash: u64) -> MutexGuard<'_, Latch> {
+        self.slots[(hash as usize) & (self.size - 1)]
```
overvenus, line 210:
```suggestion
        self.slots[(hash as usize) % self.size]
```

What about remainder to evenly distribute hashes?

Little-Wallace (approved):
LGTM

sticnarf (commented):
/lgtm

MyonKeminta (approved):

ti-srebot (approved):

PR #9160, opened 2020-12-01
compaction filter: don't use slog in thread local

hicqu:

In the current master branch, enable compaction filter could cause panic.
The root cause is using slog in a thread local context. We haven't located whether it's a bug of `thread_local` or `slog`, but we can avoid calling `slog::*` in thread local contexts.

Don't print any log in thread local contexts.

### Check List 

Tests 

- No code

Side effects

### Release note 
* No release note


Connor1996 (approved):

Little-Wallace (approved):
LGTM

ti-srebot (approved):

Issue #9171, opened 2020-12-02
TiKV server panic when running with TiCDC


amyangfei:
### What version of TiKV are you using?

master, "Git Commit Hash:   5bd92d7fa32a2b6d21e03c69258edc1565ebd7ec"

### Steps to reproduce

Running TiCDC integration test, the basic system structure is 

upstream: 3 TiKVs, 2 TiDBs, 1 PD
downstream: 1 TiKV, 1 TiDB, 1 PD

Use TiCDC to replicate upstream data change to downstream

### What did you expect?

Running CI normally

### What did happened?

CI source:

- https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/cdc_ghpr_integration_test/detail/cdc_ghpr_integration_test/3365/pipeline/87/
- https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/cdc_ghpr_integration_test/detail/cdc_ghpr_integration_test/3366/pipeline

TiKV panic happens

- panic 1, original log: https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/cdc_ghpr_integration_test/runs/3365/nodes/84/steps/244/log/?start=0
```
[2020-12-02T06:01:05.442Z] [2020/12/02 14:00:24.065 +08:00] [FATAL] [lib.rs:482] ["7480000000000000335F728000000000000001@421236263768031235, commit@421236263768031237 is not tracked, region 65"] [backtrace="stack backtrace:
   0: tikv_util::set_panic_hook::{{closure}}
             at components/tikv_util/src/lib.rs:481
   1: std::panicking::rust_panic_with_hook
             at library/std/src/panicking.rs:581
   2: std::panicking::begin_panic_handler::{{closure}}
             at library/std/src/panicking.rs:484
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at library/std/src/sys_common/backtrace.rs:141
   4: rust_begin_unwind
             at library/std/src/panicking.rs:483
   5: std::panicking::begin_panic_fmt
             at library/std/src/panicking.rs:437
   6: resolved_ts::Resolver::untrack_lock
             at components/resolved_ts/src/lib.rs:92
   7: cdc::delegate::Delegate::sink_data
             at components/cdc/src/delegate.rs:586
      cdc::delegate::Delegate::on_batch
             at components/cdc/src/delegate.rs:436
   8: cdc::endpoint::Endpoint<T>::on_multi_batch
[...]

PR #9172, opened 2020-12-02
txn: remove invalid lock num assertion in 1PC

sticnarf:

Reported by https://github.com/tikv/tikv/issues/9171.

The lock number assertion is not valid when there are some mutations that needn't be written (like `CheckNotExists`).

Remove this invalid assertion.

### Check List 

Tests 

- Unit test

### Release note 

- Part of the async commit.


youjiali1995 (commented):
lgtm

MyonKeminta (approved):

ti-srebot (approved):

PR #9173, opened 2020-12-02
txn: remove extra_op related things from MvccTxn

longfangsong:

Problem Summary:

`ExtraOp` is a very unatural thing to be handled in MvccTxn, in fact, the only situation we need it is for getting the old value for cdc during the prewrite process. So maybe we can move it to there.

What's Changed:

Move `ExtraOp` and old_value getting staff to actions module, and change the prewrite part to adopt it

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No Release Note

### DIFF
--- a/components/txn_types/src/types.rs
+++ b/components/txn_types/src/types.rs
@@ -244,6 +244,12 @@ pub enum MutationType {
     Other,
 }
 
+impl MutationType {
+    pub fn may_have_old_value(&self) -> bool {
+        matches!(self, MutationType::Put | MutationType::Delete)
+    }
+}
+
 /// A row mutation.
 #[derive(Debug, Clone)]
 pub enum Mutation {
--- a/components/txn_types/src/types.rs
+++ b/components/txn_types/src/types.rs
@@ -1,4 +1,5 @@
 use super::timestamp::TimeStamp;
+use crate::Write;
 use byteorder::{ByteOrder, NativeEndian};
 use collections::HashMap;
 use kvproto::kvrpcpb;
@@ -322,17 +323,57 @@ impl From<kvrpcpb::Mutation> for Mutation {
     }
 }
 
-#[derive(Default, Debug, Clone, PartialEq)]
-pub struct OldValue {
-    pub short_value: Option<Value>,
-    pub start_ts: TimeStamp,
+/// `OldValue` is used by cdc to read the previous value associated with some key during the prewrite process
+#[derive(Debug, Clone, PartialEq)]
+pub enum OldValue {
+    /// A real `OldValue`
+    Value {
+        short_value: Option<Value>,
+        start_ts: TimeStamp,
+    },
+    /// `None` means we don't found a previous value
+    None,
+    /// `Unspecified` means the user doesn't care about the previous value
+    Unspecified,
+}
+
+impl Default for OldValue {
+    fn default() -> Self {
+        OldValue::Unspecified
+    }
+}
+
+impl From<Option<Write>> for OldValue {
+    fn from(write: Option<Write>) -> Self {
+        match write {
+            Some(w) => OldValue::Value {
+                short_value: w.short_value,
+                start_ts: w.start_ts,
+            },
+            None => OldValue::None,
+        }
+    }
 }
 
 impl OldValue {
+    pub fn specified(&self) -> bool {
+        !matches!(self, OldValue::Unspecified)
+    }
+
+    pub fn exists(&self) -> bool {
+        matches!(
+            self,
+            OldValue::Value {..}
+        )
+    }
+
     pub fn size(&self) -> usize {
-        let value_size = match self.short_value {
-            Some(ref v) => v.len(),
-            None => 0,
+        let value_size = match self {
+            OldValue::Value {
+                short_value: Some(v),
+                ..
+            } => v.len(),
+            _ => 0,
         };
         value_size + std::mem::size_of::<TimeStamp>()
     }
@@ -342,7 +383,7 @@ impl OldValue {
 // key with current ts -> (short value of the prev txn, start ts of the prev txn).
 // The value of the map will be None when the mutation is `Insert`.
 // MutationType is the type of mutation of the current write.
-pub type OldValues = HashMap<Key, (Option<OldValue>, MutationType)>;
+pub type OldValues = HashMap<Key, (OldValue, MutationType)>;
 
 // Extra data fields filled by kvrpcpb::ExtraOp.
 #[derive(Default, Debug, Clone)]
@@ -354,22 +395,8 @@ pub struct TxnExtra {
 }
 
 impl TxnExtra {
-    pub fn add_old_value(
-        &mut self,
-        key: Key,
-        value: Option<OldValue>,
-        mutation_type: MutationType,
-    ) {
-        self.old_values.insert(key, (value, mutation_type));
-    }
-
     pub fn is_empty(&self) -> bool {
         self.old_values.is_empty()
-    }
-
-    pub fn extend(&mut self, other: &mut Self) {
-        self.old_values
-            .extend(std::mem::take(&mut other.old_values))
     }
 }
 
--- a/components/txn_types/src/write.rs
+++ b/components/txn_types/src/write.rs
@@ -324,9 +324,12 @@ impl WriteRef<'_> {
         size
     }
 
-    /// Assume the current `Write` record is the latest version found by reading at `read_ts`, check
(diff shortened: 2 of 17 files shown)


components/txn_types/src/write.rs
```diff
@@ -53,6 +53,10 @@ impl WriteType {
             WriteType::Rollback => FLAG_ROLLBACK,
         }
     }
+
+    pub fn valid(&self) -> bool {
```
longfangsong, line 57: However I think `valid` is not the best name here, it's kind of too wide, but I don't have a better one.

src/storage/txn/actions/prewrite.rs
```diff
@@ -23,9 +26,10 @@ pub fn prewrite<S: Snapshot>(
     mutation: Mutation,
     secondary_keys: &Option<Vec<Vec<u8>>>,
     is_pessimistic_lock: bool,
-) -> Result<TimeStamp> {
+    need_old_value: bool,
+) -> Result<(TimeStamp, Option<Option<OldValue>>)> {
```
longfangsong, line 30: `Option<Option<OldValue>>` looks stange, the outer `Option` stands for "whether we need to find the old value" and the inner one stands for "whether we do found an old value".
Wonder how can we improve this further.

5kbpers: How about defining OldValue as an enum like
```rust
enum OldValue {
    Value {...},
    NotExist,
}
```

longfangsong:
@5kbpers @nrc PTAL

5kbpers (approved):
LGTM, thanks a lot!

src/storage/txn/actions/prewrite.rs
```diff
@@ -23,9 +26,10 @@ pub fn prewrite<S: Snapshot>(
     mutation: Mutation,
     secondary_keys: &Option<Vec<Vec<u8>>>,
     is_pessimistic_lock: bool,
-) -> Result<TimeStamp> {
+    need_old_value: bool,
```
nrc, line 29: I think this can be added to `TransactionProperties` since it is a property of the whole transaction, not individual mutations.

src/storage/txn/actions/prewrite.rs
```diff
@@ -52,20 +56,20 @@ pub fn prewrite<S: Snapshot>(
         };
 
     if mutation.should_not_write {
-        return Ok(TimeStamp::zero());
+        return Ok((TimeStamp::zero(), old_value));
     }
 
     if lock_status.fallback_from_async_commit() {
```
nrc, line 62: I think we `let old_value = ` here, return `None` earlier and avoid the mutable variable, which is a little more idiomatic

components/txn_types/src/types.rs
```diff
@@ -244,6 +244,12 @@ pub enum MutationType {
     Other,
 }
 
+impl MutationType {
+    pub fn may_has_old_value(&self) -> bool {
```
nrc, line 248:
```suggestion
    pub fn may_have_old_value(&self) -> bool {
```

components/txn_types/src/types.rs
```diff
@@ -316,17 +322,34 @@ impl From<kvrpcpb::Mutation> for Mutation {
...
-pub struct OldValue {
-    pub short_value: Option<Value>,
-    pub start_ts: TimeStamp,
+#[derive(Debug, Clone, PartialEq)]
+pub enum OldValue {
+    Value {
+        short_value: Option<Value>,
+        start_ts: TimeStamp,
+    },
+    NotExist,
```
nrc, line 331: What is the difference between `Some(NotExist)` and `None`? Perhaps we can add another variant to the enum with a more descriptive name.

Please add some docs to the enum and variants.

components/txn_types/src/write.rs
nrc: Could you document what the function is for? I'm not sure what it does so I can't suggest a better name

nrc (commented):
This is just a partial review, I'll finish it later...

components/txn_types/src/types.rs
longfangsong: See discussion [here](https://github.com/tikv/tikv/pull/9173#discussion_r534277441).
So maybe we can add a new variant `DontCare`?

components/txn_types/src/write.rs
longfangsong: > Could you document what the function is for? I'm not sure what it does so I can't suggest a better name

The definition of `valid` can be viewed [here](https://github.com/tikv/tikv/blob/a68cb230a818f34b3316874a523bea8a2e17e8a7/src/storage/mvcc/reader/scanner/mod.rs#L388), however I'm afraid `valid` is not a good name for that function too, maybe we should make all these `valid` into `has_old_value` or something alike.

components/txn_types/src/types.rs
nrc: Yeah, I think having a three-valued enum is a good idea. Maybe `Value`, `None`, and `Unspecified`?

components/txn_types/src/write.rs
nrc: I see. I'm not sure about that function, but I think this one would be better called `has_old_value` or similar

src/storage/txn/actions/get_old_value.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// Read the old value for key for CDC.
+/// `prev_write` stands for the previous write record of the key
+/// it must be read in the caller and be passed in for optimization
+pub fn get_old_value<S: Snapshot>(
+    txn: &mut MvccTxn<S>,
+    key: &Key,
+    prev_write: Option<Write>,
+) -> MvccResult<OldValue> {
+    if let Some(w) = prev_write {
```
nrc, line 15: Use match rather than `if let ... else ...`

src/storage/mod.rs
```diff
@@ -742,8 +743,7 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         start_key: Key,
         end_key: Key,
         notify_only: bool,
-        callback: Callback<()>,
-    ) -> Result<()> {
+    ) -> Result<impl Future<Output = Result<()>>> {
```
nrc, line 746: Can you make it an async function instead of using an implicit future?

components/txn_types/src/types.rs
```diff
@@ -316,17 +322,34 @@ impl From<kvrpcpb::Mutation> for Mutation {
...
 
 impl OldValue {
+    pub fn exists(&self) -> bool {
+        !matches!(self, OldValue::NotExist)
+    }
     pub fn size(&self) -> usize {
-        let value_size = match self.short_value {
-            Some(ref v) => v.len(),
-            None => 0,
+        let value_size = if let OldValue::Value { short_value, .. } = self {
```
nrc, line 345: Use match rather than two `if let ... else ...`s

src/storage/mod.rs
longfangsong: Sorry it's from another branch, I made a rebase mistake ……

components/txn_types/src/write.rs
```diff
@@ -53,6 +53,10 @@ impl WriteType {
             WriteType::Rollback => FLAG_ROLLBACK,
         }
     }
+
+    pub fn has_old_value(&self) -> bool {
```
nrc, line 57: Could this be a method on Write instead of WriteType?

src/storage/txn/actions/get_old_value.rs
```diff
@@ -0,0 +1,193 @@
...
+                Some(w)
+            } else {
+                last_valid_write(txn, key)?
+            };
+            Ok(valid_write
+                .map(|w| OldValue::Value {
+                    short_value: w.short_value,
+                    start_ts: w.start_ts,
+                })
+                .unwrap_or(OldValue::None))
```
nrc, line 27: Can you impl `From<Option<Write>>` for Write, so that this can be `valid_write.into()`?

nrc (approved):
lgtm, thanks for the changes!

ti-srebot (approved):

components/txn_types/src/write.rs
```diff
@@ -218,6 +218,10 @@ impl Write {
             gc_fence: self.gc_fence,
         }
     }
+
+    pub fn may_has_old_value(&self) -> bool {
```
nrc, line 222:
```suggestion
    pub fn may_have_old_value(&self) -> bool {
```

nrc (approved):
lgtm (one comment on improving the name of a function)

PR #9174, opened 2020-12-03
*: fix clippy

5kbpers:

Problem Summary:
`clippy-all` script did not check for tests.

What's Changed:
Fix the script and make clippy happy.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A


Little-Wallace (approved):
LGTM

ti-srebot (approved):

components/tidb_query_datatype/src/codec/mysql/time/mod.rs
```diff
@@ -2735,6 +2735,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::excessive_precision)]
```
BusyJay, line 2738: I don't get it, what can trigger it?

scripts/clippy-all
```diff
@@ -20,24 +20,18 @@ ALLOWED_CLIPPY_LINTS=(-A clippy::module_inception  \
 	-A clippy::new_ret_no_self \
 	-A clippy::unnecessary_sort_by)
 
-if [[ "${TIKV_ENABLE_FEATURES}" == *prost-codec* ]] ; then
-	cargo clippy --workspace --all-targets --no-default-features \
-		--exclude cdc --exclude backup --exclude tests --exclude cmd \
-		--exclude fuzz-targets --exclude fuzzer-honggfuzz --exclude fuzzer-afl --exclude fuzzer-libfuzzer \
+cargo clippy --workspace --all-targets --no-default-features \
```
BusyJay, line 23: The original style is on purpose to speed up the checking speed.

components/tidb_query_datatype/src/codec/mysql/time/mod.rs
5kbpers: Some test cases use float value which excess precision, not sure if we need to modify these tests.

scripts/clippy-all
5kbpers: But it doesn't cover all code snippets.

components/tidb_query_datatype/src/codec/mysql/time/mod.rs
BusyJay: /cc @iosmanthus PTAL

scripts/clippy-all
BusyJay: Even after adding `--all-targets`?

components/tidb_query_datatype/src/codec/mysql/time/mod.rs
iosmanthus: LGTM

scripts/clippy-all
5kbpers: Tried to remove the for loop and found more warnings.😅

BusyJay (commented):
LGTM

PR #9175, opened 2020-12-03
copr: support analyze common handle and columns by scan once table rows

lzmhhh123:
Signed-off-by: lzmhhh123 <lzmhhh123@gmail.com>

Issue Number: close https://github.com/pingcap/tidb/issues/20243

related changes: 
- https://github.com/pingcap/tipb/pull/197
- https://github.com/pingcap/tidb/pull/21381

What's Changed:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more MEM

### Release note 

- none.


src/coprocessor/statistics/analyze.rs
```diff
@@ -211,6 +249,8 @@ struct SampleBuilder<S: Snapshot> {
     cm_sketch_depth: usize,
     cm_sketch_width: usize,
     columns_info: Vec<tipb::ColumnInfo>,
+    common_handle_req: Option<AnalyzeIndexReq>,
```
iosmanthus, line 252: Seems this field is useless, if `common_handle_col_ids.len() > 0` then this field is a `Some` value. Otherwise, it is a None or `common_handle_col_ids` is empty.

components/tidb_query_executors/src/runner.rs
```diff
@@ -115,6 +115,11 @@ impl BatchExecutorsRunner<()> {
                     BatchTopNExecutor::check_supported(&descriptor)
                         .map_err(|e| other_err!("BatchTopNExecutor: {}", e))?;
                 }
+                // For TiFlash MPP engine. Not support now.
+                ExecType::TypeJoin
```
iosmanthus, line 119: Could we return some error instead?

src/coprocessor/statistics/analyze.rs
lzmhhh123: If a table hash common handle. Then `common_handle_col_ids.len() > 0` is always satisfied when analyze columns. But `common_handle_req` only in the mixed type of analyze request.

components/tidb_query_executors/src/runner.rs
lzmhhh123: OK

src/coprocessor/statistics/analyze.rs
iosmanthus, line 252: This field could use a bool flag instead.

src/coprocessor/statistics/analyze.rs
```diff
@@ -330,7 +401,16 @@ impl<S: Snapshot> SampleBuilder<S> {
                 }
             }
         }
-        Ok((collectors, pk_builder))
+        let mut idx_res: Option<tipb::AnalyzeIndexResp> = None;
```
iosmanthus, line 404: Prefer to transform the result into protobuf in `handle_mixed` instead of in this function. We could add a struct to wrap this two result and transform it by `into_proto` in `handle_mixed`.

iosmanthus:
Seems we need to add some integration tests to this pull request.

iosmanthus:
@coocood @breeswish @zhongzc @andylokandy PTAL

src/coprocessor/statistics/analyze.rs
```diff
@@ -297,6 +329,26 @@ impl<S: Snapshot> SampleBuilder<S> {
...
+                    let mut data = vec![];
+                    for handle_id in &self.common_handle_col_ids {
+                        let mut handle_col_val = vec![];
+                        columns_slice[*handle_id as usize].encode(
+                            *logical_row,
+                            &columns_info[*handle_id as usize],
+                            &mut EvalContext::default(),
+                            &mut handle_col_val,
+                        )?;
+                        data.extend_from_slice(&handle_col_val);
```
skyzh, line 343: What about using `append` here?

src/coprocessor/statistics/analyze.rs
```diff
@@ -297,6 +329,26 @@ impl<S: Snapshot> SampleBuilder<S> {
                 columns_info = &columns_info[1..];
             }
 
+            if self.analyze_common_handle {
+                for logical_row in &result.logical_rows {
+                    let mut data = vec![];
+                    for handle_id in &self.common_handle_col_ids {
+                        let mut handle_col_val = vec![];
```
skyzh, line 336: I would recommend use `Vec::with_capacity` if we can know vector length in advance.

src/coprocessor/statistics/analyze.rs
```diff
@@ -297,6 +329,26 @@ impl<S: Snapshot> SampleBuilder<S> {
...
+                        columns_slice[*handle_id as usize].encode(
+                            *logical_row,
+                            &columns_info[*handle_id as usize],
+                            &mut EvalContext::default(),
+                            &mut handle_col_val,
+                        )?;
+                        data.extend_from_slice(&handle_col_val);
+                        if let Some(common_handle_cms) = common_handle_cms.as_mut() {
+                            common_handle_cms.insert(&data);
```
skyzh, line 345: It seems that we would append `handle_col_val` from previous iteration into `common_handle_cms`. Is this intended?

src/coprocessor/statistics/analyze.rs
```diff
@@ -330,7 +382,14 @@ impl<S: Snapshot> SampleBuilder<S> {
                 }
             }
         }
-        Ok((collectors, pk_builder))
+        let mut idx_res: Option<AnalyzeIndexResult> = None;
```
skyzh, line 385: Use `let mut idx_res  = None;` is enough.

skyzh: By the way, we can also write:
```
let idx_res = if xxx { ...; Some(...); } else { None };
```

src/coprocessor/statistics/analyze.rs
lzmhhh123: Yes, the same as MockTiKV in TiDB.

src/coprocessor/statistics/analyze.rs
lzmhhh123: It's hard to estimate the length.

sticnarf (approved):
LGTM, except for there is no unit test for the added code :/

src/coprocessor/statistics/analyze.rs
```diff
@@ -71,21 +71,35 @@ impl<S: Snapshot> AnalyzeContext<S> {
...
         };
         Ok(res_data)
     }
 
+    async fn handle_mixed(builder: &mut SampleBuilder<S>) -> Result<Vec<u8>> {
+        let (col_res, idx_res) = builder.collect_columns_stats().await?;
+
+        let res_data = {
+            let resp: tipb::AnalyzeMixedResp;
+            match idx_res {
```
iosmanthus, line 88: We could use `ok_or_else` instead.

src/coprocessor/statistics/analyze.rs
```diff
@@ -292,15 +320,34 @@ impl<S: Snapshot> SampleBuilder<S> {
             max_sample_size: req.get_sample_size() as usize,
             cm_sketch_depth: req.get_cmsketch_depth() as usize,
             cm_sketch_width: req.get_cmsketch_width() as usize,
+            stats_version: if let Some(ref idx_req) = common_handle_req {
```
iosmanthus, line 323: We could use `map`, `unwrap_or_else` combinator

src/coprocessor/statistics/analyze.rs
```diff
@@ -346,6 +387,60 @@ impl<S: Snapshot> SampleBuilder<S> {
                 columns_info = &columns_info[1..];
             }
 
+            if self.analyze_common_handle {
+                // cur_val recording the current value's data and its counts when iterating index's rows.
+                // Once we met a new value, the old value will be pushed into the topn_heap to maintain the
+                // top-n information.
+                let mut cur_val: (u32, Vec<u8>) = (0, vec![]);
```
iosmanthus, line 394: a little bit confusing convention to use a tuple, could we use pattern matching instead?

iosmanthus (approved):

src/coprocessor/statistics/analyze.rs
lzmhhh123: `topn_heap` uses the tuple as the element in it. I prefer to keep it.

skyzh:
Seems that this PR doesn't satisfy cargo clippy. Please refer to CI log and fix it before merging.

PR #9177, opened 2020-12-03
raftstore: use async ready raft

gengliqi:

Problem Summary:
Update the raft-rs version and use the async ready raft.
This PR is part of https://github.com/tikv/tikv/pull/8855.

What's Changed:
Remove the early-apply config because this feature has already existed in raft-rs.

### Related changes

Tests 

- Integration test

Side effects

- No

### Release note 
* No Release Note


components/test_raftstore/src/common-test.toml
```diff
@@ -41,7 +41,6 @@ capacity = "64MB"
 [metric]
 
 [raftstore]
-sync-log = false
```
BusyJay, line 44: Should be in next PR.

tests/integrations/config/test-custom.toml
```diff
@@ -94,7 +94,6 @@ endpoints = [
 job = "tikv_1"
 
 [raftstore]
-sync-log = false
```
BusyJay, line 97: Should be in next PR.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -657,8 +654,13 @@ where
...
         self.on_check_merge();
         // Apply committed entries more quickly.
-        if self.fsm.peer.raft_group.store().committed_index()
+        // Or if it's a leader. This implicitly means it's a singleton
+        // because it becomes leader in `Peer::new` when it's a
+        // singleton. It has a no-op entry that need to be persisted,
+        // committed, and then it should apply it.
+        if self.fsm.peer.raft_group.store().commit_index()
             > self.fsm.peer.raft_group.store().applied_index()
+            || self.fsm.peer.is_leader()
```
BusyJay, line 663: Any test case?

components/test_raftstore/src/common-test.toml
gengliqi: The `sync-log` has already been removed in master.

components/raftstore/src/store/fsm/peer.rs
gengliqi: Add a simple test `test_node_apply_no_op`.

tests/failpoints/cases/test_early_apply.rs
```diff
@@ -1,92 +0,0 @@
...
-use raft::eraftpb::MessageType;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
-use std::thread;
-use std::time::*;
-use test_raftstore::*;
-
-/// Tests early apply is disabled for singleton.
-#[test]
-fn test_singleton_early_apply() {
```
BusyJay, line 12: The case should be reserved to see if early apply is still working as expected.

gengliqi: Ok, I will move it to `test_single` file

tests/integrations/raftstore/mod.rs
```diff
@@ -18,6 +17,7 @@ mod test_region_heartbeat;
 mod test_region_info_accessor;
 mod test_replica_read;
 mod test_replication_mode;
+mod test_restart;
```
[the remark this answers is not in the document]
gengliqi: Because these tests are used for testing the behavior of handing commit index after restarting and there is no individual `early apply` feature.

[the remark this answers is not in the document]
BusyJay: The reason why we need to handle commit index is because we apply committed entries before persisting commit index. That's what "early apply" means. If the cases are to check restarting, then there are other things need to be checked. For example, a restarted node should not lose updates, should be able to rejoin the group, etc.

tests/failpoints/cases/test_early_apply.rs
gengliqi: I change them back and split it into two tests and add some comments

BusyJay: I don't see them in diff. Did you commit it?

gengliqi: My fault. I forget to use `git add`.

gengliqi:
PTAL

components/raftstore/src/store/peer_storage.rs
```diff
@@ -515,70 +545,52 @@ fn validate_states<EK: KvEngine, ER: RaftEngine>(
...
-
-    if commit_index < apply_state.get_last_commit_index() {
-        return Err(box_err!(
-            "raft state {:?} not match apply state {:?} and can't be recovered.",
-            raft_state,
-            apply_state
-        ));
-    }
     let recorded_commit_index = apply_state.get_commit_index();
+    let state_str = format!(
```
BusyJay, line 549: How about make it a function for lazy evaluation?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -835,10 +847,22 @@ where
...
 
     #[inline]
-    pub fn committed_index(&self) -> u64 {
+    pub fn commit_index(&self) -> u64 {
         self.raft_state.get_hard_state().get_commit()
     }
 
+    #[inline]
+    pub fn set_commit_index(&mut self, commit: u64) {
+        if commit > self.commit_index() {
```
BusyJay, line 856: When will `commit < self.commit_index()`?

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -1439,16 +1491,13 @@ where
...
                 self.post_pending_read_index_on_replica(ctx);
+                // If there is a snapshot, it must belongs to the last ready.
+                self.raft_group
+                    .on_persist_ready(self.last_unpersisted_number);
             }
             CheckApplyingSnapStatus::Idle => {}
         }
 
-        if !self.pending_messages.is_empty() {
-            fail_point!("raft_before_follower_send");
```
BusyJay, line 1447: I'm afraid they may be used in external tests.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1002,23 +1026,23 @@ where
...
-        for entry in entries {
-            if !ready_ctx.sync_log() {
-                ready_ctx.set_sync_log(get_sync_log_from_entry(entry));
-            }
-        }
+        let update_cache = if let Some(ref mut cache) = self.cache {
+            Some((cache, entries.clone()))
+        } else {
+            None
+        };
```
NingLin-P, line 1040: Why we need `update_cache`? Also, maybe we can append entries to the cache first then the `WriteBatch` to avoid cloning.

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/peer_storage.rs
gengliqi: You are right. The code is updated.

components/raftstore/src/store/peer.rs
gengliqi: I change them back.

components/raftstore/src/store/peer_storage.rs
gengliqi: For now, the answer is no so I change to `assert!(commit >= self.commit_index())`

components/raftstore/src/store/peer_storage.rs
gengliqi: Done.

components/raftstore/src/store/peer.rs
```diff
@@ -1045,17 +1045,18 @@ where
...
                     resp.to = m.from;
-                    resp.index = self.get_store().committed_index();
+                    // Must use the commit index of `PeerStorage` instead of the commit index
+                    // in raft-rs which may be greater than the former one.
+                    // For more details, see the annotations above `on_leader_commit_idx_changed`.
+                    resp.index = self.get_store().commit_index();
                     resp.set_entries(m.take_entries());
 
-                    self.pending_messages.push(resp);
+                    self.raft_group.raft.msgs.push(resp);
```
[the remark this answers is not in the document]
gengliqi: Maybe this response can be sent immediately?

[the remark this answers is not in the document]
gengliqi: I use `self.raft_group.raft.msgs.push(resp)` because I want to remain the order. But I think it doesn't matter.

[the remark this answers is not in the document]
gengliqi: I find the `pre_transfer_leader` uses the same logic.

gengliqi:
I move the `PollContext::need_flush_trans` to `Transport`(add a `need_flush` function) so we don't need to set `need_flush_trans` every time after sending msgs.

NingLin-P:
> I move the `PollContext::need_flush_trans` to `Transport`(add a `need_flush` function) so we don't need to set `need_flush_trans` every time after sending msgs. PTAL again @BusyJay @NingLin-P

Looks like `RaftClient::flush` already have a `need_flush` check, maybe we can just remove the `need_flush` function /cc @BusyJay 
https://github.com/tikv/tikv/blob/e146fae0cac3acf4a8a293860c52d720ccd8182d/src/server/raft_client.rs#L892-L894

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1033,18 +1034,12 @@ where
             (e.get_index(), e.get_term())
         };
 
-        let update_cache = if let Some(ref mut cache) = self.cache {
-            Some((cache, entries.clone()))
-        } else {
-            None
-        };
+        if let Some(ref mut cache) = self.cache {
+            cache.append(&self.tag, &entries);
```
BusyJay, line 1038: What if L1041 returns error, the cache may be wrong.

src/server/transport.rs
```diff
@@ -44,12 +49,20 @@ where
...
             Err(reason) => Err(raftstore::Error::Transport(reason)),
         }
     }
 
+    fn need_flush(&self) -> bool {
+        self.need_flush
+    }
+
     fn flush(&mut self) {
+        self.need_flush = false;
```
BusyJay: May need to check the flag before flushing.

BusyJay (commented):
> maybe we can just remove the need_flush function

It's an implementation of master. Not sure whether the old implement also has the check. But it's a good idea.

components/raftstore/src/store/peer_storage.rs
gengliqi: Yes. This is why I did it before. But I find if `append` returns error, TiKV will panic so it doesn't matter that cache is wrong.

src/server/transport.rs
gengliqi: I remove this line. You can see the new commit.

components/raftstore/src/store/peer_storage.rs
BusyJay: Can you add a comment to the line which panics? So we can be warned for future rewrite.

gengliqi: Added.

gengliqi:
PTAL, again and again @BusyJay @NingLin-P

NingLin-P (approved):
LGTM

BusyJay:
LGTM

ti-srebot (approved):

PR #8865, opened 2020-10-21
encryption: Turn file dictionary into log style

hunterlxt:
### What problem does this PR solve?

Issue Number: close https://github.com/tikv/tikv/issues/8212

Problem Summary:
When file number is large, file dictionary size is large too, so TiKV needs more time to rewrite the file dictionary.

### What is changed and how it works?

Add a new file format `LogFile`, it supports append writing.

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Side effects

- Breaking backward compatibility
use new file format to store file dictionary.

### Release note 
- Turn file dictionary into log style

PR #9178, opened 2020-12-03
encryption: pipeline save file dict

Little-Wallace:

Problem Summary:
Encryption manager has to save the file dict on disk every time when there is a new file created or updated. It will cost too much time to do this operation. https://github.com/tikv/tikv/pull/8865 has solved thie problem but it is hard to pick this change to release-4.0 because it also changes the file format, if user upgrade one cluster from release-4.0 to https://github.com/tikv/tikv/pull/8865, he can not go back to release-4.0 again. 

This PR will make all of thread which want to change file dict in a pipeline queue. In origin implement, if there is  4 thread trying to call `save_file_dict`,  it may wait 3 * IO round and cost one IO round to sync the file dict again. But it is not necessary, because the second thread can sync the result of the 3rd and the 4th thread together, the 3rd and the 4th only need to wait 2 * IO round and cost no time to sync itself changes.
In this PR, no matter how many threads there are trying to call `save_file_dict`,  every of them only need to wait 2 IO round and do nothing at most.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

- No release note.

### DIFF
--- a/components/encryption/src/manager/mod.rs
+++ b/components/encryption/src/manager/mod.rs
@@ -20,15 +20,18 @@ use crate::io::EncrypterWriter;
 use crate::master_key::{create_backend, Backend, PlaintextBackend};
 use crate::metrics::*;
 use crate::{Error, Result};
+use dashmap::DashMap;
 use std::collections::vec_deque::VecDeque;
 
 const KEY_DICT_NAME: &str = "key.dict";
 const FILE_DICT_NAME: &str = "file.dict";
 const ROTATE_CHECK_PERIOD: u64 = 600; // 10min
+const DEFAULT_FILES_CACHE_CAPACITY: usize = 1024 * 128;
 
 struct Dicts {
     // Maps data file paths to key id and metadata. This file is stored as plaintext.
     file_dict: Mutex<FileDictionary>,
+    files_cache: DashMap<String, FileInfo>,
     // Lock to make sure there's no concurrent update operations to file dictionary.
     write_queue: Mutex<VecDeque<(ThreadId, Arc<AtomicBool>, Arc<AtomicBool>)>>,
     log_cond: Condvar,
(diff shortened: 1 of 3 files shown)


components/encryption/src/manager/mod.rs
```diff
@@ -18,16 +20,21 @@ use crate::io::EncrypterWriter;
...
 
 const KEY_DICT_NAME: &str = "key.dict";
 const FILE_DICT_NAME: &str = "file.dict";
 const ROTATE_CHECK_PERIOD: u64 = 600; // 10min
+const DEFAULT_FILES_CACHE_CAPACITY: usize = 1024 * 128;
 
 struct Dicts {
     // Maps data file paths to key id and metadata. This file is stored as plaintext.
     file_dict: Mutex<FileDictionary>,
+    files_cache: DashMap<String, FileInfo>,
```
hunterlxt, line 34: Add comments for this?

components/encryption/src/manager/mod.rs
```diff
@@ -175,31 +190,80 @@ impl Dicts {
...
+            }
+            assert!(write_queue.len() > 0);
+            write_queue.len()
+        };
+        let ret = self.save_file_dict();
+        let save_err = ret.is_err();
+        if let Err(e) = ret {
+            error!("encryption: failed to save file dictionary."; "error" => ?e);
+        }
+        let mut write_queue = self.write_queue.lock().unwrap();
```
hunterlxt, line 243: The logic is complicated... I read the part several times, it should be okay. @yiwu-arbug Pls check again.

Little-Wallace: I will add some comment to make it easy to understand.

yiwu-arbug:
I incline to just cherry-pick #8865 in a somehow -forward-compatible way. The pipelined approach is error-prone and don't bring much value - we only need it for 4.0 release.

Little-Wallace:
We will cherry-pick  #8865 to solve this problem because the logic of this PR is too complicated.

PR #8728, opened 2020-09-24
copr: Implement non-nullable for string related functions

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

### What problem does this PR solve?

Issue Number: https://github.com/tikv/tikv/issues/8491

### What is changed and how it works?

Implement non-nullable for all string related functions.

### Check List

Tests

- Unit test
- Integration test

PR #9179, opened 2020-12-03
copr: add cast year to int

ichn-hu:

Problem Summary: MySQL treat 0 as `0000` when casting from year to string, in order to match the bahavior, we should introduce a new cast function.

What's Changed:

### Related changes

- N/A

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- N/A

### Release note 

- fix casting year 0 as string 0000


components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -3574,6 +3595,35 @@ mod tests {
...
+
+        let ref_cs = helper_get_cs_ref(&cs);
+
+        test_as_string_helper(
+            ref_cs,
+            |ctx, extra, val| {
+                let val = val.map(|x| *x as i64);
+                cast_uint_as_string(ctx, extra, val.as_ref())
+            },
+            "cast_year_as_string",
```
Xuanwo, line 3623: https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/36921/pipeline#step-97-log-39

`cargo fmt` is required here.

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -604,6 +606,25 @@ fn cast_any_as_string<T: ConvertTo<Bytes> + Evaluable + EvaluableRet>(
     }
 }
 
+#[rpn_fn(nullable, capture = [ctx, extra])]
```
Xuanwo, line 609: Please considering remove `nullable` here so we don't need to check this, take [this PR](https://github.com/tikv/tikv/pull/8728) as ref.

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -604,6 +606,25 @@ fn cast_any_as_string<T: ConvertTo<Bytes> + Evaluable + EvaluableRet>(
...
+#[inline]
+fn cast_year_as_string(
+    ctx: &mut EvalContext,
+    extra: &RpnFnCallExtra,
+    val: Option<&Int>,
+) -> Result<Option<Bytes>> {
+    match val {
+        None => Ok(None),
+        Some(val) => {
+            let val = match *val as i64 {
```
Xuanwo, line 619: Maybe use `if` directly looks more clear?

Xuanwo (changes requested):

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -3574,6 +3595,35 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_year_as_string() {
+        test_none_with_ctx_and_extra(cast_year_as_string);
+
+        let cs: Vec<(i64, Vec<u8>, String)> = vec![
```
andylokandy, line 3602: Please add more tests according to thttps://dev.mysql.com/doc/refman/8.0/en/year.html

components/tidb_query_expr/src/impl_cast.rs
ichn-hu: I want to try pattern matching, that's amazing!

components/tidb_query_expr/src/impl_cast.rs
ichn-hu: done

components/tidb_query_expr/src/impl_cast.rs
ichn-hu: I think we don't need to match the behavior on the document here, I am migrating the function from TiDB to TiKV, and the only special case is for `0`, as for cases like 69 -> 2069, it is handled on insert, and will not use this function.

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -604,6 +606,25 @@ fn cast_any_as_string<T: ConvertTo<Bytes> + Evaluable + EvaluableRet>(
...
+fn cast_year_as_string(
+    ctx: &mut EvalContext,
+    extra: &RpnFnCallExtra,
+    val: Option<&Int>,
+) -> Result<Option<Bytes>> {
+    match val {
+        None => Ok(None),
+        Some(val) => {
+            let val = match *val as i64 {
+                0 => "0000".as_bytes().to_vec(),
```
skyzh, line 620: you may use `b"0000".to_vec()` here.

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -3574,6 +3595,35 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_year_as_string() {
+        test_none_with_ctx_and_extra(cast_year_as_string);
+
+        let cs: Vec<(i64, Vec<u8>, String)> = vec![
+            (
+                0,
+                "0000".as_bytes().to_vec(),
```
skyzh, line 3605: ditto. May use `b"0000"` here.

components/tidb_query_expr/src/impl_cast.rs
Xuanwo: Pattern matching is cool, but it's too verbose here compared with:

```rust
let val = if *val == 0 { abc } else { def }
```

skyzh (approved):
Rest LGTM. By the way, it seems that `tipb` signature has not been added here. You may send a PR to `tipb` and run integration test with that.

components/tidb_query_expr/src/impl_cast.rs
ichn-hu: Nope, I figured out that I don't need to update tipb, we have FieldType in `get_cast_fn_rpn_meta` and it can be used to distinguish cast_year_to_int.

components/tidb_query_expr/src/impl_cast.rs
ichn-hu: Cool, nice point.

components/tidb_query_expr/src/impl_cast.rs
andylokandy: Please fix this, rest lgtm

ichn-hu: According to Alex Chi, writer can't be used together with ctx, and the amoung of bytes written is limited, therefore we can keep it this way.

Xuanwo (approved):

components/tidb_query_expr/src/impl_cast.rs
Xuanwo: Remove `nullable` and add `writer` is different thing.

Xuanwo (approved):
LGTM

andylokandy:
LGTM

ti-srebot (approved):

breeswish:
Do we need cherry pick?

ichn-hu:
> Do we need cherry pick?

I think it is okay not to cherrypick, given it is a rather corner case.

PR #9182, opened 2020-12-03
copr: fetch restore data while new collation is meet in `BatchTableScan`

iosmanthus:
Signed-off-by: iosmanthus <myosmanthustree@gmail.com>

This pull requests ignore `is_key_only` scanner option while encountering field type that is probably a new collation string and fetches restore data. (TiDB has a global variable that acts as a switch of new collation but not pushed down yet).

Problem Summary:

If a column is probably in a new collation format, then we try to fetch its value.
What's Changed:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

- copr: fetch restore data while new collation is meet in `BatchTableScan`

### DIFF
--- a/components/tidb_query_executors/src/table_scan_executor.rs
+++ b/components/tidb_query_executors/src/table_scan_executor.rs
@@ -5,7 +5,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use kvproto::coprocessor::KeyRange;
-use tidb_query_datatype::{EvalType, FieldTypeAccessor};
+use tidb_query_datatype::{EvalType, FieldTypeAccessor, FieldTypeFlag};
 use tikv_util::collections::HashMap;
 use tipb::ColumnInfo;
 use tipb::FieldType;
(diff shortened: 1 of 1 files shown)


lysu:
hi @iosmanthus  please help resolve conflict and help merge... this PR block https://github.com/pingcap/tidb/pull/23096 thanks

iosmanthus:
@wjhuang2016 @lzmhhh123 PTAL

components/tidb_query_executors/src/table_scan_executor.rs
```diff
@@ -68,6 +68,7 @@ impl<S: Storage> BatchTableScanExecutor<S> {
                 handle_indices.push(index);
             } else {
                 if !primary_column_ids_set.contains(&ci.get_column_id())
+                    || (ci.is_string_like() && !ci.flag().contains(FieldTypeFlag::BINARY))
```
wjhuang2016, line 71: Use the collation info to check the binary instead of the flag.

lzmhhh123 (approved):

PR #9193, opened 2020-12-07
[DNM] encryption: Add config to gate the log style file dictionary optimization

yiwu-arbug:

Cherry-pick of #9191.

Problem Summary:
#8865 introduces an optimization to greatly reduce IO and mutex contention when encryption is enabled, by turning the file dictionary file into log style. The optimization is also useful to improve Lightning data import speed. We would like to include the optimization in 4.0.x branch as well.

However, the optimization introduce a data format that is not recognizable by TiKV <= 4.0.8. To address the problem, this PR introduce the `security.encryption.enable-file-dictionary-log` config to allow turning off the optimization and convert to previous data format. The option will be default to true for TiKV >= 5.0, and default to off for TiKV < 5.0 but >= 4.0.9. User can downgrade TiKV cluster with encryption turned on to <= 4.0.8 with the following steps:

1. set `security.encryption.enable-file-dictionary-log = false` and restart TiKV. After TiKV successfully start, the data format is converted back to the same as before the optimization.
2. downgrade to version between (included) 4.0.0 and 4.0.8 (since encryption is not supported pre-4.0).

What's Changed:
introduce the `security.encryption.enable-file-dictionary-log` config to allow turning off the optimization and convert to previous data format.

### Related changes

- #8865 introduces the backward-incompatible change.
- Will update user docs

### Check List 

Tests 

- New unit tests
- manual testing for downgrade to previous file dictionary format
- Will do bank2 test for correctness

### Release note 
* Reduce IO and mutex contention when encryption-at-rest is used. The optimization is gated behind the  `security.encryption.enable-file-dictionary-log` config.


yiwu-arbug:
DNM before #9193 landed.

Issue #9188, opened 2020-12-04
The unified pool cannot schedule kv_scan, raw_scan and analyze requests well


sticnarf:
### What version of TiKV are you using?

TiKV 4.0 or master

### Steps to reproduce

Set TiKV to use the unified thread pool for both storage and coprocessor. Run a TP workload. Then, run large kv_scan requests.

### What did you expect?

The unified thread pool should let the kv_scan request not affect TP workload.

### What did happened?

It affects the performance of the TP workload now. Even if the priority of kv_scan is set to low, it can still affect TP performance.

sticnarf:
Now analyze requests and kv_scan requests are not transformed into splittable coroutines. Then, it is not possible for yatp to schedule such a large request.

sticnarf:
Typical affected cases: 

- TiDB analyze
- Add index DDL backfilling
- BR backup
- TiSpark analysis

The first three use cases should have a low concurrency, so the severity should be not high.
In the last case, TiSpark may have a high concurrency, TP performance may be affected greatly even if the scan requests are marked low priority.

BusyJay:
Should the change be rolled back?

sticnarf:
> Should the change be rolled back?

Oh. 4.0.9 is not released yet. I think we should not enable it by default now.

PR #8887, opened 2020-10-26
*: update yatp to reduce allocations in the multilevel mode and enable it for storage by default (#8008)

ti-srebot:
cherry-pick #8008 to release-4.0

---

### What problem does this PR solve?

Issue Number: close #7976 

Problem Summary:

The unified read pool is not enabled for the storage module.

### What is changed and how it works?

What's Changed:

Update yatp to include the fix https://github.com/tikv/yatp/pull/37. It reduces allocations when using the multilevel mode.

This PR also enables the unified thread pool for storage by default.

Previously the unified read pool is not enabled for the storage module because there is major performance regression report on a high-end machine. That's a special environment where there is TiDB and TiKV deployed on the same machine. TiKV and TiDB may interfere with each. If they are isolated, performance regression will be quite minor.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - There could be minor performance regression if point select is the only workload.

### Release note 

Enable the unified read pool for the storage module by default.

PR #9194, opened 2020-12-07
config: revert setting unified pool as default for storage

sticnarf:

Do not use unified pool for storage by default.

Due to https://github.com/tikv/tikv/issues/9188, setting the unified pool for storage by default can slow
down normal TP requests when there are TiSpark requests.

### Check List 

Tests 

- Unit test

### Release note 

- Revert part of https://github.com/tikv/tikv/pull/8887. Do not set the unified pool as storage default now.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1588,26 +1588,6 @@ macro_rules! readpool_config {
                 }
             }
 
-            pub fn use_unified_pool(&self) -> bool {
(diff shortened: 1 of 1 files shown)


hunterlxt (approved):
LGTM

BusyJay (approved):

ti-srebot (approved):

PR #9195, opened 2020-12-07
encryption: Cherry-pick log style file dictionary changes to 4.0.x

yiwu-arbug:

Cherry-picking #8865 and #9191 into 4.0 branch

Problem Summary:

What's Changed:
#8865 introduces an optimization to greatly reduce IO and mutex contention when encryption is enabled, by turning the file dictionary file into log style. The optimization is also useful to improve Lightning data import speed. We would like to include the optimization in 4.0.x branch as well.

However, the optimization introduce a data format that is not recognizable by TiKV <= 4.0.8. To address the problem, #9191 introduce the `security.encryption.enable-file-dictionary-log` config to allow turning off the optimization and convert to previous data format. The option will be default to true for TiKV >= 5.0, and default to off for TiKV < 5.0 but >= 4.0.9. User can downgrade TiKV cluster with encryption turned on to <= 4.0.8 with the following steps:

1. set `security.encryption.enable-file-dictionary-log = false` and restart TiKV. After TiKV successfully start, the data format is converted back to the same as before the optimization.
2. downgrade to version between (included) 4.0.0 and 4.0.8 (since encryption is not supported pre-4.0).

### Related changes

- #8865, #9191
- user docs: https://github.com/pingcap/docs/pull/4351

### Check List 

Tests 

- unit test

Side effects
- Breaking backward compatibility with TiKV <= 4.0.8 when encryption is used, but can be workaround by turning off `security.encryption.enable-file-dictionary-log` and restart TiKV.

### Release note 
* Reduce IO and mutex contention when encryption-at-rest is enabled. The change is backward incompatible. If user need to downgrade to versions prior to 4.0.9, `security.encryption.enable-file-dictionary-log` has to be turned off and TiKV has to be restarted once before being downgrade.


hunterlxt:
LGTM

Connor1996 (approved):
LGTM

ti-srebot (approved):

yiwu-arbug:
working on fix the test

yiwu-arbug:
`test_ingest_key_manager_delete_file_failed` failure does not seems to related to this PR and TDE in general. Checking with @glorv and @Little-Wallace.

Issue #9156, opened 2020-11-30
*: TiKV IO rate limiter - phase 1


yiwu-arbug:
## Development Task

The goal of the project is to provide a generic IO rate limiter within a TiKV process. Currently for separate tasks we have rate limiter of different form (e.g. rocksdb compaction IO rate limiter, BR/Lightning rate limiter). By providing a generate IO rate limiter we can have one single IO upper bound for all IO tasks. The project mainly target for users of cloud disk, such as EBS, where there's hard limit of IO bandwidth. But it would also be useful in general to have fine granularity control of TiKV IO behavior.

This issue tracks phase 1 of the IO rate limiter project, whose scope is described in the [spec](https://docs.google.com/document/d/1Jdsi7Ah1YtlZ3trCenqOL3BQ0FyeeubPJwyGwPeUdXM/edit?usp=sharing).

Tasks breakdown:
## interface
- [x] Rate limiter and file abstraction code skeleton @tabokie #9149 
## file system abstraction
- [x] File system abstraction - TiKV @tabokie #9149
- [x] File system abstraction - RocksDB @tabokie https://github.com/tikv/rocksdb/pull/216 https://github.com/tikv/rocksdb/pull/218 https://github.com/tikv/rust-rocksdb/pull/575 https://github.com/tikv/rust-rocksdb/pull/589 https://github.com/tikv/rust-rocksdb/pull/590
## tagging IO flows
- [x] Tag foreground read and coprocessor read @Connor1996 https://github.com/tikv/tikv/pull/9440
- [x] Tag foreground writes @Connor1996 https://github.com/tikv/tikv/pull/9440
- [x] Tag snapshot @Connor1996 https://github.com/tikv/tikv/pull/9440
- [x] Tag RocksDB flush and compaction @Connor1996 https://github.com/tikv/tikv/pull/9440
- [ ] Tag BR / Lightning / CDC
## measure IO throughput
- [x] Using bcc API to measure actual IO throughput @Connor1996 #9153 
- [x] Expose the metrics to Grafana, add dashboard @Connor1996 #9153
## rate limiter implementation
- [ ] Basic rate limiter implementation @tabokie #9197 
## integration test
- [ ] bank2
- [ ] Manual run TPCC and check metrics
- [ ] benchmark (surge of compaction, snapshot, GC) 
- [ ] benchmark (backup and restore, Lightning)

yiwu-arbug:
This is done as of v5.1.0.

PR #9197, opened 2020-12-07
Implement IO Rate Limiter

tabokie:

What's Changed:
Part of #9156, a priority-based limiter implementation.

TODOs:
- [x] unify sync/async semaphore to speedup synchronous call
- [x] optimize no-op code path
- [x] unit tests
- [x] microbenchmarks
- [x] add metrics for request wait

~Microbenchmark results:~
condvar wait_timeout (wait duration subtracted)
test condvar::tests::bench_condvar_async        ... bench:   1,108,886 ns/iter (+/- 138,515)
test condvar::tests::bench_condvar_sync         ... bench:   64,381 ns/iter (+/- 21,718)
test condvar::tests::bench_std_condvar          ... bench:   69,385 ns/iter (+/- 37,206)
limiter
test rate_limiter::tests::bench_acquire_limiter ... bench:          43 ns/iter (+/- 6)
test rate_limiter::tests::bench_limited_fast    ... bench:         702 ns/iter (+/- 833)
test rate_limiter::tests::bench_noop_limiter    ... bench:           1 ns/iter (+/- 0)
test rate_limiter::tests::bench_not_limited     ... bench:       1,008 ns/iter (+/- 1,081)

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
Implement a priority-based write I/O rate limiter.
This feature is controlled by `storage.io-rate-limit` configuration group.
```


tabokie:
Early feedback is welcome @yiwu-arbug @Connor1996.

tabokie:
Unit tests are still a little thin, but the main body is ready for review now. @yiwu-arbug @Connor1996

yiwu-arbug (commented):
sorry for delay. I will go through this PR tomorrow.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,318 @@
...
+fn get_priority(io_type: IOType) -> IOPriority {
+    match io_type {
+        IOType::Flush | IOType::Compaction => IOPriority::Low,
+        _ => IOPriority::High,
+    }
+}
+
+/// Used to calibrate actual bytes through after last reset.
+#[derive(Debug)]
+struct BytesCalibrator {
```
yiwu-arbug, line 25: Let's make this a trait. The default implementation will consume IO budget on request.

components/file_system/src/lib.rs
```diff
@@ -25,26 +29,28 @@ use std::sync::{Arc, Mutex};
...
-    Write,
-    Coprocessor,
-    Flush,
-    Compaction,
-    Replication,
-    LoadBalance,
-    Import,
-    Export,
+    Other = 0,
+    Read = 1,
```
Connor1996, line 41: is it really necessary?

components/file_system/src/lib.rs
```diff
@@ -25,26 +29,28 @@ use std::sync::{Arc, Mutex};
...
+    Coprocessor = 3,
+    Flush = 4,
+    Compaction = 5,
+    Replication = 6,
+    LoadBalance = 7,
+    Import = 8,
+    Export = 9,
 }
 
+pub const IO_TYPE_VARIANTS: usize = 10;
```
Connor1996, line 52: proc macro can help you do this, see https://docs.rs/variant_count/1.0.0/variant_count/

components/file_system/src/lib.rs
```diff
@@ -266,13 +272,33 @@ impl<R: Read> Read for Sha256Reader<R> {
     }
 }
 
+const SPACE_PLACEHOLDER_FILE: &str = "space_placeholder_file";
+
+// create a file with hole, to reserve space for TiKV.
+pub fn reserve_space_for_recover<P: AsRef<Path>>(data_dir: P, file_size: u64) -> io::Result<()> {
```
Connor1996, line 278: why move to this crate

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,350 @@
...
+                continue;
+            }
+            if let Some(calibrator) = &self.calibrator {
+                let calibrated = calibrator.calibrate();
+                self.consumed.store(calibrated, Ordering::Relaxed);
+                if calibrated < cached_bytes_per_refill {
+                    continue;
+                }
+            }
+            let now = Instant::now();
```
Connor1996, line 171: `now_coarse()` is enough?

components/file_system/src/lib.rs
```diff
@@ -25,26 +29,28 @@ use std::sync::{Arc, Mutex};
...
-    Coprocessor,
-    Flush,
-    Compaction,
-    Replication,
-    LoadBalance,
-    Import,
-    Export,
+    Other = 0,
+    Read = 1,
+    Write = 2,
```
Connor1996, line 42: I want to change these two types cause it's confusing with IOOp. Maybe call it `Storage` and `Raftstore`?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,350 @@
...
 
-    /// Request for token for bytes and potentially update statistics. If this
+    /// Requests for token for bytes and potentially update statistics. If this
     /// request can not be satisfied, the call is blocked. Granted token can be
     /// less than the requested bytes, but must be greater than zero.
-    pub fn request(&self, _io_type: IOType, _io_op: IOOp, bytes: usize) -> usize {
-        std::cmp::min(self.refill_bytes.load(atomic::Ordering::Relaxed), bytes)
+    pub fn request(&self, io_type: IOType, io_op: IOOp, bytes: usize) -> usize {
+        let prio = get_priority(io_type);
+        let mut bytes = self.total_limiters[IOType::Other as usize].request(bytes, prio);
```
Connor1996, line 287: why use `IOType::Other`

components/file_system/src/lib.rs
tabokie: So tikv_util doesn't depend on file_system anymore.

Connor1996: seems there is a lot of `std::fs` need to be replaced? it depends on file_system anyway

components/file_system/src/rate_limiter.rs
tabokie: Here uses `Other` slot to store total limiter, which is the sum of all r/w IOs. Just for convenience because `Other` itself doesn't own a dedicated limiter.

components/file_system/src/lib.rs
yiwu-arbug: I think this is good. Making the enum -> int mapping explicit.

components/file_system/src/lib.rs
tabokie: What do you mean? file_system now depends on tikv_util to use the time_util, therefore tikv_util has some uses of std::fs that can't be pruned.

components/file_system/src/lib.rs
tabokie: As per the original proposal, `IOType::Read` refers to foreground read query (I believe we have read request going through raftstore too?), and `IOType::Write` refers to write transaction (kvdb WAL included). So `Storage` and `Raftstore` isn't ideal either.

tabokie: I think the confusion part exists but is manageable, because `IOOp` is only exposed to core components related to file_system.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,363 @@
...
 
-/// No-op limiter
 /// An instance of `IORateLimiter` should be safely shared between threads.
 #[derive(Debug)]
 pub struct IORateLimiter {
-    refill_bytes: atomic::AtomicUsize,
+    // IOType::Other slot is used to store type-less limiter
+    write_limiters: [PerTypeIORateLimiter; IOType::VARIANT_COUNT],
+    read_limiters: [PerTypeIORateLimiter; IOType::VARIANT_COUNT],
+    total_limiters: [PerTypeIORateLimiter; IOType::VARIANT_COUNT],
```
Connor1996, line 268: Is there any case for limiting read+write IO? I think we just need to limit write/read separately?

tabokie: I think cloud disk for instance requires to limit total throughput.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,363 @@
...
+        if cached_bytes_per_refill == 0 {
+            return bytes;
+        }
+        if priority == IOPriority::High {
+            self.consumed.fetch_add(bytes, Ordering::Relaxed);
+            // Unlimited requestor don't do refills themselves, therefore the first
+            // limited IO in long period will take penalty for a mandotory refill.
+            return bytes;
+        }
+        loop {
```
Connor1996, line 167: maybe need to check whether `bytes_per_refill` is updated every loop?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,363 @@
...
+                break bytes;
+            }
+            let mut last_refill_time = self.last_refill_time.lock().unwrap();
+            // double check if bytes have been refilled by others
+            if self.consumed.load(Ordering::Relaxed) < cached_bytes_per_refill {
+                continue;
+            }
+            if let Some(calibrator) = &self.calibrator {
+                let calibrated = calibrator.calibrate();
+                self.consumed.store(calibrated, Ordering::Relaxed);
```
Connor1996, line 178: not quite understand the logic about `calibrator`

components/file_system/src/rate_limiter.rs
```diff
@@ -1,43 +1,363 @@
...
+    }
+
+    #[inline]
+    fn refill_and_request(&self, bytes: usize) -> usize {
+        let token = std::cmp::min(self.bytes_per_refill.load(Ordering::Relaxed), bytes);
+        self.consumed.store(token, Ordering::Relaxed);
+        if let Some(calibrator) = &self.calibrator {
+            calibrator.reset();
+        }
+        self.condv.notify_all();
```
Connor1996, line 152: if the consumed is already `bytes_per_refill`, no need to notify? or just notify_one

components/file_system/src/condvar.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// timing out after a specified duration.
+    pub fn wait_timeout<'a, T>(
+        &self,
+        guard: MutexGuard<'a, T>,
+        timeout: Duration,
+    ) -> (MutexGuard<'a, T>, bool) {
+        // mutable just to indulge NonNull
+        let mut node = CondvarNode::new_sync();
+        self.enqueue(&mut node);
+        // alternative: std::thread::park_timeout suffers from spurious wake
```
Connor1996, line 156: consider using https://docs.rs/parking_lot/0.11.1/parking_lot/struct.Condvar.html

tabokie: parking_lot::Condvar is used in combo with parking_lot::Mutex, I would rather our implementation of Condvar can integrate well with standard Mutex. Also, I think a fair portion of it's optimizations aim at multiple waiter on the same condvar, which isn't touched by our use case (linked wait).

components/file_system/src/rate_limiter.rs
tabokie: Good point, this relates to a part I haven't decided yet. Say we have a thread always asking for large chunks of bytes and in the exact pace as refilling, it will eventually starve other threads, calling `notify_one` won't help either. The best approach would be setting some limits for the bytes we are giving out to each request.

components/file_system/src/rate_limiter.rs
tabokie: For read limiter, it calculates bytes consumption based on buffered read initially. When the quotas are drained, it will call calibrator (if any) to calibrate the consumption estimation against some ground truths. This design is to reduce calibrating frequency.

components/file_system/src/condvar.rs
Connor1996: you can use parking_lot::Mutex instead of standard Mutex as parking_lot::Mutex is 1.5x faster than std::sync::Mutex when uncontended

components/file_system/src/condvar.rs
```diff
@@ -0,0 +1,299 @@
...
+/// waking up all nodes at once, sort them to a seperate waiting queue and only
+/// wake up the leader.
+#[derive(Debug)]
+pub struct Condvar {
+    head: Cell<Option<NonNull<CondvarNode>>>,
+    tail: Cell<Option<NonNull<CondvarNode>>>,
+}
+
+unsafe impl Send for Condvar {}
+unsafe impl Sync for Condvar {}
```
Connor1996, line 89: How do you promise the thread-safety? 
By the mutex of `last_refill_time` to make sure `enqueue` and `dequeue` not called in concurrency?

components/file_system/src/rate_limiter.rs
Connor1996: the request would be placed at the tail of the queue, so that seems it won't starve other threads?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,57 +1,634 @@
...
+        token
+    }
+
+    pub fn request(&self, bytes: usize, priority: IOPriority) -> usize {
```
Connor1996, line 157: how about writing like `block_on(async_request)` to reuse the logic

components/file_system/src/condvar.rs
```diff
@@ -0,0 +1,303 @@
...
+        mu: &'a Mutex<T>,
+        guard: MutexGuard<'b, T>,
+        timeout: Duration,
+    ) -> (MutexGuard<'a, T>, bool) {
+        let mut node = CondvarNode::new_async();
+        self.enqueue(&mut node);
+        std::mem::drop(guard);
+        let timed_out = {
+            let f = match node {
+                CondvarNode::Async(ref sem, _) => sem.acquire().fuse(),
```
Connor1996, line 180: use tokio::sync::Notify?

components/file_system/src/condvar.rs
tabokie: Yes, that's why condvar is always paired with a mutex.

components/file_system/src/rate_limiter.rs
tabokie: block_on a async function has a considerable overhead. The microbenchmark also shows a nearly 2x latency.

components/file_system/src/condvar.rs
tabokie: Seems the only difference is whether the notifications can accumulate, not sure the performance-wise which is the winner. I'll leave a todo here for future reference.

components/file_system/src/rate_limiter.rs
tabokie: Not if this thread with "perfect" timing can always do the refill itself before others waking up. Also besides this case, generally it's negative for fair scheduling. e.g. queries on wide table take more quotas than those on normal table.

components/file_system/src/condvar.rs
```diff
@@ -177,7 +177,7 @@ impl Condvar {
         std::mem::drop(guard);
         let timed_out = {
             let f = match node {
-                CondvarNode::Async(ref sem, _) => sem.acquire().fuse(),
+                CondvarNode::Async(ref not, _) => not.notified().fuse(),
```
Connor1996, line 180: Can we use `notified().await`? Actually, I don't understand what's the purpose of `fuse()` and the `select!`.

Connor1996: oh, seems there is no good way to await within a timeout?

Connor1996: seems https://docs.rs/tokio/1.0.1/tokio/time/fn.timeout.html timeout would help you to await within a timeout

tabokie: nice

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -146,52 +145,19 @@ pub fn get_io_type() -> IOType {
...
-                        })
-                        .or_insert(stats.clone());
-                }
-
-                return self.io_stats_map.replace(map).unwrap();
+pub(crate) fn fetch_io_bytes(io_type: IOType) -> IOBytes {
+    unsafe {
+        if let Some(ctx) = BPF_CONTEXT.as_mut() {
+            let mut buf = [0 as u8; std::mem::size_of::<IOType>()];
+            std::ptr::write(buf.as_mut_ptr() as *mut IOType, io_type);
```
Connor1996, line 152: this may cause write-write conflict

tabokie: You mean `buf`? it's a local variable.

Connor1996: oh sorry, I misunderstood it.

tabokie:
Added a new metrics to monitor request wait:

image

components/file_system/src/rate_limiter.rs
```diff
@@ -1,133 +1,362 @@
...
+                    let wait = DEFAULT_REFILL_PERIOD - since_last_refill;
+                    RATE_LIMITER_REQUEST_WAIT_DURATION
+                        .with_label_values(&[$priority.as_str()])
+                        .observe(wait.as_secs_f64());
+                    sleep_impl!(wait, $mode);
+                }
+                // our request is already recorded in `pending_ios`.
+                if pending <= cached_ios_per_refill {
+                    return amount;
+                }
```
yiwu-arbug, line 130: can we do some refactoring to move these lines out of the `if Some(locked)` block, so to avoid `drop(locked)`?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,133 +1,362 @@
...
+    };
+    ($duration:expr, "async") => {
+        tokio::time::delay_for($duration).await
+    };
+    ($duration:expr, "non-blocking") => {
+        return 0
+    };
+}
+
+macro_rules! request_impl {
```
yiwu-arbug, line 95: Add some inline comment to describe the approach here?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,133 +1,362 @@
...
+}
+
 /// An instance of `IORateLimiter` should be safely shared between threads.
 #[derive(Debug)]
 pub struct IORateLimiter {
-    refill_bytes: usize,
-    enable_statistics: bool,
+    priority_map: [IOPriority; IOType::VARIANT_COUNT],
+    bytes: Arc<RawIORateLimiter>,
+    // TODO: 1) move this to background pool. 2) pause it when limit is toggled off.
```
yiwu-arbug, line 264: also 3. stop on server shutdown?

components/file_system/src/lib.rs
```diff
@@ -103,6 +104,80 @@ impl std::ops::Sub for IOBytes {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Copy, VariantCount)]
+pub enum IOPriority {
+    Low = 0,
+    Medium = 1,
+    High = 2,
+}
+
+impl IOPriority {
```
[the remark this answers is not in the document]
tabokie: Added `FromStr`, `as_str` rather than `into_str` is still needed though.

components/file_system/src/rate_limiter.rs
tabokie: I think it will be shutdown when Limiter is dropped, or managed by background pool.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,133 +1,362 @@
...
+    ($duration:expr, "non-blocking") => {
+        return 0
+    };
+}
+
+macro_rules! request_impl {
+    ($self:expr, $priority:expr, $amount:expr, $mode:tt) => {{
+        let priority_idx = $priority as usize;
+        loop {
+            let cached_ios_per_refill = $self.ios_per_sec[priority_idx].load(Ordering::Relaxed);
```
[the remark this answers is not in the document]
tabokie: Now that we don't have complex logic for high priority IOs, I think it's a good idea. Since EBS operates on a much lower level, the performance should be better (e.g. latency being more predictable)

tabokie:
Deviation of actual usage from user limit within 5%:

image

components/engine_rocks/src/write_batch.rs
```diff
@@ -72,6 +73,7 @@ impl engine_traits::WriteBatch<RocksEngine> for RocksWriteBatch {
     }
 
     fn write_opt(&self, opts: &WriteOptions) -> Result<()> {
+        let _io_type_guard = WithIOType::new(IOType::ForegroundWrite);
```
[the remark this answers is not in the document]
Connor1996: I don't understand why doing this. It's called under `ForegroundWrite` already

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
-        let stats = limiter.statistics();
-        set_io_rate_limiter(Some(limiter));
-        (
-            WithIORateLimit {
-                previous_io_rate_limiter,
-            },
-            stats,
-        )
-    }
+pub struct LocalIORateLimiterDaemon {
```
Connor1996, line 344: only used in test?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+/// Actual implementation for requesting IOs from RawIORateLimiter.
+/// An attempt will be recorded first. If the attempted amount exceeds the available quotas of
+/// current epoch, the requester will register itself and sleep until next epoch.
+macro_rules! request_imp {
```
Connor1996, line 134: why can't make it a funcation?

tabokie: Because a function can't be both synchronous and asynchronous, macro is the only way to dedup these code.

yiwu-arbug (commented):
sorry for delay. I'm trying to read the whole PR again but haven't finished. Some comments here may be incorrect, as I didn't fully understand the code.

components/file_system/src/metrics.rs
```diff
@@ -49,4 +49,13 @@ lazy_static! {
             &["type", "op"],
             exponential_buckets(1.0, 2.0, 22).unwrap() // max 4s
         ).unwrap();
+
+    pub static ref RATE_LIMITER_REQUEST_WAIT_DURATION: HistogramVec = register_histogram_vec!(
+            "tikv_rate_limiter_request_wait_duration",
+            "Bucketed histogram of IO rate limiter request wait duration",
+            &["type"],
+            // 0 ~ 1s, 50ms resolution
+            linear_buckets(0.0, 0.05, 22).unwrap()
```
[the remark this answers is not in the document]
tabokie: I prefer linear because wait duration is bounded within `refill_period`, there is no need to sacrifice resolution for larger range. Speaking of which I might reconfigure this bucket again since `refill_period` has been updated to finer grain.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
+            }
+            let now = Instant::now_coarse();
+            let (next_refill_time, pending) = {
+                let mut locked = $self.protected.lock();
+                // a small delay in case a refill slips in after `ios_per_epoch` was fetched.
+                if locked.next_refill_time + Duration::from_millis(1) >= now + DEFAULT_REFILL_PERIOD
+                {
+                    continue;
+                }
+                locked.pending_ios[priority_idx] += amount;
```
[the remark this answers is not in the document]
tabokie: If it's refill time, the request thread won't block either.

components/file_system/src/lib.rs
```diff
@@ -45,20 +49,20 @@ pub enum IOOp {
...
-    LoadBalance,
-    Gc,
-    Import,
-    Export,
+    ForegroundWrite = 2,
+    Flush = 3,
+    Compaction = 4,
+    Replication = 5,
+    LoadBalance = 6,
+    Gc = 7,
```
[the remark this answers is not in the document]
tabokie: I'm not familiar with that, but it seems that filter only makes decisions without issuing IOs? It would be unfair to tag the entire compaction as GC just for a few keys being GCed.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
+
+/// Limit total IO flow below provided threshold by throttling lower-priority IOs.
+/// Rate limit is disabled when total IO threshold is set to zero.
+#[derive(Debug)]
+struct RawIORateLimiter {
+    // IO amount passed through within current epoch
+    ios_through: [CachePadded<AtomicUsize>; IOPriority::COUNT],
+    // Maximum IOs permitted within current epoch
+    ios_per_epoch: [CachePadded<AtomicUsize>; IOPriority::COUNT],
+    protected: Mutex<RawIORateLimiterProtected>,
```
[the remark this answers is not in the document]
tabokie: `park` seems like an overkill with ordering and signaling ability to cope with this racy case. IMO a optional spin when wait duration is extremely small would probably serve better to reduce syscalls.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
+                }
+                locked.pending_ios[priority_idx] += amount;
+                (locked.next_refill_time, locked.pending_ios[priority_idx])
+            };
+            if next_refill_time > now {
+                let wait = next_refill_time - now;
+                RATE_LIMITER_REQUEST_WAIT_DURATION
+                    .with_label_values(&[$priority.as_str()])
+                    .observe(wait.as_secs_f64());
+                do_sleep!(wait, $mode);
```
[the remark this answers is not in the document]
tabokie: That'a a good thought, right now I don't have a solution in mind. Further on we could probably use a fallback queue to manage those highly throttled request in order, then again, the switching between two modes would be painful.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
+            let ios_through =
+                $self.ios_through[priority_idx].fetch_add(amount, Ordering::AcqRel) + amount;
+            if ios_through <= cached_ios_per_refill {
+                return amount;
+            }
+            let now = Instant::now_coarse();
+            let (next_refill_time, pending) = {
+                let mut locked = $self.protected.lock();
+                // a small delay in case a refill slips in after `ios_per_epoch` was fetched.
+                if locked.next_refill_time + Duration::from_millis(1) >= now + DEFAULT_REFILL_PERIOD
```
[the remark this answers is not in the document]
tabokie: In order to speed up this specific condition? Because a mutex seems necessary to make sure a `CAS`-like semantics between `refill_time` and `pending_ios`

cmd/src/server.rs
```diff
@@ -862,31 +860,31 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+            .build(!io_snooper_on /*enable_statistics*/);
+        let fetcher = if io_snooper_on {
+            BytesFetcher::FromIOSnooper()
         } else {
-            let limiter = IORateLimiter::new(0, true);
-            (BytesFetcher::FromRateLimiter(limiter.statistics()), limiter)
+            BytesFetcher::FromRateLimiter(limiter.statistics().unwrap())
         };
         set_io_rate_limiter(Some(Arc::new(limiter)));
+        start_global_io_rate_limiter_daemon(&self.background_worker);
```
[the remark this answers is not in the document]
tabokie: I added a few lines to make sure `refill` can't be delayed for more than half of `refill_period`. PTAL

components/file_system/src/lib.rs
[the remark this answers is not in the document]
yiwu-arbug: The compaction filter only filter write CF, then use the info and send a GC task to another thread pool to GC the default CF. The GC task is done by writing rocksdb.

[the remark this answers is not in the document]
tabokie: So these GC tasks are scheduled in normal gc threads? Then they should be tagged by `GcRunner`.

components/file_system/src/rate_limiter.rs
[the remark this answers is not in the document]
yiwu-arbug: I'm wondering if we can refactor the code using parking_lot queue (by using park/unpark primitives). But we can leave it for later.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,568 @@
...
+                locked.pending_bytes[priority_idx] += amount;
+                (locked.next_refill_time, locked.pending_bytes[priority_idx])
+            };
+            if next_refill_time > now {
+                let wait = next_refill_time - now;
+                RATE_LIMITER_REQUEST_WAIT_DURATION
+                    .with_label_values(&[$priority.as_str()])
+                    .observe(wait.as_secs_f64());
+                do_sleep!(wait, $mode);
+            } else if next_refill_time + DEFAULT_REFILL_PERIOD / 2 < now {
```
yiwu-arbug, line 166: let's see if we can get rid of background task, and use the request thread to do the refill.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,568 @@
...
+        if locked.next_refill_time > now + DEFAULT_REFILL_PERIOD / 2 {
+            // already refilled
+            return;
+        }
+
+        // keep in sync with a potentially skewed clock
+        locked.next_refill_time = now + DEFAULT_REFILL_PERIOD;
+
+        debug_assert!(IOPriority::High as usize > IOPriority::Medium as usize);
+        for p in &[IOPriority::High, IOPriority::Medium] {
```
yiwu-arbug, line 231: inline comment each iteration is to calculate the budget for next priority.

components/file_system/src/metrics.rs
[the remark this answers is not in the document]
tabokie: Changing to exponential since we record total-wait now.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,527 @@
...
+
+const DEFAULT_REFILL_PERIOD: Duration = Duration::from_millis(50);
+
+/// Limit total IO flow below provided threshold by throttling lower-priority IOs.
+/// Rate limit is disabled when total IO threshold is set to zero.
+#[derive(Debug)]
+struct PriorityBasedIORateLimiter {
+    // IO amount passed through within current epoch
+    bytes_through: [CachePadded<AtomicUsize>; IOPriority::COUNT],
+    // Maximum IOs permitted within current epoch
```
Connor1996, line 80:
```suggestion
    // Maximum bytes permitted within current epoch
```

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,527 @@
...
+    };
+}
+
+const DEFAULT_REFILL_PERIOD: Duration = Duration::from_millis(50);
+
+/// Limit total IO flow below provided threshold by throttling lower-priority IOs.
+/// Rate limit is disabled when total IO threshold is set to zero.
+#[derive(Debug)]
+struct PriorityBasedIORateLimiter {
+    // IO amount passed through within current epoch
```
Connor1996, line 78:
```suggestion
    // Total bytes passed through within current epoch
```

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,527 @@
...
+            return amount;
+        }
+        let now = Instant::now_coarse();
+        let mut wait = Duration::from_millis(0);
+        // hold a snapshot ticket of pending bytes
+        let pending_snapshot = {
+            let mut locked = $limiter.protected.lock();
+            locked.pending_bytes[priority_idx] += amount;
+            let pending_snapshot = locked.pending_bytes[priority_idx];
+            println!(
```
Connor1996, line 130: use `debug!`?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,527 @@
...
+    bytes_through: [CachePadded<AtomicUsize>; IOPriority::COUNT],
+    // Maximum IOs permitted within current epoch
+    bytes_per_epoch: [CachePadded<AtomicUsize>; IOPriority::COUNT],
+    protected: Mutex<PriorityBasedIORateLimiterProtected>,
+}
+
+#[derive(Debug)]
+struct PriorityBasedIORateLimiterProtected {
+    next_refill_time: Instant,
+    // IOs that are can't be fulfilled in current epoch
```
Connor1996, line 88:
```suggestion
    // Bytes that are can't be fulfilled in current epoch
```

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,527 @@
...
+            $limiter.bytes_through[priority_idx].fetch_add(amount, Ordering::AcqRel) + amount;
+        if bytes_through <= cached_bytes_per_refill {
+            return amount;
+        }
+        let now = Instant::now_coarse();
+        let mut wait = Duration::from_millis(0);
+        // hold a snapshot ticket of pending bytes
+        let pending_snapshot = {
+            let mut locked = $limiter.protected.lock();
+            locked.pending_bytes[priority_idx] += amount;
```
Connor1996, line 128: should it be `amount - (cached_bytes_per_refill - bytes_through)`?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,527 @@
...
+        let amount = std::cmp::min($amount, cached_bytes_per_refill);
+        let bytes_through =
+            $limiter.bytes_through[priority_idx].fetch_add(amount, Ordering::AcqRel) + amount;
+        if bytes_through <= cached_bytes_per_refill {
+            return amount;
+        }
+        let now = Instant::now_coarse();
+        let mut wait = Duration::from_millis(0);
+        // hold a snapshot ticket of pending bytes
+        let pending_snapshot = {
```
Connor1996, line 126: what do you mean by `snapshot`?

tabokie: Added a comment. Basically we use the `pending_bytes` value at this point to deduce when this slice of newly added pending request can actually be served.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,558 @@
...
-        for i in self.read.iter() {
-            i.store(0, Ordering::Relaxed);
+    fn set_bytes_per_sec(&self, bytes_per_sec: usize) {
+        let now = (bytes_per_sec as f64 * DEFAULT_REFILL_PERIOD.as_secs_f64()) as usize;
+        let before = self.bytes_per_epoch[IOPriority::High as usize].swap(now, Ordering::Relaxed);
+        if before == 0 || now == 0 {
+            // toggle on/off rate limit.
+            // we hold this lock so a concurrent refill can't negate our effort.
+            let _locked = self.protected.lock();
+            for p in &[IOPriority::Medium, IOPriority::Low] {
```
Connor1996, line 178: why change medium and low only when zero

components/engine_rocks/src/write_batch.rs
```diff
@@ -72,6 +73,9 @@ impl engine_traits::WriteBatch<RocksEngine> for RocksWriteBatch {
     }
 
     fn write_opt(&self, opts: &WriteOptions) -> Result<()> {
+        // Indiscriminately increase the priority because RocksDB merges write
+        // batches with different IO types.
+        let _io_type_guard = WithIOType::new(IOType::ForegroundWrite);
```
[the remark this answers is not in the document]
tabokie: GC

components/file_system/src/rate_limiter.rs
tabokie: Limits are propagated from high priority to low priority in `refill`. Even modified here, they will be overwritten later.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,570 @@
...
+    ///    total IO flow will exceed global threshold.
+    /// 3) Highest priority IO alone must not exceed global threshold.
+    fn refill(&self, locked: &mut MutexGuard<PriorityBasedIORateLimiterProtected>, now: Instant) {
+        const UPDATE_BUDGETS_EVERY_N_EPOCHS: usize = 5;
+        // Epochs can be skipped for refill, which happens fairly rare under production workload.
+        // We will ignore these epochs when making decision.
+        // assuming clock can be skewed for no more than half an epoch
+        let skipped_epochs = ((now - locked.next_refill_time).as_secs_f64()
+            / DEFAULT_REFILL_PERIOD.as_secs_f64()
+            + 0.5) as usize;
```
tabokie, line 218: Right now the unit tests could pass, but I'm still not very confident in capturing all the corner cases. Generally I'm leaning towards using a separate daemon refill thread so to not worry about `skipped_epochs` or other unrevealed racy cases.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+        // acquire a snapshot of pending bytes, which denotes a position of logical queue
+        // for this request to wait in.
+        let pending_snapshot = {
+            let mut locked = $limiter.protected.lock();
+            // already served partially by current epoch when consumption overflow bytes is
+            // smaller than requested amount
+            let remains = std::cmp::min(bytes_through - cached_bytes_per_refill, amount);
+            // when there is a recent refill, double check if bytes consumption has been reset
+            if now + DEFAULT_REFILL_PERIOD < locked.next_refill_time + Duration::from_millis(1)
+                && $limiter.bytes_through[priority_idx].fetch_add(remains, Ordering::Relaxed)
```
yiwu-arbug, line 236: The second condition `bytes_through + remains <= cached_bytes_per_epoch` alone is probably enough?

components/file_system/src/rate_limiter.rs
yiwu-arbug: If it simplifies the implementation, let's do that. but how about use a separate thread, instead of the bg_worker pool?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+            let pending_snapshot = locked.pending_bytes[priority_idx];
+            if locked.next_refill_time <= now {
+                $limiter.refill(&mut locked, now);
+            } else {
+                wait += locked.next_refill_time - now;
+            }
+            pending_snapshot
+        };
+        // calculate wait duration with queue_length / service_rate.
+        wait += DEFAULT_REFILL_PERIOD * ((pending_snapshot - 1) / cached_bytes_per_refill) as u32;
```
yiwu-arbug, line 252: why `-1`?

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+    ///    consume the same IO amount as the last few epochs within global threshold.
+    /// 2) Higher priority may temporarily use lower priority's IO budgets. When this happens,
+    ///    total IO flow could exceed global threshold.
+    /// 3) Highest priority IO alone must not exceed global threshold.
+    fn refill(&self, locked: &mut MutexGuard<PriorityBasedIORateLimiterProtected>, now: Instant) {
+        const UPDATE_BUDGETS_EVERY_N_EPOCHS: usize = 5;
+        // Epochs can be skipped for refill, which happens fairly rare under production workload.
+        // We will ignore these epochs when making decision.
+        // assuming clock can be skewed for no more than half an epoch
+        let skipped_epochs = ((now - locked.next_refill_time).as_secs_f64()
```
yiwu-arbug, line 308: `debug_assert!(now >= next_refill_time);`

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+    }
+
+    /// Update and refill IO budgets for next epoch based on IO priority.
+    /// Here we provide best-effort priority control:
+    /// 1) Limited IO budget is assigned to lower priority to make sure higher priority can always
+    ///    consume the same IO amount as the last few epochs within global threshold.
+    /// 2) Higher priority may temporarily use lower priority's IO budgets. When this happens,
+    ///    total IO flow could exceed global threshold.
+    /// 3) Highest priority IO alone must not exceed global threshold.
+    fn refill(&self, locked: &mut MutexGuard<PriorityBasedIORateLimiterProtected>, now: Instant) {
```
yiwu-arbug, line 303:
```suggestion
    fn refill(&self, protected: &mut PriorityBasedIORateLimiterProtected, now: Instant) {
```

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+        // Epochs can be skipped for refill, which happens fairly rare under production workload.
+        // We will ignore these epochs when making decision.
+        // assuming clock can be skewed for no more than half an epoch
+        let skipped_epochs = ((now - locked.next_refill_time).as_secs_f64()
+            / DEFAULT_REFILL_PERIOD.as_secs_f64()
+            + 0.5) as usize;
+        // keep in sync with a potentially skewed clock
+        locked.next_refill_time = now + DEFAULT_REFILL_PERIOD;
+        let mut limit = self.bytes_per_epoch[IOPriority::High as usize].load(Ordering::Relaxed);
+        if limit == 0 {
```
yiwu-arbug, line 314: we shall let the refill thread to know limit is disabled and not waiting anymore.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+            self.write_bytes[i].store(0, Ordering::Relaxed);
         }
     }
+}
 
-    #[allow(dead_code)]
-    pub fn reset(&self) {
-        for i in self.read.iter() {
-            i.store(0, Ordering::Relaxed);
+const DEFAULT_REFILL_PERIOD: Duration = Duration::from_millis(50);
```
yiwu-arbug, line 151: nit: move all constants to the top of the file?

components/file_system/src/rate_limiter.rs
tabokie: This tries to optimize out a `fetch_add` for most cases.

components/file_system/src/rate_limiter.rs
tabokie: Ideally it should be `pending_snapshot.saturating_sub(cached_bytes_per_refill)` because we have already count the wait until next epoch. I'll add an explanatory comment.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,148 +1,723 @@
...
+        let priority_idx = $priority as usize;
+        let cached_bytes_per_refill =
+            $limiter.bytes_per_epoch[priority_idx].load(Ordering::Relaxed);
+        // flow control is disabled when limit is zero
+        if cached_bytes_per_refill == 0 {
+            return $amount;
+        }
+        let amount = std::cmp::min($amount, cached_bytes_per_refill);
+        let bytes_through =
+            $limiter.bytes_through[priority_idx].fetch_add(amount, Ordering::AcqRel) + amount;
```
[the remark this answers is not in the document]
tabokie: It's actually not very useful, removing them now.

components/file_system/src/rate_limiter.rs
tabokie: Ideally it should be `ceiling(pending_snapshot.saturating_sub(cached_bytes_per_refill) as f32 / cached_bytes_per_refill)`, will add a comment.

components/file_system/src/rate_limiter.rs
tabokie: I think the unreproducible bugs in CI has been fixed now. Maybe a separate refiller isn't needed after all.

yiwu-arbug (commented):
Looking good to me. @tabokie do you have other pending changes?

tabokie:
> Looking good to me. @tabokie do you have other pending changes?

No other changes at the moment. I plan to do some more tests before merging it.

yiwu-arbug (approved):
Awesome work!

tabokie:
Latest updates contains:
1) Stop altering IOType in `write_opt` to avoid erasing the type information. Instead change GC priority to high.
2) Account for skipped epochs when calculating budgets changes. And use max filter instead of avg filter because the latter isn't accurate enough in certain workloads.
3) Limit the wait duration of one single request by returning early with partial quotas. This is to avoid a temporary spike from high-priority IOs causes an indefinite freeze in low-priority IOs.

src/storage/config.rs
```diff
@@ -333,6 +334,14 @@ impl IORateLimitConfig {
             );
             self.other_priority = IOPriority::High;
         }
+        if self.gc_priority != self.foreground_write_priority {
+            warn!(
+                "GC writes are merged with foreground writes. To avoid priority inversion, change \
+                  priority for IOType::Gc from {:?} to {:?}",
+                self.gc_priority, self.foreground_write_priority,
+            );
+            self.gc_priority = self.foreground_write_priority;
```
yiwu-arbug, line 343: When compaction filter is enabled, does it affect compaction IO priority?

tabokie: No, the `Gc` type guard is only scoped around actual writes without contaminating compaction thread.

tabokie:
To avoid accumulating RocksDB IO backlog (measured with memtable count, L0 count, pending bytes) which eventually leads to write stalls, several potential solutions are:
1. control feedback loop between RocksDB backlog metrics and their corresponding IO threshold
2. smart IO calculation that deduces the right amount of IO needed for background work for given foreground IO
3. escalate background IO requests when write stall happens or is about to

(1) is relatively easy to implement but unable to respond to high-frequency target such as memtable/L0 (e.g. 100MB/s write flow can easily creates 200 L0 files in 10s); (2) breaks the current "priority" abstraction and will be a pain to maintain; (3) causes jitter which contradicts our original intent here.

Given the situation, my current design will simply move "flush" which accounts for memtable backlog, and "level-zero-compaction" which accounts for L0 backlog to high-priority group; and use a probabilistic method which converts the amount of pending-bytes backlog to a probability for "compaction" request to "jump" to high-priority.

tabokie:
@yiwu-arbug @Connor1996 This PR is ready for yet another review now.
In latest code LevelZeroCompaction is of medium priority, and Compaction is of low priority, the rate limit is configured as R+W total threshold.
I'm also adjusting rate limit to `rw_rate_limit * (1 + pending_bytes/pending_bytes_soft_limit) / 2`, and this adjustment is only visible to low-priority IOs. The rationale is write IOs is normally at least half of total IOs, when pending bytes are low, we can restrain low-priority IOs towards that target.

tabokie:
I'm squashing all the housekeeping changes into this PR: #10223, for the ease of review.

yiwu-arbug (approved):
Is this ready again after merging #10223? LGTM overall.

components/server/src/server.rs
```diff
@@ -1257,3 +1285,97 @@ impl<R: RaftEngine> EngineMetricsManager<R> {
         }
     }
 }
+
+pub struct EnginesResourceInfo {
+    kv_engine: RocksEngine,
+    raft_engine: Option<RocksEngine>,
+    normalized_num_mem_tables: MovingMaxU32,
```
Connor1996, line 1292: why these statistics are collected in moving max instead of moving average

tabokie: Originally use `max` to avoid hitting write stalls. Updated to use `max(avg, current)`.

tabokie:
@Connor1996 @yiwu-arbug We're good to go.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,74 +1,431 @@
...
+            } else {
+                // `(a-1)/b` is equivalent to `roundup(a.saturating_sub(b)/b)`.
+                locked.next_refill_time - now
+                    + DEFAULT_REFILL_PERIOD
+                        * ((locked.pending_bytes[priority_idx] - 1) / cached_bytes_per_epoch) as u32
+            }
+        };
+        if wait > MAX_WAIT_DURATION_PER_REQUEST {
+            // Long wait duration could freeze request thread not to react to latest budgets
+            // adjustment. Exit early by returning partial quotas.
```
Connor1996, line 264: what will the caller do when the write quota is partially satisified.

components/file_system/src/rate_limiter.rs
```diff
@@ -1,74 +1,431 @@
...
+    /// 3) Highest priority IO alone must not exceed global threshold (in strict mode).
+    fn refill(&self, locked: &mut PriorityBasedIORateLimiterProtected, now: Instant) {
+        let mut total_budgets =
+            self.bytes_per_epoch[IOPriority::High as usize].load(Ordering::Relaxed);
+        if total_budgets == 0 {
+            // It's possible that rate limit is toggled off in the meantime.
+            return;
+        }
+        debug_assert!(now >= locked.next_refill_time);
+        let skipped_epochs_mul_16 = ((now - locked.next_refill_time).as_secs_f32() * 16.0
```
Connor1996, line 340: better use float directly, it harms the readability.

components/file_system/src/rate_limiter.rs
tabokie: We have retry logic in `File` and `rocksdb::Env` for this case.

src/storage/config.rs
```diff
@@ -225,3 +242,107 @@ impl BlockCacheConfig {
         None
     }
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Configuration)]
+#[serde(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct IORateLimitConfig {
+    pub max_bytes_per_sec: ReadableSize,
```
yiwu-arbug, line 250: I forget if I commented on this, but can we make it a `Option<ReadableSize>`, to avoid using "0MB/s" to denote no rate limiter?

tabokie: Do you mean an additional representation for no rate limit? Or we should also raise an error for using 0MB?

tabokie:
Pin @Connor1996

components/server/src/server.rs
```diff
@@ -1272,3 +1303,73 @@ impl<R: RaftEngine> EngineMetricsManager<R> {
...
+    pub fn update(&self, _duration: Duration) {
+        let mut normalized_pending_bytes = 0;
+
+        fn fetch_engine_cf(engine: &RocksEngine, cf: &str, normalized_pending_bytes: &mut u32) {
+            if let Ok(cf_opts) = engine.get_options_cf(cf) {
+                if let Ok(Some(b)) = engine.get_cf_compaction_pending_bytes(cf) {
+                    if cf_opts.get_soft_pending_compaction_bytes_limit() > 0 {
+                        *normalized_pending_bytes = std::cmp::max(
+                            *normalized_pending_bytes,
+                            (b * 100 / cf_opts.get_soft_pending_compaction_bytes_limit()) as u32,
```
Connor1996, line 1337: please use const variable instead of magic number

components/file_system/src/rate_limiter.rs
```diff
@@ -1,74 +1,431 @@
...
+                // `(a-1)/b` is equivalent to `roundup(a.saturating_sub(b)/b)`.
+                locked.next_refill_time - now
+                    + DEFAULT_REFILL_PERIOD
+                        * ((locked.pending_bytes[priority_idx] - 1) / cached_bytes_per_epoch) as u32
+            }
+        };
+        if wait > MAX_WAIT_DURATION_PER_REQUEST {
+            // Long wait duration could freeze request thread not to react to latest budgets
+            // adjustment. Exit early by returning partial quotas.
+            amount = std::cmp::max(
```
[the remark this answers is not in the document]
tabokie: Good catch. Here the large wait is contributed by accumulated `pending_bytes`, `bytes_through` will always be saturated, so I'll subtract from `pending_bytes` instead.

Connor1996 (commented):
rest LGTM

Connor1996 (approved):
LGTM

PR #7794, opened 2020-05-11
Ingest SST instead of writing memtable directly when delete a large range of keys.

Little-Wallace:
### What problem does this PR solve?

Issue Number: close [#7296](https://github.com/tikv/tikv/issues/7296)

Problem Summary:
TiKV deleted range by key and this would negatively affected write performance in apply-thread. Delete keys of each region mixed in one memtable also cause write-amplification. This PR is to solve this problem. It will insert the keys in a single SST and ingest it into RocksDB. If this file is not overlap with memtable and files in L0, it will be ingested into L1 or higher level. So it will avoid compacting with files in L0 and reduce write-amplification.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- No release note

PR #9200, opened 2020-12-08
raftstore: Fix delete by ingest cause too many L0 files

Little-Wallace:

Issue Number: close https://github.com/tikv/tikv/issues/10680

Problem Summary:

https://github.com/tikv/tikv/pull/7794  used `ingest` for destroy region instead of deleting them key by key. But there is a problem that when RocksDB has been `WriteStall` status, the region-worker will still ingest file into RocksDB, because I did not check `ingest_maybe_stall` when calling `clean_stale_ranges` in timeout tick. So the number of files in L0 of RocksDB may be larger and larger, and TiKV can not apply any snapshot because applying snapshot acquires to pass the check of `ingest_maybe_stall`.

Check `ingest_maybe_stall` every time before calling `ingest_maybe_stall`.

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```
None
```


components/raftstore/src/store/worker/region.rs
```diff
@@ -721,7 +734,9 @@ where
         self.handle_pending_applies();
         self.clean_stale_tick += 1;
         if self.clean_stale_tick >= STALE_PEER_CHECK_TICK {
-            self.ctx.clean_stale_ranges();
+            if !self.ctx.ingest_maybe_stall() {
+                self.ctx.clean_stale_ranges();
```
yiwu-arbug, line 738: just trying to understand. Is it true that in case of write stall, the worker thread will wait for the next tick to reschedule the stale range cleanup?

yiwu-arbug: I'm wondering whether we should notify PD to stop scheduling more region migration, if there is a backlog of clean_stale_range backlog.

components/raftstore/src/store/worker/region.rs
```diff
@@ -457,6 +460,9 @@ where
         if overlap_ranges.is_empty() {
             return Ok(());
         }
+        SNAP_COUNTER_VEC
```
BusyJay, line 463: I don't get it, why add the metrics to `SNAP_COUNTER_VEC`?

Little-Wallace: There are many kind of counter in `SNAP_COUNTER_VEC ` ....

Little-Wallace: In actually, almost all counter metrics in region.rs are include in `SNAP_COUNTER_VEC `.

components/raftstore/src/store/worker/region.rs
Little-Wallace: There are some limit in PD. If we stop consume the request to region-worker, PD will stop scheduling more region migration. Please check that whether there is other negative impact @BusyJay

components/raftstore/src/store/worker/metrics.rs
```diff
@@ -99,6 +99,12 @@ lazy_static! {
         "Total number of tikv pending delete range of stale peer"
     )
     .unwrap();
+    pub static ref CLEAN_COUNTER_VEC: IntCounterVec = register_int_counter_vec!(
+        "tikv_raftstore_clean_region_count",
+        "Total number of raftstore snapshot processed.",
```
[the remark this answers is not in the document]
Little-Wallace: Done

components/raftstore/src/store/worker/region.rs
BusyJay: I don't think so. `SNAP_COUNTER_VEC` is used for counting the generated and applied snapshots. overlap is not a valid snapshot type. Neither `delete_by_ingest` is a valid snapshot status.

metrics/grafana/tikv_details.json
```diff
@@ -16810,6 +16817,109 @@
...
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "stack": false,
+          "steppedLine": true,
+          "targets": [
+            {
+              "expr": "sum(tikv_pending_delete_ranges_of_stale_peer{tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\"}) by (instance)",
```
BusyJay, line 16872: Can you post an image in description to show how it looks?

metrics/grafana/tikv_details.json
```diff
@@ -16856,11 +16966,18 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum(delta(tikv_raftstore_snapshot_total{tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\"}[1m])) by (type, status)",
+              "expr": "sum(rate(tikv_raftstore_snapshot_total{tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\"}[1m])) by (type, status)",
```
BusyJay, line 16969: Why change this? Actually I think `delta` is more straightforward here. For example, 2 snapshot in 60 seconds can be 0.03 with rate, but can show 2 if using delta.

Little-Wallace: OK.... I used to think it is not accurately.

metrics/grafana/tikv_details.json
Little-Wallace: image
image

components/raftstore/src/store/worker/region.rs
```diff
@@ -457,6 +460,7 @@ where
         if overlap_ranges.is_empty() {
             return Ok(());
         }
+        CLEAN_COUNTER_VEC.with_label_values(&["overlap"]).inc();
```
BusyJay, line 463: This can be triggered either by applying snapshot or inserting overlap ranges.

components/raftstore/src/store/worker/region.rs
```diff
@@ -522,6 +526,10 @@ where
             .stale_ranges(oldest_sequence)
             .map(|(region_id, s, e)| (region_id, s.to_vec(), e.to_vec()))
             .collect();
+        if cleanup_ranges.is_empty() {
+            return;
+        }
+        CLEAN_COUNTER_VEC.with_label_values(&["tick"]).inc_by(1);
```
BusyJay, line 532: This can either be triggered by `tick` or destroy task.

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/worker/region.rs
```diff
@@ -489,21 +497,28 @@ where
...
+                .inc();
+            if let Err(e) = self.cleanup_overlap_regions(overlap_ranges) {
+                warn!("cleanup_overlap_ranges failed";
+                    "region_id" => region_id,
+                    "start_key" => log_wrappers::Value::key(start_key),
+                    "end_key" => log_wrappers::Value::key(end_key),
+                    "err" => %e,
+                );
+            } else {
+                info!("register deleting data in range";
```
BusyJay, line 515: This should be printed when `overlap_ranges` is empty.

BusyJay (approved):

Connor1996 (approved):
LGTM

zhouqiang-cl:
@ti-chi-bot

PR #9207, opened 2020-12-09
txn_types: Add GC fence feild to Write records

MyonKeminta:

Issue Number: 

Adds a `gc_fence` field to `Write` records which will be used to resolve the compatibility issue between async commit and gc compaction filter/cdc/tiflash.

Added a `gc_fence` field.

### Related changes

-

### Check List 

Tests 

- Unit test

Side effects

### Release note 

* No release note.


hicqu (approved):

components/txn_types/src/write.rs
```diff
@@ -68,6 +70,47 @@ pub struct Write {
     /// Also note that `has_overlapped_rollback` field is only necessary when the Rollback record
     /// should be protected.
     pub has_overlapped_rollback: bool,
```
youjiali1995, line 72: Seems it can be replaced by `gc_fence`?

sticnarf (commented):
If it's produced because commit_ts == rollback_ts in the lock, it doesn't need a gc fence.

components/txn_types/src/write.rs
sticnarf: If it's produced because commit_ts == rollback_ts in the lock, it doesn't need a gc fence.

components/txn_types/src/write.rs
```diff
@@ -68,6 +70,47 @@ pub struct Write {
...
+    /// The meaning of the field:
+    /// * `None`: A record that haven't been rewritten
+    /// * `Some(0)`: A commit record that has been rewritten due to overlapping rollback, but it
+    ///   doesn't have an newer version.
+    /// * `Some(ts)`: A commit record that has been rewritten due to overlapping rollback,
+    ///   and it's next version's `commit_ts` is `ts`
+    pub gc_fence: Option<TimeStamp>,
 }
 
 impl std::fmt::Debug for Write {
```
sticnarf, line 116: `Debug` needs updating

components/txn_types/src/write.rs
```diff
@@ -217,6 +277,10 @@ impl WriteRef<'_> {
         if self.has_overlapped_rollback {
             b.push(FLAG_OVERLAPPED_ROLLBACK);
         }
+        if let Some(ts) = self.gc_fence {
```
sticnarf, line 280: Maybe we can calculate a better capacity estimation like https://github.com/tikv/tikv/blob/6b4466ba8a/components/txn_types/src/lock.rs#L161

components/txn_types/src/write.rs
```diff
@@ -68,6 +70,47 @@ pub struct Write {
...
+    ///
+    /// For CDC and TiFlash, when they receives a commit record with `gc_fence` field set, it can
+    /// determine that it must be caused by an overlapped rollback instead of an actual commit.
+    ///
+    /// The meaning of the field:
+    /// * `None`: A record that haven't been rewritten
+    /// * `Some(0)`: A commit record that has been rewritten due to overlapping rollback, but it
+    ///   doesn't have an newer version.
+    /// * `Some(ts)`: A commit record that has been rewritten due to overlapping rollback,
+    ///   and it's next version's `commit_ts` is `ts`
```
nrc, line 112: This is a great comment! 👍👍👍

components/txn_types/src/write.rs
```diff
@@ -68,6 +70,47 @@ pub struct Write {
...
+    /// 1. `Key_100_put`, `Key_120_del` applied
+    /// 2. GC with safepoint = 130 started and `Key_100_put`, `Key_120_del` are deleted
+    /// 3. Finished applying `Key_100_put_R`, which means to rewrite `Key_100_put`
+    /// 4. Read at `140` should get nothing (since it's MVCC-deleted at 120) but finds `Key_100_put`
+    ///
+    /// To solve the problem, when marking `has_overlapped_rollback` on an already-existed commit
+    /// record, add a special field `gc_fence` on it. If there is a newer version after the record
+    /// being rewritten, the next version's `commit_ts` will be recorded. When MVCC reading finds
+    /// a commit record with a GC fence timestamp but the corresponding version that matches that ts
+    /// doesn't exest, the current version will be believed to be already GC-ed and ignored.
```
nrc, line 102: Could you say when and what the gc fence would be set to in the example?

nrc (approved):
lgtm, one query about the comment

sticnarf (approved):
The rest lgtm.

ti-srebot (approved):

components/txn_types/src/write.rs
```diff
@@ -217,9 +291,26 @@ impl WriteRef<'_> {
...
+        if let Some(ts) = self.gc_fence {
+            b.push(GC_FENCE_PREFIX);
+            b.encode_u64(ts.into_inner()).unwrap();
+        }
         b
     }
 
+    fn pre_allocate_size(&self) -> usize {
+        // write_type + start_ts + has_overlapped_rollback
+        let mut size = 1 + MAX_VAR_U64_LEN + 1;
```
sticnarf, line 303:
```suggestion
        let mut size = 1 + MAX_VAR_U64_LEN + self.has_overlapped_rollback as usize;
```

MyonKeminta: Good idea

components/txn_types/src/write.rs
```diff
@@ -217,9 +291,25 @@ impl WriteRef<'_> {
         if self.has_overlapped_rollback {
             b.push(FLAG_OVERLAPPED_ROLLBACK);
         }
+        if let Some(ts) = self.gc_fence {
+            b.push(GC_FENCE_PREFIX);
+            b.encode_u64(ts.into_inner()).unwrap();
```
youjiali1995, line 296: Why not `encode_var_u64`?

sticnarf: VarInt saves space when ts is 0. But VarInt occupies more space (9 bytes vs 8 bytes) and is slower when ts is a real timestamp.
It has little difference to me because overlapping happens rarely...

MyonKeminta: I thought it doesn't matter much which encoding to use here, but encode_u64 looks faster than var_u64

youjiali1995: But it's strange to encode start_ts in var int and this ts in u64.

MyonKeminta: You can see in Lock's encoding, everything new is using encode_u64...

PR #9215, opened 2020-12-10
raftstore: negotiate before hibernating

### LINKED ISSUE #7747: Fail to evict leader when rolling update a new hibernated cluster
## Bug Report

If there are too many leaders or transfer leader just goes too slow, rolling update a cluster can take too much time. If the cluster enables hibernate region at the time, it's possible that all updated nodes will not have any leader because updated leader becomes hibernated very quickly and outdated follower will start election when lost heartbeat from leader for a long time. So it becomes slower because more leaders need to be evicted during rolling update.

A possible solution is to add a hibernate timeout configuration. Leader will not go to sleep before the timeout. So it's safe as long as the hibernate timeout is larger than the update time. And the timeout can be set to a smaller value after the cluster has enabled hibernate region on all nodes.

Another solution is make the configuration be changed online. So users can enable/disable hibernate regions whenever they feel safe.

We need to fix the issue in 3.0.x, 3.1.x and 4.0.

BusyJay:

Close #7747.

This PR implements the mechenism mentioned in https://github.com/tikv/tikv/issues/7747#issuecomment-737117530.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- Negotiate before hibernating to support rolling update


tests/integrations/raftstore/test_hibernate.rs
```diff
@@ -275,3 +276,137 @@ fn test_split_delay() {
...
+    let a = awakened.clone();
+    cluster.add_send_filter(CloneFilterFactory(
+        RegionPacketFilter::new(1, 1)
+            .direction(Direction::Send)
+            .set_msg_callback(Arc::new(move |_| {
+                a.store(true, Ordering::SeqCst);
+            }))
+            .when(filter.clone()),
+    ));
+    thread::sleep(cluster.cfg.raft_store.raft_heartbeat_interval() * 2);
```
NingLin-P, line 312: Should we sleep more than 2 to make the test more stable?

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -3231,11 +3280,9 @@ where
             "peer_id" => self.peer.get_id(),
             "msg_type" => ?msg_type,
```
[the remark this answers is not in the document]
BusyJay: They will be sent using `send_extra_message` method instead.

tests/integrations/raftstore/test_hibernate.rs
BusyJay: We only need to sleep a heartbeat timeout to receive a heartbeat, using `* 2` is already considering the occasional slowness in tests.

NingLin-P (approved):

components/raftstore/src/store/hibernate_state.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    pub fn should_bcast(&self, gate: &FeatureGate) -> bool {
+        gate.can_enable(NEGOTIATION_HIBERNATE)
+    }
+
+    pub fn maybe_hibernate(&mut self, my_id: u64, region: &Region) -> bool {
+        let peers = region.get_peers();
+        let v = match &mut self.leader {
```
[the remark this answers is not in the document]
BusyJay: Without `&mut`, pattern matching on a field will move it.

components/raftstore/src/store/peer.rs
```diff
@@ -3231,11 +3280,9 @@ where
...
             "msg_size" => msg.compute_size(),
-            "from" => from_peer.get_id(),
```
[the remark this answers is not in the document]
BusyJay: `from_peer` is `self.peer.clone()`, which can be known by `peer_id` in the same log.

components/raftstore/src/store/hibernate_state.rs
```diff
@@ -0,0 +1,93 @@
...
+        self.group = group_state;
+        if group_state != GroupState::Idle {
+            self.leader = LeaderState::Awaken;
+        }
+    }
+
+    pub fn count_vote(&mut self, from: u64) {
+        if let LeaderState::Poll(v) = &mut self.leader {
+            if !v.contains(&from) {
+                v.push(from);
```
gengliqi, line 60: How about checking the vote state here?

gengliqi (commented):
Rest LGTM

gengliqi (approved):

ti-srebot (approved):

components/raftstore/src/store/hibernate_state.rs
BusyJay: Currently, a leader will not go to sleep if internal ticks don't reach 0. Checking vote count early can make the situation complicated, I'm prefer to keep checking it after a round of ticking.

Issue #9210, opened 2020-12-10
build: build failed on macOS


longfangsong:
## Development Task

Fix build failed under macOS, caused by #9206.

```
error[E0432]: unresolved import `crate::collections`
  --> components/tikv_util/src/metrics/threads_dummy.rs:12:12
   |
12 | use crate::collections::HashMap;
   |            ^^^^^^^^^^^
   |            |
   |            unresolved import
   |            help: a similar path exists: `std::collections`

error[E0432]: unresolved import `crate::collections`
  --> components/tikv_util/src/metrics/threads_dummy.rs:12:12
   |
12 | use crate::collections::HashMap;
   |            ^^^^^^^^^^^
   |            |
   |            unresolved import
   |            help: a similar path exists: `std::collections`

error: aborting due to previous error
```

Issue #8570, opened 2020-09-02
Cannot build under macOS


longfangsong:
### What version of TiKV are you using?
master

### What operating system and CPU are you using?
macOS 10.15.4

### Steps to reproduce

Build TiKV (eg. by using `make dev`)

### What did you expect?

Build successfully.

### What did happened?

Some error about raft-engine happens:

```shell
error[E0433]: failed to resolve: could not find `FallocateFlags` in `fcntl`
```

Seems caused by [raft-engine#22](https://github.com/tikv/raft-engine/pull/22) and introduced into TiKV by #8554.

As [code in `nix`](https://github.com/nix-rust/nix/blob/6f0ca030a30fd813ad74f7dc4d63d3d5538090b0/src/fcntl.rs#L537) imply, `fcntl::FallocateFlags` only works on 
linux.

PR #9219, opened 2020-12-11
build, ci: Check whether developers can develop TiKV under macOS

longfangsong:

Issue Number: related with #9210 and #8570

Problem Summary:

Since many our developers are using macOS, we should prevent building errors on macOS happen again by let the ci check whether master can be built under macOS.

What's Changed: Use Github Actions to check whether `master` can build under macOS

### Check List 

Tests 

- No code


longfangsong:
I'll try to use `Swatinem/rust-cache@v1` to cache the build result to make it run faster.

sticnarf:
Now the four tasks use the same cache key. I saw warnings about it. I think you need to set a different `key` for building the test.

longfangsong:
Now the GitHub Actions runs for around 15min, I think that's acceptable. Since our jenkins/tikv_ghpr_test also lasts for around 15min.
PTAL @sticnarf

sticnarf:
Looks good to me. The name of the check may be polished.

What do you think about it? cc @BusyJay @zhouqiang-cl

BusyJay:
Actually, jenkins test only last for 12 minutes, new build adds 25% of time. I think just doing link check when pull request on MacOS is enough, which is simple `make format clippy`.

.github/workflows/develop_under_macos.yml
```diff
@@ -0,0 +1,17 @@
+on: [
+    push,
+    pull_request
+]
+name: make sure can develop under macOS
```
longfangsong, line 5:
```suggestion
name: make sure can develop under macOS
concurrency:
  group: ${{ github.ref }}-${{ github.workflow }}
  cancel-in-progress: true
```

PR #9222, opened 2020-12-11
txn: append GC fence with overlapped rollback flag if necessary

hicqu:

This PR is based on #9207 to make async-commit be compatible with machine local GC.

Append overlapped rollback flag with a **GC fence** if necessary. So that if a given `key_mvcc_put` is rewritten when appending a rollback flag, MvccReader can know whether it has been cleared by GC or not according to the GC fence.

### Check List 

Tests 

- Unit test

Side effects

### Release note 
* No release note


src/storage/mvcc/txn.rs
```diff
@@ -19,26 +22,68 @@ pub struct GcInfo {
     pub is_completed: bool,
 }
 
+/// Get the GC fence for `key`. PTAL at `txn_types::Write::gc_fence`.
+pub struct GcFenceGetter<'a, S: Snapshot> {
```
sticnarf, line 26: Using this getter brings extra costs. But I think both `CheckTxnStatus` and `CheckSecondaryLocks` search the commit record, so we can get the fence TS through this search.

sticnarf (commented):
Mostly LGTM. And I recommend adding more `must_get_overlapped_rollback` assertions for all `check_txn_status`. `check_txn_status` also writes overlapped rollbacks.

sticnarf:
LGTM

src/storage/mvcc/reader/reader.rs
```diff
@@ -28,6 +30,13 @@ pub enum TxnCommitRecord {
     OverlappedRollback { commit_ts: TimeStamp },
 }
 
+#[derive(Clone, Debug)]
+pub struct OverlappedWrite {
+    pub write: Write,
+    /// GC fence for `overlapped_write`. PTAL at `txn_types::Write::gc_fence`.
+    pub gc_fence: TimeStamp,
```
MyonKeminta, line 37: It looks redundant to add the `gc_fence` field here, since `Write` already has the field.

sticnarf: It's something different. It's actually the "potential" gc_fence to be write.

hicqu: In fact we can merge them into one. I have updated code for this.

src/storage/mvcc/reader/reader.rs
```diff
@@ -269,9 +270,16 @@ impl<S: Snapshot> MvccReader<S> {
                     return Ok(TxnCommitRecord::OverlappedRollback { commit_ts });
                 }
                 return Ok(TxnCommitRecord::None {
-                    overlapped_write: Some(write),
+                    overlapped_write: Some(
+                        write
+                            .set_overlapped_rollback(true)
+                            .set_gc_fence(Some(gc_fence)),
+                    ),
```
MyonKeminta, line 277: Why immediately set the overlapped rollback and gc fence here? The `get_txn_commit_record` is used to load the transaction's status from disk, it doesn't assume what the result will be used to do. Also the record might already has overlapped rollback or gc fence.

For example, this function is also used when committing a transaction but lock is not found, to check what's going on about the transaction. No rollback will happen here.

sticnarf: I prefer the implementation one commit ago: store potential gc fence separately in `OverlappedWrite`.

MyonKeminta (approved):
LGTM.

ti-srebot (approved):

PR #9237, opened 2020-12-12
raftstore: stop background worker

Little-Wallace:

Issue Number: close https://github.com/tikv/tikv/issues/9151 

Problem Summary:
TiKV may crash if background worker stop before raftstore pool shutdown.

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

- No release note.

### DIFF
--- a/components/batch-system/src/router.rs
+++ b/components/batch-system/src/router.rs
@@ -220,7 +220,13 @@ where
                     .unwrap()
                     .force_send(m, &self.normal_scheduler)
             }
-            Err(TrySendError::Disconnected(m)) => Err(SendError(m)),
+            Err(TrySendError::Disconnected(m)) => {
+                if self.is_shutdown() {
+                    Ok(())
+                } else {
+                    Err(SendError(m))
+                }
+            }
         }
     }
 
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -2696,17 +2696,21 @@ where
             );
             self.fsm.peer.heartbeat_pd(self.ctx);
         }
-        self.ctx
-            .router
-            .force_send(
+        if let Err(e) = self.ctx.router.force_send(
+            source.get_id(),
+            PeerMsg::SignificantMsg(SignificantMsg::MergeResult {
+                target_region_id: self.fsm.region_id(),
+                target: self.fsm.peer.peer.clone(),
+                result: MergeResultKind::FromTargetLog,
+            }),
+        ) {
+            panic!(
+                "{} failed to send merge result(FromTargetLog) to source region {}, err {}",
+                self.fsm.peer.tag,
                 source.get_id(),
-                PeerMsg::SignificantMsg(SignificantMsg::MergeResult {
-                    target_region_id: self.fsm.region_id(),
-                    target: self.fsm.peer.peer.clone(),
-                    result: MergeResultKind::FromTargetLog,
-                }),
-            )
-            .unwrap();
+                e
+            );
+        }
     }
 
     /// Handle rollbacking Merge result.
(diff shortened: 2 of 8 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2943,18 +2937,17 @@ where
...
-            if let Err(e) = self.ctx.router.force_send(
-                r.get_id(),
-                PeerMsg::SignificantMsg(SignificantMsg::MergeResult {
-                    target_region_id: self.fsm.region_id(),
-                    target: self.fsm.peer.peer.clone(),
-                    result: MergeResultKind::FromTargetSnapshotStep2,
-                }),
-            ) {
-                if !self.ctx.router.is_shutdown() {
-                    panic!("{} failed to send merge result(FromTargetSnapshotStep2) to source region {}, err {}", self.fsm.peer.tag, r.get_id(), e);
```
gengliqi, line 2955: The log should remain.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2696,23 +2696,17 @@ where
...
-                result: MergeResultKind::FromTargetLog,
-            }),
-        ) {
-            if !self.ctx.router.is_shutdown() {
-                panic!(
-                    "{} failed to send merge result(FromTargetLog) to source region {}, err {}",
-                    self.fsm.peer.tag,
-                    source.get_id(),
-                    e
-                );
```
gengliqi, line 2713: The log should remain.

gengliqi (approved):
LGTM

MyonKeminta (approved):

ti-srebot (approved):

components/batch-system/src/router.rs
```diff
@@ -220,7 +220,13 @@ where
                     .unwrap()
                     .force_send(m, &self.normal_scheduler)
             }
-            Err(TrySendError::Disconnected(m)) => Err(SendError(m)),
+            Err(TrySendError::Disconnected(m)) => {
+                if self.is_shutdown() {
```
BusyJay, line 224: I don't think it's a good idea. Only caller knows what's the most appropriate approach to handle shutdown.

gengliqi: Maybe it's general enough. I can't imagine a case that the caller needs to know this send error during the shutdown.

BusyJay: It's the problem of API design choice. The change assumes callers are always OK to drop messages without reasoning the error type, which may not be the case for future. For example, if callers need to clean up some resources for error, while holding resource when succeed.

If it's a generic case, then it should be implemented as a different API and give it a clear name.

BusyJay: @Little-Wallace I don't see it's commented or adjusted.  :)

Little-Wallace: If this router has shutdown, it means that all of the message shall be ignore, because there is no thread to deal with these messages.

Little-Wallace: Sorry for I missing your comment.... I think we can assume that when the router is `shut_down`, all the message has been send successfully but there is no thread to deal with them. The following two logics is equal:
```
thread1: send(msg).is_ok();
thread2: recv(msg).is_ok(); if is_shutdown() {  drop(msg);  exit(); }
```

```
thread1: send(msg).is_err();
thread2: if is_shutdown() { exit(); }
```

BusyJay: It means all of the messages are not being handled by the threads, doesn't necessary mean to be ignored or reported as error.

Little-Wallace: If shutdown occurs,  TiKV do not need to care about whether the messages have send successfully.

BusyJay: > The following two logics is equal:

Actually, TiKV doesn't just drop all messages unconditionally. It will invoke the callbacks if there are any.

> If shutdown occurs, TiKV do not need to care about whether the messages have send successfully.

Like I said above, "It's the problem of API design choice. The change assumes callers are always OK to drop messages without reasoning the error type, which may not be the case for future. For example, if callers need to clean up some resources for error, while holding resource when succeed.". Or maybe someone just tries to use `force_send` to send a command.

Issue #9016, opened 2020-11-11
Coprocessor functions migration from non-vec framework


andylokandy:
## Feature Request

Coprocessor historically has two implementations. One is based on the vectorized framework, and the other one is based on the plain (non-vec) framework, which was removed recently. A few of the functions in the legacy non-vec framework have not been ported into the vectorized framework yet.

The following is the list of functions that existed in non-vec framework but not in vectorized framework. You may also want to look into the [non-vec framework before it was removed](https://github.com/tikv/tikv/tree/1a88f12ebf50064b992fc5efc7e7f55795210521/components/tidb_query_normal_expr/src).

- [x] AddDateAndDuration
- [ ] AddDateAndString
- [x] AddDatetimeAndDuration
- [x] AddDatetimeAndString
- [x] AddDurationAndDuration
- [x] AddDurationAndString
- [ ] AddTimeDateTimeNull
- [ ] AddTimeDurationNull
- [ ] AddTimeStringNull
- [ ] Compress
- [x] Date
- [ ] DateDiff
- [ ] Instr
- [x] Locate2ArgsUtf8
- [x] Locate3ArgsUtf8
- [ ] Lower
- [x] NullTimeDiff
- [x] Quote
- [x] RegexpSig
- [x] RegexpUtf8Sig
- [ ] RpadUtf8
- [x] SubDateAndDuration
- [ ] SubDateAndString
- [x] SubDatetimeAndDuration
- [ ] SubDatetimeAndString
- [x] SubDurationAndDuration
- [x] SubDurationAndString
- [ ] SubTimeDateTimeNull
- [ ] SubTimeDurationNull
- [ ] Substring2Args
- [ ] Substring2ArgsUtf8
- [ ] Substring3Args
- [ ] Substring3ArgsUtf8
- [ ] Trim2Args
- [ ] TruncateDecimal
- [ ] TruncateUint
- [ ] Uncompress
- [x] WeekWithoutMode
- [x] YearWeekWithMode
- [x] YearWeekWithoutMode

wangggong:
AddDatetimeAndDuration & SubDatetimeAndDuration done

andylokandy:
> AddDatetimeAndDuration & SubDatetimeAndDuration done

Do know which PR added these two?

wangggong:
> > AddDatetimeAndDuration & SubDatetimeAndDuration done
> 
> Do know which PR added these two?

@andylokandy This one: https://github.com/tikv/tikv/pull/8999

[...]

Issue #8986, opened 2020-11-06
copr: UTF8 functions should consider collation


andylokandy:
Some [UTF8 functions](https://github.com/tikv/tikv/blob/master/components/tidb_query_vec_expr/src/impl_string.rs) in TiKV coprocessor do not properly take collation into consideration. For example, [`instr_utf8`](https://github.com/tikv/tikv/blob/44f189c39460ccec75c3f0a5d0210c82eeddc6fa/components/tidb_query_vec_expr/src/impl_string.rs#L561) now compares two strings by their lowercase representation, which is not the right definition.

Related functions:
- [ ] ord (may need to split into ord_utf8)
- [ ] ~~instr_utf8~~
- [x] find_in_set
- [x] strcmp
- [ ]  ~~Locate2ArgsUTF8~~
- [ ] ~~Locate3ArgsUTF8~~

andylokandy:
Useful reference: https://github.com/tikv/tikv/blob/308c641fce76c65313f9ffa491f0b4a5be8b1f70/components/tidb_query_vec_expr/src/impl_like.rs#L12

TszKitLo40:
Can I try `strcmp` and `strcmp_utf8`?

andylokandy:
@TszKitLo40 

> Can I try strcmp and strcmp_utf8? Is strcmp like the function in C++?

Yes, sure! You need to use a [collator](https://tikv.github.io/doc/tidb_query_datatype/codec/collation/trait.Collator.html) to compare strings [here](https://github.com/tikv/tikv/blob/515df8d552cce67111991fc6b205ec2905716c2b/components/tidb_query_expr/src/impl_string.rs#L684). In order to retrieve a collator, you need to expose the collator as a generic parameter like [this](https://github.com/tikv/tikv/blob/515df8d552cce67111991fc6b205ec2905716c2b/components/tidb_query_expr/src/impl_string.rs#L101). At last, the [function entrance](https://github.com/tikv/tikv/blob/515df8d552cce67111991fc6b205ec2905716c2b/components/tidb_query_expr/src/lib.rs#L589) needs to be replaced by a collation mapper like [this](https://github.com/tikv/tikv/blob/515df8d552cce67111991fc6b205ec2905716c2b/components/tidb_query_expr/src/lib.rs#L93) .

TszKitLo40:
> @TszKitLo40
> 
> > Can I try strcmp and strcmp_utf8? Is strcmp like the function in C++?
> 
[...]

PR #9238, opened 2020-12-12
copr: Implement regexp & regexp_utf8

wangggong:
Signed-off-by: wangggong <793160615@qq.com>

Issue Number: #9016 

Problem Summary:

What's Changed:

Add regexp & regexp_utf8

### Check List 

Tests 

- Unit test

Side effects

### Release note 

No release note.


wangggong:
@andylokandy PTAL

Maybe these two methods need to use Collation like https://github.com/tikv/tikv/issues/8986 ?

Xuanwo:
Please considering use the title `corp: Implement regexp & regexp_utf8`?

Xuanwo:
@wangggong Yes, I think we should use `Collation` in `regexp_utf8`

andylokandy (commented):
> @wangggong Yes, I think we should use `Collation` in `regexp_utf8`

Seems that tidb did not make `regexp_utf8` collation-aware.

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,6 +68,41 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+        Err(err) => return Err(box_err!("invalid input value: {:?}", err)),
+    };
+    let pattern = match String::from_utf8(pattern.to_vec()) {
+        Ok(pattern) => pattern,
+        Err(err) => return Err(box_err!("invalid input value: {:?}", err)),
+    };
+    let pattern = format!("(?i){}", pattern);
+
+    // TODO: cache compiled result
+    match Regex::new(&pattern) {
```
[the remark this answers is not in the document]
andylokandy: PS: we can make the metadata to be of type`Option<Regex>`; to be `None` when the pattern is ColumnRef.

andylokandy:
> Yes, I think we should use Collation in regexp_utf8

But the TiDB did not make regex_utf8 collation-aware...

Xuanwo:
> But the TiDB did not make regex_utf8 collation-aware...

So we should implement `regex_utf8` without collation until TiDB supported?

wangggong:
> > But the TiDB did not make regex_utf8 collation-aware...
> 
> So we should implement `regex_utf8` without collation until TiDB supported?

@Xuanwo @andylokandy Or I can support it in TiDB myself before this PR; if it's not in a hurry, maybe I can handle it this weekend?

andylokandy:
> > > But the TiDB did not make regex_utf8 collation-aware...
> > 
> > 
> > So we should implement `regex_utf8` without collation until TiDB supported?
> 
> @Xuanwo @andylokandy Or I can support it in TiDB myself before this PR; if it's not in a hurry, maybe I can handle it this weekend?

Since TiDB and TiKV's regex depends on 3rd-party-library, it becomes quite difficult to make it collation-aware. That's why TiDB ignores the collation. Do you have any suggestion?

wangggong:
> > > > But the TiDB did not make regex_utf8 collation-aware...
> > > 
> > > 
> > > So we should implement `regex_utf8` without collation until TiDB supported?
> > 
> > 
> > @Xuanwo @andylokandy Or I can support it in TiDB myself before this PR; if it's not in a hurry, maybe I can handle it this weekend?
> 
> Since TiDB and TiKV's regex depends on 3rd-party-library, it becomes quite difficult to make it collation-aware. That's why TiDB ignores the collation. Do you have any idea to implement it?

Maybe I'll read how MySQL implements it? Or just implement `regex_utf8` without collation like now?

andylokandy:
> > > > > But the TiDB did not make regex_utf8 collation-aware...
> > > > 
> > > > 
> > > > So we should implement `regex_utf8` without collation until TiDB supported?
> > > 
> > > 
> > > @Xuanwo @andylokandy Or I can support it in TiDB myself before this PR; if it's not in a hurry, maybe I can handle it this weekend?
> > 
> > 
> > Since TiDB and TiKV's regex depends on 3rd-party-library, it becomes quite difficult to make it collation-aware. That's why TiDB ignores the collation. Do you have any idea to implement it?
> 
> Maybe I'll read how MySQL implements it? Or just implement `regex_utf8` without collation like now?

I think you can implement it without collation first and to investigate the collation solution later on.

components/tidb_query_expr/src/impl_like.rs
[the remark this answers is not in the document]
wangggong: done

PTAL

andylokandy:
@Xuanwo @zhongzc PTAL

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,6 +70,87 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+    match metadata {
+        Some(regex) => Ok(Some(regex.is_match(target.as_ref()) as i64)),
+        None => {
+            let pattern = match String::from_utf8(pattern.to_vec()) {
+                Ok(pattern) => pattern,
+                Err(err) => return Err(box_err!("invalid input value: {:?}", err)),
+            };
+            let pattern = format!("(?i){}", pattern);
+
+            match BytesRegex::new(&pattern) {
```
Xuanwo, line 94: Why we need to use `BytesRegex` for utf-8 here? Both `target` and `pattern` should be valid utf-8 value here.

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -68,38 +70,84 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+                Err(err) => Err(box_err!("invalid regex pattern: {:?}", err)),
+            }
+        }
+    }
+}
+
+fn init_regexp_utf8_data(expr: &mut Expr) -> Result<Option<Regex>> {
+    let children = expr.mut_children();
+    let n = children.len();
+    assert!(n > 1);
```
Xuanwo, line 127: It's OK to panic here?

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -68,38 +70,84 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+
+    let tree_node = &mut children[1];
+    match tree_node.get_tp() {
+        ExprType::ScalarFunc | ExprType::ColumnRef | ExprType::Null => Ok(None),
+        _ => match Regex::new(String::from_utf8(tree_node.take_val()).unwrap().as_ref()) {
```
Xuanwo, line 132: We don't need to care about other `ExprType`? What if we got a `ExprType::DateTime` instead of `ExprType:Bytes` here?

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,6 +70,87 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+
+fn init_regexp_data(expr: &mut Expr) -> Result<Option<BytesRegex>> {
+    let children = expr.mut_children();
+    let n = children.len();
+    assert!(n > 1);
+
+    let tree_node = &mut children[1];
+    match tree_node.get_tp() {
+        ExprType::ScalarFunc | ExprType::ColumnRef | ExprType::Null => Ok(None),
+        _ => match BytesRegex::new(String::from_utf8(tree_node.take_val()).unwrap().as_ref()) {
```
andylokandy, line 147: Would be better to use `?` instead of unwrap

components/tidb_query_expr/src/impl_like.rs
wangggong: Hmm, maybe just returning `Ok(None)` for unexpected input would be better. I'll fix it.

components/tidb_query_expr/src/impl_like.rs
wangggong: In order to prevent unexpected `ExprType`, I just consider `Bytes` for `regexp` and `String` for `regexp_utf8` temporary.

components/tidb_query_expr/src/impl_like.rs
wangggong: It is tried, while ` the trait ``std::convert::From<std::string::FromUtf8Error>`` is not implemented for ``tidb_query_common::error::EvaluateError`` ` will be reported.

Maybe we can implement it?

components/tidb_query_expr/src/impl_like.rs
Xuanwo: Maybe we can handle `Bytes` and `String` here, and return `Ok(None)` for other `ExprType`?

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,6 +70,87 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+            Ok(regex) => Ok(Some(regex)),
+            Err(_) => Ok(None),
+        },
+    }
+}
```
andylokandy, line 152:
```suggestion
    match tree_node.get_tp() {
         ExprType::Bytes => {
              let pattern = match String::from_utf8(tree_node.take_val()) {
                   Ok(pattern) => pattern,
                   Err(err) => return Err(box_err!("invalid input value: {:?}", err)),
              };
              match BytesRegex::new(&pattern) {
                Ok(bytes_regex) => Ok(Some(bytes_regex.is_match(target.as_ref()) as i64)),
                Err(err) => Err(box_err!("invalid regex pattern: {:?}", err)),
            }
         }
         _ => Ok(None)
    }
```

components/tidb_query_expr/src/impl_like.rs
wangggong: done

PTAL

components/tidb_query_expr/src/impl_like.rs
wangggong: done

components/tidb_query_expr/src/impl_like.rs
wangggong: done

wangggong: BTW, in my comprehension `RpnFnScalarEvaluator::evaluate_raw` just sets field type and tp of each param, and does not set val of param; which will make test fail (`init_regexp_data` / `init_regexp_utf8_data` just read val part of the param as the `pattern`, and this will get a null value). Is it reasonable? Or the code is misunderstood?

https://github.com/tikv/tikv/blob/master/components/tidb_query_expr/src/types/test_util.rs#L91

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,13 +70,105 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+fn init_regexp_utf8_data(expr: &mut Expr) -> Result<Option<Regex>> {
+    let children = expr.mut_children();
+    if children.len() <= 1 {
+        return Ok(None);
+    }
+
+    let tree_node = &mut children[1];
+    match tree_node.get_tp() {
+        ExprType::Bytes | ExprType::String => {
+            match Regex::new(String::from_utf8(tree_node.take_val()).unwrap().as_ref()) {
```
andylokandy, line 133: Please don't `unwrap`

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,13 +70,105 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+fn init_regexp_data(expr: &mut Expr) -> Result<Option<BytesRegex>> {
+    let children = expr.mut_children();
+    if children.len() <= 1 {
+        return Ok(None);
+    }
+
+    let tree_node = &mut children[1];
+    match tree_node.get_tp() {
+        ExprType::Bytes | ExprType::String => {
+            match BytesRegex::new(String::from_utf8(tree_node.take_val()).unwrap().as_ref()) {
```
andylokandy, line 151: Please don't `unwrap`

andylokandy:
rest LGTM

components/tidb_query_expr/src/impl_like.rs
andylokandy: But seems your tests are passing. What's wrong with the `evaluate_raw`?

wangggong: `evaluate_raw` does not set val of param, which makes the Regex will have a null value to `new`; while I'm not sure whether set value for `evaluate_raw` will make other side effect. Thus in the case `RpnExpression` is just generated manually.

components/tidb_query_expr/src/impl_like.rs
wangggong: okay

done

components/tidb_query_expr/src/impl_like.rs
wangggong: ditto

done

Xuanwo:
title should be `copr` instead of `corp`.

Xuanwo (approved):

wangggong:
> corp

Got it.

done~

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -66,13 +70,113 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+                Ok(val) => val,
+                Err(_) => return Ok(None),
+            };
+            match BytesRegex::new(val.as_ref()) {
+                Ok(regex) => Ok(Some(regex)),
+                Err(_) => Ok(None),
+            }
+        }
+        _ => Ok(None),
+    }
```
zhongzc, line 165: Consider refactoring the repeat code

zhongzc (approved):

Issue #6402, opened 2020-01-03
Engine abstraction


brson:
This issue is to organize and  track progress on abstracting TiKV over generic storage
engines (previous issue: https://github.com/tikv/tikv/issues/4184). There are many tasks to be done and contributions are welcome.

In scope: the process of abstracting the entire RocksDB engine API into the `engine_traits` and `engine_rocks` crates, decoupling TiKV and its test suite from RocksDB., and prototyping new engine bindings. Out of scope: refactoring `engine_traits` to better support with non-RocksDB engines.

The [`engine_traits` crate docs][engine_traits] describe the design and
migration process.

The issue is updated to contain a checklist of tasks that definitely need to
happen, along with name of who is working on it. When you decide to take a task,
say so on the issue before attempting it so that someone else doesn't duplicate
your work.

It's hard to know _in what order_ these need to happen, nor exactly how to
complete each of them, as each tends to present new challenges. Instead I have
documented here, in [the `engine_traits` crate docs][engine_traits], and in a [status report](todo), my experiences so far and suggestions for how to proceed.

[engine_traits]: https://github.com/tikv/tikv/blob/8b21c0f364b13ee3452f1c6c4d0526169e3d6481/components/engine_traits/src/lib.rs

The most important thing to keep in mind is:

Make small changes: small commits, small PRs. When you find that something else
needs to change first, stop what you are doing, and make that change, and submit
that PR. Don't let your patches grow out of control. Don't be afraid to throw away
your work and try something else.

### Migrating the `engine` abstractions

Take some abstraction defined in `engine`, implement it in `engine_traits` and
`engine_rocks`, change all callers to use the new version, delete the old
version. This abstraction can be a single function, an implementation of a
trait, a type, or an entire trait.

- [x] SstReader/Writer (brson)
- [x] IOLimiter (brson)
- [x] Snapshot (brson)
- [ ] Mutable for WriteBatch
- [ ] Peekable for DB
- [ ] Iterable for DB (hicqu)
- [ ] cf module
- [ ] util module
- [ ] rocks/util module
- [ ] Errors
- [ ] Engines

### Eliminating direct-use of `rocksdb` re-exports

`engine` reexports many types from `rocksdb`. Find one of those types, add an
equivalent abstraction to `engine_traits`, implement it for `engine_rocks`,
migrate all callers to use `engine_rocks`, delete the reexport so that it cannot
be used again.

- [ ] TablePropertiesCollection (brson)
[...]

PR #9243, opened 2020-12-13
Make PerfContext abstract and use it in raftstore

brson:

Problem Summary:

PerfContext is the last concrete rocks type used in raftstore. In order to remove engine_rocks as a dependency of raftstore, it must be abstracted in some way.

This PR makes PerfContext another engine trait. Unlike in rocks, where it is accessed as a global, this patch follows the existing engine_traits pattern, and access PerfContext through accessors on an engine instance. This seems more reasonable as a general-purpose abstraction.

On the other hand, this doesn't make any attempt to make the PerfContext measurements reasonable for any non-rocks engine. Instead, it makes the PerfContext accessor, now  called `get_perf_context` return an Option. For now, other engines are just expected to return None.

This patch then entirely removes engine_rocks::PerfContext and PerfLevel from raftstore. There are a few usages in the tikv crate that I have not touched yet. The next step is to remove engine_rocks as a dependency from raftstore.

### Related changes

### Check List 

- Unit test

### Release note 

- No release note.


Little-Wallace:
I think we do not need implement every field for an abstract PerfContext. I prefer follows:
```
trait PerfContext {
      fn start_observe(&self);
      fn report_metrics(&self);
}
impl PerfContext for RocksPerfContext {
     fn report_metrics(&self) {
          let mutex_time = self.0. db_condition_wait_nanos();
          APPLY_PERF_CONTEXT_TIME_HISTOGRAM_STATIC.with_label_values(&[self.name, "mutex_wait"]).observe(mutex_time);
         ....
     }
}
```

brson:
@Little-Wallace I am working to adapt this patch to do as you suggest.

That will be sufficient to cover the use cases in raftstore, which is good enough to let me get rid of the raftstore dependency on engine_rocks. It will not be sufficient to cover the use cases in the tikv crate though, which I am happy to tackle later.

brson:
@Little-Wallace I have completed the requested rework.

Now the API looks like

```rust
#[derive(Copy, Clone, Debug, PartialEq)]
pub enum PerfLevel {
    Uninitialized,
    Disable,
    EnableCount,
    EnableTimeExceptForMutex,
    EnableTimeAndCPUTimeExceptForMutex,
    EnableTime,
    OutOfBounds,
}

pub trait PerfContextExt {
    type PerfContext: PerfContext;

    fn get_perf_context(&self, level: PerfLevel, kind: PerfContextKind) -> Self::PerfContext;
}

#[derive(Eq, PartialEq, Copy, Clone)]
pub enum PerfContextKind {
    RaftstoreApply,
    RaftstoreStore,
}

pub trait PerfContext: Send {
    fn start_observe(&mut self);
    fn report_metrics(&mut self);
}
```

This is still a leaky abstraction, with rocks-specific perf levels and tikv-specific context-'kind'. The 'kind' is needed to distinguish which subsystem is reporting metrics. It is better than the previous API though. In the near-term it's not really feasible to get rid of these rocks-specific perf levels, since this enum is deserialized directly from the tikv config files. Doing anything about the abstraction of these perf levels probably needs to wait until we tackle multi-engine configuration, which will probably be one of the very last tasks involved in enabling alternate engines in a running tikv instance.

As I mentioned previously, there will need to be more work done to support the use-cases within the tikv crate.

Please review @Little-Wallace @5kbpers @hicqu @Connor1996

Little-Wallace:
Sorry for review this PR so later..... Shall we merge `APPLY_PERF_CONTEXT_TIME_HISTOGRAM_STATIC ` and `STORe_PERF_CONTEXT_TIME_HISTOGRAM_STATIC` to one? We can distinguish them by labels. Rest LGTM

Little-Wallace (approved):
LGTM

Connor1996 (approved):
LGTM

brson:
> Sorry for review this PR so later..... Shall we merge `APPLY_PERF_CONTEXT_TIME_HISTOGRAM_STATIC ` and `STORe_PERF_CONTEXT_TIME_HISTOGRAM_STATIC` to one? We can distinguish them by labels.

@Little-Wallace I don't understand prometheus enough to know how to do this. Can you explain it a bit more?

brson:
I've updated this to master, but as indicated in my last comment I don't know how to make the change you've suggested @Little-Wallace. I'm happy to if you explain a bit more how.

brson:
Updated to master.

Little-Wallace:
@brson Oh, It is enough. Maybe we can keep it and make refactor another day.....

brson:
Thanks @Little-Wallace !

PR #9245, opened 2020-12-14
txn: Unit test to check scheduler handle response_policy correctly

longfangsong:

Currently we don't have a test to check whether the scheduler handle `response_policy` correctly.

What's Changed:

Write tests for checking whether the correct stage the callback is called  according to `response_policy`. This is done by introducing a `MockEngine` which is a wrapper around `RocksEngine` but with the ability to assert the in comming writes, which should be also useful in other tests.

### Check List 

Tests 

- Unit test

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -1716,7 +1716,9 @@ mod tests {
     use super::{test_util::*, *};
 
     use crate::config::TitanDBConfig;
+    use crate::storage::kv::{ExpectedWrite, MockEngineBuilder};
     use crate::storage::mvcc::LockType;
+    use crate::storage::txn::commands::Prewrite;
     use crate::storage::{
         config::BlockCacheConfig,
         kv::{Error as EngineError, ErrorInner as EngineErrorInner},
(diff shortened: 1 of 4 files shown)


src/storage/mod.rs
```diff
@@ -5698,4 +5700,154 @@ mod tests {
         let res = rx.recv().unwrap().unwrap();
         assert_eq!(res.min_commit_ts, 1001.into());
     }
+
+    // this test shows that the scheduler take `response_policy` in `WriteResult` serious,
```
longfangsong, line 5704: This test should be in `scheduler.rs`, but I haven't found a good way to mock a scheduler.

longfangsong:
@nrc @youjiali1995 PTAL, maybe after #9244 got merged.

src/storage/mod.rs
```diff
@@ -5698,4 +5700,154 @@ mod tests {
...
+    #[test]
+    fn test_scheduler_response_policy() {
+        struct Case<T: 'static + StorageCallbackType + Send> {
+            expected_writes: Vec<ExpectedWrite>,
+
+            command: TypedCommand<T>,
+            pipelined_pessimistic_lock: bool,
+        }
+
+        fn run_case<T: 'static + StorageCallbackType + Send>(case: Case<T>) {
```
nrc, line 5715: Make this a method on Case?

nrc (approved):
one nit, otherwise lgtm

youjiali1995 (commented):
lgtm

ti-srebot (approved):

PR #9248, opened 2020-12-14
txn: Documentation for TypedCommand and command macro

pramodbiligiri:

Issue Number:  Addresses the need for code level documentation in transactions, as per https://github.com/tikv/sig-transaction/issues/25

What's Changed:
Documentation added for `TypedCommand`, the `command` macro and a couple of other traits.

I found it hard to follow the flow of the Request -> Command handling and thought these comments might help others in future. My difficulty lay in putting 3-4 pieces together: the macros (and their enums), the `From` traits invoked via .into() and the usage of PhantomData. There's also the StorageCallback and some async code there that I haven't dug into much.

Let me know whether the documentation I have added is correct, and whether it is worth keeping around in the code. If not in the code, I'll keep it around somewhere at least for my own reference. Also, if my docs show a lack of use of the right Rust terminology, please point out that as well.

Tests:
- No code

Side effects:
- None

/cc @andylokandy  and @nrc

### DIFF
--- a/src/storage/txn/commands/macros.rs
+++ b/src/storage/txn/commands/macros.rs
@@ -37,8 +37,8 @@ macro_rules! command {
         }
 
         impl $cmd {
-            /// Return a `TypedCommand` encapsulating the result of executing this command.
-            /// The result is of type `StorageCallbackType`
+            /// Return a `TypedCommand` that encapsulates the result of executing this command.
+            /// The actual result of the command implements the `StorageCallbackType` trait.
             pub fn new(
                 $($arg: $arg_ty,)*
                 ctx: crate::storage::Context,
--- a/src/storage/txn/commands/macros.rs
+++ b/src/storage/txn/commands/macros.rs
@@ -11,7 +11,7 @@ macro_rules! ctx {
     };
 }
 
-/// Generate the struct definition and Debug, Display methods for a passed in
+/// Generate the struct definition and Debug, Display methods for a passed-in
 /// storage command. An instance of this type is created from the `From`
 /// trait implementations of `TypedCommand`, when processing incoming storage
 /// requests of different types.
--- a/src/storage/txn/commands/macros.rs
+++ b/src/storage/txn/commands/macros.rs
@@ -12,9 +12,7 @@ macro_rules! ctx {
 }
 
 /// Generate the struct definition and Debug, Display methods for a passed-in
-/// storage command. An instance of this type is created from the `From`
-/// trait implementations of `TypedCommand`, when processing incoming storage
-/// requests of different types.
+/// storage command.
 /// Parameters:
 /// cmd -> Used as the type name for the generated struct. A variant of the
 /// enum `storage::txns::commands::Command` must exist whose name matches the
--- a/src/storage/txn/commands/macros.rs
+++ b/src/storage/txn/commands/macros.rs
@@ -19,9 +19,6 @@ macro_rules! ctx {
 /// value of `cmd` and which accepts one parameter whose type name matches
 /// the value of `cmd`.
 /// cmd_ty -> The type of the result of executing this command.
-/// A `From` conversion should exist in txn/commands/mod.rs from a `Command` enum
-/// variant corresponding to the value of `cmd` to a `TypedCommand`
-/// parameterized with the value of `cmd_ty`.
 /// display -> Information needed to implement the `Display` trait for the command.
 /// content -> The fields of the struct definition for the command.
 macro_rules! command {
--- a/src/storage/txn/commands/macros.rs
+++ b/src/storage/txn/commands/macros.rs
@@ -11,6 +11,15 @@ macro_rules! ctx {
     };
 }
 
+/// Generate the struct definition and Debug, Display methods for a passed in
+/// storage command.
+/// Parameters:
+/// cmd -> Used as the type name for the struct. It must match one of the variants of the enum
+/// `storage::txns::commands::Command`.
+/// cmd_ty -> The type that encapsulates the result of executing this command. It is passed as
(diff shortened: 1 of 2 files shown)


longfangsong (approved):
LGTM

(However `TypedCommand` is a very unatural thing, but it's so hard to remove it……)

src/storage/txn/commands/mod.rs
```diff
@@ -600,10 +623,16 @@ impl Debug for Command {
     }
 }
 
+/// Commands that do not need to modify the database during execution will implement this trait.
+/// Currently, this refers to the following variants of `storage::txn::commands::Command`:
```
nrc, line 627: Here and below, I don't think listing the commands is a good idea - it is difficult to keep such lists up to date.

src/storage/txn/commands/macros.rs
```diff
@@ -11,6 +11,21 @@ macro_rules! ctx {
     };
 }
 
+/// Generate the struct definition and Debug, Display methods for a passed in
```
nrc, line 14:
```suggestion
/// Generate the struct definition and Debug, Display methods for a passed-in
```

src/storage/txn/commands/macros.rs
```diff
@@ -11,6 +11,21 @@ macro_rules! ctx {
...
+/// storage command. An instance of this type is created from the `From`
```
nrc, line 15: This is not quite correct, Commands are created from grpc messages, the `From` impl is for conversion between typed and untyped versions.

src/storage/txn/commands/macros.rs
```diff
@@ -11,6 +11,21 @@ macro_rules! ctx {
...
+/// trait implementations of `TypedCommand`, when processing incoming storage
+/// requests of different types.
+/// Parameters:
+/// cmd -> Used as the type name for the generated struct. A variant of the
+/// enum `storage::txns::commands::Command` must exist whose name matches the
+/// value of `cmd` and which accepts one parameter whose type name matches
+/// the value of `cmd`.
+/// cmd_ty -> The type that encapsulates the result of executing this command.
```
nrc, line 23:
```suggestion
/// cmd_ty -> The type of the result of executing this command.
```

src/storage/txn/commands/macros.rs
```diff
@@ -11,6 +11,21 @@ macro_rules! ctx {
...
+/// A `From` conversion should exist in txn/commands/mod.rs from a `Command` enum
+/// variant corresponding to the value of `cmd` to a `TypedCommand`
+/// parameterized with the value of `cmd_ty`.
+/// display -> Information needed to implement the `Display` trait for the command.
+/// content -> The contents of the struct definition for the command.
```
nrc, line 28:
```suggestion
/// content -> The fields of the struct definition for the command.
```

src/storage/txn/commands/mod.rs
```diff
@@ -86,8 +86,31 @@ pub enum Command {
     MvccByStartTs(MvccByStartTs),
 }
 
+/// Stores the `Command` along with its return type, stored as generic parameter `T`.
```
nrc, line 89:
```suggestion
/// A `Command` with its return type reified as the generic parameter `T`.
```

src/storage/txn/commands/mod.rs
```diff
@@ -86,8 +86,31 @@ pub enum Command {
...
+///
+/// Incoming protobuf requests (like `CommitRequest`, `PrewriteRequest`) are converted to
```
nrc, line 91:
```suggestion
/// Incoming grpc requests (like `CommitRequest`, `PrewriteRequest`) are converted to
```

nrc (approved):
Looks good, thanks! I have a few suggestions inline. LGTM with those changes

ti-srebot (approved):

src/storage/txn/commands/mod.rs
pramodbiligiri: Done.

src/storage/txn/commands/macros.rs
pramodbiligiri: Done.

src/storage/txn/commands/macros.rs
pramodbiligiri: I need some clarification here: From what I can make out, both the Command enum variant and the Command struct instances are created in this flow that I've described above (Steps 3 and Steps 4 in my doc). Which is the part of the codebase that creates a Command from the grpc request?

src/storage/txn/commands/macros.rs
pramodbiligiri: Done.

src/storage/txn/commands/macros.rs
pramodbiligiri: Done.

src/storage/txn/commands/mod.rs
pramodbiligiri: Done.

src/storage/txn/commands/mod.rs
pramodbiligiri: Done.

src/storage/txn/commands/macros.rs
nrc: I think a Command is only created at the same time as a TypedCommand at the moment, but there is no reason a Command should not be created without a TypedCommand (this is done in tests, for example).

pramodbiligiri: Ah, make sense. I've removed that line. I just checked that the usage of the macro can be figured out using IntelliJ's Find Usages also. And now there is enough documentation around this piece, removing that one line won't lose much.

PR #9256, opened 2020-12-14
Add key and size threshold for load-base-split 

### LINKED ISSUE #9255: Add key and size threshold for load-base-split

The current load-base-split only receives the limitation of qps, so sometimes it may cut out a region with only a few keys, which is not good, especially in scenarios where there are more scan requests。

Add limits on the size and key of the region before split.

If we have a more accurate distribution of keys, we can even ensure that the region after split is not smaller than the threshold of key or size, avoiding the situation of Salami tactics.

lhy1024:

Issue Number: close #9255

Problem Summary:：

Add key and size threshold for load-base-split  to avoid unnecessarily split

add threshold and unit test

### Check List 

Tests 

- Unit test


nolouch (commented):
Could you share the test result and the workaround?

lhy1024:
image

the left is master, the right is this pr

the work load is:
```
1*tidb+1*pd+6*tikv

threads=256

sysbench --config-file=config --table_size=100000 --range_size=100 --point_selects=0 --sum_ranges=0 --order_ranges=0 --distinct_ranges=0 oltp_read_only run
```

disksing:
@BusyJay Could you help to review or assign some one to review?

BusyJay (approved):

HunDunDM (approved):

ti-srebot (approved):

PR #9068, opened 2020-11-19
raftstore: keep wake on isolation

BusyJay:
### What problem does this PR solve?

Issue Number: close #9022  

### What is changed and how it works?

What's Changed:

Prevent leader going to sleep when there are still peer catching up for logs to reduce corner cases.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 
- Prevent hibernation when there are still peers catching up logs

PR #6648, opened 2020-02-19
raftstore: wake up hibernated leader after failed read index

5kbpers:
Signed-off-by: 5kbpers <tangminghua@pingcap.com>

###  What have you changed?
Fix https://github.com/tikv/tikv/issues/6596
* Broadcast `MsgRegionWakeUp` to wake up potential hibernated leader when follower does not know who is the leader before requesting read index
* Change group state to `Ordered` after receiving a `MsgReadIndex` from other peers.

###  What is the type of the changes?

- Bugfix (a change which fixes an issue)

###  How is the PR tested?

- Unit test
- Integration test

###  Does this PR affect documentation (docs) or should it be mentioned in the release notes?

No.

###  Does this PR affect `tidb-ansible`?

No.

PR #6348, opened 2019-12-27
raftstore: retry pending read index requests

hicqu:
Signed-off-by: qupeng <qupeng@pingcap.com>

###  What have you changed?

On follwer replicas pending read index requests will be retried. So that read commands won't be blocked forever in raftstore.

###  What is the type of the changes?
Bugfix.

###  How is the PR tested?
- [ ] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

###  Does this PR affect documentation (docs) or should it be mentioned in the release notes?
No.

###  Does this PR affect `tidb-ansible`?
No.

###  Any examples? (optional)

PR #9260, opened 2020-12-14
tests: add test for replica read with split isolation

### LINKED ISSUE #6667: Broadcast WakeUp message for awakening hibernated regions after node restarts

After enable `hibernate regions`, the leader may go into `Idle` State. If a region splits during a node is down, after that node restarts, the newly created region will not be created on the node until the leader is awakened.

It may hurt the performance of replica read. We could add an **option** for broadcasting `MsgWakeUp` after a node is restarted.

5kbpers:

Issue Number: close #6667  

Problem Summary:
Previously tiflash gets read-index through `read_index` RPC, now it has changed to use follower read.
With #9068, a region that has pending peers will not go to sleep, and with #6648 and #6348, the failed follower read will be retried.
So for now we can consider that #6667 has been addressed.

What's Changed:
Add a test for covering the case.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A

### DIFF
--- a/tests/integrations/raftstore/test_replica_read.rs
+++ b/tests/integrations/raftstore/test_replica_read.rs
@@ -370,7 +370,7 @@ fn test_split_isolation() {
         if !resp.get_header().has_error() {
             return;
         }
-        thread::sleep(Duration::from_millis(100));
+        thread::sleep(Duration::from_millis(200));
     }
     panic!("test failed");
 }
--- a/tests/integrations/raftstore/test_replica_read.rs
+++ b/tests/integrations/raftstore/test_replica_read.rs
@@ -334,7 +334,7 @@ fn test_split_isolation() {
     cluster.must_split(&r1, b"k2");
     let idx = cluster.truncated_state(1, 1).get_index();
     // Trigger a log compaction, so the left region ['', 'k2'] cannot created through split cmd.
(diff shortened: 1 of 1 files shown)


tests/integrations/raftstore/test_replica_read.rs
```diff
@@ -309,3 +309,68 @@ fn test_read_index_out_of_order() {
...
+    must_get_equal(&cluster.get_engine(2), b"k2", b"v2");
+    cluster.must_transfer_leader(1, new_peer(1, 1));
+
+    cluster.stop_node(2);
+    // Split region into ['', 'k2') and ['k2', '')
+    let r1 = cluster.get_region(b"k2");
+    cluster.must_split(&r1, b"k2");
+    let idx = cluster.truncated_state(1, 1).get_index();
+    // Trigger a log compaction, so the left region ['', 'k2'] cannot created through split cmd.
+    for i in 0..cluster.cfg.raft_store.raft_log_gc_count_limit * 2 {
```
NingLin-P, line 337:
```suggestion
    for i in 2..cluster.cfg.raft_store.raft_log_gc_count_limit * 2 {
```
Maybe should start from `k2` so all put will send to the region 1

NingLin-P (approved):
Rest LGTM

BusyJay (commented):
LGTM

ti-srebot (approved):

PR #9266, opened 2020-12-15
copr: fix the different behavior between tidb and tikv in convert function.

wshwsh12:

Issue Number: close https://github.com/pingcap/tidb/issues/21695 

Problem Summary:
The condition value is NULL, but TiKV return the result.

Proposal: [xxx](url) 

What's Changed:
In the function Duration.from_i64, return a MaxDurationTime is meaningless.

When it throw an overflow error, the behavior should be the two situations:
1. If Flag::OVERFLOW_AS_WARNING is set, TiKV should handle the error, and return a NULL value.
2. If Flag::OVERFLOW_AS_WARNING is not set, TiKV should return the error. 

But now, the overflow number will be convert to MaxDurationTime, and make the condition ture. 
So I try to throw the error directly, can handle the error in `cast_int_as_duration`

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

- Fix the wrong behavior for convert function.

### DIFF
--- a/components/tidb_query_datatype/src/codec/mysql/duration.rs
+++ b/components/tidb_query_datatype/src/codec/mysql/duration.rs
@@ -499,16 +499,7 @@ impl Duration {
                     return t.convert(ctx);
                 }
             }
-            ctx.handle_overflow_err(Error::overflow("Duration", n))?;
-            // Returns max duration if overflow occurred
-            return Self::new_from_parts(
-                n.is_negative(),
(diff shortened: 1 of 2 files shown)


skyzh (approved):

Xuanwo (approved):
LGTM

ti-srebot (approved):

PR #9269, opened 2020-12-15
engine_rocks: rename and support dynamically changing auto-tuned mode of rate limiter

tabokie:

Issue Number: close #xxx 

Problem Summary:

Tune off auto-tuned in case it causes unexpected troubles. Also rename the config to `rate-limiter-auto-tuned`

What's Changed: support dynamically changing auto-tuned mode of rate limiter

### Related changes

- docs pingcap/docs#4422

### Check List 

Tests 

- Unit test

### Release note 

Support dynamically changing auto-tuned mode of RocksDB rate limiter


components/engine_panic/src/db_options.rs
```diff
@@ -36,6 +36,14 @@ impl DBOptions for PanicDBOptions {
         panic!()
     }
 
+    fn get_auto_tuned(&self) -> Option<bool> {
```
yiwu-arbug, line 39: can we get a better name, like `get_rate_limiter_is_auto_tuned`?

yiwu-arbug (approved):
LGTM

components/engine_panic/src/db_options.rs
tabokie: It does look a little off, I'll try changing the config name entirely to avoid similar confusion.

tabokie: Change made, I'm now sure how we handle the backward compatibility in this case, what do you think? @yiwu-arbug

yiwu-arbug: Since the underlying approach is changed, I think it is fine to not providing the old config name.

etc/config-template.toml
```diff
@@ -468,13 +468,12 @@
...
 ##    3 : AllIo
-## 4. auto_tuned enables dynamic adjustment of rate limit within the range
+## 4. rate-limiter-auto_tuned enables dynamic adjustment of rate limit within the range
 ##    [10MB/s, rate_bytes_per_sec], according to the recent demand for background I/O.
 # rate-bytes-per-sec = "10GB"
 # rate-limiter-refill-period = "100ms"
 # rate-limiter-mode = 2
-# auto-tuned = true
-
+# rate-limiter-auto-tuned = true
```
yiwu-arbug, line 476: let's also update docs.

components/engine_panic/src/db_options.rs
yiwu-arbug: regardless of whether the old auto-tuned was enabled, since we rename the config, all users will be using the new auto-tuned by default, which I don't see the down side.

src/config.rs
```diff
@@ -911,8 +911,7 @@ pub struct DbConfig {
     #[serde(with = "rocks_config::rate_limiter_mode_serde")]
     #[config(skip)]
     pub rate_limiter_mode: DBRateLimiterMode,
-    #[config(skip)]
-    pub auto_tuned: bool,
```
Connor1996, line 915: make it deprecated instead of delete it. you can refer to `end_point_max_task`

src/config.rs
```diff
@@ -937,6 +937,12 @@ pub struct DbConfig {
     pub ver_defaultcf: VersionCfConfig,
     #[config(skip)]
     pub titan: TitanDBConfig,
+
+    // deprecated. use rate_limiter_auto_tuned.
+    #[config(skip)]
+    #[doc(hidden)]
+    #[serde(skip_serializing)]
+    pub auto_tuned: Option<bool>,
```
Connor1996, line 945:
```suggestion
    pub auto_tuned: bool,
```

src/config.rs
```diff
@@ -2510,6 +2517,13 @@ impl TiKvConfig {
                 "raft_store.clean_stale_peer_delay",
             );
         }
+        if self.rocksdb.auto_tuned.is_some() {
+            warn!(
+                "deprecated configuration, {} is no longer used and ignored, please use {}.",
+                "rocksdb.auto_tuned", "rocksdb.rate_limiter_auto_tuned",
+            );
+            self.rocksdb.auto_tuned = None;
```
Connor1996, line 2525: let's override it
```suggestion
            self.rocksdb.rate_limiter_auto_tuned = self.rockdsb.auto_tuned;
```

Connor1996 (approved):

ti-srebot (approved):

PR #8115, opened 2020-06-20
rocksdb: Compaction guard: split SST by region boundaries

yiwu-arbug:
### What problem does this PR solve?

Problem Summary: Split SST files by TiKV region boundaries to reduce compaction IO.

### What is changed and how it works?

What's Changed: This PR adds the following configs (with default value. only showing default CF, same for other CFs):
```
[rocksdb.defaultcf]
enable_compaction_guard = false
compaction_guard_min_output_file_size = 8MB
compaction_guard_max_output_file_size = 128MB
```
SST files will be split by region boundary, i.e. if two keys falls in two TiKV region, they will be on different SST file. With the exception that 1. if the file size is smaller than compaction_guard_min_output_file_size, the split is not triggered, 2. if the file size is larger than compaction_guard_max_output_file_size, split is forced.

The region boundary obtained when splitting the SSTs, i.e. region can get split or merged between we obtain region boundary and compaction splitting SSTs. Also we lazily rely on compaction to split SST. We do not actively split SSTs after region split or merge.

The compaction_guard_min_output_file_size and compaction_guard_max_output_file_size is to prevent having too small or too large SST files. By having compaction_guard_min_output_file_size = 8MB and TiKV default region size of 96MB, compaction guard will mostly only split SST files at the bottom two levels of LSM. 

### Related changes

- https://github.com/tikv/rocksdb/pull/184
- Need to update docs.

### Check List 

Tests

Performance test. Need to add correctness test (e.g. bank2) later.

Side effects

- Compaction IO is expected to reduce by 5%-20%, depend on workload
- average SST file size is expected to grow to ~40MB, greatly reduce number of SST compare to default config (8M sst).

### Release note 

* Add `enable_compaction_guard` config to split SST by region boundaries to reduce compaction IO.

PR #9270, opened 2020-12-15
rocksdb: turn on compaction guard by default and add metrics

yiwu-arbug:

Problem Summary: This PR turn on compaction guard by default and improve metrics and unit test for the feature.

What's Changed:
* Turn on compaction guard for default CF and write CF.
* When `enable-compaction-guard = true`, override `target-file-size-base` with the value of `compaction-guard-max-output-file-size`, which defaults to 128M instead of 8M.
* compaction guard can only apply to default CF and write CF of kvdb.
* Add metrics to monitor compaction guard behavior, such as creating a new sst partitioner instance, splitting SST at region boundaries, or decided to skip to split SST file due to file being too small.
* add a unit test that enable compaction guard in rocksdb and check SST content.

### Related changes

- compaction guard was added in #8115.
- docs: https://github.com/pingcap/docs/pull/4420

### Check List 

Tests 
- new unit test
- will do manual testing with tpcc workload.

### Release note 
* Enable compaction guard by default, to split rocksdb SST files at TiKV region boundaries, to reduce overall compaction IO.


yiwu-arbug:
Will update config template and tikv-details grafana dashboard. But the rest of the PR is ready for review.

src/config.rs
```diff
@@ -493,7 +490,7 @@ impl Default for DefaultCfConfig {
             max_write_buffer_number: 5,
             min_write_buffer_number_to_merge: 1,
             max_bytes_for_level_base: ReadableSize::mb(512),
-            target_file_size_base: ReadableSize::mb(8),
+            target_file_size_base: ReadableSize::mb(128),
```
tabokie, line 493: I think the naming here is trickier than it seems. So the semantics of this config actually changes as we toggle the compaction guard, which could cause serious confusion. Maybe we can remove `compaction_guard_min_output_file_size` and implicitly takes the old `target_file_size_base` value as `compaction_guard_min_output_file_size` internally when guard is on? As for the maximum size limit we can keep the `compaction_guard_max_output_file_size` with less advertising.

yiwu-arbug: good suggestion.

yiwu-arbug: I think I'll keep the meaning and default value for `target_file_size_base`, but use `compaction_guard_max_output_file_size` to override it when compaction guard is enabled.

Connor1996 (commented):
most LGTM. And please update grafana template too

yiwu-arbug:
Updates:
* When enable-compaction-guard = true, override target-file-size-base with the value of compaction-guard-max-output-file-size, which defaults to 128M instead of 8M.
* compaction guard can only apply to default CF and write CF of kvdb.

yiwu-arbug:
Sample dashboard:
WeChatWorkScreenshot_007fe59a-6888-49b0-b56d-f7cafe4a35e7

yiwu-arbug:
I'm done. ready for review.

tabokie (approved):
LGTM

Connor1996 (approved):

ti-srebot (approved):

PR #9272, opened 2020-12-15
txn: change check txn status logic considering resolving lock type

cfzjywxk:

Problem Summary:

Make the returned status certain for check txn results.

What's Changed:

Add logic with the fields added in https://github.com/pingcap/kvproto/pull/705.

### Check List 

Tests 

- Unit test

Side effects

### Release note
- No release note, part of changing the txn api.


cfzjywxk:
@sticnarf @youjiali1995 @MyonKeminta PTAL

src/storage/txn/actions/check_txn_status.rs
```diff
@@ -99,6 +110,9 @@ pub fn check_txn_status_missing_lock<S: Snapshot>(
                 }
                 .into());
             }
+            if resolving_pessimistic_lock {
+                return Ok(TxnStatus::LockNotExistDoNothing);
+            }
```
sticnarf, line 115: Should it be before L106? In this case, I think it should return `LockNotExistDoNothing` instead of `TxnNotFound` error.

cfzjywxk: For resolving pessimistic lock, the `rollback_if_not_exist` will not be set at first and `txnNotFound` error is returned.
If we get a `txnNotFound` error resolving a secondary pessimistic lock, we could keep waiting until the seconrary lock itself expired(maybe we remove lock primary pessimistic lock first in the future), and then the `LockNotExistDoNothing ` will be returned, and we'll pessimistic rollback the secondary pessimistic lock.

sticnarf: Got it.

src/storage/txn/commands/check_txn_status.rs
```diff
@@ -874,10 +959,113 @@ pub mod tests {
...
+            k,
+            ts(3, 0),
+            ts(3, 0),
+            ts(4, 0),
+            true,
+            false,
+            true,
+            |s| s == LockNotExistDoNothing,
+        );
+        must_get_none(&engine, k, ts(5, 0));
```
youjiali1995, line 987: Seems it should be ` must_get_rollback_ts_none()`. `must_get_none` gets the value of the key.

src/storage/txn/commands/check_txn_status.rs
```diff
@@ -874,10 +959,113 @@ pub mod tests {
...
+            ts(10, 0),
+            ts(21, 0),
+            ts(21, 0),
+            true,
+            false,
+            true,
+            |s| s == PessimisticRollBack,
+        );
+        must_unlocked(&engine, k);
+        must_get_none(&engine, k, ts(22, 0));
```
youjiali1995, line 1022: `must_get_rollback_ts_none`

src/storage/txn/commands/check_txn_status.rs
```diff
@@ -874,10 +959,113 @@ pub mod tests {
...
+
+        // Path: the prewrite primary key lock dose exist, and it's not expired yet.
```
youjiali1995, line 1024:
```suggestion
        // Path: the prewrite primary key lock does exist, and it's not expired yet.
```

src/storage/txn/commands/check_txn_status.rs
sticnarf, line 987: `must_get_none` only checks there is no committed value. You can use `must_seek_write_none`.

src/storage/txn/commands/check_txn_status.rs
```diff
@@ -874,10 +959,113 @@ pub mod tests {
...
+            ts(41, 0),
+            ts(41, 0),
+            true,
+            false,
+            true,
+            |s| s == TtlExpire,
+        );
+        must_unlocked(&engine, k);
+        must_get_rollback_ts(&engine, k, ts(30, 0));
+    }
```
youjiali1995, line 1067: Need one more case: `resolving_pessimistic_lock` is false and the primary key is a pessimistic lock. Rollback record should be written and the transaction status is certain.

src/storage/txn/commands/check_txn_status.rs
```diff
@@ -874,10 +959,113 @@ pub mod tests {
...
+            true,
+            |s| s == LockNotExistDoNothing,
+        );
+        must_get_none(&engine, k, ts(5, 0));
+
+        // Path: there is no commit or rollback record, error should be reported if
+        // rollback_if_not_exist is set to false.
+        must_err(&engine, k, ts(3, 0), ts(5, 0), ts(5, 0), false, false, true);
+
+        // Path: the pessimistic primary key lock dose exist, and it's not expired yet.
```
sticnarf, line 993: does
```suggestion
        // Path: the pessimistic primary key lock does exist, and it's not expired yet.
```

sticnarf: And same for below

src/storage/txn/commands/check_txn_status.rs
sticnarf: And same for below

sticnarf: ah there's an existing comment now

sticnarf:
LGTM

src/storage/txn/commands/check_txn_status.rs
sticnarf: I think it's covered in L826.

sticnarf: I think it's covered in L826

youjiali1995 (commented):
lgtm

ti-srebot (approved):

PR #9283, opened 2020-12-16
backup: support split big region into small backup files

### LINKED ISSUE #9144: TiKV OOM because BR read a large region in memory

release-4.0

BR will read all data of a region and fill it in a SST writer. But it is in-memory, see: https://github.com/tikv/tikv/blob/master/components/backup/src/writer.rs#L138.
So if there is a huge region, TiKV may crash for OOM because of keeping all data of this region in memory.

### What did happened?

- TiKV crashed because of OOM.

lichunzhu:

Issue Number: close #9144 

Problem Summary: BR will read all data of a region and fill it in a SST writer. But it is in-memory. If there is a huge region, TiKV may crash for OOM because of keeping all data of this region in memory.

What's Changed: Record the written txn entries' size. When it reaches `region_max_size`, we will save the data cached in RocksDB to a SST file and then switch to the next file.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
1. Set `sst-max-size` to 15MiB.
```
mysql> select * from CLUSTER_CONFIG where `TYPE`="tikv";
+------+-----------------+---------------------------------------------------------------+------------------------------------------------------+
| TYPE | INSTANCE        | KEY                                                           | VALUE                                                |
+------+-----------------+---------------------------------------------------------------+------------------------------------------------------+
| tikv | 127.0.0.1:20160 | backup.batch-size                                             | 8                                                    |
| tikv | 127.0.0.1:20160 | backup.num-threads                                            | 9                                                    |
| tikv | 127.0.0.1:20160 | backup.sst-max-size                                           | 15MiB                                                |
...
```
2. Backup around 100MB data(without compaction) successfully.
```
$ ./br backup full -s ./backup --pd http://127.0.0.1:2379 
Full backup <--------------------------------------------------------------------------------------------------------------------------------------------------------------------> 100.00%
Checksum <-----------------------------------------------------------------------------------------------------------------------------------------------------------------------> 100.00%
[2020/12/31 14:39:12.534 +08:00] [INFO] [collector.go:60] ["Full backup Success summary: total backup ranges: 2, total success: 2, total failed: 0, total take(Full backup time): 4.273097395s, total take(real time): 8.133315406s, total kv: 8000000, total size(MB): 361.27, avg speed(MB/s): 84.55"] ["backup checksum"=901.754111ms] ["backup fast checksum"=6.09384ms] ["backup total regions"=10] [BackupTS=421893700168974340] [Size=48023090]
```
3. The big region can be split into several files:
```
-rw-r--r-- 1 * * 1.5M Dec 31 14:39 1_60_28_74219326eeb0a4ae3a0f5190f7784132bb0e44791391547ef66862aaeb668579_1609396745730_write.sst
-rw-r--r-- 1 * * 1.2M Dec 31 14:39 1_60_28_b7a5509d9912c66a21589d614cfc8828acd4051a7eeea3f24f5a7b337b5a389e_1609396746062_write.sst
-rw-r--r-- 1 * * 1.5M Dec 31 14:39 1_60_28_cdcc2ce1c18a30a2b779b574f64de9f0e3be81c2d8720d5af0a9ef9633f8fbb7_1609396745429_write.sst
-rw-r--r-- 1 * * 2.4M Dec 31 14:39 1_62_28_4259e616a6e7b70c33ee64af60230f3e4160af9ac7aac723f033cddf6681826a_1609396747038_write.sst
-rw-r--r-- 1 * * 2.4M Dec 31 14:39 1_62_28_5d0de44b65fb805e45c93278661edd39792308c8ce90855b54118c4959ec9f16_1609396746731_write.sst
-rw-r--r-- 1 * * 2.4M Dec 31 14:39 1_62_28_ef7ab4b5471b088ee909870e316d926f31f4f6ec771754690eac61af76e8782c_1609396747374_write.sst
-rw-r--r-- 1 * * 1.5M Dec 31 14:39 1_64_29_74211aae8215fe9cde8bd7ceb8494afdcc18e5c6a8c5830292a577a9859d38e1_1609396746671_write.sst
-rw-r--r-- 1 * * 1.2M Dec 31 14:39 1_64_29_81e152c98742938c1662241fac1c841319029e800da6881d799a16723cb42888_1609396747010_write.sst
-rw-r--r-- 1 * * 1.5M Dec 31 14:39 1_64_29_ce0dde9826aee9e5ccac0a516f18b9871d3897effd559ff7450b8e56ac449bbd_1609396746349_write.sst
-rw-r--r-- 1 * *   78 Dec 31 14:39 backup.lock
-rw-r--r-- 1 * * 229K Dec 31 14:39 backupmeta
```
4. Restore backuped data. It works successfully and passes the manual check.
```
./br restore full -s ./backup --pd http://127.0.0.1:2379
Full restore <-------------------------------------------------------------------------------------------------------------------------------------------------------------------> 100.00%
[2020/12/31 14:42:49.983 +08:00] [INFO] [collector.go:60] ["Full restore Success summary: total restore files: 27, total success: 27, total failed: 0, total take(Full restore time): 5.063048828s, total take(real time): 7.84620924s, total kv: 8000000, total size(MB): 361.27, avg speed(MB/s): 71.36"] ["split region"=26.217737ms] ["restore checksum"=4.10792638s] ["restore ranges"=26] [Size=48023090]
```

### Release note 

- Fix the problem that TiKV OOM when we backup a huge region.


src/config.rs
```diff
@@ -2160,6 +2160,7 @@ mod readpool_tests {
 pub struct BackupConfig {
     pub num_threads: usize,
     pub batch_size: usize,
+    pub region_max_size: ReadableSize,
```
kennytm, line 2163: this seems irrelevant to `coprocessor.region-max-size` besides the init value

perhaps better name it something like `sst_max_size` / `backup_file_max_size` / `sst_max_kv_size` / ...

src/storage/txn/store.rs
```diff
@@ -170,6 +170,18 @@ impl TxnEntry {
...
+        match self {
+            TxnEntry::Commit {
+                default: _, write, ..
+            } => Ok(Key::from_encoded(write.0.clone())
+                .truncate_ts()?
+                .into_raw()?),
+            // Prewrite are not support
+            _ => unreachable!(),
+        }
+    }
```
kennytm, line 184:
```suggestion
    pub fn as_key(&self) -> Result<&[u8]> {
        match self {
            TxnEntry::Commit { write, .. } => Ok(Key::truncate_ts_for(&write.0)?),
            // Prewrite are not support
            _ => unreachable!(),
        }
    }
```

not sure if we should ignore the `default` cf

components/backup/src/endpoint.rs
```diff
@@ -302,38 +353,40 @@ impl BackupRange {
...
-        let mut writer = match BackupWriter::new(
+        let writer = match BackupWriter::new(
             db,
+            store_id,
             &file_name,
             storage.limiter.clone(),
             compression_type,
             compression_level,
+            region_max_size,
+            region,
```
overvenus, line 373: It accepts too many arguments, could refactor into builder pattern? https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

src/config.rs
```diff
@@ -2535,6 +2538,7 @@ impl TiKvConfig {
                     + self.raftdb.defaultcf.block_cache_size.0,
             });
         }
+        self.backup.region_max_size = self.coprocessor.region_max_size;
```
overvenus, line 2541: We should sanitize the value is in a reasonable range. Eg, a large region max size could leads to OOM during backup.

overvenus (commented):
Could add tests to cover the changes?

src/storage/txn/store.rs
lichunzhu: From @3pointer, the keys of `default` and `write` are the same.

src/config.rs
lichunzhu: Will this happen?

components/backup/src/endpoint.rs
```diff
@@ -224,17 +227,65 @@ impl BackupRange {
...
             }
             debug!("backup scan entries"; "len" => batch.len());
+
+            let entries = batch.drain();
+            if writer.need_split_keys() {
+                let res = {
+                    entries.as_slice().get(0).map_or_else(
+                        || Err(Error::Other(box_err!("get entry error"))),
+                        |x| match x.as_key() {
+                            Ok(k) => writer.rebuild(Option::from(Key::from_raw(&k))),
```
kennytm, line 237:
```suggestion
                            Ok(k) => writer.rebuild(Some(Key::from_raw(&k))),
```

kennytm (commented):
rest LGTM

src/config.rs
```diff
@@ -2570,6 +2573,7 @@ impl TiKvConfig {
                     + self.raftdb.defaultcf.block_cache_size.0,
             });
         }
+        self.backup.sst_max_size = self.coprocessor.region_max_size;
```
overvenus, line 2576: Looks like we can adjust the sst_max_size via config file? It always be set to `region_max_size`?

overvenus (commented):
Please add a test case to make sure it can write a large range into multiple SST files.

components/backup/src/writer.rs
```diff
@@ -352,14 +430,25 @@ mod tests {
         let storage = external_storage::create_storage(&backend).unwrap();
 
         // Test empty file.
-        let mut writer =
-            BackupWriter::new(db.get_sync_db(), "foo", Limiter::new(INFINITY), None, 0).unwrap();
+        let mut r = kvproto::metapb::Region::default();
+        r.set_id(1);
+        r.mut_peers().push(new_peer(1, 1));
+        let mut writer = BackupWriter::new(db.get_sync_db(), "foo", None, 0)
```
overvenus, line 436: Could you add a new struct `BackupWriterBuilder`?

src/config.rs
```diff
@@ -2574,6 +2577,13 @@ impl TiKvConfig {
                     + self.raftdb.defaultcf.block_cache_size.0,
             });
         }
+        if self.backup.sst_max_size == default_coprocessor.region_split_size {
+            warn!(
+                "override backup.sst-max-size with coprocessor.region-max-size, {:?}",
+                self.coprocessor.region_max_size
+            );
+            self.backup.sst_max_size = self.coprocessor.region_max_size;
+        }
```
overvenus, line 2586: I think we should make sure the value is in a reasonable range, eg [default_region_max_size/10, default_region_max_size * 2]

components/backup/tests/integrations/test_backup.rs
```diff
@@ -462,9 +464,106 @@ fn test_backup_and_import() {
...
+        vec![],   // start
+        vec![],   // end
+        0.into(), // begin_ts
+        backup_ts,
+        &make_unique_dir(tmp.path()),
+    );
+    let resps3 = block_on(rx.collect::<Vec<_>>());
+    assert_eq!(files1.len(), resps3[0].files.len());
+
+    // different backup should be the same except the file names(because since_the_epoch are different)
```
overvenus, line 552: Could you set file range based on the keys in SST file instead of region range? So we can avoid restoring a huge region.

overvenus (commented):
Rest LGTM

src/storage/txn/store.rs
```diff
@@ -170,6 +170,16 @@ impl TxnEntry {
             _ => unreachable!(),
         }
     }
+    /// This method will generate this kv pair's key
+    pub fn as_key(&self) -> Result<Key> {
```
kennytm, line 174: as this method is no longer O(1) this method should be called `.to_key()`.

kennytm:
PTAL @overvenus (i've already [given LGT before](https://github.com/tikv/tikv/pull/9283#issuecomment-750452191) so not gonna send the command again 🙃)

components/backup/tests/integrations/test_backup.rs
```diff
@@ -457,14 +489,102 @@ fn test_backup_and_import() {
...
+        backup_ts,
+        &storage_path,
+    );
+    let resps1 = block_on(rx.collect::<Vec<_>>());
+    // Only leader can handle backup.
+    assert_eq!(resps1.len(), 1);
+    let files1 = resps1[0].files.clone();
+    // Short value is piggybacked in write cf, so we get 1 sst at least.
+    assert!(!resps1[0].get_files().is_empty());
+    assert_eq!(files1.len(), 2);
```
overvenus, line 522: Please make sure these two files have different start/end key.

overvenus (commented):
Does raw kv support split huge regoin?

components/backup/tests/integrations/test_backup.rs
lichunzhu: addressed in https://github.com/tikv/tikv/pull/9283/commits/0601fbb192d932a4469ba2bafa075c7883418292

lichunzhu:
> Does raw kv support split huge regoin?

No. I think we can do this in another PR.

overvenus:
LGTM

Issue #9115, opened 2020-11-25
encryption: handle non-atomic file operations


yiwu-arbug:
## Bug Report

When encryption (TDE) is enabled, file operations like delete file/rename file/copy file are non-atomic. For these operations, we need to update encryption metadata via `encryption::DataKeyManager`, which cannot combine with the actual file operation into one atomic operation. When TiKV crash in between file operation and encryption metadata update, encryption metadata could be inconsistent with the file system. This is fine, providing that as long as an encrypted file exist in the file system, the corresponding entry in encryption metadata always exists. And whenever we find there's an obsolete entry in the metadata without corresponding file, the entry should be ignored and removed. Current implementation is not consistent for all `DataKeyManager` call sites. We need to refactor the code to provide helper method that work in the following way:

* delete file: works in the order 1. fs::delete_file(), 2. key_manager.delete_file()
* rename file: works in the order 1. key_manager.new_file(dst), 2. fs::rename_file(), 3. key_manager.delete_file(src)
* link (copy) file: works in the order 1. key_manager.new_file(dst), 2. fs::copy_file()

And whenever these operations see unexpected existing entry in metadata (for example, when link file and see metadata for dst already exists), check if the file exists in the file system. If not, ignore and remove the entry.

The task include changing TiKV code and RocksDB code.

### What version of TiKV are you using?
master, 4.0.x

### What operating system and CPU are you using?
N/A

### Steps to reproduce
#9099 for example

### What did you expect?
encryption logic should handle non-atomic file operations like delete/rename/link files correctly

PR #9285, opened 2020-12-16
Check stale file information and remove `Rename` operation in encryption module

hunterlxt:

Issue Number: close https://github.com/tikv/tikv/issues/9115

Problem Summary:

This PR is to solve the stale information encountered in the encryption dictionary

- Delete: When deleting a file but failing to execute `delete_file` on `file_dict`, the stale information leaves
- Rename: When a new `dst_fname` is added to the dictionary file, but the real `rename` is not successful or execute `rename` but failed to delete `src_fname`
- Link: When a new `dst_fname` is added to the dictionary file, the subsequent operations don't finish.

In the above situation, some stale file information will appear in the encrypted file dictionary. We need to deal with these errors instead of generating and returning errors.

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- Check stale file information from encryption file dict


yiwu-arbug (commented):
Looking good. Shall we also do the following?
* extend `file_system::copy` to handle encryption logic, by accepting `DataKeyManager` as parameter. implement a similar `file_system::rename` helper method.
* update all TiKV logic that call `DataKeyManager::link_file` and `DataKeyManager::rename_file` previously to use new helper methods.

components/encryption/src/manager/mod.rs
```diff
@@ -250,12 +250,7 @@ impl Dicts {
                     return Ok(());
                 }
             };
-            if file_dict.files.get(dst_fname).is_some() {
-                return Err(Error::Io(IoError::new(
-                    ErrorKind::AlreadyExists,
-                    format!("file already exists, {}", dst_fname),
-                )));
-            }
+            check_stale_file_exist(dst_fname, &mut file_dict, &mut file_dict_file)?;
```
Connor1996, line 253: add some comments here to explain why there may be some stale files.

Connor1996 (commented):
rest LGTM

yiwu-arbug:
> Looking good. Shall we also do the following?
> 
> * extend `file_system::copy` to handle encryption logic, by accepting `DataKeyManager` as parameter. implement a similar `file_system::rename` helper method.
> * update all TiKV logic that call `DataKeyManager::link_file` and `DataKeyManager::rename_file` previously to use new helper methods.

discussed with @hunterlxt offline. These will be included in another PR.

yiwu-arbug (approved):
LGTM

Connor1996 (approved):
LGTM

ti-srebot (approved):

hunterlxt:
Waiting https://github.com/tikv/rocksdb/pull/222 and related change in rust-rocksdb. Those changes should be included into this PR.

components/raftstore/src/store/snap.rs
```diff
@@ -1470,7 +1470,8 @@ impl SnapManagerCore {
             let dst = cf_file.path.to_str().unwrap();
             // It's ok that the cf file is moved but machine fails before `mgr.rename_file`
             // because without metadata file, saved cf files are nothing.
-            mgr.rename_file(src, dst)?;
+            mgr.new_file(dst)?;
```
yiwu-arbug, line 1473: should be `link_file`?

yiwu-arbug: add a method to encapsulate the `link_file` then `delete_file` logic?

components/engine_rocks/Cargo.toml
```diff
@@ -48,7 +48,8 @@ protobuf = "2"
 fail = "0.4"
 
 [dependencies.rocksdb]
-git = "https://github.com/tikv/rust-rocksdb.git"
+git = "https://github.com/ti-srebot/rust-rocksdb"
```
yiwu-arbug, line 51: revert this?

hi-rustin:
@tabokie @hunterlxt  You don't have permission because this pr doesn't have a sig label.

Please don't add the status/can-merge label manually, add the sig label and you'll have access.

Also we don't need to keep triggering `/merge` now, please read bot's reply carefully. We just need to trigger the test.

PR #9286, opened 2020-12-16
copr: implement json cast as bool

wshwsh12:
Signed-off-by: wshwsh12 <793703860@qq.com>

Issue Number: close https://github.com/pingcap/tidb/issues/12233 

Problem Summary: port  https://github.com/pingcap/tidb/pull/18948

Proposal: [xxx](url) 

What's Changed:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 
- Implement cast json as bool.


components/tidb_query_datatype/src/codec/datum.rs
```diff
@@ -374,6 +375,7 @@ impl Datum {
             Datum::Time(t) => Some(!t.is_zero()),
             Datum::Dur(d) => Some(!d.is_zero()),
             Datum::Dec(d) => Some(ConvertTo::<f64>::convert(&d, ctx)?.round() != 0f64),
+            Datum::Json(j) => Some(j.as_ref().as_mysql_bool(ctx).unwrap()),
```
skyzh, line 378: Let's replace `unwrap()` with `?` here, in case that future work makes `as_mysql_bool` returns Err.

skyzh (commented):
Good work!

components/tidb_query_datatype/src/codec/mysql/json/mod.rs
```diff
@@ -203,6 +203,19 @@ impl<'a> JsonRef<'a> {
...
+    // Return whether the value is zero.
+    pub(crate) fn is_zero(&self) -> bool {
+        match self.get_type() {
+            JsonType::Object => false,
+            JsonType::Array => false,
+            JsonType::Literal => false,
+            JsonType::I64 => self.get_i64() == 0,
+            JsonType::U64 => self.get_u64() == 0,
+            JsonType::Double => self.get_double() == 0f64,
+            JsonType::String => false,
```
Xuanwo, line 216: We don't need to check the json string's length here?

wshwsh12: I don't think so. I have a test in MySQL8.0.  
```
MySQL [test]> select *,json_extract(j, '$.test') from testjson where json_extract(j, '$.test');
+----+-----------------+---------------------------+
| id | j               | json_extract(j, '$.test') |
+----+-----------------+---------------------------+
|  1 | {"test": ""}    | ""                        |
|  2 | {"test": ""}    | ""                        |
|  3 | {"test": true}  | true                      |
|  4 | {"test": false} | false                     |
+----+-----------------+---------------------------+
4 rows in set, 1 warning (0.000 sec)
```

components/tidb_query_datatype/src/codec/datum.rs
wshwsh12: Addressed.

components/tidb_query_datatype/src/codec/mysql/json/mod.rs
Xuanwo: Yes, the logic here is correctly, here always return `false` no matter the value is.

Maybe we can add ref: https://dev.mysql.com/doc/refman/8.0/en/json.html#Converting%20between%20JSON%20and%20non-JSON%20values here?

```
MariaDB [test]> CREATE TABLE JSON (c JSON);
MariaDB [test]> insert into json values row('{"a":"","b":"abc"}');
MariaDB [test]> select if(JSON_EXTRACT(c, '$.a') = true, 'TRUE', 'FALSE') from JSON;
+----------------------------------------------------+
| if(JSON_EXTRACT(c, '$.a') = true, 'TRUE', 'FALSE') |
+----------------------------------------------------+
| FALSE                                              |
+----------------------------------------------------+
1 row in set (0.001 sec)
MariaDB [test]> select if(JSON_EXTRACT(c, '$.b') = true, 'TRUE', 'FALSE') from JSON;
+----------------------------------------------------+
| if(JSON_EXTRACT(c, '$.b') = true, 'TRUE', 'FALSE') |
+----------------------------------------------------+
| FALSE                                              |
+----------------------------------------------------+
1 row in set (0.001 sec)
```

Xuanwo (approved):
LGTM

skyzh (approved):

ti-srebot (approved):

PR #9292, opened 2020-12-18
raftstore: refactor local read to reduce the number of read-proposal redirect to raftstore

### LINKED ISSUE #8724: Reduce the number of read-proposal send to raft-store-thread
### Problem Description
image

We found that when TiKV call `raftkv.async_snapshot`,  there are a lot of read-proposal send to raft-store-thread, but only few of them execute as `read-index` and most of them execute as `local-read`. It means that most of request are still in a leader lease but they are still send to raft-store-thread. LocalReader may believe the lease of leader is expired mistakenly.

We need to investigate this problem and find how to make local reader judge lease more accurately.

NingLin-P:

Issue Number: close #8724 

Problem Summary:

Although we have local reader, a read request may still be redirected to raftstore if:
- the `ReadDelegate`'s remote lease is expired
- Some checks not pass, it may cause by the local `ReadDelegate`'s stale infos (i.e region epoch) 

What's Changed:

We use `AtomicU64` as the remote lease, if the lease is expired, it is most likely the leader lease is also expired so we have to redirect to raftstore to renew lease. But for the local `ReadDelegate`'s infos we can always refresh it from `StoreMeta` with careful `lock()` calling 

Tests 

- Unit test

### Release note 
- No release note


BusyJay:
What's the improvement?

NingLin-P:
> What's the improvement?

Make local `ReadDelegate` more up-to-date with `StoreMeta.readers` and less `store_meta.lock()` calling, also with some code refactor.

Little-Wallace (approved):
LGTM

NingLin-P:
https://github.com/tikv/tikv/blob/2f526ab7776544ad3cdc07db8e1fe207211a5144/components/raftstore/src/store/worker/read.rs#L462-L464

If the local `readDelegate`'s infos (like `epoch`, `applied_index_term`, etc) is stale, the `break` at the above code will execute (and cause read request redirect to `raftstore`). `self.delegates.get(&region_id).is_some()` will be true because `delegates: HashMap<u64, Option<ReadDelegate>>` and 

https://github.com/tikv/tikv/blob/2f526ab7776544ad3cdc07db8e1fe207211a5144/components/raftstore/src/store/worker/read.rs#L368-L370

NingLin-P:
I had done some benchmarks with a read-only workload, the results look good:
before:
截屏2021-01-06 下午1 32 04
after:
截屏2021-01-06 下午2 00 31

NingLin-P:
More surprisingly, this PR even reduce the local reader's `reject-by-lease_expire` count and little improvement on the QPS:

`reject-by-lease_expire` before:
截屏2021-01-06 下午1 31 26

`reject-by-lease_expire` after:
截屏2021-01-06 下午2 10 04

QPS before:
截屏2021-01-06 下午1 30 36

QPS after:
截屏2021-01-06 下午2 09 13

components/raftstore/src/store/worker/read.rs
```diff
@@ -146,6 +146,47 @@ pub struct ReadDelegate {
...
     max_ts_sync_status: Arc<AtomicU64>,
+
+    track_ver: TrackVer,
+}
+
+#[derive(Debug)]
+struct TrackVer {
+    version: Arc<AtomicU64>,
+    local_ver: u64,
+    source: bool,
```
BusyJay, line 157: What does `source` mean?

components/raftstore/src/store/worker/read.rs
```diff
@@ -264,7 +307,7 @@ where
     kv_engine: E,
     metrics: ReadMetrics,
     // region id -> ReadDelegate
-    delegates: HashMap<u64, Option<ReadDelegate>>,
+    delegates: HashMap<u64, Arc<ReadDelegate>>,
```
BusyJay, line 310: Can it cause memory leak? For example, after many round of splitting and transferring replicas, some peer may be deleted while local reader still keeps a copy of delegate.

components/raftstore/src/store/worker/read.rs
NingLin-P: It means the one created by `new` instead of `clone`, more specific, only the `ReadDelegate` created by `ReadDelegate::new` (store at `StoreMeta`) can increase version, because only these `ReadDelegate` will invoke `ReadDelegate::update`

components/raftstore/src/store/worker/read.rs
NingLin-P: Yes, but it is not caused by changing `Option` to `Arc`. A peer will call `mark_invalid` when it is destroyed, and when the local reader handles requests, it will check its local `ReadDelegate`'s `invalid` flag and remove the `ReadDelegate` if the flag is true, but if there is no incoming request the local `ReadDelegate` will not be removed.

components/raftstore/src/store/worker/read.rs
BusyJay: Can you also add it as comment?

NingLin-P: Done, PTAL again

components/raftstore/src/store/worker/read.rs
```diff
@@ -987,5 +1036,86 @@ mod tests {
...
+        {
+            let mut meta = store_meta.lock().unwrap();
+            meta.readers
+                .get_mut(&1)
+                .unwrap()
+                .update(Progress::applied_index_term(2));
+        }
+        assert_eq!(reader.get_delegate(1).unwrap().applied_index_term, 2);
+
+        {
```
BusyJay, line 1111: I don't think you need to put them in brackets. Now compiler is smart enough to infer the correct lifetime.

BusyJay (commented):
Because lease doesn't record the left bound, can it be a problem when not dropping the delegate? For example, in some case a lease should be [3, 10), but a request checking the lease using 2, it should not be in lease but we can't tell it at the moment.

components/raftstore/src/store/worker/read.rs
```diff
@@ -146,6 +146,51 @@ pub struct ReadDelegate {
     invalid: Arc<AtomicBool>,
     pub txn_extra_op: Arc<AtomicCell<TxnExtraOp>>,
     max_ts_sync_status: Arc<AtomicU64>,
+
+    track_ver: TrackVer,
+}
+
+#[derive(Debug)]
+struct TrackVer {
+    version: Arc<AtomicU64>,
```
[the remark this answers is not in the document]
NingLin-P: I think it is not a good idea, they serve different purposes, mix-use of them will cause UB.

NingLin-P:
> Because lease doesn't record the left bound, can it be a problem when not dropping the delegate? For example, in some case a lease should be [3, 10), but a request checking the lease using 2, it should not be in lease but we can't tell it at the moment.

It won't be a problem, when handling a request, the delegate's `invalid` flag will be checked first, and the delegate will be removed and the request will redirect to raftstore in this case.

components/raftstore/src/store/worker/read.rs
NingLin-P: Done, PTAL https://github.com/tikv/tikv/issues/9469

BusyJay (approved):

components/raftstore/src/store/worker/read.rs
```diff
@@ -987,5 +1036,76 @@ mod tests {
...
+            track_ver: TrackVer::new(),
+        };
+        let mut meta = store_meta.lock().unwrap();
+        meta.readers.insert(1, read_delegate);
+        let d = reader.get_delegate(1).unwrap();
+        assert_eq!(&*d.region, &region);
+        assert_eq!(d.term, 1);
+        assert_eq!(d.applied_index_term, 1);
+        assert!(d.leader_lease.is_none());
+        drop(d);
```
BusyJay, line 1082: You need to also drop meta at the end of every block.

components/raftstore/src/store/worker/read.rs
NingLin-P: I revert the change, remove the bracket will make the test blocking, because `reader.get_delegate` may also try to `store_meta.lock()`

BusyJay (approved):

PR #6427, opened 2020-01-07
raftstore: add a new tick to renew lease in advance

5kbpers:
ref #11579

###  What have you changed?
Currently, the read lease of a leader is renewed after proposing a write or executing a read request with read index.
In some situations, a client may always try to read data after the read lease expired, read requests would be handled with read index, then the duration of read request could be higher than expected.
This PR introduces following changes:
* Solve this problem by renewing the read lease in a tick.
* Fix the metrics of awaken hibernated regions

###  What is the type of the changes?

- New feature (a change which adds functionality)
- Improvement (a change which is an improvement to an existing feature)
- Bugfix (a change which fixes an issue)
- Engineering (engineering change which doesn't change any feature or fix any issue)
- Misc (other changes)

###  How is the PR tested?

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

###  Does this PR affect documentation (docs) or should it be mentioned in the release notes?
No

###  Does this PR affect `tidb-ansible`?
No

###  Benchmark result if necessary (optional)

**TODO**

### Release note 
```release-note
renew lease on a new tick to cover lease till next tick
```

PR #9307, opened 2020-12-21
raftstore: renew leader lease in advance when handle read request

NingLin-P:

ref #11579

Problem Summary:

Currently, the leader lease will be renewed when there are successful writes or read index, but if there are only read requests on the leader, the lease will not be renewed until it is expired and then the leader send read index request to renew it, which cause latency jitter, https://github.com/tikv/tikv/pull/6427 try to renew lease when raft basic tick triggered, but it is not compatible with the hibernate region feature which will skip raft basic tick.

Blocked by #6427

What's Changed:

When the local reader handle a read request, it also checks if the region's leader lease is gonna expired in near future, if so it sends a message to `raftstore` and tries to renew lease

TODO: add test

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 
```release-note
raftstore: renew leader lease in advance when handle read request
```


BusyJay:
Any stats or graph to show the improvement?

slow-zhang:
what need to do if I want this merge this PR? @NingLin-P

NingLin-P:
> what need to do if I want this merge this PR? @NingLin-P

@slow-zhang  We need benchmark (a read-only workload) to verify the improvement, the key metrics to verify are the same as  https://github.com/tikv/tikv/pull/9292#issuecomment-755105740

slow-zhang:
> > what need to do if I want this merge this PR? @NingLin-P
> 
> @slow-zhang We need benchmark (a read-only workload) to verify the improvement, the key metrics to verify are the same as [#9292 (comment)](https://github.com/tikv/tikv/pull/9292#issuecomment-755105740)

Can I help here? How to set up and what kind of benchmark? Is there any relevant documentation?

NingLin-P:
> > > what need to do if I want this merge this PR? @NingLin-P
> > 
> > 
> > @slow-zhang We need benchmark (a read-only workload) to verify the improvement, the key metrics to verify are the same as [#9292 (comment)](https://github.com/tikv/tikv/pull/9292#issuecomment-755105740)
> 
> Can I help here? How to set up that kind of benchmark? Is there any relevant documentation?
> 
> BTW, there is one Jenkins job failed.

Sure! You can use [go-ycsb](https://github.com/pingcap/go-ycsb) and [workloadc](https://github.com/pingcap/go-ycsb/blob/master/workloads/workloadc), and check out this [blog](https://en.pingcap.com/blog/building-running-and-benchmarking-tikv-and-tidb#benchmarking) for guide. Also, you can set up a cluster quickly with  [tiup playgound](https://docs.pingcap.com/zh/tidb/stable/tiup-playground) (don't forget to replace the TiKV binary though

slow-zhang:
> Sure! You can use [go-ycsb](https://github.com/pingcap/go-ycsb) and [workloadc](https://github.com/pingcap/go-ycsb/blob/master/workloads/workloadc), and check out this [blog](https://en.pingcap.com/blog/building-running-and-benchmarking-tikv-and-tidb#benchmarking) for guide. Also, you can set up a cluster quickly with [tiup playgound](https://docs.pingcap.com/zh/tidb/stable/tiup-playground) (don't forget to replace the TiKV binary though

I can not download the /release tar.gz, so can I just run `make` in that branch to create a binary?

can I just use my mac(with ssd) for that test or need I to run in ubuntu?

NingLin-P:
> I can not download the /release tar.gz, so can I just run `make` in that branch to create a binary?

You can run `make dist_release` to get a release build

> can I just use my mac(with ssd) for that test or need I to run in ubuntu?

Mac will be okay.

gotoxu:
Is this work still advancing?  @NingLin-P

hykych:
@NingLin-P may I ask why the work is paused? this should improve the read performance.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -864,7 +864,6 @@ where
             CasualMessage::RegionOverlapped => {
                 debug!("start ticking for overlapped"; "region_id" => self.region_id(), "peer_id" => self.fsm.peer_id());
                 // Maybe do some safe check first?
-                self.fsm.reset_hibernate_state(GroupState::Chaos);
```
BusyJay, line 867: Why remove this?

tests/integrations/raftstore/test_lease_read.rs
```diff
@@ -769,3 +770,46 @@ fn test_infinite_lease() {
...
+    let region = cluster.get_region(key);
+    let region_id = region.get_id();
+    let peer = new_peer(1, 1);
+    cluster.must_transfer_leader(region_id, peer.clone());
+
+    let detector = LeaseReadFilter::default();
+    cluster.add_send_filter(CloneFilterFactory(detector.clone()));
+
+    // election_timeout_ticks * base_tick_interval * 3
+    let hibernate_wait = 20 * Duration::from_millis(100) * 3;
```
BusyJay, line 804: 6 seconds is too long for a test case.

5kbpers: Changed to 3 seconds.

components/raftstore/src/store/util.rs
```diff
@@ -587,6 +610,20 @@ impl RemoteLease {
     fn renew(&self, bound: Timespec) {
         self.expired_time
             .store(timespec_to_u64(bound), AtomicOrdering::Release);
+        self.renewing.store(false, AtomicOrdering::Release);
+    }
+
+    pub fn need_renew(&self, ts: Timespec) -> bool {
+        self.inspect(Some(ts + self.advance_renew_lease)) == LeaseState::Expired
+            && self
```
BusyJay, line 618: You mean:
```
!self.renewing.swap(true, Ordering::Relaxed)
```
?

components/raftstore/src/store/worker/read.rs
```diff
@@ -257,6 +257,33 @@ impl ReadDelegate {
...
+    ) {
+        if !self
+            .leader_lease
+            .as_ref()
+            .map(|lease| lease.need_renew(ts))
+            .unwrap_or(false)
+        {
+            return;
+        }
+        // TODO: add metric
```
BusyJay, line 276: How about adding it in this PR?

components/raftstore/src/store/worker/read.rs
```diff
@@ -573,7 +600,14 @@ where
                             self.redirect(RaftCommand::new(req, cb));
                             return;
                         }
-                        self.execute(&req, &delegate.region, None, read_id)
+                        let response = self.execute(&req, &delegate.region, None, read_id);
+                        // Try renew lease in advance
+                        delegate.maybe_renew_lease_advance(
```
BusyJay, line 605: Any benchmark to show the impact?

components/raftstore/src/store/util.rs
```diff
@@ -615,15 +615,7 @@ impl RemoteLease {
...
-            && self
-                .renewing
-                .compare_exchange(
-                    false,
-                    true,
-                    AtomicOrdering::Relaxed,
-                    AtomicOrdering::Relaxed,
-                )
-                .is_ok()
+            && self.renewing.swap(true, AtomicOrdering::Relaxed)
```
NingLin-P, line 618: I think it should be `!self.renewing.swap(true, AtomicOrdering::Relaxed)`

components/raftstore/src/store/worker/read.rs
NingLin-P: What it cost is only two atomic operations, checking the `expired_time` and the `renewing` flag. The `expired_time` op can be further eliminated by loading once and reuse for both lease checking and renew lease checking, but it makes the code complicated though.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1528,6 +1532,7 @@ where
             self.ctx.raft_metrics.propose.read_index += 1;
             let read_proposal = ReadIndexRequest::noop(id, current_time);
             self.fsm.peer.push_pending_read(read_proposal, true);
+            self.fsm.has_ready = true;
```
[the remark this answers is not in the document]
NingLin-P: What do you mean by `existing implementations`? 

Also, I think the `has_ready` flag is quite a pitfall, can we replace it with some simple checkings like `self.raft_group.has_ready()` instead?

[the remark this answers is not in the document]
BusyJay: What I mean is use the existing propose method. `has_ready` is not the same as `raft_group.has_ready`. For example, we may rely it to drive snapshot state.

components/raftstore/src/store/util.rs
```diff
@@ -445,12 +446,13 @@ pub enum LeaseState {
 }
 
 impl Lease {
-    pub fn new(max_lease: Duration) -> Lease {
+    pub fn new(max_lease: Duration, advance_renew_lease: Duration) -> Lease {
         Lease {
             bound: None,
             max_lease,
```
BusyJay, line 453: Unnecessary blank line.

components/raftstore/src/store/worker/read.rs
BusyJay: Read itself is also a lightweight operation. I think point get benchmark should show the difference if any.

components/raftstore/src/store/util.rs
```diff
@@ -545,13 +547,32 @@ impl Lease {
...
         // Clone the remote.
         let remote_clone = remote.clone();
         self.remote = Some(remote);
         Some(remote_clone)
     }
+
+    /// Check if the lease will be expired in near future, if so return a
+    /// future timestamp in which the lease will be expired, if not `None`
+    /// will return
+    pub fn need_renew(&self, ts: Timespec) -> Option<Timespec> {
```
BusyJay, line 563: When will it be used?

NingLin-P: It is staled, should be removed.

5kbpers:
@BusyJay @NingLin-P Just realized we need a config item for disabling that local reader renewing the lease in advance in some tests. So I added `renew_leader_lease_advance_duration` to set the time that checks the lease ahead.

5kbpers:
sysbench read_only with 256 concurrency and 15k regions shows that there is a little improvement with this patch. There are still some regions that will go hibernate in my test.
image
image

components/raftstore/src/store/util.rs
5kbpers: It is used in `need_renew_lease_at`

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
5kbpers: `has_ready` should be true because a read request was proposed here.

The reason I don't want to propose read index with existing implementations is that the read may be returned as a follower read.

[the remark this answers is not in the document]
5kbpers: Changed to propose through `propose_raft_command`. (

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1516,18 +1518,18 @@ where
...
-            let read_proposal = ReadIndexRequest::noop(id, current_time);
-            self.fsm.peer.push_pending_read(read_proposal, true);
+        if self.fsm.peer.need_renew_lease_at(self.ctx, current_time) {
+            let mut cmd = new_read_index_request(
+                self.region_id(),
+                self.region().get_region_epoch().clone(),
+                self.fsm.peer.peer.clone(),
+            );
+            cmd.mut_header().set_read_quorum(true);
+            self.propose_raft_command(
```
BusyJay, line 1528: You can call `propose_raft_command_internal` as order doesn't matter.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
NingLin-P: Perhaps `fsm.has_ready` should be removed at all. What `fsm.has_ready` may save is the snapshot and `raft_group.has_ready` checking, both of them should be lightweigth.

BusyJay:
Any updates?

BusyJay (approved):

5kbpers (approved):

Issue #9249, opened 2020-12-14
security: not all logs being redact when log redaction is enabled


yiwu-arbug:
### What version of TiKV are you using?
4.0, master

### What operating system and CPU are you using?
N/A

### Steps to reproduce
Run TiKV with `security.redact-info-log = true`, still seeing user keys being printed in hex form in info log. example:

<img width="842" alt="WeChatWorkScreenshot_e00b32ca-0c42-407b-86a4-2511118488d9" src="https://user-images.githubusercontent.com/2606959/102050509-db16e980-3d97-11eb-8145-2afc3f3d6997.png">

### What did you expect?
all info logs should be redact. We'd better add a CI to prevent misuse of `hex::encode_upper` or `hex::encode_lower` to print keys into debug log.

### What did happened?
some info logs are not redacted

PR #9250, opened 2020-12-14
security: add log redaction check

skyzh:
### What problem does this PR solve?

Issue Number: close https://github.com/tikv/tikv/issues/9249

Problem Summary:

Fix some logs not being redact when `security.redact-info-log = true`

### What is changed and how it works?

What's Changed:

We have added a script to find all occurrence of `hex::encode_upper`. If one is found in any file, then it will complain.

If you want to print key hex in log, you should always wrap it with `log_wrappers::Value::key(key)`.

If you really want to use `hex::encode_upper`, use `log_wrappers::encode_upper` instead.

### Check List 

Tests 

- Unit test

Side effects

### Release note 

Fix some logs not being redact when `security.redact-info-log = true`

PR #9315, opened 2020-12-21
security: add log redaction check (#9250)

ti-srebot:
cherry-pick #9250 to release-5.0-rc

---

Issue Number: close https://github.com/tikv/tikv/issues/9249

Problem Summary:

Fix some logs not being redact when `security.redact-info-log = true`

What's Changed:

We have added a script to find all occurrence of `hex::encode_upper`. If one is found in any file, then it will complain.

If you want to print key hex in log, you should always wrap it with `log_wrappers::Value::key(key)`.

If you really want to use `hex::encode_upper`, use `log_wrappers::encode_upper` instead.

### Check List 

Tests 

- Unit test

Side effects

### Release note 

Fix some logs not being redact when `security.redact-info-log = true`


yiwu-arbug (approved):
LGTM

hunterlxt:
LGTM

youjiali1995 (approved):

PR #9329, opened 2020-12-22
gc : skip for flushing to the bottommost level

hicqu:

When inserting data into a new constructed cluster, SST files will be flushed from memtables to L6(the bottommost level) directly. It's unnecessary to run GC for them because all data should be fresh instead of stale.

Update `check_need_gc` function for GC in compaction filter:
For a given SST file if its min ts is larger than safe point, skip GC for it.

### Check List 

Tests 

- Unit test

### Release note 
* gc: skip SSTs flushed to the bottommost level

### DIFF
--- a/src/server/gc_worker/compaction_filter.rs
+++ b/src/server/gc_worker/compaction_filter.rs
@@ -663,7 +663,7 @@ fn check_need_gc(
     ratio_threshold: f64,
     context: &CompactionFilterContext,
 ) -> bool {
-    if ratio_threshold < 1.0 || context.is_bottommost_level() {
+    if ratio_threshold < 1.0 {
         // According to our tests, `split_ts` on keys and `parse_write` on values
         // won't utilize much CPU.
         return true;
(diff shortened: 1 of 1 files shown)


yiwu-arbug (approved):
LGTM

src/server/gc_worker/compaction_filter.rs
```diff
@@ -663,16 +663,20 @@ fn check_need_gc(
     ratio_threshold: f64,
     context: &CompactionFilterContext,
 ) -> bool {
-    if ratio_threshold < 1.0 || context.is_bottommost_level() {
-        // According to our tests, `split_ts` on keys and `parse_write` on values
-        // won't utilize much CPU.
+    if ratio_threshold < 1.0 {
```
Connor1996, line 666: still return true when props.min_ts > safe_point

Connor1996 (approved):

ti-srebot (approved):

PR #9340, opened 2020-12-23
storage: tag RocksDB background IO

tabokie:

What's Changed:

Tag RocksDB compaction and flush IO flow with event listener.

### Check List 

Tests 

- Unit test

### Release note 

No release note

### DIFF
--- a/cmd/src/server.rs
+++ b/cmd/src/server.rs
@@ -19,10 +19,14 @@ use std::{
 
 use concurrency_manager::ConcurrencyManager;
 use encryption::DataKeyManager;
-use engine_rocks::{encryption::get_env, RocksEngine};
+use engine_rocks::{
+    encryption::get_env as get_encrypted_env, file_system::get_env as get_inspected_env,
+    RocksEngine,
+};
 use engine_traits::{
     compaction_job::CompactionJobInfo, Engines, MetricsFlusher, RaftEngine, CF_DEFAULT, CF_WRITE,
 };
+use file_system::EngineFileSystemInspector;
 use fs2::FileExt;
 use futures::executor::block_on;
 use grpcio::{EnvBuilder, Environment};
@@ -355,7 +359,7 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         // We truncate a big file to make sure that both raftdb and kvdb of TiKV have enough space
         // to compaction when TiKV recover. This file is created in data_dir rather than db_path,
         // because we must not increase store size of db_path.
-        tikv_util::reserve_space_for_recover(
+        file_system::reserve_space_for_recover(
             &self.config.storage.data_dir,
             self.config.storage.reserve_space.0,
         )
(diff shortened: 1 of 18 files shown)


components/engine_rocks/src/event_listener.rs
```diff
@@ -42,6 +56,19 @@ impl rocksdb::EventListener for RocksEventListener {
...
+        }
+    }
+
+    fn on_subcompaction_begin(&self, _info: &SubcompactionJobInfo) {
+        set_io_type(IOType::Compaction);
+    }
+
+    fn on_subcompaction_completed(&self, _info: &SubcompactionJobInfo) {
+        if get_io_type() == IOType::Compaction {
+            set_io_type(IOType::Other);
```
Connor1996, line 70: `Other` is tagged for some missing tagged threads or when an error occurs for the io. I'm not sure whether it's better to tag `Other` here.

components/engine_rocks/Cargo.toml
```diff
@@ -48,8 +49,9 @@ protobuf = "2"
 fail = "0.4"
 
 [dependencies.rocksdb]
-git = "https://github.com/tikv/rust-rocksdb.git"
+git = "https://github.com/tabokie/rust-rocksdb.git"
```
Connor1996, line 52: please change it

cmd/src/server.rs
```diff
@@ -943,7 +950,10 @@ impl TiKVServer<RocksEngine> {
 
 impl TiKVServer<RaftLogEngine> {
     fn init_raw_engines(&mut self) -> Engines<RocksEngine, RaftLogEngine> {
-        let env = get_env(self.encryption_key_manager.clone(), None /*base_env*/).unwrap();
+        let env =
+            get_encrypted_env(self.encryption_key_manager.clone(), None /*base_env*/).unwrap();
+        let env =
+            get_inspected_env(Some(Arc::new(EngineFileSystemInspector::new())), Some(env)).unwrap();
```
Connor1996, line 956: the inspector should be passed to limiter?

components/engine_rocks/src/event_listener.rs
tabokie: The IO type after compaction is unspecified. For instance flush can be scheduled to low-priority threads used for compaction before.

components/engine_rocks/Cargo.toml
tabokie: This PR currently depends on tikv/rust-rocksdb#595

cmd/src/server.rs
tabokie: Inspector is constructed from limiter, which is globally accessible.

tabokie:
IMO we should merge this ASAP as it provides a minimal integration of trackable limiter into the system, making it possible to write unit tests without the actual implementation being merged. @yiwu-arbug @Connor1996

components/engine_rocks/src/event_listener.rs
Connor1996: I mean just keep the type, future flush or compaction will set the type at the beginning anyway.

tabokie: True. But I still don't see a strong enough reason to make the compromise. It's just *more correct* to set IO type to unknown than leaving it potentially wrong.

Connor1996: Alright, let's keep it

Connor1996 (approved):
LGTM

yiwu-arbug (commented):
LGTM overall. Left comment on the rust-rocksdb PR.

yiwu-arbug (approved):
LGTM

Issue #9346, opened 2020-12-24
*: support handling stale read request on replica


NingLin-P:
## Development Task

This is a subtask of https://github.com/pingcap/tidb/issues/21094, only for `TxnKV` (not `RawKV`)

Currently, TiKV only support reading the most up-to-date data, the leader need holding leader lease and applied in its term, and the follower need `ReadIndex`. For stale read request, these checks in unneeded, because stale read can response stale data.

What stale read request need to ensure is it must not violate the transaction isolation guarantees TiKV provided. Let's define `safe_ts` as the max timestamp that the replica can read on locally (without `ReadIndex`) and without violate the transaction isolation guarantees, `safe_ts` should be change as the replica have more data (apply more data).

So that when handle stale read request, the replica checks `safe_ts` >= `read_ts`, if it is true then the replica read data directly, if it is false the replica need response  the `DataIsNotReady` error to indicate it doesn't have enough data to handle this request.

PR #9347, opened 2020-12-24
*: support handling stale read request on replica

NingLin-P:

What's Changed:

This PR is port of https://github.com/pingcap/tidb/issues/21094

Implement https://github.com/tikv/tikv/issues/9346

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 
- *: support handling stale read request on replica


NingLin-P:
PTAL @BusyJay

components/cdc/src/observer.rs
```diff
@@ -191,7 +191,7 @@ impl RoleObserver for CdcObserver {
                 let deregister = Deregister::Region {
                     region_id,
                     observe_id,
-                    err: CdcError::Request(store_err.into()),
+                    err: CdcError::request(store_err.into()),
```
BusyJay, line 194: Can the changes to error be extracted as a separate PR?

NingLin-P: We need these changes because clippy complain about `large_enum_variant` after we add `DataIsNotReady` to `kvproto::errorpb::Error` which is a variant of `CdcError`

components/raftstore/src/store/peer.rs
```diff
@@ -3137,6 +3141,32 @@ where
                 };
             }
         }
+        let flags = WriteBatchFlags::from_bits_truncate(req.get_header().get_flags());
+        if flags.contains(WriteBatchFlags::STALE_READ) {
+            let read_ts =
+                decode_var_u64(&mut req.mut_header().take_flag_data().as_slice()).unwrap();
+            let safe_ts = self.safe_ts.load(Ordering::Relaxed);
```
BusyJay, line 3148: These should be put into coprocessor.

NingLin-P: I will extract all of them after this PR and https://github.com/tikv/tikv/pull/9385 are merged

components/raftstore/src/store/peer.rs
```diff
@@ -3137,6 +3141,32 @@ where
                 };
             }
         }
+        let flags = WriteBatchFlags::from_bits_truncate(req.get_header().get_flags());
+        if flags.contains(WriteBatchFlags::STALE_READ) {
+            let read_ts =
+                decode_var_u64(&mut req.mut_header().take_flag_data().as_slice()).unwrap();
```
BusyJay, line 3147: Probably using fixed u64 is better. `ts` is usually too large to be encoded as varint efficiently.

components/raftstore/src/store/peer.rs
```diff
@@ -3575,6 +3606,11 @@ pub trait RequestInspector {
             return Ok(RequestPolicy::ProposeNormal);
         }
 
+        let flags = WriteBatchFlags::from_bits_truncate(req.get_header().get_flags());
```
BusyJay, line 3609: Probably should panic on unrecognized flags.

components/raftstore/src/store/peer.rs
BusyJay: Then there should be a todo.

components/tikv_util/src/lib.rs
```diff
@@ -537,6 +538,18 @@ pub fn empty_shared_slice<T>() -> Arc<[T]> {
     Vec::new().into()
 }
 
+bitflags! {
+    /// Additional flags for a write batch.
+    /// They should be set in the `flags` field in `RaftRequestHeader`.
+    pub struct WriteBatchFlags: u64 {
```
BusyJay, line 544: `txn_types` seems a more appropriate place.

components/raftstore/src/store/peer.rs
NingLin-P: I think this could break forward compatibility.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3248,7 +3249,13 @@ where
             }
         }
         let allow_replica_read = read_only && msg.get_header().get_replica_read();
-        if !(self.fsm.peer.is_leader() || is_read_index_request || allow_replica_read) {
+        let flags = WriteBatchFlags::from_bits_truncate(msg.get_header().get_flags());
+        let allow_stale_read = read_only && flags.contains(WriteBatchFlags::STALE_READ);
+        if !(self.fsm.peer.is_leader()
```
BusyJay, line 3254: It's hard to read, how about `!is_leader && !is_read_index_request && !allow_replica_read && !allow_stale_read`?

components/raftstore/src/store/peer.rs
BusyJay: It's on purpose. New features should be enabled when all binaries are upgraded. Otherwise different instances can behave differently.

NingLin-P: The flags of read request are set by tikv itself, not from the network, but I add the check for robustness though.

components/raftstore/src/store/peer.rs
```diff
@@ -3205,6 +3211,31 @@ where
                 };
             }
         }
+        let flags = WriteBatchFlags::from_bits_check(req.get_header().get_flags());
+        if flags.contains(WriteBatchFlags::STALE_READ) {
+            let read_ts = decode_u64(&mut req.mut_header().take_flag_data().as_slice()).unwrap();
+            let safe_ts = self.safe_ts.load(Ordering::Relaxed);
```
BusyJay, line 3217: Why use `Relaxed`?

components/txn_types/Cargo.toml
```diff
@@ -31,3 +32,7 @@ tikv_alloc = { path = "../tikv_alloc" }
 tikv_util = { path = "../tikv_util", default-features = false }
 collections = { path = "../collections" }
 log_wrappers = { path = "../log_wrappers" }
+
+[dev-dependencies]
+bitflags = "1.0"
+rand = "0.7"
```
BusyJay, line 38: Why are they necessary?

NingLin-P: `bitflags` is a mistake, `rand` is used for test

components/raftstore/src/store/peer.rs
NingLin-P: `safe_ts` just an incrementing counter for now, and only needs an atomic read.

BusyJay: Why? I think it stands for "some data are written before this ts". Using `Relaxed` no ordering is guaranteed.

NingLin-P:
There are some changes related to the replica read compatibility of async commit, PTAL @sticnarf

components/raftstore/src/store/worker/read.rs
```diff
@@ -507,36 +513,68 @@ where
...
+                        }
                     }
-                    response.txn_extra_op = delegate.txn_extra_op.load();
-                    cb.invoke_read(response);
-                } else {
-                    // Forward to raftstore.
-                    self.redirect(RaftCommand::new(req, cb));
+                    _ => unreachable!(),
+                }
+                let mut response = self.execute(&req, &delegate.region, None, read_id);
```
BusyJay, line 571: What if follower is deleted right before this line?

components/txn_types/src/types.rs
```diff
@@ -408,10 +409,58 @@ pub trait TxnExtraScheduler: Send + Sync {
...
+            WriteBatchFlags::ONE_PC
+        );
+        assert_eq!(
+            WriteBatchFlags::from_bits_check(WriteBatchFlags::STALE_READ.bits()),
+            WriteBatchFlags::STALE_READ
+        );
+    }
+
+    #[test]
+    #[should_panic]
```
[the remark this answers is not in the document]
NingLin-P: It won't print backtrace if the test is passed

src/storage/mod.rs
```diff
@@ -673,24 +677,19 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
                         .observe(begin_instant.elapsed().as_secs_f64());
                 }
 
-                let snap_ctx = if need_check_locks_in_replica_read(&ctx) {
+                let mut snap_ctx = SnapContext {
+                    pb_ctx: &ctx,
+                    start_ts,
```
BusyJay, line 682: @sticnarf PTAL

src/coprocessor/endpoint.rs
```diff
@@ -364,11 +364,11 @@ impl<E: Engine> Endpoint<E> {
     ) -> impl std::future::Future<Output = Result<E::Snap>> {
         let mut snap_ctx = SnapContext {
             pb_ctx: &ctx.context,
+            start_ts: ctx.txn_start_ts,
```
BusyJay, line 367: @sticnarf PTAL if it's OK.

components/txn_types/src/types.rs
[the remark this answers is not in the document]
BusyJay: It will. You can grep "panicked" in the [case logs](https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/tikv_ghpr_test/runs/47208/log/?start=0) from jenkins. Verbose stacktrace is the exact reason why `recover_safe` is used in every cases that expect to panic.

sticnarf:
This PR reuses the `start_ts` field in `SnapContext`, right? The async commit related part looks good to me.

components/raftstore/src/store/worker/read.rs
```diff
@@ -263,6 +268,35 @@ impl ReadDelegate {
 
         false
     }
+
+    fn is_safe_to_read<S: Snapshot>(
```
BusyJay, line 272: It feels odd and unclear to return `Option` for an `is_xxx` method. How about naming it `check_stale_read_safe` and return `Result<(), ReadResponse<S>>`?

components/raftstore/src/store/peer.rs
```diff
@@ -3205,6 +3211,31 @@ where
                 };
             }
         }
+        let flags = WriteBatchFlags::from_bits_check(req.get_header().get_flags());
+        if flags.contains(WriteBatchFlags::STALE_READ) {
+            let read_ts = decode_u64(&mut req.mut_header().take_flag_data().as_slice()).unwrap();
```
BusyJay, line 3216: I think you can write it as `&mut req.get_header().get_flag_data()`.

BusyJay: Why mark as resolved?

Cargo.lock
```diff
@@ -2107,7 +2107,7 @@ dependencies = [
 [[package]]
 name = "kvproto"
 version = "0.0.2"
-source = "git+https://github.com/pingcap/kvproto.git#6ed99a08e262d8a32d6355dcba91cf99cb92074a"
+source = "git+https://github.com/NingLin-P/kvproto.git?branch=stale-read#78ca596a8112b4f78653c73f8db302f17a21bced"
```
BusyJay, line 2110: Should use master version.

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/peer.rs
NingLin-P: Oh, there is one more place, it is changed now.

BusyJay (approved):

NingLin-P:
The follower may not able to handle stale read requests while applying snapshot, I will leave a TODO for now, and handle it at the next PR. /cc @BusyJay

components/raftstore/src/store/peer.rs
```diff
@@ -3203,6 +3209,31 @@ where
...
+        if flags.contains(WriteBatchFlags::STALE_READ) {
+            let read_ts = decode_u64(&mut req.get_header().get_flag_data()).unwrap();
+            let safe_ts = self.safe_ts.load(Ordering::Acquire);
+            if safe_ts < read_ts {
+                debug!(
+                    "read rejected by safe timestamp";
+                    "safe ts" => safe_ts,
+                    "read ts" => read_ts,
+                    "tag" => &self.tag
+                );
```
nolouch, line 3222: need metrics?

nolouch (commented):
rest LGTM

nolouch (approved):
lgtm

PR #9349, opened 2020-12-24
*: update raft

BusyJay:

What's Changed:
Latest raft contains fix for liveness problem in joint consensus.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No release note.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3129,7 +3129,7 @@ dependencies = [
 [[package]]
 name = "raft"
 version = "0.6.0-alpha"
-source = "git+https://github.com/tikv/raft-rs#34829a8152c845ab642a8094a9ad5887dc64c85e"
+source = "git+https://github.com/tikv/raft-rs#c7c230f63c59d408176e9288966da00c579ef4ba"
 dependencies = [
  "fxhash",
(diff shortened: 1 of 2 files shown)


NingLin-P (approved):

gengliqi (approved):

ti-srebot (approved):

BusyJay:
CI fails because raft receives a second snapshot before the first one is finished. So `on_persist_ready` of the first snapshot will not update `persisted` because term check fails due to no log. And following advance_apply will panic because `persisted` is too small.

One way to fix it is when `advance_apply` is provided an index less than first index, skip checking `persisted`.

BusyJay:
I get around the problem in this PR by informing raft the snapshot is persisted in `handle_raft_ready_advance` by using `on_persist_ready`. raft-rs is still needed to be adjusted for async ready.

PTAL @NingLin-P @gengliqi

gengliqi (approved):

NingLin-P (approved):

PR #9350, opened 2020-12-24
server: register consistency checker correctly

hicqu:

Currently there is a bug that consistency checker can't be registered correctly, so that users can't enable consistency check at all.

In the PR we register consistency checker before starting batch system, so that it can work as expected.

### Check List 

Tests 
  1. Start a tikv server with configuration `raftstore.consistency-check-interval = '5s'`
  2. Wait 5 seconds, and then check the TiKV's log
  3. Grep "consistency check pass", if we can see something it says the patch works

Side effects

### Release note 
* server: register consistency checker correctly

### DIFF
--- a/cmd/src/server.rs
+++ b/cmd/src/server.rs
@@ -14,7 +14,7 @@ use std::{
     fs::{self, File},
     net::SocketAddr,
     path::{Path, PathBuf},
-    sync::{Arc, Mutex},
+    sync::{atomic::AtomicU64, Arc, Mutex},
 };
 
 use concurrency_manager::ConcurrencyManager;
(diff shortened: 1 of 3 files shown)


hunterlxt:
Can you post the related issue here to describe how the bug happens?

hicqu:
> how the bug happens
In expected users can enable TiKV's consistency check with such configuration
```
raftstore.consistency-check-interval = "1m"
```
However it can't work exactly because of this bug, so that no consistency check will be run at all.

hunterlxt (approved):
LGTM

NingLin-P (approved):

ti-srebot (approved):

PR #9366, opened 2020-12-25
[DNM] storage&txn: Add debug logs for async commit/1pc

MyonKeminta:

Adds logs for debugging async commit/1pc based on version 3bae18b0d23723de15deed3bd1bba990cf827910

Adds some logs.

### Release note 

- No release note.


sticnarf:
Maybe we need to track when the memory locks are released from the lock table. We can add logs to the drop method of KeyHandleGuard:

```
impl Drop for KeyHandleGuard {
    fn drop(&mut self) {
        let l = self.handle.lock_store.lock();
        info!("drop KeyHandleGuard"; "lock" => ?l);
        *l = None;
    }
}
```

PR #9375, opened 2020-12-28
cdc: fix resolved ts stoped after term changed

5kbpers:

Problem Summary:

leader info was not updated after term changed.

Proposal: [xxx](url) 

What's Changed:
update leader info after term changed.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A

### DIFF
--- a/components/cdc/tests/integrations/test_cdc.rs
+++ b/components/cdc/tests/integrations/test_cdc.rs
@@ -1343,9 +1343,12 @@ fn test_term_change() {
     suite
         .cluster
         .must_transfer_leader(region.id, new_peer(2, 2));
+    // Simulate network partition.
     let recv_filter =
         Box::new(RegionPacketFilter::new(region.get_id(), 1).direction(Direction::Recv));
     suite.cluster.sim.wl().add_recv_filter(1, recv_filter);
+    // Transfer leader to peer 3 and then change it back to peer 2.
+    // Peer 1 would not get a new SoftState.
     suite
         .cluster
         .must_transfer_leader(region.id, new_peer(3, 3));
--- a/components/cdc/tests/integrations/test_cdc.rs
+++ b/components/cdc/tests/integrations/test_cdc.rs
@@ -1334,6 +1334,10 @@ fn test_cdc_filtering_gc_fence() {
     }
 }
 
+// This test is created for covering the case that term was increased without leader change.
+// Ideally leader id and term in StoreMeta should be updated together with a yielded SoftState,
+// but sometimes the leader was transferred to another store and then changed back,
+// a follower would not get a new SoftState.
 #[test]
 fn test_term_change() {
     let cluster = new_server_cluster(0, 3);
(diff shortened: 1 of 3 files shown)


components/raftstore/src/store/peer.rs
```diff
@@ -1378,17 +1387,27 @@ where
...
-        leader_id: u64,
-        term: u64,
+        leader_id: Option<u64>,
+        term: Option<u64>,
     ) {
+        let mut meta = ctx.store_meta.lock().unwrap();
+        match meta.leaders.entry(self.region_id) {
+            HashEntry::Occupied(mut e) => {
+                let term = term.unwrap_or_else(|| e.get().0);
+                let leader_id = leader_id.unwrap_or_else(|| e.get().1);
```
BusyJay, line 1397: I don't think it's correct. term and leader id can't be mixed arbitrary.

5kbpers: It just means (term, leader id) could be updated partly.

BusyJay: That's exactly what I mean. For example, (4, 5) can't be updated to (5, 5). Because 5 is leader at term 4 but may not be leader at term 5.

5kbpers: Ideally, leader id and term should be updated together with a yielded `SoftState`, but sometimes the leader was transferred to another store and then changed back, a follower would not get a new `SoftState`. In this situation, we can only update the term.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2353,6 +2353,7 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
         let regions = leaders
             .into_iter()
             .map(|leader_info| {
+                let store_id = meta.store_id.unwrap();
```
BusyJay, line 2356: Why not `self.fsm.store.id`?

components/raftstore/src/store/peer.rs
BusyJay: Then follower will also record the leader id in soft state, you can query it by `self.leader_id()`.

components/raftstore/src/store/peer.rs
```diff
@@ -1312,11 +1313,19 @@ where
                 }
                 _ => {}
             }
-            self.on_leader_changed(ctx, ss.leader_id, self.term());
+            self.on_leader_changed(ctx, Some(ss.leader_id), Some(self.term()));
             ctx.coprocessor_host
                 .on_role_change(self.region(), ss.raft_state);
             self.cmd_epoch_checker.maybe_update_term(self.term());
         }
+        if ready.must_sync() {
```
BusyJay, line 1321: I think you need an else to avoid calling `on_leader_changed` twice.

BusyJay (commented):
LGTM

components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -1333,3 +1333,53 @@ fn test_cdc_filtering_gc_fence() {
         }
     }
 }
+
+#[test]
+fn test_term_change() {
```
overvenus, line 1338: Please add some comment about the test, thanks!

overvenus (commented):
LGTM

ti-srebot (approved):

PR #9377, opened 2020-12-28
Test/fix mac

### LINKED ISSUE #9411: Security check doesn't work in MacOS

master

MacOS

### Steps to reproduce

```shell
make dev
```

### What did happened?

```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
Security check passed.
```

Clearly the grep of the security check does not work in MacOS. And it correctly reports security check pass incorrectly as well.

gregwebs:
Fix Mac warnings.

Resolves #9411

### Release note 

None

### DIFF
--- a/components/test_util/src/lib.rs
+++ b/components/test_util/src/lib.rs
@@ -28,10 +28,6 @@ pub use crate::security::*;
 
 pub fn setup_for_ci() {
     if env::var("CI").is_ok() {
-        if env::var("LOG_FILE").is_ok() {
-            logging::init_log_for_test();
-        }
-
         // HACK! Use `epollex` as the polling engine for gRPC when running CI tests on
         // Linux and it hasn't been set before.
         // See more: https://github.com/grpc/grpc/blob/v1.17.2/src/core/lib/iomgr/ev_posix.cc#L124
@@ -41,6 +37,10 @@ pub fn setup_for_ci() {
             if env::var("GRPC_POLL_STRATEGY").is_err() {
                 env::set_var("GRPC_POLL_STRATEGY", "epollex");
             }
+        }
+
+        if env::var("LOG_FILE").is_ok() {
+            logging::init_log_for_test();
         }
     }
 
--- a/scripts/check-redact-log
+++ b/scripts/check-redact-log
@@ -1,9 +1,21 @@
 #!/usr/bin/env bash
 # This script checks if there is plain (not redacted) log in TiKV.
+set -euo pipefail
 
-if grep -r -n -P '(?<!hex_)encode_upper' -C 1 --color=always --include \*.rs --exclude hex.rs --exclude-dir target . ; then
-    echo "To print user data into info logs or error messages, use log_wrappers::Value() instead of hex::encode_upper. The former will respect \`security.redact-info-log\` config and filter out user data from info log if needed. Otherwise, use \`log_wrappers::hex_encode_upper\` to get around the lint error. See https://github.com/tikv/tikv/pull/9250 for more information." 
-    exit 1
+function error_msg() {
+        echo "To print user data into info logs or error messages, use log_wrappers::Value() instead of hex::encode_upper. The former will respect \`security.redact-info-log\` config and filter out user data from info log if needed. Otherwise, use \`log_wrappers::hex_encode_upper\` to get around the lint error. See https://github.com/tikv/tikv/pull/9250 for more information." >&2
+}
+
+if [[ "$(uname)" == "Darwin" ]] ; then
+   if grep -r -n --color=always --include '*.rs' --exclude hex.rs --exclude-dir target 'encode_upper' . | grep -v log_wrappers ; then
+	error_msg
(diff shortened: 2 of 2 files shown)


components/test_util/src/lib.rs
```diff
@@ -28,10 +28,6 @@ pub use crate::security::*;
 
 pub fn setup_for_ci() {
     if env::var("CI").is_ok() {
-        if env::var("LOG_FILE").is_ok() {
```
gregwebs, line 31: Clippy gave a collapsible if warning here previously.

scripts/check-redact-log
```diff
@@ -1,9 +1,21 @@
...
+        echo "To print user data into info logs or error messages, use log_wrappers::Value() instead of hex::encode_upper. The former will respect \`security.redact-info-log\` config and filter out user data from info log if needed. Otherwise, use \`log_wrappers::hex_encode_upper\` to get around the lint error. See https://github.com/tikv/tikv/pull/9250 for more information." >&2
+}
+
+if command -v uname >/dev/null && [[ "$(uname)" == "Darwin" ]] ; then
+   if grep -r -n --color=always --include '*.rs' --exclude hex.rs --exclude-dir target 'encode_upper' . | grep -v log_wrappers ; then
+	error_msg
+	exit 1
+   fi
+else
+    if grep -r -n -P '(?<!hex_)encode_upper' -C 1 --color=always --include \*.rs --exclude hex.rs --exclude-dir target . ; then
```
gregwebs, line 15: Should we look behind for l`og_wrappers` instead?

skyzh: It is possible that some SQL functions relies on `log_wrappers::hex_encode_upper` in `tidb_query_vec_expr`. Therefore, I applied look behind regex for all files instead of only for `log_wrappers`.

skyzh (approved):

scripts/check-redact-log
skyzh: I think it's okay to look behind for `log_wrappers` instead of `hex_`.

scripts/check-redact-log
```diff
@@ -1,9 +1,21 @@
...
+set -euo pipefail
 
-if grep -r -n -P '(?<!hex_)encode_upper' -C 1 --color=always --include \*.rs --exclude hex.rs --exclude-dir target . ; then
-    echo "To print user data into info logs or error messages, use log_wrappers::Value() instead of hex::encode_upper. The former will respect \`security.redact-info-log\` config and filter out user data from info log if needed. Otherwise, use \`log_wrappers::hex_encode_upper\` to get around the lint error. See https://github.com/tikv/tikv/pull/9250 for more information." 
-    exit 1
+function error_msg() {
+        echo "To print user data into info logs or error messages, use log_wrappers::Value() instead of hex::encode_upper. The former will respect \`security.redact-info-log\` config and filter out user data from info log if needed. Otherwise, use \`log_wrappers::hex_encode_upper\` to get around the lint error. See https://github.com/tikv/tikv/pull/9250 for more information." >&2
+}
+
+if command -v uname >/dev/null && [[ "$(uname)" == "Darwin" ]] ; then
```
BusyJay, line 9: What's the purpose of `command -v uname`?

gregwebs: I am used to checking for the existence of an executable. Probably not needed in this case so I am happy to remove it.

BusyJay: Please keep it simple.

scripts/check-redact-log
```diff
@@ -1,9 +1,21 @@
...
+   if grep -r -n --color=always --include '*.rs' --exclude hex.rs --exclude-dir target 'encode_upper' . | grep -v log_wrappers ; then
```
BusyJay, line 10: Probably just using the same command is enough for all platforms.

gregwebs:
@BusyJay PTAL

scripts/check-redact-log
BusyJay: Why is it marked resolved? I can see different commands are still used for different platforms.

skyzh: On macOS, `grep` doesn't support look back pattern (`{except hex_}encode_upper`). If we use grep once and then `grep -v`, the output will lose context information `-C`, which shows adjacent lines around `encode_upper` and is friendly to developers. Therefore, to retain context information correctly, we need to use one grep with look back pattern, and have to use two separate commands for different platforms. The result is that, on macOS, no context information is shown, and on Linux, adjacent lines will be shown.

BusyJay: Get it, thanks!

BusyJay (approved):

yiwu-arbug (approved):
LGTM

PR #9385, opened 2020-12-29
raftstore: forward the replica's safe timestamp

### LINKED ISSUE #9384: raftstore: forward the replica's safe timestamp
## Development Task

This is a subtask of https://github.com/pingcap/tidb/issues/21094, only for `TxnKV` (not `RawKV`)

To support stale read, we need `safe_ts` to check whether a replica can handle a read request safely and to let a replica be able to handle as many requests as possible, we should keep its `safe_ts` as new as possible (close to `TSO.now`).

The replica can forward its `saft_ts` by calculating its own `saft_ts` according to the data it has (it has applied). But when the region is inactive, the replica's `safe_ts` can't be forwarded because there are no new data coming.  To overcome this, the leader replica needs to forward its own `safe_ts` (with a new timestamp from `PD`) and synchronize `safe_ts` to other replicas.

There are two periodical tasks that can synchronize `safe_ts` from leader replica to other replicas:
- `CheckLeader` RPC
- raft heartbeat

Also, the leader needs to send `(safe_ts, applied_index)` to other replicas, so that only when the replica's `apply_index` is larger or equal to `applied_index` it can use that `safe_ts`.

NingLin-P:

Issue Number: close #9384 

Forward the replica's `safe_ts` by `CheckLeader` RPC and raft heartbeat.

TODO: add test

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- No release note


components/raftstore/src/store/fsm/store.rs
```diff
@@ -2393,6 +2413,146 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
...
+                    .unwrap()
+                    .safe_ts
+                    .load(Ordering::Relaxed)
+            })
+            .min()
+            .unwrap_or(0)
+    }
+}
+
+/// `RegionReadProgress` is used to keep track of the replica's `safe_ts`, the replica can handle a read
```
BusyJay, line 2454: Can you move it out of store.rs? It's more than 2500 lines.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2393,6 +2413,41 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
             .collect();
         cb(regions);
     }
+
+    fn on_get_store_safe_ts(&self, key_range: KeyRange, cb: Box<dyn FnOnce(u64) + Send>) {
+        let meta = self.ctx.store_meta.lock().unwrap();
```
BusyJay, line 2418: If there are a lot of requests, this can become bottle neck easily.

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,153 @@ impl Display for MsgType<'_> {
     }
 }
 
+/// `RegionReadProgress` is used to keep track of the replica's `safe_ts`, the replica can handle a read
+/// request directly without requiring leader lease or read index iff `safe_ts` >= `read_ts` (the `read_ts`
+/// is usually stale i.e seconds ago).
+///
+/// `safe_ts` is forward by the `(apply index, safe ts)` item:
```
BusyJay, line 844: This doesn't seem like a common usage. Do you have any recommendations? /cc @WT-Liu

components/raftstore/src/store/fsm/store.rs
NingLin-P: It won't be accessed per-request or per-transaction, instead, TiDB will get the store safe ts periodically in the interval of seconds.

BusyJay: Can you add a comment explaining the context?

NingLin-P: Sure!

components/raftstore/src/store/util.rs
WT-Liu: In this context, "update" might be more appropriate.

components/raftstore/src/store/peer.rs
```diff
@@ -1797,6 +1805,9 @@ where
             let mut meta = ctx.store_meta.lock().unwrap();
             meta.readers
                 .insert(self.region_id, ReadDelegate::from_peer(self));
+            if let Some(rp) = meta.region_read_progress.get(&self.region_id) {
+                rp.update_applied(self.get_store().applied_index());
```
BusyJay, line 1809: I don't think it's correct to update apply index here as data has not been written to disk yet.

components/raftstore/src/store/peer.rs
```diff
@@ -3464,6 +3477,18 @@ where
             send_msg.set_start_key(region.get_start_key().to_vec());
             send_msg.set_end_key(region.get_end_key().to_vec());
         }
+
+        if self.is_leader()
+            && msg.get_msg_type() == MessageType::MsgHeartbeat
+            && self.has_applied_to_current_term()
```
BusyJay, line 3483: Is there a way to send the payload only when it's necessary?

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,153 @@ impl Display for MsgType<'_> {
...
+            safe_ts: AtomicU64::from(0),
+        }
+    }
+
+    pub fn update_applied(&self, applied: u64) {
+        let mut core = self.core.lock().unwrap();
+        if let Some(ts) = core.update_applied(applied) {
+            self.safe_ts.fetch_max(ts, AtomicOrdering::SeqCst);
+        }
+        self.applied_index
```
BusyJay, line 881: Perhaps should panic if apply index is decreased.

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,153 @@ impl Display for MsgType<'_> {
...
+        self.safe_ts.load(AtomicOrdering::Acquire)
+    }
+}
+
+#[derive(Debug)]
+struct RegionReadProgressCore {
+    applied_index: u64,
+    safe_ts: u64,
+    // `pending_items` is a *sorted* list of `(apply index, safe ts)` item
+    pending_items: VecDeque<(u64, u64)>,
```
BusyJay, line 909: How long can the queue be? Is it possible not to store the queue at all?

NingLin-P: Currently, the capacity is `128`. 

I don't have a better idea, because the leader only sends the latest `(apply index, safe ts)` item, if the follower's apply index always lags behind the leader, without the queue the follower may never be able to update its safe ts. 

BTW, we only need one item instead of a queue if using `appiled ts` as `safe ts`, since the `appiled ts` can be updated by simply applying lagging logs.

components/raftstore/src/store/peer.rs
NingLin-P: Actually, the `CheckLeader` rpc also have such payload to sync the `safe ts` to follower, and the payload here just an optimization to sync `safe ts` more frequently (since there will be no raft heartbeat if the region is hibernated), and I think the overhead here is small.

BusyJay: Hibernate is optional before, this makes it a must. Have you check how many extra bytes will be added to the message comparing a normal heartbeat?

NingLin-P: Do you mean the payload at the `CheckLeader` rpc is unnecessary?

By setting `applied_index` to `1234` and `safe_ts` to `u64::MAX`, there are 16 extra bytes :
```
Cost of read state, current size: 48, prev size: 32
```

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,155 @@ impl Display for MsgType<'_> {
...
+            core: Mutex::new(RegionReadProgressCore::new(applied_index, cap)),
+            applied_index: AtomicU64::from(applied_index),
+            safe_ts: AtomicU64::from(0),
+        }
+    }
+
+    pub fn update_applied(&self, applied: u64) {
+        let mut core = self.core.lock().unwrap();
+        if let Some(ts) = core.update_applied(applied) {
+            self.safe_ts.fetch_max(ts, AtomicOrdering::SeqCst);
```
BusyJay, line 878: Is it possible to go backward?

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,155 @@ impl Display for MsgType<'_> {
...
+///     self.safe_ts = max(self.safe_ts, item.safe_ts)
+/// }
+/// ```
+///
+/// For the leader, the `(apply index, safe ts)` item is publish by the `resolved-ts` worker periodically.
+/// For the followers, the item is sync periodically from the leader through the `CheckLeader` rpc.
+///
+/// The intend is to make the item's `safe ts` larger (more up to date) and `apply index` smaller (require less data)
+#[derive(Debug)]
+pub struct RegionReadProgress {
```
BusyJay, line 856: If `safe_ts` and `applied_index` are wrapped into a struct, then this can be defined without lock.

```
pub struct StaleReadState {
    safe_ts: AtomicU64,
    applied_index: AtomicU64,
}

pub struct RegionReadProgress {
    state: Arc<StaleReadSTate>,
    applied_index: u64,
    safe_ts: u64,
    pending_items: VecDeque<(u64, u64)>,
}
```

NingLin-P: Besides `state`, other fields in `RegionReadProgress` could be updated by different threads

BusyJay: How? Isn't it only be updated by checkLeader RPC response? The response should be handled in peer, which can't be processed by multiple threads at the same time.

components/raftstore/src/store/util.rs
NingLin-P: It won't in this place, using `fetch_max` here for robustness.

BusyJay: Just `store` should be enough given that `update_applied` won't give a smaller ts.

components/raftstore/src/store/util.rs
NingLin-P: The response will be handled by the `StoreFsm` to avoid dispatch messages to all `PeerFsm` since there could a large amount of them, and it could be processed by different threads with the `PeerFsm` at the same time.

Also, currently, for the leader, `RegionReadProgress` will be updated at the `resolved ts` worker directly, we can let the worker send message to `raftstore` doing update instead, but it just bring the same problem.

NingLin-P:
The follower may not able to handle stale read requests while applying snapshot, I will leave a TODO for now, and handle it at the next PR. /cc @BusyJay

NingLin-P:
PTAL @BusyJay

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,163 @@ impl Display for MsgType<'_> {
...
+
+    fn clear(&mut self) {
+        self.pending_items.clear();
+        self.safe_ts = 0;
+    }
+
+    fn push_back(&mut self, item: (u64, u64)) {
+        if self.pending_items.len() >= self.cap {
+            // Randomly remove a item to make room for the incoming one
+            let idx = rand::random::<usize>() % self.cap;
```
BusyJay, line 990: How about 
```
let mut keep = false;
self.pending_items.retain(|_| {
    keep = !keep;
    keep
});
```
?

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +837,163 @@ impl Display for MsgType<'_> {
...
+        None
+    }
+
+    fn clear(&mut self) {
+        self.pending_items.clear();
+        self.safe_ts = 0;
+    }
+
+    fn push_back(&mut self, item: (u64, u64)) {
+        if self.pending_items.len() >= self.cap {
```
BusyJay, line 988: Why not use `self.pending_items.capacity()`?

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/util.rs
NingLin-P: We only want to remove one item

BusyJay: It doesn't matter. Given the capacity is 128, even remove half, there are still 64 safe ts. When the queue is full, it usually means the follower is stuck, so it probably keeps popping for sometime. Shrink it half instead of poping one can reduce the overhead of moving data around again and again.

BusyJay (approved):

NingLin-P:
PTAL @nolouch @JmPotato

components/cdc/src/endpoint.rs
```diff
@@ -987,11 +987,17 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Endpoint<T> {
                             if peer.get_role() == PeerRole::Learner {
                                 continue;
                             }
+                            let mut read_state = ReadState::default();
+                            if let Some(rp) = meta.region_read_progress.get(&region_id) {
```
nolouch, line 991: Need do it in reslove ts?

nolouch (approved):
lgtm

PR #9387, opened 2020-12-29
ci: stabilize case test_node_merge_crash_when_snapshot

hicqu:

The test case test_node_merge_crash_when_snapshot is not stable.

Here is a failure result: https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/tikv_ghpr_test/runs/39212/nodes/98/steps/537/log/?start=0 . 

It shows that it's possible to meet `proposal dropped` in `Cluster::put_cf`.

### Check List 

Tests 

- Unit test

Side effects

### Release note 
* No release note.

### DIFF
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -772,6 +772,16 @@ impl<T: Simulator> Cluster<T> {
                 sleep_ms(100);
                 continue;
             }
+            if resp
+                .get_header()
+                .get_error()
+                .get_message()
+                .contains("proposal dropped")
+            {
+                warn!("seems transferring leadership, let's retry");
+                sleep_ms(100);
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


components/test_raftstore/src/cluster.rs
```diff
@@ -772,6 +772,16 @@ impl<T: Simulator> Cluster<T> {
                 sleep_ms(100);
                 continue;
             }
+            if resp
+                .get_header()
+                .get_error()
+                .get_message()
+                .contains("proposal dropped")
```
BusyJay, line 779: Have you tried to verify if it's really fixing the issue? I can't grep the string in code, which probably means the fix is a noop.

hicqu: It's in `raft-rs`.

PR #9389, opened 2020-12-29
raftstore: match ready by batch_offset

### LINKED ISSUE #9388: ready can be advanced by a wrong peer
## Bug Report

Supposing peer (1, 1) is split into (1, 2) and (1, 1) on store 1. If it takes too long for peer 1 to apply snapshot and then split, an uninitialized peer 2 can be created first and then the split one will replace then uninitialized one.

If before peer 1 is split, peer 2 receives a message, then the uninitialized peer and split peer may be fetched in one batch. Because batch system maps ready to peer by ID, so if the uninitialized peer doesn't generate readiness, then the readiness generated by split peer will be mapped to uninitialized one as they have the same region ID.

In the past, uninitialized peer has to be a follower and can't be in merge state, so all messages in readiness will be sent and no other actions will be performed. Because peer is uninitialized, so no proposal has been handled, no read states will be queued. For the split peer, although ready is not advanced, but raft will happily create a duplicated ready and all paused progresses will be resumed except committed logs. Duplicated committed logs can make apply thread panic. Fortunately, there will usually be no committed logs for a just split peer.

Now we have used async ready, unmatched ready will lead to panic directly as the numbers probably don't matched.

/cc @NingLin-P @gengliqi

BusyJay:

Issue Number: close #9388 

What's Changed:

This PR adds a new field batch_offset to ready collection, which
indicates the offset of origin peer in the batch.

Stores the ready inside Fsm is an alternative, but it can occupy
unnecessary memory.

Using memory address of Fsm is also an alternative, though seems a
little hacky.

Both of the alternatives require a loop, using offset only needs to
query once. It may have improvement in clusters that have a lot of
regions and ticking usually doesn't generate readiness.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- Fix wrong mapping between readies and peers by offset

### DIFF
--- a/tests/failpoints/cases/test_split_region.rs
+++ b/tests/failpoints/cases/test_split_region.rs
@@ -384,7 +384,7 @@ fn test_split_not_to_split_exist_tombstone_region() {
 
 /// A filter that collects all snapshots.
 ///
-/// It's different from the one in simulate_transport in two aspects:
+/// It's different from the one in simulate_transport in three aspects:
 /// 1. It will not flush the collected snapshots.
 /// 2. It will not report error when collecting snapshots.
 /// 3. It callers can access the collected snapshots.
@@ -430,13 +430,14 @@ impl Filter for CollectSnapshotFilter {
             }
         }
         msgs.extend(to_send);
+        check_messages(msgs)?;
         Ok(())
     }
 }
 
-/// If the uninitialized peer and split peer are fetched into one batch, and the front
+/// If the uninitialized peer and split peer are fetched into one batch, and the first
 /// one doesn't generate ready, the second one does, ready should not be mapped to the
-/// front one.
+/// first one.
 #[test]
 fn test_split_duplicated_batch() {
     let mut cluster = new_node_cluster(0, 3);
--- a/tests/failpoints/cases/test_split_region.rs
+++ b/tests/failpoints/cases/test_split_region.rs
@@ -1,6 +1,6 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::sync::atomic::AtomicBool;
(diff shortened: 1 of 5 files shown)


NingLin-P (approved):

tests/failpoints/cases/test_split_region.rs
```diff
@@ -380,3 +381,162 @@ fn test_split_not_to_split_exist_tombstone_region() {
...
+            if is_pending {
+                pending_msg.insert(from_peer_id, msg);
+                let sender = self.pending_count_sender.lock().unwrap();
+                sender.send(pending_msg.len()).unwrap();
+            } else {
+                to_send.push(msg);
+            }
+        }
+        msgs.extend(to_send);
+        Ok(())
```
gengliqi, line 433: How about using `check_messages(msgs)`?

tests/failpoints/cases/test_split_region.rs
```diff
@@ -380,3 +381,162 @@ fn test_split_not_to_split_exist_tombstone_region() {
     fail::remove("on_raft_gc_log_tick");
     fail::remove(peer_check_stale_state_fp);
 }
+
+/// A filter that collects all snapshots.
+///
+/// It's different from the one in simulate_transport in two aspects:
+/// 1. It will not flush the collected snapshots.
+/// 2. It will not report error when collecting snapshots.
+/// 3. It callers can access the collected snapshots.
```
gengliqi, line 390: It seems three aspects.

gengliqi (commented):
Rest LGTM

gengliqi (approved):

NingLin-P (approved):

ti-srebot (approved):

PR #9396, opened 2020-12-30
ci: allow to not truncate log file

hicqu:

When running one test case, for example:
```bash
cargo test --all test_node_merge_transfer_leader
```
Both `integrations-*` and `failpoints-` will be runned, which will open and truncate the target log file twice.

Add a new environment variable `LOG_APPEND`.
Setting it to any value will use append instead of truncate when opening log file.

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
* No release note.

### DIFF
--- a/components/test_util/src/logging.rs
+++ b/components/test_util/src/logging.rs
@@ -1,6 +1,6 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::io::prelude::*;
 use std::sync::{Mutex, Once};
 use std::{env, fmt, io};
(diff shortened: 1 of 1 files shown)


components/test_util/src/logging.rs
```diff
@@ -100,7 +100,19 @@ pub fn init_log_for_test() {
             &env::var("LOG_LEVEL").unwrap_or_else(|_| "debug".to_owned()),
         )
         .unwrap();
-        let writer = output.map(|f| Mutex::new(File::create(f).unwrap()));
+        let writer = if env::var("LOG_APPEND").is_ok() {
+            output.map(|f| {
+                Mutex::new(
+                    OpenOptions::new()
+                        .create(true)
+                        .append(true)
```
BusyJay, line 108: How about replacing `true` with `env::var("LOG_APPEND").is_ok()`?

BusyJay (approved):

Fullstop000 (approved):
It seems this is relate to the failing tests

PR #9412, opened 2020-12-31
concurrency manager: fix incorrect key drop when weak succeeds to upgrade

sticnarf:

Previously, if a reader holds an `Arc<KeyHandle>` which in the lock table and a prewrite calls `lock_key`, `lock_key` will upgrade a weak in the lock table. But it also creates a useless `KeyHandle` which drops when `lock_key` returns. When dropping `KeyHandle`, it removes the key from the table incorrectly.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

- Fix a bug that memory locks are removed incorrectly.

### DIFF
--- a/components/concurrency_manager/src/key_handle.rs
+++ b/components/concurrency_manager/src/key_handle.rs
@@ -2,7 +2,7 @@
 
 use super::lock_table::LockTable;
 use parking_lot::Mutex;
-use std::{mem, sync::Arc};
+use std::{cell::UnsafeCell, mem, sync::Arc};
 use tokio::sync::{Mutex as AsyncMutex, MutexGuard as AsyncMutexGuard};
 use txn_types::{Key, Lock};
 
@@ -10,16 +10,16 @@ use txn_types::{Key, Lock};
 /// key.
 pub struct KeyHandle {
     pub key: Key,
-    table: LockTable,
+    table: UnsafeCell<Option<LockTable>>,
     mutex: AsyncMutex<()>,
     lock_store: Mutex<Option<Lock>>,
 }
 
 impl KeyHandle {
-    pub fn new(key: Key, table: LockTable) -> Self {
+    pub fn new(key: Key) -> Self {
         KeyHandle {
             key,
-            table,
+            table: UnsafeCell::new(None),
             mutex: AsyncMutex::new(()),
             lock_store: Mutex::new(None),
         }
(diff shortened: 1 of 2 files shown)


hi-rustin:
> /auto-cc

Just testing the auto-assignment, hope I'm not bothering you.

components/concurrency_manager/src/lock_table.rs
```diff
@@ -21,13 +21,19 @@ impl Default for LockTable {
 impl LockTable {
     pub async fn lock_key(&self, key: &Key) -> KeyHandleGuard {
         loop {
-            let handle = Arc::new(KeyHandle::new(key.clone(), self.clone()));
+            let handle = Arc::new(KeyHandle::new(key.clone()));
```
cfzjywxk, line 24: We could add some comments to demonstrate that this handle object's relationship with the `LockTable` like that it may or may not be the used one.

components/concurrency_manager/src/lock_table.rs
```diff
@@ -299,4 +305,19 @@ mod test {
...
+        let guard = lock_table.lock_key(&key).await;
+        let handle = lock_table.get(&key).unwrap();
+        drop(guard);
+
+        let guard2 = lock_table.lock_key(&key).await;
+        drop(handle);
+
+        let handle = lock_table.get(&key).unwrap();
+        assert_eq!(Arc::as_ptr(&handle), Arc::as_ptr(guard2.handle()));
+    }
```
cfzjywxk, line 322: How about add some more lines dropping both the `handle` and `guard2` and try `lock_key` again, and check the pointer?

youjiali1995 (commented):
lgtm

youjiali1995 (approved):

components/concurrency_manager/src/key_handle.rs
```diff
@@ -39,14 +39,29 @@ impl KeyHandle {
...
         f(&*self.lock_store.lock())
     }
+
+    /// Set the LockTable that the KeyHandle is in.
+    ///
+    /// This method is not thread safe. Make sure that no other threads access
+    /// `table` at the same time.
+    pub(crate) unsafe fn set_table(&self, table: LockTable) {
+        *self.table.get() = Some(table);
+    }
```
MyonKeminta, line 49: Must it be unsafe? Do we have any other way?
I don't mind it much though, just being afraid that the more unsafe code, the more unsafe 🤔

cfzjywxk:
lgtm

MyonKeminta (approved):

components/concurrency_manager/src/key_handle.rs
sticnarf: I agree. I just think that code touching this unsafe is quite limiting now, so it's CURRENTLY easy to check its safety.
If it will be accessed in more places, then it's better to use a Mutex.

MyonKeminta:
Forgot to link this PR to https://github.com/pingcap/tidb/issues/21900.

Issue #8906, opened 2020-10-29
Support GCP KMS for data encryption at rest


tennix:
### Is your feature request related to a problem? Please describe:

When running TiKV on GCP, users cannot use KMS of GCP.

### Describe the feature you'd like:

Support GCP KMS for data encryption at rest.

### Describe alternatives you've considered:

Users can use AWS KMS, but it's weird that GCP users uses keys in AWS KMS.

### Teachability, Documentation, Adoption, Migration Strategy:

https://cloud.google.com/kms/docs

Issue #7236, opened 2020-03-25
Backup to S3 panics for a table ~30 GiB


kennytm:
### What version of TiKV are you using?

```
TiKV
Release Version:   4.1.0-alpha
Git Commit Hash:   50ceedd57dcc01ac1f8670a0672eabb79329ee0b
Git Commit Branch: HEAD
UTC Build Time:    2020-03-23 02:12:58
Rust Version:      rustc 1.42.0-nightly (0de96d37f 2019-12-19)
Enable Features:   jemalloc portable sse protobuf-codec
Profile:           dist_release
```

### What operating system and CPU are you using?

CentOS 7.

`Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz` × 40

### Steps to reproduce

1. Prepare an database of size ≥ 5 GiB (sysbench with 4 tables and 58 million rows each).

    ```sh
    sysbench /usr/share/sysbench/tests/include/oltp_legacy/oltp.lua 
        --mysql-host=172.16.4.95 --mysql-port=4000 --mysql-user=root 
        --oltp-test-mode=complex 
        --oltp-tables-count=10 --oltp-table-size=150000000 
        --threads=10 --time=120 --report-interval=10 
        prepare
    ```

2. Perform a BR backup to an AWS S3 storage.

### What did you expect?

Succeed

### What did happened?

<details><summary>Hyper crashed with this fatal log:</summary>

```
[2020/03/24 20:00:34.405 +08:00] [FATAL] [lib.rs:480] ["dispatch dropped without returning error"] [backtrace="stack backtrace:
   0: tikv_util::set_panic_hook::{{closure}}
             at components/tikv_util/src/lib.rs:479
   1: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:475
   2: std::panicking::begin_panic
             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/panicking.rs:404
   3: hyper::client::conn::SendRequest<B>::send_request_retryable::{{closure}}
             at /home/jenkins/agent/workspace/tikv_ghpr_build_release/tikv/<::std::macros::panic macros>:3
      <futures_util::future::future::then::Then<Fut1,Fut2,F> as core::future::future::Future>::poll::{{closure}}
             at /rust/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.4/src/future/future/then.rs:44
[...]

PR #9415, opened 2021-01-03
encryption: support multiple cloud providers

gregwebs:
Signed-off-by: Greg Weber <greg@pingcap.com>

Problem Summary:

TiKV needs to support different KMS encryption providers. For example, issue #8906 

What's Changed:

The KMS backend now has a KmsProvider which defaults to AwsKms.
It is now straightforward to write a different backend.

Some associated refactorings were made:
* Simplify KmsBackend internal state management
* Creating a backend does not happen inside Arc::new()
* Reduce code pollution of testing where `#[cfg(test)]` was used
* Add a Debug trait to backends
* Add some newtypes for safety
* Make some error messages more detailed
* Fix KMS was marked as not send but can change how it uses the APIs instead

### Related changes

None, cherry-pick to a release branch may be needed in the future.

### Check List 

Tests 

- Unit test
- Integration test (tested manually with file backend and KMS backend to see that encryption is enabled and that keys are rotated)

### Release note 

Support alternative KMS encryption backends

### DIFF
--- a/components/encryption/src/config.rs
+++ b/components/encryption/src/config.rs
@@ -49,21 +49,6 @@ pub struct FileConfig {
 #[serde(rename_all = "kebab-case")]
 pub struct KmsConfig {
     pub key_id: String,
-
-    // It is preferable to use the cloud provider SDK to automatically rotate credentials
-    // Using access_key and secret_access_key creates a security risk.
-    // These fields are required by the macro new_client! that is currently in use
-    #[doc(hidden)]
-    // We don't want to write access_key and secret_access_key to config file
-    // accidentally.
-    #[serde(skip_serializing)]
-    #[config(skip)]
-    pub access_key: String,
-    #[doc(hidden)]
-    #[serde(skip_serializing)]
-    #[config(skip)]
-    pub secret_access_key: String,
-
     pub region: String,
     pub endpoint: String,
 }
@@ -167,8 +152,6 @@ mod tests {
             master_key: MasterKeyConfig::Kms {
                 config: KmsConfig {
                     key_id: "key_id".to_owned(),
-                    access_key: "access_key".to_owned(),
-                    secret_access_key: "secret_access_key".to_owned(),
                     region: "region".to_owned(),
                     endpoint: "endpoint".to_owned(),
                 },
@@ -187,8 +170,6 @@ mod tests {
         [master-key]
         type = "kms"
         key-id = "key_id"
-        access-key = "access_key"
-        secret-access-key = "secret_access_key"
         region = "region"
         endpoint = "endpoint"
         "#;
--- a/components/encryption/src/config.rs
+++ b/components/encryption/src/config.rs
@@ -3,11 +3,6 @@
 use kvproto::encryptionpb::EncryptionMethod;
 use tikv_util::config::ReadableDuration;
 
-#[cfg(test)]
-use crate::master_key::Backend;
-#[cfg(test)]
-use std::sync::Arc;
-
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Configuration)]
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
@@ -55,10 +50,11 @@ pub struct FileConfig {
 pub struct KmsConfig {
     pub key_id: String,
 
-    // Providing access_key and secret_access_key is recommended as it has
-    // security risk.
+    // It is preferable to use the cloud provider SDK to automatically rotate credentials
+    // Using access_key and secret_access_key creates a security risk.
+    // These fields are required by the macro new_client! that is currently in use
     #[doc(hidden)]
-    // We don's want to write access_key and secret_access_key to config file
+    // We don't want to write access_key and secret_access_key to config file
     // accidentally.
     #[serde(skip_serializing)]
     #[config(skip)]
@@ -71,18 +67,6 @@ pub struct KmsConfig {
     pub region: String,
     pub endpoint: String,
 }
-
-#[cfg(test)]
-#[derive(Clone, Debug)]
-pub struct Mock(pub Arc<dyn Backend>);
-#[cfg(test)]
-impl PartialEq for Mock {
-    fn eq(&self, _: &Self) -> bool {
-        false
-    }
-}
-#[cfg(test)]
-impl Eq for Mock {}
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "kebab-case", tag = "type")]
@@ -104,10 +88,6 @@ pub enum MasterKeyConfig {
         #[serde(flatten)]
         config: KmsConfig,
     },
-
-    #[cfg(test)]
-    #[serde(skip)]
-    Mock(Mock),
 }
 
 impl Default for MasterKeyConfig {
--- a/components/encryption/src/crypter.rs
+++ b/components/encryption/src/crypter.rs
@@ -141,14 +141,14 @@ impl AesGcmTag {
 /// An Aes256-GCM crypter.
 pub struct AesGcmCrypter<'k> {
     iv: Iv,
-    key: &'k [u8],
+    key: &'k PlainKey,
 }
 
 impl<'k> AesGcmCrypter<'k> {
     /// The key length of `AesGcmCrypter` is 32 bytes.
     pub const KEY_LEN: usize = 32;
 
-    pub fn new(key: &'k [u8], iv: Iv) -> AesGcmCrypter<'k> {
+    pub fn new(key: &'k PlainKey, iv: Iv) -> AesGcmCrypter<'k> {
         AesGcmCrypter { iv, key }
     }
 
@@ -157,7 +157,7 @@ impl<'k> AesGcmCrypter<'k> {
         let mut tag = AesGcmTag([0u8; GCM_TAG_LEN]);
         let ciphertext = symm::encrypt_aead(
             cipher,
-            self.key,
+            &self.key.0,
             Some(self.iv.as_slice()),
             &[], /* AAD */
             &pt,
@@ -170,7 +170,7 @@ impl<'k> AesGcmCrypter<'k> {
         let cipher = OCipher::aes_256_gcm();
         let plaintext = symm::decrypt_aead(
             cipher,
-            self.key,
+            &self.key.0,
             Some(self.iv.as_slice()),
             &[], /* AAD */
             &ct,
@@ -195,6 +195,46 @@ pub fn verify_encryption_config(method: EncryptionMethod, key: &[u8]) -> Result<
         }
     }
     Ok(())
+}
+
+// PlainKey is a newtype used to mark a vector a plaintext key.
+// It requires the vec to be a valid AesGcmCrypter key.
+pub struct PlainKey(Vec<u8>);
+
+impl PlainKey {
+    pub fn new(key: Vec<u8>) -> Result<Self> {
+        if key.len() != AesGcmCrypter::KEY_LEN {
+            return Err(box_err!(
+                "encryption method and key length mismatch, expect {} get {}",
+                AesGcmCrypter::KEY_LEN,
+                key.len()
+            ));
+        }
+        Ok(Self(key))
+    }
+}
+
+impl std::ops::Deref for PlainKey {
+    type Target = Vec<u8>;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+// Don't expose the key in a debug print
+impl std::fmt::Debug for PlainKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("PlainKey")
+            .field(&"REDACTED".to_string())
+            .finish()
+    }
+}
+
+// Don't expose the key in a display print
+impl std::fmt::Display for PlainKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
 }
 
 #[cfg(test)]
@@ -248,7 +288,7 @@ mod tests {
 
         let pt = Vec::from_hex(pt).unwrap();
         let ct = Vec::from_hex(ct).unwrap();
-        let key = Vec::from_hex(key).unwrap();
+        let key = PlainKey::new(Vec::from_hex(key).unwrap()).unwrap();
         let iv = Iv::from_slice(Vec::from_hex(iv).unwrap().as_slice()).unwrap();
         let tag = Vec::from_hex(tag).unwrap();
 
--- a/components/encryption/src/encrypted_file/mod.rs
+++ b/components/encryption/src/encrypted_file/mod.rs
@@ -70,7 +70,7 @@ impl<'a> EncryptedFile<'a> {
             .create(true)
             .write(true)
             .open(&tmp_path)
-            .unwrap();
+            .unwrap_or_else(|_| panic!("EncryptedFile::write {}", &tmp_path.to_str().unwrap()));
 
         // Encrypt the content.
         let encrypted_content = master_key
--- a/components/encryption/src/manager/mod.rs
+++ b/components/encryption/src/manager/mod.rs
@@ -12,7 +12,10 @@ use engine_traits::{EncryptionKeyManager, FileEncryptionInfo};
 use kvproto::encryptionpb::{DataKey, EncryptionMethod, FileDictionary, FileInfo, KeyDictionary};
 use protobuf::Message;
 
-use crate::config::{EncryptionConfig, MasterKeyConfig};
+use crate::config::EncryptionConfig;
+#[cfg(test)]
+use crate::config::MasterKeyConfig;
+
 use crate::crypter::{self, compat, Iv};
 use crate::encrypted_file::EncryptedFile;
 use crate::file_dict_file::FileDictionaryFile;
@@ -169,18 +172,9 @@ impl Dicts {
         )
     }
 
-    fn get_file(&self, fname: &str) -> FileInfo {
+    fn get_file(&self, fname: &str) -> Option<FileInfo> {
         let dict = self.file_dict.lock().unwrap();
-        let file_info = dict.files.get(fname);
-        match file_info {
-            None => {
-                // Return Plaintext if file not found
-                let mut file = FileInfo::default();
-                file.method = compat(EncryptionMethod::Plaintext);
-                file
-            }
-            Some(info) => info.clone(),
-        }
+        dict.files.get(fname).cloned()
     }
 
     fn new_file(&self, fname: &str, method: EncryptionMethod) -> Result<FileInfo> {
@@ -210,6 +204,8 @@ impl Dicts {
         Ok(file)
     }
 
+    // If the file does not exist, return Ok(())
+    // In either case the intent that the file not exist is achieved.
     fn delete_file(&self, fname: &str) -> Result<()> {
         let mut file_dict_file = self.file_dict_file.lock().unwrap();
         let (file, file_num) = {
@@ -238,7 +234,7 @@ impl Dicts {
         Ok(())
     }
 
-    fn link_file(&self, src_fname: &str, dst_fname: &str) -> Result<()> {
+    fn link_file(&self, src_fname: &str, dst_fname: &str) -> Result<Option<()>> {
         let mut file_dict_file = self.file_dict_file.lock().unwrap();
         let (method, file, file_num) = {
             let mut file_dict = self.file_dict.lock().unwrap();
@@ -247,7 +243,7 @@ impl Dicts {
                 None => {
                     // Could be a plaintext file not tracked by file dictionary.
                     info!("link untracked plaintext file"; "src" => src_fname, "dst" => dst_fname);
-                    return Ok(());
+                    return Ok(None);
                 }
             };
             // When an encrypted file exists in the file system, the file_dict must have info about
@@ -267,7 +263,7 @@ impl Dicts {
         } else {
             info!("link plaintext file"; "src" => src_fname, "dst" => dst_fname);
         }
-        Ok(())
+        Ok(Some(()))
     }
 
     fn rotate_key(&self, key_id: u64, key: DataKey, master_key: &dyn Backend) -> Result<()> {
@@ -355,7 +351,7 @@ fn check_stale_file_exist(
 fn run_background_rotate_work(
     dict: Arc<Dicts>,
     method: EncryptionMethod,
-    master_key: Arc<dyn Backend>,
+    master_key: &dyn Backend,
     terminal_recv: channel::Receiver<()>,
 ) {
     let check_period = std::cmp::min(
@@ -367,7 +363,7 @@ fn run_background_rotate_work(
         select! {
             recv(tick(check_period)) -> _ => {
                 info!("Try to rotate data key, current method:{:?}", method);
-                dict.maybe_rotate_data_key(method, master_key.as_ref())
+                dict.maybe_rotate_data_key(method, master_key)
                     .expect("Rotating key operation encountered error in the background worker");
             },
             recv(terminal_recv) -> _ => {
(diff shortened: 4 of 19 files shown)


components/external_storage/src/s3.rs
```diff
@@ -27,22 +27,15 @@ use kvproto::backup::S3 as Config;
...
-        Self::check_config(config)?;
-        let client = new_client!(S3Client, config);
-        Ok(S3Storage {
-            config: config.clone(),
-            client,
-            _not_send: PhantomData::default(),
-        })
+        // Need to explicitly create a dispatcher
+        // See https://github.com/tikv/tikv/issues/7236.
+        let dispatcher = HttpClient::new().unwrap();
```
yiwu-arbug, line 37: return error instead of panic.

components/encryption/src/master_key/kms.rs
```diff
@@ -16,30 +15,70 @@ use tokio::runtime::{Builder, Runtime};
...
+impl std::fmt::Debug for AwsKms {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let kms_client = KmsClientDebug {
+            region: self.region.clone(),
+            endpoint: self.endpoint.clone(),
+        };
+        f.debug_struct("AwsKms")
+            .field("client", &kms_client)
+            .field("current_key_id", &self.current_key_id)
+            .field("runtime", &self.runtime)
```
yiwu-arbug, line 72: nit: just have `region`, `endpoint` and `current_key_id` as fields, no need to define a `KmsClientDebug` type and not printing runtime?

components/encryption/src/master_key/kms.rs
```diff
@@ -52,28 +91,35 @@ impl AwsKms {
...
 
-    fn check_config(config: &KmsConfig) -> Result<()> {
-        if config.key_id.is_empty() {
-            return Err(box_err!("KMS key id can not be empty"));
-        }
-        Ok(())
+    pub fn new(config: KmsConfig) -> Result<AwsKms> {
+        // We must create our own dispatcher
+        // See https://github.com/tikv/tikv/issues/7236
+        let dispatcher = HttpClient::new().unwrap();
```
yiwu-arbug, line 104: return error instead of panic as much as possible.

components/encryption/src/master_key/kms.rs
```diff
@@ -148,92 +197,71 @@ where
...
+        self.encryption_backend = MemAesGcmBackend::new(datakey.plaintext)?;
         Ok(())
     }
+
+    fn cached(&self, ciphertext_key: &Encrypted) -> bool {
+        *ciphertext_key == self.cached_ciphertext_key
+    }
+}
+
+pub struct DataKey {
```
yiwu-arbug, line 226:
```suggestion
pub struct KmsDataKey {
```
Want to distinguish it with `kvproto::encryptionpb::DataKey`.

components/encryption/src/master_key/kms.rs
```diff
@@ -52,28 +91,35 @@ impl AwsKms {
...
+
+impl KmsProvider for AwsKms {
+    fn name(&self) -> &[u8] {
+        AWS_KMS_VENDOR_NAME
     }
 
     // On decrypt failure, the rule is to return WrongMasterKey error in case it is possible that
     // a wrong master key has been used, or other error otherwise.
-    fn decrypt(&mut self, ciphertext: &[u8]) -> Result<Vec<u8>> {
+    fn decrypt(&mut self, ciphertext: EncryptedRef) -> Result<Vec<u8>> {
```
yiwu-arbug, line 116: rename the type `EncryptedRef` to `CiphertextKey`, if you want to use the newtype ideom. Definition of the type should be moved to this file, as it is KMS specific.

yiwu-arbug: It would be nice to also define a new type for the plaintext key.

components/encryption/src/master_key/kms.rs
```diff
@@ -148,92 +197,71 @@ where
...
+    fn generate_data_key(&mut self) -> Result<DataKey>;
+    fn decrypt(&mut self, ciphertext: EncryptedRef) -> Result<Vec<u8>>;
+    fn name(&self) -> &[u8];
+}
+
+#[derive(Debug)]
 pub struct KmsBackend {
-    inner: Mutex<Inner>,
+    state: Mutex<State>,
+    kms_provider: Mutex<Box<dyn KmsProvider>>,
```
yiwu-arbug, line 240: Since `KmsProvider` is `Sync`, we don't need a mutex to guard it.

components/encryption/src/master_key/kms.rs
```diff
@@ -148,92 +197,71 @@ where
...
     }
+
+    fn cached(&self, ciphertext_key: &Encrypted) -> bool {
+        *ciphertext_key == self.cached_ciphertext_key
+    }
+}
+
+pub struct DataKey {
+    pub encrypted: Encrypted,
+    pub plaintext: Vec<u8>,
```
yiwu-arbug, line 228: should avoid the plaintext key being printed.

components/encryption/src/master_key/kms.rs
yiwu-arbug, line 226:
```suggestion
pub struct DataKeyPair {
```

components/encryption/src/master_key/kms.rs
```diff
@@ -148,92 +197,71 @@ where
...
 impl KmsBackend {
-    pub fn new(config: KmsConfig) -> Result<KmsBackend> {
-        let inner = Inner {
-            backend: None,
-            config,
-            cached_ciphertext_key: Vec::new(),
-        };
-
+    pub fn new(mut kms_provider: Box<dyn KmsProvider>) -> Result<KmsBackend> {
+        let state = State::new_from_datakey(kms_provider.generate_data_key()?)?;
```
yiwu-arbug, line 245: Shall we keep the existing behavior to lazily initialize the kms data key? In case preexisting encryption metadata presents, we don't need this extra `generate_data_key` call.

components/encryption/src/master_key/kms.rs
gregwebs: The Mutex is to avoid requiring an `&mut` for `KmsBackend` in the `decrypt` function.

components/encryption/src/config.rs
```diff
@@ -55,8 +50,8 @@ pub struct FileConfig {
 pub struct KmsConfig {
     pub key_id: String,
 
-    // Providing access_key and secret_access_key is recommended as it has
-    // security risk.
+    // It is preferable to use the cloud provider SDK to automatically rotate credentials
+    // Using access_key and secret_access_key creates a security risk.
```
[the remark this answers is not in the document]
yiwu-arbug: yes, do you mind help me update the macro?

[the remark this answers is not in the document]
gregwebs: I think it would be better to make this change separately. I have experience doing the same thing in our go code and always found that the solution to these issues is just to delete any custom aws credentials code and use what the SDK provides by default.

components/encryption/src/manager/mod.rs
```diff
@@ -246,8 +243,13 @@ impl Dicts {
                 Some(file_info) => file_info.clone(),
                 None => {
                     // Could be a plaintext file not tracked by file dictionary.
-                    info!("link untracked plaintext file"; "src" => src_fname, "dst" => dst_fname);
-                    return Ok(());
+                    return Err(Error::Io(IoError::new(
```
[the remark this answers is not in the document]
gregwebs: In that case can we check to see that the pre-existing plaintext file exists on the file system?

components/encryption/src/config.rs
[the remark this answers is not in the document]
yiwu-arbug: "use what the SDK provides by default` is implemented in the `new_client!` macro. We just need a variant of the macro that don't look at the custom credential but just use the default credentials.

components/encryption/src/crypter.rs
```diff
@@ -197,6 +198,81 @@ pub fn verify_encryption_config(method: EncryptionMethod, key: &[u8]) -> Result<
...
+}
+
+impl Deref for EncryptedRef<'_> {
+    type Target = [u8];
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+// PlainKey is a newtype used to mark a vector a plain text key.
```
yiwu-arbug, line 243:
```suggestion
// PlainKey is a newtype used to mark a vector a plaintext key.
```

components/encryption/src/manager/mod.rs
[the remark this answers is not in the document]
yiwu-arbug: we always assume the source file exists on the file system when `link_file` is called. If the file entry exists in the file dictionary, we assume the file is encrypted, otherwise unencrypted.

components/encryption/src/config.rs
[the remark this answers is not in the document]
yiwu-arbug: the explicit credentials is not being used by anyone for the KMS flow. It is only used for S3 for BR.

components/encryption/src/crypter.rs
```diff
@@ -197,6 +198,81 @@ pub fn verify_encryption_config(method: EncryptionMethod, key: &[u8]) -> Result<
     Ok(())
 }
 
+// Encrypted is a newtype used to mark data as encrypted
+// It requires the vec to be non-empty
+#[derive(PartialEq, Clone, Debug)]
+pub struct Encrypted(Vec<u8>);
```
yiwu-arbug, line 204: how about `EncryptedKey`, to be consistent with `PlainKey`?

components/encryption/src/manager/mod.rs
```diff
@@ -586,17 +603,30 @@ impl DataKeyManager {
...
 
 impl EncryptionKeyManager for DataKeyManager {
     // Get key to open existing file.
     fn get_file(&self, fname: &str) -> IoResult<FileEncryptionInfo> {
         let (method, key_id, iv) = {
-            let file = self.dicts.get_file(fname);
-            (file.method, file.key_id, file.iv)
+            match self.dicts.get_file(fname) {
+                Some(file) => (file.method, file.key_id, file.iv),
+                None => {
```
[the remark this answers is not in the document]
yiwu-arbug: then why we need to refactor `get_file`? `get_file_exists` looks redundant. Again if the file entry doesn't exist in file dictionary, it could mean the file does not exist, or the file exists but not encrypted. In the key manager we don't care about if the file exists on the file system. We return file info with method = plaintext and let the caller decide what to do.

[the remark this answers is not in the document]
yiwu-arbug: Or in another word, we always assume the caller of `get_file` knows the file exists.

components/encryption/src/master_key/kms.rs
yiwu-arbug: do you have a sample output for the `runtime`?

components/encryption/src/master_key/kms.rs
yiwu-arbug: well, in this context it is the encrypted data key. it is more clear if the type name is specific.

components/encryption/src/master_key/kms.rs
yiwu-arbug: why you want to avoid `&mut`?

components/encryption/src/master_key/kms.rs
```diff
@@ -256,19 +308,31 @@ impl KmsBackend {
...
-        let ciphertext_key = content.metadata.get(MetadataKey::KmsCiphertextKey.as_str());
-        if ciphertext_key.is_none() {
-            return Err(box_err!("KMS ciphertext key not found"));
+        let ciphertext_key = match content.metadata.get(MetadataKey::KmsCiphertextKey.as_str()) {
+            None => return Err(box_err!("KMS ciphertext key not found")),
+            Some(key) => Encrypted::new(key.to_owned())?,
+        };
+
+        let mut state = self.state.lock().unwrap();
+        self.lazy_load_state(&mut *state)?;
```
yiwu-arbug, line 323: Add comment to say the KMS data key is rotated every time on restart? This is different from previous behavior but I think it is a good change.

components/encryption/src/crypter.rs
gregwebs: PlainKey differentiates a key from other unencrypted data. `Encrypted` can be used to mark different types of data (a key or a payload) as encrypted. I could see this developing into `Key<T>` where `T` could be `Encrypted` or not.

components/encryption/src/master_key/kms.rs
gregwebs: The other backends and the interface don't need `&mut`, and `encrypt` does not need it, so I was hesitant to change the interface to `&mut` just for `decrypt`: to me that interface definition would look odd. I suppose I can add some comments saying that it is mutable for the purposes of caching internal state.

components/encryption/src/crypter.rs
yiwu-arbug: I don't see the code being more clear after introducing this type. It doesn't serve any functional purpose, and don't make the logic more clear.

components/encryption/src/master_key/kms.rs
yiwu-arbug: I think you can remove `&mut` from `KmsProvider::generate_data_key`, then you don't need a mutex for the `kms_provider`.

gregwebs: In `generate_data_key` `retry()` requires `runtime` to be mutable.

components/encryption/src/crypter.rs
gregwebs: I think I agree with you because we already have an EncryptedContent type in use.

components/encryption/src/manager/mod.rs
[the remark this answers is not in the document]
gregwebs: It is strange to have to assume so much. I am wondering if this can be improved. However, it seems that some of this behavior is forced by the RocksDB encryption implementation?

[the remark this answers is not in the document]
yiwu-arbug: not really. we make use of some of original rocksdb encryption implementation, but the key manager logic are new. The design of the key manager and the encryption module is that it has no knowledge of the actual file system and only care about encryption metadata. Interaction with file system should be handled in the higher level, i.e. the work-in-progress `file_system` crate.

components/encryption/src/master_key/kms.rs
yiwu-arbug: Not familiar with tokio, but maybe `runtime` should not be member of `AwsKms`, instead being created for each invocation of `decrypt`?

yiwu-arbug: cc @overvenus

yiwu-arbug:
cc @overvenus for logic related to #7236 fix.

components/encryption/src/crypter.rs
```diff
@@ -197,6 +197,39 @@ pub fn verify_encryption_config(method: EncryptionMethod, key: &[u8]) -> Result<
...
+
+impl std::ops::Deref for PlainKey {
+    type Target = Vec<u8>;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+// Don't expose the Key in a debug print
+impl std::fmt::Debug for PlainKey {
```
overvenus, line 225: Could you also sanitize `std::fmt::Display`?

components/encryption/src/encrypted_file/mod.rs
```diff
@@ -70,7 +70,7 @@ impl<'a> EncryptedFile<'a> {
             .create(true)
             .write(true)
             .open(&tmp_path)
-            .unwrap();
+            .unwrap_or_else(|_| panic!("EncryptedFile::write {}", &tmp_path.to_str().unwrap()));
```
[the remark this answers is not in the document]
overvenus: Panic here might be simpler than propagating errors, this error is fatal, callers may panic anyway.

components/external_storage/src/s3.rs
```diff
@@ -27,22 +27,16 @@ use kvproto::backup::S3 as Config;
 pub struct S3Storage {
     config: Config,
     client: S3Client,
-    // The current implementation (rosoto 0.43.0 + hyper 0.13.3) is not `Send`
-    // in practical. See more https://github.com/tikv/tikv/issues/7236.
-    // FIXME: remove it.
-    _not_send: PhantomData<*const ()>,
```
overvenus, line 33: Without this phantom, `S3Storage` is error-prone, it may cause panic if it is used in other threads.

gregwebs: From my reading the underlying github issues, the problem is from using the global dispatcher. The code below here is changed to instead create our own dispatcher. Let me know if you have any specific tests for this that I could run.

components/encryption/src/manager/mod.rs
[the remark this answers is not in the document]
gregwebs: Can the caller check for this case before calling the function? Then we don't need any assumptions.

components/encryption/src/master_key/kms.rs
gregwebs: Runtimes are generally designed to be persistent. I suspect we should not be creating a runtime but instead passing one in from a top level. That will still require a mutable runtime. However, in either case if we assume all KmsProvider needs the tokio runtime we can pass it in to the decrypt_data_key function.

gregwebs: Okay, runtime is now in State and passed into the backend, so the backend is now immutable and does not need a Mutex.

components/encryption/src/config.rs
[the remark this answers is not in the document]
gregwebs: I think we are in agreement. I will make a separate PR for this though.

components/encryption/src/encrypted_file/mod.rs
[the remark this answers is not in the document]
gregwebs: It was already performing a panic. I need to limit the scope of my improvements in this large PR. There are unwraps everywhere in our code base.

components/encryption/src/config.rs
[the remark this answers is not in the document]
yiwu-arbug: okay

components/encryption/src/manager/mod.rs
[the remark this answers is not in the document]
yiwu-arbug: That's not an assumption. The point is the encryption crate should not aware of file system stuff, and just focus on encryption metadata.

components/encryption/src/master_key/kms.rs
```diff
@@ -228,20 +227,41 @@ impl State {
...
-    fn cached(&self, ciphertext_key: &Encrypted) -> bool {
+    fn cached(&self, ciphertext_key: &EncryptedKey) -> bool {
         *ciphertext_key == self.cached_ciphertext_key
     }
 }
 
+// EncryptedKey is a newtype used to mark data as an encrypted key
+// It requires the vec to be non-empty
+#[derive(PartialEq, Clone, Debug)]
+pub struct EncryptedKey(Vec<u8>);
```
yiwu-arbug, line 226: maybe we don't even want the encrypted key to be printed?

components/encryption/src/crypter.rs
```diff
@@ -198,7 +198,7 @@ pub fn verify_encryption_config(method: EncryptionMethod, key: &[u8]) -> Result<
 }
 
 // PlainKey is a newtype used to mark a vector a plaintext key.
-// It requires the vec to be non-empty.
+// It requires the vec to be a valied AesGcmCrypter key.
```
yiwu-arbug, line 201: typo: valid

components/encryption/src/master_key/kms.rs
```diff
@@ -256,19 +323,32 @@ impl KmsBackend {
...
-        let mut inner = self.inner.lock().unwrap();
-        let ciphertext_key = content.metadata.get(MetadataKey::KmsCiphertextKey.as_str());
-        if ciphertext_key.is_none() {
-            return Err(box_err!("KMS ciphertext key not found"));
+        let ciphertext_key = match content.metadata.get(MetadataKey::KmsCiphertextKey.as_str()) {
+            None => return Err(box_err!("KMS ciphertext key not found")),
+            Some(key) => EncryptedKey::new(key.to_vec())?,
+        };
+
+        let mut state = self.state.lock().unwrap();
```
yiwu-arbug, line 337: scoping the mutex lock for readablity:
```
{
    let mut state = self.state.lock().unwrap();
    // the rest of code that uses the state
}
```

components/encryption/src/master_key/kms.rs
gregwebs: In theory it is safe to publish encrypted data publicly, so I am not worried about the security. I think we need some way to identify the data key when debugging.

components/encryption/src/encrypted_file/mod.rs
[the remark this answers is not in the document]
yiwu-arbug: well, i actually prefer propagate the error to upper layer, which prints the error message of encryption failure, before panicking. I hope we can reduce the unnecessary panics one by one, but since the panic exists previously, it is okay not to update it.

components/encryption/src/master_key/kms.rs
gregwebs: it's the entire rest of the function so the behavior and the compiler analysis is exactly the same. Is this better for developers to read the code? It is also possible it will be confusing to someone reading the code since they don't have any effect.

components/encryption/src/master_key/kms.rs
yiwu-arbug: cool

components/external_storage/src/s3.rs
yiwu-arbug: @gregwebs I mainly have two questions for the issue, do you mind explain a bit?
1. how does `tokio::runtime::Builder::basic_scheduler()` work? From the document it says the runtime will execute tasks on "current thread". What if I create a basic runtime on thread T1, send it over to thread T2, and execute some task? Does T1 or T2 count for "current thread"?
2. how does creating a new dispatcher for each `AmsKms` and `S3Storage` instances (which are `Sync + Send`, so they allow people to send them to different threads)  fixes the issue?

components/encryption/src/master_key/kms.rs
yiwu-arbug: okay

components/encryption/src/master_key/kms.rs
yiwu-arbug: It is a common practices to scope the lock guard for readability. It make it crystal clear how far is the critical session.

yiwu-arbug: for example in an later iteration people add more unrelated logic at the end of this function, and may forget to unlock the mutex. explicit scoping avoids that.

components/external_storage/src/s3.rs
gregwebs: 1) I don't know. The behavior after this PR is merged should be the same as previous.

2) My reading of the issue is that there is a global dispatcher that is initialized with lazy_static and it turns out it is not actually safe to use it. But instead you can just create a dispatcher at runtime yourself.

gregwebs: This is the default shared client: https://github.com/rusoto/rusoto/blob/d59d716f097f64a713b3bbcd5fc7b84e210647e4/rusoto/core/src/client.rs#L16
This is how it is used: https://github.com/rusoto/rusoto/search?q=Client%3A%3Ashared

I don't know specifically what the problem is, but I can easily believe that this would cause problems.

components/encryption/src/manager/mod.rs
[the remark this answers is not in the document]
gregwebs: We talked about this and agreed that in a separate PR we can change this from `Ok(()` to `None` and return an `Option` type.

components/external_storage/src/s3.rs
gregwebs: Since S3 does not need this change I will keep S3 as is with a note about how this is unnecessary now. I will remove the phantom from KMS because it shouldn't be necessary and even if there were a race condition in hyper it would be difficult for KMS to hit it now that it uses its own client since it very rarely makes http calls.

components/encryption/src/master_key/kms.rs
gregwebs: okay, the prior behavior is restored now

yiwu-arbug:
Waiting for testing and @overvenus 's review for the tokio runtime related logic, before approving.

gregwebs:
@yiwu-arbug I went ahead and removed the static credentials as per your suggestion. They were interfering with my testing.

components/external_storage/src/s3.rs
gregwebs: [This comment](https://github.com/hyperium/hyper/issues/2112#issuecomment-630857751) indicates that avoiding the lazy_static client (connected to a global dispatcher) as the code does now fixes the issue.

components/encryption/src/manager/mod.rs
```diff
@@ -238,7 +234,7 @@ impl Dicts {
         Ok(())
     }
 
-    fn link_file(&self, src_fname: &str, dst_fname: &str) -> Result<()> {
+    fn link_file(&self, src_fname: &str, dst_fname: &str) -> Result<Option<()>> {
```
overvenus, line 237: Why returning `Result<Option<()>>`? Is it possible to replace it with `Result<bool>`?

overvenus (commented):
Rest LGTM

components/encryption/src/manager/mod.rs
gregwebs: Both represent two states. The problem with using `bool` is knowing what it means. For `is_empty` it is obvious, but in cases like these it is not. The file apis here are being changed to return an `Option` with `None` meaning the operation did not occur because the file did not exist. So `Option<T>` is the consistent choice for these APIs and in this case `Option<()>` will be understandable to someone familiar with the APIs.

components/sst_importer/src/util.rs
```diff
@@ -61,7 +61,15 @@ pub fn prepare_sst_for_ingestion<P: AsRef<Path>, Q: AsRef<Path>>(
...
-        key_manager.link_file(path, clone)?;
+        match key_manager.link_file(path, clone) {
+            // Ignore NotFound error
+            // In this case the file dict does not have an entry for the source file
+            // The encryption method must be plaintext for that file
+            Err(e) if e.kind() != io::ErrorKind::NotFound => {
+                return Err(std::convert::From::from(e))
+            }
+            _ => {}
+        }
```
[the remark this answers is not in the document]
gregwebs: I changed the behavior of link_file back to how it was previously, so I should remove this change now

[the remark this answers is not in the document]
gregwebs: This is changed back now

overvenus:
LGTM (including S3 client change)

overvenus:
LGTM

gregwebs:
I rebased the PR to 2 commits. I did a diff of the rebase and what was previously pushed for the branch to ensure they were the same.

hunterlxt:
Sorry late for response, the PR is large so I read several times.

hunterlxt (approved):
/lgtm

yiwu-arbug (approved):
LGTM

PR #9358, opened 2020-12-24
*: fix make build PROST=1

kennytm:
### What problem does this PR solve?

Issue Number: close #9352

Problem Summary: `make build PROST=1` failed to compile.

### What is changed and how it works?

What's Changed:

* Changed TiKV's Cargo.toml to use "resolver = 2" so that `cmd` can recognize `feature = "prost-codec"`.
* In Prost, we have the `FieldType::tp()` inherent method which has name conflict with `FieldTypeAccessor::tp()` trait method. Use the FQN to disambiguate.
* Use `as _` to bridge enum-to-int conversion.
* Use `.into()` to bridge T-to-Box&lt;T&gt; conversion.

### Related changes

- Need to cherry-pick to the release branch
    * release-5.0-rc

### Check List 

Tests 

- Manual test (add detailed scripts or steps below)
   * make build PROST=1
   * make build

Side effects

### Release note 

No release notes

PR #9430, opened 2021-01-04
Fix build of all crates under default features, and prost-codec / test-engines-panic

### LINKED ISSUE #9406: Many crates in workspace no longer build on their own
PR https://github.com/tikv/tikv/pull/9358 changed cargo to use `resolver = 2`, which is described [here](https://doc.rust-lang.org/cargo/reference/unstable.html#resolver).

This changes, among other things, feature resolution to be based on the crates selected to build/check, instead of the workspace crate. After this change, building many crates results in the appropriate protobuf-codec/prost-codec features not being set, as the individual crates were relying on the tikv crate to properly configure all crates.

Using the newer resolver is probably the right thing to do long term though, since that is how future-cargo is expected to work.

My workflow typically involves running `cargo check` on individual crates, so I expect all crates to be buildable by default. In some cases it is probably reasonable for crates to _not_ have a buildable default feature set, but in tikv's case, I think it's best for the contributor experience that every crate can build on its own. Otherwise, one needs to look at each crate's manifest and figure out what features to turn on.

Unfortunately, realizing this the obvious way requires adding a _lot_ more boilerplate to all the manifests in the workspace, setting `default = ["protobuf-codec"]`, and correctly setting `default-features = false` everywhere necessary.

I've started working on a patch that does this, as well as introducing a script that ensures that all crates always build under a variety of configurations. It does though feel like something is missing in the rust tooling that managing complex trees of features is so incredibly arduous.

brson:

Issue Number: close #9406

Problem Summary:

Building many individual crates with `-p` fails, particularly after https://github.com/tikv/tikv/pull/9358 changed cargo to use the v2 resolver.

In some cases the default configuration is unbuildable for not specifying the `protobuf-codec` feature, in some cases `prost-codec` is broken, in some cases `test-engines-panic` is broken, and there are a few other broken cases.

What's Changed:

This patch makes sure every crate that needs `protobuf-codec` to build specifies it as a default feature, and that every crate that depends on a crate with a `protobuf-codec` default feature turns _off_ the default feature, and instead uses its own `protobuf-codec` / `prost-codec` feature to turn on the correct codec. This creates a connected tree of crates with `protobuf-codec / `prost-codec` features.

It does the same for `test-engines-rocksdb` / `test-engines-panic`.

It further fixes up some other cases where crates don't build on their own.

Note that the "external_storage: make sure openssl is vendored" commit contains a hack, where I use the grpcio crate to correctly include the openssl dependency. I could not otherwise figure out how to vendor the correct openssl directly (other means either had missing or duplicate symbols).

Note that this approach assumes that we _want_ all crates in the workspace to build by default. IMO we do - at least I do a lot of development on a single crate at a time, and I think it makes the contributor experience better. On the other hand, not having a default feature set simplifies some of the manifest maintenance, while making the cargo invocation for building any given crate require extra work.

In the final commit I add the script I used to verify that all crates build correctly. I am not sure what to do with this script as it takes 100m to run locally, so it is neither suitable as part of the test suite, nor as part of CI. But _some kind_ of automation is needed to keep the build working in all its configurations - the feature setup is so complex it is sure to break again. At least this script can be used to help fix it later.

### Release note 

- No release note

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1160,7 +1160,6 @@ dependencies = [
  "slog-global",
  "structopt",
  "tempfile",
- "test_util",
  "tikv_alloc",
  "tikv_util",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1286,6 +1286,7 @@ dependencies = [
  "futures-executor",
  "futures-io",
  "futures-util",
+ "grpcio",
  "http",
  "hyper",
  "hyper-tls",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,12 +61,10 @@ protobuf-codec = [
 ]
 testexport = ["raftstore/testexport"]
 test-engines-rocksdb = [
-  "engine_test/test-engine-kv-rocksdb",
-  "engine_test/test-engine-raft-rocksdb",
+  "engine_test/test-engines-rocksdb",
 ]
 test-engines-panic = [
-  "engine_test/test-engine-kv-panic",
-  "engine_test/test-engine-raft-panic",
+  "engine_test/test-engines-panic",
 ]
 bcc-iosnoop = [
   "file_system/bcc-iosnoop",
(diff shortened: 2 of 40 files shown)


brson:
cc @zhouqiang-cl This pr contains a script that ensures every crate in the workspace has its cargo features configured correctly. It is not run as part of the test suite though because it takes a long time to run. But not running, or something like it, guarantees there will be regressions in the future. If tikv has a periodic batch testing job that is allowed to be slow, this script would be a good candidate to run there.

nrc (approved):
lgtm

yiwu-arbug (approved):

brson:
Thanks @yiwu-arbug

PR #9441, opened 2021-01-05
resolved_ts: introduce cmd encoder

5kbpers:

Problem Summary:
Separate from https://github.com/tikv/tikv/pull/9435
* Introduce cmd encoder.

What's Changed:
* implement cmd encoder to encode change log.
* move `Resolver` to `resolver.rs`

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A


components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,288 @@
...
+                    None => continue,
+                },
+                "" | "default" => {
+                    let key = Key::from_encoded(put.take_key());
+                    let start_ts = key.decode_ts().unwrap();
+                    let key = key.truncate_ts().unwrap();
+                    match change_logs.get_mut(&key) {
+                        Some(ChangeRow::Prewrite { value, lock, .. }) => {
+                            assert!(value.is_none());
+                            assert_eq!(start_ts, lock.ts);
```
NingLin-P, line 154: How about doing these checks (`ChangeRow::Prewrite` & `ChangeRow::OnePc`) after all requests are processed, in case L115 or L140 return a `None` and not incoming put on `default`.

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,288 @@
...
+                        }
+                    }
+                    None => continue,
+                },
+                "lock" => match decode_lock(put.get_key(), put.get_value()) {
+                    Some(lock) => {
+                        let key = Key::from_encoded(put.take_key());
+                        let value = pending_default.remove(&key);
+                        let l = change_logs
+                            .insert(key.clone(), ChangeRow::Prewrite { key, lock, value });
```
NingLin-P, line 142: The lock may contain a `short_value`, should we set the `ChangeRow::Prewrite::value` here in this case?

5kbpers: Not necessary. Just notice to handle it 🤔.

components/resolved_ts/src/cmd.rs
5kbpers: If a prewrite has a short value, it will never get into this path.

NingLin-P: What I mean is to check whether the `CmdBatch` is completed (panic when both `ChangeRow::Prewrite::value` and `ChangeRow::Prewrite::lock::short_value` is `None` or `pending_default` is not empty after all `Request` are processed), I think this is what the `assert` here checking about.

NingLin-P (approved):

components/resolved_ts/Cargo.toml
```diff
@@ -7,18 +7,54 @@ publish = false
...
+futures = "0.3"
+grpcio = { version = "0.7", default-features = false, features = ["openssl-vendored"] }
+hex = "0.4"
+kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
+log_wrappers = { path = "../log_wrappers" }
+pd_client = { path = "../pd_client" }
+prost = "0.6"
+protobuf = "2.8"
+raft = { version = "0.6.0-alpha", default-features = false }
+raftstore = { path = "../raftstore" }
```
BusyJay, line 52: Isn't stale read going to depend on resolved ts? Isn't it a cyclic dependency?

src/storage/kv/mock_engine.rs
```diff
@@ -13,7 +13,15 @@ use std::sync::{Arc, Mutex};
 #[derive(Clone)]
 pub struct MockEngine {
     base: RocksEngine,
-    expected_modifies: Arc<ExpectedWriteList>,
+    expected_modifies: Option<Arc<ExpectedWriteList>>,
```
BusyJay, line 16: Why make it an `Option`?

src/storage/txn/actions/tests.rs
```diff
@@ -440,3 +440,30 @@ pub fn must_rollback_err<E: Engine>(engine: &E, key: &[u8], start_ts: impl Into<
     let mut txn = MvccTxn::new(snapshot, start_ts, true, cm);
     assert!(txn::cleanup(&mut txn, Key::from_raw(key), TimeStamp::zero(), false).is_err());
 }
+
+pub fn must_one_pc<E: Engine>(
```
BusyJay, line 444: @sticnarf PTAL

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,288 @@
...
+use collections::HashMap;
+use kvproto::errorpb;
+use kvproto::raft_cmdpb::{AdminCmdType, CmdType, Request};
+use raftstore::coprocessor::{Cmd, CmdBatch};
+use raftstore::errors::Error as RaftStoreError;
+use tikv::server::raftkv::WriteBatchFlags;
+use txn_types::{Key, Lock, LockType, TimeStamp, Value, Write, WriteRef, WriteType};
+
+#[derive(Debug)]
+pub enum ChangeRow {
```
BusyJay, line 12: @overvenus PTAL

components/resolved_ts/Cargo.toml
5kbpers: No, raftstore will not depend on resolved ts.

src/storage/kv/mock_engine.rs
5kbpers: To indicate whether we need to check it.

src/storage/txn/actions/tests.rs
5kbpers: Remove it since not be used...

5kbpers:
@sticnarf @MyonKeminta PTAL

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,347 @@
...
+            })
+            .filter_map(|v| v)
+            .collect()
+    }
+
+    fn encode_rows(changes: HashMap<Key, RowChange>, is_one_pc: bool) -> Vec<ChangeRow> {
+        changes
+            .into_iter()
+            .map(|(key, row)| {
+                if row.write.is_some() {
```
sticnarf, line 93: How about organizing these branches using `match`?

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,347 @@
...
+                    let (mut commit_ts, write) = row.write.unwrap().into_put();
+                    if let Some(write) = decode_write(key.as_encoded(), &write, true) {
+                        let value = row.default.map(|v| v.into_put().1);
```
sticnarf, line 96: `map`? It means we don't have value if `default` is `None`.

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,347 @@
...
+use raftstore::errors::Error as RaftStoreError;
+use tikv::server::raftkv::WriteBatchFlags;
+use txn_types::{Key, Lock, LockType, TimeStamp, Value, Write, WriteRef, WriteType};
+
+#[derive(Debug)]
+pub enum ChangeRow {
+    Prewrite {
+        key: Key,
+        lock: Lock,
+        value: Option<Value>,
```
sticnarf, line 16: Now we have two fields that may contain the real value: the `short_value` in `lock` and this `value`. I prefer to avoid this. What about flattening the `Lock` struct and only keeping the necessary fields? Then we combine the short value and long value into one.
What do you think?

sticnarf: And the same for `OnePc`

5kbpers: I'm afraid of adding too many fields in the future. Keep both `short_value` and `value` is consistent with our original design, so I think it's ok.

components/resolved_ts/src/cmd.rs
5kbpers: Done

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,340 @@
...
+            })
+            .filter_map(|v| v)
+            .collect()
+    }
+
+    fn encode_rows(changes: HashMap<Key, RowChange>, is_one_pc: bool) -> Vec<ChangeRow> {
+        changes
+            .into_iter()
+            .map(|(key, row)| match (row.write, row.lock, row.default) {
+                (Some(write_op), None, default) => {
```
sticnarf, line 93: lock should be "delete" for normal commit

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,372 @@
...
+        value: Option<Value>,
+    },
+    Commit {
+        key: Key,
+        write_type: WriteType,
+        start_ts: TimeStamp,
+        commit_ts: Option<TimeStamp>,
+        // In some cases a rollback will be done by just deleting the lock, need to lookup Resolver
+        // to untrack lock and get start ts from it.
+        lack_start_ts: bool,
```
sticnarf, line 26: What about `start_ts: Option<TimeStamp>`?

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,372 @@
...
+                }
+                (None, Some(KeyOp::Put(_, lock)), default) => decode_lock(key.as_encoded(), &lock)
+                    .map(|lock| {
+                        let Lock {
+                            short_value,
+                            ts,
+                            lock_type,
+                            ..
+                        } = lock;
+                        let value = default.map_or_else(|| short_value, |v| Some(v.into_put().1));
```
sticnarf, line 133: ditto

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,372 @@
...
+        let engine = MockEngineBuilder::from_rocks_engine(rocks_engine).build();
+
+        must_prewrite_put(&engine, b"k1", b"v1", b"k1", 1);
+        must_commit(&engine, b"k1", 1, 2);
+
+        must_prewrite_put(&engine, b"k1", b"v2", b"k1", 3);
+        must_rollback(&engine, b"k1", 3);
+
+        must_prewrite_put(&engine, b"k1", &[b'v'; 512], b"k1", 4);
+
```
sticnarf, line 290: I think some cases are missing:

- delete keys
- 1pc

sticnarf (commented):
The rest lgtm

sticnarf (approved):

components/resolved_ts/src/cmd.rs
```diff
@@ -0,0 +1,401 @@
...
+                }
+            })
+            .filter_map(|v| v)
+            .collect()
+    }
+
+    fn encode_rows(changes: HashMap<Key, RowChange>, is_one_pc: bool) -> Vec<ChangeRow> {
+        changes
+            .into_iter()
+            .map(|(key, row)| match (row.write, row.lock, row.default) {
```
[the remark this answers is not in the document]
5kbpers: I think it seems is cleaner to use an iterator.

PR #9458, opened 2021-01-07
engine: refactor `MetricsFlusher` and add fallback storage IO measuring

tabokie:

What's Changed:

Remove metrics flusher in engine_traits and use background worker instead, add IO metrics task which reports statistics from iosnooper, or IO rate limiter if BCC is unavailable.

### Check List 

Tests 

- Unit test

### Release note 

No release note


cmd/src/server.rs
```diff
@@ -107,9 +110,14 @@ pub fn run_tikv(config: TiKvConfig) {
         ($ER: ty) => {{
             let enable_io_snoop = config.enable_io_snoop;
             let mut tikv = TiKVServer::<$ER>::init(config);
-            if enable_io_snoop {
+            let recorder = if enable_io_snoop {
                 tikv.init_io_snooper();
-            }
+                None
+            } else {
+                let recorder = Arc::new(BytesRecorder::new());
```
Connor1996, line 117: can we make the iosnoop to be a kind of BytesRecorder?

tabokie: Since iosnoop doesn't interact with limiter the same way as bytes recorder, I don't think so. It might be better to extract the global IOContext out. I'll leave it to another PR where we actually use IOContext.

Connor1996: Seems the difference is the `add()`  of `BytesRecorder`. For iosnoop, we can just do nothing in `add()` and use `IOContext` to get the difference for `fetch()`

tabokie: Latency isn't included in bytes recorder, and doing that means spreading the optional compilation directive all over the place.

Connor1996: Latency is just for metrics, you don't consider it in BytesRecorder. And seems optional compilation can be bounded by file_system providing a uniform interface like `create_bytes_recorder(some_config)`.

tabokie: Got it, i thought latency is stored in IOContext too.

tabokie:
Metrics sample:
image

brson:
Just from the perspective of the engine_traits abstraction this looks fine to me. I do think that MetricsTask doesn't really belong in engine_traits since it is just building on the engine_traits and not an integral part of a storage engine. But since it is not leaking any engine details I don't mind having it in engine_traits for now. It is easy enough to move somewhere else someday, if it ever really matters.

I can't really evaluate the details of what it is doing with I/O and metrics. Because of that I think others should approve this and not me.

tabokie:
some benchbot results:
oltp_update_non_index: 17088.60(+2.7%) per sec, latency(ms) 95th 10.84(-), avg 5.85(-2.6%)
oltp_point_select: 77922.99(-0.95%) per sec, latency(ms) 95th 2.91(-), avg 1.28(-)

components/engine_traits/src/metrics_task.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use tikv_util::IntervalRunnable;
+
+use crate::engine::KvEngine;
+use crate::engines::Engines;
+use crate::raft_engine::RaftEngine;
+
+const FLUSHER_RESET_INTERVAL: Duration = Duration::from_millis(60_000);
+
+pub struct MetricsTask<K: KvEngine, R: RaftEngine> {
```
yiwu-arbug, line 13: I agree with @brson this does not belong to the engine trait. Maybe move to tikv_util?

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -146,52 +145,19 @@ pub fn get_io_type() -> IOType {
...
-
-                return self.io_stats_map.replace(map).unwrap();
+pub(crate) fn fetch_io_bytes(io_type: IOType) -> IOBytes {
+    unsafe {
+        if let Some(ctx) = BPF_CONTEXT.as_mut() {
+            let mut buf = [0 as u8; std::mem::size_of::<IOType>()];
+            std::ptr::write(buf.as_mut_ptr() as *mut IOType, io_type);
+            if let Ok(e) = ctx.stats_table.get(&mut buf) {
+                let read = std::intrinsics::atomic_load(e.as_ptr() as *const u64) as i64;
+                let write = std::intrinsics::atomic_load((e.as_ptr() as *const u64).add(1)) as i64;
```
yiwu-arbug, line 155: btw what does `add(1)` mean here?

components/file_system/src/metrics_task.rs
```diff
@@ -0,0 +1,112 @@
...
+use tikv_util::IntervalRunnable;
+
+use crate::iosnoop::{fetch_io_bytes, flush_io_latency_metrics};
+use crate::metrics::IO_BYTES_VEC;
+use crate::BytesRecorder;
+use crate::IOBytes;
+use crate::{IOOp, IOType};
+
+pub enum BytesFetcher {
+    ByRateLimiter(Arc<BytesRecorder>),
```
yiwu-arbug, line 14: I assume this does not involve rate limit? If so rename it to ByBytesRecorder?

yiwu-arbug: Add some comment here about the difference between the two types?

yiwu-arbug (commented):
Looking good overall.

components/engine_traits/src/metrics_task.rs
brson: This can't live in tikv_util because this code depends on code in engine_traits, and engine_traits depends on tikv_util. It could live it tikv directly, or in some other intermediate crate between engine_traits and tikv.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -146,52 +145,19 @@ pub fn get_io_type() -> IOType {
...
-                        })
-                        .or_insert(stats.clone());
-                }
-
-                return self.io_stats_map.replace(map).unwrap();
+pub(crate) fn fetch_io_bytes(io_type: IOType) -> IOBytes {
+    unsafe {
+        if let Some(ctx) = BPF_CONTEXT.as_mut() {
+            let mut buf = [0 as u8; std::mem::size_of::<IOType>()];
+            std::ptr::write(buf.as_mut_ptr() as *mut IOType, io_type);
```
brson, line 152: Prompted by @yiwu-arbug's question I thought a bit about the unsafe code here, and these two lines don't look sound to me. If size_of<IOType> != 1 then I don't think there's any guarantee that this buffer is correctly aligned, which it needs to be for the cast back to `*mut IOType`. That is, I don't _think_ Rust has any guarantee that on-stack u8 arrays will be aligned to anything other than a u8.

It _looks_ to me that IOType is probably a u32 since it as a `repr(C)` enum.

You might test this with an assert that size of IOType == 1.

In general, it is not sound to cast u8 buffers into pointers of types that require larger than u8 alignment. If you do then you must _know_ that buffer is correctly aligned.

As written this type of code will often work most of the time, while still introducing undefined behavior, and could fail in rare scenarios.

The point of this code looks to be to pass an `&mut [u8]` that represents the given `IOType` to `ctx.stats_table.get`.

(Aside: it isn't clear to me why this buffer needs to be mutable, that is - why `bcc::Table::get` takes a mutable slice, as the only thing _in_ this buffer is a temporary C-style enum).

Instead of creating a u8 buffer and byte-copying the `io_type` into it, I think this code should instead be taking the approach of casting the given IOType to a buffer of u8.

You might do this by

- making io_type mutable
- casting a mutable pointer to `*mut u8`
- constructing a mutable u8 slice with `slice::from_raw_parts_mut(u8_ptr, size_of::<IOType>)`

components/file_system/src/iosnoop/biosnoop.rs
brson: Similar to my previous comment these casts to `*mut u64` don't look sound because they assume the `Vec<u8>` returned by `bcc::Table::get` is at least u64-aligned, and, unless `Table` is provide some special guarantee abut its return value, Rust does not guarantee the the underlying buffer here is u64-aligned.

Furthermore, I don't see why atomics are needed here at all. `bcc::Table::get` returns an owned `Vec<u8>` - there's no way another thread has access to this data.

These atomic loads look like they can be changed to safe conversions from `&[u8]` to u64. The easiest way I know of offhand to that conversion is with [byte_slice_cast::AsSliceOf](https://docs.rs/byte-slice-cast/1.0.0/byte_slice_cast/trait.AsSliceOf.html).

brson:
Both of my comments on the unsafe code here have me wondering if I'm just reading the code wrong and the `stats_table` is not actually a `bcc::Table` type...

components/file_system/src/iosnoop/biosnoop.rs
brson: Furthermore, this double-cast to a u64 pointer _then_ to an i64 seems suspicous. They are both 64-bits. Why not cast directly to an i64 pointer?

brson: Sorry for all the comments, but I just need to note that offhand I don't know the most simple and correct way to convert from `&[u8]` to `u64`. There must be some function somewhere that does this correctly in the presence of misalignment, but I don't know where to find it.

One part of the solution is [`u64::from_ne_bytes`](https://doc.rust-lang.org/stable/std/primitive.u64.html#method.from_ne_bytes) but to call that you first need a `[u8; 8]`.

To get that array you can probably use [`slice::as_chunks`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.as_chunks).

brson:
I know at this point I've left a lot of thoughts about `fetch_io_bytes` and it might not be clear exactly what I think should be done, but I do think it should mostly be rewritten, unless I'm misunderstanding how it uses `bcc` and `IOType`, which is very possible.

Here's a sketch of I rewrite using all my suggestions:

```rust
pub(crate) fn fetch_io_bytes(mut io_type: IOType) -> IOBytes {
    if let Some(ctx) = BPF_CONTEXT.as_mut() {
        unsafe {
            let io_type_buf_ptr = &mut io_type as *mut IOType as *mut u8;
            let io_type_buf = std::slice::from_raw_parts_mut(io_type_buf_ptr, std::mem::size_of::<IOType>);
            if let Ok(e) = ctx.stats_table.get(&mut io_type_buf) {
                let (chunks, _) = e.as_chunks();
                assert!(chunks.len() > 2);
                let read = std::u64::from_ne_bytes(chunks[0]);
                let write = std::u64::from_ne_bytes(chunks[1]);
                return IOBytes { read, write };
            }
        }
    }
    IOBytes::default()
}
```

tabokie:
@brson Thanks for your generous review, your rewrite looks neat. I looked into the rust-bcc [code](https://github.com/rust-bpf/rust-bcc/blob/c4e5ef83568b220c8bf271a6f80e9579b2df1e5e/src/table.rs#L65) and the value returned is actually a simple on-stack u8 Vec. But since this part of bcc-related code is still experimental, I'll leave a todo here for later investigation.

components/file_system/src/iosnoop/biosnoop.rs
yiwu-arbug: @brson the same buffer is passed via eBPF and modified in kernel space, so they can be accessed concurrently.

Connor1996: @brson you're right. I was thought that the `iter()` returns a ptr of the value, but actually, it's a ptr of on-stack vec.  So the atomics are not necessary here. 

For the u64-alignment, cause the definition of `stats_table` in `biosnoop.c` is 
```c
struct stats_t {
  u64 read;
  u64 write;
};
BPF_HASH(stats_by_type, io_type, struct stats_t);
```, seems we can sure that the returned value should be u64-alignment.

File: components/file_system/src/metrics_task.rs (hunk 4 above)
Comment (tabokie) [reply components/file_system/src/metrics_task.rs:14]: I'm inclined to keep the name because I plan to make limiter the owner of statistics and add toggle for recording (#9197).

File: components/file_system/src/iosnoop/biosnoop.rs (hunk 3 above)
Comment (tabokie) [reply components/file_system/src/iosnoop/biosnoop.rs:155]: I don't think the stats_t has any effect here. Maybe kernel side the memory is aligned, but the `rust-bcc` actually copied them into an on-stack vector ([code](https://github.com/rust-bpf/rust-bcc/blob/c4e5ef83568b220c8bf271a6f80e9579b2df1e5e/src/table.rs#L65)), making the alignment here suspicious.

File: components/file_system/src/metrics_task.rs (hunk 4 above)
Comment (tabokie) [reply components/file_system/src/metrics_task.rs:14]: Made the change to reflect this thought.

File: components/file_system/src/iosnoop/biosnoop.rs (hunk 3 above)
Comment (Connor1996) [reply components/file_system/src/iosnoop/biosnoop.rs]: I got what you mean. Seems we can simply write like that `ptr::read_unaligned(e.value.as_ptr() as *const IOStats)`.

Comment (Connor1996) [reply components/file_system/src/iosnoop/biosnoop.rs]: I take a look at the example of rust-bcc and it doesn't consider the alignment either. https://github.com/rust-bpf/rust-bcc/blob/master/examples/contextswitch.rs#L123 @brson Is there anything promising it implicitly?

Comment (yiwu-arbug) [reply components/file_system/src/iosnoop/biosnoop.rs]: it does look like the map value is obtained via `bpf_map_lookup_elem(fd, key, value)` call and the result is stored in an owned buffer from rust-bcc. I don't seems to find thread-safe guarantee of `bpf_map_lookup_elem` call. But anyway, let's remove the use of atomic here.

**Review by yiwu-arbug — APPROVED** [review]
LGTM on my side.

File: components/tikv_util/src/interval_driver.rs [hunk 6]
```diff
@@ -70,8 +69,13 @@ impl<K: KvEngine, R: RaftEngine> MetricsFlusher<K, R> {
...
         drop(self.sender.take().unwrap());
         if let Err(e) = h.unwrap().join() {
-            error!("join metrics flusher failed"; "err" => ?e);
-            return;
+            error!("join thread failed"; "err" => ?e, "name" => self.name.as_str());
         }
     }
 }
+
+pub trait IntervalRunnable: Send + Sync {
```
Connor1996, line 77: why not use `Runnable`?

components/tikv_util/src/interval_driver.rs
```diff
@@ -4,61 +4,60 @@ use std::io;
...
-use crate::engines::Engines;
-use crate::raft_engine::RaftEngine;
+use std::time::Duration;
 
 const DEFAULT_FLUSH_INTERVAL: Duration = Duration::from_millis(10_000);
-const FLUSHER_RESET_INTERVAL: Duration = Duration::from_millis(60_000);
 
-pub struct MetricsFlusher<K: KvEngine, R: RaftEngine> {
-    pub engines: Engines<K, R>,
+pub struct IntervalDriver {
```
Connor1996, line 11: I think `MetricsFlusher` is better. It's only used for flushing metrics. Other interval tasks are scheduled in their own worker.

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -146,52 +145,23 @@ pub fn get_io_type() -> IOType {
...
-
-                return self.io_stats_map.replace(map).unwrap();
+pub(crate) fn fetch_io_bytes(mut io_type: IOType) -> IOBytes {
+    unsafe {
+        if let Some(ctx) = BPF_CONTEXT.as_mut() {
+            let io_type_buf_ptr = &mut io_type as *mut IOType as *mut u8;
+            let mut io_type_buf =
+                std::slice::from_raw_parts_mut(io_type_buf_ptr, std::mem::size_of::<IOType>());
+            if let Ok(e) = ctx.stats_table.get(&mut io_type_buf) {
+                // TODO: the returned slice could be misaligned
```
Connor1996, line 155: If this cannot figure the misalignment, why change that instead of just removing the atomic load.

components/tikv_util/src/interval_driver.rs
tabokie: `Runnable` is a machine abstraction, `IntervalRunnable` is a task abstraction, I'll rename it to `IntervalTask` instead.

components/file_system/src/iosnoop/biosnoop.rs
tabokie: After reading [this](https://dwrensha.github.io/capnproto-rust/2020/01/11/unaligned-memory-access.html) and [this](https://stackoverflow.com/questions/42499049/transmuting-u8-buffer-to-struct-in-rust), it seems that `ptr::read` is vulnerable to misalignment, `transmute` or `T::from_ne_bytes` (which essentially uses transmute) is safe because they copy the value, `align_to` can be used to determine the origin slice's alignment. So @brson 's code is solid, we can remove the comment.

components/tikv_util/src/interval_driver.rs
tabokie: I don't really like it, because the class itself has no involvement in actual logic. If you think the name isn't friendly enough, it is actually only used in server.rs, all the task definitions are self-explanatory.

components/file_system/src/iosnoop/biosnoop.rs
Connor1996: Agree. This resolve the misalignment. But I prefer `ptr::read_unaligned(e.value.as_ptr() as *const IOStats)` which is neat.

components/tikv_util/src/interval_driver.rs
Connor1996: I mean move it into `server` instead of `tikv_util` and make `MetricsFlusher` which implements RunnableWithTimer or just keep the origin implementation.  Now the `IntervalDriver` is quite duplicated with `tikv_util::Worker`.

tabokie: We can't do that because metrics task are defined in submodules. I'll try the reuse to see if they look better.

Connor1996: You don't need metrics task in that way

components/file_system/src/iosnoop/biosnoop.rs
tabokie: Yah it's better, making the choice of method more explicit.

components/tikv_util/src/interval_driver.rs
tabokie: You mean fusing all those module logic into server.rs? Be aware that each metrics flushing routine has their own states, exposing them to server.rs would be chaotic.

tabokie: On a second thought, I decided to make some less generic wrapper for those states, so we can hardcode them into a runner in server.rs.

Connor1996: Yes, I mean the level of `server` mod, namely, a runner in `server.rs`

tabokie:
@Connor1996 The code is updated, but I'm not sure they look any better. The `DummyTask` workaround is kind of ugly. And BTW I'm not sure what's `LazyWorker` for, they look very similar to me.

cmd/src/lib.rs
```diff
@@ -6,5 +6,6 @@ extern crate tikv_util;
 #[macro_use]
 pub mod setup;
 pub mod dump;
+mod metrics_flusher;
```
Connor1996, line 9: I think it should be in server not cmd

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -145,18 +145,25 @@ pub fn get_io_type() -> IOType {
...
                 std::slice::from_raw_parts_mut(io_type_buf_ptr, std::mem::size_of::<IOType>());
             if let Ok(e) = ctx.stats_table.get(&mut io_type_buf) {
-                let (chunks, _) = e.as_slice().as_chunks();
-                assert!(chunks.len() == 2);
-                let read = u64::from_ne_bytes(chunks[0]) as i64;
-                let write = u64::from_ne_bytes(chunks[1]) as i64;
-                return IOBytes { read, write };
+                assert!(e.len() == std::mem::size_of::<RawStats>());
+                let raw_stats = std::ptr::read_unaligned(e.as_ptr() as *const RawStats);
+                return IOBytes {
```
Connor1996, line 163: Can we make IOBytes be u64, is there any case that the bytes would be negative?

components/tikv_util/src/interval_driver.rs
Connor1996, line 11: Can be deleted now

components/file_system/src/metrics_task.rs
```diff
@@ -41,20 +49,8 @@ impl MetricsTask {
...
-        let delta_bytes = bytes - $last_fetch;
-        $last_fetch = bytes;
-        IO_BYTES_VEC.$metrics.read.inc_by(delta_bytes.read);
-        IO_BYTES_VEC.$metrics.write.inc_by(delta_bytes.write);
-    };
-}
 
-impl IntervalTask for MetricsTask {
-    fn on_tick(&mut self) {
+    pub fn on_tick(&mut self) {
```
Connor1996, line 53: call it flush_metrics? and rename MetricsTask to something else?

Connor1996:
> @Connor1996 The code is updated, but I'm not sure they look any better. The `DummyTask` workaround is kind of ugly. And BTW I'm not sure what's `LazyWorker` for, they look very similar to me.

`DummyTask` is unavoidable and looks fine for me. `LazyWorker` is just a worker that can get the scheduler before calling start.

tabokie:
I made some updates to remove the ugly mocking. The only downside is exposing thread pool handle, which I think is fine because we legally own the pool.

cmd/src/server.rs
```diff
@@ -853,10 +882,19 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
-        );
+        let remote = self.background_worker.clone_yatp_handle();
+        let mut interval = GLOBAL_TIMER_HANDLE
+            .interval(Instant::now(), DEFAULT_METRICS_FLUSH_INTERVAL)
+            .compat();
+        let mut engine_task =
+            EngineMetricsTask::new(self.engines.as_ref().unwrap().engines.clone());
+        let mut io_task = IOMetricsTask::new(fetcher);
+        remote.spawn(async move {
+            while let Some(Ok(_)) = interval.next().await {
```
Connor1996, line 893: how can it stop?

tabokie: I am not sure. Can yatp stop itself when there are pending tasks? If it's a problem we can use a channel like before.

tabokie: Seems to me that `Worker` itself doesn't stop the channel either.

components/file_system/src/lib.rs
```diff
@@ -73,6 +77,30 @@ impl Drop for WithIOType {
...
+        IOBytes { read: 0, write: 0 }
+    }
+}
+
+impl std::ops::Sub for IOBytes {
+    type Output = Self;
+
+    fn sub(self, other: Self) -> Self::Output {
+        Self {
+            read: self.read - other.read,
```
Connor1996, line 98: use saturating_sub

components/file_system/src/rate_limiter.rs
```diff
@@ -57,25 +58,35 @@ impl BytesRecorder {
...
-        let bytes = std::cmp::min(self.refill_bytes.load(atomic::Ordering::Relaxed), bytes);
-        if let Some(recorder) = &self.recorder {
-            recorder.add(io_type, io_op, bytes);
+        let bytes = if self.refill_bytes > 0 {
+            std::cmp::min(self.refill_bytes, bytes)
+        } else {
+            bytes
+        };
+        if self.enable_statistics {
+            self.stats.add(io_type, io_op, bytes);
```
Connor1996, line 89: can we make the `BytesRecorder` to be a trait `{ fn add(...); fn fetch(...) }` and give an empty `fn add(...)` implementation for IOsnoop. So we don't need to pass the enable_statistics and later io rate limiter can utilize the stats from IOsnoop too.

tabokie: I appreciate your effort to unite the two, my concern is where do we actually need to ignore their differences. For metrics flusher, it's ok; but for IO rate limiter statistics part, it isn't necessary (one could even argue it's worse because a virtual function cal is unlikely to be elided); for IO rate limiter calibration part, the decision to utilize stats is based on whether it is from a external source (calibrate based on our own estimation is a waste of work), which means even if the interface is unified, a extra toggle is required.

tabokie: Also I think wrapping the whole "utilizing external stats" part into a separate object is slightly better than trying to reply on one shared statistics object: the calibration logic needs to maintain some more states than vanilla stats vec; and at this phrase we are really not sure if there will be any other fallback methods to choose from.

Connor1996: Okay, I didn't take a look at the calibration logic, so I was thinking it just needs to get the delta bytes periodical. But for the concern of dynamic dispatch, don't worry, you can utilize the https://docs.rs/enum_dispatch/0.3.5/enum_dispatch/ to convert it into static dispatch.

Connor1996 (approved):

PR #9482, opened 2021-01-10
engine_traits: Add delete_range (no cf) methods

brson:

The other two mutation methods, put and delete, have two versions each: one that takes a CF, and one that doesn't (it mutates CF_DEFAULT). The `delete_range_cf` method though has no corresponding `delete_range` method.

This PR adds the method.

This is not _actually_ needed by tikv - the one place it calls `delete_range_cf(CF_DEFAULT...)` makes sense to continue using the "cf" version of the method.

As a matter of API design though, these methods should exist.

But furthermore, the engine test suite I'm developing will be more self-consistent if the non-cf version of this method is available - various tests are parameterized over both the "cf" and "non-cf" versions of these mutation methods.

I've included no unit tests, but the engine_traits_test suite will include tests for it, and I intend to submit a PR for that in the next week or so.

Add `delete_range` methods to `SyncMutable` and `Mutable` traits in engine_traits.

### Release note 

- No release note.

### DIFF
--- a/components/engine_panic/src/engine.rs
+++ b/components/engine_panic/src/engine.rs
@@ -55,6 +55,9 @@ impl SyncMutable for PanicEngine {
     fn delete_cf(&self, cf: &str, key: &[u8]) -> Result<()> {
         panic!()
     }
+    fn delete_range(&self, begin_key: &[u8], end_key: &[u8]) -> Result<()> {
+        panic!()
+    }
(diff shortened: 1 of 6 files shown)


brson:
Rebased.

kennytm (commented):
/lgtm

brson:
@kennytm ti-chi-bot didn't recognize your lgtm. Can you try it again?

Issue #9466, opened 2021-01-08
Performance decline in TPCC bench due to the absence of GC


hunterlxt:
### What version of TiKV are you using?

master 7757bd416b5f087f8a67bb02fd827feb78376aa6

### What operating system and CPU are you using?

CentOS 3.10.0-1127.13.1.el7.x86_64

### Steps to reproduce

```
# 1TiDB/1PD/1TiKV
# default config

go-tpc tpcc -P 20000 --warehouses 1000 --parts 1 prepare
go-tpc tpcc -P 20000 -T 200 --warehouses 1000 --parts 1 run
```

### What did you expect?
Stable QPS

### What did happened?
Declining QPS

### Supplement
I found that `next` operation increased in `acquire_pessimistic_lock` which means there were more and more old versions. But, I checked `GC` panel and found there is no data in GC Speed. It might be a bug.

![image](https://user-images.githubusercontent.com/18193301/104012533-eae4ea80-51ea-11eb-9eff-ed0da2362f60.png)

![image](https://user-images.githubusercontent.com/18193301/104012453-c983fe80-51ea-11eb-8953-529ba8ca17fd.png)

![image](https://user-images.githubusercontent.com/18193301/104012646-25e71e00-51eb-11eb-9a16-65e7fa1e7c26.png)

![image](https://user-images.githubusercontent.com/18193301/104012756-4fa04500-51eb-11eb-82ae-204af1ad040f.png)

MyonKeminta:
According to the metrics, GC normally proceeds to the final distributed GC phase (compaction filter is not enabled).

![image](https://user-images.githubusercontent.com/9948422/104014505-1917f980-51ee-11eb-93d5-6cb1f689b5c2.png)


Then according to the log, auto gc founds no regions in the store.

```
[2021/01/08 18:41:43.776 +08:00] [INFO] [future.rs:146] ["starting working thread"] [worker=gc-worker]
[2021/01/08 18:41:43.871 +08:00] [ERROR] [gc_manager.rs:352] ["failed to get safe point from pd"] [err_code=KV:Storage:Unknown] [err="Other(\"[src/server/gc_worker/gc_worker.rs:56]: failed to get safe point from PD: Other(\\\"[components/pd_client/src/util.rs:298]: request retry exceeds limit\\\")\")"]
[2021/01/08 18:53:33.966 +08:00] [INFO] [gc_manager.rs:428] ["gc_worker: auto gc starts"] [safe_point=422078711892017152]
[...]

PR #9486, opened 2021-01-11
start gc after node to fetch node id

hicqu:

There is a bug in the master branch that GC can't work at all. The reason is `0` is passed into the GC worker as `store_id`, which will cause all regions get skiped when doing GC.

Close https://github.com/tikv/tikv/issues/9466.

The bug is introduced in https://github.com/tikv/tikv/pull/9350.

Start gc worker after node. So that `node_id` passed into the GC worker will must be valid.

### Check List 

Tests 
a) starting a new tikv cluster
b) run a sysbench update_index workload, then check the gc worker cpu panel
c) check the result:
  
图片
图片

### Release note 
* No release note

### DIFF
--- a/cmd/src/server.rs
+++ b/cmd/src/server.rs
@@ -636,6 +636,7 @@ impl<ER: RaftEngine> TiKVServer<ER> {
 
         let auto_split_controller = AutoSplitController::new(split_config_manager);
 
+        // `ConsistencyCheckObserver` must be registered before `Node::start`.
         let safe_point = Arc::new(AtomicU64::new(0));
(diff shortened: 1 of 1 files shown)


hunterlxt (approved):
LGTM

gengliqi (approved):

MyonKeminta (approved):
LGTM. Can you also test manually to ensure the problem is fixed?

hicqu:
@MyonKeminta I'm testing it. Result will be posted later.

BusyJay:
I remembered there was a benchmark result about compaction filter gc, was the result still correct?

Issue #8783, opened 2020-10-04
maybe_create_peer_internal fails due to hashmap entry not found


overvenus:
## Bug Report

Panic found when enables the random-merge scheduler.

```
[2020/10/04 08:27:59.846 +08:00] [FATAL] [lib.rs:481] ["no entry found for key"] [backtrace="stack backtrace:
   0: tikv_util::set_panic_hook::{{closure}}
             at components/tikv_util/src/lib.rs:480
   1: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:524
   2: rust_begin_unwind
             at src/libstd/panicking.rs:431
   3: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   4: core::option::expect_failed
             at src/libcore/option.rs:1261
   5: core::option::Option<T>::expect
             at /home/stn/.rustup/toolchains/nightly-2020-07-01-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/option.rs:344
      <std::collections::hash::map::HashMap<K,V,S> as core::ops::index::Index<&Q>>::index
             at /home/stn/.rustup/toolchains/nightly-2020-07-01-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/collections/hash/map.rs:1034
      raftstore::store::fsm::store::StoreFsmDelegate<EK,ER,T,C>::maybe_create_peer_internal
             at /home/stn/tikv/components/raftstore/src/store/fsm/store.rs:1728
      raftstore::store::fsm::store::StoreFsmDelegate<EK,ER,T,C>::maybe_create_peer
             at /home/stn/tikv/components/raftstore/src/store/fsm/store.rs:1671
      raftstore::store::fsm::store::StoreFsmDelegate<EK,ER,T,C>::on_raft_message
             at /home/stn/tikv/components/raftstore/src/store/fsm/store.rs:1603
   6: raftstore::store::fsm::store::StoreFsmDelegate<EK,ER,T,C>::handle_msgs
             at /home/stn/tikv/components/raftstore/src/store/fsm/store.rs:575
      <raftstore::store::fsm::store::RaftPoller<EK,ER,T,C> as batch_system::batch::PollHandler<raftstore::store::fsm::peer::PeerFsm<EK,ER>,raftstore::store::fsm::store::StoreFsm<EK>>>::handle_control
             at /home/stn/tikv/components/raftstore/src/store/fsm/store.rs:823
[...]

PR #8084, opened 2020-06-11
raftstore: handle the race between creating new peer and splitting correctly

gengliqi:
Signed-off-by: Liqi Geng <gengliqiii@gmail.com>

### What problem does this PR solve?

Problem Summary:

The new region which is created by splitting may has already created before in some sophisticated cases.

I tried to fix it by changing the split flow. close https://github.com/tikv/tikv/pull/7503
But it's too complicated and slower than before.
With @NingLin-P's help, this new method for this PR is to add a new `pending_create_peers`. It has lower performance loss than the previous fix PR.

### What is changed and how it works?

What's Changed:
Add `StoreMeta` to ApplyContext to handle the race of creating peer and splitting.

The test will be added.

### Related changes

- Need to cherry-pick to the release branch

Tests 

- Integration test

Side effects

No.

### Release note 
* Fix an issue that split may panic and corrupt the metadata if the split process is too slow and region merge is open.

PR #9495, opened 2021-01-13
raftstore: fix the region meta inconsistency bug during splitting

gengliqi:

Issue Number: close https://github.com/tikv/tikv/issues/8783

Problem Summary:
https://github.com/tikv/tikv/issues/8783#issuecomment-708298415
The bug is located in function `on_ready_split_region` which is introduced by https://github.com/tikv/tikv/pull/8084.
https://github.com/tikv/tikv/blob/7c9850040ba58d48f0f1a93345c0392dffb2fc5b/components/raftstore/src/store/fsm/peer.rs#L2084-L2112
Line 2111 `continue` means this region has already existed so it should be skipped.
But line 2087-2090 has already added this region to `region_ranges` which leads to metadata inconsistency.

What's Changed:
Fix this bug and add 2 tests.

Run `test_split_not_to_split_existed_tombstone_region ` and `test_split_not_to_split_existed_different_uninitialied_peer` can panic in different place in previous code.

### Related changes

### Check List 

Tests 

- Integration test

Side effects

- No

### Release note 
* No release note


tests/failpoints/cases/test_split_region.rs
```diff
@@ -248,7 +248,7 @@ impl Filter for PrevoteRangeFilter {
 // region id has already existed. In previous implementation, it can be created and panic
 // will happen because there are two initialized peer with the same region id.
 #[test]
-fn test_split_not_to_split_exist_region() {
+fn test_split_not_to_split_existed_region() {
```
BusyJay, line 251: s/existed/existing/

BusyJay (commented):
Rest LGTM

tests/failpoints/cases/test_split_region.rs
```diff
@@ -378,8 +378,126 @@ fn test_split_not_to_split_exist_tombstone_region() {
...
+    cluster.must_split(&region, b"k2");
+    let left = pd_client.get_region(b"k1").unwrap();
+    assert_eq!(left.get_id(), 1000);
+    let left_peer_2 = find_peer(&left, 2).cloned().unwrap();
+
+    pd_client.must_remove_peer(left.get_id(), left_peer_2);
+    pd_client.must_add_peer(left.get_id(), new_peer(2, 4));
+
+    let before_check_snapshot_1000_2_fp = "before_check_snapshot_1000_2";
+    fail::cfg(before_check_snapshot_1000_2_fp, "pause").unwrap();
```
NingLin-P, line 485: How about setting the fail point before splitting?

NingLin-P (commented):
Rest LGTM

tests/failpoints/cases/test_split_region.rs
gengliqi: There is no difference because there is a send filter of store 2.

NingLin-P (approved):

BusyJay (approved):

Issue #9395, opened 2020-12-30
Wrong available space calculation


gengliqi:
### What version of TiKV are you using?
3.0, 3.1, 4.0, master

### What operating system and CPU are you using?
doesn't matter

### Steps to reproduce
The available space which TiKV calculates is wrong.

### What did you expect?
Make it right.

Since linux will reserve some space to avoid filesystem fragmentation(default 5%), we should calculate available space by using the `available_space` of `FsStat` instead of `total_space` or `free_space`.
https://github.com/tikv/tikv/blob/15282cb2c7c67c7cff4b7bf81cba1d0e7075a5be/components/raftstore/src/store/worker/pd.rs#L652-L689

![image](https://user-images.githubusercontent.com/21253569/103338295-e3ba3000-4ab8-11eb-8e33-39a15f601d48.png)
Note that `3737671118848/1024 = 3650069452`

PR #9497, opened 2021-01-13
reserve space with a better strategy

hicqu:

Currently TiKV reserves 2GB disk space to ensure recovering from `no space left` error must success. It's not enough exactly.

And, `available_space` reported to PD is wrong.

Issue Number: close https://github.com/tikv/tikv/issues/9395.

TiKV reserve 5GB space at least and 20GB at most. The lower limit can be adjusted in config file.

### Check List 

Tests 

  a. Start a TiKV instance without the patch
  b. Stop it and switch to the branch, restart it
  c. size of file `space_placeholder_file` increases larger

### Release note 
* server: reserve space with a better strategy

### DIFF
--- a/cmd/src/server.rs
+++ b/cmd/src/server.rs
@@ -70,7 +70,7 @@ use tikv::{
     },
     storage::{
         self,
-        config::{StorageConfigManger, MAX_RESERVED_SPACE_SIZE},
+        config::{StorageConfigManger, MAX_RESERVED_SPACE_GB},
         mvcc::MvccConsistencyCheckObserver,
     },
 };
@@ -376,10 +376,10 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         file_system::reserve_space_for_recover(
             &self.config.storage.data_dir,
             cmp::min(
-                ReadableSize::gb(MAX_RESERVED_SPACE_SIZE).0,
+                ReadableSize::gb(MAX_RESERVED_SPACE_GB).0,
                 // Max one of configured `reserve_space` and `storage.capacity * 5%`.
                 cmp::max(
-                    (capacity as f64 / 20f64) as u64,
+                    (capacity as f64 * 0.05) as u64,
                     self.config.storage.reserve_space.0,
                 ),
             ),
--- a/cmd/src/server.rs
+++ b/cmd/src/server.rs
@@ -9,6 +9,7 @@
 //! We keep these components in the `TiKVServer` struct.
 
 use std::{
+    cmp,
     convert::TryFrom,
     env, fmt,
     fs::{self, File},
@@ -67,9 +68,13 @@ use tikv::{
         status_server::StatusServer,
         Node, RaftKv, Server, CPU_CORES_QUOTA_GAUGE, DEFAULT_CLUSTER_ID, GRPC_THREAD_PREFIX,
     },
-    storage::{self, config::StorageConfigManger, mvcc::MvccConsistencyCheckObserver},
+    storage::{
+        self,
+        config::{StorageConfigManger, MAX_RESERVED_SPACE_SIZE},
+        mvcc::MvccConsistencyCheckObserver,
+    },
 };
-use tikv_util::config::VersionTrack;
+use tikv_util::config::{ReadableSize, VersionTrack};
 use tikv_util::{
     check_environment_variables,
     config::ensure_dir_exist,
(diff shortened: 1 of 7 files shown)


src/storage/config.rs
```diff
@@ -17,13 +17,18 @@ const DEFAULT_GC_RATIO_THRESHOLD: f64 = 1.1;
...
 // According to "Little's law", assuming you can write 100MB per
 // second, and it takes about 100ms to process the write requests
 // on average, in that situation the writing bytes estimated 10MB,
 // here we use 100MB as default value for tolerate 1s latency.
 const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
 
+const DEFAULT_RESERVED_SPACE_SIZE: u64 = 5;
+// 20GB for reserved space is enough because size of one compaction is limited,
+// generally less than 2GB.
+pub const MAX_RESERVED_SPACE_SIZE: u64 = 20;
```
tabokie, line 30:
```suggestion
const MIN_RESERVED_SPACE_GB: u64 = 5;
// 20GB for reserved space is enough because size of one compaction is limited,
// generally less than 2GB.
pub const MAX_RESERVED_SPACE_GB: u64 = 20;
```

cmd/src/server.rs
```diff
@@ -361,11 +366,23 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+            capacity = cmp::min(capacity, self.config.raft_store.capacity.0);
+        }
         file_system::reserve_space_for_recover(
             &self.config.storage.data_dir,
-            self.config.storage.reserve_space.0,
+            cmp::min(
+                ReadableSize::gb(MAX_RESERVED_SPACE_SIZE).0,
+                // Max one of configured `reserve_space` and `storage.capacity * 5%`.
+                cmp::max(
+                    (capacity as f64 / 20f64) as u64,
```
tabokie, line 382: maybe `* 0.05` is more readable?

src/storage/config.rs
```diff
@@ -43,6 +48,8 @@ pub struct Config {
     pub scheduler_pending_write_threshold: ReadableSize,
     #[config(skip)]
     // Reserve disk space to make tikv would have enough space to compact when disk is full.
+    // Values less than `5GB` will be sanitized to `5GB` (2GB for `max_compaction_bytes`, and
```
tabokie, line 51: What if `max_compaction_bytes` is altered?

hicqu: Seems no better solutions except explain it in config file?

etc/config-template.toml
```diff
@@ -209,6 +209,13 @@
 ## latency jittering when apply duration is not stable.
 # enable-async-apply-prewrite = false
 
+## Reserve some space to ensure recovering the store from `no space left` must success. The
+## configured value only indicates the lower limit, and 20GB is the upper limit. In this range,
+## `max(reserve-space, capacity * 5%)` will be reserved exactly.
+##
+## Generally it's not necessary to change it. This configuration could be deprecated in future.
+# reserve-space = "5GB"
```
tabokie, line 217: Add this in test-custom.toml too?

tabokie (approved):

src/storage/config.rs
tabokie: Yes, 5G should be safe enough.

gengliqi (commented):
LGTM. I suggest adding a switch to control the behavior of clearing the tombstone regions' data because the `delete_file_in_range` may leave much data.

src/storage/config.rs
```diff
@@ -17,13 +17,18 @@ const DEFAULT_GC_RATIO_THRESHOLD: f64 = 1.1;
...
 // According to "Little's law", assuming you can write 100MB per
 // second, and it takes about 100ms to process the write requests
 // on average, in that situation the writing bytes estimated 10MB,
 // here we use 100MB as default value for tolerate 1s latency.
 const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
 
+const MIN_RESERVED_SPACE_GB: u64 = 5;
+// 20GB for reserved space is enough because size of one compaction is limited,
+// generally less than 2GB.
+pub const MAX_RESERVED_SPACE_GB: u64 = 20;
```
zhangjinpeng1987, line 30: I don't think it is enough for a 4TB RocksDB instance

src/storage/config.rs
```diff
@@ -17,13 +17,18 @@ const DEFAULT_GC_RATIO_THRESHOLD: f64 = 1.1;
...
 const MAX_SCHED_CONCURRENCY: usize = 2 * 1024 * 1024;
-const DEFAULT_RESERVER_SPACE_SIZE: u64 = 2;
+
 // According to "Little's law", assuming you can write 100MB per
 // second, and it takes about 100ms to process the write requests
 // on average, in that situation the writing bytes estimated 10MB,
 // here we use 100MB as default value for tolerate 1s latency.
 const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
 
+const MIN_RESERVED_SPACE_GB: u64 = 5;
```
zhangjinpeng1987, line 27: How about a TiKV with 10GB disk?

src/storage/config.rs
hicqu: It's decided based on `max compaction size`, which of the default value is `2GB`. And `max-background-jobs` generally is less than 10 or 12, and some background threads are reserved for flush. So I choose `2GB * 10` as the max limit.

src/storage/config.rs
hicqu: In this case users need to set `reserve-space` to a less value. Maybe we can remove the configuration from config.toml, leave it self-adaptable.

src/storage/config.rs
zhangjinpeng1987: 2GB is not a hard limit.

PR #9500, opened 2021-01-14
*: Update readme for clients link

andylokandy:
Signed-off-by: Andy Lok <andylokandy@hotmail.com>


README.md
```diff
@@ -100,16 +100,13 @@ For the production environment, use [TiDB Ansible](https://github.com/pingcap/ti
 
 ## Client drivers
 
-Currently, the interfaces to TiKV are the [TiDB Go client](https://github.com/pingcap/tidb/tree/master/store/tikv) and the [TiSpark Java client](https://github.com/pingcap/tispark/tree/master/tikv-client/src/main/java/com/pingcap/tikv).
-
 These are the clients for TiKV:
 
-- [Go](https://github.com/tikv/client-go) (The most stable and widely used)
+- [Go](https://tikv.org/docs/4.0/reference/clients/go/) (The most stable and widely used)
```
nrc, line 105: Is it really stable and widely used? I thought it was out of date and not very good which is why we're re-doing it

andylokandy: It's true that tikv client in tidb is the most widely used. (still not much tho...)

nrc: iiuc, there is the client inside TiDB which is stable and most used, and there is the standalone go client which is not widely used and that we are planning to replace?

andylokandy: that's correct!

PR #9501, opened 2021-01-14
test: fix test_pd_client_deadlock

youjiali1995:

Issue Number: close https://github.com/tikv/tikv/issues/9470

Problem Summary:

`region_heartbeat` in the `test_pd_client_deadlock` case sometimes hangs forever.

What's Changed:

`region_heartbeat` interface in the PD client creates a stream when reconnecting to the PD leader which means `block_on`ing it can't return until reconnection. It results in the unstable test. It's changed to return a future instead of a stream which is more intuitive.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
* No release note

### DIFF
--- a/tests/failpoints/cases/test_pd_client.rs
+++ b/tests/failpoints/cases/test_pd_client.rs
@@ -84,14 +84,14 @@ fn test_pd_client_deadlock() {
             func();
             tx.send(()).unwrap();
         });
+        // Only allow to reconnect once for a func.
+        client.handle_reconnect(move || {
+            fail::cfg(leader_client_reconnect_fp, "return").unwrap();
+        });
         // Remove the fail point to let the PD client thread go on.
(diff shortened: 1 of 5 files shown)


tests/failpoints/cases/test_pd_client.rs
```diff
@@ -84,14 +84,14 @@ fn test_pd_client_deadlock() {
             func();
             tx.send(()).unwrap();
         });
+        // Only allow to reconnect once for a func.
```
sticnarf, line 87: What's it for?

youjiali1995: The reconnection interval is 100ms. It's possible the test becomes fragile due to frequent reconnection. I ban it after the first reconnection.

sticnarf (commented):
/LGTM

MyonKeminta (approved):

PR #9502, opened 2021-01-14
*: Add more failpoints about transaction

MyonKeminta:

Problem Summary: This PR tries to add more failpoints as well as enhance existed failpoints about transaction.

What's Changed: This PR tries to add more failpoints as well as enhance existed failpoints about transaction.

### Related changes

  - Maybe can be partially picked to release-4.0

### Check List 

Tests 

- Unit test

Side effects

-

### Release note 

- No release note


MyonKeminta:
So many redundant labels 🤔

src/server/gc_worker/applied_lock_collector.rs
```diff
@@ -131,28 +131,28 @@ impl LockObserver {
...
-            };
-            send_fp();
-        }
+        let injected_full = (|| {
+            fail_point!("lock_observer_send_full", |_| {
+                info!("[failpoint] injected lock observer channel full"; "locks" => ?locks);
+                true
+            });
+            false
+        })();
```
MyonKeminta, line 140: This kind of thinks produces annoying compiler warnings..

sticnarf:
Maybe only the `component/bench-test` label is valid? The others don't even have a description.

sticnarf (approved):
LGTM

sticnarf:
PTAL @youjiali1995

components/concurrency_manager/src/lib.rs
```diff
@@ -10,6 +10,9 @@
 //! In order to mutate the lock of a key stored in the lock table, it needs
 //! to be locked first using `lock_key` or `lock_keys`.
 
+#[macro_use(fail_point)]
+extern crate fail;
```
sticnarf, line 14: I prefer simple `use fail::failpoint` for each file which uses the macro.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -36,11 +36,13 @@ use crate::store::{CasualMessage, PeerMsg, RaftCommand, RaftRouter, StoreMsg};
 
 use collections::HashMap;
 use concurrency_manager::ConcurrencyManager;
+use futures::FutureExt;
 use pd_client::metrics::*;
 use pd_client::{Error, PdClient, RegionStat};
 use tikv_util::metrics::ThreadInfoStatistics;
 use tikv_util::time::UnixSecs;
 use tikv_util::worker::{FutureRunnable as Runnable, FutureScheduler as Scheduler, Stopped};
+use tokio::time::delay_for;
```
sticnarf, line 45: Tokio's timer needs it's own reactor. I'm afraid it doesn't work here. It would be a better choice to use `GLOBAL_TIMER_HANDLE`.

youjiali1995 (commented):
Any plan to use these failpoints?

sticnarf:
@youjiali1995 @MyonKeminta provided presets of failpoints for testing async commit and those failpoints contain them.

MyonKeminta:
> Any plan to use these failpoints?

I think these can be used in random tests for checking correctness.
...and I think I still need to think if these stuff can really protect us against the bugs like those we found recently..

zyguan:
> Any plan to use these failpoints?

We plan to run tiacid, jepsen and various bank-like cases with these failpoints enabled this week.

sticnarf (approved):
DCO is broken...

MyonKeminta:
Sorry

MyonKeminta:
@youjiali1995 PTAL thx

components/raftstore/src/store/worker/pd.rs
```diff
@@ -992,7 +995,22 @@ where
...
+        #[cfg(feature = "failpoints")]
+        let delay = (|| {
+            fail_point!("delay_update_max_ts", |_| true);
+            false
+        })();
+        #[cfg(not(feature = "failpoints"))]
+        let delay = false;
+
+        if delay {
+            info!("[failpoint] delay update max ts for 1s"; "region_id" => region_id);
```
youjiali1995, line 1008: Can it print a lot? What's the usage of this failpoint, long time tests?

MyonKeminta: I think this will be triggered when region split ,merge or leader transfer happens.
This failpoint is expected to delays updating the max_ts in these cases, which makes MaxTsNotSynced error more likely to happen.

youjiali1995: I concern this one.
```rust
            fail_point!("async_commit_1pc_force_fallback", |_| {
                info!("[failpoint] injected fallback for async commit/1pc transaction"; "start_ts" => start_ts);
                true
            });
            false
```

MyonKeminta: Do  you think it's unnecessary? I don't have any strong opinion about whether it should be removed, just thinking it might help us debugging when something went  wrong during tests, since it tells us what failpoint the transaction has encountered. There's also similar logs in many of TiDB's failpoints. 😕

youjiali1995 (approved):

PR #9511, opened 2021-01-15
raftstore: check valid leader before sleeping 

### LINKED ISSUE #9505: Leader can be missing for a long time when joint consensus and hibernate regions are both enabled

Joint consensus can make a leader step down as a learner during conf change. And if hibernate region is enabled, the leader steps down, there can be a long time that there is no leader in the region until requests are sent to followers or hibernate timeout. It's expected client should try other peers when hibernate region is enabled, so this is not a bug in my opinion. But currently TiDB can't handle it well, see also pingcap/tidb#22400.

We can make it hardly happen by stop followers from hibernating by checking if leader is not a learner in current configuration.

BusyJay:

Issue Number: close #9505

What's Changed:

If leader is demoted and the region is hibernated, there can be no
leader for quite a long time. It's by design that client should try
other peer when leader is not available, but currently ticlient can't
handle the issue well. On the other hand, PD can still keep sending
scheduling command to the demoted leader, which will also timeout.

So this PR adds an additional check before sleeping to get around it.
This should fix the pd scheduling issue, but ticlient can still timeout
if demoted peer is restarted, client should still retry other peers when
necessary.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- No release note.

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -898,12 +898,25 @@ where
             // Checking term to make sure campaign has finished and the leader starts
             // doing its job, it's not required but a safe options.
             state != GroupState::Chaos
-                && self.raft_group.raft.leader_id != raft::INVALID_ID
+                && self.has_valid_leader()
                 && self.raft_group.raft.raft_log.last_term() == self.raft_group.raft.term
                 && !self.has_unresolved_reads()
                 // If it becomes leader, the stats is not valid anymore.
                 && !self.is_leader()
         }
+    }
+
+    #[inline]
+    pub fn has_valid_leader(&self) -> bool {
+        if self.raft_group.raft.leader_id == raft::INVALID_ID {
+            return false;
+        }
+        for p in self.region().get_peers() {
+            if p.get_id() == self.raft_group.raft.leader_id && p.get_role() != PeerRole::Learner {
+                return true;
+            }
+        }
+        false
     }
 
     /// Pings if followers are still connected.
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -733,7 +733,8 @@ impl<T: Simulator> Cluster<T> {
     ) -> RaftCmdResponse {
         let timer = Instant::now();
         let mut tried_times = 0;
-        while tried_times < 10 || timer.elapsed() < timeout {
+        // At least retry once.
+        while tried_times < 2 || timer.elapsed() < timeout {
             tried_times += 1;
             let mut region = self.get_region(key);
             let region_id = region.get_id();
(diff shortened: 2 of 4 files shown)


components/test_raftstore/src/cluster.rs
```diff
@@ -733,7 +733,8 @@ impl<T: Simulator> Cluster<T> {
     ) -> RaftCmdResponse {
         let timer = Instant::now();
         let mut tried_times = 0;
-        while tried_times < 10 || timer.elapsed() < timeout {
+        // At least retry once.
+        while tried_times < 2 || timer.elapsed() < timeout {
```
NingLin-P, line 737: Looks like it is `at most retry once` (if the timer does not fire first), also why we need this change?

tests/integrations/raftstore/test_hibernate.rs
```diff
@@ -410,3 +410,89 @@ fn test_hibernate_feature_gate() {
...
+    for id in 1..=2 {
+        cluster.add_send_filter(CloneFilterFactory(
+            RegionPacketFilter::new(r, id)
+                .msg_type(MessageType::MsgRequestPreVote)
+                .direction(Direction::Send),
+        ));
+    }
+    // Leave joint.
+    cluster.async_exit_joint(r).unwrap();
+    // Ensure peer 3 can campaign.
```
NingLin-P, line 452: Can we just transfer leadership to peer 3 then leave joint, while leaving the joint state, tikv will not check and reject leave joint if the leader will be demoted.

components/test_raftstore/src/cluster.rs
BusyJay: Without the patch, it can retry at least 10 times. Some operation inside the loop can take more than 5 seconds, so the method can wait for 50 seconds, which is too long in the test of hibernate regions and exceeds hibernate timeout eventually.

tests/integrations/raftstore/test_hibernate.rs
BusyJay: > ...tikv will not check and reject leave joint if the leader will be demoted.

Then it should be a bug.

components/test_raftstore/src/cluster.rs
NingLin-P: Perhaps we should change `||` to `&&`?

BusyJay: Using `||` is OK. It's possible that the first try stuck for a while in TiKV and fail. Make it try another time in this case should make cases stable.

NingLin-P: I mean keeping the try times at 10 and use `||`.

tests/integrations/raftstore/test_hibernate.rs
NingLin-P: https://github.com/tikv/tikv/pull/9533 add check to prevent it, PTAL

components/raftstore/src/store/peer.rs
```diff
@@ -898,14 +898,27 @@ where
...
         }
     }
 
+    #[inline]
+    pub fn has_valid_leader(&self) -> bool {
+        if self.raft_group.raft.leader_id == raft::INVALID_ID {
+            return false;
+        }
+        for p in self.region().get_peers() {
+            if p.get_id() == self.raft_group.raft.leader_id && p.get_role() != PeerRole::Learner {
```
gengliqi, line 915: How about `p.get_role() == PeerRole::Voter`?

gengliqi (commented):
Rest LGTM

components/raftstore/src/store/peer.rs
BusyJay: It can still go to sleep if leader is incoming voter or outgoing voter.

gengliqi (approved):

NingLin-P (approved):

PR #9514, opened 2021-01-18
Storage: refactor reader

nrc:

The mvcc::reader module shows up a lot in profiles, but is pretty crufty. MvccTxn is a bit of a dumping ground for anything vaguely transaction related.

This PR attempts to tidy things up a bit in the reader module and MvccTxn. Mostly this is simple code motion and removing some dead code. There are a few interesting bits and pieces though:

* I create a SnapshotReader abstraction which represents a reader at a given timestamp. It takes some responsibility from MvccTxn and MvccReader.
* I remove the check for locks in MvccReader::get and the isolation level from MvccReader. The isolation level is only used in the check for locks. `get` is only called by acquire_pessimistic_lock and in tests, in non-test code it is always called with `skip_lock_check = true` so the locks are never checked and the isolation level is irrelevant.
* I simplify the rollback collapse logic a little (though it is still a mess). It is worth closely checking these changes, I'm not confident in them. The gist is that in some places the code for checking whether a rollback should be collapsed can be simplified, in some cases by only having a single option to collapse or not, rather than an option to protect rollbacks and an option to collapse rollbacks.

PTAL @longfangsong @MyonKeminta 

The commits are relatively organised and it is probably easier to review this PR one commit at a time.

### Release note 

* No release note.


longfangsong:
I think you need to update the tests in `storage::mvcc::reader` to make it work with the new iterface of the readers.

src/storage/mvcc/reader/reader.rs
```diff
@@ -1,77 +1,78 @@
...
-    /// `commit_ts` of another transaction on this key.
-    OverlappedRollback { commit_ts: TimeStamp },
+use txn_types::{Key, Lock, OldValue, TimeStamp, Value, Write, WriteRef, WriteType};
+
+/// Read from an MVCC snapshot, i.e., a logical view of the database at a specific timestamp (the
+/// start_ts).
+///
+/// This represents the view of the database from a single transaction.
+///
+/// Confusingly, there are two meanings of the word 'snapshot' here. In the name of the struct,
```
longfangsong, line 18: How about make `SnapshotReader` into `MvccSnapshotReader` to make it clear.
Or we can also make `Snapshot` into `EngineSnapshot`

nrc: I made the second change. I feel that the `MvccSnapshotReader` us a bit long

nrc:
rebased and comment addressed, PTAL @longfangsong @MyonKeminta

src/storage/txn/actions/tests.rs
```diff
@@ -417,9 +427,16 @@ pub fn must_rollback<E: Engine>(engine: &E, key: &[u8], start_ts: impl Into<Time
     let snapshot = engine.snapshot(Default::default()).unwrap();
     let start_ts = start_ts.into();
     let cm = ConcurrencyManager::new(start_ts);
-    let mut txn = MvccTxn::new(snapshot, start_ts, true, cm);
-    txn.collapse_rollback(false);
-    txn::cleanup(&mut txn, Key::from_raw(key), TimeStamp::zero(), false).unwrap();
```
longfangsong, line 422: Now `must_rollback` and `must_rollback_collapsed` are identical, we can keep just one of them.
However some tests are based on non-`collapse_rollback` version of this function, I'm trying to do something.

src/storage/txn/actions/cleanup.rs
```diff
@@ -33,11 +33,21 @@ pub fn cleanup<S: Snapshot>(
...
-            txn.check_write_and_rollback_lock(key, lock, is_pessimistic_txn)
+            rollback_lock(
+                txn,
+                reader,
+                key,
+                lock,
+                is_pessimistic_txn,
+                !protect_rollback,
+                overlapped_write,
+            )
```
longfangsong, line 46: I'm not very sure about this.

The origin behaviour here by `check_write_and_rollback_lock` is:
```rust
TxnCommitRecord::None => rollback_lock
TxnCommitRecord::SingleRecord => if write is not WriteType::Rollback panic!
			         else unlock_key
TxnCommitRecord::OverlappedRollback => unlock_key
```

But the new behaviour here is:
```rust
TxnCommitRecord::None => rollback_lock
_ => panic!
```

Did I miss anything?

sticnarf: I think the change is from my latest PR https://github.com/tikv/tikv/pull/9656, and this PR is not updated yet.

sticnarf (commented):
In practice, `collapse_rollback` is always enabled. Maybe we can remove this option?

sticnarf:
The rest lgtm.

nrc:
> In practice, collapse_rollback is always enabled. Maybe we can remove this option?

This is not the case when rollback_lock is called via cleanup

longfangsong (approved):
/lgtm (finally!)

nrc:
@sticnarf review comments addressed, rebased, and tests passing. PTAL.

src/server/gc_worker/mod.rs
```diff
@@ -13,8 +13,178 @@ use compaction_filter::{
...
 pub use gc_manager::AutoGcConfig;
 pub use gc_worker::{sync_gc, GcSafePointProvider, GcTask, GcWorker, GC_MAX_EXECUTING_TASKS};
+use txn_types::TimeStamp;
 
 #[cfg(any(test, feature = "failpoints"))]
 pub use compaction_filter::test_utils::{gc_by_compact, TestGCRunner};
+
+// Returns true if it needs gc.
+// This is for optimization purpose, does not mean to be accurate.
+pub fn check_need_gc(safe_point: TimeStamp, ratio_threshold: f64, props: &MvccProperties) -> bool {
```
sticnarf, line 26: This function seems unnecessary to be pub now.

sticnarf (commented):
/lgtm

sticnarf (approved):

PR #9515, opened 2021-01-18
cdc: fix old value config glitch when changefeeds with different settings connect to one region

liuzix:

Problem Summary:
When two TiCDC changefeeds connect to the same region, but one changefeed requires old value but the other doesn't, the one that doesn't need old value still receives old value.

What's Changed: Now we record the old value setting in each `Downstream` and filters the old value appropriately before sinking the events. 

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
- Fix old value config glitch when changefeeds with different settings connect to one region

### DIFF
--- a/components/cdc/tests/integrations/test_cdc.rs
+++ b/components/cdc/tests/integrations/test_cdc.rs
@@ -924,15 +924,12 @@ fn test_old_value_multi_changefeeds() {
     suite.must_kv_prewrite(1, vec![m2], k1.clone(), ts3);
     let ts4 = block_on(suite.cluster.pd_client.get_tso()).unwrap();
     suite.must_kv_commit(1, vec![k1], ts3, ts4);
-    println!("test_old_value_multi_changefeeds: 2");
     let mut event_count = 0;
     loop {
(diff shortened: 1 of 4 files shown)


components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -890,6 +890,96 @@ fn test_old_value_basic() {
...
+    m2.value = vec![b'3'; 5120];
+    let ts3 = block_on(suite.cluster.pd_client.get_tso()).unwrap();
+    suite.must_kv_prewrite(1, vec![m2], k1.clone(), ts3);
+    let ts4 = block_on(suite.cluster.pd_client.get_tso()).unwrap();
+    suite.must_kv_commit(1, vec![k1], ts3, ts4);
+    println!("test_old_value_multi_changefeeds: 2");
+    let mut event_count = 0;
+    loop {
+        let events = receive_event_1(false).events.to_vec();
+        println!("test_old_value_multi_changefeeds: 3");
```
5kbpers, line 931: Please remove these `println`

amyangfei (commented):
LGTM

Issue #9327, opened 2020-12-22
Can not change region-split-check-diff through config file


rleungx:
## Bug Report
region-split-check-diff is exposed to the users, but it can not be changed through the config file. 
See 
https://github.com/tikv/tikv/blob/4251f091297cd763ddf9679efca3a28ebe1beffa/etc/config-template.toml#L279-L283
and
https://github.com/tikv/tikv/blob/4251f091297cd763ddf9679efca3a28ebe1beffa/src/config.rs#L2362-L2368

### What version of TiKV are you using?
master (3b2c5337c837b8bf9d37f954a4c17a407e253251)

### Steps to reproduce

1. change the `region-split-check-diff` to `10MB`
2. restart the tikv cluster

### What did you expect?
The value of `region-split-check-diff` should be `10MB` 

### What did happened?
The value of `region-split-check-diff` is still `6MB`

PR #9525, opened 2021-01-19
allow to custom region-split-check-diff

hicqu:

Issue Number: close https://github.com/tikv/tikv/issues/9327.

### Check List 

Tests 
- Unit test

Side effects

### Release note 
* No release note.


src/config.rs
```diff
@@ -3631,4 +3629,12 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_validate_tikv_config() {
+        let mut cfg = TiKvConfig::default();
+        cfg.raft_store.region_split_check_diff = ReadableSize::mb(20);
```
gengliqi, line 3636: How about setting it to `cfg.raft_store.region_split_check_diff + 1`? Thus we can ensure this value is different from the original one.

gengliqi (commented):
Rest LGTM

rleungx (approved):

gengliqi (approved):
LGTM

PR #9527, opened 2021-01-19
raftstore: fix local reader delegate keeps stale id

### LINKED ISSUE #9517: Local reader delegate keeps stale ID
Nightly 95951d04e6e8c0bb267c6a5591fb336ed5483b41

### Steps to reproduce
Deploy a cluster with at least four TiKV nodes and enable shuffle scheduling.
Kill a node randomly and check if all tables can be accessed and then restart it.

### What did you expect?
The operation can be repeated infinitely.

### What did happened?
TiDB keeps failing with error region unavailable.

The detail error is "mismatch peer id 4519007 != 4513038". From the code we can see that 4513038 is the delegate id.

https://github.com/tikv/tikv/blob/d9405c71878f70a315ee7550764ccfcaf324d0ef/components/raftstore/src/store/util.rs#L334-L345

And from the TiKV log, 4513038 is removed and then 4519007 is readded on the same node. So there should be bugs that delegate ID is not updated correctly.

NingLin-P:

Issue Number: close #9517 

Problem Summary:

After applying snapshot the `ReadDelegate` stored at `StoreMeta` will be replaced with the new one, but the removed one didn't call `mark_invalid` to notify the `ReadDelegate` store at `LocalReader` to remove itself and cause the `LocalReader` always return ` mismatch peer id`

https://github.com/tikv/tikv/blob/35c66d38ff8180f3652643e35575590d9ab57fc1/components/raftstore/src/store/peer.rs#L1750-L1755

What's Changed:

People can easily forget to call `mark_invalid` when a `ReadDelegate` is removed from `StoreMeta`, instead, we can call `mark_invalid` when the `ReadDelegate` (the one stored at `StoreMeta`) is dropped.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- raftstore: fix local reader delegate keeps stale id


hicqu:
/LGTM.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

PR #9529, opened 2021-01-20
resolver: untracking locks by key

### LINKED ISSUE #9523: Resolved TS is not updated with async commit transactions

Master (35c66d38ff8180f3652643e35575590d9ab57fc1)

### Steps to reproduce

Enable failpoints in TiDB (with unmerged https://github.com/pingcap/tidb/pull/22431):

```
github.com/pingcap/tidb/store/tikv/asyncCommitDoNothing=50%return
github.com/pingcap/tidb/store/tikv/runTTLManager=return
github.com/pingcap/tidb/store/tikv/shortPessimisticLockTTL=return
github.com/pingcap/tidb/store/tikv/twoPCShortLockTTL=return
```

Set the interval of updating PD physical TS to 2s.

Run tipocket/pbank with async commit and 1PC enabled:

```
./pbank -case=multi_bank -prepare-sql='set global tidb_enable_async_commit = 1;set global tidb_enable_1pc = 1;'
```

### What did you expect?

CDC works well.

### What did happened?

From the monitor, we see some TiKV failed to update the resolved TS, so CDC didn't work.

image
image

sticnarf:

Issue Number: close #9523

Let the resolver maintain a HashMap that maps lock key to start_ts. Then, when we untrack a lock, we only need the key.

### Check List 

Tests 

- Unit test
- Manual test using the reproducing steps in #9523

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

- Fix that CDC's resolved_ts may not advance when using async commit.


components/cdc/src/delegate.rs
```diff
@@ -394,11 +389,7 @@ impl Delegate {
         for lock in pending.take_locks() {
             match lock {
                 PendingLock::Track { key, start_ts } => resolver.track_lock(start_ts, key),
-                PendingLock::Untrack {
-                    key,
-                    start_ts,
-                    commit_ts,
-                } => resolver.untrack_lock(start_ts, commit_ts, key),
+                PendingLock::Untrack { key } => resolver.untrack_lock(&key),
```
5kbpers, line 392: Better to keep `start_ts` and `commit_ts`, it could help debugging.

sticnarf: But actually we don't have ways to know them.

5kbpers: Just in the case of deleting lock?

sticnarf: Yes. Locks are all untracked by deleting lock events.

5kbpers: hmmm, got.

MyonKeminta (approved):
I can't find any more problem... LGTM

components/resolved_ts/src/lib.rs
```diff
@@ -50,65 +54,29 @@ impl Resolver {
...
-        &mut self,
-        start_ts: TimeStamp,
-        commit_ts: Option<TimeStamp>,
-        key: Vec<u8>,
-    ) {
+    pub fn untrack_lock(&mut self, key: &[u8]) {
+        let start_ts = if let Some(start_ts) = self.locks_by_key.remove(key) {
+            start_ts
+        } else {
+            return;
```
5kbpers, line 66: Please add a warn or debug log here.

sticnarf: OK. I add a debug log. Otherwise, I worry such logs would bloat after leader transfer.

5kbpers (approved):

Issue #9351, opened 2020-12-24
Stale read when using read_index request


BusyJay:
## Bug Report

Similar to #9208, ReadIndex request can be given a stale index after just transferring leader. The API doesn't seem to be used anymore, so the best way to solve it is removing the API completely.

@NingLin-P @gengliqi PTAL

/cc @solotzg, just make sure tiflash will not use the API in the future.

solotzg:
since 4.0.8, tiflash won't use ReadIndex to get read-index from leader directly.

BusyJay:
I remember there is a solution to use proxy to do reading across DC, which also use read index. @hicqu Is it still a valid solution? Is the stale read proposal a good alternative to it?

hicqu:
@BusyJay that solution is paused and I think the stale read proposal is more useful than that.

Is it necessary to fix the bug for `TiFlash ^4.0.7`? If so we need to fix it, or we can remove the API directly.

BusyJay:
The only way to fix for them is using a hotfix. I suggest to ask them to upgrade to v4.0.9 instead.

hicqu:
Can it be closed now?

BusyJay:
Why? The API still exists and it can return wrong result.

PR #9531, opened 2021-01-20
mark read-index API deprecated

hicqu:

Deprecated the read-index API. Close https://github.com/tikv/tikv/issues/9351.

Code about `CmdType::ReadIndex` is not removed from `components/raftstore` for backforward compatibility.

### Check List 

Tests 
- No code

### Release note 
* No release note.


BusyJay:
> Code about CmdType::ReadIndex is not removed from components/raftstore for backforward compatibility.

I suggest to return error when those commands are used.

Issue #3572, opened 2018-09-05
Potential two different leaders on a same term


BusyJay:
Suppose a region A contains three peers a, b, c at term n. If c is uninitialized, a is stuck for some reason, then b can become leader at term n + 1. If c is restarted before accepting a snapshot, and a is recovered, then a can still start a campaign. Because we don't save hard state of an uninitialized peer, so c's term go back to 0. Hence a can win the campaign and become leader at term n + 1. So both a and b are the leaders of term n + 1, which violate the raft election safety.

However, such cases seems no harm except the violation itself. If no log is committed when b is leader, then the replica of b keeps the same as a. If there are logs committed, then a should be recovered or c has accepted a snapshot, in both cases, a can't become leader at term less than n + 1.

The root cause of the problem is that hard state is not saved when a peer is uninitialized, which introduce the risk of stepping back of term on the node.

gengliqi:
It's the same as https://github.com/tikv/tikv/issues/8381, I think we must add some check before proposing `conf change` to forbid promoting an uninitialized learner to voter. /cc @NingLin-P

NingLin-P:
> It's the same as #8381, I think we must add some check before proposing `conf change` to forbid promoting an uninitialized learner to voter. /cc @NingLin-P

This forbids is not enough, if PD adds a new peer as voter (it is allowed currently, and use by a lot of tests) this issue can still happen. Maybe an uninitialized voter should just ignore vote request, this is intuitive, since the voter can not persist its vote, it should not vote at all. But this also means a voter won't vote for election and log replication until it is initialized. /cc @gengliqi  @BusyJay

gengliqi:
> > It's the same as #8381, I think we must add some check before proposing `conf change` to forbid promoting an uninitialized learner to voter. /cc @NingLin-P
> 
> This forbids is not enough, if PD adds a new peer as voter (it is allowed currently, and use by a lot of tests) this issue can still happen. Maybe an uninitialized voter should just ignore vote request, this is intuitive, since the voter can not persist its vote, it should not vote at all. But this also means a voter won't vote for election and log replication until it is initialized. /cc @gengliqi @BusyJay

It is enough.
If PD wants to directly add a new peer as a voter, the leader must prevent it.
For tests, I suggest changing all adding voters to 2 steps, i.e. adding a learner then promoting it to a voter.
If a voter can ignore vote request, the available issue will happen even when the majority of peers are alive.

BusyJay:
[...]

PR #9533, opened 2021-01-20
raftstore: add more checks before proposing joint confchange

### LINKED ISSUE #9445: raftstore::test_joint_consensus::test_restart_in_joint_state failed
Latest build: <a href="https://internal.pingcap.net/idc-jenkins/job/tikv_ghpr_test/39932/display/redirect">https://internal.pingcap.net/idc-jenkins/job/tikv_ghpr_test/39932/display/redirect</a>

NingLin-P:

Issue Number: close #9445 #9332 #8789 #8775

Problem Summary:

Currently, TiKV doesn't check leave joint command. If the leader is `DemotingVoter`, it will step down after applying the leave joint command and lead to region unavailable until a new leader elect. Normally, this should not happened, because PD (with its acknowledge about the region) won't send such command to the leader,  and as such command can send to the leader with the right `RegionEpoch` meaning PD has up-to-date acknowledge. But we should add such check for robustness.

What's Changed:

This PRs add checks to prevent leave joint while the leader is `DemotingVoter`, and also add checks to prevent split or merge region while the region is in joint state, again this also should not happen because of the epoch check.

This PRs also fix some unstable tests,  which cause by checking `must_get_none` on peers that were added as learner with joint confchange. This PRs fix them by checking the roles of these peers.

### Check List 

Tests 

- Integration test

### Release note 
- raftstore: add more checks before proposing joint confchange

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -3107,6 +3107,22 @@ where
         self.ctx.store_stat.engine_total_keys_written += metrics.written_keys;
     }
 
+    fn check_joint_state(&self, msg: &RaftCmdRequest) -> Result<()> {
+        if !self.fsm.peer.in_joint_state() {
+            return Ok(());
+        }
+        let cmd = msg.get_admin_request();
+        if cmd.has_prepare_merge() || cmd.has_commit_merge() || cmd.has_split() || cmd.has_splits()
+        {
+            return Err(box_err!(
+                "{} region in joint state, can not propose split or merge command, command: {:?}",
+                self.fsm.peer.tag,
+                cmd
+            ));
+        }
+        Ok(())
+    }
+
     /// Check if a request is valid if it has valid prepare_merge/commit_merge proposal.
     fn check_merge_proposal(&self, msg: &mut RaftCmdRequest) -> Result<()> {
         if !msg.get_admin_request().has_prepare_merge()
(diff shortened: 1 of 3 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3283,17 +3299,32 @@ where
...
-                "peer_id" => self.fsm.peer_id(),
-                "message" => ?msg,
-                "err" => %e,
-                "error_code" => %e.error_code(),
-            );
-            cb.invoke_with_response(new_error(e));
-            return;
+        if msg.has_admin_request() {
+            if let Err(e) = self.check_joint_state(&msg) {
+                warn!(
```
BusyJay, line 3304: Does PD follows the same constraint?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3107,6 +3107,22 @@ where
         self.ctx.store_stat.engine_total_keys_written += metrics.written_keys;
     }
 
+    fn check_joint_state(&self, msg: &RaftCmdRequest) -> Result<()> {
+        if !self.fsm.peer.in_joint_state() {
+            return Ok(());
+        }
+        let cmd = msg.get_admin_request();
+        if cmd.has_prepare_merge() || cmd.has_commit_merge() || cmd.has_split() || cmd.has_splits()
```
BusyJay, line 3115: Shouldn't the pipeline checker filter them automatically?

gengliqi:
I suggest fixing https://github.com/tikv/tikv/issues/3572 in this PR.
Maybe it's better to mention this issue is about correctness, not availability.

components/raftstore/src/store/fsm/peer.rs
NingLin-P: Yes, when a region in joint state PD won't try to split/merge this region /cc @HunDunDM

components/raftstore/src/store/fsm/peer.rs
gengliqi: Do you mean epoch checker? The epoch checker can only filter the concurrent admin cmd.

gengliqi: Actually both the merge and split cmd are ok. I agree that rejecting the merge cmd for simpleness. But maybe it's not necessary to reject split cmd?

NingLin-P: Yes, I removed the check for split cmd, PTAL again

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3115,6 +3115,14 @@ where
             return Ok(());
         }
 
+        if self.fsm.peer.in_joint_state() {
```
BusyJay, line 3118: I think you should only prevent prepare merge otherwise it can be an incompatible change.

BusyJay (approved):
Either fix #3572 in this PR or a separate one LGTM.

gengliqi (approved):

NingLin-P:
> Either fix #3572 in this PR or a separate one LGTM.

I will fix it in the next PR

Issue #9513, opened 2021-01-16
`validate_endpoints` does not check `leader` field of `GetMembersResponse`


5kbpers:
### What version of TiKV are you using?

nightly

### What did happened?
During the initialization of TiKV, it will check pd connection in `validate_endpoints` with following steps:
1. call `get_members` for all pd endpoints
2. Use the leader information in the first `GetMembersResponse` to connect leader
By default the retry time is unlimited. Sometimes the response may not contain the `leader` field, then TiKV will be hung with the failure to connect leader.

nolouch:
there are three members, named pd0,pd1, pd2. and the pd0 is leader, pd1 is unhealth, which means pd1 may not know the leader. the TiKV start with endpoints="pd0,pd1,pd2", which may cause cannot connect to PD forever because cannot know the leader of pd members.
![image](https://user-images.githubusercontent.com/6428910/104907611-5172c100-59c0-11eb-9e37-eb9848946cb4.png)

PR #9537, opened 2021-01-21
pd_client: fix connect to PD failed forever

Yisaer:

Issue Number: close https://github.com/tikv/tikv/issues/9513

Problem Summary:

Detailed description in https://github.com/tikv/tikv/issues/9513#issuecomment-762175968

Shuffle the members before connect to pd members in order to avoid connecting to failure pd members forever.

### Check List 

Tests 

- Unit test

### Release note 

* No Release Note

### DIFF
--- a/components/pd_client/src/util.rs
+++ b/components/pd_client/src/util.rs
@@ -30,6 +30,9 @@ use tikv_util::timer::GLOBAL_TIMER_HANDLE;
 use tikv_util::{Either, HandyRwLock};
 use tokio_timer::timer::Handle;
 
+use rand::seq::SliceRandom;
+use rand::thread_rng;
+
 use super::{Config, Error, FeatureGate, PdFuture, Result, REQUEST_TIMEOUT};
 
 pub struct Inner {
@@ -340,8 +343,7 @@ pub async fn validate_endpoints(
     security_mgr: Arc<SecurityManager>,
 ) -> Result<(PdClientStub, GetMembersResponse)> {
     let len = cfg.endpoints.len();
-    let mut endpoints_set = HashSet::with_capacity_and_hasher(len, Default::default());
-
+    let mut endpoints_set = HashSet::with_capacity_and_hasher(len, Default::default());    
     let mut members = None;
     let mut cluster_id = None;
     for ep in &cfg.endpoints {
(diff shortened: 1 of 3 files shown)


components/pd_client/src/util.rs
```diff
@@ -425,8 +427,13 @@ pub async fn try_connect_leader(
     let members = previous.get_members();
     let cluster_id = previous.get_header().get_cluster_id();
     let mut resp = None;
+    let mut shuffule_members: Vec<kvproto::pdpb::Member> = Vec::new();
```
longfangsong, line 430: You can clone directly, ie.
```rust
let mut shuffle_member = members.clone();
```

Yisaer: updated.

hi-rustin:
I'm sorry for looking at the wrong pr. I hope I'm not disturbing you.

components/pd_client/src/util.rs
```diff
@@ -425,8 +427,10 @@ pub async fn try_connect_leader(
     let members = previous.get_members();
     let cluster_id = previous.get_header().get_cluster_id();
     let mut resp = None;
+    let mut shuffle_members: Vec<_> = members.to_vec();
```
JmPotato, line 430:
```suggestion
    let mut shuffle_members = members.to_vec();
```

I think the type annotation here is unnecessary.

components/pd_client/src/util.rs
```diff
@@ -425,8 +427,10 @@ pub async fn try_connect_leader(
...
+    shuffle_members.shuffle(&mut thread_rng());
     // Try to connect to other members, then the previous leader.
-    'outer: for m in members
+    'outer: for m in shuffle_members
         .iter()
         .filter(|m| *m != previous_leader)
         .chain(&[previous_leader.clone()])
```
5kbpers, line 436: How about just checking the `leader` field and skipping the valid response?

JmPotato (approved):

components/pd_client/src/util.rs
nolouch: PTAL @overvenus

tests/integrations/pd/test_rpc_client.rs
```diff
@@ -218,6 +218,30 @@ fn test_validate_endpoints() {
...
+    let eps_count = 3;
+    let server = MockServer::with_case(eps_count, Arc::new(Split::new()));
+    let env = Arc::new(
+        EnvBuilder::new()
+            .cq_count(1)
+            .name_prefix(thd_name!("test-pd"))
+            .build(),
+    );
+    let eps = server.bind_addrs();
+    let mut copy_eps = eps.clone();
```
5kbpers, line 232: Seems we do not need to clone a new vector here? (just iterate over the reference of the vector)

5kbpers (approved):
reset LGTM

Issue #9488, opened 2021-01-12
TiKV crashes continuously if access AWS KMS failed at the first time


aylei:
### What version of TiKV are you using?

`v4.0.9`

Note: `v4.0.8` does not have this issue

### What operating system and CPU are you using?

CentOS 7

### Steps to reproduce

1. Configure TiKV to use TDE with AWS KMS
2. Run TiKV with no AWS credentials
3. TiKV crashes
4. Restart TiKV with proper credentials
5. TiKV will still crash
6. Remove the empty `dict` and `key*..tmp` files
7. TiKV can then start normally

Logs below:

TiKV crash
```
db-tikv-0 tikv [2021/01/12 02:51:39.496 +00:00] [INFO] [mod.rs:335] ["starting working thread"] [worker=addr-resolver]
db-tikv-0 tikv [2021/01/12 02:51:39.496 +00:00] [INFO] [mod.rs:335] ["starting working thread"] [worker=region-collector-worker]
db-tikv-0 tikv [2021/01/12 02:51:39.502 +00:00] [ERROR] [mod.rs:121] ["encryption: failed to load key dictionary."]
[...]

PR #9540, opened 2021-01-21
Make TiKV not to create dict files when master_key cannot work

hunterlxt:

Issue Number: close https://github.com/tikv/tikv/issues/9488

Problem Summary:
At the first time TiKV access AWS KMS failed, it leaves corrupted dict file which causes TiKV to crash continuously.

check the `master_key` before creating the dicts.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- Fix the bug that TiKV could not start normally after failing to access AWS KMS.


components/encryption/src/encrypted_file/mod.rs
```diff
@@ -16,7 +16,7 @@ use crate::Result;
 mod header;
 pub use header::*;
 
-pub const TMP_FILE_SUFFIX: &str = ".tmp";
+pub const TMP_FILE_SUFFIX: &str = "tmp";
```
gregwebs, line 19: Can you add a test case for this change?

components/encryption/src/manager/mod.rs
```diff
@@ -473,6 +473,10 @@ impl DataKeyManager {
             // Encryption is being enabled.
             (Ok(None), _) => {
                 info!("encryption is being enabled. method = {:?}", args.method);
+                // If `master_key` doesn't work correctly, dict files will not be created.
+                let plaintext_temp = "";
+                let _ = master_key.encrypt(plaintext_temp.as_bytes())?;
```
gregwebs, line 478: Can you add a test case for this change?

components/encryption/src/manager/mod.rs
```diff
@@ -473,6 +473,10 @@ impl DataKeyManager {
             // Encryption is being enabled.
             (Ok(None), _) => {
                 info!("encryption is being enabled. method = {:?}", args.method);
+                // If `master_key` doesn't work correctly, dict files will not be created.
+                let plaintext_temp = "";
```
gregwebs, line 477: It would be better to have some content here. It would be valid to reject an encrypt operation because the content is empty.

components/encryption/src/manager/mod.rs
hunterlxt: added

components/encryption/src/encrypted_file/mod.rs
hunterlxt: This is just a small fix, `set_extension` automatically add a `.`

components/encryption/src/manager/mod.rs
```diff
@@ -477,6 +477,10 @@ impl DataKeyManager {
...
+                let _ = master_key.encrypt(plaintext_temp.as_bytes())?;
```
yiwu-arbug, line 482: I think this is not a good fix. It doesn't prevent us from intermittent failure. For example the test call here could success, while the real call fail. In that case we still leave behind an empty key dict.

yiwu-arbug (changes requested):

components/encryption/src/manager/mod.rs
hunterlxt: Empty file dict is considered as non-existent a `file.dict` now.

components/encryption/src/file_dict_file.rs
```diff
@@ -187,6 +189,16 @@ impl FileDictionaryFile {
                 }
             }
         }
+
+        // If the file dict is empty, it must be deleted which avoids continuing TiKV crashes.
+        if file_dict.files.is_empty() {
+            std::fs::remove_file(self.base.join(&self.name))?;
```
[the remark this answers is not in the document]
gregwebs: I think this should be done from the manager though. This problem occurs due to the manager `open` function. Fixing it here means that now the file dictionary knows how the manager operates and we have to add a comment explaining how the logic is intertwined.

I think in the manager open function we can fix this just by reading the master key first before writing out key.dict.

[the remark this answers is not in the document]
gregwebs: Sorry, the solution above was a little off. I think the problem is that manager open calls FileDictionaryFile::open before checking if the master key is valid. So I think we can check the master key before the FileDictionaryFile::open either with the previous solution or by immediately checking the error of the encrypted file read.

components/encryption/src/file_dict_file.rs
```diff
@@ -187,6 +189,16 @@ impl FileDictionaryFile {
                 }
             }
         }
+
+        // If the file dict is empty, it must be deleted which avoids continuing TiKV crashes.
```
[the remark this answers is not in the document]
hunterlxt: updated

components/encryption/src/file_dict_file.rs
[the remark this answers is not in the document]
hunterlxt: Just like what @yiwu-arbug said, reading the master key once does not guarantee that the next read will be successful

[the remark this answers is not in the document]
gregwebs: You are right that my suggestion is still not good enough. I still suggest we solve it in the manager open function. It is this function that has setup a state machine that it doesn't respect on restart. We can simply add a case in the match statement to check if key_bytes is NotFound but file_dict_file is found.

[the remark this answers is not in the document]
hunterlxt: use a new solution as you suggested. @yiwu-arbug PTAL

components/encryption/src/manager/mod.rs
```diff
@@ -121,6 +121,14 @@ impl Dicts {
             // ...else, return either error.
             (file_dict_file, key_bytes) => {
                 if let Err(key_err) = key_bytes {
+                    let (file_dict_file, file_dict) = file_dict_file.unwrap();
```
yiwu-arbug, line 124: avoid unwrap. In fact we can make a match arm for it ?
```
match (log_content, key_bytes) {
    ...
    (Ok(file_dict_file, Err(key_bytes)) if file_dict_file.0.files.is_empty() => {
        ... // delete empty file dict
    }
    // ...else, return either error
    (file_dict_file, key_bytes) => {
         ...
    }
}
```
(I might be using incorrect syntax though)

yiwu-arbug (commented):
rest looking good to me.

yiwu-arbug (approved):
LGTM

gregwebs (approved):

yiwu-arbug:
@hunterlxt can you fix lint?

hunterlxt:
I tested and confirmed that it can work normally after failure

PR #9549, opened 2021-01-22
avoid stale result for read index command

### LINKED ISSUE #9351: Stale read when using read_index request

hicqu:

Close #9351 .

Avoid stale result for read index command.

### Check List 

Tests 
- Integration test

Side effects

### Release note 
* No release note.


components/raftstore/src/store/peer.rs
```diff
@@ -1946,10 +1946,15 @@ where
                 continue;
             }
             if !replica_read {
-                if read_index.is_none() {
+                match (read_index, read.read_index) {
+                    // `read_index` could be less than `read.read_index` because the former is
+                    // filled with `committed index` when proposed, and the latter is filled after
+                    // a read-index procedure finished.
+                    (Some(i_test), Some(i_got)) if i_test < i_got => read_index = read.read_index,
```
BusyJay, line 1953: Confusing names. How about `lower_bound` and `batch_index`?

NingLin-P (approved):
LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -1953,7 +1953,14 @@ where
                 continue;
             }
             if !replica_read {
-                if read_index.is_none() {
+                if read_index.is_some() && read.read_index.is_some() {
```
gengliqi, line 1956: It seems the `read.read_index` can not be `None` when calling `response_read` in any case.

gengliqi: I prefer to change here. https://github.com/tikv/tikv/blob/bf92a4a9b56f410f3c0eefe8056c22923a08ceee/components/raftstore/src/store/peer.rs#L2599 
If `self.get_store().term(commit_index).unwrap() == self.term()` is not true, we should not set it.
If so, here logic is the same as before. The only place that needs to be updated is the comment "Actually, the read_index is none if and only if..."

gengliqi: Hmmm, I think we can use the commit index in raft-rs directly here. It's right because
1. the first read index request has already checked if the term of last committed entry is the same as current term. (we can add an assert here)
2. the response is sent after getting the raft ready so the `on_leader_commit_idx_changed` is called.

gengliqi: @NingLin-P  said using raft-rs's commit index may be too large and the caller needs to wait for a long time.
I am ok with these two choices.

BusyJay: Checking `commit_index` before setting it should be the same as this patch. This patch can be a last protection against any wrong modification. Do you have any reasons to prefer checking term instead?

5kbpers (approved):

5kbpers:
Please resolve conflicts.

PR #9045, opened 2020-11-14
raftstore: move sst cleanup (compact range) task to its own thread

yiwu-arbug:
Signed-off-by: Yi Wu <yiwu@pingcap.com>

### What problem does this PR solve?

Issue Number: close #9044

Problem Summary: move sst cleanup task to its own thread. It fixes deadlock when compaction guard is enabled, and there are few thread in the background_worker pool.

### What is changed and how it works?

What's Changed: Revert changes done in #8821 which move sst cleanup task to background_worker pool. The cleanup task will fire compact range, which is a blocking call and can take up to minutes. Meanwhile when compaction guard is enabled, the feature will schedule a task to fetch regions in the compaction range, which is also executed in the background_worker pool. When there's few background_worker pool thread (the pool size is `min(4, max(1, (#cpu +7)/8))`, so there could be only 1 thread when #cpu <= 8), it could cause a deadlock.

Despite the deadlock, having SST cleanup task sharing the background_worker pool is bad, since it could take an extensive amount of time, it could delay other tasks in the thread pool.

### Related changes

compaction guard was added in #8115. SST cleanup task worker is merged with background worker in #8821.

### Check List 

Tests

CI

### Release note 

- No release notes.

Issue #9044, opened 2020-11-13
compaction guard: deadlock when creating compaction partitioner


yiwu-arbug:
### What version of TiKV are you using?
master

### Steps to reproduce
enable compaction guard with config `rocksdb.[defaultcf|writecf].enable-compaction-guard = true`.

### What did you expect?
no deadlock

### What did happened?
Deadlock when TiKV try to create compaction partitioner instance:
```
Thread 63 (Thread 0x7fea467ff700 (LWP 18955)):
#0  0x00007fea4f8d8bf9 in syscall () from /lib64/libc.so.6
#1  0x000056375a11e99a in futex_wait () at library/std/src/sys/unix/futex.rs:18
#2  park () at library/std/src/sys_common/thread_parker/futex.rs:50
#3  park () at library/std/src/thread/mod.rs:881
#4  std::sync::mpsc::blocking::WaitToken::wait::hae3a13d09b5b5f41 () at library/std/src/sync/mpsc/blocking.rs:64
#5  0x000056375a323fc0 in rocksdb::sst_partitioner::sst_partitioner_factory_create_partitioner::ha9a189958e3f3bfd (ctx=0x7fea4f3f8150, context=<optimized out>) at /rustc/b1496c6e606dd908dd651ac2cce89815e10d7fc5/library/std/src/sync/mpsc/oneshot.rs:140
#6  0x000056375a881b60 in crocksdb_sst_partitioner_factory_impl_t::CreatePartitioner (this=<optimized out>, partitioner_context=...) at crocksdb/c.cc:5917
#7  0x000056375ac410fb in rocksdb::Compaction::CreateSstPartitioner (this=<optimized out>) at /rust/git/checkouts/rust-rocksdb-a9a28e74c6ead8ef/c6b58a4/librocksdb_sys/rocksdb/db/compaction/compaction.cc:554
#8  0x000056375ac567d8 in rocksdb::CompactionJob::ProcessKeyValueCompaction (this=this@entry=0x7fea467d9a40, sub_compact=0x7fea44c72480) at /rust/git/checkouts/rust-rocksdb-a9a28e74c6ead8ef/c6b58a4/librocksdb_sys/rocksdb/db/compaction/compaction_job.cc:906
[...]

PR #8821, opened 2020-10-13
*: Merge some single thread of TiKV into one thread pool.

Little-Wallace:
### What problem does this PR solve?

Issue Number: close #xxx 

Problem Summary:
There are a lot of threads in TiKV. But most of them are only working a while and then suspend themself and wait for waking up. If there are a log of threads waiting for waking up, system will cost much time to schedule and we can not know how much CPU that this thread use.

### What is changed and how it works?

Proposal: [xxx](url) 
This PR is to merged most of worker into one or two threads so that they can run without suspend and schedule.

I refactor Worker in components/tikv_util/src/worker/mod.rs so that it could run multiple Runner in one threads.

What's Changed:
The following thread will be merged into background-pool:
- snapshot-worker
- consistency-check
- cleanup-worker
- raft-gc-worker
- split-check
- addr-resolver
- backup-endpoint
- cdc

### Related changes

- This PR can not be merged until https://github.com/tikv/yatp/issues/46 has been solved. Because `Runner` in yatp pool may never drop because of bug of yatp pool.

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 

- Add one background thread pool for each background worker. Configure it in `server.background-thread-count`.

PR #9552, opened 2021-01-25
raftstore: 2nd attempt to fix deadlock when creating compaction guard

yiwu-arbug:

Issue Number: #9044

Problem Summary:
After the introduction of compaction guard (#8115), RocksDB compaction threads can call region info accessor to create SST partitioner, which blocks the compaction thread. After #8821 both of region info accessor and region tasks in moved into a unified background worker pool. Per #8990, the thread pool would contain 2-3 threads. The region tasks, such as `RegionTask::Apply` and `RegionTask::Destroy`, can execute file ingestions, which leads to a blocking flush in `crocksdb_ingest_external_file_optimized` in rust-rocksdb. It can create a deadlock when compaction is waiting in the background pool queue to create SST partitioner and multiple region tasks are waiting for RocksDB threads to run flush.

Since the region tasks contain blocking flush call, which could run for a while, it doesn't sound like a good idea to have them run in the background pool, which could block other background pool tasks. Moving region info access to its own worker is a precaution to avoid similar deadlock happen in the future.

What's Changed: 
Move region info accessor and region tasks each out of the unified background worker pool to their own worker thread. 

Also in compaction_guard.rs, access region info provider in `should_partition()` instead of `create_partitioner()`, since the later can be invoked by RocksDB while holding db_mutex.

### Related changes

compaction guard was added in #8115. 
SST cleanup task worker is merged with background worker in #8821. 
The first attempt to fix the deadlock was #9045.
background thread pool size was default to 2 instead of 1 in #8990, which should relieve the problem, but apparently it is not.

This PR depends on https://github.com/tikv/rust-rocksdb/pull/598.

### Check List 

Tests 

Will do manual test.

### Release note 

- Fix deadlock when creating compaction guard.


yiwu-arbug:
WIP for pending test.

yiwu-arbug:
@Little-Wallace @Connor1996 any suggestion on how can I implement an unit test?

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -490,23 +490,31 @@ impl RunnableWithTimer for RegionCollector {
...
     scheduler: Scheduler<RegionInfoQuery>,
 }
 
 impl RegionInfoAccessor {
     /// Creates a new `RegionInfoAccessor` and register to `host`.
     /// `RegionInfoAccessor` doesn't need, and should not be created more than once. If it's needed
     /// in different places, just clone it, and their contents are shared.
-    pub fn new(host: &mut CoprocessorHost<impl KvEngine>, worker: &Worker) -> Self {
+    pub fn new(host: &mut CoprocessorHost<impl KvEngine>) -> Self {
+        let worker = WorkerBuilder::new("region-info").create();
```
Connor1996, line 507: name it "region-collector-worker" too?

Connor1996:
> @Little-Wallace @Connor1996 any suggestion on how can I implement an unit test?

Schedule destroy-region task and do a manual compact?

yiwu-arbug:
> > @Little-Wallace @Connor1996 any suggestion on how can I implement an unit test?
> 
> Schedule destroy-region task and do a manual compact?

Should it be at raftstore level or at server level? Since the issue involve the server level `background_worker`.

yiwu-arbug:
I'm to move region accessor query from `SstPartitionerFactory::create_partitioner` to `SstPartitioner::should_partition`, since rocksdb can call `create_partitioner` while holding db_mutex, which can block other TiKV operations that read/write rocksdb.

Cargo.toml
```diff
@@ -218,6 +218,9 @@ raft-proto = { git = "https://github.com/tikv/raft-rs", branch = "master", defau
 protobuf = { git = "https://github.com/pingcap/rust-protobuf", rev = "65e9df20fbcbcf2409d5ee86a2332ecd04c534f8" }
 protobuf-codegen = { git = "https://github.com/pingcap/rust-protobuf", rev = "65e9df20fbcbcf2409d5ee86a2332ecd04c534f8" }
 
+[patch.'https://github.com/tikv/rust-rocksdb']
+rocksdb = { git = "https://github.com/yiwu-arbug/rust-rocksdb", branch = "guard_mut" }
```
Connor1996, line 222: Please update this

Connor1996 (approved):
rest LGTM

yiwu-arbug:
> > > > @Little-Wallace @Connor1996 any suggestion on how can I implement an unit test?
> > > 
> > > 
> > > Schedule destroy-region task and do a manual compact?
> > 
> > 
> > Should it be at raftstore level or at server level? Since the issue involve the server level `background_worker`.
> 
> Maybe in integrations/raftstore?

will update the unit test in a separate PR.

components/raftstore/src/store/compaction_guard.rs
```diff
@@ -48,60 +48,87 @@ impl<P: RegionInfoProvider> CompactionGuardGeneratorFactory<P> {
...
-
-impl SstPartitioner for CompactionGuardGenerator {
-    fn should_partition(&self, req: &SstPartitionerRequest) -> SstPartitionerResult {
-        let mut pos = self.pos.get();
+impl<P: RegionInfoProvider> SstPartitioner for CompactionGuardGenerator<P> {
+    fn should_partition(&mut self, req: &SstPartitionerRequest) -> SstPartitionerResult {
+        if !self.initialized {
+            self.initialize();
+        }
+        if !self.use_guard {
```
Little-Wallace, line 128: Should we check `use_guard` before `initialize `?

Little-Wallace (approved):
LGTM

PR #9554, opened 2021-01-25
raftstore: use lru cache store local read delegate

### LINKED ISSUE #9469: raftstore: the invalid read delegates in local reader may not be removed forever
## Bug Report

Although https://github.com/tikv/tikv/pull/4765 introduce the `invalid` flag to remove the invalid `ReadDelegate`s on `LocalReader`, but the flag only be checked when handling a request to the `ReadDelegate`, if the `ReadDelegate` is marked as invalid and no incoming request to it, the `ReadDelegate` not be removed and consuming memory.

Although `mem::size_of::<ReadDelegate>()` is 122 bytes, but if this case happened the `ReadDelegate` will not be removed forever, and the memory occupied by these invalid `ReadDelegate` will only increase not decrease as TiKV running.

NingLin-P:

Issue Number: close #9469 

What's Changed:

Using `LruCache` instead of `HashMap` to store local `ReadDelegate`

### Release note 
- raftstore: use lru cache store local read delegate

### DIFF
--- a/components/raftstore/src/store/worker/read.rs
+++ b/components/raftstore/src/store/worker/read.rs
@@ -424,11 +424,18 @@ where
                 debug!("update local read delegate"; "region_id" => region_id);
                 self.metrics.rejected_by_cache_miss += 1;
 
+                let (meta_len, meta_reader) = {
+                    let meta = self.store_meta.lock().unwrap();
+                    (
+                        meta.len(),
+                        meta.readers.get(&region_id).cloned().map(Arc::new),
+                    )
+                };
+
                 // Remove the stale delegate
                 self.delegates.remove(&region_id);
-
-                let meta = self.store_meta.lock().unwrap();
-                match meta.readers.get(&region_id).cloned().map(Arc::new) {
+                self.delegates.resize(meta_len);
+                match meta_reader {
                     Some(reader) => {
                         self.delegates.insert(region_id, Arc::clone(&reader));
                         Some(reader)
--- a/components/raftstore/src/store/worker/read.rs
+++ b/components/raftstore/src/store/worker/read.rs
@@ -425,7 +425,7 @@ where
                 let (meta_len, meta_reader) = {
                     let meta = self.store_meta.lock().unwrap();
                     (
-                        meta.len(),
+                        meta.readers.len(),
                         meta.readers.get(&region_id).cloned().map(Arc::new),
                     )
                 };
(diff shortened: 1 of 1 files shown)


NingLin-P:
Do you have any suggestions about the capacity of the `LruCache`? @BusyJay @hicqu

components/raftstore/src/store/worker/read.rs
```diff
@@ -373,7 +375,7 @@ where
             cache_read_id,
             store_id: Cell::new(None),
             metrics: Default::default(),
-            delegates: HashMap::default(),
+            delegates: LruCache::with_capacity(DELEGATE_LRU_CACHE_CAPACITY),
```
BusyJay, line 378: I suggest to use default. There can be no allocation at all for learners.

BusyJay:
You can use `resize` to shrink the cache at every modification operations.

components/raftstore/src/store/worker/read.rs
NingLin-P: There is not `LruCache::default`,  underneath `LruCache::with_capacity` using `HashMap::default` to initialize, so it won't allocate large memory until it needs to.

hicqu:
There are some `FIXME`s can be removed in `read.rs`.

NingLin-P:
> You can use `resize` to shrink the cache at every modification operations.

If so, the problem still exists, local read will not call `delegate.remove` for those invalid delegates.

NingLin-P:
> There are some `FIXME`s can be removed in `read.rs`.

The only `FIXME` is removed.

hicqu (approved):

BusyJay:
> If so, the problem still exists, local read will not call delegate.remove for those invalid delegates.

What I mean is `cache.resize(meta.delegates.len())`.

NingLin-P:
I also remove the `invalid` flag, PTAL again @hicqu @BusyJay

hicqu:
/LGTM.

components/raftstore/src/store/worker/read.rs
```diff
@@ -154,11 +153,8 @@ pub struct ReadDelegate {
 
 impl Drop for ReadDelegate {
     fn drop(&mut self) {
-        // `mark_invalid` if the source `ReadDelegate` is dropped
-        if self.track_ver.source {
-            self.mark_invalid();
-            self.track_ver.inc();
-        }
+        // call `inc` to notify the source `ReadDelegate` is dropped
+        self.track_ver.inc();
```
BusyJay, line 157: Why not check `source` flag?

NingLin-P: `inc` check the flag inside.

BusyJay (approved):

PR #9558, opened 2021-01-25
storage, gc_worker: Update max_ts when invoking scan_lock and register_lock_observer

### LINKED ISSUE #9526: incompatibility between gc and async commit
## Bug Report

Because min_commit_ts is not updated for resolving locks in gc, so async commit may commit a transaction using a ts smaller than safe ts, which is dangerous and can break transaction easily.

MyonKeminta:

Issue Number: close #9526 

Problem Summary: In some very-rare corner cases, if prewrite requests of async commit transactions arrives  so late that when it arrives, GC safepoint already exceeds its start_ts, it's possible that GC breaks the transaction's atomicity.

What's Changed:

This PR mades a few changes to fix the issue.

* Move `scan_lock` out of scheduler, and uses similar handling logic as other kv read requests like `get`, `batch_get`, `scan`, etc.
  * Including updating `max_ts` of concurrency manager and checking memory locks. This fixes the issue for non green gc.
* Add an optional `end_key` parameter for `scan_lock`.
* Update `max_ts` of concurrency manager and check memory locks when `register_lock_observer` is invoked. This fixes the issue for green gc.

### Related changes

- Though this PR fixes #9526 , but for non green gc, the scan_lock phase may not be very efficient since each scan_lock request will cause scanning memory lock from that region until the very end. Therefore we need an additional PR to TiDB to fill the `end_key` parameter of scan_lock reuqest.

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

* Fix an issue that in some very rare cases, GC may break the atomicity of async commit transactions.


Cargo.toml
```diff
@@ -218,6 +218,9 @@ raft-proto = { git = "https://github.com/tikv/raft-rs", branch = "master", defau
 protobuf = { git = "https://github.com/pingcap/rust-protobuf", rev = "65e9df20fbcbcf2409d5ee86a2332ecd04c534f8" }
 protobuf-codegen = { git = "https://github.com/pingcap/rust-protobuf", rev = "65e9df20fbcbcf2409d5ee86a2332ecd04c534f8" }
 
+[patch.'https://github.com/pingcap/kvproto.git']
```
sticnarf, line 221: Your own branch is already deleted :(

MyonKeminta:
This PR doesn't solve all problems around the GC boundary. There still exists some corner cases but it's even much more rare than this case. Making TiKV aware of the latest safepoint before GC starts and blocking stale requests may solve those problems, but the implementation might be kind of difficult.

src/server/service/kv.rs
```diff
@@ -1254,6 +1254,51 @@ fn future_batch_get<E: Engine, L: LockManager>(
...
+        Ok(resp)
+    }
+}
+
+// txn_command_future!(future_scan_lock, ScanLockRequest, ScanLockResponse, (v, resp) {
+//     match v {
+//         Ok(locks) => resp.set_locks(locks.into()),
+//         Err(e) => resp.set_error(extract_key_error(&e)),
+//     }
+// });
```
sticnarf, line 1300: Remove it?

src/server/gc_worker/applied_lock_collector.rs
```diff
@@ -407,6 +417,15 @@ impl AppliedLockCollector {
...
     pub fn start_collecting(&self, max_ts: TimeStamp, callback: Callback<()>) -> Result<()> {
+        // Before starting collecting, check the concurrency manager to avoid later prewrite
+        // requests uses a min_commit_ts less than the safepoint.
+        // `max_ts` here is the safepoint of the current round of GC.
+        self.concurrency_manager.update_max_ts(max_ts);
+        self.concurrency_manager
+            .read_range_check(None, None, |k, l| {
+                Lock::check_ts_conflict(Cow::Borrowed(l), k, max_ts, &TsSet::Empty)
+                    .map_err(MvccError::from)
+            })?;
```
sticnarf, line 428: Is there a test for this?

src/storage/mod.rs
```diff
@@ -686,6 +689,124 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
+
+                KV_COMMAND_COUNTER_VEC_STATIC.get(CMD).inc();
+                SCHED_COMMANDS_PRI_COUNTER_VEC_STATIC
+                    .get(priority_tag)
+                    .inc();
+
+                let command_duration = tikv_util::time::Instant::now_coarse();
+
+                concurrency_manager.update_max_ts(max_ts);
+                // TODO: Use better error handling here
```
sticnarf, line 729: Could you describe it more in detail? Do you mean returning error will make this round of GC fail, so a better way is to keep these locks and scan some more locks to reach the limit?

src/storage/mod.rs
```diff
@@ -686,6 +689,124 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
+                    statistics.add(reader.get_statistics());
+                    let (kv_pairs, _) = result?;
+                    let mut locks = Vec::with_capacity(kv_pairs.len());
+                    for (key, lock) in kv_pairs {
+                        let mut lock_info = LockInfo::default();
+                        lock_info.set_primary_lock(lock.primary);
+                        lock_info.set_lock_version(lock.ts.into_inner());
+                        lock_info.set_key(key.into_raw().map_err(txn::Error::from)?);
+                        lock_info.set_lock_ttl(lock.ttl);
+                        lock_info.set_txn_size(lock.txn_size);
```
sticnarf, line 785: Maybe better to use `into_lock_info`?

src/server/gc_worker/applied_lock_collector.rs
```diff
@@ -382,10 +384,14 @@ impl Runnable for LockCollectorRunner {
 pub struct AppliedLockCollector {
     worker: Mutex<Worker>,
     scheduler: Scheduler<LockCollectorTask>,
+    concurrency_manager: ConcurrencyManager,
```
cfzjywxk, line 387: Feels we've made many things coupled and there may still be some possible corner cases.

src/storage/mod.rs
```diff
@@ -686,6 +689,124 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         }
     }
 
+    pub fn scan_lock(
```
cfzjywxk, line 692: Do we have test case for the latter arrived async commit transaction and checking its `commitTS` with safe point ts?

src/storage/mod.rs
MyonKeminta: Oh.. this is copied from the old code. I didn't even notice the problem before, there are many fields missing..

src/storage/mod.rs
youjiali1995: How bout collecting all locks in the `concurrency_manager` and merging it with the result of `scan_lock`?

src/server/gc_worker/applied_lock_collector.rs
MyonKeminta: For the current implementation of TiDB, GC will simply fail.

src/storage/mod.rs
MyonKeminta: Is it necessary? How do  you think @sticnarf 
and how about Green  GC mode?

MyonKeminta:
I've added more tests. @sticnarf @cfzjywxk PTAL thanks

src/storage/mod.rs
youjiali1995: OK. It's nearly impossible to happen.

src/storage/mod.rs
```diff
@@ -686,6 +689,132 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
+        let res = self.read_pool.spawn_handle(
+            async move {
+                if let Ok(start_key) = start_key
+                    .as_ref()
+                    .map(|k| k.to_raw())
+                    .unwrap_or_else(|| Ok(vec![]))
+                {
+                    let mut key = vec![];
+                    if let Some(end_key) = &end_key {
+                        if let Ok(end_key) = end_key.to_owned().into_raw() {
```
youjiali1995, line 714:
```suggestion
                        if let Ok(end_key) = end_key.to_raw() {
```

src/storage/mod.rs
```diff
@@ -686,6 +689,132 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
+                        read_id: None,
+                        start_ts: max_ts,
+                        key_ranges: vec![key_range],
+                    }
+                } else {
+                    SnapContext {
+                        pb_ctx: &ctx,
+                        ..Default::default()
+                    }
+                };
```
youjiali1995, line 767: It's better to extract a `prepare_range_snap_ctx()` because there are many duplicated codes in `scan()`.

tests/failpoints/cases/test_kv_service.rs
```diff
@@ -36,3 +37,92 @@ fn test_kv_scan_memory_lock() {
...
+        std::thread::sleep(Duration::from_millis(200));
+
+        let client1 = client.clone();
+        let ctx1 = ctx.clone();
+        let handle2 = std::thread::spawn(move || {
+            if *use_green_gc {
+                let mut req = RegisterLockObserverRequest::default();
+                req.set_max_ts(ts + 20);
+                let resp = client1.register_lock_observer(&req).unwrap();
+                assert_ne!(resp.error.len(), 0);
```
youjiali1995, line 101:
```suggestion
                assert!(!resp.error.is_empty());
```

youjiali1995 (commented):
rest LGTM

sticnarf:
I don't have extra comments on it.

tests/failpoints/cases/test_kv_service.rs
```diff
@@ -36,3 +37,92 @@ fn test_kv_scan_memory_lock() {
     assert!(!resp.get_error().get_abort().is_empty());
     fail::remove("raftkv_async_snapshot_err");
 }
+
+#[test]
+fn test_scan_lock_push_async_commit() {
```
cfzjywxk, line 42: How about adding a case that scan requests reach first and async prewrite is executed with smaller `startTS`, checking the returned `min_commit_ts`.

MyonKeminta: In this case I did two scan lock requests, one is before the prewrite request, and another meets the memory lock of the prewrite request. I think the second one is equivalent to the case you said. How do you think?

cfzjywxk: I see, ignore this comment~

src/storage/mod.rs
MyonKeminta: Ah... their checking logic is a little different...

MyonKeminta: I just notice that everywhere else uses `Lock::check_ts_conflict` to check the memory lock, including `read_index`'s logic. It ignores LockType::Lock, However for scan_lock it can't be ignored.. so scan_lock can't be handled by follower read

sticnarf: Emm.. Currently, TiDB doesn't use follower read for ScanLock. Maybe we can ignore this problem now... (or returning error if the request wants to use replica read for safety)

MyonKeminta: Ah, in the last commit I set the replica read field to false directly so that it can return an not-leader error in this case, I think.

sticnarf:
LGTM

youjiali1995 (approved):

PR #9564, opened 2021-01-25
raftstore: force voter initialized

### LINKED ISSUE #3572: Potential two different leaders on a same term

BusyJay:

Issue Number: close #3572 

Problem Summary:

Not saving hardstate of uninitialized peer can cause two leader at a term. This PR adds a check to deny promoting uninitialized learner to voter. So every voting peer should be initialized.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- require voter be initialized to fix potential violation of election safety

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -2324,8 +2324,22 @@ where
                         cp
                     ));
                 }
+                (ConfChangeType::AddNode, PeerRole::Voter) => {
+                    let pr = match current_progress.get(peer.get_id()) {
+                        Some(pr) => pr,
+                        None => {
+                            return Err(box_err!("{} invalid conf change request: {:?}, voter should be added as learner first.", self.tag, cp));
+                        }
+                    };
+                    if pr.matched == 0 {
+                        return Err(box_err!(
+                            "{} reject adding uninitialized peer {:?}",
+                            self.tag,
+                            cp
+                        ));
+                    }
+                }
                 (ConfChangeType::RemoveNode, _)
-                | (ConfChangeType::AddNode, PeerRole::Voter)
                 | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {}
                 _ => {
                     return Err(box_err!(
--- a/components/test_raftstore/src/pd.rs
+++ b/components/test_raftstore/src/pd.rs
@@ -1017,6 +1017,12 @@ impl TestPdClient {
     }
 
     pub fn must_add_peer(&self, region_id: u64, peer: metapb::Peer) {
+        if peer.get_role() == PeerRole::Voter {
+            let mut learner = peer.clone();
+            learner.set_role(PeerRole::Learner);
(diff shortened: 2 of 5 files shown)


components/raftstore/src/store/peer.rs
```diff
@@ -2324,8 +2324,22 @@ where
...
                     ));
                 }
+                (ConfChangeType::AddNode, PeerRole::Voter) => {
+                    let pr = match current_progress.get(peer.get_id()) {
+                        Some(pr) => pr,
+                        None => {
+                            return Err(box_err!("{} invalid conf change request: {:?}, voter should be added as learner first.", self.tag, cp));
+                        }
+                    };
+                    if pr.matched == 0 {
```
hicqu, line 2334: If the peer hasn't send any messages to the leader  in the leader's current term, `pr.matched` should be  0. However the peer can have been already initialized.

components/test_raftstore/src/pd.rs
```diff
@@ -1017,6 +1017,12 @@ impl TestPdClient {
     }
 
     pub fn must_add_peer(&self, region_id: u64, peer: metapb::Peer) {
+        if peer.get_role() == PeerRole::Voter {
+            let mut learner = peer.clone();
+            learner.set_role(PeerRole::Learner);
+            self.add_peer(region_id, learner.clone());
+            self.must_have_peer(region_id, learner);
+        }
```
NingLin-P, line 1025: Perhaps should check whether the peer already a learner first? Also, need to change `must_joint_confchange` too.

components/raftstore/src/store/peer.rs
gengliqi: Yes. But it seems there is no better way to recognize this situation.

NingLin-P:
Perhaps, we should add a warning (or error) log when an uninitialized voter do try to vote for other peers for better debugging?

components/raftstore/src/store/peer.rs
hicqu: How about comment it?

gengliqi:
This PR can not make all voters initialized. 
There is another bug that the voter may uninitialized if it comes from splitting.
See https://github.com/tikv/tikv/issues/9579 for more details.

PR #9576, opened 2021-01-27
[WIP] cloud: organize cloud provider code into separate crates

gregwebs:

We would like to have clear interfaces and organization for cloud providers. 

This PR
* moves all cloud provider code into separate crates (cloud/aws and cloud/gcp)
* demonstrates how a cloud provider can be loaded dynamically
* allows cloud providers to be switched off with feature flags (cloud-aws and cloud-gcp)

To avoid cyclical crate dependencies, an additional "export" crate is added. These crates should be imported by users of this functionality.
* external_storage/export
* encryption/export

Actual code changes in this PR are relatively few where they were required to break up dependencies.
There are no changes to how existing code works. 

This PR is marked as WIP because some additional manual testing is needed and kvproto changes should be merged.

### Check List

Tests

- Unit test

Side effects

- Performance: there is some new additional copying but this is all for relatively rarely used control plane structures so this shouldn't be noticeable
- Breaking backward compatibility: all changes are backwards compatible with existing TiKV.

### Release note 

Organize cloud provider code into separate crates.


brson:
I just have some comments:

_Crates in subdirectories_. I find that any crate that doesn't live in `components` is basically hidden and doesn't receive the same attention as if they are at the components level.

This applies to the two `_export` crates, and the two `cloud` implementations. My preference would be to have them all at the top `components` level, at least for now, where most crates are at this same level in the directory structure. If there was some organized push to clean up `components` and make it a multi-level organization I would not be opposed.

_The `*_export` crate naming convention`_, as far as I can tell, is new to tikv. Here `export` means the crate that is depended on by the application, where non-export is the interface used by multiple  implementors; as cloud/aws depends on and implements things from encryption and external_storage, while tikv depends on encryption_export and external_storage_export.

Here it looks like encryption_export and external_storage_export are providing constructors over multiple implementations of encryption and export.

TiKV's conventions are pretty disorganized, but I note that the engine traits, the systems I'm familiar with most, are organized differently: `engine_traits` is the interface implemented by many crates, which here corresponds to `encryption` and `external_storage`. Engines don't currently have an equivalent to the "export" crates here, but there is similar functionality inside the "engine_test" crate, that constructs every type of engine, and which I was planning to move to its own crate eventually, perhaps called "engine_ctors".

I'm not sure if there are other instances of this "interface crate - implementation crates - constructor crate" pattern in tikv, but it would be best for them all to follow the same pattern.

I am not _opposed_ to the naming pattern here: unadorned name is the interface, "_export" is the constructor; but I do think it is not particularly clear what eachs's role is from the naming.

_Implementation crates_. It looks to me like the implementations of the "encryption" and "external_storage" live inside `cloud/aws` and `cloud/gcp`. Maybe there are dependency-based reasons that the implementations of encryption and external_storage have to be combined into a single "cloud" crate, but I think it would be nice if the implementations of each interface have their own crate, `encryption_aws`, `encryption_gcp`, etc.

_Dynamic loading_. `external_storage_export` has code for dynamically loading an external storage. It does seem to just be a demo of what's possible, since there's no code here to build such a library. Have you successfully built a dynamic library and run it this way? What kind of problems did you run into? There are some important ABI, and other, considerations when doing this, and I don't know offhand what kind of guarantees Rust provides for it, or best practices for it.

Aside: the name of the dylib `external_storage` is loading is `libtikv_cloud*.so`. It seems like there's some abstraction missing here since the "cloud" concept seems to exist at a higher level than both external storage and encryption. Maybe some "cloud" crate needs to be responsible for the dynamic loading.

Those are just my thoughts for now. I don't know exactly what I'd want to change here, so I'm hoping others will review and comment.

gregwebs:
@brson Thanks for your careful review!

> Crates in subdirectories. I find that any crate that doesn't live in components is basically hidden and doesn't receive the same attention as if they are at the components level.

What do you think about putting the cloud crates at the top-level (I would suggest "cloud-aws" rather than just "aws" so they get grouped together). And then leaving the export crates as sub crates. The export crates are designed to have as little code as possible.

> I am not opposed to the naming pattern here: unadorned name is the interface, "_export" is the constructor; but I do think it is not particularly clear what eachs's role is from the naming.

I am open to changing the naming. Another possibility I thought of was calling the interface "_lib" and removing the "_export". Or using both "_lib" and "_export".

I have a preference for a broader name like export over ctor because the export crate can potentially do more than just constructors. Generally speaking it is a crate to help connect a library crate with the outside world. One thing I also tried to do in the external_storage_export crate was to get rid of protocol buffers there (the cloud provider crate though still has to import them to define its own conversion) so that we end up with a library that has logic operating independently of the outside world (for control plane operations there isn't much concern about efficiency).

>  think it would be nice if the implementations of each interface have their own crate, encryption_aws, encryption_gcp, etc.

I am not entirely opposed to this, but it seems to me to be a dramatic change from the status quo that may be too far in the other direction. I am not sure what the upside is. The downside is that
* it will contribute to components pollution if every crate is at the top level
* more Cargo manifest duplication to maintain
* common functionality for a cloud needs its own crate. We do already have a rusoto_util crate for AWS, but we would end up needing yet another crate for GCP and every cloud provider as well.

> Dynamic loading. external_storage_export has code for dynamically loading an external storage. It does seem to just be a demo of what's possible

Yes, this is definitely a demo at best right now. I have successfully built the cloud providers as a cdylib but not actually tested it yet. The point right now is mostly to show that putting code in a separate crate facilitates being able to do this in the future.

gregwebs:
@brson personally I would favor organizing components with hierarchy. I think the components are already at a point of there being too many in one directory, and we seem to only be adding more. Just switching `tidb_*` to `tidb/`, `engine_` to `engine/`, and `test_` to `test/` would be good enough for now. I would be happy to send a PR for that.

components/encryption/src/master_key/kms.rs
```diff
@@ -38,168 +25,11 @@ impl KeyId {
     }
 }
 
-pub struct AwsKms {
```
gregwebs, line 41: This part is now in the cloud aws crate as kms.rs

components/cloud/aws/src/kms.rs
```diff
@@ -0,0 +1,278 @@
+use std::future::Future;
```
gregwebs, line 1: This code is taken unchanged from the encryption kms.rs file. https://github.com/tikv/tikv/pull/9576/files#diff-0f18341af3dc671328d8f26275bacc9676378054ce65b6eba9e561d69511b4c9

components/cloud/aws/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[dependencies]
+bytes = "0.5"
+encryption = { path = "../../encryption", default-features = false }
+futures = "0.3"
+futures-util = { version = "0.3", default-features = false, features = ["io"] }
+# This is only a dependency to vendor openssl for rusoto. It's not clear exactly
+# how openssl is built for tikv, but it seems to be controlled by grpcio. This
+# makes `cargo test -p external_storage` link correctly.
+grpcio = { version = "0.7", default-features = false, features = ["openssl-vendored"] }
```
yiwu-arbug, line 26: does changing to depend on openssl-sys directly work?
```
openssl-sys = { version = "0.10", optional = true, features = ["vendored"] }
```

yiwu-arbug:
Personally I also prefer adhere to what engine_trait doing now. We can have a cloud_trait crate (instead of having cloud_* crates depend on encryption and external_storage) that abstracting cloud functionalities like blob store and KMS, then cloud_aws and cloud_gcp that implementing the interfaces, then external_storage and encryption crate depends on maybe a crate call cloud_export for dynamic or statically loading cloud libs and providing ctor.

gregwebs:
Currently what I have done is try to use the fewest crates possible without having a cycle:
encryption -> cloud_aws -> encryption_export

@yiwu-arbug I am trying to understand your suggestion. AFAICT it creates a cycle:
encryption -> cloud -> cloud_aws -> cloud_export -> encryption

yiwu-arbug:
@gregwebs (copy from offline chat) My proposal is like
encryption
       | -> cloud_export -> cloud_aws -> cloud_trait
       | -> cloud_trait

For example KMS API is independent of encryption, and can be defined in cloud_trait.
Basically rename your encryption_export to cloud_trait, and also move things specific to encryption back to encryption crate, leaving only the KMS interface.
In my proposal cloud_export contain only ctor for the underlying implementations.

gregwebs:
If I understand your suggestion is to add a cloud_trait crate in which the KMS interface would reside separately from encryption. However KMS does not imply a cloud provider. It could instead be for example a self operated instance of Hachicorp Vault.
If we expanded the definition of cloud to mean any external system that requires a network request then this would make more sense.
Even still, I think it is presumptuous to think that a functionality that we use the cloud for now such as KMS will always be handled by a network call to an external system. For example I advocated an encryption design where TiKV would ask PD for a data key and PD would handle all master key management.

yiwu-arbug:
> If I understand your suggestion is to add a cloud_trait crate in which the KMS interface would reside separately from encryption. However KMS does not imply a cloud provider. It could instead be for example a self operated instance of Hachicorp Vault.
> If we expanded the definition of cloud to mean any external system that requires a network request then this would make more sense.
> Even still, I think it is presumptuous to think that a functionality that we use the cloud for now such as KMS will always be handled by a network call to an external system. For example I advocated an encryption design where TiKV would ask PD for a data key and PD would handle all master key management.

Now I understand your concern. But I still prefer not to have the cloud code depend on any encryption or external_storage specific logic. Both of encryption and external_storage can keep their existing abstraction to handle, for example, Hachicorp Vault or a KMS service by PD.

yiwu-arbug:
> > If I understand your suggestion is to add a cloud_trait crate in which the KMS interface would reside separately from encryption. However KMS does not imply a cloud provider. It could instead be for example a self operated instance of Hachicorp Vault.
> > If we expanded the definition of cloud to mean any external system that requires a network request then this would make more sense.
> > Even still, I think it is presumptuous to think that a functionality that we use the cloud for now such as KMS will always be handled by a network call to an external system. For example I advocated an encryption design where TiKV would ask PD for a data key and PD would handle all master key management.
> 
> Now I understand your concern. But I still prefer not to have the cloud code depend on any encryption or external_storage specific logic. Both of encryption and external_storage can keep their existing abstraction to handle, for example, Hachicorp Vault or a KMS service by PD. That way the cloud lib can be a truly standalone library.

@brson do you have any suggestion?

gregwebs:
> I still prefer not to have the cloud code depend on any encryption or external_storage specific logic.

The cloud code needs to depend on the types from encryption, etc that are in the interface. This is what it does now. There is no dependency on logic from encryption, etc other than the logic that is inherent to using the types of the interface.

It is possible to duplicate both the encryption interface and type needed by it in a cloud crate by importing it and re-exporting it or aliasing it. For example, define a trait `CloudKmsProvider` with `KmsProvider` imported as a dependent trait that is exactly the same as `KmsProvider`. I think aliases like this trait alias create boilerplate will end up being confusing busywork without any benefit. However, I can see a benefit to doing a straight-up re-export from a cloud crate: it would minimize boilerplate (not create any new names so that searching would work the same) and it would show all the interfaces and types used by a cloud provider.

gregwebs:
@yiwu-arbug I added an additional commit which adds a cloud crate. Currently it is pass through but maybe it will make sense to have a cloud specific adapter in the future or to otherwise share some code among cloud providers.

gregwebs:
@yiwu-arbug convinced me to try to reverse the dependency flow from the cloud crate so that encryption will import it. Same for external_storage. This may end up requiring duplication of the trait and types, but that code is small.

yiwu-arbug (commented):
This PR is becoming huge. Can we split it into smaller ones, for easier review? Thanks.

gregwebs:
> This PR is becoming huge. Can we split it into smaller ones, for easier review? Thanks.

I will see what I can do. I could split encryption and external_storage into separate PRs, but one will still depend on the other unless I can first commit a useless PR that adds some empty structure. I usually try to clean up my PRs into separately reviewable commits (which I think everyone usually ignores).

gregwebs:
@yiwu-arbug I sent a PR to split out one small change: https://github.com/tikv/tikv/pull/9629

gregwebs:
closing in favor of newer PRs

Issue #7163, opened 2020-03-19
Restoring a subrange of backed up rawkv data may import keys that are out of range


MyonKeminta:
### What version of TiKV are you using?

Master

### What operating system and CPU are you using?

(Not related)

### Steps to reproduce

The bug was found in code so the following reproduce steps have not been tested yet.

1. Write following raw keys to TiKV: "a", "b", "c", "c\x00", "c\x00\x00", "d"
2. backup rawkv range ["a", "e")
3. Restore subrange ["b", "c")
4. The result is expected to be: "b". However in the current implementation, it may be: "b", "c", "c\x00", "c\x00\x00"

### Some Explaination

1. When BR restoring a subrange of rawkv data, it sents the whole SSTs that contains the needed data and changes the range in SstMeta to tell TiKV to ignore the data outside the range.

2. The restoring logic of RawKv in TiKV is the same as restoring table data. The range passed through the SstMeta is inclusive ([start_key, end_key]), and it uses adds 1 to the last byte of endKey to make it exclusive. This way it suits perfectly with TiDB and the rewrite mechanism, However for rawkv and plain transactional kv it may doesn't works properly.

PR #7196, opened 2020-03-20
Fix the issue that end key of a partial rawkv-restore range is inclusive

MyonKeminta:
###  What have you changed?

Partially fixes https://github.com/tikv/tikv/issues/7163 , but it still doesn't work for restoring plain transactional kv (without TiDB).

This PR adds is_raw_kv field to DownloadRequest to distinguish rawkv restore requests, and adds end_key_exclusive field to SstMeta so that TiKV knows whether the endkey need to be included.l

###  What is the type of the changes?

- Bugfix (a change which fixes an issue)

###  How is the PR tested?

- Unit test
- Manual test (add detailed scripts or steps below)

###  Does this PR affect documentation (docs) or should it be mentioned in the release notes?

(I'm not sure if it's necessary)
* Fix the issue that when restoring partial of backed up raw kv data, TiKV may write out-of-range keys.

###  Refer to a related PR or issue link (optional)

- [x] kvproto https://github.com/pingcap/kvproto/pull/581
- [ ] Corresponding PR of BR

### Release note

* No release note

PR #9583, opened 2021-01-27
Fix the issue that end key of a partial rawkv-restore range is inclusive (#7196)

MyonKeminta:
cherry-pick #7196 to release-4.0

---

###  What have you changed?

Partially fixes https://github.com/tikv/tikv/issues/7163 , but it still doesn't work for restoring plain transactional kv (without TiDB).

This PR adds is_raw_kv field to DownloadRequest to distinguish rawkv restore requests, and adds end_key_exclusive field to SstMeta so that TiKV knows whether the endkey need to be included.l

###  What is the type of the changes?

- Bugfix (a change which fixes an issue)

###  How is the PR tested?

- Unit test

###  Does this PR affect documentation (docs) or should it be mentioned in the release notes?

(I'm not sure if it's necessary)
* Fix the issue that when restoring partial of backed up raw kv data, TiKV may write out-of-range keys.

###  Refer to a related PR or issue link (optional)

- [x] kvproto https://github.com/pingcap/kvproto/pull/581
- [ ] Corresponding PR of BR

### Release note

* No release note


3pointer:
LGTM

PR #9587, opened 2021-01-27
storage: return invalid after seeking out of bound

youjiali1995:

Issue Number: close https://github.com/tikv/tikv/issues/3378

Problem Summary:

`Cursor` uses `min_key/max_key` to reduce the cost of seeking when the sought key is out of bound. Even though I think this optimization can't be widely applicable, I still keep it. However, `valid()` may return true even if `seek()` returns false which is dangerous. We should make them consistent.

Proposal: [xxx](url) 

What's Changed:

Use the `in_bound` field to indicate that the previous `seek()` has returned false and `valid()` returns false in this case.

### Check List 

Tests 

- Unit test

### Release note 
* No release note


src/storage/kv/cursor.rs
```diff
@@ -119,6 +125,7 @@ impl<I: Iterator> Cursor<I> {
     /// around `key`, otherwise you should use `seek` instead.
     pub fn near_seek(&mut self, key: &Key, statistics: &mut CfStatistics) -> Result<bool> {
         assert_ne!(self.scan_mode, ScanMode::Backward);
+        self.in_bound = true;
```
sticnarf, line 128: I prefer to add unit test cases for calling `near_seek` directly because of the logics added here.

tabokie:
Ping @youjiali1995

PR #9593, opened 2021-01-28
resolve: detect tombstone correctly

### LINKED ISSUE #9590: raft client doesn't recognize tombstone store
## Bug Report

pd client will filter tombstone store and return error

https://github.com/tikv/tikv/blob/6cd51f9d07aa60fea651dea7ecb6a39c7425273a/components/pd_client/src/client.rs#L343-L347

but raft client depends on the store meta to recognize tombstone store correctly, so it will keep failing.

https://github.com/tikv/tikv/blob/6cd51f9d07aa60fea651dea7ecb6a39c7425273a/src/server/resolve.rs#L100-L103

BusyJay:

Issue Number: close #9590

Problem Summary:

PD client filter tombstone store and return error instead. Resolver should recognize the error and handle it correctly.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- Fix repeated tombstone logs when sunset nodes


src/server/raft_client.rs
```diff
@@ -800,25 +800,27 @@ where
         let (s, pool_len) = {
             let mut pool = self.pool.lock().unwrap();
             if pool.tombstone_stores.contains(&store_id) {
+                let pool_len = pool.connections.len();
+                drop(pool);
```
hicqu, line 804: Is it necessary to drop it explicitly?

src/server/resolve.rs
```diff
@@ -82,7 +81,14 @@ where
 
     fn get_address(&self, store_id: u64) -> Result<String> {
         let pd_client = Arc::clone(&self.pd_client);
-        let mut s = box_try!(pd_client.get_store(store_id));
+        let mut s = match pd_client.get_store(store_id) {
+            Ok(s) => s,
+            Err(pd_client::Error::StoreTombstone(_)) => {
+                RESOLVE_STORE_COUNTER_STATIC.tombstone.inc();
+                return Err(box_err!("store {} has been removed", store_id));
```
hicqu, line 88: How about comment here about where the error is handled so that other guys won't make mistakes again?

hicqu:
rest LGTM. Please fix the clippy warning.

hicqu (approved):

src/server/raft_client.rs
BusyJay: I drop it earlier to reduce the time to hold the lock.

src/server/raft_client.rs
```diff
@@ -800,25 +800,27 @@ where
...
+                        store_id,
+                        queue: queue.clone(),
+                        builder: self.builder.clone(),
+                    };
+                    self.future_pool
+                        .spawn(start(back_end, conn_id, self.pool.clone()));
+                    queue
+                })
+                .clone();
+            (conn, pool.connections.len())
```
hunterlxt, line 823: No changes seem to be found for me

hunterlxt (approved):
Most LGTM

PR #9594, opened 2021-01-28
storage: yield in kv scan to avoid occupying cpu resource for too long

### LINKED ISSUE #9188: The unified pool cannot schedule kv_scan, raw_scan and analyze requests well

sticnarf:

Issue Number: close #9188

This PR makes scan interface asynchronous. Then, we can reschedule a long scan task during its execution. It should help yatp schedule these tasks better. So these long scan tasks will not affect normal requests greatly.

### Check List 

Tests 

### Release note 

```
Schedule long scan requests better when using the unified thread pool for the storage module.
```

### DIFF
--- a/src/coprocessor/mod.rs
+++ b/src/coprocessor/mod.rs
@@ -183,40 +183,6 @@ impl ReqContext {
     }
 
     pub fn build_task_id(&self) -> u64 {
-        const ID_SHIFT: u32 = 16;
-        const MASK: u64 = u64::max_value() >> ID_SHIFT;
-        const MAX_TS: u64 = u64::max_value();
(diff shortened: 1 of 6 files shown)


sticnarf:
Already implemented by other PR

PR #9595, opened 2021-01-28
*: Add TTL support for rawkv

Connor1996:

Issue Number:  https://github.com/tikv/tikv/issues/9612 

Problem Summary: Add TTL support for rawkv

What's Changed:

- Add a `ttl` field for raw put requests
- Calculate a `expire_ts` on storage level, and bring it with `PutCmd` of `RaftCmdRequest`
- Append the `expire_ts` to the value when applying the put cmd
- Strip the `expire_ts` from values when doing get and scan at `RegionSnapshot` level
- Add a `TTLCompactionFilter` to filter the invalidated value
- Add a `TTLCheckAndCompact` task in cleanup worker to compact the invalidated SSTs in time.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

- Add TTL support for rawkv


components/engine_rocks/src/compact.rs
```diff
@@ -92,14 +86,33 @@ impl CompactExt for RocksEngine {
...
+        let cf_opts = db.get_options_cf(handle);
+        let output_level = output_level.unwrap_or(cf_opts.get_num_levels() as i32 - 1);
+        let output_compression = cf_opts
+            .get_compression_per_level()
+            .get(output_level as usize)
+            .cloned()
+            .unwrap_or(DBCompressionType::No);
+        let output_file_size_limit = cf_opts.get_target_file_size_base() as usize;
+
+        let max_subcompactions = cmp::min(num_cpus::get(), 32);
```
Little-Wallace, line 109: May it be too large?

components/raftstore/src/store/worker/compact.rs
```diff
@@ -169,6 +200,49 @@ where
...
+                    return;
+                }
+
+                let mut files = Vec::new();
+                for (file_name, v) in collection.iter() {
+                    let prop = match TTLProperties::decode(&v.user_collected_properties()) {
+                        Ok(v) => v,
+                        Err(_) => continue,
+                    };
+                    if prop.max_expire_ts <= current_ts {
```
Little-Wallace, line 232: I think we do not need to compact every file which concludes an expired key, because if this region is updated frequently, the file may be compacted automatic. We only need to compact files which has not been compacted for a long time. Maybe we could add the time one or two hour?

src/storage/mod.rs
```diff
@@ -1444,6 +1376,182 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
     }
 }
 
+enum RawStore<S: Snapshot> {
```
Little-Wallace, line 1379: This file is too large.... We shall move this code into another independent file

src/storage/ttl.rs
```diff
@@ -0,0 +1,179 @@
...
+            s,
+            current_ts: UnixSecs::now().into_inner(),
+        }
+    }
+}
+
+impl<S: Snapshot> Snapshot for TTLSnapshot<S> {
+    type Iter = TTLIterator<S::Iter>;
+
+    fn get(&self, key: &Key) -> Result<Option<Value>> {
```
Little-Wallace, line 51: Please check it by  `find_valid_value`.

src/storage/mod.rs
```diff
@@ -993,17 +988,22 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         cf: String,
         key: Vec<u8>,
         value: Vec<u8>,
+        ttl: u64,
         callback: Callback<()>,
     ) -> Result<()> {
         check_key_size!(Some(&key).into_iter(), self.max_key_size, callback);
-
+        let mut m = Modify::Put(Self::rawkv_cf(&cf)?, Key::from_encoded(key), value);
+        if self.enable_ttl {
```
Little-Wallace, line 996: Do we need to add another key space (prefix) when `enable_ttl` ? Or we shall add in `RawKVClient`?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +23,30 @@ pub fn check_key_in_range(
...
+    value.encode_u64(expire_ts).unwrap();
+}
+
+pub fn get_expire_ts(value_with_ttl: &[u8]) -> Result<u64> {
+    let len = value_with_ttl.len();
+    if len < number::U64_SIZE {
+        return Err(Error::Codec(codec::Error::ValueLength));
+    }
+    let mut ts = &value_with_ttl[len - number::U64_SIZE..];
+    Ok(number::decode_u64(&mut ts)?.into())
```
Little-Wallace, line 37: We must change `0` to `u64::MAX`.

components/engine_rocks/src/compact.rs
Connor1996: seems yes, don't know why it's set that large before, will check

src/storage/mod.rs
Connor1996: will do

src/storage/ttl.rs
Connor1996: it's checked in `map_value`

components/raftstore/src/store/worker/compact.rs
Connor1996: note that it's comparing to the `max_expire_ts`, so the file is compacted only when all keys are expired. Also, considering the interval of checking SSTs' TTL may be hours(24h by default), I don't think it won't be compacted so frequently.

Little-Wallace: But if user write a log of keys today. The time in tomorrow will trigger a lot of manual compaction. However, we only need to wait a hour, maybe most of them will be compacted automatic. And then we only need to compact few regions.

components/engine_traits/src/util.rs
Connor1996: let's keep it and check outside (forget to check 0 in `map_value`)

src/storage/mod.rs
Connor1996: Have discussed with @BusyJay before, adding the key prefix is wasteful. If user wants to use keyspace, it's the user's duty to write the key with the prefix.

components/raftstore/src/store/worker/compact.rs
Connor1996: why most of them will be compacted automatically tomorrow?

src/server/ttl_checker.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let mut round_start_time = Instant::now();
+        let mut key = vec![];
+        loop {
+            let (tx, rx) = mpsc::channel();
+            let res = self.region_info_provider.seek_region(
+                &key,
+                Box::new(move |iter| {
+                    for info in iter {
+                        let _ = tx.send(Some(info.region.clone()));
```
Little-Wallace, line 48: Could we seek more region once?

Connor1996: of course, maybe 10?

PR #9597, opened 2021-01-28
storage, gc_worker: Update max_ts when invoking scan_lock and register_lock_observer (#9558)

### LINKED ISSUE #9526: incompatibility between gc and async commit

ti-srebot:
cherry-pick #9558 to release-5.0-rc
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/9597
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/9597:release-5.0-rc-579f5a4943a3
```

---

Issue Number: close #9526 

Problem Summary: In some very-rare corner cases, if prewrite requests of async commit transactions arrives  so late that when it arrives, GC safepoint already exceeds its start_ts, it's possible that GC breaks the transaction's atomicity.

What's Changed:

This PR mades a few changes to fix the issue.

* Move `scan_lock` out of scheduler, and uses similar handling logic as other kv read requests like `get`, `batch_get`, `scan`, etc.
  * Including updating `max_ts` of concurrency manager and checking memory locks. This fixes the issue for non green gc.
* Add an optional `end_key` parameter for `scan_lock`.
* Update `max_ts` of concurrency manager and check memory locks when `register_lock_observer` is invoked. This fixes the issue for green gc.

### Related changes

- Though this PR fixes #9526 , but for non green gc, the scan_lock phase may not be very efficient since each scan_lock request will cause scanning memory lock from that region until the very end. Therefore we need an additional PR to TiDB to fill the `end_key` parameter of scan_lock reuqest.

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

* Fix an issue that in some very rare cases, GC may break the atomicity of async commit transactions.


BusyJay:
5.0-rc is not maintained anymore.

PR #9600, opened 2021-01-28
raftstore: Move engine_rocks to dev-dependencies

brson:

Part of engine abstraction https://github.com/tikv/tikv/issues/6402

Problem Summary:

This moves engine_rocks to a dev-dependency of raftstore, where only two test modules depend on it. It will no longer be possible to write code in raftstore that depends on rocksdb directly.

What's Changed:

### Check List 

Tests 

### Release note 

- No release note


Connor1996 (approved):
LGTM

PR #9607, opened 2021-01-30
copr: add feature json_quote

ZingLix:
Signed-off-by: ZingLix <me@zinglix.xyz>

Issue Number:  #5751  

Problem Summary: `JsonQuoteSig` did not have a vectorized version in TiKV.

What's Changed: Implemented `json_quote`

### Check List 

Tests 

- Unit test

### Release note

- Add copr feature: json_quote


components/tidb_query_expr/src/impl_json.rs
```diff
@@ -156,6 +156,47 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
     Ok(Some(Json::merge(jsons)?))
 }
 
+#[rpn_fn(nullable, raw_varg)]
```
Xuanwo, line 159: Maybe we can remove `nullable` and add `writer` here?

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -156,6 +156,47 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
...
+#[inline]
+fn json_quote(arg: &[ScalarValueRef]) -> Result<Option<Bytes>> {
```
Xuanwo, line 161: Maybe we can use `BytesRef` here directly?

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -156,6 +156,47 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
...
+    quote(arg[0].as_bytes())
+}
+
+fn quote(input: Option<BytesRef>) -> Result<Option<Bytes>> {
```
Xuanwo, line 165: Is this implementation the same as `strconv.Quote` in Golang?

ref: https://github.com/pingcap/tidb/blob/master/expression/builtin_json.go#L1122

ZingLix: It that ok to import serde here? I think this should be more reliable.

Xuanwo: Why we need to import `serde` here?

ZingLix: Use `serde_json::to_string()` to replace this  `quote`.

Xuanwo: Are they consistent with the TiDB side of the implementation?

components/tidb_query_expr/src/impl_json.rs
Xuanwo: We don't need `raw_varg` for `JSON_QUOTE` as it only accept a `string` input: https://dev.mysql.com/doc/refman/5.7/en/json-creation-functions.html#function_json-quote

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -156,45 +156,13 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
...
-            }
-            result.push(b'\"');
-            Ok(Some(result))
-        }
-        _ => Ok(None),
-    }
+fn json_quote(input: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
+    let str = String::from_utf8(input.to_vec()).map_err(tidb_query_datatype::codec::Error::from)?;
+    let result =
+        Bytes::from(serde_json::to_string(&str).map_err(tidb_query_datatype::codec::Error::from)?);
```
ZingLix, line 164: `serde_json::to_string(String)` should be same with `strconv.Quote()` since both of them want to convert a string to a valid string literals in json (quoted Go string literals for `Quote` actually).
But maybe there is any function in `codec::mysql::Json` to format strings so that we don't need to import `serde` here?
@Xuanwo

Xuanwo: We need to implement it like https://github.com/tikv/tikv/blob/master/components/tidb_query_datatype/src/codec/mysql/json/json_unquote.rs

Xuanwo: And I don't think import `serde_json` is a good idea.

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -159,10 +159,47 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
...
+}
+
+fn quote(bytes: BytesRef) -> Result<Option<Bytes>> {
+    let mut result = Vec::with_capacity(bytes.len() * 2 + 2);
+    result.push(b'\"');
+    for byte in bytes.iter() {
+        if *byte == b'\"' || *byte == b'\\' {
+            result.push(b'\\');
+            result.push(*byte)
+        } else if *byte == b'\x07' {
```
Xuanwo, line 172: Please take https://dev.mysql.com/doc/refman/8.0/en/json-modification-functions.html#json-unquote-character-escape-sequences a look.

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -159,10 +159,47 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
...
+            // \a alert
+            result.push(b'\\');
+            result.push(b'a');
+        } else if *byte == b'\x08' {
```
Xuanwo, line 176: Maybe we need to export https://github.com/tikv/tikv/blob/master/components/tidb_query_datatype/src/codec/mysql/json/json_unquote.rs#L12 ?

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -557,8 +594,8 @@ mod tests {
                 Some(r#""hello,\"宽字符\",world""#),
```
Xuanwo, line 594: BTW, maybe we need to encode UTF-8 bytes for Unicode value ?

components/tidb_query_expr/src/impl_json.rs
ZingLix: yep, but it seems to be different from `strconv` [here](https://golang.org/src/strconv/quote.go?s=2262:2273#L82)

Xuanwo: OK, Let's keep the same with TiDB for now.

components/tidb_query_expr/src/impl_json.rs
ZingLix: `'\x0A'` is same with `b'\n'`.

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -156,6 +156,52 @@ pub fn json_merge(args: &[Option<JsonRef>]) -> Result<Option<Json>> {
...
+fn json_quote(input: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
+    Ok(writer.write(quote(input)?))
+}
+
+fn quote(bytes: BytesRef) -> Result<Option<Bytes>> {
+    let mut result = Vec::with_capacity(bytes.len() * 2 + 2);
+    result.push(b'\"');
+    for byte in bytes.iter() {
+        if *byte == b'\"' || *byte == b'\\' {
+            result.push(b'\\');
```
Xuanwo, line 170: Please considering use `writer.partial_write` to reuse the buffer.

components/tidb_query_expr/src/impl_json.rs
Xuanwo: I'm lacking some understanding on `json_quote`, ping @andylokandy to take a look.

andylokandy:
I'm not familiar with json_quote either... Would mind having a look @ichn-hu?

ichn-hu:
MySQL's documentation doesn't clearly state the specification of the behavior of `json_quote`, so I think it is okay to firstly match TiDB's behavior.

andylokandy (approved):

andylokandy:
@Xuanwo  PATL~

Xuanwo (approved):

PR #9618, opened 2021-02-01
copr: add truncate decimal & truncate uint

wangggong:
Signed-off-by: wangggong <793160615@qq.com>

Issue Number: #9016 

Problem Summary:

What's Changed:

### Check List 

Tests 

- Unit test

Side effects

### Release note 

No release note.


components/tidb_query_expr/src/impl_math.rs
```diff
@@ -460,6 +483,28 @@ pub fn truncate_int_with_uint(arg0: &Int, _arg1: &Int) -> Result<Option<Int>> {
     Ok(Some(*arg0))
 }
 
+#[inline]
+#[rpn_fn]
+pub fn truncate_uint_with_int(arg0: &Int, arg1: &Int) -> Result<Option<Int>> {
+    let x = arg0;
+    let d = arg1;
```
Xuanwo, line 490: Maybe we can remove this assignment, as they add extra burden for code reading.

components/tidb_query_expr/src/impl_math.rs
```diff
@@ -460,6 +483,28 @@ pub fn truncate_int_with_uint(arg0: &Int, _arg1: &Int) -> Result<Option<Int>> {
...
+    Ok(Some(if *d >= 0 {
+        *x
+    } else if *d <= -(I64_TEN_POWS.len() as i64) {
```
Xuanwo, line 493: Logic in tidb:

```go
	if d == mathutil.MinInt {
		return 0, false, nil
	}
```

components/tidb_query_expr/src/impl_math.rs
wangggong: OK.

done

components/tidb_query_expr/src/impl_math.rs
wangggong: An `u64` is between `[0, 1 << 64)`, which means an `u64` is not longer than `20` digits.

Thus we can just take at most `20` digits in consideration. This is the reason of the original logic in `tikv`.

See also [here](https://github.com/tikv/tikv/blob/1a88f12ebf50064b992fc5efc7e7f55795210521/components/tidb_query_normal_expr/src/builtin_math.rs#L424)

Xuanwo: Maybe we can introduce a constant like something called `MAX_DIGIT_LENGTH` instead of depends on `I64_TEN_POWS.len()` ?

components/tidb_query_expr/src/impl_math.rs
```diff
@@ -493,6 +530,28 @@ fn truncate_real(x: Real, d: i32) -> Real {
...
+#[inline]
+#[rpn_fn]
+pub fn truncate_decimal_with_int(arg0: &Decimal, arg1: &Int) -> Result<Option<Decimal>> {
+    let d = if *arg1 >= 0 {
+        *arg1.min(&127) as i8
+    } else {
+        *arg1.max(&-128) as i8
+    };
+
+    let res: codec::Result<Decimal> = arg0.to_owned().round(d, RoundMode::Truncate).into();
```
Xuanwo, line 542: - Do we need to use `to_owned()` here? `Decimal` is ok to `Copy`.
- Can we use `into()?` directly here?

skyzh (approved):

components/tidb_query_expr/src/impl_math.rs
Xuanwo: ping @wangggong to take a look.

components/tidb_query_expr/src/impl_math.rs
wangggong: okay, let me remove it.

done

components/tidb_query_expr/src/impl_math.rs
wangggong: 1. Yes, `Decimal` is ok to `Copy`. I removed  it.
2. If the code just like

```rust
let res: codec::Result<Decimal> = arg0.into();
```

The error will be raised:

```
^^^^ the trait `std::convert::From<&tidb_query_datatype::codec::data_type::Decimal>` is not implemented for `std::result::Result<tidb_query_datatype::codec:
 :data_type::Decimal, tidb_query_datatype::codec::Error>`
```

Thus here `round(d, RoundMode::Truncate)` is not removed now.

Xuanwo: What I mean is:

```
let res = arg0.round(d, RoundMode::Truncate).into()?; 
```

I don't know whether or not this can be compiled.

wangggong:
```rust
let res = arg0.round(d, RoundMode::Truncate).into()?;  // This will raise error: error[E0283]: type annotations needed
// let res: Decimal = arg0.round(d, RoundMode::Truncate).into()?;  // This will also raise the error, cause `into` cannot infer type.
```

Thus here the code will not be modified.

Xuanwo: OK, I got it.

Xuanwo (approved):

PR #9619, opened 2021-02-01
Fix redundant closure warning

gregwebs:

```
warning: try not to call a closure in the expression where it is declared
    --> components/raftstore/src/store/fsm/apply.rs:3220:9
     |
3220 |         (|| fail_point!("apply_on_handle_snapshot_sync", |_| { need_sync = true }))();
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try doing something like: `{}`
     |
     = note: `#[warn(clippy::redundant_closure_call)]` on by default
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_call
```

remove redundant closure

### Check List

Tests

Compile after change and warning does not occur.

### Release note

* None


yiwu-arbug:
cases::test_pending_peers::test_pending_snapshot failing. Is it related?

BusyJay:
The closure is necessary in test. It means execute something in the tests. How about adding `#[cfg(feature = failpoint)]` before the closure?

gregwebs:
@BusyJay thanks, that worked! PR is updated.

I guess clippy does not understand macros. I did not understand fail point macros either, but now I see the closure is to deal with the fact that the macro always does an early return. So a better fix might be to support a non-returning form of the macro.

BusyJay (approved):

yiwu-arbug (approved):

PR #9623, opened 2021-02-03
copr: implement find_in_set considering collation

### LINKED ISSUE #8986: copr: UTF8 functions should consider collation

TszKitLo40:

Issue Number: close #8986  

Problem Summary:

What's Changed:
implement find_in_set considering collation
### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No release note

### DIFF
--- a/components/tidb_query_expr/src/impl_string.rs
+++ b/components/tidb_query_expr/src/impl_string.rs
@@ -702,15 +702,15 @@ pub fn instr_utf8(s: BytesRef, substr: BytesRef) -> Result<Option<Int>> {
 
 #[rpn_fn]
 #[inline]
-pub fn find_in_set(s: BytesRef, str_list: BytesRef) -> Result<Option<Int>> {
+pub fn find_in_set<C: Collator>(s: BytesRef, str_list: BytesRef) -> Result<Option<Int>> {
     if str_list.is_empty() {
         return Ok(Some(0));
     }
 
-    let s = String::from_utf8_lossy(s);
+    use std::cmp::Ordering::*;
     let result = String::from_utf8_lossy(str_list)
         .split(',')
-        .position(|str_in_set| str_in_set == s)
+        .position(|str_in_set| C::sort_compare(str_in_set.as_bytes(), s).unwrap() == Equal)
         .map(|p| p as i64 + 1)
         .or(Some(0));
 
@@ -3253,18 +3253,25 @@ mod tests {
     #[test]
     fn test_find_in_set() {
         let cases = vec![
-            ("foo", "foo,bar", 1),
-            ("foo", "foobar,bar", 0),
-            (" foo ", "foo, foo ", 2),
-            ("", "foo,bar,", 3),
-            ("", "", 0),
-            ("a,b", "a,b,c", 0),
+            ("foo", "foo,bar", Collation::Utf8Mb4Bin, 1),
+            ("foo", "foobar,bar", Collation::Utf8Mb4Bin, 0),
+            (" foo ", "foo, foo ", Collation::Utf8Mb4Bin, 2),
+            ("", "foo,bar,", Collation::Utf8Mb4Bin, 3),
+            ("", "", Collation::Utf8Mb4Bin, 0),
+            ("a,b", "a,b,c", Collation::Utf8Mb4Bin, 0),
+            ("a,b", "A,B,C", Collation::Utf8Mb4GeneralCi, 0),
         ];
 
-        for (s, str_list, exp) in cases {
+        for (s, str_list, collation, exp) in cases {
             let s = Some(s.as_bytes().to_vec());
             let str_list = Some(str_list.as_bytes().to_vec());
             let got = RpnFnScalarEvaluator::new()
+                .return_field_type(
+                    FieldTypeBuilder::new()
+                        .tp(FieldTypeTp::LongLong)
+                        .collation(collation)
+                        .build(),
+                )
                 .push_param(s)
                 .push_param(str_list)
                 .evaluate::<Int>(ScalarFuncSig::FindInSet)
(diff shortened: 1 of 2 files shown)


components/tidb_query_expr/src/impl_string.rs
```diff
@@ -702,15 +702,15 @@ pub fn instr_utf8(s: BytesRef, substr: BytesRef) -> Result<Option<Int>> {
...
 #[inline]
-pub fn find_in_set(s: BytesRef, str_list: BytesRef) -> Result<Option<Int>> {
+pub fn find_in_set<C: Collator>(s: BytesRef, str_list: BytesRef) -> Result<Option<Int>> {
     if str_list.is_empty() {
         return Ok(Some(0));
     }
 
-    let s = String::from_utf8_lossy(s);
+    use std::cmp::Ordering::*;
     let result = String::from_utf8_lossy(str_list)
```
andylokandy, line 711: Please use https://docs.rs/bstr/0.2.14/bstr/trait.ByteSlice.html#method.split_str

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -702,15 +702,15 @@ pub fn instr_utf8(s: BytesRef, substr: BytesRef) -> Result<Option<Int>> {
...
 #[rpn_fn]
 #[inline]
-pub fn find_in_set(s: BytesRef, str_list: BytesRef) -> Result<Option<Int>> {
+pub fn find_in_set<C: Collator>(s: BytesRef, str_list: BytesRef) -> Result<Option<Int>> {
     if str_list.is_empty() {
         return Ok(Some(0));
     }
 
-    let s = String::from_utf8_lossy(s);
+    use std::cmp::Ordering::*;
```
andylokandy, line 710: It's not necessary to import it.

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -702,15 +702,15 @@ pub fn instr_utf8(s: BytesRef, substr: BytesRef) -> Result<Option<Int>> {
...
     let result = String::from_utf8_lossy(str_list)
         .split(',')
-        .position(|str_in_set| str_in_set == s)
+        .position(|str_in_set| C::sort_compare(str_in_set.as_bytes(), s).unwrap() == Equal)
```
andylokandy, line 713:
```suggestion
        .position(|str_in_set| C::sort_compare(str_in_set.as_bytes(), s).filter(|o| o == Ordering::Equal).is_some())
```

components/tidb_query_expr/src/impl_string.rs
TszKitLo40: So should we use `bstr::ByteSlice` instead of `String::from_utf8_lossy`?

TszKitLo40: So should we convert the `str_list` to bstr:ByteSilce instead of using `String::from_utf8_lossy`?

TszKitLo40: It seems that `bstr::ByteSlice` can't be constructed by a `BytesRef`.

components/tidb_query_expr/src/impl_string.rs
TszKitLo40: After `C::sort_compare(str_in_set.as_bytes(), s)`, `filter` can't be used.

components/tidb_query_expr/src/impl_string.rs
TszKitLo40: It seems that we need to, otherwise Ordering::Equal is not imported.

components/tidb_query_expr/src/impl_string.rs
mashenjun: Maybe use `ok()` to convert `Result<T>` to `Option<T>` before using `filter()`?

components/tidb_query_expr/src/impl_string.rs
andylokandy: you can add `use std::cmp::Ordering` to the top of the flie.

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -3253,18 +3254,25 @@ mod tests {
...
-            ("", "foo,bar,", 3),
-            ("", "", 0),
-            ("a,b", "a,b,c", 0),
+            ("foo", "foo,bar", Collation::Utf8Mb4Bin, 1),
+            ("foo", "foobar,bar", Collation::Utf8Mb4Bin, 0),
+            (" foo ", "foo, foo ", Collation::Utf8Mb4Bin, 2),
+            ("", "foo,bar,", Collation::Utf8Mb4Bin, 3),
+            ("", "", Collation::Utf8Mb4Bin, 0),
+            ("a,b", "a,b,c", Collation::Utf8Mb4Bin, 0),
+            ("a,b", "A,B,C", Collation::Utf8Mb4GeneralCi, 0),
```
andylokandy, line 3263: Please add a ci case that matches.

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -3253,18 +3259,26 @@ mod tests {
...
-            ("", "", 0),
-            ("a,b", "a,b,c", 0),
+            ("foo", "foo,bar", Collation::Utf8Mb4Bin, 1),
+            ("foo", "foobar,bar", Collation::Utf8Mb4Bin, 0),
+            (" foo ", "foo, foo ", Collation::Utf8Mb4Bin, 2),
+            ("", "foo,bar,", Collation::Utf8Mb4Bin, 3),
+            ("", "", Collation::Utf8Mb4Bin, 0),
+            ("a,b", "a,b,c", Collation::Utf8Mb4Bin, 0),
+            ("foo", "A,FOO,BAR", Collation::Utf8Mb4GeneralCi, 2),
+            ("b", "A,B,C", Collation::Utf8Mb4GeneralCi, 2),
```
Xuanwo, line 3269: Add cases for wide-width chars, like `测试` in `中文,测试`

TszKitLo40: OK, PTAL

PR #9624, opened 2021-02-03
sst_importer: make ingest reentrant

3pointer:
Signed-off-by: luancheng <luancheng@pingcap.com>

Problem Summary:

Before this fix, the ingested file won't execute `validate_sst_for_ingestion`(reset `global seqno`) before ingests when meta doesn't have crc32.https://github.com/tikv/tikv/blob/515df8d552cce67111991fc6b205ec2905716c2b/components/sst_importer/src/sst_importer.rs#L671-L677

Since crc32 and length are no longer check in `validate_sst_for_ingestion`. we can always do validate_sst_for_ingestion before ingests.

### Related changes

### Check List 

Tests 
- Integration test
before the fix: 
屏幕快照 2021-02-03 下午8 13 16

### Release note 

- Fix the issue that ingests operation cannot reentrant.


components/sst_importer/src/sst_importer.rs
```diff
@@ -664,21 +664,14 @@ impl ImportDir {
...
-        // we need re-compute length & crc32 and fill back to sstMeta.
-        if length != 0 {
-            if crc32 != 0 {
-                // we only validate if the length and CRC32 are explicitly provided.
-                engine.validate_sst_for_ingestion(cf, &path.clone, length, crc32)?;
-            }
-            IMPORTER_INGEST_BYTES.observe(length as _)
-        }
+        // length & crc32 is no longer used in validate_sst_for_ingestion any more.
+        engine.validate_sst_for_ingestion(cf, &path.clone, 0, 0)?;
```
kennytm, line 668: wait, what??? why?

kennytm, line 668: perhaps we should just apply the TODO and rename `validate_sst_for_ingestion` to `reset_global_seq` to avoid the confusion why you pass 0 to the length & crc32 fields.

https://github.com/tikv/tikv/blob/79a0cfbca2c9015e9fe804505ce91d6c70cf04c4/components/engine_rocks/src/import.rs#L34-L35

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1351,6 +1351,11 @@ where
         let exec_res = if !ranges.is_empty() {
             ApplyResult::Res(ExecResult::DeleteRange { ranges })
         } else if !ssts.is_empty() {
+            (|| {
+                fail_point!("dont_delete_ingested_sst", |_| {
+                    ssts.drain(..);
```
kennytm, line 1356:
```suggestion
                    ssts.clear();
```

kennytm (commented):
Rest LGTM

components/sst_importer/src/sst_importer.rs
3pointer: Done, PTAL

kennytm:
(rustfmt failed)

components/sst_importer/src/sst_importer.rs
```diff
@@ -664,21 +664,13 @@ impl ImportDir {
...
-        // currently we can not handle sst file size after rewrite,
-        // we need re-compute length & crc32 and fill back to sstMeta.
-        if length != 0 {
-            if crc32 != 0 {
-                // we only validate if the length and CRC32 are explicitly provided.
-                engine.validate_sst_for_ingestion(cf, &path.clone, length, crc32)?;
-            }
-            IMPORTER_INGEST_BYTES.observe(length as _)
-        }
+        engine.reset_global_seq(cf, &path.clone, 0, 0)?;
```
BusyJay, line 667: If it's not going to check the size and checksum, why keep them as parameters? Raftstore uses additional steps to check the size and checksum.

3pointer: I see, I'll fix it.

kennytm:
```
[2021-02-04T08:49:17.127Z] error: unused variable: `size`
[2021-02-04T08:49:17.127Z]    --> components/sst_importer/src/util.rs:161:13
[2021-02-04T08:49:17.127Z]     |
[2021-02-04T08:49:17.127Z] 161 |         let size = fs::metadata(&sst_path).unwrap().len();
[2021-02-04T08:49:17.127Z]     |             ^^^^ help: if this is intentional, prefix it with an underscore: `_size`
[2021-02-04T08:49:17.127Z]     |
[2021-02-04T08:49:17.127Z]     = note: `-D unused-variables` implied by `-D warnings`
[2021-02-04T08:49:17.128Z] 
[2021-02-04T08:49:17.128Z] error: unused variable: `checksum`
[2021-02-04T08:49:17.128Z]    --> components/sst_importer/src/util.rs:162:13
[2021-02-04T08:49:17.128Z]     |
[2021-02-04T08:49:17.128Z] 162 |         let checksum = calc_crc32(&sst_path).unwrap();
[2021-02-04T08:49:17.128Z]     |             ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_checksum`
```

BusyJay:
> currently we can not handle sst file size after rewrite

Can you explain it in more details? As raftstore snapshot can re-ingest and verify the snapshot multiple times.

3pointer:
> > currently we can not handle sst file size after rewrite
> 
> Can you explain it in more details? As raftstore snapshot can re-ingest and verify the snapshot multiple times.

Currently, TiKV performs key perfix [rewrite](https://github.com/pingcap/br/blob/master/docs/cn/2019-09-09-BR-key-rewrite-disscussion.md) in sst_importer#download. this is because `table id` in restore cluster is usually different with the backup cluster. so the sst file length may change after key rewrite. we can not compare length of ingested file with download file for now.

components/sst_importer/src/sst_importer.rs
```diff
@@ -668,21 +668,14 @@ impl ImportDir {
...
-        // FIXME perform validate_sst_for_ingestion after we can handle sst file size correctly.
-        // currently we can not handle sst file size after rewrite,
-        // we need re-compute length & crc32 and fill back to sstMeta.
-        if length != 0 {
-            if crc32 != 0 {
-                // we only validate if the length and CRC32 are explicitly provided.
-                engine.validate_sst_for_ingestion(cf, &path.clone, length, crc32)?;
-            }
-            IMPORTER_INGEST_BYTES.observe(length as _)
-        }
```
overvenus, line 681: Lightning import may need to validate crc32 and size.

3pointer: but `validate_sst_for_ingestion` is no longer need crc32 and length any more.

overvenus:
LGTM

PR #9629, opened 2021-02-04
external_storage: move util to tikv_util/stream

gregwebs:

In principle this util file shouldn't belong exclusively to external_storage and instead should be shareable among other crates.

In practice I needed to make this change for my cloud library refactoring [PR that is WIP](https://github.com/tikv/tikv/pull/9576): cloud code depends on this and external_storage will depend on cloud code. That PR is getting very large and I am trying to break out some separate changes from it.

We already have a future.rs file and perhaps the two of these should be re-organized appropriately. However, I am already working on massive refactoring so would like to leave that for someone else to do.

/cc @kennytm @yiwu-arbug @BusyJay 

move external_storage/util to tikv_util/stream

### Related changes

- #9576 still WIP will rebase to use this commit.

### Check List 

Tests

- No code changes, so I am relying on the compiler and test suite

Side effects

None

### Release note

None


components/tikv_util/src/stream.rs
```diff
@@ -123,3 +124,18 @@ where
 
     result
 }
+
+impl<E> RetryError for RusotoError<E> {
```
yiwu-arbug, line 128: I don't know if this is feasible but can we move this into the `cloud` crate? Basically I want to hide `rusoto` inside the `cloud` crate and other crates in tikv only depend on it independently via the `cloud` crate.

yiwu-arbug (commented):
Looking good overall.

components/tikv_util/src/stream.rs
gregwebs: rusoto is hidden in the aws crate in my other WIP PR. But rusoto_core may be imported anywhere where a trait is defined due to Rust's restrictions on where traits can be implemented. This is just what we have to do due to the way Rust works.
The only other option is to implement the trait for a new wrapper trait of the rusoto_core type (RusotoError) in this case. I refer to this wrapping as the newtype pattern. I think we could use this pattern in this case, but in general it is not always possible to use it for this purpose. I did already use this pattern once in my WIP to define Debug for a rusoto type when improving encryption.

yiwu-arbug: Can we move this `RetryError` impl into cloud crate in your other PR? Let's not worry about other use case of `rusoto` in the rest of non-cloud tikv code.

yiwu-arbug (approved):

PR #9630, opened 2021-02-04
Add engine_traits tests and docs

brson:

Those adding new storage engines to tikv need to know how they should behave.
This is the start of a test suite just for storage engines.

It tests the engine_traits interface, constructing engines via the engine_test crate,
and thus can be used to test any engine that implements engine_traits, and has a constructor in engine_test.

As of now that is just engine_rocks and engine_panic, but I also have an out of tree engine that passes these tests,
as well as most of the raftstore tests. I do not intend to submit that engine for inclusion in tikv any time soon -
I am just using it as a guide to implement this test suite.

Along the way I have documented much of engine_traits.

As of now this covers most of the essential engine features:

- point gets/writes
- range deletes
- iterators
- snapshots
- write batches
- column families

It does not yet cover the more challenging and esoteric features needed by tikv, but I do intend it to have full coverage eventually.

This is almost entirely additive and makes essentially no changes to tikv, with one exception:
I discovered a problem in the rocksdb iterator API, where if it is called incorrectly the underlying C++ code will abort the program. As that makes testing challenging, I added runtime checks on the Rust side to prevent crashing. Since this is conceivably a performance problem, I also added a "nortcheck" feature to engine_rocks to turn these checks off. This is probably overkill though, and this feature should not be turned off.

### Release note 

- No release note


kennytm (commented):
/lgtm

components/engine_traits_tests/src/lib.rs
```diff
@@ -0,0 +1,100 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: Should be 2021

components/engine_traits_tests/src/misc.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: ditto

components/engine_traits_tests/src/read_consistency.rs
```diff
@@ -0,0 +1,81 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: ditto

components/engine_traits_tests/src/scenario_writes.rs
```diff
@@ -0,0 +1,288 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: ditto

components/engine_traits_tests/src/snapshot_basic.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: ditto

components/engine_traits_tests/src/write_batch.rs
```diff
@@ -0,0 +1,726 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: ditto

5kbpers (approved):
rest LGTM :shipit:

brson:
@5kbpers I've updated the copyright year.

brson:
Updated.

brson:
Updated to master.

components/engine_traits/src/write_batch.rs
```diff
@@ -45,9 +96,36 @@ pub trait Mutable: Send {
...
+/// exactly the same data as previously, Replacing any keys that may have
+/// changed in between the two batch writes.
+///
+/// Commands issued to write batches can be rolled back prior to being committed
+/// by use of _save points_. At any point in the life of a write batch a save
+/// point can be recorded. Any number of save points can be recorded to a stack.
+/// Calling `rollback_to_save_point` reverts all commands issued since the last
+/// save point, and pops the save point from the stack.
 pub trait WriteBatch<E: WriteBatchExt + Sized>: Mutable {
+    /// Create a write batch with a given command capacity
```
tabokie, line 122:
```suggestion
    /// Create a WriteBatch with a given command capacity
```

tabokie (approved):
LGTM

brson:
Thanks for the reviews @tabokie !

PR #9635, opened 2021-02-06
copr: add vectorized rpad_utf8 for strings

Jibbow:

Implements `rpad_utf8()` for coprocessor.

Also moved the test case `test_lpad_utf8()` to the other `lpad` tests so they are more closely together in the file.

### Check List  #5751, #9016

Tests 

- Unit test

### Release note 

- coprocessor supports `rpad_utf8()`


components/tidb_query_expr/src/impl_string.rs
```diff
@@ -304,6 +304,37 @@ pub fn rpad(arg: BytesRef, len: &Int, pad: BytesRef, writer: BytesWriter) -> Res
...
 
+#[rpn_fn(writer)]
+#[inline]
+pub fn rpad_utf8(
+    arg: BytesRef,
+    len: &Int,
+    pad: BytesRef,
+    writer: BytesWriter,
+) -> Result<BytesGuard> {
+    let input = match str::from_utf8(&*arg) {
```
Xuanwo, line 315:
```suggestion
    let input = str::from_utf8(&*arg)?;
```

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -304,6 +304,37 @@ pub fn rpad(arg: BytesRef, len: &Int, pad: BytesRef, writer: BytesWriter) -> Res
...
+        Ok(arg) => arg,
+        Err(err) => return Err(box_err!("invalid input value: {:?}", err)),
+    };
+    let pad = match str::from_utf8(&*pad) {
```
Xuanwo, line 319: ditto

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -304,6 +304,37 @@ pub fn rpad(arg: BytesRef, len: &Int, pad: BytesRef, writer: BytesWriter) -> Res
...
+    match validate_target_len_for_pad(*len < 0, *len, input_len, 4, pad.is_empty()) {
+        None => Ok(writer.write(None)),
+        Some(0) => Ok(writer.write_ref(Some(b""))),
+        Some(target_len) => {
+            let r = input
+                .chars()
+                .chain(pad.chars().cycle())
+                .take(target_len)
+                .collect::<String>();
+            Ok(writer.write(Some(r.into_bytes())))
```
Xuanwo, line 333: We can use `write_ref` instead.

Jibbow: I tried using `write_ref` but it didn't work because we create a new (owned) string with `chain` and `take` and `collect`. Not sure if I missed something though.

Xuanwo: Can we reduce the allocation here? For exmaple, we can calculate the length first and call `write_partial` to write them without extra allocation.

Jibbow: Ah okay got it :)
Should I change the implementation of `rpad()` to use `partial_write()` as well?

skyzh: Yes, I suggest doing so.

Jibbow: Hi, I removed the allocations for both `rpad` and `lpad` functions with the latest commit. The resulting code is not as nice as it was with iterators. Any ideas on how to improve this or should we leave it as it is?

skyzh (approved):
LGTM, thanks!

components/tidb_query_expr/src/impl_string.rs
skyzh: I think we could leave it as-is.

Xuanwo (approved):

PR #9637, opened 2021-02-06
copr: add lower() for strings

Jibbow:

Implements `lower()` for strings in coprocessor

But I'm not sure if I did handling of binary cases correctly.

### Check List   #5751, #9016

Tests 

- Unit test

### Release note 

- coprocessor supports `lower()` for strings


components/tidb_query_expr/src/impl_string.rs
```diff
@@ -499,6 +499,22 @@ pub fn upper(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
     Ok(writer.write_ref(Some(arg)))
 }
 
+#[rpn_fn(writer)]
+#[inline]
+pub fn lower(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
```
Xuanwo, line 504: ping @andylokandy , Do we need to care about utf-8 collection here?

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -499,6 +499,22 @@ pub fn upper(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
...
+    match str::from_utf8(arg) {
+        Ok(s) => Ok(writer.write_ref(Some(s.to_lowercase().as_bytes()))),
+        Err(err) => Err(box_err!("invalid input value: {:?}", err)),
+    }
+}
+
+#[rpn_fn(writer)]
+#[inline]
+pub fn lower_binary(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
```
Xuanwo, line 513: `lower` and `lower_utf8` are different funtions, take `upper` and `upper_utf-8` a look maybe?

components/tidb_query_expr/src/lib.rs
```diff
@@ -603,6 +618,7 @@ fn map_expr_node_to_rpn_func(expr: &Expr) -> Result<RpnFnMeta> {
         ScalarFuncSig::RightUtf8 => right_utf8_fn_meta(),
         ScalarFuncSig::UpperUtf8 => upper_utf8_fn_meta(),
         ScalarFuncSig::Upper => upper_fn_meta(),
+        ScalarFuncSig::Lower => map_lower_sig(value, children)?,
```
Xuanwo, line 621: ditto

components/tidb_query_expr/src/impl_string.rs
Jibbow: Ah sure yes! I wasn't aware of that because in the old coprocessor and in TiDB there was only `lower`, but not `lower_utf8`. Is `lower()` also just a noop?

andylokandy, line 513:
```suggestion
pub fn lower(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
```

andylokandy: > Is lower() also just a noop?

Yes!

components/tidb_query_expr/src/impl_string.rs
andylokandy, line 504:
```suggestion
pub fn lower_utf8(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
```

andylokandy: It should be utf8 encoding since we've checked it not `is_binary_like`

andylokandy (commented):
rest LGTM

Jibbow:
Thanks for the review! I renamed the functions as suggested. 

But I'm still a bit confused: Is it okay now or should there also be a dedicated `lower()` function (without UTF8), because there is a `Upper` and `UpperUTF8`, but only a `Lower` but no `LowerUTF8` in tipb (https://github.com/pingcap/tipb/blob/master/proto/expression.proto#L623)

components/tidb_query_expr/src/lib.rs
```diff
@@ -274,9 +274,9 @@ fn map_lower_sig(value: ScalarFuncSig, children: &[Expr]) -> Result<RpnFnMeta> {
         ));
     }
     if children[0].get_field_type().is_binary_string_like() {
-        Ok(lower_binary_fn_meta())
-    } else {
         Ok(lower_fn_meta())
+    } else {
+        Ok(lower_utf8_fn_meta())
```
Xuanwo, line 279: We don't need this check.

Xuanwo:
Maybe we need to add `LowerUTF8` in tipb.

skyzh:
> Maybe we need to add `LowerUTF8` in tipb.

Yes. I think we may merge this PR after `LowerUTF8` is available in tipb.

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -499,6 +499,22 @@ pub fn upper(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
     Ok(writer.write_ref(Some(arg)))
 }
 
+#[rpn_fn(writer)]
+#[inline]
+pub fn lower_utf8(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
+    match str::from_utf8(arg) {
+        Ok(s) => Ok(writer.write_ref(Some(s.to_lowercase().as_bytes()))),
+        Err(err) => Err(box_err!("invalid input value: {:?}", err)),
```
Xuanwo, line 507: As https://github.com/tikv/tikv/pull/9640 has been merged, please use `str::from_utf8(arg)?` to return error.

Jibbow:
Alright, thanks for the review! How should we go on from here? What needs to be done to add `LowerUTF8` to tipb?

skyzh:
> Alright, thanks for the review! How should we go on from here? What needs to be done to add `LowerUTF8` to tipb?

<del>If you want, you may send a PR to https://github.com/pingcap/tipb/blob/master/proto/expression.proto , update tipb dependency in TiKV, and `LowerUTF8` could be used in TiKV.</del>

I have went through code in TiDB. It seems that TiDB itself doesn't have `LowerUTF8` function https://github.com/pingcap/tidb/blob/master/expression/builtin_string.go . If TiDB pushes down `ScalarFuncSig::Lower`, it is meant to be a `UTF8` lower function. As seen in https://github.com/pingcap/tidb/blob/master/expression/builtin_string.go#L721 , binary won't be pushed down.

Sorry for my mistake before. So for this PR, you don't have to implement `lower` (binary) or check if field type is binary. Just implement `lower` function as UTF8 lower is enough.

components/tidb_query_expr/src/lib.rs
skyzh: We don't need check for `is_binary_string_like`. If `lower` is pushed to TiKV, it is for UTF-8 ones.

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -499,6 +499,22 @@ pub fn upper(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
...
+    }
+}
+
+#[rpn_fn(writer)]
+#[inline]
+pub fn lower(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
+    // Noop for binary strings
+    Ok(writer.write_ref(Some(arg)))
```
skyzh, line 515: `lower` should behave as current `lower_utf8`. You may substitute body of this function for those in `lower_utf8` and remove `lower_utf8`. Please also add a comment to describe why we implement `lower` different from `upper` and `upper_utf8`.

skyzh:
After some investigation, I found that:

* When implementing `upper[_utf8]`, TiDB will use different signature depending on whether data is binary or not https://github.com/pingcap/tidb/blob/master/expression/builtin_string.go#L868
* When implementing `lower`, TiDB uses the same signature. https://github.com/pingcap/tidb/blob/master/expression/builtin_string.go#L697

Therefore, implementation in this PR is correct. We don't need `LowerUTF8` signature in tipb. Sorry for my mistake before.

components/tidb_query_expr/src/lib.rs
skyzh: This check is needed. See my comment below.

skyzh (approved):
LGTM

components/tidb_query_expr/src/lib.rs
skyzh: This check is needed. See my comment below.

Jibbow:
Cool, thanks for looking into this! :blush:

components/tidb_query_expr/src/impl_string.rs
Jibbow: Done :+1:

andylokandy (approved):
/lgtm

PR #8729, opened 2020-09-24
Implements Trim2Args (Vectorized)

hbina:
Part of the functions migration process from TiDB.
Related issue #5751.

### What problem does this PR solve?

Issue Number: #5751  

Problem Summary:

### What is changed and how it works?

What's Changed:

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note

PR #9638, opened 2021-02-06
copr: add vectorized Trim2Args

pvid:
Hi! I am dipping my toes slowly into contributing to TiKV by this small PR

Issue Number: part of https://github.com/tikv/tikv/issues/9016

Problem Summary: implement vectorized `Trim2Args`

I have just reused the implementation for `Trim3Args` to reduce chances of subtle differences in their behaviour in the future. My question is - do you want to go forward with a specialized implementation for `Trim1Arg`? All three could share the same implementation, but a specialized implementation  for `Trim1Arg` that does not bother with conversions to strings also makes sense.

### Check List 

Tests 

- Unit test: migrated from old non-vec implementation

### Release note 
- No release note

### DIFF
--- a/components/tidb_query_expr/src/impl_string.rs
+++ b/components/tidb_query_expr/src/impl_string.rs
@@ -744,6 +744,13 @@ pub fn trim_1_arg(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
 
 #[rpn_fn(writer)]
 #[inline]
+pub fn trim_2_args(arg: BytesRef, pat: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
+    let trimmed = trim(arg, pat, TrimDirection::Both);
+    Ok(writer.write_ref(Some(trimmed)))
+}
+
+#[rpn_fn(writer)]
+#[inline]
 pub fn trim_3_args(
     arg: BytesRef,
     pat: BytesRef,
@@ -752,9 +759,8 @@ pub fn trim_3_args(
 ) -> Result<BytesGuard> {
     match TrimDirection::from_i64(*direction) {
         Some(d) => {
-            let arg = String::from_utf8_lossy(arg);
-            let pat = String::from_utf8_lossy(pat);
-            Ok(writer.write(Some(trim(&arg, &pat, d))))
+            let trimmed = trim(arg, pat, d);
+            Ok(writer.write_ref(Some(trimmed)))
         }
         _ => Err(box_err!("invalid direction value: {}", direction)),
     }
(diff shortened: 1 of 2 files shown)


skyzh (approved):
LGTM. Thanks for your contribution! You can make `trim_1_arg` share the same implementation if you want in another PR.

Xuanwo (approved):

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -742,6 +742,14 @@ pub fn trim_1_arg(arg: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
     Ok(writer.write_ref(Some(result)))
 }
 
+#[rpn_fn(writer)]
+#[inline]
+pub fn trim_2_args(arg: BytesRef, pat: BytesRef, writer: BytesWriter) -> Result<BytesGuard> {
+    let arg = String::from_utf8_lossy(arg);
```
andylokandy, line 748: I think `Trim3Args` is buggy because of the usage of `from_utf8_lossy`. I think we should use [`bstr`](https://docs.rs/bstr/0.2.14/bstr/trait.ByteSlice.html#method.trim_start ).

pvid: There is a test case for invalid utf8 for `Trim1Arg`. I'll port it to `Trim2Args` and `Trim2Args` tests. I've looked into TiDB tests, but they are quite terse and don't deal with non-utf8 characters. However, Go strings do not have to be valid utf-8, so `Trim3Args` is indeed buggy. I'll make the changes and ask for another review

pvid: unfortunately, `bstr` does not support trimming by patterns longer than one characted. I ended up implementing it for byte slices and expanded the test suite

skyzh (commented):
Rest LGTM

Xuanwo:
After this PR merged, we need to close https://github.com/tikv/tikv/pull/8729 for duplicated.

components/tidb_query_expr/src/impl_string.rs
```diff
@@ -752,9 +759,8 @@ pub fn trim_3_args(
 ) -> Result<BytesGuard> {
     match TrimDirection::from_i64(*direction) {
         Some(d) => {
-            let arg = String::from_utf8_lossy(arg);
-            let pat = String::from_utf8_lossy(pat);
-            Ok(writer.write(Some(trim(&arg, &pat, d))))
+            let trimmed = trim(arg, pat, d);
+            Ok(writer.write(Some(trimmed.to_vec())))
```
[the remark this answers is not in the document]
pvid: changed it. Thanks for the catch!

pvid:
> /retest

I am not able to do this 😅  because the PR is not trusted. I believe the test failure is spurious because it comes from `test raftstore::test_merge` that this PR should not influence

Xuanwo (approved):

skyzh (approved):

Issue #4822, opened 2019-06-03
coprocessor/expression: supported aggregate functions


lonng:
## Feature Request

This issue is used to track aggregate functions implementation in the coprocessor.

### Batch

- [ ] AVG
- [ ] BIT_AND
- [ ] BIT_OR
- [ ] BIT_XOR
- [ ] COUNT
- [ ] COUNT(DISTINCT)
- [ ] GROUP_CONCAT
- [ ] JSON_ARRAYAGG
- [ ] JSON_OBJECTAGG
- [ ] MAX
- [ ] MIN
- [ ] STD
- [ ] STDDEV
- [ ] STDDEV_POP
- [ ] STDDEV_SAMP
- [ ] SUM
- [ ] VAR_POP
- [ ] VAR_SAMP
- [ ] VARIANCE

### Non-Batch

- [ ] AVG
- [ ] BIT_AND
- [ ] BIT_OR
- [ ] BIT_XOR
- [ ] COUNT
- [ ] COUNT(DISTINCT)
- [ ] GROUP_CONCAT
- [ ] JSON_ARRAYAGG
- [ ] JSON_OBJECTAGG
- [ ] MAX
- [ ] MIN
- [ ] STD
- [ ] STDDEV
- [ ] STDDEV_POP
- [ ] STDDEV_SAMP
- [ ] SUM
- [ ] VAR_POP
- [ ] VAR_SAMP
- [ ] VARIANCE

pvid:
Hi! I'd like to take a stab at STD/STDDEV/STDDEV_POP/STDDEV_SAMP/VAR_POP/VAR_SAMP/VARIANCE

Based on their implementation in `tidb`, the partial results returned from coprocessors should be the same for all of them.

BTW, is there an accompanying issue in the `tidb` repository? From what I managed to read, there is some more work to do on that side. I have also encountered [copr-test](https://github.com/tikv/copr-tes) and I am not sure how that enters the picture. 

Could you please provide some guidance?

PR #9642, opened 2021-02-08
copr: variance agg func (population and sample)

Jibbow:

Implement variance aggregate function in coprocessor with sample and population variance.

Problem Summary:

Implemented VARIANCE for common types including `Set` and `Enum` similar to e.g. `AVG`.
Similar to `MIN_MAX`, you can specify via a generic parameter if you want the population variance or sample variance. Feedback on that? Is this a good approach?

Also I wasn't quite sure about how to design proper trait bounds for `Real` and `Decimal`. Instead, I quickly extended the `Summable` trait with some more functions, but it's not very nice. Help here is appreciated :hugs: 
One problem I ran into was that I can't just require `T: Add + Sub + Mul + Div` because `Decimal` implements these only for `&Decimal` and `Real` implements them only for `Real` (reference vs no reference). Not sure how to go on from there.

Also fixed a few typos I came across.

### Check List  #4822 

Tests 

- Unit test

### Release note 

- coprocessor supports `VARIANCE()`, `VAR_POP()`, `VAR_SAMP()`


pvid:
too bad that you did not leave a comment under the issue - I have [started to research the issue a few days ago](https://github.com/tikv/tikv/issues/4822#issuecomment-774513129). You can adapt the implementation to support all 6 of the variance and standard deviation functions, because they use use the same same partial results and differ only in the [AppendFinalResult2Chunk](https://github.com/pingcap/tidb/blob/eb81a67b33e763b8ba94f8176b77ce70d931f7a5/executor/aggfuncs/func_stddevsamp.go#L27) implementation which is executed only on the tidb side. 

Except for parsing, the implementation can be oblivious to which of these has actually been pushed down. You could even cut the `VarianceType` trait and computation of the final result

Jibbow:
Ah yes sorry, I also started a few days ago but forgot to mention it. I was too hooked by the implementation so I didn't really pay attention to the issue anymore :D
Anyways, good point with STDDEV!! Didn't think of that! But before going deeper into that I would kindly refer to Slack because @skyzh and I had a very [interesting discussion there](https://tikv-wg.slack.com/archives/CPCS8ADL4/p1612800303121200). 

Adding a new aggregation function might be a bit more work than expected and changes might need to be made to the TiDB side as well. If the maintainers decide to go on with adding a new aggregate function, we could work together on it if you want! :blush:

pvid:
> Ah yes sorry, I also started a few days ago but forgot to mention it. I was too hooked by the implementation so I didn't really pay attention to the issue anymore :D
> Anyways, good point with STDDEV!! Didn't think of that! But before going deeper into that I would kindly refer to Slack because @skyzh and I had a very [interesting discussion there](https://tikv-wg.slack.com/archives/CPCS8ADL4/p1612800303121200).
> 
> Adding a new aggregation function might be a bit more work than expected and changes might need to be made to the TiDB side as well. If the maintainers decide to go on with adding a new aggregate function, we could work together on it if you want! 😊

Sure! I completely missed that channel, thanks!

andylokandy:
Seems that the lint is broken

andylokandy (commented):
Most LGTM

andylokandy:
Do you think it is possible to use [`num-trait`](https://docs.rs/num-traits/0.2.14/num_traits/index.html) and deprecate `Summable`?

skyzh:
> num-trait

We need to implement related traits for `Decimal` if we want to use `num-trait`. I think current design is okay.

Xuanwo:
Please let `cargo clippy` happy:

```
[2021-02-08T14:59:15.151Z] error: using `clone` on type `tidb_query_datatype::codec::data_type::Decimal` which implements the `Copy` trait

[2021-02-08T14:59:15.151Z]    --> components/tidb_query_aggr/src/impl_variance.rs:329:33

[2021-02-08T14:59:15.151Z]     |

[2021-02-08T14:59:15.151Z] 329 |             target[1].push(Some(self.sum.clone()));

[2021-02-08T14:59:15.151Z]     |                                 ^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `self.sum`

[2021-02-08T14:59:15.151Z]     |

[2021-02-08T14:59:15.151Z]     = note: `-D clippy::clone-on-copy` implied by `-D warnings`

[2021-02-08T14:59:15.151Z]     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy

```

andylokandy:
> We need to implement related traits for Decimal. 

We can implement what-we-need.

Jibbow:
Re: How to deal with `Summable` or `num-traits`

The main issue here is that `Decimal` implements all `std::ops::*` for `&'a Decimal` (reference) and `Real` implements them for `Real` (no reference). Both is fine, but I wasn't able to figure out a way to define a trait bound that says "impl `Add`, ... either for references or for owned values; both is fine". So I went back to extending the `Summable` trait.
To conclude: I think we don't necessarily need `num-traits`; normal `std::ops::*` should be fine - but please correct me if I'm wrong.

From the discussion here + some own observations, possible solutions might look like:
 * implement `std::ops::*` for `Decimal` (no reference) alongside(?) `&'a Decimal`. Then we can also use `Summable` as a marker trait if we want, but we could also fall back to requiring a trait bound on `std::ops::Add` directly.
 * implement a new trait alongside `Summable` that also allows for `Sub`, `Mul`, `Div` and leave `Summable` is it was before.

skyzh (approved):
Already looks good to me. By the way, I suggest using `macro_rules` to avoid writing nearly identical code for primitive number types and Enum/Set.

andylokandy (approved):

PR #9634, opened 2021-02-05
copr: add sub_datetime_and_string

Jibbow:
Hi there :wave: 
This is my first PR to the TiKV in an attempt to apply for the LFX Mentorship Program.
I am happy about all kind of feedback. Thank you! :blush: 

### What problem does this PR solve?
Implements `sub_datetime_and_string()` for coprocessor.

### Check List #5751, #9016 

Tests 

- Unit test

### Release note 

- coprocessor supports `sub_datetime_and_string()`

PR #9643, opened 2021-02-08
copr implement vectorized AddDateAndString and SubDateAndString

pvid:
🚨 needs https://github.com/tikv/tikv/pull/9634 to be merged first 🚨 
some parts of the PR depend on it and are commented out for the time being.

Issue Number: part of #9016

Problem Summary:

This PR implements vectorized `AddDateAndString` and `SubDateAndString` functions.

Since `DATE` and `DATETIME` types have the same representation ([`Time`](https://github.com/tikv/tikv/blob/acc1aff38602920a2d475954729bb9eebcbcc1ee/components/tidb_query_datatype/src/codec/mysql/time/mod.rs#L162)), the implementation just delegates to implementation of `AddDatetimeAndString` and `SubDatetimeAndString`.  However, I added separate tests so that the test suite does not depend on an implementation detail.

I have also refactored the `(add|sub)_(date|datetime)_and_string` functions so that they just parse the string to duration and delegate to  `(add|sub)_(date|datetime)_and_duration`. It reduces the number of times the same logic is reimplemented.

Another "cosmetic" change is the removal of  `tidb_query_datatype::codec::mysql::Time` import from the test suite. The implementation and some tests used the `DateTime` alias from [here](https://github.com/tikv/tikv/blob/master/components/tidb_query_datatype/src/codec/data_type/mod.rs#L23). It was quite difficult to read.

### Check List

Tests 

- Unit test

### Release note 

- coprocessors support AddDateAndString and SubDateAndString functions


andylokandy:
`builtinSubDatetimeAndStringSig` and `builtinSubDateAndStringSig` seems to be different in TiDB. `builtinSubDatetimeAndStringSig` casts the value into `Time` and then substrate, while `builtinSubDateAndStringSig` directly subtract on `Duration`. I don't know whether there is any differences between their output. Maybe we should enable the pushdown in tidb and the whitelist in `copr-test` to ensure that.

pvid:
> `builtinSubDatetimeAndStringSig` and `builtinSubDateAndStringSig` seems to be different in TiDB. `builtinSubDatetimeAndStringSig` casts the value into `Time` and then substrate, while `builtinSubDateAndStringSig` directly subtract on `Duration`. I don't know whether there is any differences between their output. Maybe we should enable the pushdown in tidb and the whitelist in `copr-test` to ensure that.

I've read throught the TiDB implemetation and you're right! I thought that `AddDateAndString` and `SubDateAndString` implement the `ADDDATE`/`SUBDATE` MySQL functions, but they implement `ADDTIME`/`SUBTIME` MySQL functions 🙈 . The name tripped me up. I'll modify the PR to return corrent types.

I think the integration tests in `copr-test` will work reliably only after [all of the 12 branches for various arg types](https://github.com/pingcap/tidb/blob/6e7613fec3b4531f9149c3fde0e332f637e162af/expression/builtin_time.go#L5218) are implemented. It might be useful to document the correspondence between [tipb expressions](https://github.com/pingcap/tipb/blob/2870a085203756a63d19d3913672e44c25163cb7/proto/expression.proto#L460) and `ADDDATE`/`SUBDATE` and `ADDTIME`/`SUBTIME` MySQL functions. 

BTW, any ideas why `ADDTIME(<DATE>, <STRING>)` results in a `STRING`? I'd expect a `DATETIME`

pvid:
> > `builtinSubDatetimeAndStringSig` and `builtinSubDateAndStringSig` seems to be different in TiDB. `builtinSubDatetimeAndStringSig` casts the value into `Time` and then substrate, while `builtinSubDateAndStringSig` directly subtract on `Duration`. I don't know whether there is any differences between their output. Maybe we should enable the pushdown in tidb and the whitelist in `copr-test` to ensure that.
> 
> I've read throught the TiDB implemetation and you're right! I thought that `AddDateAndString` and `SubDateAndString` implement the `ADDDATE`/`SUBDATE` MySQL functions, but they implement `ADDTIME`/`SUBTIME` MySQL functions 🙈 . The name tripped me up. I'll modify the PR to return corrent types.
> 
> I think the integration tests in `copr-test` will work reliably only after [all of the 12 branches for various arg types](https://github.com/pingcap/tidb/blob/6e7613fec3b4531f9149c3fde0e332f637e162af/expression/builtin_time.go#L5218) are implemented. It might be useful to document the correspondence between [tipb expressions](https://github.com/pingcap/tipb/blob/2870a085203756a63d19d3913672e44c25163cb7/proto/expression.proto#L460) and `ADDDATE`/`SUBDATE` and `ADDTIME`/`SUBTIME` MySQL functions.
> 
> BTW, any ideas why `ADDTIME(<DATE>, <STRING>)` results in a `STRING`? I'd expect a `DATETIME`

I delved more inte the `tidb` implementation played a bit with `tidb` using `tiup` to validate test cases.

`ADDTIME`/`SUBTIME` does a cast to duration, when the first argument is `DATE`. The desired arg types can be seen in [getBf4TimeAddSub](https://github.com/pingcap/tidb/blob/c9e85ec7011f87a5914fb29f92bbf40f1112a4cb/expression/builtin_time.go#L5113) and a cast is added in [newBaseBuiltinFuncWithTp](https://github.com/pingcap/tidb/blob/7f9c716854e9c2b2cd753a0e4b623e1c040f64ae/expression/builtin.go#L166) which is called by `getBf4TimeAddSub`

@andylokandy, could you please take another look?

andylokandy:
@ichn-hu PTAL

wjhuang2016:
@pvid I think it's a good idea to reuse the code. But I have some concerns:
1. Is there any corner case that needs to handle differently?
2. If there are some changes in TiDB, then it's hard to port these changes to TiKV.

May I propose a solution?
1. Refactoring the code on the TiDB side and see if it can pass the all tests.
2. Refactoring the code on the TiKV side.

skyzh:
Hi @pvid , thanks for your work! Do you have any free time to resolve the conflicts? If not, I would let https://github.com/tikv/tikv/pull/9931 to follow and work on this PR.

PR #9647, opened 2021-02-09
encryption: move AWS to a separate cloud crate

gregwebs:

We are organizing the cloud code in a separate crate. This will facilitate the cloud team being able to own the code and modifying builds to include or exclude cloud providers or in the future dynamically link them.

This is the encryption portion split off from this cloud re-organization PR: https://github.com/tikv/tikv/pull/9576
After this PR is merged, I will create a second PR that organizes the external_storage code.

* AWS code is moved from the encryption crate into the cloud/aws crate
* To further separate out the cloud crate, the encryption crate depends on the `KmsProvider` trait that it exposes. A few types (DataKeyPair, EncryptedKey) are duplicated in the cloud encryption crate to avoid any dependency.
* components/encryption/export is added to avoid cyclical dependencies. Cloud constructors of encryption traits are imported here and then re-exported along with the encryption crate for end users of encryption
* rusoto_util code is moved to aws/util. This is currently duplicated because external_storage still depends on rusoto_util. The next PR will organize the external_storage code and remove the rusoto_util crate.

/cc @yiwu-arbug 

### Related changes

This is the encryption portion split off from this cloud re-organization PR: https://github.com/tikv/tikv/pull/9576
A second PR will be based off of that to organize the external_storage code.

### Check List

Tests

- Unit test: new unit tests covering the functionality
- Integration test: no new integration tests
- Manual test: basic test of running with encryption and rotating keys

Compatibility

- New cloud error codes are used for AWS KMS operations.
- Programs built from the TiKV lib (tikv-server, etc) now import the encryption_export crate instead of encryption directly.

### Release note

Move KMS cloud provider code out of encryption and into separate cloud provider crates.


components/cloud/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+[dependencies]
+error_code = { path = "../error_code", default-features = false }
+failure = "0.1"
+futures-io = "0.3"
+kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
+openssl = "0.10"
+protobuf = "2.8"
+tokio = { version = "0.2.13", features = ["time"] }
+tikv_util = { path = "../tikv_util", default-features = false }
+rusoto_core = "0.45.0"
```
yiwu-arbug, line 25: nit: sort by crate name.

components/cloud/aws/src/kms.rs
```diff
@@ -0,0 +1,279 @@
...
+            plaintext: PlainKey::new(plaintext_key)?,
+        })
+    }
+}
+
+fn classify_generate_data_key_error(err: RusotoError<GenerateDataKeyError>) -> Error {
+    if let RusotoError::Service(e) = &err {
+        match &e {
+            GenerateDataKeyError::InvalidKeyUsage(_) => {
+                Error::KmsError(KmsError::WrongMasterKey(err.into()))
```
yiwu-arbug, line 130: According to rusoto docs, `InvalidKeyUsage` means the `KEY_USAGE` field is incorrect (i.e. not being `ENCRYPT_DECRYPT` in our case), which should not translate to `WrongMasterKey` error.

components/cloud/src/error.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl ErrorTrait for Error {}
+
+#[derive(Debug, Fail)]
+pub enum KmsError {
+    #[fail(display = "Wrong master key {}", _0)]
+    WrongMasterKey(Box<dyn error::Error + Sync + Send>),
+    #[fail(display = "Empty key {}", _0)]
+    EmptyKey(String),
```
yiwu-arbug, line 44: where is this error type being used?

components/encryption/export/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "encryption_export"
```
yiwu-arbug, line 2: why we need this extra export crate? We can have `encryption` crate depend directly on `cloud` crate.

components/cloud/src/kms.rs
```diff
@@ -0,0 +1,155 @@
...
+    type Target = String;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+// EncryptedKey is a newtype used to mark data as an encrypted key
+// It requires the vec to be non-empty
+#[derive(PartialEq, Clone, Debug)]
+pub struct EncryptedKey(Vec<u8>);
```
yiwu-arbug, line 57: Can we remove redundant types like `EncryptedKey` and `DataKeyPair` from encryption crate? They are not being used elsewhere.

components/encryption/export/Cargo.toml
gregwebs: We could. The normal approach is that a trait is defined in a crate and it is implemented for the standard library and some other light-weight crates. Other crates implementing the interface import the interface. Reversing this means that crates that rely on encryption are now being imported by encryption instead. 

It would be odd to see the encryption lib depend on cloud-aws and thus the AWS SDK. The AWS SDK libraries here will soon include S3. We could make a separate AWS cloud crate for encryption and storage, but still we have our encryption library depending on AWS KMS even though it shouldn't know anything about this. Normally dependencies flow in the opposite direction.

It would make sense for cloud-aws to import encryption for KMS encryption. We are choosing to do extra work to de-couple these through a cloud interface. Would we follow this pattern for other crates? We are only allowed to by the compiler if we go to the extra trouble of creating additional traits that allow for de-coupling.

I am open to being convinced that this is a pragmatic choice. However, just because we can now reverse the dependency flow doesn't necessarily mean we should.

components/cloud/src/kms.rs
gregwebs: I am not sure what you mean: they are used in the encryption crate.

components/cloud/src/error.rs
gregwebs: This is used by `empty_key_contents`. I can get rid of that helper function.

components/encryption/export/Cargo.toml
yiwu-arbug: I'm seeing your point that encryption crate need not depend on cloud-aws. But I still want to get rid of encryption-export which is an extra indirection. How about we do this?

* split `cloud` crate into `cloud-trait` and `cloud`. `cloud-trait` contain the traits for KMS and blob storage. `cloud` contain instantiation of cloud providers that you are putting in `xxx-export` crates.
* `encryption` crate depend on `cloud-trait`.
* Move `encryption::KMSConfig` config struct into the `cloud-trait` crate (becomes `cloud-trait::KMSConfig` probably). `encryption::KMSConfig` become an alias of `cloud-trait::KMSConfig`.
* in the server layer where `encryption::DataKeyManager` is created, the KMS provider is constructed and passed into the `DataKeyManager`'s ctor. Something like:

```
// in cmd/src/server.rs
let encryption_config = tikv_config.security_encryption;
let kms_provider = if let MasterKeyConfig::KMS { kms_config } = encryption_config.master_key {
  Some(cloud::kms_provider_from_config(kms_config))
} else {
  None
};
let encryption_key_manager = encryption::DataKeyManager::new(..., kms_provider);
```

What do you think?

yiwu-arbug (commented):
Sorry for delay. Replied some comments.

components/encryption/export/Cargo.toml
gregwebs: This would be fine if we were sure the server will always be the only code using this. But we do already have tikv-ctl using this so it already gets repetitive. It might work okay for encryption which currently has few end users, but this is not a broad pattern that can be followed. external storage already has backup and sst_importer as end users in the code base and could end up with more end users.

I think that the "extra indirection" has its purpose in this general pattern to differentiate between libraries that implement the interface and end users that constructor it and call its methods. It simplifies usage to only have a restricted subset of encryption available for import (as encryption_export).

yiwu-arbug: I don't see encryption crate or the external_storage crate being use in say 10 places in the foreseeable future. The added code to pass the param is almost just a one liner. Change both of these crate to accept KMSProvider and BlobStorageProvider as param also make them more testable - for dependency injection.

components/cloud/src/kms.rs
gregwebs: I think you are pointing to something where the split between the two crates is not quite right. I will make some improvements here and see what can be made of it.

components/encryption/export/Cargo.toml
gregwebs: I changed things so that neither crate depends on each other and instead they are bound together in the export crate via newtype delegation from cloud to encryption. I think this is a nice organization.

components/cloud/src/kms.rs
gregwebs: I was able to stop passing the runtime around and cleanup the interface. I then followed up with the above comment about completely separating the two crates.

components/encryption/export/Cargo.toml
gregwebs: Again I think your suggestions are pragmatic enough for the situation we have right now. However, I am much more interested in establishing patterns that we can use everywhere, are easy to follow, and don't have too much overhead.

components/cloud/aws/src/kms.rs
```diff
@@ -0,0 +1,262 @@
...
+            key_spec: Some(AWS_KMS_DATA_KEY_SPEC.to_owned()),
+            number_of_bytes: None,
+        };
+        self.client
+            .generate_data_key(generate_request)
+            .await
+            .map_err(classify_generate_data_key_error)
+            .and_then(|response| {
+                let ciphertext_key = response.ciphertext_blob.unwrap().as_ref().to_vec();
+                let plaintext_key = response.plaintext.unwrap().as_ref().to_vec();
```
yiwu-arbug, line 115: avoid unwrap and handle the error.

components/encryption/src/master_key/kms.rs
```diff
@@ -68,10 +98,13 @@ impl KmsBackend {
         let mut opt_state = self.state.lock().unwrap();
         if opt_state.is_none() {
             let mut runtime = self.runtime.lock().unwrap();
-            let data_key = self.kms_provider.generate_data_key(&mut runtime)?;
+            let timeout_duration = Duration::from_secs(10);
```
yiwu-arbug, line 95: make it a constant?

components/encryption/export/src/lib.rs
```diff
@@ -38,20 +47,87 @@ pub fn create_backend(config: &MasterKeyConfig) -> Result<Box<dyn Backend>> {
     result
 }
 
+pub fn create_cloud_backend(config: &KmsConfig) -> Result<Box<dyn Backend>> {
+    Ok(match config.vendor.as_str() {
+        #[cfg(feature = "aws")]
+        "aws" | "" => {
```
yiwu-arbug, line 55: make it a constant

yiwu-arbug (commented):
I still don't like having a encryption-export crate and don't like duplicating `DataKeyPair` and `EncryptedKey` types, but the patch looking good to me otherwise.

yiwu-arbug (approved):

components/cloud/aws/src/kms.rs
gregwebs: This code was just moved from encryption/master_key/kms.rs and is unchanged.

components/encryption/export/src/lib.rs
gregwebs: Done

components/encryption/src/master_key/kms.rs
gregwebs: moved it into the struct. It should be configurable in the future.

gregwebs:
> I still don't like having a encryption-export crate and don't like duplicating DataKeyPair and EncryptedKey types, but the patch looking good to me otherwise.

The encryption-export crate is following the same flow that @brson has in mind as well for constructors (he had some different specifics about naming conventions and having lots more small crates). I would like to try to stick with one pattern that works everywhere if possible and I hope we can converge when his work lands.

I am not entirely happy with the duplicated types either. If cloud were the only possible kind of KMS provider I would probably import them from cloud. Having the types defined in encryption better aligns with having non-cloud KMS providers.

components/cloud/aws/src/kms.rs
```diff
@@ -0,0 +1,263 @@
...
+        };
+        f.debug_struct("AwsKms")
+            .field("client", &kms_client)
+            .field("current_key_id", &self.current_key_id)
+            .finish()
+    }
+}
+
+impl AwsKms {
+    fn new_creds_dispatcher<Creds, Dispatcher>(
```
hunterlxt, line 40:
```suggestion
    fn new_with_creds_dispatcher<Creds, Dispatcher>(
```

components/cloud/aws/src/util.rs
```diff
@@ -0,0 +1,113 @@
...
+        Ok(Region::default())
+    }
+}
+
+pub struct CredentialsProvider(AutoRefreshingProvider<DefaultCredentialsProvider>);
+
+impl CredentialsProvider {
+    pub fn new() -> io::Result<CredentialsProvider> {
+        Ok(CredentialsProvider(
+            AutoRefreshingProvider::new(DefaultCredentialsProvider::default()).map_err(|e| {
```
hunterlxt, line 58: why not use `rusoto_core::DefaultCredentialsProvider`

components/cloud/aws/src/kms.rs
yiwu-arbug: and we can change them by the way.

components/cloud/aws/src/util.rs
yiwu-arbug: The credential provider is a combination of `rusoto_core::DefaultCredentialsProvider` and `WebIdentityProvider`. The later handles IAM in k8s.

components/cloud/src/kms.rs
hunterlxt: anything progressed?

components/cloud/aws/src/util.rs
```diff
@@ -0,0 +1,113 @@
...
+            format!("create aws http client error: {}", e),
+        )
+    })
+}
+
+pub fn get_region(region: &str, endpoint: &str) -> io::Result<Region> {
+    if !endpoint.is_empty() {
+        Ok(Region::Custom {
+            name: region.to_owned(),
+            endpoint: endpoint.to_owned(),
```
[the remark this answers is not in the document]
gregwebs: This file was existing code that is unchanged. It is just a copy from rusoto_util. In my next PR for external_storage I will delete the rusoto_util crate. I was asked to send 2 separate PRs.

components/cloud/aws/src/util.rs
gregwebs: see above comment that this is the pre-existing rusoto_util code.

components/cloud/aws/src/kms.rs
gregwebs: I know, I have changed so much code already. I need to stop refactoring this code now and instead commit it.

components/cloud/aws/src/util.rs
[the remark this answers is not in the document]
gregwebs: This was mentioned in the PR description BTW

components/cloud/aws/src/kms.rs
gregwebs: 👍 changed.

components/cloud/src/kms.rs
gregwebs: These small structs are duplicated because neither crate depends on each other as per the other conversation. the encryption_export crate binds the two crates together. Duplication in general is a problem when using non type-safe languages (we are not) or when implementations that should be the same drift apart. However, our entire goal of the cloud library is to provide an interface that lets the cloud library make changes independently of TiKV. So we do not want the implementations to be the same. There is already some difference in EncryptedKey that cannot be resolved by importing it from cloud to encryption. We very much want to allow these definitions to drift apart as needed. The cloud just needs to maintain the interface: `EncryptedKey::new` (takes a vec) and the two DataKeyPair fields support `.to_vec`.

hunterlxt (approved):
/lgtm

PR #9648, opened 2021-02-10
copr: support enum pushdown

dyzsr:

Problem Summary:

- This PR enables enum processing in TiKV coprocessor together with https://github.com/pingcap/tidb/pull/22686

**What's Changed:**

- Encoding: add `EnumEncoder`, `EnumDatumPayloadChunkEncoder` and modify `chunk::Column`:
   - Encode enums from raw data using `Column::append_enum_datum`:
      - `EnumDatumPayloadChunkEncoder::write_enum_to_chunk_by_datum_payload_compact_bytes` for compact bytes.
      - `EnumDatumPayloadChunkEncoder::write_enum_to_chunk_by_datum_payload_uint` for unsigned int.
      - `EnumDatumPayloadChunkEncoder::write_enum_to_chunk_by_datum_payload_var_uint` for variant unsigned int.
   - Encode enums from vector value using `Column::append_enum`:
      - `EnumEncoder::write_enum` for decoded enum data.

- Decoding: add `EnumDecoder`, modify `DatumPayloadDecoder`, and impl `RawDatumDecoder<Enum>` for `'a [u8]`:
   - Decode enums from raw data using `RawDatumDecoder<Enum>::decode`:
      - `DatumPayloadDecoder::read_datum_payload_enum_compact_bytes` calls `EnumDecoder::read_enum_compact_bytes`.
      - `DatumPayloadDecoder::read_datum_payload_enum_uint` calls `EnumDecoder::read_enum_uint`.
      - `DatumPayloadDecoder::read_datum_payload_enum_var_uint` calls `EnumDecoder::read_enum_var_uint`.

- Modify `Enum`, `EnumRef`:
   - Store the name as `Vec<u8>` instead of the reference to element names (in the format of `Arc<BufferVec>`).
      - This is because `FieldType::elems` has the type `[String]`, which cannot be shared with `Arc<BufferVec>`, otherwise we have to copy all the `elems` each time we create an `Enum`.
   - Impl `ToInt`, `ToString`, and `ConvertTo<f64>` traits for casting `Enum` to `Int`, `Bytes`, and `Real`.

- Re-Impl `Evaluable` for `Int` to enable type-conversion from `Enum` to `Int`.
- Re-Impl `EvaluableRef` for `BytesRef` to enable type-conversion from `Enum` to `Bytes`.
- Re-Impl `ChunkedVecEnum` to accommodate `Evaluable for Int` and `EvaluableRef for BytesRef`:
   - `Evaluable::borrow_vector_value` requires conversion from `ChunkedVecEnum` to `ChunkedVecSized<Int>`
   - `EvaluableRef::borrow_vector_value` requires conversion from `ChunkedVecEnum` to `ChunkedVecBytes`.
   - All `ChunkedVecEnum`, `ChunkedVecSized<Int>` and `ChunkedVecBytes` have lifetime `'a`, which means we're not able to create owned values for the destination types, as they are owned by the function `borrow_vector_value` and cannot be returned with a lifetime `'a`.
   - Therefore we have to store `ChunkedVecSized<Int>` and `ChunkedVecBytes` as owned fields inside `ChunkedVecEnum`, and return their references when calling `borrow_vector_value`.
   - This implementation is ugly and not space efficient, but it works and conforms with Rust's lifetime requirements.

- Modify `validate_expr_return_type` to loose the type requirements:
   - Allow using `Enum` as `Int` or as `Bytes`.
- Auto type-conversion during the evaluation of `RpnFnMeta`:
   - `ArgConstructor::eval` will be called during evaluation to compute each argument, and it invokes `borrow_vector_value`, so that we can automatically convert `Enum` to `Int` or to `Bytes`.

**Problems:**

- TiDB doesn't have separate eval types for `Enum/Set`, so functions that returning enums have to either return `Int` or `String`, such as `if` and `case`. This causes incompatible evaluation results between TiKV and MySQL:
   - For example, `select e from t where if(e>1, e, e)`, `select e from t where case e when 1 then e else e end`.
   TiKV uses `Bytes` as the return type of function `if` and `case`, however enums and bytes behave differently in case of converting to MySQL boolean type. This can cause inconsistency when using these functions as selection conditions.

Related issue: https://github.com/pingcap/tidb/issues/23114

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

- Support enum pushdown from TiDB


Cargo.toml
```diff
@@ -185,7 +185,7 @@ tikv_alloc = { path = "components/tikv_alloc" }
 tikv_util = { path = "components/tikv_util", default-features = false }
 collections = { path = "components/collections" }
 time = "0.1"
-tipb = { git = "https://github.com/pingcap/tipb.git", default-features = false }
+tipb = { git = "https://github.com/Xuanwo/tipb.git", branch = "patch-1", default-features = false }
```
gregwebs, line 188: You can add a single patch in the top-level Cargo.toml for this WIP change.

```
[patch.'https://github.com/pingcap/tipb']
tipb = { branch = "patch-1", git = "https://github.com/Xuanwo/tipb.git", default-features = false }
```

dyzsr: Thanks!

skyzh:
```
Store the name as Vec<u8> instead of the reference to element names (in the format of Arc<BufferVec>).
This is because FieldType::elems has the type [String], which cannot be shared with Arc<BufferVec>, otherwise we have to copy all the elems each time we create an Enum.
```

`Vec<u8>` plus offset is nearly identical to `BufferVec`. By using `Arc<BufferVec>`, we only need one copy of same column of elements in memory. I think we could utilize the fields and design provided by `Enum` and `EnumRef`.

components/tidb_query_datatype/src/codec/data_type/mod.rs
```diff
@@ -257,7 +257,39 @@ macro_rules! impl_evaluable_type {
...
+impl Evaluable for Int {
+    const EVAL_TYPE: EvalType = EvalType::Int;
+
+    #[inline]
+    fn borrow_scalar_value(v: &ScalarValue) -> Option<&Self> {
+        match v {
+            ScalarValue::Int(x) => x.as_ref(),
+            ScalarValue::Enum(x) => x
+                .as_ref()
+                .map(|x| unsafe { std::mem::transmute(x.value_ref()) }),
```
skyzh, line 269: Why do we need to use `std::men::transmute` here? This should be able to compile without this transmute.

skyzh (commented):
I'm okay with current design. From my perspective, this patch derives ideas from TiDB, where we represent each enum with an integer and a byte string. This also means that, we gave up on previous approach, where each owned enum element is represented by its index and all of its possible values (stored in `Arc<BufferVec>`).

The upside of current design is that, we nearly have identical implementation to TiDB. But the downside is that, we use more memory (there might be multiple copies of string representation of same-value enum in memory), need more hacks (e.g. implicit conversion from enum to bytes), and have worse locality when processing them in a vectorized fashion.

Anyway, any implementation that makes enum works in TiKV is okay, regardless of its design. We could optimize them later. And currently, this patch looks good to me overall.

skyzh:
By the way, I also suggests splitting this into multiple PRs, which makes it easier to review. It seems that we could implement
* Refactor Enum / ChunkedVecEnum
* Implement decode / encode logic
* Refactor codegen
separately.

skyzh:
Regarding this issue https://github.com/pingcap/tidb/issues/23114 . If we implement Enum / Set as separate eval types with column information embedded (as we have done previously), we could solve this issue on TiKV side (and later refactor on TiDB). I also suspect that some executors like `TopN` will malfunction on enum, if we stick to TiDB's design.

components/tidb_query_datatype/src/codec/data_type/mod.rs
dyzsr: I thought `&u64` cannot be converted into `&i64` in safe mode.

skyzh:
> > ```
> > Store the name as Vec<u8> instead of the reference to element names (in the format of Arc<BufferVec>).
> > This is because FieldType::elems has the type [String], which cannot be shared with Arc<BufferVec>, otherwise we have to copy all the elems each time we create an Enum.
> > ```
> > 
> > 
> > `Vec<u8>` plus offset is nearly identical to `BufferVec`. By using `Arc<BufferVec>`, we only need one copy of same column of elements in memory. I think we could utilize the fields and design provided by `Enum` and `EnumRef`.
> 
> 
> The problem is `Enum::new` is called when encoding or decoding, and the only place to get `elems` is from `field_type`, which is a reference and cannot be used to create `Arc` without copying all the data.

If value selections of an enum are far fewer than rows, personally, I think that this copy may reduce more string allocations in the future.

Anyway, current design seems good to me. We may save this optimization for the future.

components/tidb_query_datatype/src/codec/data_type/mod.rs
```diff
@@ -266,7 +266,7 @@ impl Evaluable for Int {
             ScalarValue::Int(x) => x.as_ref(),
             ScalarValue::Enum(x) => x
                 .as_ref()
-                .map(|x| unsafe { std::mem::transmute(x.value_ref()) }),
+                .map(|x| unsafe { &*(x.value_ref() as *const u64 as *const i64) }),
```
skyzh, line 269: I would prefer `std::mem::transmute<&u64, &i64>(a, b)` here. The reason is that, by specifying concrete type here, we could retain the original lifetime of `&u64`. Neither `transmute` without generic and pointer cast could have the same effect.

By the way, if we always get `Enum` as `i64`, why not using i64 across all enum-related structs?

dyzsr: Thanks, let me try the `transmute` you mentioned.

dyzsr: I think storing `u64` could be more consistent.

components/tidb_query_datatype/src/codec/mysql/enums.rs
```diff
@@ -131,22 +177,114 @@ impl<'a> PartialOrd for EnumRef<'a> {
...
+        String::from_utf8_lossy(self.name).to_string()
+    }
+}
+
+pub trait EnumEncoder: NumberEncoder {
+    #[inline]
+    fn write_enum(&mut self, data: EnumRef) -> Result<()> {
+        self.write_u64_le(data.value as u64)?;
+        self.write_bytes(data.name)?;
+        Ok(())
```
guo-shaoge, line 191: Do we really need to encode `data.name` ? If this member be transmited by network, there will be performance regression.

components/tidb_query_datatype/src/codec/data_type/chunked_vec_enum.rs
```diff
@@ -3,41 +3,50 @@
...
+/// TODO: Change to a compact format when TiDB has its Enum/Set EvalType
 #[derive(Debug, Clone)]
 pub struct ChunkedVecEnum {
-    data: Arc<BufferVec>,
-    bitmap: BitVec,
     // MySQL Enum is 1-based index, value == 0 means this enum is ''
-    value: Vec<usize>,
+    values: ChunkedVecSized<Int>,
+    names: ChunkedVecBytes,
 }
```
guo-shaoge, line 30: This change may cause increase of memory usage. Duplicated **bitmap** in `ChunkVecSized<Int>` and `ChunkedVecByte`.

Also origin implementation use pointer to store name, here we store real enum definition.

Maybe we can find another way to fix problem in `borrow_vector_ref()`

skyzh: > Also origin implementation use pointer to store name, here we store real enum definition.

This has been well-discussed before. Sticking to current design means that we had to store each value in both `Int` and `Bytes` format sequentially, instead of using something like pointer.

> This change may cause increase of memory usage. Duplicated bitmap in `ChunkVecSized<Int>` and `ChunkedVecByte`.

Agree with this. To make things simple, I suggest using a `Vec<Int>` and a `ChunkedVecByte` here.

skyzh: The issue lies in the design. In current PR, enum could be implicitly cast to bytes and int. Therefore, we could not reduce this memory footprint.

components/tidb_query_datatype/src/codec/data_type/vector.rs
```diff
@@ -260,8 +260,22 @@ impl VectorValue {
                 }
                 size
             }
-            // TODO: implement here after we implement enum/set encoding
-            VectorValue::Enum(_) => unimplemented!(),
+            VectorValue::Enum(vec) => {
+                let mut size = logical_rows.len() + 10;
```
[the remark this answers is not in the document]
dyzsr: Sorry I forgot to update this part😂.
I think maybe `logical_rows.len() * 9 + logical_rows.len() * 8 + sum(string_length) + 10` instead? Since we write both int and string to chunks.

components/tidb_query_datatype/src/codec/mysql/enums.rs
dyzsr: There are some problems in adjusting TiDB code for this purpose. We can leave it for further optimizations.

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -67,6 +67,7 @@ fn get_cast_fn_rpn_meta(
                 cast_json_as_any_fn_meta::<Int>()
             }
         }
+        (EvalType::Enum, EvalType::Int) => cast_enum_as_any_fn_meta::<Int>(),
```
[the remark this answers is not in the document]
dyzsr: We still have to implement them for explicitly called cast functions.

components/tidb_query_datatype/src/codec/mysql/enums.rs
```diff
@@ -131,22 +177,114 @@ impl<'a> PartialOrd for EnumRef<'a> {
...
+    ) -> Result<()> {
+        let value = src_payload.read_var_u64()?;
+        let name = Enum::get_value_name(value, field_type.get_elems());
+        self.write_enum_to_chunk(value, name)
+    }
+}
+
+impl<T: BufferWriter> EnumDatumPayloadChunkEncoder for T {}
+
+pub trait EnumDecoder: NumberDecoder {
```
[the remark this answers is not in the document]
dyzsr: Will add them later

skyzh (approved):

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -1352,6 +1345,37 @@ fn cast_json_as_bytes(ctx: &mut EvalContext, val: Option<JsonRef>) -> Result<Opt
     }
 }
 
+#[rpn_fn(nullable, capture = [ctx])]
```
Xuanwo, line 1348: Myabe we can remove `nullable` here? `rpn_fn` should already implement it.

components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -1352,6 +1345,37 @@ fn cast_json_as_bytes(ctx: &mut EvalContext, val: Option<JsonRef>) -> Result<Opt
...
+    match val {
+        None => Ok(None),
+        Some(val) => {
+            let val = To::convert_from(ctx, val.to_owned())?;
+            Ok(Some(val))
+        }
+    }
+}
+
+#[rpn_fn(nullable, capture = [ctx, extra])]
```
Xuanwo, line 1363: ditto.

components/tidb_query_expr/src/impl_cast.rs
dyzsr: Seems it cannot compile if I remove `nullable`.

Xuanwo: OK, let's figure it out in another PR.

components/tidb_query_datatype/src/codec/data_type/mod.rs
dyzsr: This is a recommendation from CI. CI will not be happy if I write `transmute` here😂

components/tidb_query_expr/src/impl_cast.rs
skyzh: I didn't implement non-nullable with context support

components/tidb_query_datatype/src/codec/data_type/mod.rs
Xuanwo: How about copy here instead of using unsafe magic?

skyzh: Copied item only have lifetime throughout this function. We need it to be as long as value itself.

Xuanwo: OK, I got it.

Xuanwo (approved):

skyzh:
Are we ready to merge? I suggest running some integration test against the TiDB PR before merging.

dyzsr:
> Are we ready to merge? I suggest running some integration test against the TiDB PR before merging.

I think we can merge it next week. Let @wshwsh12 take a look first.

wshwsh12:
LGTM

components/tidb_query_datatype/src/codec/data_type/mod.rs
```diff
@@ -257,7 +257,41 @@ macro_rules! impl_evaluable_type {
...
+            _ => unimplemented!(),
+        }
+    }
+
+    #[inline]
+    fn borrow_scalar_value_ref(v: ScalarValueRef) -> Option<&Self> {
+        match v {
+            ScalarValueRef::Int(x) => x,
+            ScalarValueRef::Enum(x) => {
+                x.map(|x| unsafe { &*(x.value_ref() as *const u64 as *const i64) })
```
sticnarf, line 279: This unsafe looks unsound.

Because x is `EnumRef` which implements `Copy`, the `x` inside the closure is a local variable inside of a function. So I think this closure returns a reference to a local variable.

sticnarf: @skyzh

sticnarf: To fix the problem, I think `EnumRef` may be defined like

```rust
pub struct EnumRef<'a> {
    name: &'a [u8],
    value: &'a u64,
}
```

skyzh: I did not even remember EnumRef implements Copy. I'll check soon.

skyzh: Agree. This is unsound. Will be fixed soon.

skyzh: What about changing it to `i64`? cc @dyzsr

dyzsr: Agree

skyzh:
https://github.com/tikv/tikv/pull/10074 a fix is in progress.

PR #9649, opened 2021-02-10
rocksdb: increase default value of max-background-flushes to avoid write stall

yiwu-arbug:

Problem Summary: I experience write stall due to high number of lock CF memtables, when running TPCC prepare on AWS 16c host. Apparently the default max-background-flushes = 2 is not enough when load is high and there are concurrent flushes for default CF and write CF.

What's Changed: Adjusting the formula so that `max-background-flushes` would be 3 and 4 respectively for 8c and 16c. Increases `max-background-jobs` accordingly.

### Related changes

none

### Check List 

Tests 
Manually test on a 40c host and see max-background-flushes = 4. Sample config output: https://gist.githubusercontent.com/yiwu-arbug/a04d47ddfbde7b878258e2cd8b401371/raw/942cce5c412fa19b569a83a20786955f280245ff/gistfile1.txt

### Release note 

* Increase default value of `rocksdb.max-background-flushes` to avoid write stall.


src/config.rs
```diff
@@ -158,40 +158,45 @@ impl TitanCfConfig {
...
     // At the minimum, we should have two background jobs: one for flush and one for compaction.
     // Otherwise, the number of background jobs should not exceed cpu_num - 1.
     // By default, rocksdb assign (max_background_jobs / 4) threads dedicated for flush, and
     // the rest shared by flush and compaction.
-    let max_background_jobs: i32 =
-        cmp::max(2, cmp::min(default_background_jobs, (cpu_num - 1.0) as i32));
-    // Scale flush threads proportionally to background jobs. Also make sure the number of flush
+    let max_background_jobs = cmp::max(
+        2,
+        cmp::min(defaults.max_background_jobs, (cpu_num - 1.0) as u32),
```
gregwebs, line 177: It looks like cpu_num can be < 1. Then we would cast a negative number as unsigned?
Maybe at the top we can have `let cpu_num = min(1, SysQuota...)` or we can push this guarantee down into SysQuota itself?

gregwebs: I would like to see the usage of `trunc()` to explicitly say that we are trying to truncate a number rather than implicitly truncating by casting.

src/config.rs
```diff
@@ -158,40 +158,45 @@ impl TitanCfConfig {
...
 }
 
-fn get_background_job_limit(
-    default_background_jobs: i32,
-    default_background_flushes: i32,
-    default_sub_compactions: u32,
-    default_background_gc: i32,
-) -> (i32, i32, u32, i32) {
+#[derive(Clone, Copy, Debug)]
+struct BackgroundJobLimits {
```
gregwebs, line 162: This is much better, thanks! As a rule of thumb I try to avoid having 2 arguments of the same type given to a function.

src/config.rs
```diff
@@ -158,40 +158,45 @@ impl TitanCfConfig {
...
+    let max_background_jobs = cmp::max(
+        2,
+        cmp::min(defaults.max_background_jobs, (cpu_num - 1.0) as u32),
+    );
+    // Scale flush threads proportionally to cpu cores. Also make sure the number of flush
     // threads doesn't exceed total jobs.
-    let max_background_flushes: i32 = cmp::min(
-        cmp::max(default_background_flushes, max_background_jobs / 4),
+    let max_background_flushes = cmp::min(
+        cmp::min(defaults.max_background_flushes, (cpu_num as u32) / 4),
```
tabokie, line 182: The old constraint shouldn't be dropped, which is reserving at least 2 flush workers even when `cpu_num` is smaller than 8.

src/config.rs
yiwu-arbug: updated

src/config.rs
yiwu-arbug: updated. Adding a min_background_flushes param to enforce that.

src/config.rs
yiwu-arbug: I think arguments are fine. They are named. I'm bother by the unnamed return tuple.

src/config.rs
```diff
@@ -158,40 +158,48 @@ impl TitanCfConfig {
...
+    let max_background_jobs = cmp::max(2, cmp::min(defaults.max_background_jobs, cpu_num - 1));
+    // Scale flush threads proportionally to cpu cores. Also make sure the number of flush
     // threads doesn't exceed total jobs.
-    let max_background_flushes: i32 = cmp::min(
-        cmp::max(default_background_flushes, max_background_jobs / 4),
+    let max_background_flushes = cmp::min(
+        cmp::min(
+            cmp::max(min_background_flushes, (cpu_num as u32) / 4),
+            defaults.max_background_flushes,
+        ),
```
tabokie, line 185: For simplicity, maybe `max_background_jobs / 3` suffice?

tabokie (approved):

src/config.rs
```diff
@@ -158,40 +158,48 @@ impl TitanCfConfig {
...
-        cmp::max(2, cmp::min(default_background_jobs, (cpu_num - 1.0) as i32));
-    // Scale flush threads proportionally to background jobs. Also make sure the number of flush
+    let max_background_jobs = cmp::max(2, cmp::min(defaults.max_background_jobs, cpu_num - 1));
+    // Scale flush threads proportionally to cpu cores. Also make sure the number of flush
     // threads doesn't exceed total jobs.
-    let max_background_flushes: i32 = cmp::min(
-        cmp::max(default_background_flushes, max_background_jobs / 4),
+    let max_background_flushes = cmp::min(
+        cmp::min(
+            cmp::max(min_background_flushes, (cpu_num as u32) / 4),
```
gregwebs, line 183: cpu_num is already a `u32` so this cast shouldn't be needed? It is odd to see division of a truncated number. 3.999 would end up as 0 here instead of 1. Maybe this number should be rounded instead of truncated?

src/config.rs
gregwebs: The return tuple was of variables (named) at both the function definition return and the function call return. The calling bothered me more because there were unnamed numbers in some spots. They both bother me a lot, thanks again for cleaning it up.

src/config.rs
yiwu-arbug: Cleaned it up.

src/config.rs
yiwu-arbug: Simplified.

yiwu-arbug:
Simplified and adjusted the formula. Now it has 3 and 4 flush threads for 8c and 16c respectively.
```
cpu 1, limits BackgroundJobLimits { max_background_jobs: 2, max_background_flushes: 1, max_sub_compactions: 1, max_titan_background_gc: 1 }
cpu 2, limits BackgroundJobLimits { max_background_jobs: 2, max_background_flushes: 1, max_sub_compactions: 1, max_titan_background_gc: 2 }
cpu 3, limits BackgroundJobLimits { max_background_jobs: 3, max_background_flushes: 1, max_sub_compactions: 1, max_titan_background_gc: 3 }
cpu 4, limits BackgroundJobLimits { max_background_jobs: 4, max_background_flushes: 2, max_sub_compactions: 1, max_titan_background_gc: 4 }
cpu 5, limits BackgroundJobLimits { max_background_jobs: 5, max_background_flushes: 2, max_sub_compactions: 2, max_titan_background_gc: 4 }
cpu 6, limits BackgroundJobLimits { max_background_jobs: 6, max_background_flushes: 2, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 7, limits BackgroundJobLimits { max_background_jobs: 7, max_background_flushes: 3, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 8, limits BackgroundJobLimits { max_background_jobs: 8, max_background_flushes: 3, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 9, limits BackgroundJobLimits { max_background_jobs: 9, max_background_flushes: 3, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 10, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 11, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 12, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 13, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 14, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 15, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
cpu 16, limits BackgroundJobLimits { max_background_jobs: 10, max_background_flushes: 4, max_sub_compactions: 3, max_titan_background_gc: 4 }
```

gregwebs (approved):
@yiwu-arbug this needs `make format`

Connor1996 (approved):

Connor1996 (approved):
LGTM

PR #9650, opened 2021-02-11
ci-build: verify that error codes have been generated

gregwebs:
Don't bother trying to track this in the Makefile.
Now this task is only ran by "make all" so it isn't run much.

error_code.toml is not up to date

Fail the ci build if error_code.toml is not generated.

### Check List 

Testing
* ci build should pass
* test that this script failed when there were missing error codes but passes now.

### Release note

None


ci-build/test.sh
```diff
@@ -11,9 +11,14 @@ panic() {
 # Move to project root
 cd "$(dirname "$0")/.."
 
-if [[ -z "$SKIP_FORMAT_CHECK" ]]; then
```
overvenus, line 14: Maybe it should be `SKIP_FORMAT_CHECK`?

gregwebs: okay, I made dirty checking a separate flag from the individual checks.

ci-build/test.sh
```diff
@@ -8,12 +8,26 @@ panic() {
     exit 1
```
BusyJay, line 8: I don't think this file is used in CI check. Now CI steps are written in job configurations.

gregwebs: Where is the CI code?

BusyJay: Maybe you can ask @zhouqiang-cl for help.

PR #9659, opened 2021-02-18
txn_types: ignore unknown bytes when parsing Write/Lock

youjiali1995:

Problem Summary:

To make downgrading easier, Ignore unknown bytes when parsing Write/Lock. https://github.com/tikv/sig-transaction/pull/93

What's Changed:

Ignore unknown bytes when parsing Write/Lock.

### Check List 

Tests 

- Unit test

### Release note 
- No release note.


sticnarf (approved):
LGTM

nrc (approved):
lgtm

youjiali1995:
@MyonKeminta Downgrading is a dangerous operation, we will provide detailed downgrading docs of each version for users and users should follow it. If some new fields can't be ignored, we need to provide solutions case by case and I don't think we can handle all cases now.

youjiali1995:
BTW, we ignore unknown fields in this PR. Should we ignore unknown types too?
```rust
        let lock_type = LockType::from_u8(b.read_u8()?).ok_or(ErrorInner::BadFormatLock)?;
```
```rust
        let write_type_bytes = b
            .read_u8()
            .map_err(|_| Error::from(ErrorInner::BadFormatWrite))?;
```

PR #9669, opened 2021-02-19
skip to examine snapshots when GC

hicqu:

If there are too many stale snapshots, GC snapshot in raftstore thread could be very slow.

图片

When GC snapshots, skip to examine checksum for them.

### Check List 

Tests 

- Integration test

No new cases are added, but all cases about snapshot GC cover this change.

### Release note 
* speed up snapshot GC

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -677,7 +677,7 @@ where
         let is_applying_snap = s.is_applying_snapshot();
         for (key, is_sending) in snaps {
             if is_sending {
-                let s = match self.ctx.snap_mgr.get_snapshot_for_sending(&key) {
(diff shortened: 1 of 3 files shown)


Connor1996 (commented):
I'm wondering if the delete io operation would cause raftstore latency spike? can we move it out of raftstore thread.

hicqu:
> I'm wondering if the delete io operation would cause raftstore latency spike? can we move it out of raftstore thread

The logic will access `peer_storage` which is not available out from raftstore threads. However it's still possible because we can access RaftDB directly instead, with some more complex code. I prefer to implement it in another PR after this is merged, if it's still necessary. After this PR is merged, no disk I/O will be involved for snapshot GC.

Connor1996 (commented):
LGTM

BusyJay (approved):

Issue #9612, opened 2021-02-01
Support the TTL for rawkv


Connor1996:
## Development Task

Currently, TiKV doesn't support TTL feature, so some users who use rawkv directly need to implement the TTL logic on the applications side, like https://github.com/tikv/tikv/issues/3337. 

To make TiKV self productization, we're planning to support TTL for rawkv with demands:
- Every key can have different TTLs or non-TTL
- When a key-value exceeds the TTL, it doesn't mean the storage space is recycled, but TiKV promises that it can't be read from raw API.
- It provides linearizability within the clock gap among different TiKV nodes. Namely, once the key-value is invalid on the leader, it can't be read out from the followers or the new leader.
- TTL doesn't provide backward compatibility with origin rawkv.

## Design

Rocksdb provides TTL support natively, but it has some limitations so that we won't utilize it to support rawkv TTL.

The key point of the design to store TTL information somewhere so the key-value can be filtered on read or compaction. Like rocksdb, we decide to append the expiration time to the end of values.

With that, the write path changes are:
- user calls tikv-client's raw API with the TTL parameter
- TiKV calculated the expire timestamp by `current unix timestamp + TTL`(if is non-TTL kvs, the expire timestamp would be 0), and appends as 8 bytes to the value at `storage` level so that `raftstore` has no knowledge about TTL. 

Kvs invalidate on compactions:
- Add CompactionFilter to filter the kvs that `current unix timestamp > kv's expire timestamp`
- Some SSTs are not compacted for a long time. To make the storage space is recycled in time, add TTL related properties of SST and check them periodically to manual compact the invalided SSTs.

And last, strip the expiration timestamp on the read path.

### API changes
- [x] proto changes https://github.com/pingcap/kvproto/pull/712
- [x] Add TTL parameter for go-client
- [x] Add TTL parameter for java-client

### TiKV changes 
- [x] Add TTL compaction filter https://github.com/tikv/tikv/pull/9757
- [x] Add TTL properties collector https://github.com/tikv/tikv/pull/9756
- [x] Add TTL checker
- [x] Append expiration timestamp on write
- [x] Strip expiration timestamp on read https://github.com/tikv/tikv/pull/9758
- [x] Do compatibility check on start
- [x] Add metrics

duduainankai:
I would like to help if you don't mind.

Connor1996:
@duduainankai Sorry for not referring to the PRs. The implementation of TiKV is almost done. You can add support for the clients if you want.

[...]

PR #9672, opened 2021-02-20
*: Add TTL support for rawkv[release-5.0]

Connor1996:
Cherry pick of #9595

Issue Number:  https://github.com/tikv/tikv/issues/9612 

Problem Summary: Add TTL support for rawkv

What's Changed:

- Add a `ttl` field for raw put requests
- Calculate a `expire_ts` on storage level, and bring it with `PutCmd` of `RaftCmdRequest`
- Append the `expire_ts` to the value when applying the put cmd
- Strip the `expire_ts` from values when doing get and scan at `RegionSnapshot` level
- Add a `TTLCompactionFilter` to filter the invalidated value
- Add a `TTLCheckAndCompact` task in cleanup worker to compact the invalidated SSTs in time.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

- Add TTL support for rawkv


src/storage/raw/store.rs
```diff
@@ -0,0 +1,194 @@
...
+
+use super::ttl::TTLSnapshot;
+
+use crate::storage::kv::{Result, ScanMode, Snapshot};
+use crate::storage::Statistics;
+
+use engine_traits::{CfName, IterOptions, DATA_KEY_PREFIX_LEN};
+use txn_types::{Key, KvPair};
+
+pub enum RawStore<S: Snapshot> {
```
Little-Wallace, line 11: How about change this into a dynamic trait?

Connor1996: Introduce extra overhead

PR #9674, opened 2021-02-20
enable more logs for tikv-ctl

hicqu:

tikv-ctl can't print logs in `src/*` because it doesn't call `initial_logger`.

Make tikv-ctl print more detail logs in `tikv-ctl.log`, which is in the current directory.

### Check List 

Tests 

For example, `unsafe-recover remove-fail-stores` will record such a log:
```
[2021/02/20 14:18:03.439 +08:00] [INFO] [debug.rs:586] ["peers changed"] [new_peers="[id: 3 store_id: 1]"] [old_peers="[id: 3 store_id: 1]"] [region_id=2]
```

### Release note 
* No release note.

### DIFF
--- a/cmd/src/bin/tikv-ctl.rs
+++ b/cmd/src/bin/tikv-ctl.rs
@@ -1030,6 +1030,15 @@ fn warning_prompt(message: &str) -> bool {
 }
 
 fn main() {
+    // The config is only used to initialize the global logger.
+    let rocksdb_info_dir = tempfile::tempdir().unwrap();
+    let raftdb_info_dir = tempfile::tempdir().unwrap();
+    let mut cfg = TiKvConfig::default();
+    cfg.log_file = "./tikv-ctl.log".to_owned();
+    cfg.rocksdb.info_log_dir = rocksdb_info_dir.path().to_str().unwrap().to_owned();
+    cfg.raftdb.info_log_dir = raftdb_info_dir.path().to_str().unwrap().to_owned();
+    cmd::setup::initial_logger(&cfg);
+
     vlog::set_verbosity_level(1);
 
     let raw_key_hint: &'static str = "Raw key (generally starts with \"z\") in escaped form";
--- a/cmd/src/bin/tikv-ctl.rs
+++ b/cmd/src/bin/tikv-ctl.rs
@@ -1838,8 +1838,8 @@ fn main() {
     if let Some(log) = matches.value_of("log") {
         let mut cfg = TiKvConfig::default();
         cfg.log_file = log.to_owned();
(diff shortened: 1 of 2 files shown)


cmd/src/bin/tikv-ctl.rs
```diff
@@ -1030,6 +1030,18 @@ fn warning_prompt(message: &str) -> bool {
 }
 
 fn main() {
+    // The config is only used to initialize the global logger.
+    let rocksdb_info_dir = tempfile::tempdir().unwrap();
+    let raftdb_info_dir = tempfile::tempdir().unwrap();
+    #[allow(clippy::field_reassign_with_default)]
+    {
+        let mut cfg = TiKvConfig::default();
+        cfg.log_file = "./tikv-ctl.log".to_owned();
```
MyonKeminta, line 1039: Can the log file be configurable instead of hard coded? Also it changed tikv-ctl's behavior, a user familiar with the old version may find that a log file shows up in the working directory unexpectedly.
I suggest it do not write the log file by default, and specify the log file if it's needed.

cmd/src/bin/tikv-ctl.rs
```diff
@@ -1042,6 +1043,12 @@ fn main() {
         .version(version_info.as_ref())
         .long_version(version_info.as_ref())
         .setting(AppSettings::AllowExternalSubcommands)
+        .arg(
+            Arg::with_name("log")
+                .long("log")
+                .takes_value(true)
+                .help("Log file for tikv-ctl")
```
MyonKeminta, line 1050:
```suggestion
                .help("Log file path for tikv-ctl. Only fatal logs will be printed")
```

MyonKeminta (approved):
LGTM. Though I think it would be better if log level can be configured.

cmd/src/bin/tikv-ctl.rs
```diff
@@ -1827,6 +1834,15 @@ fn main() {
 
     let matches = app.clone().get_matches();
 
+    #[allow(clippy::field_reassign_with_default)]
+    if let Some(log) = matches.value_of("log") {
+        let mut cfg = TiKvConfig::default();
+        cfg.log_file = log.to_owned();
+        cfg.rocksdb.info_log_level = LogLevel::Fatal;
```
gregwebs, line 1841: Why only fatal? Is error to verbose?

cmd/src/bin/tikv-ctl.rs
hicqu: `Fatal` is only for RocksDB logs. For general TiKV log, the level is still `info` by default.

PR #9685, opened 2021-02-22
batch-system, raftstore: apply high-latency operations in low-priority thread pool

gozssky:

Problem Summary:

When applying, some operations probably cost more time than other operations , 
such as ingest and delete-range. It will increase the latency of normal operations.

What's Changed:

 This PR adds a low-priority thread pool in batch-system. The high-latency operations will be applied in this pool.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- apply high-latency operations in a specialized thread


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3356,6 +3383,7 @@ where
                         state.pending_msgs = drainer.collect();
                         break;
                     }
+                    self.delegate.priority = Priority::Normal;
```
Little-Wallace, line 3386: This operation will change it to `Normal` priority even if it just processed a high latency command?

gozssky: Yes, restore priority to normal so that other commands can be processed by normal thread.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -942,6 +965,10 @@ where
                     }
                 }
             }
+            if has_high_latency_operation(&cmd) && apply_ctx.priority != Priority::Low {
+                self.priority = Priority::Low;
+                return ApplyResult::Yield;
+            }
```
Little-Wallace, line 971: How about change it priority to `Normal` ?
```
if has_high_latency_operation() {
    self.priority = Priority::Low;
    if apply_ctx.priority != Priority::Low {
        return ApplyResult::Yield;
    }
}  else {
   self.priority = Priority::Normal;
}
```

gozssky: There is a little difference. Changing priority to normal here may not take effect immediately unless yield also, because current fsm may have already scheduled to the low-priority PollHandler due to previous high latency operations. 

We can simply change priority to normal after current fsm has processed finished, so that new batch msgs can always be scheduled to normal-priority PollHandler first.

Little-Wallace: OK. But you must commit the current delete if there is still some data kept in WriteBatch, which is written by previous entries.

gozssky: Yes, they have been committed before this lines.

components/batch-system/src/batch.rs
```diff
@@ -377,30 +392,51 @@ where
...
+                thd_name!(format!("{}-{}", name_prefix, i)),
+                Priority::Normal,
+                &mut builder,
+            );
         }
+        self.start_poller(
+            thd_name!(format!("{}-low", name_prefix)),
+            Priority::Low,
+            &mut builder,
+        );
```
gengliqi, line 439: The store thread pool doesn't need a low-priority thread. I suggest adding a config to fix it.

Little-Wallace: Emmmm, but I want to move the regions which have been a long time no processing data to low-priority thread so that their heartbeat can be processed independently.

gengliqi (approved):

Little-Wallace (approved):
LGTM

NingLin-P (approved):

PR #9686, opened 2021-02-22
raftstore: support batch ingest

gozssky:
Signed-off-by: Yujie Xia <xiayjchn@gmail.com>

Problem Summary:

Ingest costs more time than other operations. It will cause high latency of apply-wait.

What's Changed:

To reduce the impact on apply-wait latency, this PR collect multiple ingest requests and ingest them at a time.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

- support batch ingest in raftstore


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -492,6 +496,18 @@ where
             self.kv_wb_last_bytes = 0;
             self.kv_wb_last_keys = 0;
         }
+        if !self.pending_ssts.is_empty() {
+            let tag = self.tag.clone();
+            self.importer
+                .ingest(&self.pending_ssts, &self.engine)
```
Little-Wallace, line 502: Importer could only ingest files whose column family are same. So you must group them by cf.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1354,7 +1383,15 @@ where
                 }
                 CmdType::IngestSst => {
                     assert!(ctx.kv_wb.is_empty());
-                    self.handle_ingest_sst(&ctx.importer, &ctx.engine, req, &mut ssts)
+                    assert_eq!(requests.len(), 1);
+                    if self.pending_sst.is_none() {
+                        let sst = req.get_ingest_sst().get_sst();
+                        self.handle_ingest(&ctx.importer, sst)?;
+                        ctx.pending_ssts.push(sst.to_owned());
+                        return Ok((RaftCmdResponse::default(), ApplyResult::Yield));
```
Little-Wallace, line 1391: It is not necessary to make this fsm yield.....  In fact, we do not hope the fsm yield when there is no pending sst.
If there has been a pending sst, we only need to call `ctx.commit(self)` at once.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1354,7 +1383,15 @@ where
...
+                    }
+                    ssts.push(self.pending_sst.take().unwrap());
```
Little-Wallace, line 1393: If there is a pending sst, the current request will be discarded, will not it?

components/raftstore/src/store/fsm/apply.rs
Little-Wallace: Emmm, I see, You group them in method of `impoter`, It is ok.

components/raftstore/src/store/fsm/apply.rs
gozssky: The sst of the current request must be the same as the pending sst. Because it is saved here before last yield.

components/raftstore/src/store/fsm/apply.rs
gozssky: There are two `pending_sst` here. 😂

`ApplyContext.pending_ssts` is the real pending ssts. These ssts will be ingested in `write_to_db`.
`ApplyDelegate.pending_sst` just saved a `SSTMetaInfo` for responsing ingest request after recovering from yield. (Maybe we can find a better name)
But I actually found a bug here. In `handle_raft_entry_normal` , `ctx.commit(self)` is called every time because of the ingest request. `ApplyContext.pending_ssts` will always be empty here.

Do you have any idea that we can avoid yield when first applying a ingest request?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -959,15 +974,22 @@ where
...
-                    apply_ctx.commit(self);
+                    if has_unsynced_data {
+                        apply_ctx.commit(self);
+                    }
                     return ApplyResult::Yield;
                 }
             }
             if should_write_to_engine(&cmd) || apply_ctx.kv_wb().should_write_to_engine() {
-                apply_ctx.commit(self);
+                if has_unsynced_data {
```
Little-Wallace, line 990: If there is some admin request or write batch is too large, we must commit it....
I suggest you better handle these two situations(admin request and high latency operation) separately.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1354,7 +1376,8 @@ where
                 }
                 CmdType::IngestSst => {
                     assert!(ctx.kv_wb.is_empty());
-                    self.handle_ingest_sst(&ctx.importer, &ctx.engine, req, &mut ssts)
+                    assert_eq!(requests.len(), 1);
```
Little-Wallace, line 1379: No. There may be more than one request.

Little-Wallace: And `assert!(ctx.kv_wb.is_empty());` shall be removed.

components/sst_importer/src/sst_importer.rs
```diff
@@ -657,40 +662,52 @@ impl ImportDir {
...
 
         let mut opts = E::IngestExternalFileOptions::new();
         opts.move_files(true);
-        engine.ingest_external_file_cf(cf, &opts, &[path.clone.to_str().unwrap()])?;
-
+        for (cf, cf_paths) in paths {
+            let files: Vec<&str> = cf_paths.iter().map(|p| p.clone.to_str().unwrap()).collect();
+            engine.ingest_external_file_cf(cf, &opts, &files)?;
+        }
+        IMPORTER_INGEST_BYTES.observe(ingest_bytes as _);
```
Little-Wallace, line 706: Better add some metric(histogram) to observe that how many files are ingested in once method.

Little-Wallace (approved):
LGTM

gozssky:
@Little-Wallace PTAL again, /cc @gengliqi

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -958,21 +974,22 @@ where
         if !data.is_empty() {
             let cmd = util::parse_data_at(data, index, &self.tag);
 
-            if apply_ctx.yield_high_latency_operation && has_high_latency_operation(&cmd) {
+            if has_high_latency_operation(&cmd) && apply_ctx.yield_high_latency_operation {
```
gengliqi, line 977: Seems the former logic is more efficient if `yield_high_latency_operation` is false.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -958,21 +974,22 @@ where
...
                 self.priority = Priority::Low;
-                if apply_ctx.priority != Priority::Low {
-                    apply_ctx.commit(self);
-                    return ApplyResult::Yield;
-                }
             }
-            if should_write_to_engine(&cmd) || apply_ctx.kv_wb().should_write_to_engine() {
+            let need_sync = should_write_to_engine(&cmd)
+                && self.last_sync_apply_index != self.apply_state.get_applied_index();
```
gengliqi, line 981: When does `last_sync_apply_index` equal to `self.apply_state.get_applied_index()`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -449,11 +450,9 @@ where
     ///
     /// This call is valid only when it's between a `prepare_for` and `finish_for`.
     pub fn commit(&mut self, delegate: &mut ApplyDelegate<EK>) {
-        if self.last_applied_index < delegate.apply_state.get_applied_index() {
+        if delegate.last_sync_apply_index < delegate.apply_state.get_applied_index() {
```
gengliqi, line 453: How about changing this variable's name to `last_sync_applied_index`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -471,6 +471,21 @@ where
     /// If it returns true, all pending writes are persisted in engines.
     pub fn write_to_db(&mut self) -> bool {
         let need_sync = self.sync_log_hint;
+        // There may be put and delete requests after ingest request in the same fsm.
+        // To guarantee the correct order, we must ingest the pending_sst first, and
+        // then persist the kv write batch to engine.
+        if !self.pending_ssts.is_empty() {
```
gengliqi, line 477: I am worried about this operation may break the correctness of the transaction. I do not think carefully yet. Could you please explain why it's correct or writing comments in code is better.

components/raftstore/src/store/fsm/apply.rs
gozssky: After all the changes in `kv_wb` and `pending_sst` are write to db,  `last_sync_apply_index` will be equal to `self.apply_state.get_applied_index()`, see also  [`commit_opt`](https://github.com/gozssky/tikv/blob/batch-ingest/components/raftstore/src/store/fsm/apply.rs#L464).

components/raftstore/src/store/fsm/apply.rs
gozssky: When a fsm process an ingest command, it will ensure all the changes before this point are write to db by checking `self.last_sync_apply_index != self.apply_state.get_applied_index()` first. And then the sst meta will be added to `pending_ssts` for later ingesting. 
If some other write commands come before the above sst is ingested, they will be write to `kv_wb`. So in `write_to_db`, we must ingest the ssts first. Because the ingest command happens before the write commands in `kv_wb` for a specific fsm.

components/raftstore/src/store/fsm/apply.rs
gengliqi: Oh, I understand it. The code here just for not writing twice for the same cmd?

gengliqi:
The `last_sync_apply_index` means the applied index of synced write. It is used by snapshot though the snapshot logic can be removed and `last_sync_applied_index` can be removed too(it's a legacy). I think you can reserve the `last_sync_apply_index` and add another `last_applied_index` for your purposes.

components/raftstore/src/store/fsm/apply.rs
gengliqi: The ingest command may be reordered with the normal write command.
For example, 1. ingest(contain key A), 2. write key A, 3. ingest.
The number 2 command will be written to kv db before ingest command due to batch.
So the key A will be wrong.
Is this case right? Or I miss the point?

gozssky: For a same fsm, there is at most one ingest command in `pending_ssts`. When processing no.3 command, we will flush the previous changes to db.

components/raftstore/src/store/fsm/apply.rs
gozssky: Not really. The code here is mainly for ensuring all changes before this command are flushed, because we may bypass `kv_wb` to write data, for example ingest sst directly.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -960,19 +976,20 @@ where
...
                 apply_ctx.commit(self);
                 if let Some(start) = self.handle_start.as_ref() {
                     if start.elapsed() >= apply_ctx.yield_duration {
                         return ApplyResult::Yield;
                     }
                 }
             }
+            if self.priority != apply_ctx.priority {
+                return ApplyResult::Yield;
+            }
```
gengliqi, line 992: Before `Yield`, the `commit` must be called otherwise the data may be wrong due to reordering.

gozssky: OK, I will add it.

gengliqi (approved):
LGTM

gozssky:
> The `last_sync_apply_index` means the applied index of synced write. It is used by snapshot though the snapshot logic can be removed and `last_sync_applied_index` can be removed too(it's a legacy). I think you can reserve the `last_sync_apply_index` and add another `last_applied_index` for your purposes.

@gengliqi  I reset the usage of original `last_sync_apply_index` and use `ApplyContext.last_applied_index` instead.  PTAL

Little-Wallace:
No, I do not think `last_applied_index` of `ApplyContext` is necessary.  We can judge whether need to update apply_state to writebatch just by `last_sync_apply_index` of `delegate`.

Little-Wallace:
may be we shall update it to `last_flush_apply_index `

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -479,11 +480,9 @@ where
     ///
     /// This call is valid only when it's between a `prepare_for` and `finish_for`.
     pub fn commit(&mut self, delegate: &mut ApplyDelegate<EK>) {
-        if self.last_applied_index < delegate.apply_state.get_applied_index() {
+        if delegate.last_flush_apply_index < delegate.apply_state.get_applied_index() {
```
Little-Wallace, line 483: please update last_flush_apply_index after write_to_db.

gozssky:
@Little-Wallace @gengliqi PTAL again

gengliqi (approved):

Little-Wallace (approved):
LGTM

Issue #9627, opened 2021-02-04
Potential deadlock for GC in Compaction Filter


hicqu:
### What version of TiKV are you using?
Master. And 5.0-rc also has the defect if GC in Compaction Filter is enabled.

### Steps to reproduce
* Start a small cluster with 4 TiKV instance, 10GB data for each store.
* Change store limit from 15 (by default) to 100, and enable shuffle region scheduler.
* Enable the feature and run a sysbench client.

### What did you expect?
The sysbench client can finish successful.

### What did happened?
After about a half hour, we can see that all apply threads and RocksDB threads are blocked forever.

### Analysis
The attached file is the stack for one TiKV instance.
[stack.txt](https://github.com/tikv/tikv/files/5923415/stack.txt)

With the stack we can know how the dead lock occurs:
* In the compaction filter, `DB::write` is called to delete stale versions on default CF.
* In this time write stall occurs, so the call `DB::write` will be blocked.
* To remove the write stall condition, more compactions are necessary. But all compaction threads could be blocking at `DB::write`.
* So it will be a dead lock.

#### Don't clean default CF in write CF's compaction filter.
So that write stall in write CF's compaction filter won't happen again.
And for default CF, we can also use a compaction filter to handle stale versions like https://github.com/tikv/tikv/pull/8348.

Pros:
* Resolve the root cause for the problem

Cons:
* Compatibility with Lighting
* If the feature is disabled, GC worker can't handle orphan versions on default CF

#### Support to bypass write stall checkers in `WriteOptions`
So that write stall in write CF's compaction filter won't happen again.

Pros:
* Less change in TiKV
* Won't leave orphan versions on default CF

Cons:
* Not easy to implement

#### Use `WriteOptions::no_slowdown`
If writting a write batch into RocksDB fails with an `InComplete` error, skip to perform more GC and send the current write batch to another thread to write them.

Pros:
* Easy to implement
* Less change in TiKV

Cons:
* If the TiKV fails after the batch is sent to another thread but not written, orphan versions could be left on the default CF.

### Progress
After some discussions we choose the third solution. Patch PR will comes soon.

PR #9694, opened 2021-02-23
avoid deadlock for GC in compaction filter

hicqu:

Issue Number: close https://github.com/tikv/tikv/issues/9627.

* Use `no_slowdown` flag to delete keys on default CF in compaction filter to avoid block when stall
* GC worker helps to clean orphan versions on default CF, the task is emit from compaction filter

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
With the patch there is no deadlock any more:
图片

- Bench
*disable*: Disable the feature GC in compaction filter
*patch.1*: COMMIT HASH: a94ae74, compaction filter just sends all deletion marks to the GC worker.
*patch.2*: COMMIT HASH: 3c9d63a, compaction filter only sends deletion marks not followed by overlapped versions. This little optimization can reduce useless modifications on write CF.
图片
图片
图片

Side effects

- In an extreme case, orphan versions on default CF can be left forever.

### Release note 
* avoid dead lock  in compaction filter during write stall


src/server/gc_worker/gc_worker.rs
```diff
@@ -500,6 +509,14 @@ where
                     limit,
                 );
             }
+            GcTask::OrphanVersions(wb) => {
+                info!("handling GcTask::OrphanVersions"; "size" => wb.as_inner().count());
```
yiwu-arbug, line 513: can we give the message an ID and print both in compaction filter and here? so in the log we can match them.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -67,6 +69,10 @@ lazy_static! {
         "Compaction filter meets failure"
     )
     .unwrap();
+    static ref GC_COMPACTION_FILTER_ORPHAN_VERSIONS: IntCounter = register_int_counter!(
```
yiwu-arbug, line 72: let also add stats for the total number of orphan versions that's deleted.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -104,11 +110,24 @@ lazy_static! {
...
+    /// associated versions on default CF are deleted with `RocksEngine::delete` API before the
+    /// compaction result gets installed. However, it's possible that `RocksEngine::delete` fails
+    /// because of write stall, so it's necessary to give a fallback position for those orphan
+    /// versions instead of drop them directly.
+    ///
+    /// NOTE: `orphan_versions_handler` can only handles a `RocksWriteBatch` asynchronously
+    /// because at the time the instance has already been stalled. So it's possible that the TiKV
+    /// instance fails after a compaction result is installed but its orphan versions are not
+    /// deleted. Those orphan versions will never get cleaned until `DefaultCompactionFilter` is
+    /// introduced.
```
yiwu-arbug, line 124: btw Let's open a separate TiKV issue to track if this gets fixed in the future.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -180,13 +194,19 @@ impl CompactionFilterFactory for WriteCompactionFilterFactory {
...
         let db = Arc::clone(&gc_context.db);
+        let orphan_versions_handler = gc_context.orphan_versions_handler.clone();
 
         debug!(
             "creating compaction filter"; "feature_enable" => enable,
             "skip_version_check" => skip_vcheck,
             "ratio_threshold" => ratio_threshold,
         );
 
+        if db.get_property_int("rocksdb.is-write-stalled").unwrap() != 0 {
```
yiwu-arbug, line 205: avoid unwrap. skip compaction filter as well if this gets error.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -567,19 +625,17 @@ impl Drop for WriteCompactionFilter {
     // becomes installed into the DB instance.
     fn drop(&mut self) {
         self.handle_delete_mark();
-        self.switch_key_metrics();
-        if !self.write_batch.is_empty() {
-            self.write_batch.write().unwrap();
-            self.write_batch.clear();
-        }
+        let _ = self.flush_pending_writes_if_need(true);
```
yiwu-arbug, line 628: print error here?

src/server/gc_worker/compaction_filter.rs
```diff
@@ -491,21 +537,33 @@ impl WriteCompactionFilter {
...
+            fail_point!("write_compaction_filter_flush_write_batch", true, |_| {
+                Err("Ingested fail point".to_owned())
+            });
+            db.write_opt(wb.as_inner(), wopts)
+        }
+
+        if self.write_batch.count() > DEFAULT_DELETE_BATCH_COUNT || force {
+            let db = self.engine.as_inner();
+            let mut wopts = WriteOptions::default();
+            wopts.set_no_slowdown(true);
```
yiwu-arbug, line 559: btw do we do sync write for GC?

yiwu-arbug (commented):
rest looking good

src/server/gc_worker/compaction_filter.rs
```diff
@@ -433,21 +475,25 @@ impl WriteCompactionFilter {
...
             if need_delete {
-                self.delete_write_key(key)?;
+                wb.delete_cf(CF_WRITE, key)?;
                 self.handle_filtered_write(parse_write(value)?)?;
             } else {
                 self.mvcc_delete_skip_older += 1;
             }
             valid = write_iter.next()?;
         }
+        // TODO: Can't use `no_slowdown` until canceling compaction results is available.
```
yiwu-arbug, line 485: ah, now I see we need to cancel compaction. I remember you have a PR to add it?

src/server/gc_worker/compaction_filter.rs
hicqu: Yes, the write batch is sync written in GC worker.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -180,13 +199,23 @@ impl CompactionFilterFactory for WriteCompactionFilterFactory {
...
+        let orphan_versions_handler = gc_context.orphan_versions_handler.clone();
 
         debug!(
             "creating compaction filter"; "feature_enable" => enable,
             "skip_version_check" => skip_vcheck,
             "ratio_threshold" => ratio_threshold,
         );
 
+        if db
+            .get_property_int("rocksdb.is-write-stalled")
```
Connor1996, line 211: seems there is no property called that, but a property named `is-write-stopped`. I suggest leveraging the hook `on_stall_conditions_changed` to record a global stall status in TiKV.

hicqu: It's a new added property.

Connor1996: oh, I see

Connor1996: should consider `is-write-stopped` too

hicqu: I guess `stalled` contains `stopped`?

src/server/gc_worker/compaction_filter.rs
```diff
@@ -199,138 +189,142 @@ impl CompactionFilterFactory for WriteCompactionFilterFactory {
...
-            //   L6: |key_110_MVCC_PUT|
-            // If key_130_MVCC_DEL is handled at L0 and the SST in L6 is not involved,
-            // key_110_MVCC_PUT could be exposed incorrectly.
-            let db = filter.engine.as_inner();
-            let cf = db.cf_handle(CF_WRITE).unwrap();
-            let iter = DBIterator::new_cf(db.clone(), cf, ReadOptions::new());
-            filter.write_iter = Some(RocksEngineIterator::from_raw(iter));
+    fn schedule_gc_task(&self, task: GcTask) {
+        if let Err(e) = self.gc_scheduler.schedule(task) {
+            error!("compaction filter schedule {} fail", e.0);
```
yiwu-arbug, line 281: nit: would this cause too many info log when gc_scheduler rejects all tasks?

yiwu-arbug (commented):
rest looking good to me.

components/tikv_util/src/worker/future.rs
```diff
@@ -51,7 +51,7 @@ pub fn dummy_scheduler<T: Display>() -> Scheduler<T> {
 }
 
 impl<T: Display> Scheduler<T> {
-    fn new<S: Into<Arc<str>>>(name: S, sender: UnboundedSender<Option<T>>) -> Scheduler<T> {
+    pub fn new<S: Into<Arc<str>>>(name: S, sender: UnboundedSender<Option<T>>) -> Scheduler<T> {
```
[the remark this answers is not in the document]
hicqu: Theoretically yes it could cause OOM. However I couldn't find a way to resolve this. So how about handling it later?

src/server/gc_worker/compaction_filter.rs
```diff
@@ -74,33 +69,18 @@ lazy_static! {
...
-
     // `WriteType::Rollback` and `WriteType::Lock` are handled in different ways.
     static ref GC_COMPACTION_MVCC_ROLLBACK: IntCounter = register_int_counter!(
         "tikv_gc_compaction_mvcc_rollback",
         "Compaction of mvcc rollbacks"
     )
     .unwrap();
+
+    pub static ref GC_COMPACTION_FILTER_ORPHAN_VERSIONS: IntCounterVec = register_int_counter_vec!(
+        "tikv_gc_compaction_orphan_versions",
```
Connor1996, line 80:
```suggestion
        "tikv_gc_compaction_filter_orphan_versions",
```

src/server/gc_worker/compaction_filter.rs
```diff
@@ -199,138 +181,150 @@ impl CompactionFilterFactory for WriteCompactionFilterFactory {
...
         }
-        filter
     }
 
     fn do_filter(
         &mut self,
         _start_level: usize,
         key: &[u8],
-        sequence: u64,
+        _sequence: u64,
```
Connor1996, line 305: seems it's useless now

hicqu: Yes, we can revert relative changes in `rust-rocksdb` later.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -345,20 +339,18 @@ impl WriteCompactionFilter {
                 WriteType::Delete => {
```
Connor1996, line 339: how are the stale versions of `WriteType::Put` in other files handled?

src/server/gc_worker/gc_worker.rs
```diff
@@ -291,6 +288,44 @@ where
...
+                        "GC deleted plenty versions for a key";
+                        "key" => %key,
+                        "versions" => gc_info.deleted_versions,
+                    );
+                }
+                next_gc_key = keys.next();
+                gc_info = GcInfo::default();
+            } else {
+                Self::flush_txn(txn, &self.limiter, &self.engine)?;
+                txn = Self::new_txn(self.engine.snapshot_on_kv_engine(b"", b"")?);
```
Connor1996, line 322: why need to new_txn for every key?

Connor1996 (commented):
rest LGTM

Connor1996 (approved):
LGTM

PR #9695, opened 2021-02-23
copr: replace custom BitVec with a crate

### LINKED ISSUE #8950: copr: BitVec should implement PartialEq correctly
## Development Task

Currently we use the default derive of `BitVec`. However, when `BitVec` get truncated, default `PartialEq` implementation may not work correctly.

pvid:

Issue Number: close #8950

Problem Summary:

What's Changed:

Replaced custom `BitVec` implementation with the [bit-vec](https://crates.io/crates/bit-vec) crate

I'd still like to do at least some benchmarking before merging

### Check List

Tests

- removed custom implementation - compatibility is tested using rest of the existing test suite

### Release note

- No release note


components/tidb_query_codegen/src/rpn_function.rs
```diff
@@ -1581,7 +1581,13 @@ impl NormalRpnFn {
                 return Ok(#vec_type::cast_chunk_into_vector_value(result));
             }
 
-            for (row_index, val) in BitAndIterator::new(vecs.as_slice(), output_rows).enumerate() {
+            let merged_bitmap = vecs
+                .iter()
+                .fold(BitVec::from_elem(output_rows, true), |mut acc, x| {
+                    acc.and(x);
+                    acc
```
skyzh, line 1588: Can we use something like `BitAndIterator` before? It seems that using `iter` and `fold` will lead to an extra BitMap allocation upon calling vectorized functions.

components/tidb_query_datatype/src/lib.rs
```diff
@@ -22,6 +22,8 @@ extern crate bitflags;
 #[allow(unused_extern_crates)]
 extern crate tikv_alloc;
 
+extern crate bit_vec;
```
skyzh, line 25: Seems not necessary to add this line.

skyzh (commented):
Mostly LGTM, thanks for your work!

components/tidb_query_datatype/src/codec/data_type/chunked_vec_set.rs
```diff
@@ -75,7 +74,7 @@ impl ChunkedVec<Set> for ChunkedVecSet {
     }
 
     fn capacity(&self) -> usize {
-        self.bitmap.capacity().max(self.value.capacity())
+        self.bitmap.capacity().min(self.value.capacity())
```
Xuanwo, line 77: Why changed to `min`?

components/tidb_query_datatype/src/codec/data_type/chunked_vec_bytes.rs
```diff
@@ -44,8 +44,7 @@ impl ChunkedVecBytes {
     }
     #[inline]
     pub fn get(&self, idx: usize) -> Option<BytesRef> {
-        assert!(idx < self.len());
-        if self.bitmap.get(idx) {
+        if self.bitmap.get(idx).unwrap() {
```
Xuanwo, line 47: If we can make sure `idx` will not out of bounds, we can use `self.bitmap[idx]` instead.

pvid: missed the `Index` implementation 🤦  I'll add the assert again

```suggestion
        assert!(idx < self.len());
        if self.bitmap.get[idx] {
```

components/tidb_query_datatype/src/codec/data_type/chunked_vec_set.rs
pvid: I encountered some failing tests because the capacity of `bit_vec::BitVec` behaves a bit differently than the custom `BitVec` - it is always a multiple of 32.

My thinking is that in this instance for example the capacity should be the minimum of the two, because to add a new element without an allocation, you need to have an empty slot in *both* data structures. My reasoning is similar in `chunked_vec_enum.rs`

components/tidb_query_codegen/src/rpn_function.rs
pvid: Sure, I can get `BitAndIterator `. My plan was to go with a simple approach on my first pass and look into the performance impact before merging.

Folding with `BitVec::and` should be more efficient than doing similar work element-by-element because it is done on whole blocks. The allocation of the initial value for the fold could be eliminated in case of non-empty `vecs` - the fold can use tail of `vecs` and fold in the tail. Do you think this approach is worth pursuing?

skyzh: My previous approach of BitAndIterator is to bit-and a word at a time, instead of doing it element by element. I guess bit_vec may have similar implementation for such iterator.

PR #9696, opened 2021-02-24
error_code: remove default implementation

gregwebs:

default <T> to unknown means the compiler cannot tell us if error codes
are working. It is easy to accidentally end up with the unknown code.
It also requires all implemntations to use specialization.

Remove defaulting to unknown
use an error_unknown macro if there is no error code to set it to
unknown.

### Related changes

None

### Check List

Tests

Tested a case where the error code shows up after these changes. Previously wrappers were getting defaulted to unknown but are now properly dereferenced.

Side effects

Code may change from Unknown to a more specific code.

### Release note

Fix error code propagation

### DIFF
--- a/components/tikv_util/src/log.rs
+++ b/components/tikv_util/src/log.rs
@@ -33,6 +33,10 @@ macro_rules! error {
     };
 }
 
+// error_unknown is used the same as the above error macro
+// However, it will always use an error code of UNKNOWN
+// This is for errors that do not implement ErrorCodeExt
+// It is recommended to implement ErrorCodeExt instead of using this macro
 #[macro_export]
 macro_rules! error_unknown {
     (?$e:expr; $l:literal) => {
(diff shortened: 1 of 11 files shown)


components/tikv_util/src/log.rs
```diff
@@ -33,6 +33,25 @@ macro_rules! error {
     };
 }
 
+#[macro_export]
+macro_rules! error_unknown {
```
5kbpers, line 37: Please add some comments to explain the intention of this macro

5kbpers (commented):
rest LGTM

components/tikv_util/src/log.rs
gregwebs: 👍 comment added.

5kbpers (approved):

nrc (approved):
lgtm

PR #9698, opened 2021-02-24
compaction filter: avoid redundant seeks

hicqu:

Write CF's compaction filter utilizes too much CPU on `seek`. Here is a flame graph:
图片

The reason is that there are lots of redundant `seek` when the range is cleared by `delete_files_in_range`.

Check the iterator position before `seek` or `next` in compaction filter.

### Check List 

Tests 

- Unit test

### Release note 
* No release note.


yiwu-arbug (approved):
LGTM

tabokie (approved):

PR #9706, opened 2021-02-24
external_storage: move cloud code to separate cloud crate

gregwebs:
TODO:
- [X] merge [CloudDynamic kvproto upstream](https://github.com/pingcap/kvproto/pull/737)
- [x] merge [kvproto GRPC service definitions upstream](https://github.com/pingcap/kvproto/pull/748)

WIP:
- [X] test external_storage against S3
- [X] test external_storage against GCS
- [X] test kvproto change

We are organizing the cloud code in a separate crate. This will facilitate the cloud team being able to own the code and modifying builds to include or exclude cloud providers or in the future dynamically link them.

The encryption code reoganization is already merged here: https://github.com/tikv/tikv/pull/9647
This PR is for the external_storage code.

* AWS and GCP code are moved from the external_storage crate into the cloud/aws and cloud/gcp crates
* components/external_storage/export is added to avoid cyclical dependencies. Cloud constructors of external_storage traits are imported here and then re-exported along with the external_storage crate for end users of external_storage
* rusoto_util code was already moved to aws/util in the encryption PR and in this PR we just delete the rusoto_util crate
* There is a flag `cloud-storage-grpc` for an experimental mode where external storage is served from a separate GRPC service over a unix domain socket.
* There is a flag `cloud-storage-dylib` for an experimental mode where external storage is called via C bindings. This code uses most of the same code from the GRPC flag. This code can be linked statically or loaded dynamically.

### Check List

Tests

- Unit test
- Manual test (WIP)

scli with localstack as s3. Tested both a text file (README) and a binary file (scli)

```
AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test ./target/debug/examples/scli --storage S3 --endpoint http://localhost:4566 --region us-west-2 --bucket backup --name ext_storage --path . --file README.md save
aws s3 --endpoint http://localhost:4566 cp s3://backup/ext_storage .
diff ext_storage README.md
```

br with localstack as S3. Test dump a database, delete the database, then restore it. Restore uses same arguments as backup. Also test this after building with the flag `cloud-storage-dylib`.

```
AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test ~/go/src/github.com/pingcap/br/bin/br --version backup full -s s3://backup/full2 --s3.endpoint http://localhost:4566 --s3.provider aws --s3.region us-west-2
```

br with fake-gcs server as gcs (uses some flags that I recently added on master).

```
br --gcs.endpoint http://storage.gcs.127.0.0.1.nip.io:4443/ --no-credentials --send-credentials-to-tikv=false --skip-check-path backup -s gcs://backup/alpha full
```

Delete the database from TiDB, after restore see that it exists again

```
br --gcs.endpoint http://storage.gcs.127.0.0.1.nip.io:4443/ --no-credentials --send-credentials-to-tikv=false restore -s gcs://backup/alpha full
```

Test cloud dynamic protobuf. Tested for AWS S3 as well. Tested setting both --host and --endpoint for GCS.
Below is the save command. Then test by doing the load command and performing a diff.

```
./target/debug/examples/scli --storage cloud --cloud-name gcp --host http://localhost:4443/ --bucket backup --name ext_storage9 --path . --file README.md save
```

Test experimental GRPC service. Run backup with binaries running that are built from these steps:

```
ENABLE_FEATURES=cloud-storage-grpc make build
./scripts/env cargo build -p external_storage_export --features=protobuf-codec,cloud-gcp,cloud-aws,cloud-storage-grpc && RUST_BACKTRACE=1 RUST_LIB_BACKTRACE=1 ./target/debug/tikv-cloud-storage
```

Dynamically loading the external storage code:

```
# manually edit the Cargo manifest to build a cdylib

# now build the lib
./scripts/env cargo build -p external_storage_export --features=cloud-storage-dylib,cloud-gcp,cloud-aws,protobuf-codec

# ensure the lib is int the working directory of tikv-server
cp target/debug/libexternal_storage_export.so .
```

If the lib cannot be dynamically loaded, you will see warnings during the backup in the logs.

Compatibilty
* Programs built from the TiKV lib (tikv-server, etc) now import the encryption_export crate instead of encryption directly.

### Release note

Move external_storage cloud provider code out of external_storage and into separate cloud provider crates.


components/external_storage/export/src/export.rs
```diff
@@ -0,0 +1,230 @@
...
+    } else {
+        Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            format!("no storage backend {:?}", storage_backend),
+        ))
+    }
+}
+
+#[cfg(any(feature = "cloud-gcp", feature = "cloud-aws"))]
+fn blob_store(store: Box<dyn BlobStorage>) -> Box<dyn ExternalStorage> {
```
lcwangchao, line 38: Any difference between BlobStorage and ExternalStorage?

components/external_storage/export/src/service.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl Service {
+    /// Create a new backup service.
+    pub fn new(storages: Storages) -> Service {
+        Service { storages }
+    }
+}
+
+impl proto::ExternalStorage for Service {
+    fn start(
```
lcwangchao, line 24: May be we can  handover the storage preparing works to grpc service itself. The grpc service can read configurations which contains a storage list  from a local file when it starts up.

components/external_storage/export/src/service.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn finish(
+        &mut self,
+        _ctx: RpcContext,
+        req: proto::ExternalStorageFinishRequest,
+        sink: UnarySink<proto::ExternalStorageFinishResponse>,
+    ) {
+        let storage_id = req.get_storage_id();
+        self.storages.remove(storage_id);
+        sink.success(proto::ExternalStorageFinishResponse::default());
+    }
```
lcwangchao, line 56: What other methods look such as writing data to a file? Maybe a multipart uploading interface should be exposed

components/external_storage/export/src/service.rs
gregwebs: This could work well for many use cases. And I think we can support this pre-loading. However, currently the br tool can specify an arbitrary backend configuration when it requests a backup.

components/external_storage/export/src/service.rs
gregwebs: yes, currently this is incomplete and I am looking into the upload/download part.

components/external_storage/export/src/export.rs
gregwebs: Currently there is no difference. This gives us some flexibility to change things in the future. But the real reason for duplicate traits is to avoid crates depending on each other. external_storage is not a dependency of cloud/* and cloud/* is not a dependency of external_storage. This crate, external_storage_export, binds them both together.

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,191 @@
...
+impl BlobStorage for Box<dyn BlobStorage> {
+    fn name(&self) -> &'static str {
+        (**self).name()
+    }
+
+    fn url(&self) -> io::Result<url::Url> {
+        (**self).url()
+    }
+
+    fn write(
```
yiwu-arbug, line 34: put_object/get_object is more of "blob storage" than read/write?

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct BucketConf {
+    pub endpoint: Option<StringNonEmpty>,
+    pub region: Option<StringNonEmpty>,
+    pub bucket: StringNonEmpty,
+    pub prefix: Option<StringNonEmpty>,
+    pub storage_class: Option<StringNonEmpty>,
```
yiwu-arbug, line 114: why not use empty string to represent none? That we don't need to introduce `StringNonEmpty`.

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+
+    pub fn url(&self, scheme: &str) -> Result<url::Url, String> {
+        let path = match &self.prefix {
+            None => String::new(),
+            Some(sne) => (*sne).to_string(),
+        };
+        if let Some(ep) = &self.endpoint {
+            let mut u = url::Url::parse(&ep).expect("bucket endpoint");
```
yiwu-arbug, line 134: Do not panic. return error instead.

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+impl std::fmt::Display for StringNonEmpty {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct BucketConf {
```
yiwu-arbug, line 109: Can this be
```
pub struct BucketConf(backup::CloudDynamic);
```

components/cloud/src/blob.rs
gregwebs: Yes. Currently there is no difference with ExternalStorage so it made sense to keep the name the same until the API actually changes to better match blob storage.

components/cloud/src/blob.rs
gregwebs: I like having separate types for API and for domain modeling. protobuf types are very difficult to work with because we should be supporting both protobuf and prost and everything about their usage is different. There is an overhead to copying data around, but it doesn't matter for this metadata.

components/cloud/src/blob.rs
gregwebs: If an empty string represents none then every function must defensively check for an empty string. `StringNonEmpty` is a proof of an empty string: this check is done once and then non longer needs to be done anywhere else in the code. Making the caller prove data is already correct is a very powerful paradigm for reducing errors. Just the same as how Rust uses Option to prove that data is not null.

components/cloud/aws/src/kms.rs
```diff
@@ -13,7 +13,7 @@ use cloud::error::{Error, KmsError, Result};
 use cloud::kms::{Config, DataKeyPair, EncryptedKey, KeyId, KmsProvider, PlainKey};
 
 const AWS_KMS_DATA_KEY_SPEC: &str = "AES_256";
-const AWS_KMS_VENDOR_NAME: &[u8] = b"AWS";
+const AWS_KMS_VENDOR_NAME: &str = "AWS";
 pub const AWS_VENDOR_NAME: &str = "aws";
```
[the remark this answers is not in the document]
gregwebs: The encryption code already started using uppercase. I don't want to break backwards compatibility. The backup tools and code have tended towards lower case.

components/cloud/aws/src/s3.rs
```diff
@@ -0,0 +1,593 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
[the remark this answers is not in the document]
gregwebs: `git mv` doesn't matter. The issue is that in addition to the move some changes to the file have been made. In the past I used to try to break these things up into multiple commits but I don't think anyone ever looks at individual commits no matter how helpful it is.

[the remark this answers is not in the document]
gregwebs: You can diff this with the git CLI rather than github. Then you have the ability to tweak the diff parameters. Note that you can already do `?w=` on github, but you can't adjust `-M --find-renames`

```
git diff -w -M45% origin/master gregwebs/cloud-storage
```

components/cloud/gcp/src/gcs.rs
```diff
@@ -0,0 +1,587 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
gregwebs, line 1: View this diff with

```
git diff -w -M40% origin/master gregwebs/cloud-storage
```

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    fn url(&self) -> io::Result<url::Url>;
+
+    /// Write all contents of the read to the given path.
+    fn write(
+        &self,
+        name: &str,
+        reader: Box<dyn AsyncRead + Send + Unpin>,
+        content_length: u64,
+    ) -> io::Result<()>;
```
lcwangchao, line 19: Maybe we can make the io methods under BlobStorage async. It seems that async method are required for implementing the grpcio service and its better to usage **BlobStorage** trait directly. When tikv want use grcp to read/write file, it should use an object implemented the **ExternalStorage** by wrapping a grpcio client to make it.

gregwebs: This is implemented now in the latest commit. I did not convert the methods to be async because the compiler didn't like that idea, but I hope we will eventually.

components/external_storage/export/src/service.rs
gregwebs: The experimental GRPC service is completed now.

components/external_storage/export/src/service.rs
gregwebs: The GRPC API has changed a lot now.

components/cloud/aws/src/kms.rs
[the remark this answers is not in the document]
yiwu-arbug: Can we have better naming to denote their usage?

components/cloud/src/blob.rs
yiwu-arbug: but "read/write" is confusing in the context of blob storage, if someday someone decide to use the blob storage API somewhere else.

components/cloud/aws/src/s3.rs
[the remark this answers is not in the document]
yiwu-arbug: okay. cli is okay-ish but github has far better UI for side-by-side comparison and I can comment on it directly. 

Multiple commits would be especially helpful in this case though. They might not be useful other times.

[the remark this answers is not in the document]
yiwu-arbug: I briefly check the diff of s3.rs and gcs.rs and don't think I have further comment.

yiwu-arbug (commented):
Looking good to me. I'll approve after https://github.com/tikv/tikv/pull/9706#discussion_r607378285 is addressed.

components/cloud/aws/src/kms.rs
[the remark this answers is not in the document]
gregwebs: One already has "KMS". Should the other have "backup"?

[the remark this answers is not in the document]
yiwu-arbug: maybe `ENCRYPTION_AWS_VENDOR_NAME` and `BR_AWS_VENDOR_NAME`, and inline comment their usage?

components/cloud/src/blob.rs
gregwebs: Okay, I renamed to get/put

components/cloud/aws/src/kms.rs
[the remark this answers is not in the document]
gregwebs: okay, done in https://github.com/tikv/tikv/pull/9706/commits/28cf7b7b9e813fc23c7a621ce2e7d6820cbb75c6

yiwu-arbug (approved):

components/external_storage/export/src/dylib.rs
```diff
@@ -0,0 +1,308 @@
...
+
+#[cfg(feature = "prost-codec")]
+pub use kvproto::backup::storage_backend::Backend;
+#[cfg(feature = "protobuf-codec")]
+pub use kvproto::backup::StorageBackend_oneof_backend as Backend;
+
+struct ExternalStorageClient {
+    backend: Backend,
+    runtime: Arc<Runtime>,
+    blob_storage: Arc<Box<dyn ExternalStorage>>,
```
lcwangchao, line 28: why should we keep a blob_storage in side this client? It seems the only use of the blob storage is to build the file path of shared data

components/external_storage/export/src/dylib.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+fn call_ffi_dynamic(
+    library: &libloading::Library,
+    fn_name: &[u8],
+    bytes: Vec<u8>,
+) -> io::Result<()> {
+    let mut e = ffi_support::ExternError::default();
+    unsafe {
+        let func: FfiFn = library.get(fn_name).map_err(libloading_err_to_io)?;
```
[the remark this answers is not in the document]
gregwebs: I suspect the performance is not noticeable. Currently caching wouldn't do anything because a new ExternalStorageClient is created for each SST file so the functions are only called once.

components/external_storage/export/src/dylib.rs
gregwebs: yes, it is only used for metadata. I agree it should be removed. This code is tested to work but I still consider it experimental. We can spend more time cleaning things up if we commit to using this approach. Properly organizing the build for the cdylib is the first things I would look at cleaning up.

gregwebs: I went ahead and removed the blob_storage reference since it was an easy change to make.

components/external_storage/export/src/export.rs
```diff
@@ -0,0 +1,363 @@
...
+    )
+}
+
+fn bad_backend(backend: Backend) -> io::Error {
+    let storage_backend = StorageBackend { backend: Some(backend), ..Default::default() };
+    bad_storage_backend(&storage_backend)
+}
+
+#[cfg(any(feature = "cloud-gcp", feature = "cloud-aws"))]
+fn blob_store(store: Box<dyn BlobStorage>) -> Box<dyn ExternalStorage> {
```
kennytm, line 77: i'd suggest turning `BlobStore` to a generic type to avoid the doubling boxing.

```rust
struct BlobStore<B>(B);
impl<B: BlobStorage> ExternalStorage for BlobStore<B> { ... }
```

the `create_backend_inner` can use

```rust
Box::new(BlobStore(S3Storage::from_input(config.clone())?)))
```

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,199 @@
...
+            prefix: StringNonEmpty::opt(bucket.prefix),
+            storage_class: StringNonEmpty::opt(bucket.storage_class),
+            region: StringNonEmpty::opt(bucket.region),
+        })
+    }
+}
+
+pub fn none_to_empty(opt: Option<StringNonEmpty>) -> String {
+    if let Some(s) = opt {
+        (*s).clone()
```
kennytm, line 174:
```suggestion
        s.0
```

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    pub fn required_msg(s: String, msg: &str) -> io::Result<Self> {
+        if !s.is_empty() {
+            Ok(Self(s))
+        } else {
+            Err(Self::error_required(&format!("Empty {}", msg)))
+        }
+    }
+
+    pub fn error_required(msg: &str) -> io::Error {
```
kennytm, line 89:
```suggestion
    fn error_required(msg: &str) -> io::Error {
```

components/cloud/src/blob.rs
```diff
@@ -0,0 +1,199 @@
...
+        io::Error::new(io::ErrorKind::InvalidInput, msg)
+    }
+
+    pub fn required(s: String) -> io::Result<Self> {
+        Self::required_msg(s, "string")
+    }
+
+    pub fn stat(s: &'static str) -> Self {
```
kennytm, line 97: i'd rather you call it `static_` or `static_str` or `r#static`.

this name sounds like `stat(2)`.

components/external_storage/export/src/export.rs
gregwebs: okay. I have found it really easy to use Box but I will try to start using generics where possible.

Cargo.toml
```diff
@@ -221,6 +229,9 @@ tokio = { version = "0.2", features = ["macros", "rt-threaded", "time"] }
 zipf = "6.1.0"
 example_plugin = { path = "components/test_coprocessor_plugin/example_plugin" }
 
+[patch.'https://github.com/pingcap/kvproto']
+kvproto = { rev = "6212d77cfd9f1ea1ddc2139be0cda23895d84956", git = "https://github.com/gregwebs/kvproto.git", default-features = false }
```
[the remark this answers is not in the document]
gregwebs: This is on the checklist at the top of the PR.

components/cloud/aws/src/s3.rs
```diff
@@ -0,0 +1,598 @@
...
+        if let Some(access_key_pair) = &config.access_key_pair {
+            let cred_provider = StaticProvider::new_minimal(
+                (*access_key_pair.access_key).to_owned(),
+                (*access_key_pair.secret_access_key).to_owned(),
+            );
+            Self::new_creds_dispatcher(config, dispatcher, cred_provider)
+        } else {
+            let cred_provider = util::CredentialsProvider::new()?;
+            Self::new_creds_dispatcher(config, dispatcher, cred_provider)
+        }
```
kennytm, line 193: i don't know why we need that TODO comment. static credentials provided by the client is needed for non-cloud clusters.

kennytm (commented):
rest LGTM. but i can't approve before that kvproto PR is merged.

components/external_storage/src/grpc_client.rs
```diff
@@ -0,0 +1,132 @@
...
+        runtime: Arc::new(runtime),
+        rpc: new_rpc_client()?,
+        name,
+        url,
+    }))
+}
+
+fn new_rpc_client() -> io::Result<proto::ExternalStorageClient> {
+    let env = Arc::new(grpcio::EnvBuilder::new().build());
+    let grpc_socket_path = "/tmp/grpc-external-storage.sock";
```
overvenus, line 51: What about putting it into tikv's `data-dir` to prevent being deleted unexpectedly? 

Have you considered security for the approach?  Does it require mTLS?

components/external_storage/export/src/grpc_service.rs
```diff
@@ -0,0 +1,130 @@
...
+    (|| -> anyhow::Result<SocketService> {
+        let env = Arc::new(grpcio::EnvBuilder::new().build());
+        let storage_service = Service::new().context("new storage service")?;
+        let builder = grpcio::ServerBuilder::new(env)
+            .register_service(proto::create_external_storage(storage_service));
+        let grpc_socket_path = "/tmp/grpc-external-storage.sock";
+        let socket_addr = format!("unix:{}", grpc_socket_path);
+        let socket_path = std::path::PathBuf::from(grpc_socket_path);
+        // Keep the listener in scope: otherwise the socket is destroyed
+        let listener = bind_socket(&socket_path).context("GRPC new service create socket")?;
```
overvenus, line 30: Is it in-process RPC?

gregwebs: No. For in process one would use the C dynamic lib.

components/external_storage/src/grpc_client.rs
gregwebs: yes, data-dir might be a a better place. This would operate on the same node, so TLS is less of a concern. More important for security would be securing access to the socket.

The GRPC API is experimental and I would prefer to remove it in the future and make the case for the C dynamic lib instead. So I would prefer not to spend more time on it now.

components/external_storage/export/src/grpc_service.rs
gregwebs: The idea behind RPC is that the server is ran as a separate process: thus the need for the socket.

PR #9707, opened 2021-02-24
copr: add duration_duration_time_diff

mashenjun:
Signed-off-by: Shenjun Ma <mashenjun0902@gmail.com>

Issue Number: part of #5751

Problem Summary:

What's Changed:
- Modify the value of `MAX_NANOS` to not include `MAX_NANOS_PART`. According to [Mysql Reference](https://dev.mysql.com/doc/refman/8.0/en/time.html), the range of `TIME` tpye is '-838:59:59.000000' to '838:59:59.000000' which doesn't contain nano seconds. (Please correct me if it is deliberately to do so.)
- implement `fn duration_duration_time_diff`. The logic is similar to `fn sub_duration_and_duartion` but with extra steps to process overflow and underflow.
- add more corner case in unit tests.

### Related changes

- No related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 
implement `duration_duration_time_diff`


components/tidb_query_expr/src/impl_time.rs
```diff
@@ -700,6 +700,38 @@ pub fn add_duration_and_string(
...
+    arg2: &Duration,
+) -> Result<Option<Duration>> {
+    let res = match arg1.checked_sub(*arg2) {
+        Some(res) => res,
+        // `check_sub` returns `None` if the sub operation overflow/underflow i64 bound or mysql_time_value bound.
+        // and we need to treat these two case separately.
+        // if `arg1 - arg2` is in [`MAX_NANOS`, `i64::MAX`], return max value of mysql `TIME` type.
+        // if `arg1 - arg2` is in [`i64::MIN`, `-MAX_NANOS`], return min value of mysql `TIME` type.
+        // if `arg1 - arg2` is overflow or underflow i64, return `None`.
+        None if !arg1.is_neg() && i64::MAX - arg1.to_nanos() > -arg2.to_nanos() => {
```
Xuanwo, line 717: Maybe `arg1.to_nanos() - arg2.to_nanos() < i64::MAX` is more readable?

components/tidb_query_expr/src/impl_time.rs
```diff
@@ -700,6 +700,38 @@ pub fn add_duration_and_string(
...
+pub fn duration_duration_time_diff(
+    ctx: &mut EvalContext,
+    arg1: &Duration,
+    arg2: &Duration,
+) -> Result<Option<Duration>> {
+    let res = match arg1.checked_sub(*arg2) {
+        Some(res) => res,
+        // `check_sub` returns `None` if the sub operation overflow/underflow i64 bound or mysql_time_value bound.
+        // and we need to treat these two case separately.
+        // if `arg1 - arg2` is in [`MAX_NANOS`, `i64::MAX`], return max value of mysql `TIME` type.
```
Xuanwo, line 714: Should be (`MAX_NANOS`, `i64::MAX`) here?

mashenjun: You are right. The `if` statement and comment is not exactly correct. When reaching the `None` case, sub result can be neither `MAX_NANOS ` nor `-MAX_NANOS`. And furthermore, `i64::MAX` and `i64::MIN` should be considered as valid. So the range should be (`MAX_NANOS`, `i64::MAX`] and [`i64::MAX`,`-MAX_NANOS`).

components/tidb_query_expr/src/impl_time.rs
mashenjun: Reaching the `None` case means sub operation `arg1.to_nanos() - arg2.to_nanos()` may cause overflow/underflow. Thus, I transform the inequality to make sure the lhs is valid. Perhaps `arg1.to_nanos() - i64::MAX <= arg2.to_nanos()` is a better option here?

Xuanwo: Yes, the second style is more readable for me.

Xuanwo (approved):

Xuanwo:
PTAL @skyzh

components/tidb_query_expr/src/impl_time.rs
```diff
@@ -2084,4 +2116,59 @@ mod tests {
...
+                Some("00:00:00.000000"),
+            ),
+            // overflow or underflow case
+            (
+                Some("-00:00:01"),
+                Some("838:59:59.000000"),
+                Some("-838:59:59.000000"),
+            ),
+            (
+                Some("838:59:59.000000"),
```
andylokandy, line 2156: Please add one more test for `"838:59:59.000000" - "-838:59:59.000000"`

mashenjun: Done.

andylokandy:
@mashenjun Thank you and congratulation to your first PR to TiKV!

PR #9708, opened 2021-02-24
scripts: add build

gregwebs:
Example usage

```
./scripts/build --bin tikv-server
```

### Check List

- Manual test as per example usage

### Release note

None

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+# This script runs cargo build with a normal feature set
+# Arguments given will be passed through to "cargo build"
+# This runs in the Makefile environment via "make run"
+
+set -euo pipefail
+
+# Run from the Makefile environment
+MAKEFILE_RUN=${MAKEFILE_RUN:-""}
+if [[ -z $MAKEFILE_RUN ]] ; then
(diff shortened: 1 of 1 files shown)


BusyJay:
Are we going to add all subcommands of cargo as scripts? 😆

gregwebs:
The example given here is something I want to run all the time so it is something that should be well supported. I think TiKV devs probably don't bother specifying the bin because it is not supported so it doesn't seem worth it to save a second of build time.

One can just use `./scripts/env`. The main reason these additional scripts end up being needed is that [features cannot be set in an env variable](https://github.com/rust-lang/cargo/issues/4829) or that otherwise there are special settings defined in the Makefile or scripts rather than in cargo itself or some sort of data format.

The solution to avoiding lots of individual scripts is to put small commands in one place. This was originally done with a Makefile. But a Makefile is not designed for this. It doesn't support adding on cli args and therefore is not extensible. The solution I use for this issue that is language agnostic is [just](https://github.com/casey/just) which I can recommend from experience. I would be happy to add a Justfile to the project.

Alternatively there are always language specific solutions. In that case the main option I have seen is [cargo-make](https://github.com/sagiegurari/cargo-make). It seems like it would work but I haven't tried it previously.

PR #9709, opened 2021-02-25
file_system: cleanup eBPF on server shutdown instead of on thread exit

yiwu-arbug:

Previously we count the threads using biosnoop and cleanup the eBPF when the count drops to 0. This fails to take into account the main thread may also need to access biosnoop and the thread `IdxWrapper` does not get `drop` after program exit. This patch change to let TiKV server cleanup eBPF on shutdown. This change also fixes the broken biosnoop unit test.

It is still a problem if TiKV gets killed and the eBPF won't be cleaned up, though. 

Problem Summary:

What's Changed: This patch change to let TiKV server cleanup the eBPF for biosnoop on shutdown.

### Related changes

No

### Check List 

Tests 

build and run test in environment with bcc support and see the test pass. https://gist.github.com/yiwu-arbug/35f52102defe25a7257f1a5082b54e0f

### Release note 

- No release note

### DIFF
--- a/components/file_system/src/iosnoop/biosnoop.rs
+++ b/components/file_system/src/iosnoop/biosnoop.rs
@@ -4,9 +4,9 @@ use crate::metrics::*;
 use crate::IOBytes;
 use crate::IOType;
 
+use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::ptr;
-use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Mutex;
 
 use bcc::{table::Table, Kprobe, BPF};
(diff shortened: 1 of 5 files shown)


components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -160,6 +157,8 @@ pub(crate) fn fetch_io_bytes(mut io_type: IOType) -> IOBytes {
     IOBytes::default()
 }
 
+// For each process, it is supposed to call init_io_snooper() and stop_io_snooper_if_needed() once,
```
tabokie, line 160:
```suggestion
// For each process, it is supposed to call init_io_snooper() and stop_io_snooper() once,
```

components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -119,25 +100,41 @@ impl IdxAllocator {
...
+                    std::slice::from_raw_parts_mut(
+                        ptr as *mut u8,
+                        std::mem::size_of::<*const IOType>(),
+                    ),
+                )
+                .unwrap();
+            TYPE_TABLE_INITIALIZED.with(|initialized| {
+                initialized.replace(true);
+            });
+        }
```
tabokie, line 137: Put these and `s/initialized/registered` into `IdxWrapper`?

PR #9710, opened 2021-02-25
copr: fix cast expr when cast empty string to float (#9704)

### LINKED ISSUE #9704: Inconsisitent with Mysql When a varchar column only have empty string
master

### Steps to reproduce
drop table t;
create table t (a varchar(30));
insert into t values ('');
delete from t where a=10;

### What did you expect?
Query OK, 0 rows affected (0.01 sec) 

### What did happened?
ERROR 1265 (01000): Data Truncated

PS: It's cast expression'bug. It was fixed on TiDB [here](https://github.com/pingcap/tidb/issues/10806), but we still need to fix it in coprocessor.

guo-shaoge:

Issue Number: close #9704 

Problem Summary: 

    drop table t;
    create table t (a varchar(30));
    insert into t values ('');
    delete from t where a=10;

Delete statement should be ok. In MySQL, the result of empty stringcasting to float is zero. But in TiDB, it gives "Data Truncated" errors. Because cast expr got bug when cast empty string to float.

That was fixed in TiDB [here](https://github.com/pingcap/tidb/issues/10806), but we still need to fix it in copr. 

What's Changed:  Change function cast_string_as_signed_real(). When input string is empty, result will be 0.0

### Related changes

### Check List 

Tests 

- Unit test

Side effects: no side effects.

### Release note 

- fix cast expr when cast empty string to float in copr.

### DIFF
--- a/components/tidb_query_expr/src/impl_cast.rs
+++ b/components/tidb_query_expr/src/impl_cast.rs
@@ -519,7 +519,7 @@ fn cast_string_as_signed_real(
         None => Ok(None),
         Some(val) => {
             let r: f64;
-            if val.len() == 0 {
+            if val.is_empty() {
                 r = 0.0;
             } else {
                 r = val.convert(ctx)?;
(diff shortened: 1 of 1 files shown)


skyzh (approved):
LGTM. What about `cast_string_as_unsigned_real`?

Xuanwo (approved):

Issue #6639, opened 2020-02-18
Cannot compile TiKV on ARM64


n0vad3v:
### What version of TiKV are you using?

v4.0.0-beta

### What operating system and CPU are you using?

CentOS 7.6 with arm64.

### Steps to reproduce

`ROCKSDB_SYS_SSE=0 make dist_release`

### What did you expect?
The successful compile of TiKV.

### What did happened?
```
error[E0433]: failed to resolve: could not find `arch` in `self`
  --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/raw-cpuid-7.0.3/src/lib.rs:50:37
   |
50 |         let result = unsafe { self::arch::__cpuid_count(a, c) };
   |                                     ^^^^ could not find `arch` in `self`
   Compiling error-chain v0.12.1
   Compiling publicsuffix v1.5.3
error: aborting due to previous error
For more information about this error, try `rustc --explain E0433`.
error: could not compile `raw-cpuid`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

It seems that `raw-cpuid` does not support arm.

kennytm:
Thanks for the report.

The build failure is caused by lovesegfault/cache-size#3. We'll workaround it by making `cache-size` (and thus `raw-cpuid`) an x86-only dependency.

brson:
This bug is still present, with a slightly different version of raw-cpuid:

```
error[E0433]: failed to resolve: could not find `arch` in `self`
  --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/raw-cpuid-8.1.2/src/lib.rs:50:37
   |
50 |         let result = unsafe { self::arch::__cpuid_count(a, c) };
   |                                     ^^^^ could not find `arch` in `self`

   Compiling const-random-macro v0.1.6
error: aborting due to previous error
```

I don't see cache-size in Cargo.lock, but I do see raw-cpuid being used by the x86 crate, which is used by perfcnt, which is used by criterion-perf-events.

brson:
I put in a PR that should fix the build.

PR #9717, opened 2021-02-26
benches: Only use perfcnt on x86

brson:
Fixes ARM build.

Issue Number: close https://github.com/tikv/tikv/issues/6639

Problem Summary:

Build is currently broken due to use of the perfcnt crate (and criterion-perf-events), which does not build on arm. These crates are used in a single benchmark.

What's Changed:

Only use these crates on x86.

### Release note 

No release note.


tests/benches/coprocessor_executors/mod.rs
```diff
@@ -26,7 +26,7 @@ fn execute<M: criterion::measurement::Measurement + 'static>(c: &mut criterion::
     c.final_summary();
 }
 
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
```
kennytm, line 29: Please also update the `not` part on line 55

kennytm (commented):
Rest LGTM

tests/benches/coprocessor_executors/mod.rs
brson: Done. Good catch.

PR #9721, opened 2021-02-26
raftstore: re-propose read index with no lock checking response

### LINKED ISSUE #9713: Async snapshot may take long time if replica read is used with hibernate region

v5.0-rc release

Using the i3.4xl on aws, a tikv-server on a single machine.

### Steps to reproduce

Simple read/write requests like:
- Read. Table range scan using prefix clustered index.
- Write. Simple update serveral keys per transaction.

Enable the hibernate region feature(which is default enabled on master or v5.0-rc).
Using the follower-read in the client for each session.

### What did you expect?
The latency and qps are expected.

### What did happened?
Sometimes latency are very high, especially for the async-snapshot duration, while other metrics look good, there are just a few regions and load is not high.
image

@BusyJay has helped to analyze this phenomenon.
I believe it’s related to follower read and hibernate regions. According to the logs, one of the timeout requests is trying to request a peer of region 3848 (you can find such logs by searching error-response and looking around the context). The peer happens to be leader just before the error log. And it’s the first time it becomes leader, so the timeout request has to be accepted when it’s still follower. So TiDB should be using follower/replica read. A replica can execute read when when it has applied to the index returned by leader. I notice that hibernate region is enabled in configuration file, which can prevent leader from broadcasting its commit logs, and possibly lead to this problem. There is only a few regions in the cluster, I suggest to disable hibernate regions instead. And TiDB may also need to investigate if it’s expected to send a follower read request to a TiKV instance. As far as I known, only tiflash should accept follower read and it will keep polling unfinished requests.

- Make these two features work well together.
- Record the wait duration applying to the expected index.

sticnarf:

Closes #9713 

If some read index response that needs to check memory locks is missing, all pending reads will be blocked.

When a read index request is ready, but its lock checking is not done, we know the read index response must be lost. So we resend the same read index request to try again.

### Related changes

### Check List 

Tests 

- Integration test

### Release note 

- Fix the bug that replica reads will be blocked forever if a read index response is lost


5kbpers (approved):
LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -1974,12 +1975,26 @@ where
...
+                let (mut req, cb, _) = read.cmds.pop().unwrap();
+                assert_eq!(req.requests.len(), 1);
+                req.requests[0].set_read_index(*read_index);
+                info!(
+                    "re-propose read index request because the response is lost";
+                    "region_id" => self.region_id,
+                    "peer_id" => self.peer.get_id(),
+                );
+                RAFT_READ_INDEX_PENDING_COUNT.sub(1);
+                self.propose(ctx, cb, req);
```
NingLin-P, line 1996: After introducing async ready from https://github.com/tikv/tikv/pull/9177, we can not propose commands while there is a pending `Ready` not `advance_append` yet, maybe we should send a `PeerMsg::RaftCommand` instead. /cc @gengliqi

youjiali1995 (commented):
/LGTM

youjiali1995:
@hicqu PTAL, thanks!

components/raftstore/src/store/peer.rs
5kbpers: IMO `propose` here only proposes a read index request which doesn't depend on the current state of raft (it will check the epoch version and other states before handling the read request later).
Not sure if there is any case that breaks the correctness.

NingLin-P: Although I can't remember the sepecified case, but we come to a conclusion that `raft::step` should not be called when there is a pending `Ready` not `advance_append` yet, PTAL @gengliqi, also propose read index request may cause propose an extra empty command is some cases:
https://github.com/tikv/tikv/blob/47e03ae7906e7c32208ea01521416bae6c66ad30/components/raftstore/src/store/peer.rs#L2694-L2707

gengliqi: I choose to copy the unstable entries to `Ready` in async ready so maybe it's ok to propose `read_index` when there is a pending `Ready`.
However, there is another problem here. `post_pending_read_index_on_replica` may be called after getting the `Ready` which means the msg can not be sent until the next `Ready`.
In the previous logic, the `read_index` requests retry in `on_raft_base_tick`. It's ok because the `Ready` will be got later.

sticnarf:
@gengliqi Could you please take a look whether this works?

components/raftstore/src/store/peer.rs
```diff
@@ -1976,10 +1980,22 @@ where
     /// Responses to the ready read index request on the replica, the replica is not a leader.
     fn post_pending_read_index_on_replica<T>(&mut self, ctx: &mut PollContext<EK, ER, T>) {
         while let Some(mut read) = self.pending_reads.pop_front() {
-            // addition_request indicates an ongoing lock checking. We must wait until lock checking finished.
-            if read.addition_request.is_some() {
-                self.pending_reads.push_front(read);
-                break;
+            // The response of this read index request is lost, but we need it for
+            // the memory lock checking result. Resend the request.
+            if let Some(read_index) = read.addition_request.take() {
```
youjiali1995, line 1985: Can it result in too many retry requests?

sticnarf: In my internal tests, this happens very rarely in practice.
@gengliqi Do you have any concern about it? Could there be some case in theory that many requests are retried?

gengliqi: I believe many retry requests happen only when the network is repeatedly disconnected and connected.
So I think it's ok.

gengliqi (approved):
LGTM

PR #9722, opened 2021-02-27
concurrency manager: fix a memory leak when iterating skiplist

sticnarf:

When checking memory locks of a range, memory leaks happen.

It's due to a bug in crossbeam-skiplist. I fix it in [my own branch](https://github.com/sticnarf/crossbeam/commit/312812c95386bd3d949bb08ebee7b8950ce4aa38) and this PR will fix it.

Fixed in upstream: https://github.com/crossbeam-rs/crossbeam/pull/673

### Related changes

### Check List 

Tests 

Run long-running tests to see if memory leaks still exist manually.

### Release note 

- Fix memory leaks when checking a range for memory locks.

### DIFF
--- a/components/concurrency_manager/Cargo.toml
+++ b/components/concurrency_manager/Cargo.toml
@@ -19,9 +19,10 @@ tokio = { version = "0.2", features = ["macros", "sync", "time"] }
 txn_types = { path = "../txn_types", default-features = false }
 fail = "0.4"
 
+# FIXME: switch to the crates.io version after crossbeam-skiplist is released
 [dependencies.crossbeam-skiplist]
-git = "https://github.com/crossbeam-rs/crossbeam.git"
-tag = "crossbeam-0.7.3"
+git = "https://github.com/sticnarf/crossbeam.git"
+rev = "a58551085ce0cfbb07f28c6c0a3071c2d7d4c878"
 package = "crossbeam-skiplist" 
 
 [dev-dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -942,19 +942,6 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.8.0"
-source = "git+https://github.com/crossbeam-rs/crossbeam.git?tag=crossbeam-0.7.3#28ad2b7e015832b47db7e389dd9ebce3e94b3adb"
-dependencies = [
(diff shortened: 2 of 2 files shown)


youjiali1995 (approved):
/LGTM. Have you opened a PR to the upstream?

sticnarf:
@youjiali1995 There appears to be other more minor leaks and I am looking for the reason.

components/concurrency_manager/Cargo.toml
```diff
@@ -20,8 +20,8 @@ txn_types = { path = "../txn_types", default-features = false }
 fail = "0.4"
 
 [dependencies.crossbeam-skiplist]
-git = "https://github.com/crossbeam-rs/crossbeam.git"
-tag = "crossbeam-0.7.3"
+git = "https://github.com/sticnarf/crossbeam.git"
+branch = "tikv"
```
nrc, line 24: I think we should use a ref rather than a branch to ensure we're all building the same version

nrc (approved):
lgtm, obviously we should move back to mainline asap.

components/concurrency_manager/Cargo.toml
```diff
@@ -19,9 +19,10 @@ tokio = { version = "0.2", features = ["macros", "sync", "time"] }
 txn_types = { path = "../txn_types", default-features = false }
 fail = "0.4"
 
+# FIXME: switch to the crates.io version after crossbeam-skiplist is released
 [dependencies.crossbeam-skiplist]
-git = "https://github.com/crossbeam-rs/crossbeam.git"
-tag = "crossbeam-0.7.3"
+git = "https://github.com/sticnarf/crossbeam.git"
```
sticnarf, line 24: Although the fix is already merged to the upstream, I use my own branch to make it possible to use the crates.io version of `crossbeam-utils` and `crossbeam-epoch`.

sticnarf:
@nrc @youjiali1995 Please take a second look

components/concurrency_manager/Cargo.toml
youjiali1995: just curious, what are the advantages of the crates.io version?

sticnarf: I just don't want another copy of these dependencies...

youjiali1995 (approved):
/lgtm

PR #9723, opened 2021-02-27
*: cherry pick some async commit related fixes to 5.0-rc

sticnarf:

Cherry picks https://github.com/tikv/tikv/pull/9721 and https://github.com/tikv/tikv/pull/9722 to 5.0-rc

### Check List 

Tests 

### Release note 

- Fix the bug that replica reads will be blocked forever if a read index response is lost.
- Fix memory leaks when checking a range for memory locks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -158,12 +158,6 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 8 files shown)


BusyJay:
5.0-rc is not maintained anymore.

Issue #8486, opened 2020-08-21
Too many rocksdb tombstone may slow batch-get


Connor1996:
In the case, there are a lot of deletes for a range on the transaction level. Once GCed, it leads to a lot of tombstone marks in rocksdb. It is invisible, but iterating on them does have the same overhead as iterating on visible keys.
So when batch-get on the GCed keys, it calls `writecf.seek(key)` for each key and needs to iterate over a lot of tombstones for every key internally which hurts performance greatly. Also, coprocessor may suffer the same issue.

Some possible solution:
- use `prefix_seek` and `prefix_same_as_start`
- set `max-skippable-internal-keys=1` for `writecf.seek(key + ts)`. If it encounters a tombstone, it is enough to ensure the key is deleted at this version(without collapse rollback). So it can abort the seek directly, instead of seeking the next visible key.
- sort the keys of batch-get, and`PointGetter` records the key of the last `writecf.seek()`. If the recorded_key > now_key then now_key must be deleted. So it won't seek the range with lots of tombstones again and again.
- Use rocksdb single delete to relieve(Can't solve it completely)

PR #8514, opened 2020-08-26
storage: Optimize seek write for many tombstones

Connor1996:
### What problem does this PR solve?

optimize seek write when there are too many tombstones.
Fix https://github.com/tikv/tikv/issues/8486

### What is changed and how it works?

Three optimizations:
- for batch-get and coprocessor(point-get), use prefix_seek and prefix_same_as_start to get seek every key for write cf, so it can skip tombstones of other keys.
- for prewrite, when calculating `has_data_in_range` use `max-skippable-internals-keys` to avoid seek too many tombstones.
- check key order before calling `near_seek` for `ScanMode::Mixed` which save one seek() and several prev() calls

#### batch-get before and after
image
#### prewrite before and after
截屏2020-08-27 上午11 38 23
(before has no tombstone statistics for seek, actually there are a lot) 

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test

### Release note 
- Optimize seek write for many tombstones

PR #9729, opened 2021-03-01
storage: Optimize seek write for many tombstones (#8514)

Connor1996:
Cherry-pick of #8514

optimize seek write when there are too many tombstones.
Fix https://github.com/tikv/tikv/issues/8486

Three optimizations:
- for batch-get and coprocessor(point-get), use prefix_seek and prefix_same_as_start to get seek every key for write cf, so it can skip tombstones of other keys.
- for prewrite, when calculating `has_data_in_range` use `max-skippable-internals-keys` to avoid seek too many tombstones.
- check key order before calling `near_seek` for `ScanMode::Mixed` which save one seek() and several prev() calls

#### batch-get before and after
image
#### prewrite before and after
截屏2020-08-27 上午11 38 23
(before has no tombstone statistics for seek, actually there are a lot) 

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
- Optimize seek write for many tombstones

### DIFF
--- a/src/storage/kv/mod.rs
+++ b/src/storage/kv/mod.rs
@@ -220,7 +220,7 @@ quick_error! {
     pub enum ErrorInner {
         Request(err: ErrorHeader) {
             from()
-            description("request to underhook engine failed")
+            description(err.get_message())
             display("{:?}", err)
         }
         Timeout(d: Duration) {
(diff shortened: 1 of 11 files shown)


src/storage/kv/mod.rs
```diff
@@ -220,7 +220,7 @@ quick_error! {
     pub enum ErrorInner {
         Request(err: ErrorHeader) {
             from()
-            description("request to underhook engine failed")
+            description(err.get_message())
```
tabokie, line 223:
```suggestion
            description(err.take_message())
```

youjiali1995 (approved):
/lgtm

PR #9731, opened 2021-03-01
raftstore: return more regions in EpochNotMatch

### LINKED ISSUE #9691: Epoch not match should include all regions within the requested ranges
## Feature Request

We add region informations in `EpochNotMatch` errors to make client retry easier without requesting PD again. But current implementations only include two regions. Batch split can split a region into more than 2 ranges easily, if TiKV can't report heartbeat to PD for some reasons, client with stale region informations can always fail as some ranges are still using the old epoch.

The solution is to include all regions within the requested ranges, so client can retry without relying on the informations on PD. It should also reduce requests jitters as client can retry quicker.

BusyJay:

Close #9691.

More regions so that client can make decisions without PD, which should
reduce region miss in high load.

Tests 

- Unit test
- Integration test

### Release note 
- set more regions in EpochNotMatch to reduce region miss

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1,7 +1,7 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
 use std::borrow::Cow;
-use std::collections::Bound::{Excluded, Included, Unbounded};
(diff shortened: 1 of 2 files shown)


NingLin-P:
LGTM

5kbpers (approved):
LGTM

PR #9735, opened 2021-03-02
raftstore: disable hibernate region except on master

BusyJay:

Problem Summary:

This PR adds a function to check if the binary is built on master
branch. Hibernate region is an experimental features that should
not be enabled on release branch, but it's enabled by accident on
5.0.0-rc. Checking branch name at build time can prevent such mistake
from happening again.

The side effect of the PR is binary built in PR will also have hibernate
regions disabled even it's checkout from master.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- disable hibernate region by default as it's still unstable

### DIFF
--- a/components/raftstore/src/store/config.rs
+++ b/components/raftstore/src/store/config.rs
@@ -246,7 +246,7 @@ impl Default for Config {
             apply_batch_system: BatchSystemConfig::default(),
             store_batch_system: BatchSystemConfig::default(),
             future_poll_size: 1,
-            hibernate_regions: true,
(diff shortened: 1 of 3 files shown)


NingLin-P:
> The side effect of the PR is binary built in PR will also have hibernate
regions disabled even it's checkout from master.

People can enable `hibernate-regions` by config file if they needed.

NingLin-P (approved):
LGTM

gengliqi (approved):

PR #9737, opened 2021-03-02
raftstore: consider tick count in lease check

### LINKED ISSUE #9728: Lease check should consider tick count
## Bug Report

To guarantee lease correctness, follower should not start elections or accept vote request when with in lease. Inside raft, lease is counted by tick count, which is not accurate, the error range is within 1 tick. For example, if a tick interval is 2s, election timeout is 5 ticks, lease is also 2s * 5 = 10s. A follower is scheduled to tick at time 3:00, and leader's message arrives at 2:59, which reset its tick count to 0, so the follower may start election at 3:08. But leader expects its lease last till 2:59 + 10s = 3:09.

The correct check should ensure lease duration should not larger than election timeout - 1 tick. The default value follows the rule but we need a safe check when validating custom configuration.

TszKitLo40:

Issue Number: close #9728  

Problem Summary:

What's Changed:
consider tick count in lease check
### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
 
### Release note 
- No release note


components/raftstore/src/store/config.rs
```diff
@@ -718,5 +728,11 @@ mod tests {
         cfg = Config::new();
         cfg.future_poll_size = 0;
         assert!(cfg.validate().is_err());
+
+        cfg = Config::new();
+        cfg.raft_base_tick_interval = ReadableDuration::secs(1);
+        cfg.raft_election_timeout_ticks = 9;
+        cfg.raft_store_max_leader_lease = ReadableDuration::secs(11);
```
BusyJay, line 735: The test case will fail at L327 instead of L336.

TszKitLo40: I have fix this, PTAL

BusyJay (approved):

components/raftstore/src/store/config.rs
```diff
@@ -331,6 +331,16 @@ impl Config {
             ));
         }
 
+        let tick = self.raft_base_tick_interval.as_millis() as u64;
+        if lease > election_timeout - tick {
+            return Err(box_err!(
+                "lease {} ms should not be larger than election timeout {} ms - 1 tick({} ms)",
```
gengliqi, line 337:
```suggestion
                "lease {} ms should not be greater than election timeout {} ms - 1 tick({} ms)",
```

TszKitLo40: I have fixed it.

gengliqi (approved):

hi-rustin:
@TszKitLo40 Some cases failed. PTAL~

TszKitLo40:
> @TszKitLo40 Some cases failed. PTAL~

Done.

tests/failpoints/cases/test_merge.rs
```diff
@@ -1108,7 +1108,7 @@ fn test_node_merge_crash_before_snapshot_then_catch_up_logs() {
     cluster.cfg.raft_store.raft_base_tick_interval = ReadableDuration::millis(10);
     cluster.cfg.raft_store.raft_election_timeout_ticks = 10;
     // election timeout must be greater than lease
-    cluster.cfg.raft_store.raft_store_max_leader_lease = ReadableDuration::millis(99);
+    cluster.cfg.raft_store.raft_store_max_leader_lease = ReadableDuration::millis(90);
```
TszKitLo40, line 1111: It seems that these lines are duplicated with L1212

BusyJay:
/merge

Thank you!

PR #9738, opened 2021-03-02
*: fix build on CI

BusyJay:

grpcio-sys is updated to a new version that can detect compiler changes
correctly, so it will recompile from outdated caches.

openssl-src is replaced with an old version, because 1.1.1g to 1.1.1j
makes the check on self signed keys stricter, which can affect current
users. For example, upgrading an existing cluster can disconnect the
updated nodes due to certs error. So this PR downgrades openssl to wait
for a good solution to the problem and not blocking 5.0.0.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- No release note.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1799,9 +1799,9 @@ dependencies = [
 
 [[package]]
 name = "grpcio-sys"
-version = "0.8.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91a3435fb3d2b1367fff27e8f7bdbad1856459bf1f0fc9bd66527dce569e1523"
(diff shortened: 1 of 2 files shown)


sticnarf:
Latest openssl includes security fixes. So we'd better finally find a way to bypass the stricter check.

BusyJay:
> ...we'd better finally find a way to bypass the stricter check.

I have tried it. There are several stricter checks introduced by different commits. I don't think get around the check is a good idea, a better way is check the certificate before updating and suggesting user to renew its certificates instead. However such checks need extra work and discuss, I prefer to leave it to next sprint.

sticnarf (approved):

Connor1996 (approved):

PR #9753, opened 2021-03-05
pd_client: support forwarding requests to PD followers

rleungx:

Problem Summary: This PR is going to solve the problem that the request cannot be sent to the PD leader when there is a network partition problem between TiKV and PD leader.

What's Changed: Use the original thread, which is used to check the PD member change, to detect if the PD leader can be reached. Once the error happens, we are going to re-establish the connection to the PD follower and send the requests. PD followers will forward these kinds of requests to the leader.

### Check List 

Tests 

- Unit test
- Manual test

### Release note 
- Support forwarding requests to PD followers

### DIFF
--- a/components/pd_client/src/client.rs
+++ b/components/pd_client/src/client.rs
@@ -10,8 +10,10 @@ use futures::executor::block_on;
 use futures::future::{self, BoxFuture, FutureExt, TryFutureExt};
 use futures::sink::SinkExt;
 use futures::stream::{StreamExt, TryStreamExt};
+use grpcio::{
+    CallOption, EnvBuilder, Environment, MetadataBuilder, Result as GrpcResult, WriteFlags,
+};
 
-use grpcio::{CallOption, EnvBuilder, Environment, Result as GrpcResult, WriteFlags};
 use kvproto::metapb;
 use kvproto::pdpb::{self, Member};
 use kvproto::replication_modepb::{RegionReplicationStatus, ReplicationStatus};
@@ -25,7 +27,9 @@ use yatp::task::future::TaskCell;
 use yatp::ThreadPool;
 
 use super::metrics::*;
-use super::util::{check_resp_header, sync_request, LeaderClient, PdConnector};
+use super::util::{
+    check_resp_header, get_random_member_addr, sync_request, validate_endpoints, Client, PdConnector,
+};
 use super::{Config, FeatureGate, PdFuture, UnixSecs};
 use super::{Error, PdClient, RegionInfo, RegionStat, Result, REQUEST_TIMEOUT};
 
@@ -34,7 +38,7 @@ const CLIENT_PREFIX: &str = "pd";
 
 pub struct RpcClient {
     cluster_id: u64,
-    leader_client: Arc<LeaderClient>,
+    pd_client: Arc<Client>,
     monitor: Arc<ThreadPool<TaskCell>>,
 }
 
@@ -77,8 +81,8 @@ impl RpcClient {
                 Ok((client, members)) => {
                     let rpc_client = RpcClient {
                         cluster_id: members.get_header().get_cluster_id(),
-                        leader_client: Arc::new(LeaderClient::new(
-                            env,
+                        pd_client: Arc::new(Client::new(
+                            Arc::clone(&env),
                             security_mgr,
                             client,
                             members,
@@ -88,7 +92,7 @@ impl RpcClient {
 
                     // spawn a background future to update PD information periodically
                     let duration = cfg.update_interval.0;
-                    let client = Arc::downgrade(&rpc_client.leader_client);
+                    let client = Arc::downgrade(&rpc_client.pd_client);
                     let update_loop = async move {
                         loop {
                             let ok = GLOBAL_TIMER_HANDLE
@@ -146,18 +150,23 @@ impl RpcClient {
 
     /// Gets the leader of PD.
     pub fn get_leader(&self) -> Member {
-        self.leader_client.get_leader()
+        self.pd_client.get_leader()
     }
 
     /// Re-establishes connection with PD leader in synchronized fashion.
     pub fn reconnect(&self) -> Result<()> {
-        block_on(self.leader_client.reconnect(true))
+        block_on(self.pd_client.reconnect(true))
     }
 
     /// Creates a new call option with default request timeout.
     #[inline]
-    fn call_option() -> CallOption {
-        CallOption::default().timeout(Duration::from_secs(REQUEST_TIMEOUT))
+    fn call_option(receiver_addr: &str) -> CallOption {
+        let mut builder = MetadataBuilder::with_capacity(1);
+        builder.add_str("receiver", receiver_addr).unwrap();
+        let metadata = builder.build();
+        CallOption::default()
+            .headers(metadata)
+            .timeout(Duration::from_secs(REQUEST_TIMEOUT))
     }
 
     /// Gets given key's Region and Region's leader from PD.
@@ -172,13 +181,14 @@ impl RpcClient {
         let mut req = pdpb::GetRegionRequest::default();
         req.set_header(self.header());
         req.set_region_key(key.to_vec());
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let executor = move |client: &LeaderClient, req: pdpb::GetRegionRequest| {
+        let executor = move |client: &Client, req: pdpb::GetRegionRequest| {
             let handler = client
                 .inner
                 .rl()
                 .client_stub
-                .get_region_async_opt(&req, Self::call_option())
+                .get_region_async_opt(&req, Self::call_option(&receiver_addr))
                 .unwrap_or_else(|e| {
                     panic!("fail to request PD {} err {:?}", "get_region_async_opt", e)
                 });
@@ -200,7 +210,7 @@ impl RpcClient {
             }) as PdFuture<_>
         };
 
-        self.leader_client
+        self.pd_client
             .request(req, executor, LEADER_CHANGE_RETRY)
             .execute()
     }
@@ -211,13 +221,14 @@ impl RpcClient {
         let mut req = pdpb::GetStoreRequest::default();
         req.set_header(self.header());
         req.set_store_id(store_id);
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let executor = move |client: &LeaderClient, req: pdpb::GetStoreRequest| {
+        let executor = move |client: &Client, req: pdpb::GetStoreRequest| {
             let handler = client
                 .inner
                 .rl()
                 .client_stub
-                .get_store_async_opt(&req, Self::call_option())
+                .get_store_async_opt(&req, Self::call_option(&receiver_addr))
                 .unwrap_or_else(|e| panic!("fail to request PD {} err {:?}", "get_store_async", e));
 
             Box::pin(async move {
@@ -235,9 +246,15 @@ impl RpcClient {
             }) as PdFuture<_>
         };
 
-        self.leader_client
+        self.pd_client
             .request(req, executor, LEADER_CHANGE_RETRY)
             .execute()
+    }
+
+    /// Gets the connection address of PD.
+    /// Only used for test purpose.
+    pub fn get_address(&self) -> String {
+        self.pd_client.get_address()
     }
 }
 
@@ -270,9 +287,10 @@ impl PdClient for RpcClient {
         req.set_header(self.header());
         req.set_store(stores);
         req.set_region(region);
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let mut resp = sync_request(&self.leader_client, LEADER_CHANGE_RETRY, |client| {
-            client.bootstrap_opt(&req, Self::call_option())
+        let mut resp = sync_request(&self.pd_client, LEADER_CHANGE_RETRY, |client| {
+            client.bootstrap_opt(&req, Self::call_option(&receiver_addr))
         })?;
         check_resp_header(resp.get_header())?;
         Ok(resp.replication_status.take())
@@ -285,9 +303,10 @@ impl PdClient for RpcClient {
 
         let mut req = pdpb::IsBootstrappedRequest::default();
         req.set_header(self.header());
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let resp = sync_request(&self.leader_client, LEADER_CHANGE_RETRY, |client| {
-            client.is_bootstrapped_opt(&req, Self::call_option())
+        let resp = sync_request(&self.pd_client, LEADER_CHANGE_RETRY, |client| {
+            client.is_bootstrapped_opt(&req, Self::call_option(&receiver_addr))
         })?;
         check_resp_header(resp.get_header())?;
 
@@ -301,9 +320,10 @@ impl PdClient for RpcClient {
 
         let mut req = pdpb::AllocIdRequest::default();
         req.set_header(self.header());
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let resp = sync_request(&self.leader_client, LEADER_CHANGE_RETRY, |client| {
-            client.alloc_id_opt(&req, Self::call_option())
+        let resp = sync_request(&self.pd_client, LEADER_CHANGE_RETRY, |client| {
+            client.alloc_id_opt(&req, Self::call_option(&receiver_addr))
         })?;
         check_resp_header(resp.get_header())?;
 
@@ -318,9 +338,10 @@ impl PdClient for RpcClient {
         let mut req = pdpb::PutStoreRequest::default();
         req.set_header(self.header());
         req.set_store(store);
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let mut resp = sync_request(&self.leader_client, LEADER_CHANGE_RETRY, |client| {
-            client.put_store_opt(&req, Self::call_option())
+        let mut resp = sync_request(&self.pd_client, LEADER_CHANGE_RETRY, |client| {
+            client.put_store_opt(&req, Self::call_option(&receiver_addr))
         })?;
         check_resp_header(resp.get_header())?;
 
@@ -335,9 +356,10 @@ impl PdClient for RpcClient {
         let mut req = pdpb::GetStoreRequest::default();
         req.set_header(self.header());
         req.set_store_id(store_id);
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let mut resp = sync_request(&self.leader_client, LEADER_CHANGE_RETRY, |client| {
-            client.get_store_opt(&req, Self::call_option())
+        let mut resp = sync_request(&self.pd_client, LEADER_CHANGE_RETRY, |client| {
+            client.get_store_opt(&req, Self::call_option(&receiver_addr))
         })?;
         check_resp_header(resp.get_header())?;
 
@@ -361,9 +383,10 @@ impl PdClient for RpcClient {
         let mut req = pdpb::GetAllStoresRequest::default();
         req.set_header(self.header());
         req.set_exclude_tombstone_stores(exclude_tombstone);
+        let receiver_addr = get_random_member_addr(&self.get_leader());
 
-        let mut resp = sync_request(&self.leader_client, LEADER_CHANGE_RETRY, |client| {
-            client.get_all_stores_opt(&req, Self::call_option())
+        let mut resp = sync_request(&self.pd_client, LEADER_CHANGE_RETRY, |client| {
+            client.get_all_stores_opt(&req, Self::call_option(&receiver_addr))
         })?;
         check_resp_header(resp.get_header())?;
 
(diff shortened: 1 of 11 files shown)


components/pd_client/src/client.rs
```diff
@@ -145,18 +149,23 @@ impl RpcClient {
...
+        block_on(self.pd_client.reconnect())
     }
 
     /// Creates a new call option with default request timeout.
     #[inline]
-    fn call_option() -> CallOption {
-        CallOption::default().timeout(Duration::from_secs(REQUEST_TIMEOUT))
+    fn call_option(receiver_addr: &str) -> CallOption {
+        let mut builder = MetadataBuilder::with_capacity(1);
+        builder.add_str("receiver", receiver_addr).unwrap();
```
BusyJay, line 164: Should be `pd-receiver` to distinguish between pd requests and tikv requests.

components/pd_client/src/util.rs
```diff
@@ -87,40 +88,102 @@ impl Stream for HeartbeatReceiver {
...
+        info!("change pd client stub"; "from"=> &inner.address, "to" => &address);
+        let members = if let Some(members) = members_resp {
+            members
+        } else {
+            inner.members.clone()
+        };
+        let leader = members.get_leader();
+        let urls = leader.get_client_urls();
+        let receiver_addr = if !urls.is_empty() {
+            urls[0].to_string()
```
BusyJay, line 153: So it always chooses the first one?

components/pd_client/src/util.rs
```diff
@@ -456,13 +493,40 @@ pub async fn try_connect_leader(
...
         let leader = resp.get_leader().clone();
         for ep in leader.get_client_urls() {
             if let Ok((client, _)) = connect(Arc::clone(&env), &security_mgr, ep.as_str()).await {
                 info!("connected to PD leader"; "endpoints" => ep);
-                return Ok((client, resp));
+                return Ok((client, resp, ep.to_string()));
+            }
+        }
+
+        // Otherwise try to connect the PD cluster follower.
```
BusyJay, line 506: How about retrying some times before falling back to follower?

components/pd_client/src/util.rs
rleungx: Nope, it's only for testing now.

components/pd_client/src/util.rs
rleungx: make sense

components/pd_client/src/client.rs
rleungx: I think we don't need to distinguish them. We can use the same way to handle the requests of PD/TiKV on the PD side.

components/pd_client/src/util.rs
```diff
@@ -87,40 +88,102 @@ impl Stream for HeartbeatReceiver {
...
+            members
+        } else {
+            inner.members.clone()
+        };
+        let leader = members.get_leader();
+        let urls = leader.get_client_urls();
+        let receiver_addr = if !urls.is_empty() {
+            urls[0].to_string()
+        } else {
+            "".to_string()
```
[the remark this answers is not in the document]
rleungx: If the `urls` is empty, I think we can fall back to the old mechanism whose metadata is empty.

overvenus:
Is this PR a breaking change, it breaks all BR tests

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/6001/pipeline/

nolouch (commented):
BTW, how the connection updated from follower to leader?

components/pd_client/src/util.rs
```diff
@@ -180,46 +240,51 @@ impl LeaderClient {
...
+        }
+        let leader = resp.get_leader();
+        let members = resp.get_members();
+        let (res, has_network_error) = connector.reconnect_leader(leader).await?;
+        match res {
+            Some((client, address)) => return Ok(Some((client, address, resp))),
+            None => {
+                if has_network_error {
+                    if let Ok(Some((client, address))) =
+                        connector.reconnect_followers(members, leader).await
```
nolouch, line 277: Do we need config the enable forward like tidb?

components/pd_client/src/util.rs
```diff
@@ -456,43 +536,84 @@ impl PdConnector {
...
+            match self.connect(ep.as_str()).await {
+                Ok((client, address, _)) => {
+                    info!("connected to PD member"; "endpoints" => ep);
+                    return Ok((Some((client, address)), false));
+                }
+                Err(Error::Grpc(RpcFailure(RpcStatus { status, details: _ }))) => {
+                    if status == RpcStatusCode::UNAVAILABLE
+                        || status == RpcStatusCode::DEADLINE_EXCEEDED
+                    {
+                        network_fail_num += 1;
```
nolouch, line 562: should we log the err to find details error?

components/pd_client/src/util.rs
```diff
@@ -456,43 +536,84 @@ impl PdConnector {
...
+                    }
+                }
+                Err(_) => {}
```
nolouch, line 565: log the error?

components/test_pd/src/server.rs
```diff
@@ -136,10 +135,7 @@ fn hijack_unary<F, R, C: PdMocker>(
             );
         }
         _ => {
-            let status = RpcStatus::new(
-                RpcStatusCode::UNIMPLEMENTED,
-                Some("Unimplemented".to_owned()),
-            );
+            let status = RpcStatus::new(RpcStatusCode::UNAVAILABLE, Some("Unavailable".to_owned()));
```
nolouch, line 138: why change this?

components/pd_client/src/util.rs
```diff
@@ -180,46 +240,51 @@ impl LeaderClient {
...
+    }
+
+    async fn reconnect_pd(
+        &self,
+        connector: PdConnector,
+        members_resp: GetMembersResponse,
+        force: bool,
+    ) -> Result<Option<StubTuple>> {
+        let resp = connector.load_members(&members_resp).await?;
+        if !force && resp == members_resp {
```
nolouch, line 266: we need to reconnect to the leader if the leader network is recovered.

components/test_pd/src/server.rs
rleungx: for testing purpose

tests/integrations/pd/test_rpc_client.rs
```diff
@@ -131,6 +131,26 @@ fn test_rpc_client() {
...
+    let server = MockServer::new(2);
+    let eps = server.bind_addrs();
+    let mut cfg = new_config(eps.clone());
+    cfg.enable_forwarding = true;
+    let mgr = Arc::new(SecurityManager::new(&SecurityConfig::default()).unwrap());
+    let client = RpcClient::new(&cfg, None, mgr).unwrap();
+    fail::cfg(connect_leader_fp, "return").unwrap();
+    // RECONNECT_INTERVAL_SEC is 1s.
+    thread::sleep(Duration::from_secs(1));
+    let _ = client.alloc_id();
```
BusyJay, line 146: Should check id is allocated successfully.

components/test_pd/src/server.rs
BusyJay: I don't think it's a good idea. How about constructing the error in the failpoint directly?

components/test_pd/src/lib.rs
```diff
@@ -4,6 +4,8 @@
 extern crate tikv_util;
 #[macro_use]
 extern crate slog_global;
+#[macro_use(fail_point)]
```
BusyJay, line 7: You can import the macro by using `use fail::fail_point;`.

tests/integrations/pd/test_rpc_client.rs
rleungx: I think it's not necessary to check the result

components/pd_client/src/config.rs
```diff
@@ -30,7 +30,7 @@ pub struct Config {
     pub retry_log_every: usize,
     /// The interval at which to update PD information.
     ///
-    /// Default is 10m.
+    /// Default is 30s.
```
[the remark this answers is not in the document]
rleungx: The minimum interval of member change is 5s, 10m is too long for this check

components/pd_client/src/util.rs
```diff
@@ -456,43 +545,87 @@ impl PdConnector {
...
                 }
             }
-            None => Err(box_err!(
-                "failed to connect to {:?}",
-                previous.get_members()
-            )),
         }
     }
+
+    pub async fn reconnect_followers(
```
Yisaer, line 614: We can also record a `TODO` to connect the PD which locates in the same zone first.

components/pd_client/src/util.rs
Yisaer: How about add a metrics counter for `network_fail_num`

components/pd_client/src/util.rs
```diff
@@ -87,38 +92,91 @@ impl Stream for HeartbeatReceiver {
...
+        let _ = prev_receiver.right().map(|t| t.wake());
+        inner.client_stub = client_stub;
+        let prev_addr = inner.address.clone();
+        inner.address = address.clone();
+        inner.members = members;
+        inner.last_update = Instant::now();
+        if let Some(ref on_reconnect) = inner.on_reconnect {
+            on_reconnect();
         }
+        info!("change pd client stub"; "from"=> &prev_addr, "to" => address);
```
Yisaer, line 175: How about add a metrics counter if change pd client successfully.

Yisaer (commented):
Rest LGTM

components/pd_client/src/client.rs
```diff
@@ -146,18 +151,30 @@ impl RpcClient {
...
     pub fn reconnect(&self) -> Result<()> {
-        block_on(self.leader_client.reconnect(true))
+        block_on(self.pd_client.reconnect(true))
     }
 
     /// Creates a new call option with default request timeout.
     #[inline]
-    fn call_option() -> CallOption {
-        CallOption::default().timeout(Duration::from_secs(REQUEST_TIMEOUT))
+    fn call_option(forwarded_host: &str) -> CallOption {
```
BusyJay, line 169: Why not define it as `fn call_option(client: &Client) -> CallOption`? So `get_forwarded_host` can be removed.

components/pd_client/src/util.rs
```diff
@@ -41,6 +44,8 @@ pub struct Inner {
     >,
     pub hb_receiver: Either<Option<ClientDuplexReceiver<RegionHeartbeatResponse>>, Waker>,
     pub client_stub: PdClientStub,
+    address: String,
```
BusyJay, line 47: It makes code more complicated. If it's only for testing purpose, I suggest to change to use other way.

components/pd_client/src/util.rs
```diff
@@ -87,40 +92,90 @@ impl Stream for HeartbeatReceiver {
...
+
+        // Try to cancel an unused heartbeat sender.
+        if let Either::Left(Some(ref mut r)) = inner.hb_sender {
+            r.cancel();
+        }
+        inner.hb_sender = Either::Left(Some(tx));
+        let prev_receiver = std::mem::replace(&mut inner.hb_receiver, Either::Left(Some(rx)));
+        let _ = prev_receiver.right().map(|t| t.wake());
+        inner.client_stub = client_stub;
+        let prev_addr = inner.address.clone();
```
BusyJay, line 164:
```
let prev_addr = mem::replace(&mut inner.address, address.clone());
```

components/pd_client/src/util.rs
```diff
@@ -338,7 +421,7 @@ where
     Err(err.unwrap_or_else(|| box_err!("fail to request")))
 }
 
-pub type StubPair = (PdClientStub, GetMembersResponse);
+pub type StubTuple = (PdClientStub, String, String, GetMembersResponse);
```
BusyJay, line 424: Please define a struct instead.

components/pd_client/src/util.rs
```diff
@@ -180,56 +243,77 @@ impl LeaderClient {
...
-            // Try to cancel an unused heartbeat sender.
-            if let Either::Left(Some(ref mut r)) = inner.hb_sender {
-                debug!("cancel region heartbeat sender");
-                r.cancel();
+        self.update_client(client, address, forwarded_host, Some(members));
+        info!("tring to update PD client done"; "spend" => ?start.elapsed());
+        Ok(())
+    }
+
+    async fn reconnect_pd(
```
BusyJay, line 262: Why change the original implementation?

components/pd_client/src/util.rs
```diff
@@ -456,42 +581,90 @@ impl PdConnector {
...
+                        retry_times -= 1;
+                        continue;
+                    }
+                    return Ok((None, has_network_err));
+                }
+            }
+        }
+    }
+
+    pub async fn reconnect_followers(
```
BusyJay, line 652: I don't see why it needs to be a standalone function. Why not just keep it in `reconnect_leader`? Separating the two functions is inaccurate extraction and make code duplicated. There is no such connecting follower thing, what PD client does is always try to connect leader, either by connecting directly or forwarding by a follower. Connecting a follower is never its goal.

components/pd_client/src/util.rs
```diff
@@ -417,12 +464,41 @@ impl PdConnector {
...
-            Ok(resp) => Ok((client, resp)),
+            Ok(resp) => {
+                let leader = resp.get_leader();
+                let client_urls = leader.get_client_urls();
+                for leader_url in client_urls {
+                    if addr == leader_url {
+                        return Ok((client, addr.to_string(), "".to_string(), resp));
+                    }
+                }
+                for leader_url in client_urls {
```
BusyJay, line 475: Seems put into `try_forward` is more accurate.

components/pd_client/src/client.rs
```diff
@@ -146,18 +151,25 @@ impl RpcClient {
...
-        block_on(self.leader_client.reconnect(true))
+        block_on(self.pd_client.reconnect(true))
     }
 
     /// Creates a new call option with default request timeout.
     #[inline]
-    fn call_option() -> CallOption {
-        CallOption::default().timeout(Duration::from_secs(REQUEST_TIMEOUT))
+    fn call_option(client: &Client) -> CallOption {
+        let mut builder = MetadataBuilder::with_capacity(1);
```
BusyJay, line 165: Better extracting to a standalone function for setting metadata and should skip if `forwarded_host` is empty.

components/pd_client/src/util.rs
```diff
@@ -87,38 +91,83 @@ impl Stream for HeartbeatReceiver {
...
+        let prev_receiver = std::mem::replace(&mut inner.hb_receiver, Either::Left(Some(rx)));
+        let _ = prev_receiver.right().map(|t| t.wake());
+        inner.client_stub = client_stub;
+        inner.forwarded_host = forwarded_host;
+        inner.members = members;
+        inner.last_update = Instant::now();
+        if let Some(ref on_reconnect) = inner.on_reconnect {
+            on_reconnect();
+        }
+        info!("change pd client stub");
```
BusyJay, line 166: How about also logging `forwarded_host`?

components/pd_client/src/util.rs
```diff
@@ -456,42 +510,151 @@ impl PdConnector {
...
+        &self,
+        members_resp: GetMembersResponse,
+        pre_forwarded_host: String,
+        force: bool,
+        enable_forwarding: bool,
+    ) -> Result<Option<StubTuple>> {
+        let resp = self.load_members(&members_resp).await?;
+        let leader = resp.get_leader();
+        let members = resp.get_members();
+        let (res, has_network_error) = self.reconnect_leader(leader).await?;
```
BusyJay, line 529: This is unnecessary if `pre_forwarded_host.is_empty() && !force`.

components/pd_client/src/util.rs
```diff
@@ -87,38 +91,83 @@ impl Stream for HeartbeatReceiver {
...
             feature_gate: FeatureGate::default(),
+            enable_forwarding,
+        }
+    }
+
+    pub fn update_client(
+        &self,
+        client_stub: PdClientStub,
+        forwarded_host: String,
+        members_resp: Option<GetMembersResponse>,
```
BusyJay, line 136: Why use `Option`?

components/pd_client/src/util.rs
```diff
@@ -456,42 +505,154 @@ impl PdConnector {
...
+            match self.connect(ep.as_str()).await {
+                Ok((client, resp)) => {
+                    info!("connected to PD member"; "endpoints" => ep);
+                    return Ok((Some((client, resp)), false));
+                }
+                Err(Error::Grpc(RpcFailure(RpcStatus { status, details: _ }))) => {
+                    if status == RpcStatusCode::UNAVAILABLE
+                        || status == RpcStatusCode::DEADLINE_EXCEEDED
+                    {
+                        error!("failed to connect to PD member due to the network"; "endpoints" => ep, "status_code" => ?status);
```
BusyJay, line 577: Better log the whole error instead of just code.

components/pd_client/src/util.rs
```diff
@@ -456,42 +505,154 @@ impl PdConnector {
...
+                        network_fail_num += 1;
+                    }
+                }
+                Err(Error::Grpc(e)) => {
+                    error!("failed to connect to PD member"; "endpoints" => ep, "error" => ?e)
+                }
+                Err(e) => {
+                    error!("failed to connect to PD leader through forwarding"; "endpoints" => ep, "error" => ?e);
```
BusyJay, line 585: How do you know it's by forwarding?

components/pd_client/src/util.rs
```diff
@@ -456,42 +505,154 @@ impl PdConnector {
...
-            Some(resp) => {
-                if !force && resp == *previous {
-                    return Ok(None);
+        // Try to connect the PD cluster leader.
+        loop {
+            let (res, has_network_err) = self.connect_member(leader).await?;
+            match res {
+                Some((client, _)) => return Ok((Some((client, "".to_string())), has_network_err)),
+                None => {
+                    if has_network_err && retry_times > 0 {
```
BusyJay, line 609: How about limiting by time instead of retry times?

components/pd_client/src/util.rs
```diff
@@ -456,42 +505,154 @@ impl PdConnector {
...
+        match res {
+            Some((client, forwarded_host)) => {
+                if force || pre_forwarded_host != forwarded_host || resp != members_resp {
+                    return Ok(Some((client, forwarded_host, resp)));
+                } else {
+                    return Ok(None);
+                }
+            }
+            None => {
+                if enable_forwarding && has_network_error {
```
BusyJay, line 541: This is also unnecessary if not `force`.

components/pd_client/src/util.rs
```diff
@@ -456,42 +507,159 @@ impl PdConnector {
...
+            return Ok(None);
+        }
+        let (res, has_network_error) = self.reconnect_leader(leader).await?;
+        // There are 3 kinds of situations we will return the new client:
+        // 1. the force is true which represents the client is newly created or the original connection has some problem
+        // 2. the previous forwared host is not equal to the one right now which represents the network partition problem to leader may be recovered
+        // 3. the member information of PD has been changed
+        match res {
+            Some((client, forwarded_host)) => {
+                if force || pre_forwarded_host != forwarded_host || resp != members_resp {
```
BusyJay, line 538: The check is redundant, it's always true.

components/pd_client/src/util.rs
```diff
@@ -456,42 +507,159 @@ impl PdConnector {
...
+                    return Ok(Some((client, forwarded_host, resp)));
+                } else {
+                    return Ok(None);
+                }
+            }
+            None => {
+                // If the force is false, we could have already forwarded the requests.
+                // We don't need to try forwarding again.
+                if !force {
+                    return Ok(None);
```
BusyJay, line 548: In the past, error is returned instead.

components/pd_client/src/util.rs
```diff
@@ -456,42 +507,159 @@ impl PdConnector {
...
+                }
+                if enable_forwarding && has_network_error {
+                    if let Ok(Some((client, forwarded_host))) =
+                        self.try_forward(members, leader).await
+                    {
+                        if force || pre_forwarded_host != forwarded_host || resp != members_resp {
```
BusyJay, line 554: Redundant check.

components/pd_client/src/util.rs
```diff
@@ -456,42 +507,159 @@ impl PdConnector {
...
     }
 
     pub async fn reconnect_leader(
         &self,
-        previous: &GetMembersResponse,
-        force: bool,
-    ) -> Result<Option<StubPair>> {
-        let resp = self.load_members(&previous).await;
+        leader: &Member,
+    ) -> Result<(Option<(PdClientStub, String)>, bool)> {
```
BusyJay, line 605: String seems always be empty.

components/pd_client/src/util.rs
rleungx: Yes, but for now, I think None is more reasonable?

BusyJay: Letting caller know there is a network error seems more reasonable given that `!force` is also expected to reconnecting leader.

BusyJay (commented):
Rest LGTM

components/pd_client/src/util.rs
```diff
@@ -180,56 +230,56 @@ impl LeaderClient {
...
-        }
-        info!("updating PD client done"; "spend" => ?start.elapsed());
+        self.update_client(client, forwarded_host, members);
+        info!("tring to update PD client done"; "spend" => ?start.elapsed());
         Ok(())
     }
 }
 
+// The call option here doesn't need the timeout setting which is used by stream.
+fn call_option(forwarded_host: &str) -> CallOption {
```
hicqu, line 260: The name is not descriptive. How about `pd_forward_call_option`?

BusyJay (approved):

components/pd_client/src/util.rs
rleungx: IMO, this option is not only for forwarding, and since it is in pd client, we don't need to emphasize it is the call option for pd.

hicqu (approved):

PR #9754, opened 2021-03-05
raftstore: fix find_propose_time may return older instant

### LINKED ISSUE #9748: The propose time may be smaller than the real one
master

doesn't matter

### Steps to reproduce
doesn't matter

### What did you expect?
right propose time

### What did happened?
https://github.com/tikv/tikv/blob/311515d67331dee4bf1f34ebeb6200bd6061f711/components/raftstore/src/store/peer.rs#L104-L107
If the `idx` is got from `back.binary_search_by_key(&key, map)`, the `self.queue[idx]` is not the one we want.
The underlying data structure of `VecDeque` is a ring buffer. So if it's the back slice, the `idx` should add the length of front slice.
As an alternative, we can use the `VecDeque's binary_search_by_key` function though it's not stable yet.

Luckily, this bug can not lead to serious consequences because it may only return the smaller `propose time`.
As a result, it may cause a smaller `commit log duration` and `maybe_renew_leader_lease` will use this smaller one to update lease which doesn't break correctness.

NingLin-P:

Issue Number: close #9748 

Problem Summary:

See https://github.com/tikv/tikv/issues/9748

What's Changed:

Use `VecDeque::binary_search_by_key` to find proposal

### Check List 

Tests 

- Unit test

### Release note 
- raftstore: fix find_propose_time may return older instant

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -3762,6 +3762,7 @@ pub trait AbstractPeer {
 mod tests {
     use super::*;
     use crate::store::msg::ExtCallback;
+    use crate::store::util::u64_to_timespec;
     use kvproto::raft_cmdpb;
     #[cfg(feature = "protobuf-codec")]
     use protobuf::ProtobufEnum;
@@ -3951,16 +3952,14 @@ mod tests {
     fn test_propose_queue_find_proposal() {
         let mut pq: ProposalQueue<engine_panic::PanicSnapshot> =
             ProposalQueue::new("tag".to_owned());
-        let t = monotonic_raw_now();
         let gen_term = |index: u64| (index / 10) + 1;
         let push_proposal = |pq: &mut ProposalQueue<_>, index: u64| {
-            let renew_lease_time = if index % 3 == 0 { None } else { Some(t) };
             pq.push(Proposal {
                 is_conf_change: false,
                 index,
                 term: gen_term(index),
                 cb: Callback::write(Box::new(|_| {})),
-                renew_lease_time,
+                renew_lease_time: Some(u64_to_timespec(index)),
                 must_pass_epoch_check: false,
             });
         };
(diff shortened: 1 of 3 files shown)


NingLin-P:
This PR also add test to cover the case, I had manually check that the test will fail in the master branch and pass in this PR.

components/raftstore/src/store/peer.rs
```diff
@@ -3965,10 +3963,24 @@ mod tests {
...
                 must_pass_epoch_check: false,
             });
+        };
+        for index in 1..=100 {
+            push_proposal(&mut pq, index);
         }
         let mut pre_remove = 0;
-        for remove_i in &[0, 1, 65, 98, 100] {
-            let remove_i = *remove_i;
+        for remove_i in 1..=100 {
```
BusyJay, line 3971: How long does the case take?

components/raftstore/src/store/peer.rs
```diff
@@ -3955,8 +3953,8 @@ mod tests {
             ProposalQueue::new("tag".to_owned());
         let t = monotonic_raw_now();
         let gen_term = |index: u64| (index / 10) + 1;
-        for index in 1..=100 {
-            let renew_lease_time = if index % 3 == 1 { None } else { Some(t) };
+        let push_proposal = |pq: &mut ProposalQueue<_>, index: u64| {
+            let renew_lease_time = if index % 3 == 0 { None } else { Some(t) };
```
gengliqi, line 3957: How about using different values?

components/raftstore/src/store/peer.rs
NingLin-P: There is no time-consuming operation, the test will finish before a human can notice.

components/raftstore/src/store/peer.rs
NingLin-P: I use `u64_to_timespec` to get a unique propose time for each proposal instead, PTAL

BusyJay (approved):

gengliqi (approved):

PR #9755, opened 2021-03-05
set titan blob file zstd options

ZhenhanGong:
Signed-off-by: Zhenhan Gong<zhenhan.gong@gmail.com>

set titan blob file zstd compression options in tikv config

### Related changes
- PR of tikv/rust-rocksdb#591.

### Release note
- No release note


etc/config-template.toml
```diff
@@ -705,6 +705,13 @@
 # default: lz4
 # blob-file-compression = "lz4"
 
+# blob file zstd compression options
+window_bits = -14
+level = 32767
+stragety = 0
```
ssubu, line 711: Typo here

PR #9756, opened 2021-03-06
engine_rocks: Add ttl properties

Connor1996:

part of https://github.com/tikv/tikv/pull/9595

Add TTL properties collector

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
- None


components/engine_traits/src/util.rs
```diff
@@ -21,3 +23,30 @@ pub fn check_key_in_range(
...
+    if len < number::U64_SIZE {
+        return Err(Error::Codec(codec::Error::ValueLength));
+    }
+    let mut ts = &value_with_ttl[len - number::U64_SIZE..];
+    Ok(number::decode_u64(&mut ts)?)
+}
+
+pub fn strip_expire_ts(value_with_ttl: &[u8]) -> &[u8] {
+    let len = value_with_ttl.len();
+    &value_with_ttl[..len - number::U64_SIZE]
```
Little-Wallace, line 42: Do we need to check the length and return a `Result`?

Little-Wallace (commented):
rest LGTM.

components/engine_traits/src/util.rs
Connor1996: It was used in `value() of iter` which returns &[u8] and it is ensured to be called after `get_expire_ts()`.

Little-Wallace: OK. But it is a dangerous  design...

components/engine_rocks/src/ttl_properties.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    fn finish(&mut self) -> HashMap<Vec<u8>, Vec<u8>> {
+        if self.prop.max_expire_ts == 0 && self.prop.min_expire_ts == 0 {
+            return HashMap::default();
+        }
+        RocksTTLProperties::encode(&self.prop).0
+    }
+}
+
+pub struct TTLPropertiesCollectorFactory {}
```
BusyJay, line 99: In rust, it should be named as `TtlPropertiesCollectorFactory`.

components/engine_rocks/src/ttl_properties.rs
```diff
@@ -0,0 +1,152 @@
...
+        if entry_type != DBEntryType::Put {
+            return;
+        }
+        // only consider data keys
+        if !key.starts_with(keys::DATA_PREFIX_KEY) {
+            return;
+        }
+
+        let expire_ts = get_expire_ts(&value).unwrap_or_else(|_| {
+            error!("unexpected ttl key:{:?}, value:{:?}", key, value);
```
BusyJay, line 76: This probably violates the log redaction requirements.

Connor1996: remove the log, and the error will be recorded by metrics in later PR

BusyJay (approved):

components/engine_rocks/src/ttl_properties.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    pub fn decode<T: DecodeProperties>(props: &T) -> Result<TtlProperties> {
+        let mut res = TtlProperties::default();
+        res.max_expire_ts = props.decode_u64(PROP_MAX_EXPIRE_TS)?;
+        res.min_expire_ts = props.decode_u64(PROP_MIN_EXPIRE_TS)?;
+        Ok(res)
+    }
+}
+
+impl TtlPropertiesExt for RocksEngine {
```
hicqu, line 34: Seems this logic is same for `MvccProperties` and `TtlProperties`. Just a question, is it possible to merge them like
```rust
trait PropertiesExt {
    type Properties = X;
    fn get_range_properties_cf(...) -> Reuslt<Vec<(String, Self::Properties)>>;
}
```

components/engine_rocks/src/ttl_properties.rs
```diff
@@ -0,0 +1,159 @@
...
+        }
+        // only consider data keys
+        if !key.starts_with(keys::DATA_PREFIX_KEY) {
+            return;
+        }
+
+        let expire_ts = match get_expire_ts(&value) {
+            Ok(ts) => ts,
+            Err(e) => {
+                error!("failed to get expire ts";
```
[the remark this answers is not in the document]
Connor1996: Not possible now, but we'll introduce keyspace to support that

components/engine_rocks/src/ttl_properties.rs
Connor1996: Seems yes. I think we can leave it for engine traits' future work.

PR #9758, opened 2021-03-06
storage: Add ttl snapshot and iterator

Connor1996:

part of https://github.com/tikv/tikv/pull/9595

Add TTL snapshot and iterator

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
- None


src/storage/raw/ttl.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    fn get_cf_opt(&self, opts: ReadOptions, cf: CfName, key: &Key) -> Result<Option<Value>> {
+        self.map_value(self.s.get_cf_opt(opts, cf, key))
+    }
+
+    fn iter(&self, iter_opt: IterOptions, mode: ScanMode) -> Result<Cursor<Self::Iter>> {
+        self.s
+            .iter(iter_opt, mode)
+            .map(|c| c.into_with(|i| TTLIterator::new(i, self.current_ts)))
```
BusyJay, line 110: Confusing implementation. Why not make `S: engine_traits::Snapshot`?

Connor1996: TTLSnapshot is used in the `storage` level, so what we can get from `async_snapshot()` is a `kv::Snapshot` instead of `engine_traits::Snapshot`.

BusyJay: Why not add `ttl_get` to storage type directly?

src/storage/raw/ttl.rs
```diff
@@ -0,0 +1,377 @@
...
+
+#[derive(Clone)]
+pub struct TTLSnapshot<S: Snapshot> {
+    s: S,
+    current_ts: u64,
+}
+
+impl<S: Snapshot> TTLSnapshot<S> {
+    fn map_value(&self, mut value_with_ttl: Result<Option<Value>>) -> Result<Option<Value>> {
+        if value_with_ttl.is_err() {
```
BusyJay, line 24: Why not
```
match value_with_ttl? {
    Some(v) => {
        // check expire
    }
    None => Ok(None)
}
```

Little-Wallace (approved):
LGTM

BusyJay (approved):

PR #9759, opened 2021-03-07
*: log less on transport failure

BusyJay:

Problem Summary:

Raft client logs error already, log them again in raftstore is
useless. Instead, we can report those failure to prometheus.

This PR also adds back the missing metrics during last refactor.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- polish connection failure logs and metrics


src/server/raft_client.rs
```diff
@@ -893,8 +905,28 @@ where
         !self.need_flush.is_empty()
```
NingLin-P, line 905: Should we also check the `full_stores` here?

src/server/raft_client.rs
```diff
@@ -893,8 +905,28 @@ where
...
+        if self.full_stores.is_empty() {
+            return;
+        }
+
+        for id in &self.full_stores {
+            if let Some(s) = self.cache.get_mut(id) {
+                s.full = false;
+            }
+            REPORT_FAILURE_MSG_COUNTER
+                .with_label_values(&["full", &id.0.to_string()])
```
NingLin-P, line 918: Maybe the metric should also include the `from_store` instead of just the `to_store`.

src/server/raft_client.rs
BusyJay: No, connections are notified before they are added to `full_stores`.

src/server/raft_client.rs
BusyJay: `from_store` is in metrics already behind the field `instance`.

src/server/raft_client.rs
NingLin-P: What I mean is to flush the `full_stores` metric.

components/raftstore/src/store/peer.rs
```diff
@@ -3447,7 +3447,7 @@ where
         send_msg.set_message(msg);
 
         if let Err(e) = trans.send(send_msg) {
-            warn!(
+            debug!(
```
gengliqi, line 3450: Does it mean there is no info-level log if the error is `full`?

BusyJay: Yes, but there will be metrics in the error panel.

NingLin-P: How about adding a metric to show how many and what type of message we have dropped? We can push data to the metric wherever we have dropped a message, for example, dropping a message of a peer that we can't find in the store.

hunterlxt: Add comments to explain why only the full error log is removed?

hunterlxt (approved):
rest LGTM

gengliqi (approved):
LGTM

hunterlxt (approved):
LGTM

NingLin-P (approved):

PR #9768, opened 2021-03-09
Add metrics for load base split

lhy1024:

Problem Summary:
Add and imporve metrics for load base split

What's Changed:

~~For `READ_QPS_TOPN`, in the past we only collected the qps of the top ten regions that exceeded the threshold, now we collect the qps of all regions that exceed the threshold, considering that there are some regions whose qps is no longer higher than the threshold after a period of time, so to avoid trailing, we set the qps of these regions to 0.~~

We also add `LOAD_BASE_SPLIT_EVENT` to confirm the status of load base split, and introduce four events: 

1. `prepare_to_split` means that a suitable split key has been selected and is ready to be sent to `ask_batch_split`. 

2. `no_fit_key` means that no suitable split key is selected

~~3. `size_or_keys_too_small` means that `approximate_size` or `approximate_keys` is smaller than the threshold value~~

4. `qps_fit` means that the qps meets the threshold

~~It is worth noting that the threshold judgment is moved to after `collect_read_stats` in order to capture the case of `size_or_keys_too_small`~~
image

### Check List 

Tests 

- Unit test

### Release note 
- No Release note

### DIFF
--- a/components/raftstore/src/store/metrics.rs
+++ b/components/raftstore/src/store/metrics.rs
@@ -479,6 +479,13 @@ lazy_static! {
         &["order"]
         ).unwrap();
 
+    pub static ref LOAD_BASE_SPLIT_EVENT: IntCounterVec =
+        register_int_counter_vec!(
+            "tikv_load_base_split_event",
+            "Load base split event.",
+            &["type"]
+        ).unwrap();
+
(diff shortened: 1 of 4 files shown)


lhy1024:
It can be merged after #9794

JmPotato (approved):

nolouch (approved):
LGTM

PR #9773, opened 2021-03-09
*: support proxy over tikv nodes

BusyJay:

Problem Summary:

If clients are isolated from some of the TiKV instances, they will be
out of service eventually as all connections are waiting for isolated
TiKV instances to reply. However if TiKV can still connect with each
others, we can let TiKV instances act as proxy server that redirect the
requests from client to the target server. This can reduce
unavailability in extreme cases.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- introduce proxy on TiKV nodes to reduce unavailability


src/storage/txn/commands/mod.rs
```diff
@@ -404,13 +404,18 @@ fn find_mvcc_infos_by_key<S: Snapshot>(
 ) -> Result<LockWritesVals> {
     let mut writes = vec![];
     let mut values = vec![];
+    debug!("finding lock");
```
hicqu, line 407: How about remove all new added debug log? Seems they are only for developer's debug.

src/server/proxy.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    fn connect(&self, addr: &str) -> Option<Client> {
+        let env = match self.env.upgrade() {
+            Some(e) => e,
+            None => return None,
+        };
+
+        let cb = ChannelBuilder::new(env)
+            .max_receive_message_len(1 << 30) // 1G.
```
hicqu, line 80: These 2 limitation are `-1` for normal services.

src/server/proxy.rs
```diff
@@ -53,56 +55,95 @@ struct Client {
...
+    ) -> Option<Client> {
+        if self.last_pos == self.pool.len() {
+            if self.last_pos < cfg.proxy_max_connections_per_address {
+                let env = match env.upgrade() {
+                    Some(e) => e,
+                    None => return None,
+                };
+                let cb = ChannelBuilder::new(env)
+                    .stream_initial_window_size(cfg.grpc_stream_initial_window_size.0 as i32)
+                    .max_receive_message_len(-1) // 1G.
```
hicqu, line 86: The comment seems stale.

hicqu (approved):

hunterlxt (approved):
LGTM

rleungx (approved):
LGTM

PR #9777, opened 2021-03-10
txn_types: Check overflow in timestamp computation

### LINKED ISSUE #9771: Check overflow in timestamp computation
## Development Task
[`TimesStamp`](https://github.com/tikv/tikv/blob/master/components/txn_types/src/timestamp.rs#L8) is an u64 integer. There are some places where TiKV increases or decreases it but TiKV doesn't check overflow. Overflow can result in an infinite loop. Even if it's not possible to happen in theory, it's better to check it.

TszKitLo40:

Issue Number: close #9771 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:
Check overflow in timestamp computation
### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No release note


components/txn_types/src/timestamp.rs
```diff
@@ -33,20 +33,24 @@ impl TimeStamp {
     }
 
     pub fn next(self) -> TimeStamp {
+        assert!(self.0 + 1 >= 0);
```
andylokandy, line 36: Maybe we need this:
```suggestion
        assert!(self.0 < u64::max());
```

sticnarf (approved):
LGTM

andylokandy (approved):

PR #9784, opened 2021-03-11
storage: Add ttl tombstone statistics

Connor1996:

Add ttl tombstone statistics so that we can know the skipped ttl kvs when scanning.

Add a thread-local `TTL_TOMBSTONE`, and record the diff in `Cursor` for every action. The statistics will be flushed into scan detail metrics finnally.

### Related changes

### Check List 

Tests 

- None

### Release note 
- None


src/storage/kv/stats.rs
```diff
@@ -44,7 +46,7 @@ impl CfStatistics {
         self.get + self.next + self.prev + self.seek + self.seek_for_prev
     }
 
-    pub fn details(&self) -> [(&'static str, usize); 11] {
+    pub fn details(&self) -> [(&'static str, usize); 12] {
```
hicqu, line 49: How about making `12` as a constant?

Connor1996: good idea

hicqu (approved):

components/tikv_kv/src/stats.rs
```diff
@@ -16,6 +19,63 @@ const STAT_NEXT_TOMBSTONE: &str = "next_tombstone";
...
+            kind,
+            internal_tombstone: PerfContext::get().internal_delete_skipped_count() as usize,
+            ttl_tombstone: TTL_TOMBSTONE.with(|m| *m.borrow()),
+        }
+    }
+}
+
+impl Drop for StatsCollector<'_> {
+    fn drop(&mut self) {
+        self.stats.ttl_tombstone += TTL_TOMBSTONE.with(|m| *m.borrow()) - self.ttl_tombstone;
```
BusyJay, line 56: I'm afraid it's wrong. The future can be yield and fetched into other thread.

Connor1996: In the span of StatsCollector lifetime, we just call rocksdb iter API. We don't call a `.await` in there, so the future would not be yield.

BusyJay (approved):

PR #9788, opened 2021-03-11
pd_client: reconnect leader only when members change

### LINKED ISSUE #9690: pd connection is reconnecting every 10 minutes
## Bug Report

#7028 adds a new behavior that reconnecting PD leader every 10 minutes. The motivation is avoiding unavailability during pd member changes without restarting TiKV. However, the implementation reconnecting PD unconditionally, which can drop all pd region heartbeats every 10 minutes. If hibernate region is enabled, it can make scheduler ops from PD working very slowly.

PD doesn't need to be reconnected if memberships are the same.

Perhaps related to #9681.

BusyJay:

Issue Number: close #9690 

What's Changed:

Reconnecting leader only when members change to avoid disturbing heartbeats report unexpectedly.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- fix frequently reconnecting pd client


rleungx (approved):

overvenus (commented):
LGTM

coocood:
Can we cherry-pick to release-4.0?

BusyJay:
@coocood I'm afraid it's not a critical issue that will be landed at 4.0.

coocood:
Some users worried about the error log.

PR #9793, opened 2021-03-11
Add end key as sample key in key range for load base split

lhy1024:

Add end key as sample key in key range for load base split;
Ensure `sample` can collect enough keys.

### Check List 

Tests 

- Unit test

### Release note 
- No Release note

### DIFF
--- a/components/raftstore/src/store/worker/split_controller.rs
+++ b/components/raftstore/src/store/worker/split_controller.rs
@@ -18,7 +18,7 @@ use crate::store::worker::split_config::DEFAULT_SAMPLE_NUM;
 use crate::store::worker::{FlowStatistics, SplitConfig, SplitConfigManager};
 
 pub const TOP_N: usize = 10;
-pub const MAX_RETRY_TIME: i32 = 10000;
+pub const MAX_RETRY_TIME: i32 = 1000;
 
 pub struct SplitInfo {
     pub region_id: u64,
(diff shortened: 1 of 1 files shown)


components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -17,6 +18,7 @@ use crate::store::worker::split_config::DEFAULT_SAMPLE_NUM;
 use crate::store::worker::{FlowStatistics, SplitConfig, SplitConfigManager};
 
 pub const TOP_N: usize = 10;
+pub const MAX_RETRY_TIME: i32 = 10000;
```
nolouch, line 21: it's too large?

lhy1024: Usually , it is possible to sample up to 20 in the qps range, and this may indeed be somewhat large, how about 200？

nolouch (approved):
LGTM

lhy1024:
bench mark(all key ranges are [a,b])

in master
running 1 test
test store::worker::split_controller::tests::hub_flush                        ... bench:  31,432,192 ns/iter (+/- 6,459,350)

in this pr
running 1 test
test store::worker::split_controller::tests::hub_flush                        ... bench:  35,744,422 ns/iter (+/- 5,869,056)

nolouch:
PTAL @HunDunDM

BusyJay (approved):

HunDunDM (approved):

PR #9794, opened 2021-03-11
Add byte threshold and remove approximate threshold for load base split

lhy1024:

Add byte threshold and remove approximate threshold for load base split

Add byte threshold, because it will cover more case, such as there is a region with 1k qps and 30MB/s byte rate

Remove approximate threshold, because it is inaccurate

### Check List 

Tests 

- Unit test
- Manual test
```
set config tikv split.qps-threshold=50000;
drop database load_base_split;
create database load_base_split;
```
```
sysbench --rand-type=uniform --table_size=100000 --range_size=5000 --point_selects=0 --sum_ranges=0 --order_ranges=0 --distinct_ranges=0 --config-file=config oltp_read_only prepare
sysbench --rand-type=uniform --table_size=100000 --range_size=5000 --point_selects=0 --sum_ranges=0 --order_ranges=0 --distinct_ranges=0 --config-file=config oltp_read_only run
```

image
image

### Release note 
- No Release note

### DIFF
--- a/components/raftstore/src/store/worker/split_config.rs
+++ b/components/raftstore/src/store/worker/split_config.rs
@@ -5,16 +5,15 @@ use std::sync::Arc;
 use tikv_util::config::VersionTrack;
 
 const DEFAULT_DETECT_TIMES: u64 = 10;
-const DEFAULT_SAMPLE_THRESHOLD: i32 = 100;
+const DEFAULT_SAMPLE_THRESHOLD: u64 = 100;
 pub(crate) const DEFAULT_SAMPLE_NUM: usize = 20;
 const DEFAULT_QPS_THRESHOLD: usize = 3000;
+const DEFAULT_BYTE_THRESHOLD: usize = 30 * 1024 * 1024;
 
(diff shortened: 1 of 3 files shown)


nolouch (approved):
lgtm

HunDunDM (approved):

components/raftstore/src/store/worker/split_config.rs
```diff
@@ -25,9 +24,8 @@ pub struct SplitConfig {
     pub split_contained_score: f64,
     pub detect_times: u64,
     pub sample_num: usize,
-    pub sample_threshold: i32,
-    pub size_threshold: u64,
-    pub key_threshold: u64,
+    pub sample_threshold: u64,
+    pub byte_threshold: usize,
```
cosven, line 28: I think this change breaks backward compatibility.

If a user set the config `size_threshold` in their config files and it does not take effects after they upgrade to the newer version.

cosven: I created an issue https://github.com/tikv/tikv/issues/9816

PR #9798, opened 2021-03-12
raftstore: add region pending peer duration metrics

crazycs520:
Signed-off-by: crazycs520 <crazycs520@gmail.com>

Add region pending peer duration metrics. 

If a region has pending-peer for a long time, we need a metric to trigger alertmanager rule. such as the following rule:

```
histogram_quantile(0.9999, sum(rate(tikv_raftstore_peer_pending_duration_seconds_bucket[1m])) BY (le, instance)) > 600
```

N/A

### Related changes

- N/A

### Check List 

Tests 

- No logical code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

N/A


crazycs520:
@iosmanthus @NingLin-P PTAL

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1027,6 +1027,13 @@ where
             self.fsm.peer.mut_store().flush_cache_metrics();
             return;
         }
+
+        if self.fsm.peer.is_leader() && self.fsm.peer.peers_start_pending_time.len() > 0 {
+            let (_, pending_after) = self.fsm.peer.peers_start_pending_time[0];
```
NingLin-P, line 1032: `self.fsm.peer.peers_start_pending_time[0]` may not be the longest pending time, because [`swap_remove`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove) will not preserve ordering: 
https://github.com/tikv/tikv/blob/9029a426169d06ecf131f2624caed54784f5501f/components/raftstore/src/store/peer.rs#L1280
Perhaps we should replace `swap_remove` with [`retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain)

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/fsm/peer.rs
crazycs520: Nice catch~
How about using an iterator to find the longest pending time? use `retain` is too hack?

NingLin-P (approved):
LGTM, thanks!

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1027,6 +1027,19 @@ where
             self.fsm.peer.mut_store().flush_cache_metrics();
             return;
         }
+
+        if self.fsm.peer.is_leader() && !self.fsm.peer.peers_start_pending_time.is_empty() {
```
BusyJay, line 1031: Isn't it reported to PD already? Can alert manager use the metrics from PD directly?

NingLin-P: Before this PR, `peers_start_pending_time` is only used for logging.

BusyJay: How about updating metrics right before logging? base tick is too frequent for raft tick unrelated work.

NingLin-P: `peers_start_pending_time` only log when a pending peer has caught up log, so if a pending peer does not catch up log yet it will not show up at the metric, and when it shows up at the metric it already not a pending peer.

To make the metric accurate, we need periodically report the `peers_start_pending_time`, if 1 tick is too frequent maybe 5 or more based tick?

crazycs520: The default value of `pd-heartbeat-tick-interval` is `1m`, use metrics from PD is not in time.

BusyJay: > if 1 tick is too frequent maybe 5 or more based tick?

Why not put it into a separate tick or put it into pd-heartbeat-tick? Why is it has to be bound to base tick? I have seen people set base-tick-interval to 100ms, what do you think is a suitable tick count?

> use metrics from PD is not in time.

Why `1m` is not in time? What could go wrong if it could only be handled after 1minute?

crazycs520: > Why 1m is not in time? What could go wrong if it could only be handled after 1minute?

I thought the `pd-heartbeat-tick` is `10m` at first, `10m` is not in time.

But `pd-heartbeat-tick` is `1m` is acceptable. I will move into pd-heartbeat-tick.

components/raftstore/src/store/peer.rs
```diff
@@ -1279,6 +1285,7 @@ where
                 if progress.matched >= truncated_idx {
                     let (_, pending_after) = self.peers_start_pending_time.swap_remove(i);
                     let elapsed = duration_to_sec(pending_after.elapsed());
+                    RAFT_PEER_PENDING_DURATION.observe(elapsed);
```
BusyJay, line 1288: I don't think it's necessary. Histogram will reset automatically after fetching.

crazycs520: Why? This is used to record the finish time. Without this line, We can't know the max pending duration accurately.

BusyJay (approved):
Have you check how it looks in grafana?

crazycs520:
> Have you check how it looks in grafana?

image

PR #9781, opened 2021-03-10
copr: add coprocessor v2

Jibbow:
### What problem does this PR solve?

Issue Number: #9747

Problem Summary:

Setup a module for the new coprocessor framework.
Mostly I want to get some feedback whether I missed something obvious and about the naming of modules/structs.

### What is changed and how it works?

Proposal: [RFC: coprocessor plugin](https://github.com/andylokandy/rfcs/blob/plugin/text/2021-02-24-coprocessor-plugin.md) 

What's Changed:

### Related changes

- PR to update `kvproto`: https://github.com/pingcap/kvproto/pull/728

### Check List 

Tests 

- No code

### Release note

 * No release note

Issue #9747, opened 2021-03-04
LFX: Coprocessor Plugin


Jibbow:
## Description

TiKV is the storage component in the TiDB ecosystem, however, the distribution computation principle suggests that computation should be as close to the data source as possible. Therefore, TiKV has embedded a subset of the TiDB executor framework to push down some computation tasks when applicable.

But TiKV's capability should be far beyond that, as many distributed components can be built on top of TiKV, such as cache, full-text search engine, graph database, and NoSQL database. And same as TiDB, this product will also like to push down specific computation to TiKV, which requires the coprocessor to be customizable, aka pluggable.

For instance, a full-text searching engine will persist the original document and n-gram index on TiKV. It'll be a waste of resources if we read back and then update the index from a client. On the contrary, the coprocessor plugin can generate the index from the original document, and update the index in place. What's more, the coprocessor plugin can perform an index scan directly on TiKV.

## Document Collection

- [RFC: Coprocessor Plugin](https://github.com/tikv/rfcs/pull/63)

## LFX Program information

- Mentor of this issue: @andylokandy @skyzh 
- Recommended skills: Rust
- Estimated Workloads: 3 Man-Month

## Milestones and action items

* [x] Add a new **config option** in [`TiKvConfig`](https://github.com/tikv/tikv/blob/master/src/config.rs#L2345) for the new coprocessor. The new config should specify (multiple) plugins that are loaded at startup. Also add new config to [config-template.toml](https://github.com/tikv/tikv/blob/master/etc/config-template.toml). Plugins are loaded from the local file system. Make sure to document the config option here: https://docs.pingcap.com/tidb/stable/tikv-configuration-file
* [x] Add a new **coprocessor server endpoint** in [`tikv::server::service::kv::Service`](https://github.com/tikv/tikv/blob/master/src/server/service/kv.rs#L51).  
The old coprocessor's module is located in [coprocessor](https://github.com/tikv/tikv/tree/master/src/coprocessor). The module of the new coprocessor will be called `coprocessor_v2`.
[...]

PR #9802, opened 2021-03-14
coprv2: plugin_api and plugin manager

Jibbow:

 * Adds a new crate that contains types and traits for implementing a custom coprocessor plugin. Also adds some documentation on how to use it.
 * Adds an example plugin for the new coprocessor_v2 that can be used for integration tests. The plugin currently does nothing, but it compiles to a proper cdylib. Also, this PR adds a function to resolve package names to library names.
 * Adds a `PluginRegistry` to the new `coprocessor_v2`. The `PluginRegistry` can load `dylibs` and keep them around in memory.

Issue Number: #9747

What's Changed:

Changes are mostly noted inline as TODO comments, but a list of changes compared to the RFC are:
 * the `RawStorage` trait had some functions in the RFC that took `&mut self`. Now, it only takes `&self`.
 * the `RawStorage` trait had a bound on `Send`, but we don't necessarily need it with `#[async_trait(?Send)]`.  
   *Note: I tried having it `RawStorage: Send`, but it turned out to be quite hard, because `Engine` and `LockManager` are both "only" `Send`. And because we have `&self` references, in `RawStorage` (like `async fn get(&self, ...)`, this means that we also internally have `&Engine` and `&LockManager`, which also needs to be `Send` because of `RawStorage: Send`. And this in turn means that `Engine` and `LockManager` need to be `Sync` (and not only `Send`). The only solution I found for this was to use `#[async_trait(?Send)]`.*
 * added `on_plugin_load()` and `on_plugin_unload()` functions
 * added two new types definitions: pub type:
   * `RawRequest = [u8];`
   * `pub type RawResponse = Vec<u8>;`
 to make parameters more expressive.

### Related changes

- this PR is based on #9781, but I'm not sure how to express this in GitHub.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

 * No release notes


components/coprv2_plugin_api/src/plugin_api.rs
```diff
@@ -0,0 +1,73 @@
...
+pub trait CoprocessorPlugin: Any + Send + Sync {
+    /// Returns the name of the plugin.
+    /// Requests that are sent to TiKV coprocessor must have a matching `copr_name` field.
+    fn name(&self) -> &'static str;
+
+    /// A callback fired immediately after the plugin is loaded. Usually used
+    /// for initialization.
+    ///
+    /// The default implementation does nothing, but can be overridden by plugins.
+    fn on_plugin_load(&self) {}
```
andylokandy, line 24: If the `$constructor` in `declare_plugin!` has done the initialization, what is the difference between them?

components/coprv2_plugin_api/src/plugin_api.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// A callback fired immediately before the plugin is unloaded. Use this if
+    /// you need to do any cleanup.
+    ///
+    /// The default implementation does nothing, but can be overridden by plugins.
+    fn on_plugin_unload(&self) {}
```
andylokandy, line 30: I think the plugin can impl `Drop`.

components/coprv2_plugin_api/src/plugin_api.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// Declare a plugin type and its constructor.
+///
+/// # Notes
+/// This works by automatically generating an `extern "C"` function with a
+/// pre-defined signature and symbol name. Therefore you will only be able to
+/// declare one plugin per library.
+#[macro_export]
+macro_rules! declare_plugin {
+    ($plugin_type:ty, $constructor:path) => {
```
andylokandy, line 62: Why `plugin_type` must be specified? Isn't it sufficient that it can coerce to `$crate::CoprocessorPlugin`?

Cargo.toml
```diff
@@ -273,6 +274,7 @@ members = [
   "components/concurrency_manager",
   "components/file_system",
   "components/collections",
+  "components/coprv2_plugin_api",
```
andylokandy, line 277: I prefer `coprocessor_plugin_api`, what do you think? @skyzh

components/coprv2_plugin_api/src/storage_api.rs
```diff
@@ -0,0 +1,79 @@
...
+#[derive(Debug, Clone)]
+pub struct RegionEpoch {
+    pub conf_ver: u64,
+    pub version: u64,
+}
+
+/// Errors when operating on [`RawStorage`].
+#[derive(Debug)]
+pub enum StorageError {
+    // TODO: this introduces a dependency to kvproto. Is this okay?
```
andylokandy, line 34: I think it's not ok... Just like `Region`, we have to make reflected types for all the types we need to expose.

components/coprv2_plugin_api/src/plugin_api.rs
Jibbow: hm good point, I had in mind that `$constructor` really just creates the object and `on_plugin_load()` does some (optional) initialization - so just to make a distinction here and maybe to make a clearer API (same for `on_plugin_unload()` and `Drop`). But you are right: it might be confusing because they are the same. I'll remove the functions and instead add a comment :)

components/coprv2_plugin_api/src/storage_api.rs
Jibbow: Okay got it, thanks for the clarification :)

components/coprv2_plugin_api/src/plugin_api.rs
Jibbow: True that. A constructor should be enough.

Jibbow: What do you think is nicer: 
 * passing a constructor to `declare_plugin!` 
 * passing a type that coerces to `$crate::CoprocessorPlugin` and adding a construction function to `CoprocessorPlugin` like `fn create() -> Self` so that there definitely exists a constructor.
 * passing a type that coerces to `$crate::CoprocessorPlugin` and require `trait CoprocessorPlugin: Default`.

components/coprv2_plugin_api/src/plugin_api.rs
```diff
@@ -0,0 +1,73 @@
...
+/// Raw bytes of the request payload from the client to the coprocessor.
+pub type RawRequest = [u8];
+/// The response from the coprocessor encoded as raw bytes that are sent back to the client.
+pub type RawResponse = Vec<u8>;
+
+/// A plugin that allows users to execute arbitrary code on TiKV nodes.
+///
+/// If you want to implement a custom coprocessor plugin for TiKV, your plugin needs to implement
+/// the [`CoprocessorPlugin`] trait.
+pub trait CoprocessorPlugin: Any + Send + Sync {
```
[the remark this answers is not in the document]
Jibbow: Right; removed `Any`

Cargo.toml
skyzh: Agree.

Jibbow: Alright, renamed to `coprocessor_plugin_api` :)

components/coprv2_plugin_api/src/plugin_api.rs
andylokandy: I like the second one. The third may be less obvious on how the plugin is initialized.

Jibbow: Alright, then we'll go with the second one :+1:

Jibbow: I had to change it to `fn create() -> Self where Self: Sized` because otherwise the trait would not be object-safe (https://doc.rust-lang.org/reference/items/traits.html#object-safety)
I hope that's okay.

components/coprv2_plugin_api/src/storage_api.rs
Jibbow: I removed the dependency to `kvproto` and just left `KeyNotInRegion` and `OtherError` in the enum. I guess we'll see what other errors we need when I implement the trait for `tikv::storage::Storage`.

Jibbow:
I think the main question now is how we deal with `jemalloc` @andylokandy @skyzh

Jibbow:
Changes since last time:
 * new we use `HostAllocator` in plugin. Thanks @andylokandy for the POC! I think I could not have done this without you <3
   I changed a few things:
    * use function pointers instead of `usize`.
    * instead of dealing with the `host` feature, the `#[global_allocator]` is set and managed in `declare_plugin!` macro.
 * moved `declare_plugin!` into `util.rs` and added the name of the plugin constructor as `PLUGIN_CONSTRUCTOR_NAME`.
 * whitelisted `coprocessor_plugin_api` in https://github.com/tikv/tikv/blob/master/scripts/check-bins.py

skyzh (approved):
LGTM, great work!

components/coprocessor_plugin_api/src/allocator.rs
```diff
@@ -0,0 +1,53 @@
...
+/// to plugin. It will be passed from TiKV to the plugin.
+#[repr(C)]
+pub struct HostAllocatorPtr {
+    pub alloc_fn: AllocFn,
+    pub dealloc_fn: DeallocFn,
+}
+
+/// An allocator that forwards invocations to the host (TiKV) of the plugin.
+pub struct HostAllocator {
+    alloc_fn: AtomicPtr<AllocFn>,
```
andylokandy, line 21: `AtomicPtr` add one more indirection than `AtomicUsize`, although I don't know whether they will make any actual difference.

components/coprocessor_plugin_api/src/util.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use super::allocator::HostAllocatorPtr;
+use super::plugin_api::CoprocessorPlugin;
+
+/// Name of the exported constructor function for the plugin in the `dylib`.
+pub const PLUGIN_CONSTRUCTOR_NAME: &[u8] = b"_plugin_create";
+/// Type signature of the exported constructor function for the plugin in the `dylib`.
+pub type PluginConstructorSignature =
+    unsafe fn(host_allocator: HostAllocatorPtr) -> *mut dyn CoprocessorPlugin;
```
andylokandy, line 10: If the return type is `Box<dyn CoprocessorPlugin>`, tikv will have a chance to call the `drop()` on plugin unload.

components/coprocessor_plugin_api/src/plugin_api.rs
```diff
@@ -0,0 +1,47 @@
...
+/// A plugin that allows users to execute arbitrary code on TiKV nodes.
+///
+/// If you want to implement a custom coprocessor plugin for TiKV, your plugin needs to implement
+/// the [`CoprocessorPlugin`] trait.
+///
+/// Plugins can run setup code in their constructor and teardown code by implementing
+/// [`std::ops::Drop`].
+pub trait CoprocessorPlugin: Send + Sync {
+    /// A constructor that initializes the plugin.
+    fn create() -> Self
```
andylokandy, line 19: I realized in poc that this function will make the `CoprocessorPlugin` not object-safe. So I moved the constructor to the `plugin_declare` macro.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -0,0 +1,132 @@
...
+        Ok(plugin_name)
+    }
+}
+
+/// A wrapper around a loaded raw coprocessor plugin library.
+struct LoadedPlugin {
+    /// Pointer to a [`CoprocessorPlugin`] in the loaded `lib`.
+    plugin: Box<dyn CoprocessorPlugin>,
+    /// Underlying library file on a fixed position on the heap.
+    lib: Pin<Box<Library>>,
```
andylokandy, line 58: Why the `Library` has to be pinned? Doesn't it allocate symbols in heap? (https://github.com/nagisa/rust_libloading/blob/7fbd3a34336f7839bdd6d6aaf83a9a763c4fa1e5/src/os/unix/mod.rs#L78)

components/coprocessor_plugin_api/src/allocator.rs
```diff
@@ -0,0 +1,53 @@
...
+    dealloc_fn: AtomicPtr<DeallocFn>,
+}
+
+impl HostAllocator {
+    /// Creates a new [`HostAllocator`], but initializes the functions to the current allocator.
+    /// Make sure to call [`set_allocator()`] because otherwise the default allocator will be used.
+    pub const fn new() -> Self {
+        HostAllocator {
+            alloc_fn: AtomicPtr::new(std::alloc::alloc as *mut _),
```
andylokandy, line 30: I'll prefer using null point as default, so it'll segment fault immediately on startup when the host allocator is not set, instead of popping up some mystery bugs after dealloc.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -0,0 +1,132 @@
...
+    ///
+    /// # Safety
+    ///
+    /// The library **must** contain a function with name [`PLUGIN_CONSTRUCTOR_NAME`] and the
+    /// signature of [`PluginConstructorSignature`]. Otherwise, behavior is undefined.
+    /// See also [`libloading::Library::get()`] for more information on what restrictions apply to
+    /// [`PLUGIN_CONSTRUCTOR_NAME`].
+    pub unsafe fn new(lib: Library) -> Result<Self, DylibError> {
+        let lib = Box::pin(lib);
+        let constructor: Symbol<PluginConstructorSignature> = lib.get(PLUGIN_CONSTRUCTOR_NAME)?;
```
andylokandy, line 77: Why the ABI is not checked?

src/coprocessor_v2/plugin_registry.rs
andylokandy: And in fact, I'll suggest to leak the Library on loading the plugin because we can't control the reference to the library once the `Future` created by `CoprocessorPlugin` is spawned into a runtime, and then it's be certainly unsafe to unmount the library.

components/coprocessor_plugin_api/src/allocator.rs
andylokandy: And in this case, when a plugin defined this as the global allocator, and for some reasons the allocator ptr is not set, any allocation will result in dead loop.

andylokandy:
Where the plugin version will be defined?

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        assert_eq!(plugin_name, "example-plugin");
+    }
+
+    #[test]
+    fn move_loaded_plugin() {
+        let lib = unsafe { Library::new(pkgname_to_libname("example-plugin")).unwrap() };
+        let loaded_plugin = unsafe { LoadedPlugin::new(lib).unwrap() };
+
+        let moved = move || loaded_plugin;
```
Jibbow, line 118: Is it save to assume that loaded_plugin will be moved? Or could the move still be elided?

src/coprocessor_v2/plugin_registry.rs
Jibbow: I wanted to keep the PR a bit smaller and wanted to add a version check later on.

src/coprocessor_v2/plugin_registry.rs
andylokandy: It'll be safe to move anyhow.

src/coprocessor_v2/plugin_registry.rs
andylokandy: Ah, I see.

components/coprocessor_plugin_api/src/plugin_api.rs
Jibbow: I thought it would be object-safe with the `where Self: Sized` constraint? But not sure about this

src/coprocessor_v2/plugin_registry.rs
Jibbow: Oh true that! I somehow assumed that the library has it's own memory space because I never worked with dylibs before. But makes sense, thank you for the clarification! :)

components/coprocessor_plugin_api/src/allocator.rs
Jibbow: yes, good point! Will change it :)

components/coprocessor_plugin_api/src/plugin_api.rs
andylokandy: I tried it in rust playground:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=16b3d1bd4be438b3faf8ee5012af900b

src/coprocessor_v2/plugin_registry.rs
Jibbow: nevermind, not important anymore because as you said the library doesn't need to be pinned

components/coprocessor_plugin_api/src/plugin_api.rs
Jibbow: But it works with the trait bound: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=887606baa1dcf271206a3d9d3a5f0fb4
Do I miss anything here? can't we require `Self: Sized`?

Jibbow: Oh btw: the trait bound `where Self: Sized` in line 20 which is not included in GitHub's preview :D

Jibbow: But maybe it's a good idea to separate the construction function because then we can `impl CoprocessorPlugin for LoadedPlugin`.

And then it _should_ be memory-safe because we can't `drop()` the library as long a a reference to `LoadedPlugin` is in use. And I _think_ this shouldn't be an issue anymore:
> In fact, I'll suggest leaking the Library on loading the plugin because we can't control the reference to the library once the Future created by CoprocessorPlugin is spawned into a runtime, and then it'll be certainly unsafe to unmount the library.

What do you think? Will this work?

andylokandy: `Self: Sized` means to strip that method out of dynamic object of a trait, so I think it'll be clearer if the constructor is individual from the plugin trait.

> because then we can impl CoprocessorPlugin for LoadedPlugin.

I'm not pretty sure what is this for.

Jibbow: Ah I see, then we'll go with a separate constructor! :)

>  > because then we can impl CoprocessorPlugin for LoadedPlugin.
>
> I'm not pretty sure what this is for.

I thought that maybe we could then `impl CoprocessorPlugin for LoadedPlugin`, similar to the `FunctionProxy` struct in https://adventures.michaelfbryan.com/posts/plugins-in-rust/

This would prevent accidentally dropping the `Library` while it's still in use and then I we don't need to leak it.

Jibbow: Updated the code to use a separate constructor for plugins

andylokandy: > I thought that maybe we could then impl CoprocessorPlugin for LoadedPlugin

Why does it prevent the `Future` returned by  `LoadedPlugin` from taking the reference to the static value of the library?

Jibbow: hm, I'm not sure. I thought that references to the static value of the library are okay as long as the library is not dropped. And I haven't worked with `Futures` that much, but I assumed that `LoadedPlugin` would not be dropped as long as there is some reference to it.

components/coprocessor_plugin_api/src/allocator.rs
andylokandy: AtomicPtr<T> store the pointer to T right? And we know T=fn, where fn is a pointer. So it is a ptr to ptr.

skyzh: Sounds correct.

components/coprocessor_plugin_api/src/plugin_api.rs
Jibbow: Update: we are `mem::forget`ting the library now. But I still don't quite understand why we _have_ to do it  :see_no_evil:

Jibbow:
Also added some nicer error types in `coprocessor_plugin_api`. A `CoprocessorPlugin` now returns a `PluginError`:
```rust
pub enum PluginError {
    StorageError(StorageError),
    Other(Box<dyn std::error::Error>),
}
```

components/coprocessor_plugin_api/src/allocator.rs
Jibbow: Different question: why do we need `AtomicPtr<AllocFn>` at all and not just `Option<AllocFn>`?

components/coprocessor_plugin_api/src/plugin_api.rs
andylokandy: I got segment fault in the POC where I dropped the `Library` after spawning the future. It turns out to be the static str reference has become dangling pointer...

Jibbow: Ah yes, makes sense! Thanks! :)

I thought that we can't drop the library as long as there are references to `LoadedPlugin` because we bundle it together with `Box<dyn CoprocessorPlugin>` so we can either drop both at the same time or none. How does this conflict with futures?
```rust
struct LoadedPlugin {
  plugin: Box<dyn CoprocessorPlugin>,
  lib: Library
}
```

components/coprocessor_plugin_api/src/util.rs
Jibbow: Is there anything we should do about this?

components/coprocessor_plugin_api/src/allocator.rs
andylokandy: > why do we need AtomicPtr<AllocFn> at all and not just Option<AllocFn>?

`Option` is not mutable under `&` reference, while `AtomicXXX` is.

andylokandy: and.... https://github.com/rust-lang/rust/issues/53639

components/coprocessor_plugin_api/src/plugin_api.rs
Jibbow: Ahh, I think I got it, thanks! Sorry for asking so many questions about it :see_no_evil:

components/coprocessor_plugin_api/src/allocator.rs
Jibbow: Can't we just use 
```rust
#[global_allocator]
static mut HOST_ALLOCATOR: HostAllocator = HostAllocator::new();
```
?

I just tried it out and it seems to work :)

Jibbow: Oh sorry, I didn't see the last reply haha

Jibbow: Thanks a lot for your patience :pray:

components/coprocessor_plugin_api/src/util.rs
andylokandy: We can add it later when implementing unloading plugin.

andylokandy (approved):

skyzh:
@andylokandy Seems that we need another `make format`?

andylokandy:
I may have done something stupid...

Issue #9678, opened 2021-02-20
Import macros from other crates with the use syntax


JmPotato:
## Development Task

There are many uses of `extern crate macro_module` in TiKV. However, we now have macro naming and modularisation according to rust-lang/rfcs#1561. Importing macros from other crates with the `use` syntax will be more clear for the codes. #9651 is an example of this change.

skyline75489:
Hi @JmPotato  I've filed several PR addressing this issue in different modules (See linked PR above). Can you take a look?

JmPotato:
> Hi @JmPotato I've filed several PR addressing this issue in different modules (See linked PR above). Can you take a look?

Thx for your contribution!

skyline75489:
Hi @JmPotato do you happen to know how can I pushing those PRs forward? Who should I ask for help? I gotta say it's bit discouraging to see almost no response for more than a week.

JmPotato:
> Hi @JmPotato do you happen to know how can I pushing those PRs forward? Who should I ask for help? I gotta say it's bit discouraging to see almost no response for more than a week.

Sorry for not respond in time! I will ask some reviewers to follow up.

skyline75489:
Hi @JmPotato I've filed another 2 PRs for this issue. Mind taking a look?

PR #9803, opened 2021-03-15
backup: import macros with the use syntax

skyline75489:
Signed-off-by: skyline75489 <skyline75489@outlook.com>

Issue Number: supports #9678 

Problem Summary:

Import macros from other crates with the `use` syntax.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

- No release note


components/backup/src/lib.rs
```diff
@@ -3,12 +3,8 @@
 #![recursion_limit = "200"]
 #![feature(box_patterns)]
 
-#[macro_use]
-extern crate failure;
 #[allow(unused_extern_crates)]
 extern crate tikv_alloc;
```
skyline75489, line 7: Is this really used? Not sure If I can just remove it.

JmPotato: You can try to remove it and build again to see whether it's necessary.

sticnarf:
I am not so sure about this change. `#[macro_use]` simplifies the uses of these macros if it doesn't cause any confusion. Wdyt? @nrc

nolouch (approved):

nrc (approved):
lgtm

nrc:
> I am not so sure about this change. #[macro_use] simplifies the uses of these macros if it doesn't cause any confusion. Wdyt? @nrc

I definitely prefer using `use` to import macros - it is more consistent, more flexible, and makes it easier to lookup the macro definition.

components/backup/src/lib.rs
JmPotato: Any update on this problem?

skyline75489: Well I found this https://github.com/tikv/tikv/blob/2a2fa03da53b63f3fc24d7ea53aead40176979b5/components/tikv_alloc/src/lib.rs#L10-L12

but I'm not sure this is still true

JmPotato: Well, I think we better leave it unchanged.

PR #9804, opened 2021-03-15
tikv_util: Clippy fix

### LINKED ISSUE #8336: Fix Clippy warnings
`make clippy` results in one Clippy warning (not sure how this made it past CI) in components/tidb_query_codegen, which seems easy to fix. 

When running `make clippy` we disable a lot of warnings, when running `cargo clippy` we get the default set of lints and we see a lot of Clippy warnings. Some of those really should be ignored, but some could be easily fixed. You can see even more things that could be fixed with `-Wrust-2018-idioms` (cc #4301)

It would be great to fix a bunch of those, and perhaps re-enable the lints if appropriate.

If you have any questions about whether a warning should be fixed or not, please ask me here or in Slack.

TszKitLo40:

Issue Number: #8336 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:
Clippy fix for `tikv_uitl`
### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No release note


components/tikv_util/src/metrics/threads_dummy.rs
```diff
@@ -36,3 +36,9 @@ impl ThreadInfoStatistics {
         HashMap::default()
     }
 }
+
+impl Default for ThreadInfoStatistics {
```
BusyJay, line 40: We allow clippy::new_without_default explictly.

TszKitLo40: OK, so we can remove all the Default method is this pr?

TszKitLo40: Are there any command to allow clippy::new_without_default in clippy?

BusyJay: It's at https://github.com/tikv/tikv/blob/master/scripts/clippy#L20

TszKitLo40: So all the lints in this script is allowed?

BusyJay: `unnecessary_wraps` and `needless_range_loop` can be removed, they are not checked probably due to too many changes need to be taken.

TszKitLo40: I think it may be better to fix the clippy errors by package. And I have fixed the errors in the `tikv_util` package.

BusyJay (approved):

youjiali1995 (approved):

PR #9806, opened 2021-03-15
rocksdb: Size ratio compaction: compact sst file by size ratio violation.

CheneyDing:

Problem Summary: Compact SST files by TiKV region size ratio violation for delete scenario.

What's Changed: This PR adds the following configs (with default value. only showing default CF, same for other CFs):
```
[rocksdb.defaultcf]
enable_size_ratio_compaction = false
```
A new compaction policy that compact SST files by the size ratio violation. The size ratio violation of each region is calculated by the following formula: 
```
size ratio violation of region i = T - ( size of region i in level j ) / ( size of region i in level j+1 )
```
and the size ratio violation of SST file can be computed by weighting the region size ratio violation of regions within the SST file according to the region size in it.

when compaction pick file, we choose sst file with maximum size ratio violation. In this way, the compaction can maintain the size ratio of region in each layer as much as possible, reducing the impact of deleting data on the structure of lsm-tree. Therefore, the amount of compaction data will also be reduced.

### Related changes

- [tikv/rust-rocksdb#616](https://github.com/tikv/rust-rocksdb/pull/616)
- [tikv/rocksdb#231](https://github.com/tikv/rocksdb/pull/231)

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note


tabokie:
Is there any comparing benchmark for this?

CheneyDing:
> Is there any comparing benchmark for this?
There are still bugs, relevant test results will be added later.

PR #9823, opened 2021-03-16
copr: fix IN expr didn't handle unsigned/signed int properly

### LINKED ISSUE #9821: coprocessor IN expr not handle max_u64 properly
    TiKV
    Release Version:   5.0.0-rc.x
    Edition:           Community
    Git Commit Hash:   eca434da4e71d540f7201203d10bf078d2ece853
    Git Commit Branch: master
    UTC Build Time:    2021-02-24 08:22:44
    Rust Version:      rustc 1.51.0-nightly (1d0d76f8d 2021-01-24)
    Profile:           debug

    AMD Ryzen 5 PRO 4650U with Radeon Graphics

### Steps to reproduce

Copr bug, use following SQL can reproduce [TiDB issue](https://github.com/pingcap/tidb/issues/23198)

    use test;
    drop table if exists t;
    create table t (a int);
    insert into t values(-1);
    select * from t where a in(18446744073709551615, 1);

### What did you expect?
empty set
### What did happened?
    select a from t where a in(18446744073709551615, 1);
    +-----+
    | a   |
    |-----|
    | -1  |
    +-----+

guo-shaoge:

Issue Number: close #9821

Problem Summary: IN expr didn't distinguish between unsigned int and signed int. So -1(bigint) equals 18446744073709551615(bigint unsigned), which is not compatible with MySQL. Here is [TiDB issue](https://github.com/pingcap/tidb/issues/23198).

What's Changed: Add a fn_meta for IN expr when first arg of IN expr is INT type. In this function, we use info which records arg's flag to tell whether two integer is really equal.

### Related changes

### Check List 

Tests 

- Unit test : impl_compare_in.rs: test_unsigned_signed_int()

Side effects:
- performance regression: To fix this issue, we need to check argument's signed/unsigned flag. Also we use HashMap instread HashSet. So maybe little performance regression.

### Release note 
- fix IN expr(coprocessor) didn't handle unsigned/signed int properly

### DIFF
--- a/components/tidb_query_expr/src/impl_compare_in.rs
+++ b/components/tidb_query_expr/src/impl_compare_in.rs
@@ -218,8 +218,11 @@ where
             match metadata.lookup_map.get(base_val) {
                 None => {}
                 Some(&argi_unsigned) => {
-                    if *base_val >= 0 || (arg0_unsigned && argi_unsigned) || (!arg0_unsigned && ! argi_unsigned) {
-                        return Ok(Some(1))
+                    if (*base_val >= 0)
+                        || (arg0_unsigned && argi_unsigned)
+                        || (!arg0_unsigned && !argi_unsigned)
+                    {
+                        return Ok(Some(1));
                     }
                 }
             }
@@ -232,7 +235,11 @@ where
                     }
                     Some(v) => {
                         let v = T::map_ref(v)?;
-                        if base_val == v && ((*base_val >= 0) || (arg0_unsigned && argi_unsigned) || (!arg0_unsigned && !argi_unsigned)) {
+                        if (base_val == v)
+                            && ((*base_val >= 0)
+                                || (arg0_unsigned && argi_unsigned)
+                                || (!arg0_unsigned && !argi_unsigned))
+                        {
                             return Ok(Some(1));
                         }
                     }
@@ -283,7 +290,11 @@ fn init_compare_in_data<T: InByHash>(expr: &mut Expr) -> Result<CompareInMeta<T:
     let children = expr.mut_children();
     assert!(!children.is_empty());
     let mut unsigned_flags = vec![false; children.len()];
-    unsigned_flags[0] = children[0].get_field_type().as_accessor().flag().contains(FieldTypeFlag::UNSIGNED);
+    unsigned_flags[0] = children[0]
+        .get_field_type()
+        .as_accessor()
+        .flag()
+        .contains(FieldTypeFlag::UNSIGNED);
 
     let n = children.len();
     let mut tail_index = n - 1;
@@ -291,7 +302,11 @@ fn init_compare_in_data<T: InByHash>(expr: &mut Expr) -> Result<CompareInMeta<T:
     for i in (1..n).rev() {
         let tree_node = &mut children[i];
         let mut is_constant = true;
-        let is_unsigned = tree_node.get_field_type().as_accessor().flag().contains(FieldTypeFlag::UNSIGNED);
+        let is_unsigned = tree_node
+            .get_field_type()
+            .as_accessor()
+            .flag()
+            .contains(FieldTypeFlag::UNSIGNED);
         unsigned_flags[i] = is_unsigned;
         match tree_node.get_tp() {
             ExprType::ScalarFunc | ExprType::ColumnRef => {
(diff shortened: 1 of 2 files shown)


guo-shaoge:
@skyzh PTAL

components/tidb_query_expr/src/impl_compare_in.rs
```diff
@@ -196,6 +198,58 @@ where
     }
 }
 
+#[rpn_fn(nullable, varg, capture = [metadata], min_args = 1, metadata_mapper = init_compare_in_data::<T>)]
+#[inline]
```
skyzh, line 202: It seems that `compare_in_int_type_by_hash_fn_meta` is only used for `Int`. Can we remove generics in this function?

components/tidb_query_expr/src/impl_compare_in.rs
```diff
@@ -196,6 +198,58 @@ where
...
+    assert!(!args.is_empty());
+    let base_val = args[0];
+    let arg0_unsigned = metadata.unsigned_flags[0];
+    match base_val {
+        None => Ok(None),
+        Some(base_val) => {
+            let base_val = T::map_ref(base_val)?;
+            match metadata.lookup_map.get(base_val) {
+                None => {}
+                Some(&argi_unsigned) => {
```
skyzh, line 220: Just use `argi_unsigned` here.

skyzh (commented):
Sorry but I'm quite confused about this PR. I'm not sure if this condition is correct for all possible combinations of unsigned, signed int.

```
(*base_val >= 0)
                         || (arg0_unsigned && argi_unsigned)
                         || (!arg0_unsigned && !argi_unsigned)
```

Would you please explain this a little bit?

Personally, I would suggest removing generics of the newly-added compare function, and then adding more convincing tests which ensure this works.

components/tidb_query_expr/src/impl_compare_in.rs
skyzh:
```
if let Some(argi_unsiged) = metadata.lookup_map.get(base_val) {
}
```

components/tidb_query_expr/src/impl_compare_in.rs
guo-shaoge: use Int instread of T as generic type

components/tidb_query_expr/src/impl_compare_in.rs
guo-shaoge: done.

guo-shaoge:
> Sorry but I'm quite confused about this PR. I'm not sure if this condition is correct for all possible combinations of unsigned, signed int.
> 
> ```
> (*base_val >= 0)
>                          || (arg0_unsigned && argi_unsigned)
>                          || (!arg0_unsigned && !argi_unsigned)
> ```
> 
> Would you please explain this a little bit?
> 
> Personally, I would suggest removing generics of the newly-added compare function, and then adding more convincing tests which ensure this works.

Copied from TiDB to be compatiable with MySQL.

eg:

    col_signed_int  IN (max_uint64)

In the above case. If col_signed_int is -1, result of IN expr is false. So we need to check if unsigned/signed flags.

Why we check "*base_val >= 0"
eg:

    col_signed_int IN (col_unsigned_int)

In the above case. If both col_signed_int and col_usngned_int are 1, result of IN expr shoudle be true. So if *base_val is greater than zero, no need to check signed/unsigned falg.

components/tidb_query_expr/src/impl_compare_in.rs
```diff
@@ -155,8 +155,10 @@ impl InByCompare for DateTime {}
 
 #[derive(Debug)]
 pub struct CompareInMeta<T: Eq + Hash> {
-    lookup_set: HashSet<T>,
+    lookup_map: HashMap<T, bool>,
```
skyzh, line 158: Consider the following case: there are both `-1` and `2^63-1` (which maps to `-1`). Will this expression pass this case?

skyzh: Please also fix compile errors and lints.

skyzh (approved):
LGTM. Please fix the CI test. By the way, I think we could refactor code both on TiKV and TiDB side, to make it more efficient in the case I mentioned before...

guo-shaoge:
@lzmhhh123 PTAL

components/tidb_query_expr/src/impl_compare_in.rs
guo-shaoge: new cases are added.

lzmhhh123 (approved):

PR #8698, opened 2020-09-18
*: replace `failure` and `quick-error` with `anyhow` and `thiserror`

koushiro:
### What problem does this PR solve?

Problem Summary: 

To be honest, this is a boring PR, but I think it's worth it.

The crate `failure` has been deprecated.

And I think the syntax of `quick-error` is not as intuitive as `thiserror`.

In addition, we can unify the usage of error handling.

### What is changed and how it works?

What's Changed:

* replace `failure` with `anyhow` and `thiserror`
* replace `quick-error` with `thiserror`

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- No code

Side effects

- No

### Release note 

No

PR #9825, opened 2021-03-16
*: replace `quick-error` with `thiserror` partially

koushiro:

Problem Summary: part of #8698

What's Changed: replace `quick-error` with `thiserror` in some crates

### Check List 

Tests 

- No code

Side effects

- No

### Release note 

No


components/engine_traits/src/errors.rs
```diff
@@ -1,55 +1,41 @@
...
+    Codec(#[from] tikv_util::codec::Error),
+    #[error("The entries of region is unavailable")]
+    EntriesUnavailable,
+    #[error("The entries of region is compacted")]
+    EntriesCompacted,
+}
+
+impl From<String> for Error {
+    fn from(err: String) -> Self {
+        Error::Engine(err)
```
BusyJay, line 38: This is too broad to be implemented. How about removing the trait and using `Error::Engine` directly?

koushiro: There may be dozens of implicit conversions about this in `engine_rocks`, I am not sure whether to remove this trait.

BusyJay: So terrible. 😓

components/engine_traits/src/errors.rs
```diff
@@ -1,55 +1,41 @@
...
+pub enum Error {
+    // Engine uses plain string as the error.
+    #[error("Storage Engine {0}")]
+    Engine(String),
+    // FIXME: It should not know Region.
+    #[error(
+        "Key {} is out of [region {}] [{}, {})",
+        log_wrappers::Value::key(.0), .1, log_wrappers::Value::key(&.2), log_wrappers::Value::key(&.3)
+    )]
+    NotInRange(Vec<u8>, u64, Vec<u8>, Vec<u8>),
```
BusyJay, line 19: Can it be an anonymous struct?

components/engine_traits/src/lib.rs
```diff
@@ -254,8 +254,6 @@
 #![feature(min_specialization)]
 #![recursion_limit = "200"]
```
BusyJay, line 255: This was set for `quick_error`, is it still necessary?

components/engine_traits/src/errors.rs
koushiro: like this?

```
    #[error(
        "Key {} is out of [region {}] [{}, {})",
        log_wrappers::Value::key(&key), region_id, log_wrappers::Value::key(&start), log_wrappers::Value::key(&end)
    )]
    NotInRange {
        key: Vec<u8>, 
        region_id: u64, 
        start: Vec<u8>, 
        end: Vec<u8>,
    },
```

components/engine_traits/src/lib.rs
koushiro: no

components/engine_traits/src/errors.rs
BusyJay: Yes.

koushiro: done

BusyJay (approved):

nrc (approved):
awesome, thank you!

/lgtm

PR #9836, opened 2021-03-17
copr: add example plugin for integration tests

Jibbow:

Adds an example plugin for the new `coprocessor_v2` that can be used for integration tests. The plugin currently does nothing, but it compiles to a proper `cdylib`. Also, this PR adds a function to resolve package names to library names.

### Related changes

- based on PR #9802

### Check List 

Tests 

- The PR does not contain any tests, but rather adds an artifact that can be used for testing. See the follow-up PR (#9837)

### Release note 

 * No release notes


Jibbow:
/cc @andylokandy @skyzh 

But we first need to finish #9802 because otherwise this PR will not compile (jemalloc) and also it's harder to review because of of a larger diff.

Jibbow:
Closing, as this will be part of #9802

PR #9839, opened 2021-03-18
add test cases for GC bypassing Raft

hicqu:

Add a test case for GC bypassing Raft.
There are already some cases examining it in `src/server/gc_worker/gc_worker.rs`, but it's not enough because they are all based on a mock engine. This case uses real rocks engine.

### Check List 

Tests 

- Integration test

### Release note 
* No release note.


MyonKeminta (approved):
LGTM

youjiali1995 (approved):

Issue #9799, opened 2021-03-13
There is no limit to reconnect with PD


HunDunDM:
### What version of TiKV are you using?

```
Release Version:   5.0.0-rc.x
Edition:           Community
Git Commit Hash:   2d3c66cc44ea19a4e339ec44cf1fcd1477555b7d
Git Commit Branch: master
UTC Build Time:    2021-02-08 09:58:47
Rust Version:      rustc 1.51.0-nightly (1d0d76f8d 2021-01-24)
Enable Features:   jemalloc mem-profiling portable sse protobuf-codec test-engines-rocksdb
Profile:           dist_release
```

### Steps to reproduce

* Large-scale cluster
* For some reasons, PD does not have a leader

### What did you expect?

* TiKV performs the backoff strategy normally and waits for PD to recover.

### What did happened?

* Each Region heartbeat triggers a reconnection with PD. A large number of regions triggers similar DDoS attacks on PD. PD cannot be recovered.
* `[INFO] [util.rs:396] ["connecting to PD endpoint"]` Thousands of times per second.

coocood:
Does 4.0 has this issue?

HunDunDM:
> Does 4.0 has this issue?

yes

PR #9840, opened 2021-03-18
pd_client: prevent a large number of reconnections in a short time

HunDunDM:

Issue Number: related #9799 

Problem Summary:

Too many uninterrupted reconnection.

What's Changed:

Cancel some unnecessary reconnection.

### Related changes

### Check List 

Tests 

- Unit test
- Manual test
  - 3 PD / 3 TiKV / 1000 Region
  - Manually make PD no leader
  - Add metrics (will be in another PR)
    - Before image
    - After image

### Release note 

- pd_client: prevent a large number of reconnections in a short time


components/pd_client/src/util.rs
```diff
@@ -170,21 +174,33 @@ impl LeaderClient {
...
+        let future = {
             let inner = self.inner.rl();
-            if inner.last_update.elapsed() < Duration::from_secs(RECONNECT_INTERVAL_SEC) {
+            if inner.last_update.elapsed() < RECONNECT_UPDATE_INTERVAL {
                 // Avoid unnecessary updating.
                 return Ok(());
             }
-
-            let start = Instant::now();
+            if !force && inner.last_try_reconnect.elapsed() < ALL_NON_FORCE_RECONNECT_INTERVAL {
```
BusyJay, line 185: Why check `force`? It should probably always retry if timeout.

HunDunDM: In order to make this restriction only take effect for `reconnect` with `force=false`, I adjusted the level of `reconnect` initiated by region-related requests such as heartbeat requests.

HunDunDM: Two `reconnect`s with `force=false` that are executed need to be separated by more than 100ms, but `force=true` is not limited.

BusyJay: I don't understand why `force` matters. What could go wrong if ignoring `false`? Will the problem be also solved?

HunDunDM: In order to control the scope of influence of this PR, it currently does not limit the retry of sync_request.

BusyJay: They should also handle connection errors, and probably did.

BusyJay:
What's the test result?

HunDunDM:
Updated test results (this metrics will be in another PR). Is it possible that the unit test of CI failure is caused by `pd_client.get_store` blocking?

components/pd_client/src/util.rs
```diff
@@ -239,11 +246,17 @@ impl Client {
...
-            };
-            slow_log!(start.elapsed(), "try reconnect pd");
-            (fut, start)
+            }
         };
 
+        {
+            slow_log!(start.elapsed(), "try reconnect pd");
+            let mut inner = self.inner.wl();
+            if start > inner.last_try_reconnect {
```
BusyJay, line 255: This is always true.

components/pd_client/src/util.rs
```diff
@@ -273,7 +286,9 @@ pub fn build_forward_metadata(forwarded_host: &str) -> Metadata {
     builder.build()
 }
 
-pub const RECONNECT_INTERVAL_SEC: u64 = 1; // 1s
+pub const REQUEST_RECONNECT_INTERVAL: Duration = Duration::from_millis(1000); // 1s
```
BusyJay, line 289: Why not using the same interval?

components/pd_client/src/util.rs
HunDunDM: `start` is not taken in the write lock, I don't think it can be guaranteed.

tests/failpoints/cases/test_pd_client.rs
```diff
@@ -142,3 +142,32 @@ fn test_slow_periodical_update() {
...
+    let server = MockServer::new(1);
+    let eps = server.bind_addrs();
+
+    let mut cfg = new_config(eps);
+    let env = Arc::new(EnvBuilder::new().cq_count(1).build());
+    let mgr = Arc::new(SecurityManager::new(&SecurityConfig::default()).unwrap());
+    cfg.update_interval = ReadableDuration(Duration::from_secs(100));
+    let client = RpcClient::new(&cfg, Some(env), mgr).unwrap();
+
+    // Wait for the PD client thread blocking on the fail point.
```
BusyJay, line 159: Why will it block?

components/pd_client/src/util.rs
```diff
@@ -220,17 +223,18 @@ impl Client {
     ///
     /// If `force` is false, it will reconnect only when members change.
     pub async fn reconnect(&self, force: bool) -> Result<()> {
-        let (future, start) = {
+        let start = Instant::now();
+
+        let future = {
             let inner = self.inner.rl();
-            if inner.last_update.elapsed() < Duration::from_secs(RECONNECT_INTERVAL_SEC) {
+            if inner.last_try_reconnect.elapsed() < GLOBAL_RECONNECT_INTERVAL {
```
BusyJay, line 230: How about `.duration_since(start)`?

components/pd_client/src/util.rs
```diff
@@ -239,15 +243,26 @@ impl Client {
...
                     .await
-            };
-            slow_log!(start.elapsed(), "try reconnect pd");
-            (fut, start)
+            }
         };
 
+        {
+            let mut inner = self.inner.wl();
+            if inner.last_try_reconnect.elapsed() < GLOBAL_RECONNECT_INTERVAL {
```
BusyJay, line 251: `checked_duration_since` is safer.

tests/failpoints/cases/test_pd_client.rs
HunDunDM: typo...

components/pd_client/src/util.rs
```diff
@@ -220,17 +223,18 @@ impl Client {
     ///
     /// If `force` is false, it will reconnect only when members change.
     pub async fn reconnect(&self, force: bool) -> Result<()> {
-        let (future, start) = {
+        let start = Instant::now();
+
+        let future = {
             let inner = self.inner.rl();
-            if inner.last_update.elapsed() < Duration::from_secs(RECONNECT_INTERVAL_SEC) {
+            if start.duration_since(inner.last_try_reconnect) < GLOBAL_RECONNECT_INTERVAL {
```
BusyJay, line 230: `start` is initialized without lock, so it can be less than `last_try_reconnect`. Should use `checked_duration_since` instead.

components/pd_client/src/util.rs
```diff
@@ -220,17 +223,18 @@ impl Client {
...
                 // Avoid unnecessary updating.
-                return Ok(());
+                // Prevent a large number of reconnections in a short time.
+                return Err(box_err!("cancel reconnection due to too small interval"));
```
BusyJay, line 233: Why returning `Error` instead of `Ok` just like the past?

HunDunDM: If the request is received `reconnect ok`, it will try to resend the request immediately and will not `sleep 1s`. At present, both `async request` and `sync request` have been changed to send the request first and then try to reconnect.

tests/failpoints/cases/test_pd_client.rs
```diff
@@ -142,3 +142,31 @@ fn test_slow_periodical_update() {
...
+    let eps = server.bind_addrs();
+
+    let mut cfg = new_config(eps);
+    let env = Arc::new(EnvBuilder::new().cq_count(1).build());
+    let mgr = Arc::new(SecurityManager::new(&SecurityConfig::default()).unwrap());
+    cfg.update_interval = ReadableDuration(Duration::from_secs(100));
+    let client = RpcClient::new(&cfg, Some(env), mgr).unwrap();
+
+    // The REQUEST_RECONNECT_INTERVAL is 1s so sleeps 2s here.
+    thread::sleep(Duration::from_secs(2));
```
BusyJay, line 160: Why sleep?

HunDunDM: The initial value of `last_try_connect` is the time when the client is created. Reconnection within 100ms will be canceled and an error will be returned.

BusyJay (approved):

components/pd_client/src/util.rs
```diff
@@ -291,33 +310,32 @@ where
...
 {
-    async fn reconnect_if_needed(&mut self) -> bool {
+    async fn reconnect_if_needed(&mut self) -> Result<()> {
         debug!("reconnecting ..."; "remain" => self.reconnect_count);
-
         if self.request_sent < MAX_REQUEST_COUNT {
-            return true;
+            return Ok(());
+        }
+        if self.reconnect_count == 0 {
```
Yisaer, line 318: I think `reconnect_count` naming is misleading, can we rename it as `remain_reconnect_count` ?

Yisaer (commented):
Rest LGTM

components/pd_client/src/util.rs
```diff
@@ -239,15 +247,28 @@ impl Client {
...
-            slow_log!(start.elapsed(), "try reconnect pd");
-            (fut, start)
+            }
         };
 
+        {
+            let mut inner = self.inner.wl();
+            if start
+                .checked_duration_since(inner.last_try_reconnect)
+                .map_or(true, |d| d < GLOBAL_RECONNECT_INTERVAL)
```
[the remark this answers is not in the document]
HunDunDM: Because the first judgment is in the read lock.

Yisaer (approved):
LGTM

PR #9843, opened 2021-03-18
*: replace failure with anyhow and thiserror

koushiro:

Problem Summary: part of #8698

What's Changed: replace `failure` with `anyhow` and `thiserror` in some crates

### Check List 

Tests 

- No code

Side effects

- No

### Release note

No


nrc (approved):
lgtm, thanks!

components/backup/src/lib.rs
```diff
@@ -3,8 +3,6 @@
 #![recursion_limit = "200"]
```
BusyJay, line 3: Is it still necessary?

components/tidb_query_common/Cargo.toml
```diff
@@ -17,8 +17,8 @@ prost-codec = [
 ]
 
 [dependencies]
+anyhow = "1.0"
```
BusyJay, line 20: Is it possible to keep only depending on thiserror?

nrc: I think we need anyhow for dynamic errors. IMO, we shouldn't have any dynamic errors, but in the short term, I think it is easier to keep using them and depend on anyhow. Longer term I have some ideas to transition away.

components/backup/src/lib.rs
koushiro: no, I will remove it

BusyJay (approved):

Issue #9609, opened 2021-01-31
Remove engine_rocks dependency from tikv_kv


brson:
## Development Task

https://github.com/tikv/tikv/pull/9608 extracts tikv::storage::kv to its own crate. This crate depends on engine_rocks, but shouldn't. Remove the dependency. This will probably entail transforming the concrete RocksEngine type into a type parameterized over engine_traits.

brson:
I had a bit of a revelation this week - it's not strictly necessary to remove the rocks dependencies from tikv_kv in order to try to start up tikv with another engine, because the production implementation of the traits in this crate is `RaftKv`, and _that_ implementation can be abstracted over engines trivially, in https://github.com/tikv/tikv/pull/10013.

The tikv_kv crate does still need some work, particularly to make its use of the global PerfContext resource work with other engines, but it's a lower priority for me now.

brson:
This is not done, but I am closing it as it is not on the immediate roadmap.

PR #9845, opened 2021-03-18
Remove two uses of engine_rocks from tikv_kv

brson:

Issue: https://github.com/tikv/tikv/issues/9609

Problem Summary:

tikv_kv uses engine_rocks directly and it should not.

- Remove the get_properties / get_properties_cf methods from tikv_kv::Engine trait. These are fortunately unused.
- Use PanicEngine instead of RocksEngine as an unimplemented placeholder in BTreeEngine
   - Ultimately no concrete engine should be used here, but this is an easy way to get rid of RocksEngine, and PanicEngine is more semantically appropriate here.

### Release note 

- No release note.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5400,6 +5400,7 @@ version = "0.1.0"
 dependencies = [
  "backtrace",
  "derive_more",
+ "engine_panic",
  "engine_rocks",
  "engine_traits",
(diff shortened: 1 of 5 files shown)


5kbpers (approved):
LGTM :shipit:

PR #9846, opened 2021-03-18
server: Use KvTestEngine instead of RocksEngine in test

brson:

Issue: https://github.com/tikv/tikv/issues/6402

Problem Summary:

tikv uses RocksEngine directly and should not

This just replaces a single test instance of RocksEngine with KvTestEngine.

### Release note 

- No release note


kennytm:
/test 

#7265 (https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/tikv_ghpr_test/runs/44889/nodes/223/log/?start=0).

PR #9849, opened 2021-03-19
pd_client: add the state of forwarding metrics

rleungx:

Problem Summary: It's hard to know if the forwarding mechanism is running for now.

What's Changed: This PR adds the metrics named `tikv_pd_is_request_forwarded` to show if the forwarding is running.

### Related changes

### Check List 

Tests 

- Manual test

Start a cluster and set up a network partition problem between all tikv and pd leader.
Screen Shot 2021-03-19 at 5 30 51 PM
We can find all tikv is try to connect the pd leader 1 through another follower.

Then we kill tikv 2 and recover the network
Screen Shot 2021-03-19 at 5 31 04 PM
The newly restarted TiKV is don't need to forward and the other two are also change back to directly connection after few minutes.

### Release note 

- Add the metrics to show if the request forwarding mechanism is running


components/pd_client/src/client.rs
```diff
@@ -75,6 +75,9 @@ impl RpcClient {
         for i in 0..retries {
             match pd_connector.validate_endpoints(cfg).await {
                 Ok((client, forwarded_host, members)) => {
+                    if !forwarded_host.is_empty() {
+                        REQUEST_FORWARDED_GAUGE.set(1);
```
BusyJay, line 79: Better monitor which address is used.

components/pd_client/src/client.rs
```diff
@@ -75,6 +77,12 @@ impl RpcClient {
         for i in 0..retries {
             match pd_connector.validate_endpoints(cfg).await {
                 Ok((client, forwarded_host, members)) => {
+                    if !forwarded_host.is_empty() {
+                        let host = trim_http_prefix(&forwarded_host);
```
BusyJay, line 81: How about adding a comment explaining why not save the result of trim?

BusyJay (commented):
Rest LGTM

BusyJay (approved):

nolouch (approved):

PR #9851, opened 2021-03-19
copr: optimize limit operator next_batch method

### LINKED ISSUE #9828: coprocessor: Scan + Limit operator scan much more keys than needed.
## Bug Report

The vectorized executor initially scans 32 rows, but when the executor is just Scan + Limit(1)，we only need 1 row.

In a POC scenario, the performance increased from 60K to 130K after a temp fix.

v5.0-rc

Linux

### Steps to reproduce
- create table t (id bigint primary key, val varchar(1024));
- insert 1 million records.
- run a benchmark with query `select * from t order by id limit 1`

### What did you expect?
metrics show `TiKV Details/RocksDB kv/Seek operations`, `next` ops similar to `seek` ops.

### What did happened?
metrics show `TiKV Details/RocksDB kv/Seek operations`, `next` ops tens of times more than `seek` ops.

guo-shaoge:

Issue Number: close #9828 

Problem Summary: Performance enhancement for Limit operator. When chall child's next_batch(scan_rows), default scan_rows is 32, but if limit number is 1 or less, we scan more rows than we actually need.

What's Changed: change LimitBatchExecutor::next_batch(). See files changed, it's a small change.

### Related changes

### Check List 

Tests 

- performance test
- unitest

Side effects

- performance enhancement for limit operator. [here](https://github.com/tikv/tikv/pull/9851#issuecomment-803539262)

### Release note 
- copr: performance enhancement for limit operator


breeswish (approved):

BusyJay:
Can you post a summary of the result (images are better) in the PR description? I (and many other contributors) don't have the permission to see the doc.

components/tidb_query_executors/src/limit_executor.rs
```diff
@@ -32,7 +32,9 @@ impl<Src: BatchExecutor> BatchExecutor for BatchLimitExecutor<Src> {
 
     #[inline]
     fn next_batch(&mut self, scan_rows: usize) -> BatchExecuteResult {
-        let mut result = self.src.next_batch(scan_rows);
+        let mut result = self
+            .src
+            .next_batch(std::cmp::min(scan_rows, self.remaining_rows));
```
BusyJay, line 37: What if it's something like `select * from (select col, count(*) from table group by col) limit 1`?

coocood: This SQL will not push down LIMIT to the coprocessor.

BusyJay: What if it's something like `select * from table where non_index_col != "" limit 1`?

> ...will not push down...

How to avoid unexpected regression if it's pushed down one day? (For example, all data are within one region)

coocood: If the condition is never satisfied, then there will be regression, but the regression is only noticeable for AP request.
AP request rarely use LIMIT in statement.

BusyJay: condition with limit is very common, perhaps we need another survey? 🤣

coocood: > condition with limit is very common, perhaps we need another survey? 🤣

Common in TP requests, not in AP. Limit in AP doesn't make much sense because we need statistics for all relevent data.

guo-shaoge:
环境：
    4核 AMD Ryzen 5 PRO 4650U with Radeon Graphics
    4G内存
    3个TiDB, 1个TiKV, 1个PD

测试SQL：
    use test;
    drop table if exists t;
    create table t (gjt_col bigint primary key, val varchar(1024));
    load data local infile '/home/guo-shaoge/work/limit_perf/row_1000000_col_2' into table t fields terminated by ' ';
    select * from t order by gjt_col limit 1;

测试方法：
    客户端并发度30，以round robin方式将不同线程的请求平均发送给3个TiDB。每个TiDB收到请求后，会将select语句循环2000次

    统计的时间就是2000次执行的时间，测试时TiKV的cpu被压满，CPU占用基本到400%。测试结论如下

结果：
    Limit算子修改前: 执行时间在49.9s到1min之间，平均时间为 1min

    修改后：30个线程的执行时间在27.9s到29.7s之间，平均时间为 28s

可以看到，只有Limit算子情况下，性能提升比较大，但是带有selection的SQL有回退

测试SQL，schema等信息同上一个测试一样:
    select gjt_col from t where val > 3211489453760402549 limit 1;

测试方法：
    跟上面的测试方法一样，但是并发度改为10

结果：
   Limit算子修改前: 执行时间在1min8s到1m40s之间，平均时间为 1min25s

    修改后：30个线程的执行时间在42s到1min之间，平均时间为 55s

guo-shaoge:
But only change Limit operator will affect performance of  **selection** operator. Check the following test.

enviroments:

        4 cores. AMD Ryzen 5 PRO 4650U with Radeon Graphics
        4G memory
        3 TiDB, 1 TiKV, 1 PD

SQL(same schema and input data with above test):

    ('3211489453760402549' is a random choice, There are 733502 rows out of 1000000 satify this filter)
    SQL1: select gjt_col from t where val > '3211489453760402549' limit 1;

    ('999995990325667965' is the max value for column `val`, so table scan needs to scan all rows of table `t`)
    SQL2: select gjt_col from t where val > '999995990325667965' limit 1;

    ('100000152253498439' is the min value for column `val`, so table scan only need to scan small number of rows)
    SQL3: select gjt_col from t where val >= '100000152253498439' limit 1

how to test:

    SQL1: Same as above. But client will use 10 goroutines(Because workload is too heavy, cost too much time).

    SQL2: TiDB will execute each request 20 times, and client will only have 3 goroutines(workload is too heavy).

    SQL3: Save as above. 2000 times for each request, and client has 30 goroutines.

SQL1 result:
    
    before: exec duratoin for 30 goroutines: [42s, 1min]，avg: 55s

    after: 1min8s, 1m40s，avg: 1min25s (**regression here!**)

SQL2 result:

    before: avg: 7s

    after: avg: 18s (**regression here!**)

SQL3 result:

    before: avg: 55s

    after: avg: 53s

So need to also change implementation of selection's next_batch() method.

guo-shaoge:
> Can you post a summary of the result (images are better) in the PR description? I (and many other contributors) don't have the permission to see the doc.

Ok. Check [limit operator enhancement](https://github.com/tikv/tikv/pull/9851#issuecomment-803539262)  and [select operation regression](https://github.com/tikv/tikv/pull/9851#issuecomment-803578571)

components/tidb_query_executors/src/limit_executor.rs
guo-shaoge: Maybe we can fix regression of selection operator in another issue.

guo-shaoge:
@skyzh PTAL

components/tidb_query_executors/src/limit_executor.rs
```diff
@@ -32,7 +34,11 @@ impl<Src: BatchExecutor> BatchExecutor for BatchLimitExecutor<Src> {
 
     #[inline]
     fn next_batch(&mut self, scan_rows: usize) -> BatchExecuteResult {
-        let mut result = self.src.next_batch(scan_rows);
+        let mut real_scan_rows = scan_rows;
```
skyzh, line 37:
```rust
let real_scan_rows = if self.is_src_scan_executor {
    std::cmp::min(scan_rows, self.remaining_rows)
} else {
    scan_rows
};
```

components/tidb_query_executors/src/limit_executor.rs
```diff
@@ -225,7 +231,7 @@ mod tests {
             ],
         );
 
-        let mut exec = BatchLimitExecutor::new(src_exec, 4).unwrap();
+        let mut exec = BatchLimitExecutor::new(src_exec, 4, false).unwrap();
```
skyzh, line 234: Please also consider adding some cases for `is_src_scan_executor = true`.

components/tidb_query_executors/src/runner.rs
```diff
@@ -286,8 +287,12 @@ pub fn build_executors<S: Storage + 'static>(
                 EXECUTOR_COUNT_METRICS.batch_limit.inc();
 
                 Box::new(
-                    BatchLimitExecutor::new(executor, ed.get_limit().get_limit() as usize)?
-                        .collect_summary(summary_slot_index),
+                    BatchLimitExecutor::new(
+                        executor,
+                        ed.get_limit().get_limit() as usize,
+                        is_scan_executor,
```
skyzh, line 293: Would you please explain this logic a little bit? Seems that `is_scan_executor` is always true as this point.

guo-shaoge: `is_src_scan_executor` will be set as false when non-scan executor be created. (at the end of this for loop)

components/tidb_query_executors/src/limit_executor.rs
guo-shaoge: add test_src_exec_is_scan

components/tidb_query_executors/src/limit_executor.rs
guo-shaoge: done

skyzh (approved):
LGTM. Current design seems ad-hoc to the case that limit is the bottom-level executor. We may later let TiDB add some hint on scan rows.

components/tidb_query_executors/src/limit_executor.rs
guo-shaoge: > Maybe we can fix regression of selection operator in another issue.

This fix will not affect performance of **selection**. I will post performance test of selection later.

PR #9651, opened 2021-02-13
pd_client: refine PD client codes

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

### What problem does this PR solve?

Import macros from other crates with the `use` syntax. 

### What is changed and how it works?

rust-lang/rust#50911

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

* No release note.

PR #9854, opened 2021-03-22
pd_client: link PD client to tikv_alloc

JmPotato:

#9651 accidentally deleted the link between `tikv_alloc` and `pc_client`. This PR gets it back.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

* No release note.


components/pd_client/src/lib.rs
```diff
@@ -1,4 +1,6 @@
 // Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
+#[allow(unused_extern_crates)]
+extern crate tikv_alloc;
```
BusyJay, line 3: If it's unused, why need to link to it explictily?

JmPotato: According to the description of `tikv_alloc`:

> Writing `extern crate tikv_alloc;` will link it to jemalloc when appropriate. The TiKV library itself links to `tikv_alloc` to ensure that any binary linking to it will use jemalloc. With few exceptions, _every binary and project in the TiKV workspace should link (perhaps transitively) to tikv_alloc_. This is to ensure that tests and benchmarks run with the production allocator. In other words, binaries and projects that don't link to `tikv` should link to `tikv_alloc`.

BusyJay: Given that features is not specified in `Cargo.toml`, I don't see the point to link it everywhere. /cc @brson

JmPotato: What do you think? @brson @nrc

nrc: I think that since we depend on crates which depend on tikv_alloc, then it is not necessary here. Of course that is somewhat fragile because if we end up not depending on any crate and our transitive dependency disappears, then we wouldn't get any warning. So for the sake of two lines of boilerplate I think it is good to have a direct dependency.

BusyJay (approved):

lhy1024 (approved):

HunDunDM (approved):

PR #9855, opened 2021-03-22
cdc: fix old value cache may always missed

5kbpers:

Problem Summary:
Previously we refactored the logic about old-value in the transaction layer (see #9173). This introduces a bug that the old-value cache will be always missed because the key of the cache doesn't contains `start ts`.

Proposal: [xxx](url) 

What's Changed:
* Append start ts to the key of the cache
* Make insert operations updating the cache
* Make the cache works in the tests

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* cdc: fix old value cache maybe always missed

### DIFF
--- a/components/txn_types/src/types.rs
+++ b/components/txn_types/src/types.rs
@@ -249,6 +249,8 @@ impl MutationType {
     pub fn may_have_old_value(&self) -> bool {
         matches!(
             self,
+            // Insert operations don't have old value but need to update a flag
+            // for indicating that not seeking for old value for it.
             MutationType::Put | MutationType::Delete | MutationType::Insert
         )
     }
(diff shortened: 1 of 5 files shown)


components/txn_types/src/types.rs
```diff
@@ -247,7 +247,10 @@ pub enum MutationType {
 
 impl MutationType {
     pub fn may_have_old_value(&self) -> bool {
-        matches!(self, MutationType::Put | MutationType::Delete)
+        matches!(
+            self,
+            MutationType::Put | MutationType::Delete | MutationType::Insert
```
overvenus, line 252: Please add comments on why they may have old value.

overvenus (commented):
Rest LGTM

PR #9860, opened 2021-03-23
copr: cast invalid utf8 string to real bug

guo-shaoge:

Issue Number: close #[TiDB issue 23322](https://github.com/pingcap/tidb/issues/23322)  

Problem Summary: When cast bytes to real, we first cast bytes to utf8, so it will return ERROR when bytes are not valid utf8 string.

What's Changed: Just like bytes to int, we first get valid utf8 prefix from bytes, then convert it to float.

### Related changes

### Check List 

Tests 

- Unit test: test_bytes_to_f64_invalid_utf8()

Side effects: no

### Release note 
- copr cast invalid utf8 string to real bug fix


guo-shaoge:
@skyzh PTAL

skyzh (approved):

guo-shaoge:
@lzmhhh123 PTAL

lzmhhh123 (approved):

guo-shaoge:
@AilinKid

Issue #9859, opened 2021-03-22
raftstore: leader may not send snapshot to the follower


NingLin-P:
### What version of TiKV are you using?

5.0-rc, master

The leader will be triggered to generates snapshot for a follower in two places: 

https://github.com/tikv/tikv/blob/f02f492614c10d1cdffce8e0214feef15ce521f0/components/raftstore/src/store/peer.rs#L1662-L1671

https://github.com/tikv/tikv/blob/f02f492614c10d1cdffce8e0214feef15ce521f0/components/raftstore/src/store/peer.rs#L1881-L1886

The first place will not enter if the region has a `Ready`,  and the latter one will also not enter if the `Ready` has not committed entries because it is inside `handle_raft_committed_entries` and:
https://github.com/tikv/tikv/blob/f02f492614c10d1cdffce8e0214feef15ce521f0/components/raftstore/src/store/peer.rs#L1729-L1731

BusyJay:
Why 4.0.x is safe?

gengliqi:
> Why 4.0.x is safe?

Because the function name of `handle_raft_committed_entries` in 4.0 is `handle_raft_ready_apply`, which is always called after getting ready.

ilovesoup:
Does this affect learner?

NingLin-P:
> Does this affect learner?

Yes, it will affect learner.

PR #9863, opened 2021-03-23
raftstore: generate snapshot(if any) after getting ready

gengliqi:

Issue Number: close https://github.com/tikv/tikv/issues/9859

Problem Summary:
This bug is introduced by #9177. 
The snapshot should be generated(if any) after getting ready.
But #9177 adds one more condition so if there is no newly committed entry, the snapshot can't be generated.
In extreme cases, this bug will cause the leader to be unable to send snapshots to followers.

What's Changed:
Generate snapshot(if any) after getting ready

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No

### Release note 
* No Release Note


tests/failpoints/cases/test_snap.rs
```diff
@@ -436,3 +436,33 @@ fn test_receive_old_snapshot() {
 
     fail::remove(peer_2_handle_snap_mgr_gc_fp);
 }
+
+#[test]
+fn test_snapshot_after_isolation() {
```
BusyJay, line 441: Should add comment explaining the purpose of the case.

NingLin-P (approved):

tests/failpoints/cases/test_snap.rs
BusyJay: It can pass successfully without the patch.

NingLin-P: It can't pass with the `before_no_ready_gen_snap_task` fail point enabled, IMO the `before_no_ready_gen_snap_task` fail point is a simple way to simulate a non-empty `Ready` that without any `committed_entries`

gengliqi: Comment has been added.
This test cannot be passed without the patch on my laptop. I guess you forget to add the failpoint.

BusyJay (approved):

PR #9869, opened 2021-03-23
cdc: add version checking for resolved ts

5kbpers:

Problem Summary:
During the rolling update, the new mechanism of advancing resolved ts may do not work.

Add version checking for resolved ts.

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* cdc: add version checking for resolved ts


5kbpers:
@liuzix PTAL

components/cdc/src/endpoint.rs
```diff
@@ -46,6 +46,8 @@ use crate::metrics::*;
 use crate::service::{CdcEvent, Conn, ConnID, FeatureGate};
 use crate::{CdcObserver, Error, Result};
 
+const FEATURE_RESOLVED_TS_STORE: Feature = Feature::require(5, 0, 0);
```
overvenus, line 49: CDC has its own feature gate structure, could you unify them ?

overvenus (commented):
Rest LGTM

components/cdc/src/endpoint.rs
5kbpers: They have different purposes, this is for checking the tikv cluster version.

overvenus (commented):
/lgtm

PR #9881, opened 2021-03-24
storage: Fix ttl expire ts overflow

Connor1996:

When ttl is set to a very large value, such u64::MAX, calculating expire ts(current_ts + ttl) would overflow. 

using `saturating_add` for calculating ttl expire ts.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

- None


components/engine_traits/src/util.rs
```diff
@@ -4,6 +4,8 @@ use super::{Error, Result};
 use tikv_util::codec;
 use tikv_util::codec::number::{self, NumberEncoder};
 
+pub const TEST_CURRENT_TS: u64 = 100;
```
BusyJay, line 7: Why define it twice?

Connor1996: forget to clean up

Connor1996:
PTAL again

BusyJay (approved):

Little-Wallace (approved):
LGTM

PR #9883, opened 2021-03-24
raftstore: import macros with the use syntax 

lhy1024:

Issue Number: supports #9678

### Check List 

Tests 

- Unit test

### Release note 
- No release note


components/raftstore/src/coprocessor/config.rs
```diff
@@ -5,6 +5,9 @@ use crate::store::SplitCheckTask;
 
 use configuration::{ConfigChange, ConfigManager, Configuration};
 use engine_traits::{config as engine_config, PerfLevel};
+use serde::*;
```
JmPotato, line 8: The reason we import macros with the `use` syntax is that we want to make it easy to find the definition and source of imported macros. IMO, using `::*` will go against the original intention why we want to clean up the `extern crate xxx`.

JmPotato (commented):
LGTM

JmPotato:
@lhy1024 Please fix the lints!

lhy1024:
@JmPotato

JmPotato (approved):
LGTM

HunDunDM (approved):

PR #9888, opened 2021-03-25
txn: return min_commit_ts for check_not_exists

youjiali1995:

Problem Summary:

TiKV always returns min_commit_ts=0 for `check_not_exists` operations. It may result in unnecessary falling back for async commit.

What's Changed:

`check_not_exists` operations don't write locks or change data, so we needn't ask the concurrency manager for the max_ts. Its min_commit_ts is `max(start_ts + 1, req.min_commit_ts)`.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 
* No release note.

### DIFF
--- a/src/storage/txn/actions/prewrite.rs
+++ b/src/storage/txn/actions/prewrite.rs
@@ -66,6 +66,8 @@ pub fn prewrite<S: Snapshot>(
     };
 
     if mutation.should_not_write {
+        // `checkNotExists` is equivalent to a get operation, so it should update the max_ts.
+        txn.concurrency_manager.update_max_ts(txn_props.start_ts);
         let min_commit_ts = if mutation.need_min_commit_ts() {
             // Don't calculate the min_commit_ts according to the concurrency manager's max_ts
             // for a should_not_write mutation because it's not persisted and doesn't change data.
@@ -615,7 +617,7 @@ pub mod tests {
     #[test]
     fn test_async_commit_prewrite_min_commit_ts() {
         let engine = crate::storage::TestEngineBuilder::new().build().unwrap();
-        let cm = ConcurrencyManager::new(42.into());
+        let cm = ConcurrencyManager::new(41.into());
         let snapshot = engine.snapshot(Default::default()).unwrap();
 
         // should_not_write mutations don't write locks or change data so that they needn't ask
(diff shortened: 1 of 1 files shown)


src/storage/txn/actions/prewrite.rs
```diff
@@ -358,12 +364,15 @@ impl<'a> PrewriteMutation<'a> {
         }
     }
 
+    fn need_min_commit_ts(&self) -> bool {
```
cfzjywxk, line 367: Seems we are using different ways to judge if the current prewrite processing should use async commit logic, like 
`let need_min_commit_ts = secondaries.is_some() || self.try_one_pc;` which uses the `secondaries` field in the `prewriter`, and we've got some fields like `CommitKind`. Do we have methods to remove or reduce some fields?

cfzjywxk: I'm not sure if we have time to do this before the 5.0 GA as it may introduce extra risks, maybe we could have a try later.

youjiali1995: Agree. I tried to unify the judgment, but seems it's a little dangerous to do it for now.

cfzjywxk:
LGTM

src/storage/txn/actions/prewrite.rs
```diff
@@ -59,7 +58,14 @@ pub fn prewrite<S: Snapshot>(
     };
 
     if mutation.should_not_write {
-        return Ok((TimeStamp::zero(), OldValue::Unspecified));
+        let min_commit_ts = if mutation.need_min_commit_ts() {
+            // Don't calculate the min_commit_ts according to the concurrency manager's max_ts
+            // for a should_not_write mutation because it's not persisted and doesn't change data.
+            cmp::max(txn_props.min_commit_ts, txn_props.start_ts.next())
```
sticnarf, line 64: Do we need to also maximize it with `for_update_ts.next()`?

youjiali1995: No.
```rust
        let should_not_write = mutation.should_not_write();
        if txn_props.is_pessimistic() && should_not_write {
            return Err(box_err!(
                "cannot handle checkNotExists in pessimistic prewrite"
            ));
        }
```

sticnarf (approved):
LGTM

cfzjywxk (approved):

MyonKeminta (approved):
LGTM

PR #9019, opened 2020-11-11
*: add memory trace

BusyJay:
### What is changed and how it works?

What's Changed:
Instrumenting code to trace memory usage.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 
- Instrumenting raftstore to trace memory usage

PR #9892, opened 2021-03-25
*: fix mailbox leak

BusyJay:

Problem Summary:

#9019 shows that even after using LRU cache, a large memory are still held by router. This is because the cache is used everywhere, and waiting for all cache to be cleared can be very slow. This PR fixes the problem completely by discarding the memory when all receivers are dropped.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- Fix memory leak by slow eviction of mailbox cache


components/cdc/src/endpoint.rs
```diff
@@ -1203,8 +1204,8 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Runnable for Endpoint<T> {
                 cb,
             } => {
                 debug!("downstream was initialized"; "downstream_id" => ?downstream_id);
-                downstream_state
-                    .compare_and_swap(DownstreamState::Uninitialized, DownstreamState::Normal);
+                let _ = downstream_state
```
BusyJay, line 1207: Is it safe to ignore error? @5kbpers @overvenus

components/cdc/src/endpoint.rs
```diff
@@ -340,9 +341,9 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Endpoint<T> {
                 if is_last {
                     let delegate = self.capture_regions.remove(&region_id).unwrap();
                     if let Some(reader) = self.store_meta.lock().unwrap().readers.get(&region_id) {
-                        reader
+                        let _ = reader
```
BusyJay, line 344: Another similar case @5kbpers

components/cdc/src/endpoint.rs
5kbpers: yes

BusyJay: If so, why not just use `store`?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1324,6 +1324,8 @@ impl<EK: KvEngine, ER: RaftEngine> RaftBatchSystem<EK, ER> {
             warn!("set thread priority for raftstore failed"; "error" => ?e);
         }
         self.workers = Some(workers);
+        // This router will not be accessed again, free all caches.
+        self.router.clear_cache();
```
Connor1996, line 1328: Seems it's the main changes to fix the leak. Does the crossbeam update related to this fix?

BusyJay: See https://github.com/crossbeam-rs/crossbeam/pull/669

components/cdc/src/endpoint.rs
overvenus: Could you unwrap it?

sticnarf (approved):
LGTM

components/cdc/src/endpoint.rs
5kbpers: Because it could be `Stopped` state

components/raftstore/src/store/fsm/store.rs
gengliqi: Seems this PR has not been included in the release version of crossbeam yet?

components/cdc/src/endpoint.rs
overvenus: Is it safe to transit state from `Stopped` to `Normal`?

components/concurrency_manager/Cargo.toml
```diff
@@ -21,8 +21,8 @@ fail = "0.4"
 
 # FIXME: switch to the crates.io version after crossbeam-skiplist is released
```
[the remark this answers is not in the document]
BusyJay: Actually we still want to switch to the released version later as two patches are all merged.

components/raftstore/src/store/fsm/store.rs
BusyJay: Yes, that's the reason why a forked repository is used.

components/cdc/src/endpoint.rs
5kbpers: No, so we should not unwrap it here

components/raftstore/src/store/fsm/store.rs
NingLin-P: `self.router`'s cache only contains 1024 slots, the main fix should be the crossbeam patch?

BusyJay: Yes.

NingLin-P (approved):
LGTM

Connor1996 (approved):
LGTM

gengliqi (approved):

hi-rustin:
> /merge

You only need to trigger /merge once, and if the CI test fails, you just re-trigger the test that failed and the bot will merge the PR for you after the CI passes.

Issue #9785, opened 2021-03-11
Region size growing speed may exceed splitting speed when there is a write hotspot


MyonKeminta:
### What version of TiKV are you using?

master

### What operating system and CPU are you using?

\-

### Steps to reproduce

Construct a write hotspot scenario, such as:

In a cluster with clustered index enabled, run tpcc prepare with large thread count:

```
nohup ./go-tpc -H *** -P3306 -D tpcc tpcc prepare --threads 768 --warehouses 150000 > prepare.log 2>&1 &
nohup ./go-tpc -H *** -P3306 -D tpcc_1 tpcc prepare --threads 192 --warehouses 300000 > prepare_1.log 2>&1 &
```

### What did you expect?

The cluster runs normally

### What did happened?

The region size on the write hotspot grows too quickly, and the automatic  region splitting can't follow up.

Currently, the split check runs every 10s, and it by default generates at most 10 new regions whose size are about 96M. And if the region is generating a snapshot to send to a replica, the split will be cancelled (at most 3 times) to avoid invalidating the snapshot to be generated. So if the region is large enough and snapshot generating is slow, then the region might split only every 40s (or even slower due to other potential problem when region size is too large),  which makes the split speed even slower.

When the issue happens, the too-large region may make the cluster unable to balance data, and finally cause other problem such as using up the disk of some nodes and unable to serve.

![image](https://user-images.githubusercontent.com/9948422/110746447-ca332280-8277-11eb-8727-d1bbe52f852c.png)

PR #9897, opened 2021-03-25
raftstore: approximate split range evenly instead of against split size

Connor1996:
Signed-off-by: Connor <zbk602423539@gmail.com>

Issue Number: close https://github.com/tikv/tikv/issues/9785

Problem Summary:
When the write load is too heavy, the speed of the split may not catch up with the write speed leading to large region.  Before, we introduce the approximate split when `region_size > region_split_size * batch_split_limit * 2`, but it only split out batch_split_limit regions with region_split_size at most, so the large region would not be consumed at once. 

- If the duration of performing split command is not fast, it's possible the approximate split speed is far behind the write speed. To solve that, let approximate split the large region into batch_split_limit regions evenly. 
- Moreover, TiKV does split check for one region every 10s. Though the speed of performing split command is fast enough, it's still possible to become a large region with 10s. So it's better to split check more eagerly.

What's Changed:
- approximate split range evenly instead of against split size
- use approximate split when `region_size > region_split_size * batch_split_limit` instead of `region_split_size * batch_split_limit * 2`
- remove the `approximate_middle_key` and `divide_range` and reuse the common logic
- trigger split check immediately when the size is still larger than `region_max_size` after split

### Result
Before
截屏2021-03-29 下午2 43 01
After
image
With this PR, the large region can be consumed quickly, and eliminate the issue that follower can't catch up due to always stale snapshot)

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

- Approximate split range evenly


BusyJay:
How does it perform?

components/engine_rocks/src/range_properties.rs
```diff
@@ -379,9 +187,7 @@ impl RangePropertiesExt for RocksEngine {
             let len = keys.len();
             keys = keys.into_iter().step_by(len / 10000).collect();
```
Little-Wallace, line 188: I prefer to remove magic number.

components/engine_rocks/src/range_properties.rs
```diff
@@ -379,9 +187,7 @@ impl RangePropertiesExt for RocksEngine {
...
         }
-
         keys.sort();
-        keys.dedup();
```
Little-Wallace, line 384: Why remove dedup?

Connor1996:
> How does it perform?

will test it today

components/engine_rocks/src/range_properties.rs
Connor1996: Cause it isn't correct. Consider the case that multiple SSTs have the same key point, if you deduplicate them, the size of the range will be underestimated. Though, it's better to duplicate the split keys, will do.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2186,6 +2186,23 @@ where
         if is_leader {
             self.fsm.peer.approximate_size = estimated_size;
             self.fsm.peer.approximate_keys = estimated_keys;
+
+            if let Some(estimated_size) = self.fsm.peer.approximate_size {
```
Little-Wallace, line 2190: If some followers do not apply the `SplitRequest`, this split check task may fail because of the leader can not propose a `SplitRequest` in this term. Shall we change all `ApproximateSize` task to `SplitCheckTask` ? Because all peers will check split when they win an election and become leader.

Connor1996: this is called on the derived region, not on the newly split regions.

Connor1996:
@BusyJay the result is updated in the PR description. PTAL

Little-Wallace (approved):
LGTM

components/raftstore/src/coprocessor/split_check/half.rs
```diff
@@ -112,33 +112,12 @@ pub fn get_region_approximate_middle(
...
-) -> Result<Option<Vec<u8>>> {
-    let start_key = keys::enc_start_key(region);
-    let end_key = keys::enc_end_key(region);
-    let range = Range::new(&start_key, &end_key);
-    Ok(box_try!(db.get_range_approximate_middle_cf(
-        cfname,
-        range,
-        region.get_id()
+    Ok(box_try!(db.get_range_approximate_split_keys(range, 2).map(
+        |mut v| if v.is_empty() {
```
BusyJay, line 116: Why not just `v.pop()`?

components/raftstore/src/coprocessor/split_check/size.rs
```diff
@@ -205,47 +203,23 @@ pub fn get_region_approximate_size(
...
     let start_key = keys::enc_start_key(region);
     let end_key = keys::enc_end_key(region);
     let range = Range::new(&start_key, &end_key);
     Ok(box_try!(db.get_range_approximate_split_keys(
         range,
-        region.get_id(),
-        split_size,
-        max_size,
-        batch_split_limit
+        batch_split_limit as usize + 1
```
BusyJay, line 222: Very easy to make mistakes. I suggest to keep the same style for all functions.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2186,6 +2186,23 @@ where
...
+                if estimated_size > self.ctx.coprocessor_host.cfg.region_max_size.0 {
+                    info!(
+                        "trigger split check immediately";
+                        "region_id" => self.fsm.region_id(),
+                        "peer_id" => self.fsm.peer_id(),
+                        "size" => estimated_size,
+                    );
+                    self.fsm.peer.size_diff_hint = self.ctx.cfg.region_split_check_diff.0;
+                    // trigger a split check immediately when size is still large
+                    self.ctx
```
BusyJay, line 2200: Seems duplicated with split tick scheduled at the beginning of the function.

Connor1996: that will be scheduled after 10s, but this is intended to trigger split check at once.

BusyJay: Then the range will be scan twice.

Connor1996: The `size_diff_hint` will let the second split check skip. But it's indeed better to not schedule at the beginning if this condition is hit, will do.

src/server/debug.rs
```diff
@@ -1210,9 +1210,8 @@ fn divide_db(db: &Arc<DB>, parts: usize) -> raftstore::Result<Vec<Vec<u8>>> {
     let start = keys::data_key(b"");
     let end = keys::data_end_key(b"");
     let range = Range::new(&start, &end);
-    let region_id = 0;
     Ok(box_try!(
-        RocksEngine::from_db(db.clone()).divide_range(range, region_id, parts)
+        RocksEngine::from_db(db.clone()).get_range_approximate_split_keys(range, parts - 1)
```
BusyJay, line 1214: Why minus one?

Connor1996: it provides the number of the parts, but what the function need to specify is number of split key (equals to parts - 1). Still can not make all the same.

BusyJay (approved):

zhangjinpeng1987:
I think we should pick it to 5.0 release branch.

NingLin-P:
The issue may also affect release-4.0 branch, perhaps we should also cherry-pick this PR to release-4.0 /cc @Connor1996

Connor1996:
@NingLin-P Does any customer encounter this?

PR #9904, opened 2021-03-26
calculate snap dir size by scan

### LINKED ISSUE #9896: Store used size changes a lot after restarting

```
Release Version:   5.0.0-rc.x
Edition:           Community
Git Commit Hash:   34d72b137a3eeadbcb444613a32763e0cfa5c084
Git Commit Branch: heads/refs/tags/v5.0.0-nightly
UTC Build Time:    2021-03-17 03:10:24
Rust Version:      rustc 1.51.0-nightly (1d0d76f8d 2021-01-24)
```

Not related 

### Steps to reproduce

Run some workload and the used size will increase infinitely. And do a restart, the used size will be normal again.

### What did you expect?
the store used size is close to store size

### What did happened?
截屏2021-03-25 下午7 55 10

From PD metrics, store used size changes a lot after restarting which is reported from TiKV side.
TiKV calculates store used size by adding snapshot total size and engine used size.

截屏2021-03-25 下午7 55 27

image

From metrics, we can know that store size(engine used size) and disk space util doesn't changes a lot, **so it must be something wrong with snpashot total size.**

hicqu:

Issue Number: close #9896 .

Before the patch the snap directory size is tracked in memory. However it's hard to handle `mv *.sst.tmp *.sst` if the target path exists. Besides this maybe there are still other unknown bugs.

This PR calculates the snap directory size by scan. This operation is moved to the PD worker thread from raftstore threads. A benchmark in #9896 shows that the scan won't hurt performance too much.

### Related changes

### Check List 

Tests 

- Unit test
- Manual test
  With the patch store-used-size is very close to engine-size:
  
图片

  Without the patch store-used-size becomes far from engine-size after about 6 hours:
图片

### Release note 
* No release note.

### DIFF
--- a/components/raftstore/src/store/snap.rs
+++ b/components/raftstore/src/store/snap.rs
@@ -87,6 +87,12 @@ quick_error! {
     }
 }
 
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Error::Other(Box::new(e))
+    }
+}
+
 pub type Result<T> = result::Result<T, Error>;
 
 impl ErrorCodeExt for Error {
@@ -743,7 +749,6 @@ impl Snap {
                 // Use `kv_count` instead of file size to check empty files because encrypted sst files
                 // contain some metadata so their sizes will never be 0.
                 self.mgr.rename_tmp_cf_file_for_send(cf_file)?;
-                self.mgr.snap_size.fetch_add(cf_file.size, Ordering::SeqCst);
             } else {
                 delete_file_if_exist(&cf_file.tmp_path).unwrap();
             }
@@ -885,9 +890,7 @@ impl GenericSnapshot for Snap {
             }
 
             // Delete cf files.
-            if delete_file_if_exist(&cf_file.path).unwrap() {
-                self.mgr.snap_size.fetch_sub(cf_file.size, Ordering::SeqCst);
-            }
+            delete_file_if_exist(&cf_file.path).unwrap();
         }
         delete_file_if_exist(&self.meta_file.path).unwrap();
         if self.hold_tmp_files {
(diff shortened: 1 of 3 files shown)


components/raftstore/src/store/worker/pd.rs
```diff
@@ -670,8 +672,8 @@ where
         stats.set_capacity(capacity);
 
         // already include size of snapshot files
```
Connor1996, line 674: we can remove the comment now

components/raftstore/src/store/snap.rs
```diff
@@ -1453,8 +1453,18 @@ impl SnapManagerCore {
     fn get_total_snap_size(&self) -> Result<u64> {
         let mut total_size = 0;
         for entry in fs::read_dir(&self.base)? {
-            let entry = entry?;
-            let metadata = entry.metadata()?;
+            macro_rules! ignore_not_found {
+                ($op:expr) => {
+                    match $op {
+                        Err(e) if e.kind() == ErrorKind::NotFound => continue,
```
Connor1996, line 1459: what's the case that the entry is not found.

hicqu: `read_dir` and read entries is not atomic. It's possible that snapshots are deleted between them, in which case we can meet error `no such file or directory`.

Connor1996 (approved):
LGTM

components/raftstore/src/store/snap.rs
```diff
@@ -1448,6 +1450,34 @@ impl SnapManager {
...
+                        Err(e) => return Err(Error::from(e)),
+                        Ok(x) => x,
+                    }
+                };
+            }
+
+            let entry = ignore_not_found!(entry);
```
BusyJay, line 1466: How about
```
let res = entry.and_then(|e| (e, e.metadata()));
let (entry, metadata) = match res {
    ...
};
```

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/snap.rs
BusyJay: Or just put all body of the loop into a closure.

NingLin-P (approved):

nolouch:
Does release 3.0 exist this problem?

BusyJay:
Does it mean there is leak resource?

PR #9906, opened 2021-03-26
*: fix forward report and use more forward connections

BusyJay:

Problem Summary:

Too small value can lead to imbalance load.

The new panel looks like:
image

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

### Release note 
- no release note.


MyonKeminta (approved):
LGTM

youjiali1995 (commented):
Seems each interface has an individual `Proxy`, so the total connections count is interface count * `forward_max_connections_per_address`?

BusyJay:
No, grpc-rs tries to use only one connection if all connecting args are the same.

components/pd_client/src/util.rs
```diff
@@ -43,6 +43,38 @@ const MAX_RETRY_DURATION: Duration = Duration::from_secs(10);
...
 
+pub struct TargetInfo {
+    target_url: String,
+    via: String,
+}
+
+impl TargetInfo {
+    fn new(target_url: String, via: &str) -> TargetInfo {
+        TargetInfo {
+            target_url,
```
MyonKeminta, line 54: Is this trimmed?

BusyJay: It's not. And it's expected not to trim as PD only accepts url currently.

MyonKeminta (approved):
LGTM

youjiali1995 (approved):

PR #9912, opened 2021-03-28
coprv2: RawStorage implementation and request dispatch

Jibbow:

Implementation of `RawStorage` trait for `tikv::storageStorage` with corresponding error transformations.
Also, requests to coprocessor endpoint are now dispatched to the corresponding coprocessor.

Still open: collect version information of plugin and check version mismatch.

Issue Number: #9747

Problem Summary:

I implemented a `RawStorageImpl` that dispatches calls from a coprocessor plugin to `tikv::storage::Storage`.
We also do some error transformations and pass unrecoverable errors via `Box<dyn Any>`.

I have put some `TODO` comments into the code where I still have questions. I will also put comments into this PR with the corresponding lines of code.

What's Changed:

### Related changes

- based on PR #9802 [merged]

### Check List 

Tests 

- Unit test
- Missing: a full integration test, but I would first implement the ability to properly load a plugin.

### Release note 

 * No release note


src/coprocessor_v2/endpoint.rs
```diff
@@ -26,18 +34,74 @@ impl<E: Engine> Endpoint<E> {
...
+                .get_plugin(&req.copr_name)
+                .ok_or(CoprocessorError::Other(format!(
+                    "No registered coprocessor with name '{}'",
+                    req.copr_name
+                )))?;
+
+        let raw_storage_api = RawStorageImpl::new(req.get_context(), storage);
+        let region = Region {
+            id: req.get_context().get_region_id(),
+            start_key: vec![], // TODO how to get start_key and end_key?
```
Jibbow, line 72: I'm not sure how to get the key range of the current region.

src/coprocessor_v2/endpoint.rs
```diff
@@ -26,18 +34,74 @@ impl<E: Engine> Endpoint<E> {
...
+fn extract_region_error(error: &PluginError) -> Option<kvproto::errorpb::Error> {
+    match error {
+        PluginError::StorageError(storage_err) => {
+            match storage_err {
+                coprocessor_plugin_api::StorageError::Other(other_err) => other_err
+                    .downcast_ref::<storage::Result<()>>()
+                    .map_or(None, |e| storage::errors::extract_region_error::<()>(e)),
+                // TODO: KeyNotInRegion is not set in https://tikv.github.io/doc/src/tikv/storage/errors.rs.html#193-233
+                // so I assume we should also not set it?
+                coprocessor_plugin_api::StorageError::KeyNotInRegion { .. } => None,
```
Jibbow, line 101: I'm not entirely sure which errors we should pass back to the client as proper errors and which ones as "error messages".

There is this function `storage::errors::extract_region_error()` that is used to set specific errors from TiKV when returning the result to the client and other errors are just converted to an error message. 
For "normal" `TiKVClient::raw_put()`, ... requests, `KeyNotInRegion` is converted to a string and only the error message is returned. I do the same here, but I was not sure about this.

components/coprocessor_plugin_api/src/plugin_api.rs
```diff
@@ -0,0 +1,71 @@
...
+pub enum PluginError {
+    StorageError(StorageError),
+    /// Field for custom errors that appeared while executing a coprocessor request.
+    ///
+    /// Errors that are reported to the client via `PluginError::Other` should be of a type were
+    /// the client can't handle them. Thus, only the error's message is sent to the client.
+    /// If the error can be handled by the client, coprocessor plugins should perform proper error
+    /// reporting via their `data` fields.
+    /// TODO: Should we even include "non-recoverable user errors"?
+    Other(String),
```
Jibbow, line 22: I added a dedicated field for unknown user errors that originate in a coprocessor plugin, but I'm not sure if we need it or if error handling should exclusively be done via `data` (encoded response from plugin).

src/coprocessor_v2/raw_storage_impl.rs
```diff
@@ -0,0 +1,320 @@
...
+    fn from(error: storage::errors::Error) -> Self {
+        let inner = match *error.0 {
+            // Key not in region
+            storage::errors::ErrorInner::Engine(EngineError(box EngineErrorInner::Request(
+                ref req_err,
+            ))) if req_err.has_key_not_in_region() => {
+                let key_err = req_err.get_key_not_in_region();
+                StorageError::KeyNotInRegion {
+                    key: key_err.get_key().to_owned(),
+                    region: todo!(), // TODO: how to construct region here? We only have region_id
```
Jibbow, line 182: I'm not sure how we can construct a `Region` struct here because `StorageError::KeyNotInRegion` only has the `region_id`. How can we resolve a `region_id` to more information?

src/coprocessor_v2/raw_storage_impl.rs
```diff
@@ -0,0 +1,320 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::storage::{lock_manager::DummyLockManager, TestStorageBuilder};
+    use futures::executor::block_on;
+    use kvproto::kvrpcpb::Context;
+
+    #[test]
+    fn test_storage_api() {
```
skyzh, line 212: You could use `#[tokio::test]` to avoid writing `block_on`. Refer to https://docs.rs/tokio/0.2.25/tokio/attr.test.html for more information.

skyzh (approved):
Looks good to me overall (apart from Andy's comments). Good work!

src/coprocessor_v2/raw_storage_impl.rs
Jibbow: Ah nice, didn't know that! Makes the code much cleaner :)

Jibbow:
@andylokandy @skyzh changes incorporated. Requesting a new review if you have some time to spare :)

skyzh (approved):

andylokandy (approved):

Issue #9913, opened 2021-03-29
tikv keep panicking at components/engine_rocks/src/engine_iterator.rs:39 


hicqu:
### What version of TiKV are you using?
Master.

### What operating system and CPU are you using?
Doesn't matter.

### Steps to reproduce
Not sure.

### What did you expect?
TiKV keeps running without panicking.

### What did happened?
TiKV panics:
```
[...]

PR #9915, opened 2021-03-29
Delay iterator panic to data access

tabokie:

Mitigate #9913 

### Check List 

Tests

- Unit test

### Release note
- No release note


tabokie:
It's an UB to call `Next` on an invalid cursor, but shipping this check to production user might also cause repercussions more than we intended.

yiwu-arbug (commented):
what does "nort" stands for?

tabokie:
> what does "nort" stands for?

"no runtime"

yiwu-arbug (approved):

Connor1996 (commented):
Don't think it's a good idea. This just covers up the issue.

tabokie:
This PR is not well thought to capture bugs, just to restore things as they were.
A better solution would be:
1) In `next()`/`prev()`, check `valid()` and return error if not
2) In `key()`/`value()`, assert `valid()`

This way we don't have to panic until data integrity is actually at risk.
@yiwu-arbug @Connor1996

Connor1996:
> This PR is not well thought to capture bugs, just to restore things as they were.
> A better solution would be:
> 
> 1. In `next()`/`prev()`, check `valid()` and return error if not
> 2. In `key()`/`value()`, assert `valid()`
> 
> This way we don't have to panic until data integrity is actually at risk.
> @yiwu-arbug @Connor1996

But calling key() or value() when it is invalid is undefined behavior. It may return the wrong data to the user.

tabokie:
> > This PR is not well thought to capture bugs, just to restore things as they were.
> > A better solution would be:
> > 
> > 1. In `next()`/`prev()`, check `valid()` and return error if not
> > 2. In `key()`/`value()`, assert `valid()`
> > 
> > This way we don't have to panic until data integrity is actually at risk.
> > @yiwu-arbug @Connor1996
> 
> But calling key() or value() when it is invalid is undefined behavior. It may return the wrong data to the user.

That's where we will panic.

Connor1996:
> > > This PR is not well thought to capture bugs, just to restore things as they were.
> > > A better solution would be:
> > > 
> > > 1. In `next()`/`prev()`, check `valid()` and return error if not
> > > 2. In `key()`/`value()`, assert `valid()`
> > > 
> > > This way we don't have to panic until data integrity is actually at risk.
> > > @yiwu-arbug @Connor1996
> > 
> > 
> > But calling key() or value() when it is invalid is undefined behavior. It may return the wrong data to the user.
> 
> That's where we will panic.

But it wouldn't panic in release build.

tabokie:
> > > > This PR is not well thought to capture bugs, just to restore things as they were.
> > > > A better solution would be:
> > > > 
> > > > 1. In `next()`/`prev()`, check `valid()` and return error if not
> > > > 2. In `key()`/`value()`, assert `valid()`
> > > > 
> > > > This way we don't have to panic until data integrity is actually at risk.
> > > > @yiwu-arbug @Connor1996
> > > 
> > > 
> > > But calling key() or value() when it is invalid is undefined behavior. It may return the wrong data to the user.
> > 
> > 
> > That's where we will panic.
> 
> But it wouldn't panic in release build.

Need to re-enable nortcheck if we can all agree on this solution.

Connor1996 (approved):

yiwu-arbug (commented):
I think a rustful approach would be to have `next` (and `prev` and others) to return a result type, and caller can unwrap key and value from that type. But that will be a larger refactor. For now I'm okay just getting around the error somehow.
```
trait IterResult {
    fn key(&self) -> &[u8],
    fn value(&self) -> &[u8],
}
fn next(&mut self) -> Result<IterResult> {...}
```

yiwu-arbug:
> I think a rustful approach would be to have `next` (and `prev` and others) to return a result type, and caller can unwrap key and value from that type. But that will be a larger refactor. For now I'm okay just getting around the error somehow.
> 
> ```
> trait IterResult {
>     fn key(&self) -> &[u8],
>     fn value(&self) -> &[u8],
> }
> fn next(&mut self) -> Result<IterResult> {...}
> ```

After a second thought I think this won't work, because it fail to invalidate the `IterResult` after the subseqent `iter.next()` is called. I think a simpler approach would be to have `key()` and `value()` return `Result<&[u8]>`.

PR #9917, opened 2021-03-29
batch-system: use different default max batch size between master and release branch

gengliqi:

Problem Summary:
On master, the hibernate region is enabled by default so the max-batch-size can be larger.
However it is disabled by default on the release branch, the max-batch-size is too small which may cause many problems, e.g. too small write batch, hunger problem.
You can also see https://github.com/tikv/tikv/pull/9020 for more details.

What's Changed:
use different default max batch size between master and release branch.

### Related changes

Tests 

- No code

Side effects

- No

### Release note 
- If `hibernate-regions` is enabled, the default value of `store-max-batch-size` is 256, otherwise it's 1024.


components/batch-system/src/config.rs
```diff
@@ -13,8 +13,15 @@ pub struct Config {
 
 impl Default for Config {
     fn default() -> Config {
+        // FIXME: if hibernate regions is enabled by default on release branch, we can use
+        // the same configuration here.
+        let max_batch_size = if tikv_util::build_on_master_branch() {
```
BusyJay, line 18: How about changing it at `check` method according to the value of `hibernate-region`?

gengliqi: What if users set the `max-batch-size`?  Use a `Option`?

BusyJay: Yes.

components/test_raftstore/src/util.rs
```diff
@@ -132,7 +132,8 @@ lazy_static! {
     static ref TEST_CONFIG: TiKvConfig = {
         let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
         let common_test_cfg = manifest_dir.join("src/common-test.toml");
-        TiKvConfig::from_file(&common_test_cfg, None)
+        let mut cfg = TiKvConfig::from_file(&common_test_cfg, None);
```
BusyJay, line 135: What's the change?

components/raftstore/src/store/config.rs
```diff
@@ -718,13 +728,21 @@ mod tests {
...
+        cfg.apply_batch_system.max_batch_size = Some(0);
         assert!(cfg.validate().is_err());
 
         cfg = Config::new();
         cfg.apply_batch_system.pool_size = 0;
         assert!(cfg.validate().is_err());
 
+        cfg = Config::new();
+        cfg.store_batch_system.max_batch_size = Some(0);
+        assert!(cfg.validate().is_err());
```
BusyJay, line 740: Missing a case to check config is updated according to hibernate_region.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

NingLin-P (approved):

PR #9925, opened 2021-03-30
coprv2: hot-reloading of plugins

Jibbow:

Implements hot-reloading for coprocessor plugins, similar to the PoC.
Also adds some more public methods for `PluginRegistry` to deal with loaded plugins, e.g. the ability to unload plugins.

Again, I added some questions in `TODO` comments.

Issue Number: #9747 

Problem Summary:

The coprocessor plugin framework should watch a "plugin directory" and automatically load/unload plugins that are created/deleted in the directory.

What's Changed:

### Related changes

- based on PR #9912 

### Check List 

Tests 

- Unit test

### Release note 

 * No release notes

### DIFF
--- a/src/coprocessor_v2/plugin_registry.rs
+++ b/src/coprocessor_v2/plugin_registry.rs
@@ -357,126 +357,126 @@ fn is_library_file<P: AsRef<Path>>(path: P) -> bool {
     path.as_ref().extension() == Some(OsStr::new("dll"))
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use coprocessor_plugin_api::pkgname_to_libname;
-
-    #[test]
-    fn load_plugin() {
-        let library_path = pkgname_to_libname("example-plugin");
-        let loaded_plugin = unsafe { LoadedPlugin::new(&library_path).unwrap() };
-
-        assert_eq!(loaded_plugin.name(), "example-plugin");
-    }
-
-    #[test]
-    fn registry_load_and_get_plugin() {
-        let registry = PluginRegistry::new();
-        let library_path = pkgname_to_libname("example-plugin");
-        let plugin_name = registry.load_plugin(&library_path).unwrap();
-
-        let plugin = registry.get_plugin(plugin_name).unwrap();
-
-        assert_eq!(plugin.name(), "example-plugin");
-        assert_eq!(registry.loaded_plugin_names(), vec!["example-plugin"]);
-        assert_eq!(
-            registry
-                .get_path_for_plugin("example-plugin")
-                .unwrap()
-                .to_string_lossy(),
-            library_path
-        );
-    }
-
-    #[test]
-    fn update_plugin_path() {
-        let registry = PluginRegistry::new();
-        let library_path = pkgname_to_libname("example-plugin");
-        let library_path_2 = pkgname_to_libname("example-plugin-2");
-
-        let plugin_name = registry.load_plugin(&library_path).unwrap();
-
-        assert_eq!(
-            registry
-                .get_path_for_plugin(plugin_name)
-                .unwrap()
-                .to_str()
-                .unwrap(),
-            &library_path
-        );
-
-        registry.update_plugin_path(plugin_name, &library_path_2);
-
-        assert_eq!(
-            registry
-                .get_path_for_plugin(plugin_name)
-                .unwrap()
-                .to_str()
-                .unwrap(),
-            &library_path_2
-        );
-    }
-
-    #[test]
-    fn registry_unload_plugin() {
-        let registry = PluginRegistry::new();
-        let library_path = pkgname_to_libname("example-plugin");
-
-        let plugin_name = registry.load_plugin(&library_path).unwrap();
-
-        assert!(registry.get_plugin(plugin_name).is_some());
-
-        registry.unload_plugin(plugin_name);
-
-        assert!(registry.get_plugin(plugin_name).is_none());
-        assert_eq!(registry.loaded_plugin_names().len(), 0);
-    }
-
-    #[test]
-    fn plugin_registry_hot_reloading() {
-        let build_dir = std::env::current_exe()
-            .map(|p| p.as_path().parent().unwrap().to_owned())
-            .unwrap();
-        let coprocessor_dir = build_dir.join("coprocessors");
-        let library_path = coprocessor_dir.join(pkgname_to_libname("example-plugin"));
-        let library_path_2 = coprocessor_dir.join(pkgname_to_libname("example-plugin-2"));
-        let plugin_name = "example-plugin";
-
-        std::fs::remove_dir_all(&coprocessor_dir).unwrap();
-        std::fs::create_dir_all(&coprocessor_dir).unwrap();
-
-        let mut registry = PluginRegistry::new();
-        registry.start_hot_reloading(&coprocessor_dir).unwrap();
-
-        // trigger loading
-        std::fs::copy(
-            build_dir.join(&pkgname_to_libname("example-plugin")),
-            &library_path,
-        )
-        .unwrap();
-        std::thread::sleep(Duration::from_secs(4));
-
-        assert!(registry.get_plugin(plugin_name).is_some());
-        assert_eq!(
-            &PathBuf::from(registry.get_path_for_plugin(plugin_name).unwrap()),
-            &library_path
-        );
-
-        // trigger rename
-        std::fs::rename(&library_path, &library_path_2).unwrap();
-        std::thread::sleep(Duration::from_secs(4));
-
-        assert!(registry.get_plugin(plugin_name).is_some());
-        assert_eq!(
-            &PathBuf::from(registry.get_path_for_plugin(plugin_name).unwrap()),
-            &library_path_2
-        );
-
-        // trigger unloading
-        std::fs::remove_file(&library_path_2).unwrap();
-        std::thread::sleep(Duration::from_secs(4));
-
-        assert!(registry.get_plugin(plugin_name).is_none());
-    }
-}
+//#[cfg(test)]
+//mod tests {
+//    use super::*;
+//    use coprocessor_plugin_api::pkgname_to_libname;
+//
+//    #[test]
+//    fn load_plugin() {
+//        let library_path = pkgname_to_libname("example-plugin");
+//        let loaded_plugin = unsafe { LoadedPlugin::new(&library_path).unwrap() };
+//
+//        assert_eq!(loaded_plugin.name(), "example-plugin");
+//    }
+//
+//    #[test]
+//    fn registry_load_and_get_plugin() {
+//        let registry = PluginRegistry::new();
+//        let library_path = pkgname_to_libname("example-plugin");
+//        let plugin_name = registry.load_plugin(&library_path).unwrap();
+//
+//        let plugin = registry.get_plugin(plugin_name).unwrap();
+//
+//        assert_eq!(plugin.name(), "example-plugin");
+//        assert_eq!(registry.loaded_plugin_names(), vec!["example-plugin"]);
+//        assert_eq!(
+//            registry
+//                .get_path_for_plugin("example-plugin")
+//                .unwrap()
+//                .to_string_lossy(),
+//            library_path
+//        );
+//    }
+//
+//    #[test]
+//    fn update_plugin_path() {
+//        let registry = PluginRegistry::new();
+//        let library_path = pkgname_to_libname("example-plugin");
+//        let library_path_2 = pkgname_to_libname("example-plugin-2");
+//
+//        let plugin_name = registry.load_plugin(&library_path).unwrap();
+//
+//        assert_eq!(
+//            registry
+//                .get_path_for_plugin(plugin_name)
+//                .unwrap()
+//                .to_str()
+//                .unwrap(),
+//            &library_path
+//        );
+//
+//        registry.update_plugin_path(plugin_name, &library_path_2);
+//
+//        assert_eq!(
+//            registry
+//                .get_path_for_plugin(plugin_name)
+//                .unwrap()
+//                .to_str()
+//                .unwrap(),
+//            &library_path_2
+//        );
+//    }
+//
+//    #[test]
+//    fn registry_unload_plugin() {
+//        let registry = PluginRegistry::new();
+//        let library_path = pkgname_to_libname("example-plugin");
+//
+//        let plugin_name = registry.load_plugin(&library_path).unwrap();
+//
+//        assert!(registry.get_plugin(plugin_name).is_some());
+//
+//        registry.unload_plugin(plugin_name);
+//
+//        assert!(registry.get_plugin(plugin_name).is_none());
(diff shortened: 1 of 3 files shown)


src/coprocessor_v2/plugin_registry.rs
```diff
@@ -2,28 +2,119 @@
...
     }
 
     /// Finds a plugin by its name. The plugin must have been loaded before with [`load_plugin()`].
     ///
     /// Plugins are indexed by the name that is returned by [`CoprocessorPlugin::name()`].
-    pub fn get_plugin(&self, plugin_name: &str) -> Option<&impl CoprocessorPlugin> {
-        self.loaded_plugins.get(plugin_name)
+    // TODO: takes &self and returns Arc<_> --> allows parallel execution of plugins. Is this okay?
+    // TODO: I guess it should be okay, because `CoprocessorPlugin::on_raw_coprocessor_request()`
+    // TODO: also takes &self and thus should be concurrently usable.
```
skyzh, line 103: This is okay to use.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -33,28 +124,108 @@ impl PluginRegistry {
...
+    fn default() -> Self {
+        PluginRegistry::new()
+    }
+}
+
+#[derive(Default)]
+struct PluginRegistryInner {
+    /// Plugins that are currently loaded.
+    /// Provides a mapping from the plugin's name to the actual instance.
+    // TODO: is `Arc` here okay? Intention is that we can concurrently manipulate (load/unload)
```
skyzh, line 164: Also okay.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -78,7 +250,15 @@ impl LoadedPlugin {
         // Leak library so that we will never drop it
         std::mem::forget(lib);
```
skyzh, line 251: Current approach looks good to me, that we never drop loaded libraries. I have an idea about this: if there is no reference to `LoadedPlugin`, can we ensure that no code of plugin is running? If so, we could safely drop this lib upon last instance of `Arc<LoadedPlugin>` is dropped.

skyzh (commented):
Having a quick glance on this patch, I think it basically looks good to me. Good work! I'll come back and approve this PR after previous PRs have been merged.

src/coprocessor_v2/plugin_registry.rs
andylokandy: > if there is no reference to LoadedPlugin, can we ensure that no code of plugin is running

No, we can't. `'static` reference from the library will be leaked into the runtime even no borrow rule is violated.

Jibbow:
Not sure why, but the test`metrics::threads_linux::tests::test_thread_io_statistics` is failing.

skyzh:
> Not sure why, but the test `metrics::threads_linux::tests::test_thread_io_statistics` is failing.

That could be a temporary test failure. We may re-run the test later.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -2,28 +2,116 @@
...
-    /// Provides a mapping from the plugin's name to the actual instance.
-    loaded_plugins: BTreeMap<String, LoadedPlugin>,
+    inner: Arc<RwLock<PluginRegistryInner>>,
+    /// Active file system watcher for hot-reloading.
+    /// If dropped, the corresponding hot-reloading thread will stop.
+    fs_watcher: notify::RecommendedWatcher,
 }
 
 #[allow(dead_code)]
 impl PluginRegistry {
```
andylokandy, line 22: I think it will be better if we are able to provide a config switch for turning off the hot reload but still load plugins in the directory on startup. So I'd like to swap the `new()` and `start_hot_reload()`, in other words, when `new()` (or named `load_plugin()` or any more appropriate) the plugin will always be loaded; and when `start_hot_reload()` the watcher and thread will be created.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -2,28 +2,116 @@
...
+
+        // Create a file system watcher and background thread for hot-reloading.
+        let (tx, rx) = mpsc::channel();
+        let fs_watcher = notify::watcher(tx, Duration::from_secs(3)).unwrap();
+
+        let hot_reload_registry = registry.clone();
+        thread::spawn(move || {
+            loop {
+                match rx.recv() {
+                    Ok(DebouncedEvent::Create(file)) => {
```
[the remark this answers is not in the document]
Jibbow: Oh yes, good catch, thanks! I didn't think of that

src/coprocessor_v2/plugin_registry.rs
Jibbow: Thanks for the review! I'm not sure if I understood this correctly. Could you have a look at https://github.com/tikv/tikv/pull/9925/commits/f95fbc29d8682ac552fc9faf4dc2ef0e019a4137? Now the fs watcher and background thread will only be created after a call to `start_hot_reloading()` and not in `new()`. Also, I added another function `load_plugins_from_dir()` which attempts to load all plugins from a directory.
I still settled with just using for fs watcher and one background thread, even with multiple calls to `start_hot_reloading()` (in contrast to the POC where a thread was created for every directory that is being watched).

Now that `new()` has no side effects, is a `Default` implementation valid again?

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -2,28 +2,128 @@
...
+            let (tx, rx) = mpsc::channel();
+            let fs_watcher = notify::watcher(tx, Duration::from_secs(3)).unwrap();
+
+            let hot_reload_registry = self.inner.clone();
+            thread::spawn(move || {
+                // Simple helper functions for loading/unloading plugins.
+                let maybe_load = |file: &PathBuf| {
+                    if is_library_file(&file) {
+                        // Discard error.
+                        let _ = hot_reload_registry.write().unwrap().load_plugin(file);
```
andylokandy, line 62: Maybe better log it.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -33,28 +133,134 @@ impl PluginRegistry {
...
+        dir_name: impl Into<PathBuf>,
+    ) -> std::io::Result<Vec<&'static str>> {
+        let dir_name = dir_name.into();
+        let mut loaded_plugins = Vec::new();
+
+        for entry in std::fs::read_dir(&dir_name)? {
+            if let Ok(file) = entry.map(|f| f.path()) {
+                if is_library_file(&file) {
+                    // Ignore errors.
+                    let r = self.load_plugin(&file);
```
andylokandy, line 158: ditto

src/coprocessor_v2/plugin_registry.rs
andylokandy: Excellent! That's a good approach. I have no further option on this.

> Now that new() has no side effects, is a Default implementation valid again?

My preference on whether to implement `Default` or not depends on (1) is there any possible place `T: Default` is required on my type? (2) is it zero cost (no side effect, no allocation)?

In this case, it news an `Arc`, so I personally will not impl `Default`.

andylokandy (approved):

src/coprocessor_v2/plugin_registry.rs
andylokandy: Excellent! That's a good approach. I have no further option on this.

Now that new() has no side effects, is a Default implementation valid again?

My preference on whether to implement Default or not depends on (1) is there any possible place T: Default is required on my type? (2) is it zero cost (no side effect, no allocation)?

In this case, it news an Arc, so I personally will not impl Default.

Jibbow: Alright, thanks! I'm never really sure when to implement `Default`, so thanks for the opinion :)

src/coprocessor_v2/plugin_registry.rs
Jibbow: It's logged with `warn!`, but inside `load_plugin()`. Wasn't sure about where to place the logging. Is this okay?

src/coprocessor_v2/plugin_registry.rs
Jibbow: same as before: 
> It's logged with `warn!`, but inside `load_plugin()`.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -2,28 +2,128 @@
...
+                        }
+                        Ok(DebouncedEvent::Remove(file)) => {
+                            unload(&file);
+                        }
+                        Ok(DebouncedEvent::Write(file)) => {
+                            unload(&file);
+                            maybe_load(&file);
+                        }
+                        Ok(DebouncedEvent::Rename(old_file, new_file)) => {
+                            unload(&old_file);
```
skyzh, line 85: If a plugin has been renamed within directory, do we need to reload it?

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -98,28 +313,102 @@ impl CoprocessorPlugin for LoadedPlugin {
...
+//        let mut registry = PluginRegistry::new();
+//        registry.start_hot_reloading(&coprocessor_dir).unwrap();
+//
+//        // trigger loading
+//        std::fs::copy(
+//            build_dir.join(&library_path),
+//            coprocessor_dir.join(&library_path),
+//        )
+//        .unwrap();
+//        std::thread::sleep(Duration::from_secs(4));
```
skyzh, line 404: From my experience, tests involving non-deterministic behaviors (like waiting for x seconds) have a high possibility of random failing on CI. We could see if this test works later when we run this on CI.

skyzh (approved):
Rest LGTM

src/coprocessor_v2/plugin_registry.rs
Jibbow: I think we could get around it by just updating the associated path of the plugin. But we need to do *something* because We store the path of the plugin so we can later unload them when they are deleted. I unloaded and loaded it again because it was simpler to do in the code, but we could also update the saved path. What do you think @skyzh and @andylokandy?

src/coprocessor_v2/plugin_registry.rs
Jibbow: Yeah, I thought the same thing :see_no_evil: 
But on the other hand I think we can't really get around it? Because the file system watcher debounces events with a delay of 3s...

Jibbow: Maybe we _could_ implement the [`Watcher`](https://docs.rs/notify/4.0.15/notify/trait.Watcher.html) with our own watcher where we can manually trigger events? Maybe that's a good way to go?

src/coprocessor_v2/plugin_registry.rs
andylokandy: alright, got it. If so, the convention to discard the error is to call `.ok()` on the result.

src/coprocessor_v2/plugin_registry.rs
andylokandy: Do you think the updating will be complex? If not, we could have a try.

src/coprocessor_v2/plugin_registry.rs
skyzh: We could later decide on how to handle this. If this case randomly fails, the bot would automatically open up an issue.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -2,28 +2,128 @@
...
+                loop {
+                    match rx.recv() {
+                        Ok(DebouncedEvent::Create(file)) => {
+                            maybe_load(&file);
+                        }
+                        Ok(DebouncedEvent::Remove(file)) => {
+                            unload(&file);
+                        }
+                        Ok(DebouncedEvent::Write(file)) => {
+                            unload(&file);
```
skyzh, line 81: Will modifying content of an opened executable or dynamic library lead to undefined behavior? Maybe we should warn our users for this case.

andylokandy: On Linux, an executable can even be removed without affecting the process. And on Windows, you will be forbidden from modifying it.

skyzh: Removing / recreating won't cause issue. But rewriting would. https://unix.stackexchange.com/questions/404551/overwriting-a-running-executable-or-so

src/coprocessor_v2/plugin_registry.rs
Jibbow: Will do this today. Should be easy :)

src/coprocessor_v2/plugin_registry.rs
Jibbow: hm good point. but does it matter when we fully unload and re-load it after the write is done?

skyzh: We could not determine whether the plugin has *fully* unloaded. There might be some futures running in tokio.

src/coprocessor_v2/plugin_registry.rs
Jibbow: Okay, it was a bit more work than expected, but only because I moved the `OsString` from `LoadedPlugin` into `PluginRegistryInner`, which is nicer in my opinion. Could you have another look please?

src/coprocessor_v2/plugin_registry.rs
Jibbow: hm I see...  good point.
but not sure what we can do about it? the only thing we could do is log a warning when a plugin is overwritten on the file system. or do you have any other ideas?

skyzh: I think logging a warning to our user is enough.

(By the way, I came with another possible fault: user moves the plugin outside of directory and rewrites its content, this is something we could not detect. We just do what we can, to only handle the case that plugin is rewritten inside the plugin directory and print a warning by `eprintln!` or `warn!`.)

Jibbow: I just tried out what happens when we use `std::fs::rename` where the target is in a different directory. `notify` actually issues a `Remove` instead of a `Rename`, so we will just unload the plugin because it was basically moved outside the plugin directory. I think this behavior is fine. :)

Jibbow: @skyzh added a warning message when a plugin is overwritten :) PTAL

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -84,13 +84,16 @@ impl PluginRegistry {
                         Ok(DebouncedEvent::Remove(file)) => {
                             unload(&file);
                         }
+                        Ok(DebouncedEvent::Rename(old_file, new_file)) => {
+                            // If the file is renamed with a different parent directory, we will receive a `Remove` instead.
+                            std::debug_assert!(old_file.parent() == new_file.parent());
```
andylokandy, line 89: I think `debug_assert!` is in prelude so no need to use fully-qualified path.

PR #9885, opened 2021-03-24
cdc: flow control

liuzix:
### What problem does this PR solve?

Problem Summary:

TiCDC will cause TiKV to OOM if the generated changelog cannot be sunk to grpc fast enough. This usually happens with incremental scans, where a large amount of historical data is being rapidly read into memory. OOM can also happen if the real-time data rate is too high for the network (or the TiCDC golang part) to process.

This PR allows TiKV to function normally with TiCDC 5.0.x, which has Unified Sorter enabled and thus capable of handling large incremental scans.

### What is changed and how it works?

What's Changed:

- Now all CDC data is sent through a specially designed `RateLimiter`, which enforces the following policy: (1) the pending queue length should not be larger than `thres_close`, or otherwise the sink is closed immediately, which is the most effective way of releasing all resources related to the current CDC stream; (2) if the pending queue is longer than `thres_block_scan`, incremental scan is blocked, until the queue becomes shorter than `thres_block_scan`. This allows back-pressure propagation from grpc.
- Added a mechanism which allows incremental scan to continue after a region is split/merged. This prevents incremental scan from being interrupted by writing hotspots. The optimization is necessary for CDC to be usable under heavy writes. 

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

### Release note 

- Fix OOM problem when used with TiCDC

PR #9932, opened 2021-03-31
cdc: disable hibernate regions compactible by default

overvenus:

Problem Summary:

This is part of #9885, we cherry-pick the change in case #9885 does not get merged in time.

### Check List 

Tests 

- Unit test
- Integration test

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

- No release note

### DIFF
--- a/components/cdc/tests/integrations/test_cdc.rs
+++ b/components/cdc/tests/integrations/test_cdc.rs
@@ -1702,7 +1702,8 @@ fn test_cdc_write_rollback_when_no_lock() {
 
 #[test]
 fn test_resolved_ts_cluster_upgrading() {
-    let cluster = new_server_cluster(0, 3);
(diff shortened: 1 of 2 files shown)


components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -1702,7 +1702,8 @@ fn test_cdc_write_rollback_when_no_lock() {
 
 #[test]
 fn test_resolved_ts_cluster_upgrading() {
-    let cluster = new_server_cluster(0, 3);
+    let mut cluster = new_server_cluster(0, 3);
+    cluster.cfg.cdc.hibernate_regions_compatible = true;
```
5kbpers, line 1706: How about enabling it for all the tests?

PR #9938, opened 2021-03-31
encryption: Ignore log record parse error caused by write failure before

### LINKED ISSUE #9886: tikv can't startup after the disk is full
v4.0.11
centos-3.10.0-1127.el7.x86_64
### What did you expect?
tikv can start up without any panic
### What did happened?
here is some logs of tikv：
```
[2021/03/24 08:49:24.697 +00:00] [FATAL] [lib.rs:482] ["called `Result::unwrap()` on an `Err` value: Other(\"[components/encryption/src/file_dict_file.rs:330]: file corrupted! crc32 mismatch 1448223484 != 1104170698\")"] [backtrace="stack backtrace:\n   0: tikv_util::set_panic_hook::{{closure}}\n             at components/tikv_util/src/lib.rs:481\n   1: std::panicking::rust_panic_with_hook\n             at src/libstd/panicking.rs:475\n   2: rust_begin_unwind\n             at src/libstd/panicking.rs:375\n   3: core::panicking::panic_fmt\n             at src/libcore/panicking.rs:84\n   4: core::result::unwrap_failed\n             at src/libcore/result.rs:1188\n   5: core::result::Result<T,E>::unwrap\n             at rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/result.rs:956\n      cmd::server::TiKVServer::init_encryption\n             at cmd/src/server.rs:324\n      cmd::server::run_tikv\n             at cmd/src/server.rs:97\n   6: tikv_server::main\n
   at cmd/src/bin/tikv-server.rs:166\n   7: std::rt::lang_start::{{closure}}\n             at rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/rt.rs:67\n   8: main\n   9: __libc_start_main\n  10: <unknown>\n"] [location=src/libcore/result.rs:1188] [thread_name=main]
```

hunterlxt:

Issue Number: close #9886 

Problem Summary:

When the file dict fails to be appended and written for some reason, we believe that this record can be safely discarded because the corresponding file has not been actually modified

Add a new error type and ignore these errors.

### Related changes

- Need to cherry-pick to the release-4.x

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

N/A

### Release note 
- Fix the bug that TiKV cannot startup when the end of file dict file is damaged.

### DIFF
--- a/components/encryption/src/errors.rs
+++ b/components/encryption/src/errors.rs
@@ -19,7 +19,7 @@ pub enum Error {
     Other(#[from] Box<dyn error::Error + Sync + Send>),
     // Only when the parsing record is the last one, and the
     // length is insufficient or the crc checksum fails.
-    #[error("Recoverable types of parsing file dict error")]
+    #[error("Recoverable tail record corruption while parsing file dictionary")]
     TailRecordParseIncomplete,
     // Currently only in use by cloud KMS
     #[error("Cloud KMS error {0}")]
--- a/components/encryption/src/errors.rs
+++ b/components/encryption/src/errors.rs
@@ -17,6 +17,10 @@ pub trait RetryCodedError: Debug + Display + ErrorCodeExt + RetryError + Send + 
 pub enum Error {
     #[error("Other error {0}")]
     Other(#[from] Box<dyn error::Error + Sync + Send>),
+    // Only when the parsing record is the last one, and the
+    // length is insufficient or the crc checksum fails.
+    #[error("Recoverable types of parsing file dict error")]
+    TailRecordParseIncomplete,
     // Currently only in use by cloud KMS
     #[error("Cloud KMS error {0}")]
     RetryCodedError(Box<dyn RetryCodedError>),
@@ -70,6 +74,7 @@ impl ErrorCodeExt for Error {
     fn error_code(&self) -> ErrorCode {
         match self {
             Error::RetryCodedError(err) => (*err).error_code(),
+            Error::TailRecordParseIncomplete => error_code::encryption::PARSE_INCOMPLETE,
             Error::Rocks(_) => error_code::encryption::ROCKS,
             Error::Io(_) => error_code::encryption::IO,
             Error::Crypter(_) => error_code::encryption::CRYPTER,
@@ -88,6 +93,7 @@ impl RetryError for Error {
         // However, only Error::Tls should be encountered
         match self {
             Error::RetryCodedError(err) => err.is_retryable(),
+            Error::TailRecordParseIncomplete => true,
             Error::Rocks(_) => true,
             Error::Io(_) => true,
             Error::Crypter(_) => true,
--- a/components/encryption/src/file_dict_file.rs
+++ b/components/encryption/src/file_dict_file.rs
@@ -5,7 +5,7 @@ use file_system::{rename, File, OpenOptions};
 use kvproto::encryptionpb::{EncryptedContent, FileDictionary, FileInfo};
 use protobuf::Message;
 use rand::{thread_rng, RngCore};
-use tikv_util::{box_err, warn};
+use tikv_util::{box_err, info, set_panic_mark, warn};
 
 use crate::encrypted_file::{EncryptedFile, Header, Version, TMP_FILE_SUFFIX};
 use crate::master_key::{Backend, PlaintextBackend};
@@ -170,12 +170,14 @@ impl FileDictionaryFile {
             }
             Version::V2 => {
                 file_dict.merge_from_bytes(content)?;
+                let mut last_record_name = String::new();
                 // parse the remained records
                 while !remained.is_empty() {
                     // If the parsing get errors, manual intervention is required to recover.
                     match Self::parse_next_record(remained) {
                         Ok((used_size, file_name, mode)) => {
                             remained.consume(used_size);
+                            last_record_name = file_name.clone();
                             match mode {
                                 LogRecord::INSERT(info) => {
                                     file_dict.files.insert(file_name, info);
@@ -191,10 +193,18 @@ impl FileDictionaryFile {
                                 }
                             }
                         }
-                        Err(Error::TailRecordParseIncomplete) => {
+                        Err(e @ Error::TailRecordParseIncomplete) => {
+                            info!(
+                                "{:?} occurred and the last complete filename is {}",
+                                e, last_record_name
+                            );
                             break;
                         }
-                        Err(e) => return Err(e),
+                        Err(e) => {
+                            // This error is unrecoverable and manual intervention is required.
+                            set_panic_mark();
+                            return Err(e);
+                        }
                     }
                 }
             }
@@ -213,9 +223,10 @@ impl FileDictionaryFile {
             let file = self.append_file.as_mut().unwrap();
             let bytes = Self::convert_record_to_bytes(name, LogRecord::INSERT(info.clone()))?;
 
-            fail::fail_point!("file_dict_log_insert_error", |_| {
+            fail::fail_point!("file_dict_log_append_incomplete", |truncate_num| {
                 let mut bytes = bytes.clone();
-                let _ = bytes.pop().unwrap();
+                let truncate_num: usize = truncate_num.map_or(0, |c| c.parse().unwrap());
+                bytes.truncate(truncate_num);
                 file.write_all(&bytes)?;
                 file.sync_all()?;
                 Ok(())
@@ -314,8 +325,8 @@ impl FileDictionaryFile {
             warn!(
                 "file corrupted! record content size is too small, discarded the tail record";
                 "content size" => remained.len(),
-                "expect name length" => name_len,
-                "expect content length" =>info_len,
+                "expected name length" => name_len,
+                "expected content length" =>info_len,
             );
             return Err(Error::TailRecordParseIncomplete);
         }
@@ -329,7 +340,7 @@ impl FileDictionaryFile {
             if remained.is_empty() {
                 // Only when this record is the last one can the panic be skipped.
                 warn!(
-                    "file corrupted! crc32 mismatch";
+                    "file corrupted! crc32 mismatch, discarded the tail record";
                     "expected crc32" => crc32,
                     "checksum crc32" => crc32_checksum,
                 );
--- a/components/encryption/src/file_dict_file.rs
+++ b/components/encryption/src/file_dict_file.rs
@@ -5,11 +5,12 @@ use file_system::{rename, File, OpenOptions};
 use kvproto::encryptionpb::{EncryptedContent, FileDictionary, FileInfo};
 use protobuf::Message;
 use rand::{thread_rng, RngCore};
-use tikv_util::box_err;
+use tikv_util::{box_err, warn};
 
 use crate::encrypted_file::{EncryptedFile, Header, Version, TMP_FILE_SUFFIX};
 use crate::master_key::{Backend, PlaintextBackend};
 use crate::metrics::*;
+use crate::Error;
 use crate::Result;
 
 use std::io::BufRead;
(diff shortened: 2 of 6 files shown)


components/encryption/src/file_dict_file.rs
```diff
@@ -280,10 +288,11 @@ impl FileDictionaryFile {
 
     fn parse_next_record(mut remained: &[u8]) -> Result<(usize, String, LogRecord)> {
         if remained.len() < Self::RECORD_HEADER_SIZE {
-            return Err(box_err!(
-                "file corrupted! record header size is too small: {}",
-                remained.len()
-            ));
+            warn!(
```
gregwebs, line 291: It is better for ParseIncomplete to return this information then to just log it. If all the information is returned with the error then there is no need for a logging statement here: the error will end up getting logged higher up. Proper modeling would be to make an enum out of ParseIncomplete, but you could also have a struct of all the fields or even but everything into a single String.

yiwu-arbug: though I think it is fine to print the warning log here, where we have more context for detailed info.

components/encryption/src/file_dict_file.rs
```diff
@@ -172,21 +173,28 @@ impl FileDictionaryFile {
...
-                        LogRecord::REMOVE => {
-                            let original = file_dict.files.remove(&file_name);
-                            if original.is_none() {
-                                return Err(box_err!(
-                                    "Try to recovery from log file but remove a null entry, file name: {}",
-                                    file_name
-                                ));
+                    match Self::parse_next_record(remained) {
+                        Ok((used_size, file_name, mode)) => {
+                            remained.consume(used_size);
```
yiwu-arbug, line 178: I might misread the code, but `parse_next_record` also consume the bytes. It looks like we double consume the size here?

components/encryption/src/file_dict_file.rs
```diff
@@ -280,10 +288,11 @@ impl FileDictionaryFile {
 
     fn parse_next_record(mut remained: &[u8]) -> Result<(usize, String, LogRecord)> {
         if remained.len() < Self::RECORD_HEADER_SIZE {
-            return Err(box_err!(
-                "file corrupted! record header size is too small: {}",
-                remained.len()
-            ));
+            warn!(
+                "file corrupted! record header size is too small";
```
yiwu-arbug, line 292: can we mention this is the **tail** record that's corrupted (so it is probably fine to truncate it).

components/encryption/src/file_dict_file.rs
```diff
@@ -172,21 +173,28 @@ impl FileDictionaryFile {
...
+                                    if original.is_none() {
+                                        return Err(box_err!(
+                                            "Try to recovery from log file but remove a null entry, file name: {}",
+                                            file_name
+                                        ));
+                                    }
+                                }
                             }
                         }
+                        Err(Error::ParseIncomplete) => {
```
yiwu-arbug, line 194: upon tail record corruption, can we print some info regarding the last complete record, so we can debug what's being truncated?

components/encryption/src/file_dict_file.rs
```diff
@@ -293,23 +302,35 @@ impl FileDictionaryFile {
...
         if remained.len() < name_len + info_len {
-            return Err(box_err!(
-                "file corrupted! record content size is too small: {}, expect: {}",
-                remained.len(),
-                name_len + info_len
-            ));
+            warn!(
+                "file corrupted! record content size is too small";
+                "content size" => remained.len(),
+                "expect" => name_len + info_len,
```
yiwu-arbug, line 308: print `name_len` and `info_len` separately.

yiwu-arbug (commented):
Please add unit test. We should test both corrupted tail record and corrupted middle record cases.

components/encryption/src/file_dict_file.rs
yiwu-arbug: yeah, we can attach an error message with the error type. though contextual info, such as the remaining size here, can be printed as warning log in the current place.

yiwu-arbug: and this is not an actual error. This PR change to tolerate the tail corruption. we are using the special error type the denote the situation.

components/encryption/src/errors.rs
```diff
@@ -17,6 +17,10 @@ pub trait RetryCodedError: Debug + Display + ErrorCodeExt + RetryError + Send +
 pub enum Error {
     #[error("Other error {0}")]
     Other(#[from] Box<dyn error::Error + Sync + Send>),
+    // Only when the parsing record is the last one, and the
+    // length is insufficient or the crc checksum fails.
+    #[error("Recoverable types of parsing file dict error")]
+    ParseIncomplete,
```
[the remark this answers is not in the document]
hicqu: Seems it only used internally. So I prefer to remove it and rentrun an `Option` from `parse_next_record`,

components/encryption/src/file_dict_file.rs
hunterlxt: `parse_next_record` just borrow the `remained` content.

components/encryption/src/file_dict_file.rs
hunterlxt: maybe just warn info is enough, what is the use of returning debug information like `TailRecordParseIncomplete(discarded bytes)`?

components/encryption/src/file_dict_file.rs
gregwebs: File corruption seems like an error. I think you are saying that the error is recoverable. This function though should not know that: recovery is implemented by some caller. It might know that the error type indicates that it is recoverable, but it doesn't know if the callers properly implement recovery.
Logging here is also a case of knowing about the behavior of callers (that this doesn't need to be returned to a client or that it will be recovered).
Doing errors properly can certainly be heavyweight, so you will see me take shortcuts all the time, but I try to recognize that and differentiate it from the ideal error handling scenario.

components/encryption/src/errors.rs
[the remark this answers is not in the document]
gregwebs: Internal errors are still useful. This type can carry information up to the caller who can add additional contextual information to the error. But I am supportive if you think this is too heavyweight. I will experiment with using anyhow (recently merged) for making error handling more light-weight.

components/encryption/src/errors.rs
```diff
@@ -17,6 +17,10 @@ pub trait RetryCodedError: Debug + Display + ErrorCodeExt + RetryError + Send +
 pub enum Error {
     #[error("Other error {0}")]
     Other(#[from] Box<dyn error::Error + Sync + Send>),
+    // Only when the parsing record is the last one, and the
+    // length is insufficient or the crc checksum fails.
+    #[error("Recoverable types of parsing file dict error")]
```
yiwu-arbug, line 22: "Recoverable tail record corruption while parsing file dictionary"? I want to emphasize it is the tail record corruption in the error message.

components/encryption/src/file_dict_file.rs
```diff
@@ -293,23 +311,36 @@ impl FileDictionaryFile {
...
-            return Err(box_err!(
-                "file corrupted! crc32 mismatch {} != {}",
-                crc32,
-                crc32_checksum
-            ));
+            remained.consume(name_len + info_len);
+            if remained.is_empty() {
+                // Only when this record is the last one can the panic be skipped.
+                warn!(
+                    "file corrupted! crc32 mismatch";
```
yiwu-arbug, line 332:
```suggestion
                    "file corrupted! crc32 mismatch, discarded the tail record";
```

components/encryption/src/file_dict_file.rs
yiwu-arbug: From the info log I want to know for sure which record is discarded. Say in the info log of the key manager, we know that the last two encrypted files created are 001.sst and 002.sst. If we can print an info log here saying the last **complete** record was for 001.sst, we are sure 002.sst is the one being dropped when debugging.

components/encryption/src/file_dict_file.rs
yiwu-arbug: Yeah, we will say in the error message that it is the tail record being corrupted, and the caller need know no more than that. Contextual info like corruption offset etc should stay here IMO.

gregwebs: Contextual information must be added here, but why must it stay here? In Go we use errors.Annotate to good effect (this would be similar to anyhow, which I need to try out). We end up with a caller at some point logging the error, and when it is logged it has the full context from multiple function calls.

gregwebs (approved):

components/encryption/src/file_dict_file.rs
hunterlxt: I will attach the info for the last complete record with the error type

zhangjinpeng1987:
Seems we also need to pick it to 4.0?

hunterlxt:
> Seems we also need to pick it to 4.0?

yes. FYI https://github.com/tikv/tikv/issues/9886#issuecomment-810588646

components/encryption/src/file_dict_file.rs
```diff
@@ -204,6 +212,15 @@ impl FileDictionaryFile {
         if self.enable_log {
             let file = self.append_file.as_mut().unwrap();
             let bytes = Self::convert_record_to_bytes(name, LogRecord::INSERT(info.clone()))?;
+
+            fail::fail_point!("file_dict_log_insert_error", |_| {
```
tabokie, line 216:
```suggestion
            fail::fail_point!("file_dict_log_append_incomplete", |_| {
```

components/encryption/src/file_dict_file.rs
```diff
@@ -293,23 +311,36 @@ impl FileDictionaryFile {
...
-            return Err(box_err!(
-                "file corrupted! record content size is too small: {}, expect: {}",
-                remained.len(),
-                name_len + info_len
-            ));
+            warn!(
+                "file corrupted! record content size is too small, discarded the tail record";
+                "content size" => remained.len(),
+                "expect name length" => name_len,
+                "expect content length" =>info_len,
```
tabokie, line 318:
```suggestion
                "expected name length" => name_len,
                "expected content length" => info_len,
```

components/encryption/src/file_dict_file.rs
tabokie: I presume we can pass in a `bytes_to_write` parameter to this failpoint? Need to test both the header complete and incomplete case.

components/encryption/src/file_dict_file.rs
yiwu-arbug: I know `errors.Annotate`, and I don't think it is a must. I want this piece of code print the warning log for easy debugging, where the info is only relevant here. I don't want to assume the caller will print the error.

yiwu-arbug: @gregwebs patterns are great, but I also prefer simplicity.

components/encryption/src/file_dict_file.rs
hunterlxt: Just log before return break while loop, because `parse_next_record` only operate the remained bytes but not the parsed parts

components/encryption/src/file_dict_file.rs
hunterlxt: updated, `truncate_num` is added.

tests/failpoints/cases/test_encryption.rs
```diff
@@ -0,0 +1,51 @@
...
+        tempdir.path(),
+        "test_file_dict_file_record_corrupted_2",
+        true,
+        10, /*file_rewrite_threshold*/
+    )
+    .unwrap();
+    let info1 = create_file_info(1, EncryptionMethod::Aes256Ctr);
+    let info2 = create_file_info(2, EncryptionMethod::Unknown);
+    file_dict_file.insert("info1", &info1).unwrap();
+    fail::cfg("file_dict_log_append_incomplete", "return(8)").unwrap();
```
tabokie, line 36: Also return(9)? And add a comment for the difference

tabokie (approved):

components/encryption/src/file_dict_file.rs
gregwebs: @yiwu-arbug Why are you worried that the caller won't print the error? That would imply that a caller would just ignore an error: have you encountered this behavior in TiKV? Or are you concerned that the caller will not print it correctly?
I agree that logging immediately is much easier given the existing Rust tools. I will investigate different approaches.

components/encryption/src/file_dict_file.rs
```diff
@@ -169,24 +170,41 @@ impl FileDictionaryFile {
...
+                                        return Err(box_err!(
+                                            "Try to recovery from log file but remove a null entry, file name: {}",
+                                            file_name
+                                        ));
+                                    }
+                                }
                             }
                         }
+                        Err(e @ Error::TailRecordParseIncomplete) => {
+                            info!(
```
yiwu-arbug, line 197:
```suggestion
                            warn!(
```

yiwu-arbug (approved):

Issue #9945, opened 2021-04-02
The compression ratio of the migrate-in node and the migrate-out node is different


nolouch:
### Describe the feature you'd like:

Found the migrate-in node and the migrate-out has a very different compression ratio，which cause the different disk usage in big cluster https://github.com/tikv/pd/issues/3544.  
![image](https://user-images.githubusercontent.com/6428910/113384938-b5026d00-93b9-11eb-8522-1836976dc78f.png)

**Note** : size amplification = store_total_region_size / store_disk_usage

### Describe alternatives you've considered:

we try to use the `used ratio` (consider the compression ratio for eatch tikv) instead of `region size` as the most influential factor in https://github.com/tikv/pd/pull/3545 but that will cause redundant scheduling (due to the compaction and the delay delete strategy, the disk usage is not very effective) 
![image](https://user-images.githubusercontent.com/6428910/113384225-2c370180-93b8-11eb-96b5-2af969fe4505.png)
![image](https://user-images.githubusercontent.com/6428910/113396912-d752b580-93ce-11eb-8054-dba0be50ec19.png)

So it's better to make the difference of compression ratio less. we found it's relative to the `ingest snapshot`, because the snapshot is used `lz4` and most of the data in rocksdb (lower layers) are used zstd. this is proposal to use `zstd` for the snapshot.

### Teachability, Documentation, Adoption, Migration Strategy:

- rust

PR #9947, opened 2021-04-02
snap: use zstd to compress the snapshot

nolouch:

close https://github.com/tikv/tikv/issues/9945.

Problem Summary:

migrate region cause the compression ratio is difference in each TiKV. because the snapshot is used `lz4` and most of the data in rocksdb (lower layers) are used `zstd`.

### Check List 

Tests 
  **Note** : size amplification = store_total_region_size / store_disk_usage

    - Before:
image

     - After:
image

### Release note 
-  use `zstd` to compress the snapshot

### DIFF
--- a/components/raftstore/src/store/snap.rs
+++ b/components/raftstore/src/store/snap.rs
@@ -2496,7 +2496,7 @@ pub mod tests {
                 .unwrap();
 
             // TODO: this size may change in different RocksDB version.
-            let snap_size = 1658;
+            let snap_size = 1660;
             let max_snap_count = (max_total_size + snap_size - 1) / snap_size;
             // The first snap_size is for region 100.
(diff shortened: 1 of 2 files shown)


Connor1996 (commented):
You should change the code in `engine_rocks/src/sst.rs::L183`

Connor1996 (approved):
LGTM

nolouch:
> change `get_fastest_supported_compression_type` to make the `zstd` to be the first choice?

Besides `snapshot`, any external component use sstWritter?

Connor1996:
> > change `get_fastest_supported_compression_type` to make the `zstd` to be the first choice?
> 
> Besides `snapshot`, any external component use sstWritter?

Seems nowhere, let's keep it.

hicqu (approved):

PR #9955, opened 2021-04-04
ensure the panic log is flushed

gregwebs:
I noticed that the panic message was missing from the output.
Adding a very small sleep fixed this. I did try some other approaches to flushing, but I couldn't get any of them to work.
Additionally TiKV will now print out the panic error message just by itself as the very last output: this avoids having to look up past the stacktrace.

I also added a status server API `/debug/fail_point`. It can be used like this to trigger a panic, but it requires setting the feature `failpoints`.

Tests

- Manual test before and after

```
curl localhost:20180/fail/debug_fail_point -X PUT -d 'panic'
curl localhost:20180/debug/fail_point
```

Side effects

Additional 2ms sleep before abort.

### Release note

* Ensure panic output is flushed to the log

### DIFF
--- a/components/tikv_util/src/lib.rs
+++ b/components/tikv_util/src/lib.rs
@@ -435,31 +435,29 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
         .unwrap();
 
     let data_dir = data_dir.to_string();
-    let orig_hook = panic::take_hook();
+
     panic::set_hook(Box::new(move |info: &panic::PanicInfo<'_>| {
-        use slog::Drain;
-        if slog_global::borrow_global().is_enabled(::slog::Level::Error) {
-            let msg = match info.payload().downcast_ref::<&'static str>() {
-                Some(s) => *s,
-                None => match info.payload().downcast_ref::<String>() {
-                    Some(s) => &s[..],
-                    None => "Box<Any>",
-                },
-            };
-            let thread = thread::current();
-            let name = thread.name().unwrap_or("<unnamed>");
-            let loc = info
-                .location()
-                .map(|l| format!("{}:{}", l.file(), l.line()));
-            let bt = backtrace::Backtrace::new();
-            crit!("{}", msg;
-                "thread_name" => name,
-                "location" => loc.unwrap_or_else(|| "<unknown>".to_owned()),
-                "backtrace" => format_args!("{:?}", bt),
-            );
-        } else {
-            orig_hook(info);
-        }
+        let msg = match info.payload().downcast_ref::<&'static str>() {
+            Some(s) => *s,
+            None => match info.payload().downcast_ref::<String>() {
+                Some(s) => &s[..],
+                None => "Box<Any>",
+            },
+        };
+
+        let thread = thread::current();
+        let name = thread.name().unwrap_or("<unnamed>");
+        let loc = info
+            .location()
+            .map(|l| format!("{}:{}", l.file(), l.line()));
+        let bt = backtrace::Backtrace::new();
+        crit!("{}", msg;
+            "thread_name" => name,
+            "location" => loc.unwrap_or_else(|| "<unknown>".to_owned()),
+            "backtrace" => format_args!("{:?}", bt),
+        );
+        // This may be needed to allow the above log statements to flush
+        thread::sleep(Duration::from_millis(2));
 
         // There might be remaining logs in the async logger.
         // To collect remaining logs and also collect future logs, replace the old one with a
--- a/src/server/status_server/mod.rs
+++ b/src/server/status_server/mod.rs
@@ -750,6 +750,12 @@ where
                             (Method::GET, "/debug/pprof/profile") => {
                                 Self::dump_rsperf_to_resp(req).await
                             }
+                            (Method::GET, "/debug/fail_point") => {
+                                info!("debug fail point API start");
+                                fail_point!("debug_fail_point");
+                                info!("debug fail point API finish");
+                                Ok(Response::default())
+                            }
                             (Method::GET, path) if path.starts_with("/region") => {
                                 Self::dump_region_meta(req, router).await
                             }


components/tikv_util/src/lib.rs
```diff
@@ -436,16 +436,34 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
 
     let data_dir = data_dir.to_string();
     let orig_hook = panic::take_hook();
+
+    let plain_output = if let Ok(_) = env::var("TIKV_PANIC_PLAIN_OUTPUT") {
```
BusyJay, line 440: I don't think this is useful at all, `slog` is used in all deployment.

components/tikv_util/src/lib.rs
```diff
@@ -436,16 +436,34 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
...
+        let msg = match info.payload().downcast_ref::<&'static str>() {
+            Some(s) => *s,
+            None => match info.payload().downcast_ref::<String>() {
+                Some(s) => &s[..],
+                None => "Box<Any>",
+            },
+        };
+
+        let use_slog =
+            if !plain_output && slog_global::borrow_global().is_enabled(::slog::Level::Error) {
```
BusyJay, line 460: Actually it's quite confusing to check `Error` level here. The following usage output panic info using critical level, so here should probably check `Critical` instead. And `Critical` is always enabled, so the check is unnecessary at all. I think it's probably introduced back in the time when rust log is still used.

BusyJay:
Can you also add steps about how to reproduce the issue?

components/tikv_util/src/lib.rs
gregwebs: That is separate from the concerns of this PR so I think that change can be done in a different PR.

gregwebs:
> Can you also add steps about how to reproduce the issue?

The results may be particular to my machine and not easy to reproduce on someone else's. But it is highly reproducible on my machine and I think any panic will do it. For example this will not show panic output when applied to master:

```
diff --git a/cmd/src/server.rs b/cmd/src/server.rs
index 2b003f2..11d4958 100644
--- a/cmd/src/server.rs
+++ b/cmd/src/server.rs
@@ -744,6 +744,7 @@ impl<ER: RaftEngine> TiKVServer<ER> {
             cdc_scheduler,
         });

+        panic!("test panic");
         server_config
     }
```

components/tikv_util/src/lib.rs
gregwebs: It is very useful for me when developing. It is not designed for production usage. That is why I hid it behind an environment variable.

BusyJay: I suggest not to include the change. On the one hand, everyone has their preferred way to develop, it's impossible to make everyone happy, so keep code simple is preferred instead. On the other hand, the panic hook is executed only in deployment, keeping the same format in both developing and production cases will get more used to real world case.

To parse it, you can run `python -c "print(raw_input().decode('string_escape'))"`, and paste the log as input.

gregwebs: > To parse it, you can run python -c "print(raw_input().decode('string_escape'))", and paste the log as input.

This is a painful workflow since the string is bigger than my terminal window.  I don't think I have any particular preference here. I just want to see a readable backtrace, which is a requirement for development.
One way this can be handled in an automated way externally is by having a log parser and then intelligently re-formatting the log. I do this with json log output piped to jq for example. But I am at a loss for how to make this work here since we already have a plain text oriented format.

gregwebs: I think I might prefer this setting in a deployment scenario as well, depending on the exact tooling I am using. Some log tools will make it easy to see the multi-line backtrace string, but others will not, and they might happily pass through the log lines that they do not understand.

BusyJay: > since the string is bigger than my terminal window

How about `grep backtrace log | python -c "import sys; map(lambda a: sys.stdout.write(a.decode('string_escape')), sys.stdin)"`?

gregwebs: okay, that works 👍 

I can add this to the CONTRIBUTING guide.

gregwebs:
I added a panic API to the status server for easy reproducing. I think it is generally useful for users to be able to understand what happens in their system when there is a TiKV panic. Of course it does create an obvious way to perform a DoS attack if there is any access to the status server.

src/server/status_server/mod.rs
```diff
@@ -750,6 +750,9 @@ where
                             (Method::GET, "/debug/pprof/profile") => {
                                 Self::dump_rsperf_to_resp(req).await
                             }
+                            (Method::GET, "/debug/panic") => {
```
yiwu-arbug, line 753: Let's don't add such an API. For debugging we can add this code out of tree.

yiwu-arbug (commented):
rest looking good.

src/server/status_server/mod.rs
gregwebs: What do you mean by add out of tree?

Another approach would be to set an environment variable, something like `TIKV_DEV_ENV=1` to enable development/testing non-production features.

yiwu-arbug: or can we have this available only in dev build? as you said, it provide no usage out of dev environment and prone to attack.

yiwu-arbug: like, whoever want to do the testing can add the code themselves locally. I would prefer not to check it in.

gregwebs: how should I make it available only in the dev build? Another option is to use a feature flag.

BusyJay: If you wish to cause panic, you can run tikv-server-fail-point and use curl to trigger panic by putting instructions to `/fail/failpointname`, any existing frequent access failpoint name should be sufficient.

gregwebs: I can try that. Thinking about this more, I think a panic API in a deployed TiKV can be useful for our chaos testing and for customers to harden their deployment before a production launch (they should check that they get alerted when TiKV crashes). I think it could be guarded by a configuration item "debug-panic-api-enable". What do you think?

BusyJay: I don't think it's a good idea. It doesn't feel right to expose an API to allow triggering panic in a release TiKV. For chaos testing, failpoints are sufficient.

gregwebs: Okay, failpoints works very well for development. I changed the API to have a fail point.

As a best practice I always test the response of a production service (at least on a staging environment) for exception/panic to see that this is monitored. It is a shame there isn't a simple way for TiDB users to do this and they have to re-compile TiKV.

yiwu-arbug: I think what @BusyJay mean is remove this new API and use the failpoint API, which is gate behind the "failpoints" feature. https://github.com/tikv/tikv/blob/master/src/server/status_server/mod.rs#L715

gregwebs: I changed the new API to use the failpoints features. When compiled with the failpoints feature, it is easy to trigger a panic now. I added this snippet to the PR description as well.

```
curl localhost:20180/fail/debug_fail_point -X PUT -d 'panic'
curl localhost:20180/debug/fail_point
```

components/tikv_util/src/lib.rs
gregwebs: okay, I am going ahead with removing this now since it makes the PR easier to understand.

components/tikv_util/src/lib.rs
```diff
@@ -476,6 +481,14 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
             );
         }
 
+        // As a convenience show the panic message at the end of the output
+        // This avoids having to look above a long stack trace
+        if use_slog {
+            // This can prevent interleaving
```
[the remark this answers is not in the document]
gregwebs: thanks, that was an old comment.

components/tikv_util/src/lib.rs
```diff
@@ -476,6 +481,14 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
             );
         }
 
+        // As a convenience show the panic message at the end of the output
+        // This avoids having to look above a long stack trace
+        if use_slog {
```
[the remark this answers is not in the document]
gregwebs: @BusyJay I made some simplifications based on your advice. PTAL.

components/tikv_util/src/lib.rs
```diff
@@ -435,31 +435,32 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
...
+            .map(|l| format!("{}:{}", l.file(), l.line()));
+        let bt = backtrace::Backtrace::new();
+        crit!("{}", msg;
+            "thread_name" => name,
+            "location" => loc.unwrap_or_else(|| "<unknown>".to_owned()),
+            "backtrace" => format_args!("{:?}", bt),
+        );
+        // As a convenience show the panic message at the end of the output
+        // This avoids having to look above a long stack trace
+        crit!("{}", msg);
```
BusyJay, line 461: People may feel confused as the message is output twice. On the other hand, when using slog, backtrace is put in the same line as the message, so the message has to be seen no matter using less, tail or log collectors.

BusyJay (commented):
Rest LGTM

components/tikv_util/src/lib.rs
gregwebs: I don't think seeing the message twice is confusing as to what happened to TiKV. Quite the opposite. I agree that someone might assume it is a minor bug that it shows up twice, but I don't see why this poses a problem.

> backtrace is put in the same line as the message, so the message has to be seen no matter using less, tail or log collectors
This is true if one views the output in vim for example and I am sure in many log collectors.
However, if one uses less, tail, or otherwise views the terminal output, the backtrace takes up an entire screen and the actual panic message cannot be seen.

One solution to avoiding printing the panic message twice would be to print out the backtrace by itself without the panic message.

BusyJay: > ...but I don't see why this poses a problem.

You will when you need to handle customer's issue and being asked the same question again and again.

> ...takes up an entire screen...

Unless it's a very old screen that can only show 80 characters in a line, I don't see how backtrace can take up the whole screnn...And if it did, why not just scrolled up?

> ...print out the backtrace by itself without the panic message.

This is worse. Backtrace and panic message are put at the same line on purpose to reduce communication roundtrip during supporting.

gregwebs: > Unless it's a very old screen that can only show 80 characters in a line,  I don't see how backtrace can take up the whole screnn

I feel like you are making statements without bother to take the time looking at the panic output? It is now trivial to look at the panic output according to the instructions in this PR.
At a 134 character width while using an external monitor the panic output takes up a full screen for me (about 2 full vertical screens).

gregwebs: Okay @BusyJay , I have one last attempt at convenience: additionally printing the panic message in a slog field "panic". This way it shows up at the very bottom, but it is still all one slog logging statement. This is pushed now.

gregwebs: We don't know why, but I see 78 lines of stack traces, whereas @BusyJay normally sees 25.
@BusyJay does not want to make my use case convenient because he is afraid users will not send him a full screenshot of the last log message that includes the backtrace if they can see the last line of the log message already.

Of course the solution to this is to have an optional setting, but @BusyJay is against this as well.

src/server/status_server/mod.rs
yiwu-arbug: This new API is redundant since we have the failpoint API. Please remove it.

yiwu-arbug (changes requested):
The new API is redundant to the failpoitn API. Please remove it.

src/server/status_server/mod.rs
gregwebs: I discussed this with @yiwu-arbug. There is no redundancy here: this API is a new feature that does not exist in the failpoint API.

gregwebs:
> The new API is redundant to the failpoitn API.

I discussed this with @yiwu-arbug and there is no redundancy, this is new functionality that does not exist.

BusyJay (commented):
/LGTM

yiwu-arbug (approved):

components/tikv_util/src/lib.rs
gregwebs: The difference in stack trace seems to be compiling debug `make dev` versus release `make release`.

PR #9975, opened 2021-04-07
txn: Test 1PC prewrite request partially fail in unit tests

MyonKeminta:

What's Changed: This PR tries to cover the case that in a 1PC prewrite request with multiple keys, if it fails on one of the keys, the previously succeeded keys should never be written. I didn't find if it's already covered by any other cases, so I added these case specifically for 1PC. I'm afraid if in the future someone wants to support writing down the successful part of a request and accidentally breaks 1PC.

### Check List 

Tests 

- Unit test

### Release note 

- No release note.


sticnarf:
Oops, I always thought I wrote tests about this case but I cannot find it either.........

MyonKeminta:
> Oops, I always thought I wrote tests about this case but I cannot find it either.........

I have such impression too 😕

sticnarf (approved):

youjiali1995 (approved):

PR #9969, opened 2021-04-06
*: fix issue that some components depend on protobuf-codec unconditionally

kennytm:
### What problem does this PR solve?

Problem Summary:

Some components unconditionally imports `tikv_util` making `protobuf-codec` always be enabled even when we want `prost-codec`.

### What is changed and how it works?

What's Changed:

Change such imports to `default-features = false` and insert the appropriate `protobuf`/`prost-codec` features.

### Related changes

- Need to cherry-pick to the release branch
   * release-5.0

### Check List 

Tests 

- Manual test (add detailed scripts or steps below)
   * tikv/importer which uses `prost-codec` passes `make test` after 9968.

Side effects

### Release note 

- No release notes

PR #8793, opened 2020-10-09
*: move server setups to a separate crate and lower its optimization level

sticnarf:
### What is changed and how it works?

Previously, it always took a long time to compile the `cmd` crate but it does not contain anything heavy. Actually we don't need to give much optimization to that part.

This PR moves the setup code of tikv-server to a component crate, and sets its opt-level to 1. This optimization level avoids most of inlines. Because these code is only executed once, it's not so necessary to inline them with other crates and optimize again.

With this PR, we decrease `make release` time but affects little performance.

It does not affect the opt level of `make dist_release`.

On my Ryzen 3900X PC, this PR reduces `make release` time by about 100 seconds:

Before:

image

After:

image

### Check List 

Tests 

- Manual test

### Release note 

- No release note

PR #9979, opened 2021-04-07
*: make PROST work again

sticnarf:

From some time ago, build with `PROST=1` receives error like:

```
error[E0592]: duplicate definitions with name `generate_files`
  --> /home/yilin/.cargo/registry/src/github.com-1ecc6299db9ec823/protobuf-build-0.12.0/src/protobuf_impl.rs:63:5
   |
63 |     pub fn generate_files(&self) {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ duplicate definitions for `generate_files`
   |
  ::: /home/yilin/.cargo/registry/src/github.com-1ecc6299db9ec823/protobuf-build-0.12.0/src/prost_impl.rs:5:5
   |
5  |     pub fn generate_files(&self) {
   |     ---------------------------- other definition for `generate_files`

error: aborting due to previous error
```

This is because some components depend on protobuf-codec unconditionally. https://github.com/tikv/tikv/pull/9969 should fix most of previous mistakes but my #8793 introduces some more.

This PR fixes all new mistakes and make PROST work again.

### Check List 

Tests 

`PROST=1 make build` works now.

### Release note 

- No release note

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ protobuf-codec = [
   "tidb_query_executors/protobuf-codec",
   "tidb_query_expr/protobuf-codec",
   "tipb/protobuf-codec",
+  "tikv_kv/protobuf-codec",
   "tikv_util/protobuf-codec",
   "txn_types/protobuf-codec",
   "grpcio-health/protobuf-codec",
@@ -85,6 +86,7 @@ prost-codec = [
   "tidb_query_executors/prost-codec",
   "tidb_query_expr/prost-codec",
   "tipb/prost-codec",
+  "tikv_kv/prost-codec",
   "tikv_util/prost-codec",
   "txn_types/prost-codec",
   "grpcio-health/prost-codec",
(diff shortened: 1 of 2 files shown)


BusyJay (approved):

kennytm:
can we add `PROST=1 make build` as part of the CI (or just remove the `prost-codec` feature entirely)

sticnarf:
> can we add `PROST=1 make build` as part of the CI (or just remove the `prost-codec` feature entirely)

Previously it was part of the CI but I don't know why it disappears now.

BusyJay:
It's checked daily to speed up build.

BusyJay:
There are (at least) two things blocks switching to prost codec by default. The first is configurable recursion limit support. The second is log redaction. Maintaining two codec at the same time is a real pain. :(

kennytm:
heh from what i've heard the most recent consensus is keep protobuf-codec as default since performance of prost isn't much better 🤔

nrc:
The performance is a little better but doesn't justify migration alone. Prost has better codegen (generates more idiomatic data structures and preserves comments); iirc, it also could facilitate some better optimisations, e.g., almost zero copy gRPC.

PR #9984, opened 2021-04-08
generate snapshots with fresh RocksDB snapshots

hicqu:

For snapshot generating, RocksDB snapshot are taken in apply workers then sent to snapshot generator threads. However it could be a large gap between them. It could causes generated snapshots get stale quickly.

Take RocksDB snapshots in snapshot generator threads.

### Check List 

Tests 

- Unit test

### Release note 
* No release note.


BusyJay:
This is wrong as the index doesn't match the data. I suggest to fail fast instead. If a snapshot has to be stale, cancel it directly.

gengliqi:
The snapshot is generated in apply thread because of some historical issues. For now, it can be generated in store thread, namely, `generate_and_schedule_snapshot` can be called directly in store thread.
Also generate it in snapshot generator threads is a good idea.

PR #9987, opened 2021-04-09
deregister region when resolver fails to build

liuzix:
Signed-off-by: Zixiong Liu <liuzixiong@pingcap.com>

**This PR is to `release-5.0` because master and `release-5.0` have diverged in the handling of incremental scan.**

https://github.com/pingcap/ticdc/issues/1626

Problem Summary:

When a CDC incremental scan fails, if the connection is responsible for building the resolver for that region, it should handle its exit properly so that other connections can have a usable resolver.

What's Changed:

The current design makes it hard to handle it elegantly, so as a workaround, a failed incremental scan will deregister the whole region if it is supposed to build the resolver.

### Related changes

- Need to handle this in the master branch.
### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- Fix interference between connections to the same region.


components/cdc/src/endpoint.rs
```diff
@@ -534,16 +534,24 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Endpoint<T> {
...
+                Deregister::Downstream {
+                    region_id,
+                    downstream_id,
+                    conn_id,
+                    err: Some(err),
+                }
             };
             if let Err(e) = scheduler.schedule(Task::Deregister(deregister)) {
-                error!("cdc schedule cdc task failed"; "error" => ?e);
+                error!("schedule cdc task failed"; "error" => ?e);
```
overvenus, line 554:
```suggestion
                error!("cdc schedule cdc task failed"; "error" => ?e);
```

components/cdc/src/endpoint.rs
```diff
@@ -534,16 +534,24 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Endpoint<T> {
 
         let (cb, fut) = tikv_util::future::paired_future_callback();
         let scheduler = self.scheduler.clone();
+        let observe_id = delegate.id;
         let deregister_downstream = move |err| {
```
overvenus, line 538:
```suggestion
        let deregister = move |err| {
```

Because it may deregister a region.

components/cdc/tests/failpoints/test_endpoint.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+        let mut event_vec = event.events.to_vec();
+        while event_vec.len() > 0 {
+            match event_vec.pop().unwrap().event.unwrap() {
+                Event_oneof_event::Error(err) => {
+                    assert!(err.has_region_not_found(), "{:?}", err);
+                    break 'outer;
+                }
+                e => {
+                    println!("cdc receive_event {:?}", e);
```
overvenus, line 129:
```suggestion
                    debug!("cdc receive event {:?}", e);
```

Please do not use `println!` in tests.

components/cdc/tests/failpoints/test_endpoint.rs
```diff
@@ -0,0 +1,161 @@
...
+        let mut event_vec = event.events.to_vec();
+        println!("cdc receive_event {:?}", event_vec);
+        while event_vec.len() > 0 {
+            match event_vec.pop().unwrap().event.unwrap() {
+                Event_oneof_event::Error(err) => {
+                    assert!(err.has_region_not_found(), "{:?}", err);
+                    break 'outer1;
+                }
+                e => {
+                    println!("cdc receive_event {:?}", e);
```
overvenus, line 152:
```suggestion
                    debug!("cdc receive event {:?}", e);
```

components/cdc/src/endpoint.rs
```diff
@@ -1086,6 +1093,21 @@ impl Initializer {
 
         fail_point!("cdc_incremental_scan_start");
 
+        let deregister_region = || {
```
amyangfei, line 1096: Should `deregister_region` be called before L1150

liuzix: Technically yes, but if the program ever gets to L1150, there is likely to be a very serious congestion or deadlock, and deregistering is likely to fail as well.

components/cdc/src/endpoint.rs
```diff
@@ -1101,22 +1123,14 @@ impl Initializer {
...
-                    let deregister = Deregister::Downstream {
-                        region_id,
-                        downstream_id,
-                        conn_id,
-                        err: Some(e),
-                    };
-                    if let Err(e) = self.sched.schedule(Task::Deregister(deregister)) {
-                        error!("cdc schedule cdc task failed"; "error" => ?e, "region_id" => region_id);
-                    }
+                    deregister_region();
```
5kbpers, line 1133: If the scan is failed, the downstream should be deregistered at least.

PR #9988, opened 2021-04-09
Use RocksEngine in compaction filter

tabokie:

Issue Number: close #xxx 

Problem Summary:

Production code should interact with underlying data engine through `KvEngine` trait. In compaction filter code this rule is broken because a native write batch is relayed as an enum type, which can't be easily abstracted. Re-translate it into key vector would be approachable with some memory overhead. Here I will settle for making compaction filter interact with `RocksEngine` instead of raw `DB`.

Direct reason for this change is I made an assumption that all TiKV codes are at least using `RocksEngine` at [this line](https://github.com/tikv/tikv/blob/8d189a9b6724b5c862659530a7ad6b4089dfcbdc/components/engine_rocks/src/write_batch.rs#L76) in #9197, which turns out to be untrue. Fixing it while I'm at it.

What's Changed:

Use `RocksEngine` in compaction filter.

### Check List

Tests

- Unit test

### Release note 

- No release note


yiwu-arbug (approved):
LGTM

hicqu (approved):

PR #9989, opened 2021-04-09
coprv2: config for plugin directory

Jibbow:

Issue Number: #9747  

Problem Summary:

Adds a new configuration *section* with one *configuration option* to TiKV's configuration.
```toml
[coprocessor-v2]
coprocessor-plugin-directory = "coprocessors"
```

The config is then passed to the new coprocessor endpoint where it will be registered for hot-reloading.

What's Changed:

Instead of adding every plugin individually to TiKV's config, we just use one "plugin directory".

### Related changes

- based on PR #9925 

### Check List 

Tests 

- I'm not entirely sure how to test this. Also I'm not quite sure how to test this manually :see_no_evil: 

### Release note 

 * No release note


src/config.rs
```diff
@@ -2348,6 +2349,9 @@ pub struct TiKvConfig {
     #[config(submodule)]
     pub coprocessor: CopConfig,
 
+    #[config(skip)]
+    pub coprocessor_v2: CoprocessorV2Config,
```
Jibbow, line 2353: Not sure if `#[config(skip)]` is right here. Is this okay? I did not quite understand what `#[config(submodule)]` is.

src/coprocessor_v2/config.rs
```diff
@@ -0,0 +1,16 @@
...
+#[serde(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct Config {
+    pub coprocessor_plugin_directory: String,
+}
+
+impl Default for Config {
+    fn default() -> Config {
+        Config {
+            coprocessor_plugin_directory: "coprocessors".to_string(),
```
Jibbow, line 13: What's the difference between a default value here and the value in `etc/config-template.toml`? Are they related or are they independent?

src/coprocessor_v2/endpoint.rs
```diff
@@ -23,9 +24,17 @@ pub struct Endpoint {
...
 impl Endpoint {
-    pub fn new() -> Self {
+    pub fn new(copr_cfg: &Config) -> Self {
+        let mut plugin_registry = PluginRegistry::new();
+
+        std::fs::create_dir_all(&copr_cfg.coprocessor_plugin_directory)
+            .expect("Unable to create directory for coprocessor plugins.");
+        plugin_registry
+            .start_hot_reloading(&copr_cfg.coprocessor_plugin_directory)
+            .expect("Unable to load coprocessor plugins from directory.");
```
Jibbow, line 34: What should we do if we can't load plugins or create the directory? Just let it crash? Log it?

gfreezy:
looking forward to it.

src/config.rs
andylokandy: `#[config(submodule)]` means to allow changing the config online via an HTTP url.

src/coprocessor_v2/endpoint.rs
andylokandy: I'd like to log an error. 😄

src/config.rs
Jibbow: Cool, thanks! Got it :)

src/coprocessor_v2/endpoint.rs
Jibbow: Done, PTAL :blush: 
Also moved `std::fs::create_dir_all(&plugin_directory)?;` into `PluginRegistry::start_hot_reloading()`.

src/coprocessor_v2/config.rs
skyzh, line 13: I don't think it's a good idea to use `coprocessors` as the default config. It's better to be `None`. From my perspective, I would be surprised if there's a `coprocessors` directory popping out after running `tikv-server`, when I don't explicitly enable this feature.

skyzh (commented):
Rest LGTM

src/coprocessor_v2/config.rs
Jibbow: Ah yes, good idea! I didn't think of that. Will implement it :)

Jibbow: Done :blush:

skyzh (approved):

src/coprocessor_v2/config.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::path::Path;
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct Config {
+    pub coprocessor_plugin_directory: Option<Path>,
```
skyzh, line 9: Should be PathBuf. Path is something like [u8], which is not Sized.

src/config.rs
```diff
@@ -3782,6 +3787,7 @@ mod tests {
         // Other special cases.
         cfg.pd.retry_max_count = default_cfg.pd.retry_max_count; // Both -1 and isize::MAX are the same.
         cfg.storage.block_cache.capacity = OptionReadableSize(None); // Either `None` and a value is computed or `Some(_)` fixed value.
+        cfg.coprocessor_v2.coprocessor_plugin_directory = None; // Default is `None`, which is represented by not setting the key.
```
andylokandy, line 3790: Why is this a special case?

andylokandy: Oh, I got it.

etc/config-template.toml
```diff
@@ -368,6 +368,13 @@
 ## Set to "mvcc" to do consistency check for MVCC data, or "raw" for raw data.
 # consistency-check-method = "mvcc"
 
+[coprocessor-v2]
+## Path to the directory where compiled coprocessor plugins are located.
+## Plugins in this directory will be automatically loaded by TiKV.
+## If a coprocessor plugin file is deleted, it will be automatically unloaded.
+## If the config value is not set, hot-reloading will be disabled.
+# coprocessor-plugin-directory = "coprocessors"
```
andylokandy, line 376:
```suggestion
# coprocessor-plugin-directory = "./coprocessors"
```

andylokandy (approved):

PR #9993, opened 2021-04-12
config: Remove obsolete items from the config template file.

MyonKeminta:

Problem Summary: Some configs are deleted in the past months, but these items are not deleted from config-template.toml, which is a mistake and confuses readers.

What's Changed: This PR deleted these config items from config-template that are actually already deleted.

### Related changes

  - I think this PR should be cherry-picked to 5.0 since people may try to confirm the configs by reading the config template from code.

### Check List 

Tests 

- Unit test

### Release note 

- No release note


sticnarf (approved):
LGTM

hicqu (approved):

PR #9999, opened 2021-04-12
*: update deps

BusyJay:

To resolve RUSTSEC-2020-0151 and RUSTSEC-2021-0013.

### Check List 

Tests 

### Release note 
- No release note.


sticnarf (approved):
LGTM. Although the two security issues should only affect tests and benches.

BusyJay:
> Although the two security issues should only affect tests and benches.

Yes, but it keep waking up the alert.

Connor1996 (approved):
LGTM

BusyJay:
@5kbpers would you like to take a look at the fail case? I can reproduce it on my local machine with master.

Issue #9918, opened 2021-03-29
Add learner too slow which cause the PD operator timeout


rleungx:
### What version of TiKV are you using?
5.0 branch (4db8ee36f69b3ec2d5c5281800e1161ee7af7665)

### What operating system and CPU are you using?
Not related

### Steps to reproduce

Just offline the store.

### What did you expect?

The offline operator can be successfully executed.

### What did happened?
```
[operator_controller.go:566] ["operator timeout"] [region-id=375619799] [takes=20m3.453927516s] [operator="\"replace-rule-offline-peer {mv peer: store [442652643] to [220163081]} (kind:leader,region,replica, region:375619799(51833,1446), createAt:2021-03-29 10:16:25.152050758 +0800 CST m=+67530.662320625, startAt:2021-03-29 10:16:25.170851946 +0800 CST m=+67530.681121812, currentStep:0, steps:[add learner peer 442702203 on store 220163081, transfer leader from store 442652643 to store 3, use joint consensus, promote learner peer 442702203 on store 220163081 to voter, demote voter peer 442671526 on store 442652643 to learner, leave joint state, promote learner peer 442702203 on store 220163081 to voter, demote voter peer 442671526 on store 442652643 to learner, remove peer on store 442652643]) timeout\""]
```

And from the TiKV log:
```
[2021/03/29 10:16:25.169 +08:00] [Info] [pd.rs:824] ["try to change peer"] [peer="id: 442702203 store_id: 220163081 role: Learner"] [change_type=AddLearnerNode] [region_id=375619799]
[2021/03/29 10:16:25.169 +08:00] [Info] [peer.rs:3145] ["propose conf change peer"] [kind=Simple] [changes="[change_type: AddLearnerNode peer { id: 442702203 store_id: 220163081 role: Learner }]"] [peer_id=442671526] [region_id=375619799]
[2021/03/29 10:16:25.170 +08:00] [Info] [apply.rs:1248] ["execute admin command"] [command="cmd_type: ChangePeer change_peer { change_type: AddLearnerNode peer { id: 442702203 store_id: 220163081 role: Learner } }"] [index=91] [term=56] [peer_id=442671526] [region_id=3756
19799]
[2021/03/29 10:16:25.170 +08:00] [Info] [apply.rs:1613] ["exec ConfChange"] [epoch="conf_ver: 1446 version: 51833"] [type=AddLearner] [peer_id=442671526] [region_id=375619799]
[...]

PR #10001, opened 2021-04-12
allow to cancel snapshot generating if necessary

hicqu:

There could be a large gap between a RocksDB snapshot is taken and a Region snapshot is built based on it. It's possible that the Region snapshot become stale in the gap, for example, more Raft logs are compacted or the leadership is transferred.

This PR allows to cancel a snapshot generating task if necessary, to save I/O traffic.

It's a part to resolve tikv/tikv#9918.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
  I constructed a scenario that generated snapshots will always become stale because of raft logs are compacted very quickly. Here is the result:
  
图片

  and log:
  ```
  [2021/04/14 16:28:05.330 +08:00] [INFO] [region.rs:293] ["generate snap is canceled"] [region_id=98]
[2021/04/14 16:28:05.330 +08:00] [INFO] [region.rs:293] ["generate snap is canceled"] [region_id=2]
[2021/04/14 16:28:05.330 +08:00] [INFO] [region.rs:293] ["generate snap is canceled"] [region_id=98]
[2021/04/14 16:28:05.330 +08:00] [INFO] [region.rs:293] ["generate snap is canceled"] [region_id=2]
  ```

  The new metrics:
图片

Side effects

### Release note 
* No release note


BusyJay:
Cool, how does it perform?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -976,26 +996,23 @@ where
             panic!("{} unexpected state: {:?}", self.tag, *snap_state);
         }
 
-        if *tried_cnt >= MAX_SNAP_TRY_CNT {
```
Connor1996, line 979: why remove this

Connor1996 (commented):
seems the metrics `SNAP_COUNTER` and `SNAP_HISTOGRAM` are not shown in grafana, please add them so we can know the pending generating snapshots and duration of generating snapshot?

components/raftstore/src/store/peer_storage.rs
hicqu: Snapshot generating can be kept canceling if the snapshot generator pool is busy.

hicqu:
> seems the metrics `SNAP_COUNTER` and `SNAP_HISTOGRAM` are not shown in grafana, please add them so we can know the pending generating snapshots and duration of generating snapshot?

They are shown in the snapshot row. PTAL at http://172.16.4.78:60021/d/RDVQiEzZz/local-tikv-details?orgId=1&refresh=30s&from=now-1h&to=now.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2815,12 +2827,15 @@ impl GenSnapTask {
     where
         EK: KvEngine,
     {
+        self.index
+            .store(last_applied_state.applied_index, Ordering::Relaxed);
```
BusyJay, line 2831: Suggest to use `SeqCst` instead, we usually use `Relaxed` only when for metrics or protected by lock.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -833,6 +836,17 @@ where
...
         self.apply_state = apply_state;
+        if let SnapState::Generating {
+            ref canceled,
+            ref index,
+            ..
+        } = *self.snap_state.borrow()
+        {
+            let snap_index = index.load(Ordering::Relaxed);
+            if snap_index > 0 && self.apply_state.get_truncated_state().index > snap_index {
+                canceled.store(true, Ordering::Relaxed);
```
BusyJay, line 847: I think you can reset the state here.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -833,6 +837,19 @@ where
...
         self.apply_state = apply_state;
+        let mut snap_state = self.snap_state.borrow_mut();
+        if let SnapState::Generating {
+            ref canceled,
+            ref index,
+            ..
+        } = *snap_state
+        {
+            let snap_index = index.load(Ordering::SeqCst);
+            if snap_index > 0 && self.apply_state.get_truncated_state().index > snap_index {
```
BusyJay, line 848: How about exposing a method, and allow called from outside? So it can be called in various `on_ready_xxx`.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -939,16 +943,22 @@ where
...
+        } = *snap_state
+        {
             tried = true;
-            match recv.try_recv() {
-                Err(TryRecvError::Disconnected) => {}
-                Err(TryRecvError::Empty) => {
-                    return Err(raft::Error::Store(
-                        raft::StorageError::SnapshotTemporarilyUnavailable,
-                    ));
+            if !canceled.load(Ordering::SeqCst) {
```
BusyJay, line 953: If it's canceled, `tried_cnt` should not be counted.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1074,6 +1090,21 @@ where
             let rid = self.get_region_id();
             self.engines.raft.gc_entry_cache(rid, idx);
         }
+
+        let mut snap_state = self.snap_state.borrow_mut();
```
BusyJay, line 1094: How about making `cancel_generating_snap` accept an index parameter, so it can be reused here?

BusyJay (commented):
Conf Change should also cancel snapshot.

tests/failpoints/cases/test_snap.rs
```diff
@@ -470,3 +469,55 @@ fn test_gen_snapshot_with_no_committed_entries_ready() {
...
+    for entry in fs::read_dir(&snap_dir).unwrap() {
+        let entry = entry.unwrap();
+        fs::remove_file(entry.path()).unwrap();
+    }
+
+    fail::cfg("before_region_gen_snap", "pause").unwrap();
+    pd_client.must_add_peer(rid, new_learner_peer(4, 4));
+
+    // Snapshot generatings will be canceled by raft log GC.
+    (0..100).for_each(|_| cluster.must_put(b"kk", b"vv"));
```
BusyJay, line 504: How about `(0..15)`? I'm afraid 100 takes too long to finish.

tests/failpoints/cases/test_snap.rs
```diff
@@ -470,3 +469,55 @@ fn test_gen_snapshot_with_no_committed_entries_ready() {
     // response from peer 3.
     must_get_equal(&cluster.get_engine(3), b"k9", b"v1");
 }
+
+#[test]
+fn test_cancel_snapshot_generating() {
```
BusyJay, line 474: Please add description for the purpose and scenario.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -938,17 +942,24 @@ where
...
+        } = *snap_state
+        {
             tried = true;
-            match recv.try_recv() {
-                Err(TryRecvError::Disconnected) => {}
-                Err(TryRecvError::Empty) => {
-                    return Err(raft::Error::Store(
-                        raft::StorageError::SnapshotTemporarilyUnavailable,
-                    ));
+            last_canceled = canceled.load(Ordering::SeqCst);
```
BusyJay, line 953: Perhaps should wait it's really canceled before next try.

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2823,15 +2823,27 @@ pub struct CatchUpLogs {
 
 pub struct GenSnapTask {
     pub(crate) region_id: u64,
+    // Fill it after the RocksDB snapshot is taken.
+    pub index: Arc<AtomicU64>,
```
gengliqi, line 2827: Seems that `index` doesn't need to be `atomic`.

BusyJay (approved):

components/raftstore/src/store/fsm/apply.rs
hicqu: It's filled in region worker, but fetched in raftstore thread. So I guess `Arc` is the most simple way to deal with it.

gengliqi: I get it. BTW, it is filled in apply thread not region worker.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1335,6 +1346,27 @@ where
         self.check_applying_snap() != CheckApplyingSnapStatus::Applying
     }
 
+    /// Cancel generating snapshot, return true if there is such a job.
+    pub fn cancel_generating_snap(&mut self, compact_to: Option<u64>) -> bool {
```
gengliqi, line 1350: Seems the return value is useless. Also if the `canceled` is true, don't need to check `index` again.

gengliqi (commented):
Rest LGTM

gengliqi (approved):

PR #5593, opened 2019-10-08
*: fix thread name truncating issue

gengliqi:
Signed-off-by: gengliqi <gengliqiii@gmail.com>

###  What have you changed?
Add a tid-to-thread-name hashmap which is used in metrics.

Since thread name length in linux is restricted to 16 characters, the exceeding part will be truncated.
It's hard to differentiate between different threads when thread name is too long, e.g. raftstore thread name is "raftstore-x-y", x is store_id, y is counter. The store_id maybe larger when a new node is added to a long-running cluster. So we cannot differentiate them from each other.

The limitations of the current code is that some certain thread names can't be added to hashmap because these threads are created in other crates, such as `futures_cpupool`, `grpcio`, `rocksdb`. 

Threads from `std` and `tokio_threadpool` are handled by using wrapper to replace `spawn` and `after_start` & `before_stop` function respectively.

It is necessary to use after_start_wrapper and before_stop_wrapper together. Otherwise it may cause reporting a wrong thread name when a new thread whose name is not added in hashmap is assigned same tid as previous thread. 
So I add default `after_start` & `before_stop` in future_pool if not used. But if someone directly use tokio_threadpool and want to add thread name to hashmap, two wrapper should be used together. For now, this mistake is not exist in our code.

###  What is the type of the changes?
Bugfix.

###  How is the PR tested?
Manual test.

###  Does this PR affect documentation (docs) or should it be mentioned in the release notes?
No.

###  Does this PR affect `tidb-ansible`?
No.

PR #10004, opened 2021-04-12
*: Fix truncated thread name

Connor1996:

Problem Summary:

The max length of thread name is 16 limited by kernel. The length of thread name of unified read pool and storage read pool exceeds 16. Kernel truncates the thread name so that we can't group by the thread name in metrics conveniently. See
截屏2021-04-12 下午5 35 44

What's Changed:

- Change thread name of `unified-read-pool` to `unified-read`
- Change thread name of `store-read-low/normal/high` to `store-read-lo/nl/hi`

### Check List 

- No code

### Release note 

- None

### DIFF
--- a/metrics/grafana/tikv_details.json
+++ b/metrics/grafana/tikv_details.json
@@ -4822,7 +4822,7 @@
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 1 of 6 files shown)


sticnarf:
Could the previous way using `unified-read-po*` cause any problem?

BusyJay:
I suggest to maintain a map between thread id and thread name instead. You can see a previous implementation at #5593.

Connor1996:
> Could the previous way using `unified-read-po*` cause any problem?

can't see CPU usage for each thread

PR #10010, opened 2021-04-13
Validate `config-template.toml`

### LINKED ISSUE #9990: Provided config-template.toml is not valid

latest `master` branch.

Linux, but shouldn't matter.

### Steps to reproduce

Take the provided `config-template.toml` for TiKV, strip leading `#` from all lines and use it for as a config for TiKV.

### What did you expect?
It should "just work" and be a valid config. In the ideal case, it would be identical to `TiKvConfig::default()`.

### What did happened?
Several things were wrong:

 1. after stripping away the leading `#`, the resulting toml was *not parsable*. This was mainly because *some* lines with comments used only a single `#` instead of two `##`. Only a minor issue, but still it's inconsistent.
 2. after fixing (1), the toml was parsable as a `TiKvConfig`, but it did not pass `validate()`.
 3. after fixing (2), the resuling config differed from `TiKvConfig::default()`. Maybe that's intended, but I found it rather unusual. Here is a diff (left: `config-template.toml`, right: `TiKvConfig::default()`): https://www.diffchecker.com/Et8GSSHO

If it's okay, I would create a PR with fixes for (1) and (2) and maybe (3) depending on the comments on this issue. I would also add a test to check these issues for the future.

Jibbow:

Makes sure that the provided `config-template.toml` is valid.

Issue Number: close #9990 

Problem Summary:

As depicted in #9990, there are currently some issues with `config-template.toml` on master, e.g. inconsistent use of `#` and `##` and also it does not pass `validate()`. Further, some config options differ from default values from `TiKvConfig::Default()`.

### Check List 

Tests 

- Unit test

Side effects

- We might change some parameters in `config-template.toml` depending on the discussion.

### Release note 

 * No release note.


components/pd_client/src/config.rs
```diff
@@ -13,7 +13,7 @@ use tikv_util::config::ReadableDuration;
 pub struct Config {
     /// The PD endpoints for the client.
     ///
-    /// Default is empty.
+    /// Default is `"127.0.0.1:2379"`.
```
Jibbow, line 16: Just updated the comment because `Config::Default()` sets it to `127.0.0.1:2379`. See also https://github.com/tikv/tikv/blob/3ef232e7f2f4ad8087c8268026cefa958b637194/components/pd_client/src/config.rs#L44

src/config.rs
```diff
@@ -3787,4 +3788,26 @@ mod tests {
...
+
+    #[test]
+    fn test_config_template_matches_default() {
+        let template_config = std::include_str!("../etc/config-template.toml")
+            .lines()
+            .map(|l| l.strip_prefix('#').unwrap_or(l))
+            .join("\n");
+
+        let cfg: TiKvConfig = toml::from_str(&template_config).unwrap();
+        assert_eq!(cfg, TiKvConfig::default());
```
Jibbow, line 3811: This test is currently failing (intentionally) because I'm not sure if/how we should change the values in `config-template.toml`.
Here is a diff (left: `config-template.toml`, right: `TiKvConfig::default()`): https://www.diffchecker.com/Et8GSSHO

MyonKeminta: I'm afraid it's not easy to do the test you want. Some config items' default values are calculated rather than fixed value. For example, readpool.unified.max-thread-count is calculated from the CPU nums.

MyonKeminta: Maybe you can consider set these special items to the same value before doing the assert_eq.

Jibbow: Ah, thanks for the comment! Good point! Maybe we could also leave out that second test anyways :)

What would you prefer? Handling special items or leaving out the test?

MyonKeminta: Both is ok to me, and I personally prefer keeping the test case and handling the special items.

etc/config-template.toml
```diff
@@ -56,7 +56,7 @@
 
 [readpool.storage]
 ## Whether to use the unified read pool to handle storage requests.
-# use-unified-pool = false
+# use-unified-pool = true
```
Jibbow, line 59: Now same as 
https://github.com/tikv/tikv/blob/master/etc/config-template.toml#L86

etc/config-template.toml
```diff
@@ -123,7 +123,7 @@
 # grpc-compression-type = "none"
 
 ## Size of the thread pool for the gRPC server.
-# grpc-concurrency = 4
+# grpc-concurrency = 5
```
Jibbow, line 126: https://github.com/tikv/tikv/blob/master/src/server/config.rs#L20

etc/config-template.toml
```diff
@@ -179,12 +179,12 @@
...
 [storage]
 ## The path to RocksDB directory.
-# data-dir = "/tmp/tikv/store"
+# data-dir = "./"
 
 ## The number of slots in Scheduler latches, which controls write concurrency.
 ## In most cases you can use the default value. When importing data, you can set it to a larger
 ## value.
-# scheduler-concurrency = 2048000
+# scheduler-concurrency = 524288
```
Jibbow, line 187: https://github.com/tikv/tikv/blob/master/src/storage/config.rs#L20

etc/config-template.toml
```diff
@@ -179,12 +179,12 @@
 
 [storage]
 ## The path to RocksDB directory.
-# data-dir = "/tmp/tikv/store"
+# data-dir = "./"
```
Jibbow, line 182: https://github.com/tikv/tikv/blob/master/src/storage/config.rs#L17

etc/config-template.toml
```diff
@@ -232,7 +232,7 @@
 
 [pd]
 ## PD endpoints.
-# endpoints = []
+# endpoints = ["127.0.0.1:2379"]
```
Jibbow, line 235: https://github.com/tikv/tikv/blob/master/components/pd_client/src/config.rs#L44

etc/config-template.toml
```diff
@@ -296,7 +296,7 @@
 # raft-log-gc-threshold = 50
 
 ## When the entry count exceeds this value, GC will be forced to trigger.
-# raft-log-gc-count-limit = 72000
+# raft-log-gc-count-limit = 73728
```
Jibbow, line 299: https://github.com/tikv/tikv/blob/master/components/raftstore/src/store/config.rs#L213

etc/config-template.toml
```diff
@@ -417,7 +417,7 @@
 ## persistent storage. See https://github.com/facebook/rocksdb/wiki/How-to-persist-in-memory-RocksDB-database .
 ## If there are two disks on the machine, storing RocksDB data and WAL logs on different disks can
 ## improve performance.
-# wal-dir = "/tmp/tikv/store"
+# wal-dir = ""
```
Jibbow, line 420: https://github.com/tikv/tikv/blob/master/src/config.rs#L1031

etc/config-template.toml
```diff
@@ -789,7 +777,7 @@
 # writable-file-max-buffer-size = "1MB"
 # use-direct-io-for-flush-and-compaction = false
 # enable-pipelined-write = true
-# allow-concurrent-memtable-write = false
+# allow-concurrent-memtable-write = true
```
Jibbow, line 780: see https://github.com/tikv/tikv/issues/9990#issuecomment-817511558

etc/config-template.toml
```diff
@@ -798,7 +786,6 @@
 # info-log-keep-log-file-num = 10
 # info-log-dir = ""
 # info-log-level = "info"
-# optimize-filters-for-hits = true
```
Jibbow, line 801: Not present anymore in `RaftDbConfig` 
https://github.com/tikv/tikv/blob/master/src/config.rs#L1264

etc/config-template.toml
```diff
@@ -924,9 +911,6 @@
 ## The default value is set to min(CPU_NUM * 0.75, 32).
 # num-threads = 24
 
-## Number of ranges to backup in one batch.
-# batch = 8
```
Jibbow, line 928: There is only `batch_size`, but no `batch`
https://github.com/tikv/tikv/blob/master/src/config.rs#L2242

src/config.rs
Jibbow: Cool! I kept the test, but now handle these cases accordingly :) PTAL

Jibbow:
I'm also a bit confused by the `level_merge` config option in `TitanCfConfig`.

I added a test to detect unused keys, but the test fails because of `level_merge`. Any ideas?

MyonKeminta:
> I'm also a bit confused by the `level_merge` config option in `TitanCfConfig`.
> 
> I added a test to detect unused keys, but the test fails because of `level_merge`. Any ideas?

Looks like it should be `level-merge` (a hyphen instead of an underscore between the word).

Jibbow:
> > I'm also a bit confused by the `level_merge` config option in `TitanCfConfig`.
> > I added a test to detect unused keys, but the test fails because of `level_merge`. Any ideas?
> 
> Looks like it should be `level-merge` (a hyphen instead of an underscore between the words).

Ah yes, thanks!

etc/config-template.toml
Jibbow: Alright, you are probably right. Added it again :)

etc/config-template.toml
sticnarf: Yes, it's now on by default.

sticnarf (approved):
LGTM

src/config.rs
```diff
@@ -3787,4 +3788,82 @@ mod tests {
...
+            .map(|l| l.strip_prefix('#').unwrap_or(l))
+            .join("\n");
+
+        let mut cfg: TiKvConfig = toml::from_str(&template_config).unwrap();
+        let mut default_cfg = TiKvConfig::default();
+
+        // Some default values are computed based on the environment.
+        // Because we can't set config values for these in `config-template.toml`, we will handle
+        // them manually.
+        cfg.readpool.unified.max_thread_count = default_cfg.readpool.unified.max_thread_count;
```
andylokandy, line 3834: What is the value in `cfg` if not set in example, why does it differ from the default?

Jibbow: You are right, if they are not set in `config-template.toml`, they don't differ from the default.
When I wrote that code looked through *all* `Default` implementations on config items and added special rules for all computed ones, even if they are not set explicitly in `config-template.toml`. The idea was to avoid failures when somebody sets them in `config-template.toml` in the future.
If you want, we can also remove them because they don't make any difference.

andylokandy: > The idea was to avoid failures when somebody sets them in config-template.toml in the future

I think this is intended: to fail if the template is not relflecting the default value.

Jibbow: hm okay I see.

Just to wrap things up:
 * yes, the test should fail if a config differs from the `Default` implementation.
 * some configs are computed. thus, a fixed value in `config-template.toml` might be equal to the `Default` value on some machines, but not on others. This would lead to test failures on some machines. So we just overwrite these computed values.
 * if a value is missing in `config-template.toml`, it falls back to the `Default` value.

In this case, `max_thread_count` is set to `8` in `config-template.toml`, but actually it's a computed value that differs from machine to machine. The idea was to get rid of spurious test failures by just "ignoring" the value from `config-template.toml` and overwriting it with `Default`. The downside of this approach is that `max_thread_count` could be set to _anything_ in the config file.

But you are right: some of the computed values are **not** set in `config-template.toml` and they will then be set to `Default` while parsing. Thus, we wouldn't need to overwrite them in the test. However, if someone decides to add them to `config-template.toml` because he or she wants to document it, it might introduce test failures on some machines because it's a computed value.
So all in all I would actually keep these "overwrites" for **all** computed values in the test. But also I might lack the experience of judging this well, so please feel free to correct me and we can find a different solution :)

andylokandy: Oh! it sounds reasonable, or else we will not be able to list computed config in the example.

andylokandy (approved):

PR #10020, opened 2021-04-14
make doc great again

andylokandy:

`make doc` is broken. This PR will fix it.

### Release note 
- No release note

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,7 @@ format: pre-format
 	cargo fmt
 
 doc:
-	@cargo doc --workspace --document-private-items \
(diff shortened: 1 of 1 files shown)


zhongzc (approved):

sticnarf:
It looks really weird that `make doc` requires a toolchain different from the one defined in `rust-toolchain`.

I think we should update the toolchain of TiKV to an approriate version to solve the problem.

andylokandy:
@sticnarf Well you may be right, since https://github.com/rust-lang/rust/pull/83412 is merge.

PR #10022, opened 2021-04-14
txn_types: ignore all locks if get with MaxU64 has resolved locks

sticnarf:

It is reported that https://github.com/pingcap/tidb/pull/22789 causes some regression because more CPU is used in TiDB.

I **limit the TiDB CPU to 400%** and run sysbench point select and it shows the regression. But https://github.com/pingcap/tidb/pull/22789 optimizes point select in conflicted scenarios. (The conflicted scenaros is running sysbench point select and update_index with zipfian distribution at the same time)

图片
图片

I propose to revert https://github.com/pingcap/tidb/pull/22789 and use this PR to improve conflicted scenarios.

If the reading request is using max u64 as the snapshot ts and it brings some resolved locks, it must have resolved the first lock it met. In this case, other locks should never block this reading request.

The benchmark result is as follows.

图片
图片

The effect is not as good as the previous solution, but does not affect point select with no conflicts.

### Related changes

### Check List 

Tests 

### Release note 

- Bypass all locks when MaxU64 get has resolved any lock.

### DIFF
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -347,6 +347,14 @@ impl Lock {
             return Ok(());
         }
 
+        // When ts is `TimeStamp::max()` and `bypass_locks` is not empty, the lock
+        // read for the first time has been resolved (or whose min_commit_ts has been
+        // advanced). And because other new locks should not block such reading requests,
+        // we can ignore all seen locks in this case.
+        if ts == TimeStamp::max() && !bypass_locks.is_empty() {
+            return Ok(());
+        }
+
         let raw_key = key.to_raw()?;
 
         if ts == TimeStamp::max() && raw_key == lock.primary && !lock.use_async_commit {
@@ -683,6 +691,15 @@ mod tests {
         Lock::check_ts_conflict(Cow::Borrowed(&lock), &key, 140.into(), &empty).unwrap();
         Lock::check_ts_conflict(Cow::Borrowed(&lock), &key, 150.into(), &empty).unwrap_err();
         Lock::check_ts_conflict(Cow::Borrowed(&lock), &key, 160.into(), &empty).unwrap_err();
+
+        // Reading with `TimeStamp::max()` and non-empty `bypass_locks` can bypass all locks.
+        Lock::check_ts_conflict(
+            Cow::Borrowed(&lock),
+            &key,
+            160.into(),
+            &TsSet::from_u64s(vec![100]),
+        )
+        .unwrap();
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


components/txn_types/src/lock.rs
```diff
@@ -347,6 +347,14 @@ impl Lock {
             return Ok(());
         }
 
+        // When ts is `TimeStamp::max()` and `bypass_locks` is not empty, the lock
+        // read for the first time has been resolved (or whose min_commit_ts has been
+        // advanced). And because other new locks should not block such reading requests,
+        // we can ignore all seen locks in this case.
+        if ts == TimeStamp::max() && !bypass_locks.is_empty() {
```
MyonKeminta, line 354: So  this doesn't take effect when the first lock it meets is an async commit lock whose  min_commit_ts can't be pushed?

MyonKeminta (approved):
Looks correct. LGTM.

components/txn_types/src/lock.rs
sticnarf: 🤔 Looks strange. As what you said, this PR should not have much improvement, since I didn't modify TiDB to put resolved async commit locks in it.

youjiali1995: What's the conclusion?

components/txn_types/src/lock.rs
```diff
@@ -683,6 +691,15 @@ mod tests {
         Lock::check_ts_conflict(Cow::Borrowed(&lock), &key, 140.into(), &empty).unwrap();
         Lock::check_ts_conflict(Cow::Borrowed(&lock), &key, 150.into(), &empty).unwrap_err();
         Lock::check_ts_conflict(Cow::Borrowed(&lock), &key, 160.into(), &empty).unwrap_err();
+
+        // Reading with `TimeStamp::max()` and non-empty `bypass_locks` can bypass all locks.
+        Lock::check_ts_conflict(
+            Cow::Borrowed(&lock),
+            &key,
+            160.into(),
+            &TsSet::from_u64s(vec![100]),
```
youjiali1995, line 700:
```suggestion
            TimeStamp::max(),
            &TsSet::from_u64s(vec![101]),
```

components/txn_types/src/lock.rs
sticnarf: Use https://github.com/pingcap/tidb/pull/24084 to implement it instead.

With that solution, it is possible (but rare) that the retries keep encountering different locks. If we want to solve this problem, I think a explicit protocol is preferred instead of reusing `bypass_locks`.

sticnarf:
Use pingcap/tidb#24084  instead.

PR #10025, opened 2021-04-15
importer: auto split when import end

Little-Wallace:

Problem Summary:
If region split during import, the sst_importer::write rpc has to retry and it may hurt performance of import. And lightning may import a lot of data to one region, this may cause a large region. So I hope TiKV can split the region when import end.

TiKV will check the import mode of importer before scheduling a split-check task.  If the approximate size is too large, TiKV will schedule a split-check task.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU.  There may be a lot of split-check task after import task ending, and these tasks will cost more CPU.

### Release note 
- No release note.


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3546,6 +3560,10 @@ where
             return;
         }
 
+        if self.ctx.importer.get_mode() == SwitchMode::Import {
```
overvenus, line 3563: Please add some comments about skipping the split check.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3584,7 +3607,7 @@ where
         }
         self.fsm.peer.size_diff_hint = 0;
         self.fsm.peer.compaction_declined_bytes = 0;
-        self.register_split_region_check_tick();
+        return true;
```
overvenus, line 3610:
```suggestion
       true
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3560,6 +3575,13 @@ where
             return;
         }
 
+        // When Lightning or BR is importing data to TiKV, their ingest-request may fail because of
+        // region-epoch not matched. So we hope TiKV do not check region size and split region during
+        // importing.
+        if self.ctx.importer.get_mode() == SwitchMode::Import {
```
Connor1996, line 3581: Move it to the line between L3595 and L3596. For most cases, the check will be returned at the check of size_diff_hint, moving the import mode check can save unnecessary atomic load operations which involves memory fence.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -523,6 +524,20 @@ where
...
+                PeerMsg::SplitCheck => {
+                    // If the current `approximate_size` is larger than `region_max_size`, this region
+                    // may ingest a large file which is imported by `BR` or `lightning`, we shall check it
+                    let approximate_size =
+                        self.fsm.peer.approximate_size.unwrap_or(u64::max_value());
+                    let approximate_keys =
+                        self.fsm.peer.approximate_keys.unwrap_or(u64::max_value());
+                    if self.fsm.peer.is_leader()
+                        && (approximate_size > self.ctx.coprocessor_host.cfg.region_max_size.0
+                            || approximate_keys > self.ctx.coprocessor_host.cfg.region_max_keys)
```
Connor1996, line 536:
```suggestion
                    if self.fsm.peer.is_leader()
                        && (self.fsm.peer.approximate_size.map_or(true, |v| v > self.ctx.coprocessor_host.cfg.region_max_size.0) 
                            || self.fsm.peer.approximate_keys.map_or(true, |v| v > self.ctx.coprocessor_host.cfg.region_max_keys))
                        && (approximate_size > self.ctx.coprocessor_host.cfg.region_max_size.0
                            || approximate_keys > self.ctx.coprocessor_host.cfg.region_max_keys)
```

Connor1996 (commented):
rest LGTM

Connor1996 (approved):

PR #10030, opened 2021-04-16
coprv2: plugin version checks

Jibbow:

Introduces version checks for API version and rustc version of plugins.

Issue Number: #9747 

Problem Summary:

When loading a coprocessor plugin, we need to make sure that it actually runs. 
Thus, we need to make sure that the following things match up:
 * the version of the `coprocessor_plugin_api` crate the plugin was compiled with.
 * the rustc version the plugin was compiled with (has to be the same as the rustc version TiKV was compiled with because Rust does not guarantee a stable ABI)
 * the `target` platform the plugin was compiled for and the target platform of TiKV

All in all, it is very similar to the [POC](https://github.com/andylokandy/plugin).

Proposal: [Coprocessor V2 RFC](https://github.com/andylokandy/rfcs/blob/plugin/text/2021-02-24-coprocessor-plugin.md) 

What's Changed:

### Related changes

- based on PR #9925 

### Check List 

Tests 

- Unit test (TODO: I'm not entirely sure how to test this, help appreciated!)

### Release note 

 * No release note.

### DIFF
--- a/src/coprocessor_v2/plugin_registry.rs
+++ b/src/coprocessor_v2/plugin_registry.rs
@@ -9,25 +9,36 @@ use std::path::{Path, PathBuf};
 use std::sync::{mpsc, Arc, RwLock};
 use std::thread;
 use std::time::Duration;
+use thiserror::Error;
 
-quick_error! {
-    #[derive(Debug)]
-    pub enum PluginLoadingError {
-        Dylib(err: DylibError) {
-            from()
-            cause(err)
-            display("{}", err)
-        }
-        CompilerMismatch(plugin_rustc_version: String, tikv_rustc_version: String) {
-            display("version mismatch of rustc: plugin was compiled with {}, but TiKV with {}", plugin_rustc_version, tikv_rustc_version)
-        }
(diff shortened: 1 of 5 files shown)


Jibbow:
Not entirely sure why the build fails. Any ideas?

skyzh:
Seems that rustc ICE... Is there any way to clear cache on CI? @andylokandy

skyzh:
Also I saw `<jemalloc>: Invalid conf pair: prof:true`. Seems to be caused by jemalloc?

andylokandy:
Cleaning the cache should work...

Jibbow:
Nice, thanks!

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -10,6 +10,51 @@ use std::sync::{mpsc, Arc, RwLock};
 use std::thread;
 use std::time::Duration;
 
+quick_error! {
```
andylokandy, line 13: Prefer to use `anyhow` and `thiserror`.

PR #10048, opened 2021-04-20
copr: fix cast real, decimal to time bug on overflow

### LINKED ISSUE #10045: cast function result not same as MySQL

master branch

x86_64

### Steps to reproduce

```sql
mysql> create table t (a double);
mysql> insert into t values(-790822912);
mysql> select a from t where cast(a as time);
+------------+
| a          |
+------------+
| -790822912 |
+------------+
1 row in set, 1 warning (0.01 sec)
```

### What did you expect?

Empty result set.

### What did happened?

Related with https://github.com/pingcap/tidb/issues/23993, when cast Int, Real or Decimal into Time if there has Truncated error the result should be NULL.

blacktear23:

Issue Number: close #10045 

Problem Summary:

cast real, decimal to time if overflow or truncated should return NULL.

What's Changed:

Let cast real and decimal to time return NULL if truncated.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- No

### Release note 

- *: fix cast from real or decimal to time bug.


components/tidb_query_expr/src/impl_cast.rs
```diff
@@ -998,9 +998,15 @@ fn cast_bytes_like_as_duration(
     ctx: &mut EvalContext,
     extra: &RpnFnCallExtra,
     val: &[u8],
+    overflow_return_null: bool,
```
lzmhhh123, line 1001:
```suggestion
    overflow_as_null: bool,
```

components/tidb_query_datatype/src/codec/mysql/duration.rs
```diff
@@ -429,14 +433,25 @@ impl Duration {
     /// See: <http://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html>
     pub fn parse(ctx: &mut EvalContext, input: &str, fsp: i8) -> Result<Duration> {
         let fsp = check_fsp(fsp)?;
-        parser::parse(ctx, input, fsp, true)
+        parser::parse(ctx, input, fsp, true, false)
+            .ok_or_else(|| Error::truncated_wrong_val("TIME", input))
+    }
+
+    pub fn parse2(
```
lzmhhh123, line 440:
```suggestion
    pub fn parse_consider_overflow(
```

Why not merge it with `parse`?

blacktear23: this will break a lot of test code. If rust has parameter default value like python, Ithink it will be the best way, but not. And also no override. So I just create another function.

lzmhhh123 (approved):

skyzh (approved):
LGTM. By the way, I would suggest changing the PR title to "copr: ...." as this is indeed part of TiKV coprocessor.

At the same time, I would also suggest adding test case on `cast_real_as_duration` and `cast_decimal_as_duration`.

blacktear23:
@skyzh PTAL

skyzh:
/retitle copr: fix cast real, decimal to time bug on overflow

skyzh:
@blacktear23 Please merge the PR at TiDB side first. Otherwise, copr-integration test won't pass.

blacktear23:
@skyzh I think we should cherry-pick this PR to 5.0 branch.

PR #10050, opened 2021-04-20
cdc: limit old value cache size with memory capacity

5kbpers:

Problem Summary:
For now, the old value cache has only a count bound whose actual size is unpredictable.

* Add a `CacheSizeTrace` abstraction to `LruCache`
* Calculate the actual size in memory to limit the old value cache size

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* cdc: limit old value cache size with memory capacity


src/config.rs
```diff
@@ -2275,16 +2275,23 @@ impl Default for BackupConfig {
 #[serde(rename_all = "kebab-case")]
 pub struct CdcConfig {
     pub min_ts_interval: ReadableDuration,
+    // Deprecated! These configuration has been moved to Coprocessor.
+    // They are preserved for compatibility check.
+    #[doc(hidden)]
+    #[serde(skip_serializing)]
     pub old_value_cache_size: usize,
+    pub old_value_cache_mem_capacity: OptionReadableSize,
```
5kbpers, line 2283: I add a new option for the memory capacity. @overvenus @liuzix

src/config.rs
```diff
@@ -2275,16 +2275,23 @@ impl Default for BackupConfig {
 #[serde(rename_all = "kebab-case")]
 pub struct CdcConfig {
     pub min_ts_interval: ReadableDuration,
+    // Deprecated! These configuration has been moved to Coprocessor.
```
overvenus, line 2278: It's not a coprocessor.

components/tikv_util/src/lru.rs
```diff
@@ -132,90 +133,131 @@ impl<K> Trace<K> {
...
+    }
+}
+
+impl<K, V> Default for CountTrace<K, V> {
+    fn default() -> Self {
+        Self(0, PhantomData::default())
     }
+}
+
+pub struct RawLruCache<K, V, T> {
```
overvenus, line 169: `PhantomData` is unnecessary and we can replace LruCache type alias by [_Default Generic Type Parameters_](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#default-generic-type-parameters-and-operator-overloading).
```suggestion
pub struct CountTrace<K, V>(pub u64);

impl<K, V> CacheSizeTrace<K, V> for CountTrace<K, V> {
    fn current_size(&self) -> u64 {
        self.0
    }

    fn trace_insert(&mut self, _: &K, _: &V) {
        self.0 += 1;
    }

    fn trace_remove(&mut self, _: &K, _: &V) {
        self.0 -= 1;
    }

    fn set(&mut self, val: u64) {
        self.0 = val;
    }
}

impl Default for CountTrace {
    fn default() -> Self {
        Self(0)
    }
}

pub struct LruCache<K, V, T: CacheSizeTrace<K, V> = CountTrace> {
```

components/cdc/src/old_value.rs
```diff
@@ -0,0 +1,377 @@
...
+        .with_label_values(&["seek"])
+        .observe(start.elapsed().as_secs_f64());
+    (value, Some(statistics))
+}
+
+pub(crate) type OldValueCallback =
+    Box<dyn Fn(Key, TimeStamp, &mut OldValueCache) -> (Option<Vec<u8>>, Option<Statistics>) + Send>;
+
+#[derive(Default)]
+struct OldValueCacheSizeTrace(u64);
```
overvenus, line 70: It should be a signed integer, otherwise, we may observe an increasing curve in dashboards.

components/tikv_util/src/lru.rs
5kbpers: Default generic type may not work very well because we still need to indicate the type parameter explicitly when constructing an object or implementing some methods.
IMO defining a `Raw` type to distinguish the default manner and the custom manner is a good pattern.

components/cdc/src/old_value.rs
5kbpers: I think expectedly it should never be a negative number.

tests/integrations/config/mod.rs
```diff
@@ -659,8 +659,9 @@ fn test_serde_custom_tikv_config() {
     };
     value.cdc = CdcConfig {
         min_ts_interval: ReadableDuration::secs(4),
-        old_value_cache_size: 512,
+        old_value_cache_mem_capacity: OptionReadableSize(Some(ReadableSize::gb(1))),
```
overvenus, line 662:
```suggestion
        old_value_cache_memory_quota: OptionReadableSize(Some(ReadableSize::gb(1))),
```

overvenus:
Can we close this PR in favor of #10252?

overvenus:
A successor PR #10252 is merged.

PR #10057, opened 2021-04-21
statistics: introduce full sampling collect tech

winoros:

Problem Summary:

This pr implement the things needed in https://github.com/pingcap/tidb/pull/24089

Proposal: [xxx](url) 

What's Changed:

We assigned a random weight to each row. And then select the top n weighted rows as the sampling result.
Since the weights are assigned randomly. Merge them will not lose the randomness of the sampling.

### Related changes

- https://github.com/pingcap/tidb/pull/24089

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 
- No release note


skyzh (commented):
Generally looks good to me. I'll take some time to read through the sampling algorithm and come back for review again. Please re-request my review after tipb PR has been merged, and lint errors are fixed.

src/coprocessor/statistics/analyze.rs
```diff
@@ -271,6 +291,230 @@ impl<S: Snapshot> RequestHandler for AnalyzeContext<S> {
...
+                return;
+            }
+            self.fm_sketches[i].insert(&columns_val[i]);
+            self.total_sizes[i] += columns_val[i].len() as i64;
+        }
+        self.sampling(columns_val);
+    }
+
+    pub fn sampling(&mut self, data: Vec<Vec<u8>>) {
+        let cur_rng = self.rng.gen_range(0, i64::MAX);
```
skyzh, line 480: Could you explain how this correlates to https://en.wikipedia.org/wiki/Reservoir_sampling#Algorithm_A-Res ? 
```
r := random() ^ (1/S.Weight)
```
Are we assigning the same weight to all rows?

winoros: I think this is just the range interval is different. One is (0, 1), the other is [0, MaxInt64].
The range interval should not affect too much.

> Are we assigning the same weight to all rows?

We are assigning random value to each row.

skyzh: What confused me is `^ (1/S.Weight)` part. From the code, I guess `1/S.Weight` is selected to be 1 in this case?

winoros: Yes, we choose this one since assigning random values can guarantee randomness in distributed cases.

src/coprocessor/statistics/analyze.rs
```diff
@@ -271,6 +291,232 @@ impl<S: Snapshot> RequestHandler for AnalyzeContext<S> {
...
+        self.row_buf.clear();
+        let col_len = columns_val.len();
+        for i in 0..column_groups.len() {
+            let offsets = column_groups[i].get_column_offsets();
+            let mut has_null = true;
+            for j in 0..offsets.len() {
+                if columns_val[j][0] != NIL_FLAG {
+                    has_null = false
+                }
+                self.total_sizes[col_len + i] += columns_val[i].len() as i64
```
skyzh, line 454: The Golang version is `s.TotalSizes[colLen+i] += int64(len(cols[c].GetBytes())) - 1`. Why not minus one here?

src/coprocessor/statistics/analyze.rs
```diff
@@ -271,6 +291,232 @@ impl<S: Snapshot> RequestHandler for AnalyzeContext<S> {
...
+            row_buf: Vec::new(),
+        }
+    }
+
+    pub fn collect_column_group(
+        &mut self,
+        columns_val: &[Vec<u8>],
+        column_groups: &[tipb::AnalyzeColumnGroup],
+    ) {
+        self.row_buf.clear();
```
skyzh, line 445: Should this be moved into the inner loop?

```golang
	for i, group := range colGroups {
 		datumBuffer = datumBuffer[:0]
 		hasNull := true
```

src/coprocessor/statistics/analyze.rs
```diff
@@ -271,6 +291,232 @@ impl<S: Snapshot> RequestHandler for AnalyzeContext<S> {
...
+    }
+
+    pub fn collect_column(&mut self, columns_val: Vec<Vec<u8>>) {
+        for i in 0..columns_val.len() {
+            if columns_val[i][0] == NIL_FLAG {
+                self.null_count[i] += 1;
+                return;
+            }
+            self.fm_sketches[i].insert(&columns_val[i]);
+            self.total_sizes[i] += columns_val[i].len() as i64;
```
skyzh, line 476: ditto for possible missing `-1`.

src/coprocessor/statistics/analyze.rs
```diff
@@ -271,6 +291,232 @@ impl<S: Snapshot> RequestHandler for AnalyzeContext<S> {
...
+            }
+            self.fm_sketches[col_len + i].insert(&self.row_buf);
+        }
+    }
+
+    pub fn collect_column(&mut self, columns_val: Vec<Vec<u8>>) {
+        for i in 0..columns_val.len() {
+            if columns_val[i][0] == NIL_FLAG {
+                self.null_count[i] += 1;
+                return;
```
skyzh, line 473: Should this be `continue`?

skyzh (commented):
I'm not familiar with the statistics part, so I simply checked if this patch functions the same as the Golang version. Point me out if I misunderstood any part. Please re-request my review after fixing clippy warnings.

src/coprocessor/statistics/analyze.rs
winoros: There's a tiny difference between the two version.
For golang version, it called the `DecodeColumnValue` which called `DecodeOne` first then called the `Unflatten`. In the `DecodeOne` you can see that the first byte in the slice means the `type` information.
In rust version, it directly called a `unflatten` in the `decode_col_value` and judge the `type` by the given `ColumnInfo` so the byte slice passed here contains no `type` information. Thus we don't minus one in the rust version.

src/coprocessor/statistics/analyze.rs
winoros: Same.

skyzh (approved):

src/coprocessor/statistics/analyze.rs
```diff
@@ -271,6 +291,232 @@ impl<S: Snapshot> RequestHandler for AnalyzeContext<S> {
...
+        self.sampling(columns_val);
+    }
+
+    pub fn sampling(&mut self, data: Vec<Vec<u8>>) {
+        let cur_rng = self.rng.gen_range(0, i64::MAX);
+        if self.samples.len() < self.max_sample_size {
+            self.samples.push(Reverse((cur_rng, data)));
+            return;
+        }
+        if self.samples.len() == self.max_sample_size && self.samples.peek().unwrap().0.0 < cur_rng
```
lzmhhh123, line 487: The `samples` is a max value heap? Here should be a min value heap?

skyzh: Rust std's heap is by default max-value heap. As we wrap the original type with `Reverse`, it is now a min heap.

lzmhhh123: If we keep the max value heap. All the value which is less than the max value in heap will never be replaced?

lzmhhh123: OK. Got it.

skyzh: * Our goal is to get maximum k elements.
* The heap stores maximum k elements of first n elements, and as the heap is a min-heap, the minimum one will be at the top.
* If the next element is larger than minimum of current maximum k elements, then pop current top and add new one.

lzmhhh123 (approved):

PR #10058, opened 2021-04-21
Add a WriteBatch test to engine_traits_tests

brson:

Capturing the interaction between save points and the command count in write batches.

### Release note 

- No release note


brson:
This branch also contains improvements to assertions and an added comment about some strange rocksdb behavior with write batches plus bogus range deletes.

brson:
I'll just include these changes in a future engine_traits_tests patch

PR #10060, opened 2021-04-22
cdc: Refactor `EventBatcherSink`

liuzix:
- Refactored ``EventBatcherSink`
- Fixed a potential violation of the calling convention of future-like functions.

- Redesigned `EventBatcherSink` with an internal state that's easy to reason about.
- Ensured that tasks blocked while polling the inner sink poll the inner sink again once awaken.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 
- Optimize CDC sink

### DIFF
--- a/components/cdc/src/service.rs
+++ b/components/cdc/src/service.rs
@@ -170,8 +170,8 @@ enum EventBatcherSinkState {
 }
 
 pub struct EventBatcherSink<S, E>
-    where
-        S: Sink<(ChangeDataEvent, grpcio::WriteFlags), Error = E> + Send + Unpin,
+where
+    S: Sink<(ChangeDataEvent, grpcio::WriteFlags), Error = E> + Send + Unpin,
 {
     // the internal state machine of this sink
     state: EventBatcherSinkState,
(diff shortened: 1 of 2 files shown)


components/cdc/src/service.rs
```diff
@@ -175,27 +185,33 @@ where
...
+                EventBatcherSinkState::Sending(ref mut send_buf) => {
+                    batcher
+                        .build()
+                        .into_iter()
+                        .for_each(|event| send_buf.push_back(event));
+                }
+                _ => unreachable!(),
+            }
+        } else {
+            panic!("cdc unexpected EventBatcherSinkState");
```
amyangfei, line 213: can log the `old_state` here

components/cdc/src/service.rs
```diff
@@ -205,30 +221,28 @@ where
...
-            return Poll::Ready(Ok(()));
+        match this.state {
+            EventBatcherSinkState::Ready(ref buf) => {
+                if buf.len() >= 2 {
+                    // converts CdcEvent to ChangeDataEvent
+                    this.prepare_flush();
+                    ready!(this.poll_flush_unpin(cx))?;
+                }
+            }
+            EventBatcherSinkState::Sending(_) | EventBatcherSinkState::Flushing => {
```
amyangfei, line 240: Can we combine the state `Sending` and `Flushing` into one state

liuzix: Since the `Sending` state is associated with a `deque` but the `Flushing` state is not, combining these two will likely reduce code readability.

PR #9872, opened 2021-03-23
cdc: fix panic caused by insert

5kbpers:
Signed-off-by: 5kbpers <tangminghua@pingcap.com>

### What problem does this PR solve?

Problem Summary:

If an insert operation has an old value, tikv would be panic by an assertion.

### What is changed and how it works?

Delete the assertion.

What's Changed:

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A

PR #10072, opened 2021-04-23
cdc, txn: improve CDC old value cache hit ratio in pessimistic txn

### LINKED ISSUE #10032: CDC old value cache does not work effectively for caching pessimistic txn write
## Bug Report

0% old value cache hit for pessimistic transaction write.

This issue may relate to https://github.com/tikv/tikv/issues/9996 and https://github.com/tikv/tikv/issues/9981.

The left is optimistic transaction writes and the right is pessimistic transaction writes.
image
image

master and v5.0.0

### Steps to reproduce

```
# create a changefeed with old value feature.
sysbench --db-driver=mysql --mysql-host=172.16.5.33 --mysql-port=47904 --mysql-user=root --time=15000 --report-interval=10 --threads=64 --mysql-db=test --tables=1 --table-size=6000000 oltp_write_only prepare
sleep 60 && sysbench --db-driver=mysql --mysql-host=172.16.5.33 --mysql-port=47904 --mysql-user=root --time=15000 --report-interval=10 --threads=64 --mysql-db=test --tables=1 --table-size=6000000 oltp_write_only run
```

### What did you expect?

Old value cache works effectively.

overvenus:

Issue Number: close #10032 
Problem Summary: Improve old value cache hit ratio in pessimistic txn.

Get old value in acquire pessimistic lock stage and insert it into old value cache.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

image
image

```
# create a changefeed with old value enabled.
sysbench --db-driver=mysql --mysql-host=172.16.5.33 --mysql-port=47904 --mysql-user=root --time=15000 --report-interval=10 --threads=64 --mysql-db=test --tables=1 --table-size=6000000 oltp_write_only prepare && \
sleep 60 && sysbench --db-driver=mysql --mysql-host=172.16.5.33 --mysql-port=47904 --mysql-user=root --time=15000 --report-interval=10 --threads=64 --mysql-db=test --tables=1 --table-size=6000000 oltp_write_only run
```

### Release note 

- Improve CDC old value cache hit ratio in pessimistic txn


components/cdc/src/observer.rs
```diff
@@ -130,32 +130,34 @@ impl<E: KvEngine> CmdObserver<E> for CdcObserver {
...
-                        MutationType::Put | MutationType::Delete => match old_value {
-                            OldValue::None => (None, None),
-                            OldValue::Value {
-                                start_ts,
-                                short_value,
-                            } => {
-                                let mut statistics = None;
-                                let value = short_value.or_else(|| {
-                                    statistics = Some(Statistics::default());
+                        // Old value of an Insert is guaranteed to be None.
```
5kbpers, line 133: Sometimes it is not true, the type of mutation is decided by TiDB, see https://github.com/tikv/tikv/pull/9872

components/txn_types/src/types.rs
```diff
@@ -347,10 +346,7 @@ impl OldValue {
 
     pub fn size(&self) -> usize {
         let value_size = match self {
-            OldValue::Value {
-                short_value: Some(v),
-                ..
-            } => v.len(),
+            OldValue::Value { value } => value.len(),
```
5kbpers, line 349: The size of `ValueTimeStamp` is not zero.

5kbpers (commented):
rest LGTM

components/txn_types/src/types.rs
overvenus: `std::mem::size_of::<OldValue>()` computes OldValue enum size, inlcuding `ValueTimeStamp`.

src/storage/txn/actions/acquire_pessimistic_lock.rs
```diff
@@ -73,10 +114,17 @@ pub fn acquire_pessimistic_lock<S: Snapshot>(
...
+        return Ok((val, old_value));
     }
 
+    // Following seek_write read the previous write.
+    let (prev_write_loaded, mut prev_write) = (true, None);
     if let Some((commit_ts, write)) = reader.seek_write(&key, TimeStamp::max())? {
+        // Find a previous write.
+        if need_old_value {
+            prev_write = Some(write.clone());
+        }
```
MyonKeminta, line 126: There should be some way to avoid the cloning by moving `write` to `prev_write` and use `prev_write` instead of `write` in the if block. But it's ok to me since the current code looks clear.

MyonKeminta (approved):
LGTM

src/storage/txn/actions/acquire_pessimistic_lock.rs
overvenus: It is possible to avoid clone when need_value is false, though it may make code more complicated.

PR #10075, opened 2021-04-23
coprv2: plugin version checks

Jibbow:

Introduces version checks for API version and rustc version of plugins.

Issue Number: #9747 

Problem Summary:

When loading a coprocessor plugin, we need to make sure that it actually runs. 
Thus, we need to make sure that the following things match up:
 * the version of the `coprocessor_plugin_api` crate the plugin was compiled with.
 * the rustc version the plugin was compiled with (has to be the same as the rustc version TiKV was compiled with because Rust does not guarantee a stable ABI)
 * the `target` platform the plugin was compiled for and the target platform of TiKV

All in all, it is very similar to the [POC](https://github.com/andylokandy/plugin).

Proposal: [Coprocessor V2 RFC](https://github.com/andylokandy/rfcs/blob/plugin/text/2021-02-24-coprocessor-plugin.md) 

What's Changed:

### Related changes

- based on PR #9925 

### Check List 

Tests 

- Unit test (TODO: I'm not entirely sure how to test this, help appreciated!)

### Release note 

 * No release note.

### DIFF
--- a/components/coprocessor_plugin_api/build.rs
+++ b/components/coprocessor_plugin_api/build.rs
@@ -1,7 +1,6 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
 fn main() {
-    println!("cargo:rerun-if-changed=*");
     println!("cargo:rustc-env=API_VERSION={}", env!("CARGO_PKG_VERSION"));
     println!(
         "cargo:rustc-env=TARGET={}",
--- a/components/coprocessor_plugin_api/build.rs
+++ b/components/coprocessor_plugin_api/build.rs
@@ -1,0 +1,14 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+fn main() {
+    println!("cargo:rerun-if-changed=*");
+    println!("cargo:rustc-env=API_VERSION={}", env!("CARGO_PKG_VERSION"));
+    println!(
+        "cargo:rustc-env=TARGET={}",
+        std::env::var("TARGET").unwrap()
+    );
+    println!(
+        "cargo:rustc-env=RUSTC_VERSION={}",
+        rustc_version::version_meta().unwrap().short_version_string
+    );
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,7 @@ checksum = "59206260f98d163b3ca42fb29fe551dbcda1d43cf70a244066b2a0666a8fb2a9"
 dependencies = [
  "cc",
  "clap",
- "rustc_version",
+ "rustc_version 0.2.3",
  "xdg",
 ]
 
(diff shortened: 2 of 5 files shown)


skyzh:
What about do some hacks?

1. First modify the `test` section of Makefile by adding `cargo clean`. Wait until build succeed. Now the build cache of current branch should be updated.
2. Then revert the Makefile and build again.

Which may solve this ICE issue...

skyzh (approved):
LGTM. Let's hopefully wait until CI fixes... I'll also try solving this ICE by splitting this PR in this week.

Jibbow:
> LGTM. Let's hopefully wait until CI fixes... I'll also try solving this ICE by splitting this PR in this week.

I tried splitting the PR, but even the first commit fails (see https://github.com/tikv/tikv/pull/10079) :(

components/coprocessor_plugin_api/build.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+fn main() {
+    println!("cargo:rerun-if-changed=*");
```
skyzh, line 4: This line is causing TiKV binaries to be rebuilt after test binaries are built. I'm wondering why we need to explicitly specify `cargo:rerun-if-changed=*`

components/coprocessor_plugin_api/build.rs
```diff
@@ -0,0 +1,14 @@
...
+    println!("cargo:rustc-env=API_VERSION={}", env!("CARGO_PKG_VERSION"));
```
skyzh, line 5: If we intended to get "the version of the coprocessor_plugin_api crate the plugin was compiled with", we should move `env!("CARGO_PKG_VERSION")` inside the main program. According to https://doc.rust-lang.org/cargo/reference/environment-variables.html , seems that this env variable is not expected to be read in `build.rs`.

skyzh (approved):
Good work!

components/coprocessor_plugin_api/build.rs
andylokandy: > I'm wondering why we need to explicitly specify cargo:rerun-if-changed=*

This obviously comes from the POC. It's for forcing the cargo the recompile`plugin_api` after it's edited. Because at that time `plugin_api` carried the source hash of itself, it'll cause the hash mismatch if not recomplied.

components/coprocessor_plugin_api/build.rs
andylokandy: Seems true...

andylokandy: @Jibbow PTAL~

skyzh: I’ve tested and could get correct version number in build.rs, but this seems not to be documented...

PR #10080, opened 2021-04-25
copr: support cast enum to any type

wshwsh12:

Issue Number: close #xxx 

Problem Summary:
In pr https://github.com/tikv/tikv/pull/9648, it implements enum push down. But only implement cast enum to int/real/string. We need more types to support.

Proposal: [xxx](url) 

What's Changed:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

- No release note.


skyzh (approved):
Seems that `write_enum_to_chunk` is the same as `write_enum`. Is it necessary?

By the way, would you please explain a little bit about all these functions and when to use them? (Better in docstring)

Xuanwo (approved):

wshwsh12:
> Seems that `write_enum_to_chunk` is the same as `write_enum`. Is it necessary?
> 
Only `write_enum_to_chunk` is ok.

PR #10084, opened 2021-04-26
server: add stale read test infra and test cases

NingLin-P:

What's Changed:

Add helpful functions/structs for testing stale read
Add basic test cases for stale read

### Check List 

Tests 

- Integration test

### Release note 
- No release note

### DIFF
--- a/components/resolved_ts/src/observer.rs
+++ b/components/resolved_ts/src/observer.rs
@@ -16,6 +16,7 @@ use crate::endpoint::Task;
 pub struct Observer<E: KvEngine> {
     cmd_batches: RefCell<Vec<CmdBatch>>,
     scheduler: Scheduler<Task<E::Snapshot>>,
+    need_old_value: bool,
 }
 
 impl<E: KvEngine> Observer<E> {
@@ -23,7 +24,14 @@ impl<E: KvEngine> Observer<E> {
         Observer {
             cmd_batches: RefCell::default(),
             scheduler,
+            need_old_value: true,
         }
+    }
+
+    // Disable old value, currently only use in tests to avoid holding the snapshot
+    // and cause data can not be deleted
+    pub fn disable_old_value(&mut self) {
+        self.need_old_value = false;
     }
 
     pub fn register_to(&self, coprocessor_host: &mut CoprocessorHost<E>) {
@@ -45,6 +53,7 @@ impl<E: KvEngine> Clone for Observer<E> {
         Self {
             cmd_batches: self.cmd_batches.clone(),
             scheduler: self.scheduler.clone(),
+            need_old_value: self.need_old_value,
         }
     }
 }
@@ -74,8 +83,14 @@ impl<E: KvEngine> CmdObserver<E> for Observer<E> {
             region.mut_peers().push(Peer::default());
             // Create a snapshot here for preventing the old value was GC-ed.
             // TODO: only need it after enabling old value, may add a flag to indicate whether to get it.
-            let snapshot =
-                RegionSnapshot::from_snapshot(Arc::new(engine.snapshot()), Arc::new(region));
+            let snapshot = if self.need_old_value {
+                Some(RegionSnapshot::from_snapshot(
+                    Arc::new(engine.snapshot()),
+                    Arc::new(region),
+                ))
+            } else {
+                None
+            };
             if let Err(e) = self.scheduler.schedule(Task::ChangeLog {
                 cmd_batch: batches,
                 snapshot,
--- a/components/resolved_ts/src/resolver.rs
+++ b/components/resolved_ts/src/resolver.rs
@@ -17,7 +17,7 @@ pub struct Resolver {
     lock_ts_heap: BTreeMap<TimeStamp, HashSet<Arc<[u8]>>>,
     // The timestamps that guarantees no more commit will happen before.
     resolved_ts: TimeStamp,
-    // The region read progress used to utilize `resolved_ts` to serve stale read request
+    // The region read progress is used to utilize `resolved_ts` to serve stale read request
     read_progress: Arc<RegionReadProgress>,
     // The timestamps that advance the resolved_ts when there is no more write.
     min_ts: TimeStamp,
--- a/components/resolved_ts/src/resolver.rs
+++ b/components/resolved_ts/src/resolver.rs
@@ -1,9 +1,9 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
 use collections::{HashMap, HashSet};
+use raftstore::store::util::RegionReadProgress;
 use std::cmp;
 use std::collections::BTreeMap;
-use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use txn_types::TimeStamp;
 
@@ -16,7 +16,9 @@ pub struct Resolver {
     // start_ts -> locked keys.
     lock_ts_heap: BTreeMap<TimeStamp, HashSet<Arc<[u8]>>>,
     // The timestamps that guarantees no more commit will happen before.
-    resolved_ts: Arc<AtomicU64>,
+    resolved_ts: TimeStamp,
+    // The region read progress used to utilize `resolved_ts` to serve stale read request
+    read_progress: Arc<RegionReadProgress>,
     // The timestamps that advance the resolved_ts when there is no more write.
     min_ts: TimeStamp,
 }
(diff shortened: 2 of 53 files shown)


tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+    cluster.sim.wl().start_resolved_ts_worker();
+
+    cluster.must_transfer_leader(1, new_peer(1, 1));
+    let mut leader_client = PeerClient::new(&cluster, 1, new_peer(1, 1));
+    let mut follower_client2 = PeerClient::new(&cluster, 1, new_peer(2, 2));
+    // Set the `stale_read` flag
+    leader_client.ctx.set_stale_read(true);
+    follower_client2.ctx.set_stale_read(true);
+
+    // Filter read index request of follower 2, so it can't perform normal
```
nolouch, line 84: Why already set `stale read` need to filter "ReadIndex" msg,  can we check there is no `ReadIndex` msg?

components/resolved_ts/src/observer.rs
```diff
@@ -16,16 +16,24 @@ use crate::endpoint::Task;
...
         Observer {
             cmd_batches: RefCell::default(),
             scheduler,
+            need_old_value: true,
         }
     }
 
+    // Disable old value, currently only use in tests to avoid holding the snapshot
+    // and cause data can not be deleted
+    pub fn disable_old_value(&mut self) {
```
nolouch, line 33: If don't delete it, will big impact in the test? I think it is best not to invade the normal code logic.

nolouch (commented):
rest LGTM

tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+    // follower read
+    cluster.add_send_filter(CloneFilterFactory(
+        RegionPacketFilter::new(1, 2)
+            .direction(Direction::Send)
+            .msg_type(MessageType::MsgReadIndex),
+    ));
```
Yisaer, line 90: Why this is necessary?

tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+    let mut cluster = new_server_cluster(0, 3);
+    let pd_client = Arc::clone(&cluster.pd_client);
+    pd_client.disable_default_operator();
+
+    cluster.run();
+    cluster.sim.wl().start_resolved_ts_worker();
+
+    cluster.must_transfer_leader(1, new_peer(1, 1));
+    let mut leader_client = PeerClient::new(&cluster, 1, new_peer(1, 1));
+    let mut follower_client2 = PeerClient::new(&cluster, 1, new_peer(2, 2));
```
Yisaer, line 79:
```suggestion
    let mut follower_client = PeerClient::new(&cluster, 1, new_peer(2, 2));
```

tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    // Now we can read `value2` with the newest ts
+    follower_client2.must_kv_read_equal(
+        b"key1".to_vec(),
+        b"value2".to_vec(),
+        block_on(pd_client.get_tso()).unwrap().into_inner(),
+    );
+}
+
+// Testing how mvcc locks clould effect stale read service
```
Yisaer, line 148:
```suggestion
// Testing how mvcc locks could effect stale read service
```

tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+    let mut cluster = new_server_cluster(0, 3);
+    let pd_client = Arc::clone(&cluster.pd_client);
+    pd_client.disable_default_operator();
+
+    cluster.run();
+    cluster.sim.wl().start_resolved_ts_worker();
+
+    cluster.must_transfer_leader(1, new_peer(1, 1));
+    let leader_client = PeerClient::new(&cluster, 1, new_peer(1, 1));
+    let mut follower_client2 = PeerClient::new(&cluster, 1, new_peer(2, 2));
```
Yisaer, line 160: ditto.

tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+    follower_client2.ctx.set_stale_read(true);
+
+    // Disable read index request of follower 2, so it can't perform normal
+    // follower read
+    cluster.add_send_filter(CloneFilterFactory(
+        RegionPacketFilter::new(1, 2)
+            .direction(Direction::Send)
+            .msg_type(MessageType::MsgReadIndex),
+    ));
```
Yisaer, line 169: ditto.

components/resolved_ts/src/observer.rs
NingLin-P: There are few reasons to doing so, firstly we need to start the `resolved-ts` worker thread for our test case, but if we start it by default (for all test), the worker will panic because there are many test writing raw kv data (even directly writing to the transaction cf i.e. `write_cf`, `lock_cf`) and the worker will panic while parsing these raw data. So that we can't start the worker by default.

But even so, the apply worker will still send data to the `resolved-ts` observer, and if old value is enabled there will be a snapshot in the observer task, and since the `resolved-ts` worker is not start, the snapshot will not be dropped and causing data can't delete then some test cases (that expect data to be deleted) will fail.

tests/integrations/server/test_stale_read.rs
```diff
@@ -0,0 +1,231 @@
...
+        k2_prewrite_ts,
+    );
+    // Update `key1`
+    let commit_ts2 = leader_client.must_kv_write(
+        &pd_client,
+        vec![new_mutation(Op::Put, &b"key1"[..], &b"value2"[..])],
+        b"key1".to_vec(),
+    );
+
+    // Can not read `(key1, value2)` due to `commit_ts2` is larger than the `key2` lock's `start_ts`
```
Yisaer, line 192:
```suggestion
    // assert `(key1, value2)` can't be read with `commit_ts2` due to it's larger than then `start_ts` of `key2`.
```

tests/integrations/server/test_stale_read.rs
NingLin-P: > Why already set stale read need to filter "ReadIndex" msg

Just for robustness, to ensure that all read requests the follower performed are stale read requests

> can we check there is no ReadIndex msg?

We can add a fail point to panic if there are `ReadIndex` msg

components/resolved_ts/src/resolver.rs
```diff
@@ -16,7 +16,9 @@ pub struct Resolver {
     // start_ts -> locked keys.
     lock_ts_heap: BTreeMap<TimeStamp, HashSet<Arc<[u8]>>>,
     // The timestamps that guarantees no more commit will happen before.
-    resolved_ts: Arc<AtomicU64>,
+    resolved_ts: TimeStamp,
+    // The region read progress used to utilize `resolved_ts` to serve stale read request
```
Yisaer, line 20:
```suggestion
    // The region read progress is used to utilize `resolved_ts` to serve stale read request
```

components/resolved_ts/src/observer.rs
Yisaer: What about using `new_only_for_test` instead of `disable_old_value` ?

NingLin-P: > What about using new_only_for_test instead of disable_old_value ?

Don't seem necessary to me, `Observer::disable_old_value` and `Observer:: new` are use separately.

tests/integrations/server/test_stale_read.rs
NingLin-P: See https://github.com/tikv/tikv/pull/10084#discussion_r622985902

tests/integrations/server/test_stale_read.rs
NingLin-P: `follower_client2` stands for follower client of peer 2

tests/integrations/server/test_stale_read.rs
NingLin-P: It also tested whether the `stale_read` flag works properly.

nolouch: >> We can add a fail point to panic if there are ReadIndex msg
I think this is reasonable

components/resolved_ts/src/observer.rs
nolouch: > the worker will panic while parsing these raw data
you mean we do not support raw kv?

> the snapshot will not be dropped and causing data can't delete then some test cases
why cannot start resolved work deafaultly?

NingLin-P: > you mean we do not support raw kv?

Currently in the master branch, yes. It is because resolved worker tries to parse raw kv data as txn data,  but it is not by design and will be fixed in the next PR. 

> why cannot start resolved work in default?

"because there are many tests writing raw kv data (even directly writing to the transaction cf i.e. write_cf, lock_cf) and the worker will panic while parsing these raw data."

Some tests writing raw kv data into transaction cf (i.e. `write_cf`, `lock_cf`), it is hard for the resolved worker to distinguish test data and actually invalid data.

nolouch (approved):
lgtm

Yisaer (approved):
/lgtm

PR #10090, opened 2021-04-27
coprv2: check plugin version on coprocessor requests

Jibbow:

This PR add the ability to check plugin version on coprocessor requests.
We now collect the plugin version in `declare_plugin!` macro (either passed explicitly or from Cargo.toml).
When calling the new coprocessor endpoint, we now check whether the loaded plugin does satisfy `copr_version_constraint`.

Issue Number: #9747  

Tests 

- Unit test

### Release note 

 * No release note

### DIFF
--- a/components/coprocessor_plugin_api/src/util.rs
+++ b/components/coprocessor_plugin_api/src/util.rs
@@ -3,19 +3,24 @@
 use super::allocator::HostAllocatorPtr;
 use super::plugin_api::CoprocessorPlugin;
 
-/// Name of the exported constructor function for the plugin in the `dylib`.
+/// Name of the exported constructor with signature [`PluginConstructorSignature`] for the plugin.
 pub static PLUGIN_CONSTRUCTOR_SYMBOL: &[u8] = b"_plugin_create";
-/// Name of the exported function to get build information about the plugin.
+/// Name of the exported function with signature [`PluginGetBuildInfoSignature`] to get build
+/// information about the plugin.
 pub static PLUGIN_GET_BUILD_INFO_SYMBOL: &[u8] = b"_plugin_get_build_info";
+/// Name of the exported function with signature [`PluginGetPluginInfoSignature`] to get some
+/// information about the plugin.
+pub static PLUGIN_GET_PLUGIN_INFO_SYMBOL: &[u8] = b"_plugin_get_plugin_info";
 
-/// Type signature of the exported constructor function for the plugin in the `dylib`.
-/// See also [`PLUGIN_CONSTRUCTOR_SYMBOL`].
+/// Type signature of the exported function with symbol [`PLUGIN_CONSTRUCTOR_SYMBOL`].
 pub type PluginConstructorSignature =
     unsafe fn(host_allocator: HostAllocatorPtr) -> *mut dyn CoprocessorPlugin;
 
-/// Type signature of the exported function to get build information about the plugin.
-/// See also [`PLUGIN_GET_BUILD_INFO_SYMBOL`].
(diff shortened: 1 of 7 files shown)


src/coprocessor_v2/plugin_registry.rs
```diff
@@ -307,18 +302,18 @@ impl PluginRegistryInner {
     pub fn load_plugin<P: AsRef<OsStr>>(
         &mut self,
         filename: P,
-    ) -> Result<&'static str, PluginLoadingError> {
+    ) -> Result<String, PluginLoadingError> {
```
Jibbow, line 305: I think we could in theory still return `&'static str` if we store `name: &'static str` in `LoadedPlugin` instead of `name: String`.
But I'm not sure what is nicer? Any suggestions?

skyzh: We would have fewer constraints on lifetime if we use `String`, and I suggest doing so.

skyzh (approved):

components/coprocessor_plugin_api/src/util.rs
```diff
@@ -42,8 +47,23 @@ impl BuildInfo {
...
+    /// The name of the plugin.
+    pub name: &'static str,
+    /// The version string of the plugin. Should follow semantic versioning.
+    pub version: &'static str,
+}
+
 /// Declare a plugin for the library so that it can be loaded by TiKV.
 ///
+/// The macro has three different versions:
+/// * `declare_plugin!(plugin_name, plugin_version, plugin_ctor)` which gives you full control.
```
andylokandy, line 63: I suggest documenting the type for plugin_name and plugin_version.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -307,18 +302,18 @@ impl PluginRegistryInner {
...
             warn!("failed to load coprocessor plugin. Maybe not compiled correctly as a TiKV plugin?"; "plugin_path" => ?filename, "error" => ?err);
         }
         let plugin = plugin?;
 
-        let plugin_name = plugin.name();
+        let plugin_name = plugin.name().to_string();
 
         self.loaded_plugins.insert(
-            plugin_name.to_string(),
+            plugin_name.clone(),
```
andylokandy, line 316: Does it work?

```suggestion
            plugin.name().to_string(),
```

components/coprocessor_plugin_api/src/util.rs
Jibbow: Ah yes, good idea! Btw: would it be a good idea to require a `literal` matcher instead of an `expr`? https://doc.rust-lang.org/edition-guide/rust-next/literal-macro-matcher.html

src/coprocessor_v2/plugin_registry.rs
Jibbow: I think it should work. I moved the `to_string()` method call up a few lines (https://github.com/tikv/tikv/pull/10090/files#diff-e4204ab49a0a810cb8c0e685991d164e985fb6207a4106161a7aaf678321cc79R313), so we are cloning a `String` here.

components/coprocessor_plugin_api/src/util.rs
andylokandy: Good idea

Jibbow: done

Jibbow: Okay no, not a good idea, because then we can't use `env!()` macro to collect information from Cargo.toml. Going to revert this.

andylokandy: 😅

andylokandy (approved):

PR #10095, opened 2021-04-28
resolved_ts: use tracked index update region read progress

NingLin-P:

Problem Summary:

Currently, the `safe_ts` and `apply_index` in the leader's `RegionReadProgress` are updated in different threads, although 
both of them using `fetch_max` to ensure they are increasing, in some rare conditions there may still be problems:

One case is `safe_ts` update first:
1. resolved ts worker update the `safe_ts`
2. before the `apply_index` is updated, the leader broadcast the `(safe_ts, apply_index)` item to followers, where the `apply_index` is smaller than expected
3. the follower receives the `(safe_ts, apply_index)` item with a smaller `apply_index` and serve read request before it has enough data and may get an inconsistent result

Another case is `apply_index` update first, which will cause the follower need to catch up more data to serve read request, will fix it in the next PR

What's Changed:

This PR fixes the first case (`safe_ts` update first) by not directly update the `safe_ts` in resoled ts worker, and instead using `RegionReadProgress::update_safe_ts`, which will wait for the `apply_index` catch up before update `safe_ts`

### Check List 

Tests 

- Integration test

### Release note 
- No release note


components/resolved_ts/src/endpoint.rs
```diff
@@ -78,55 +80,80 @@ impl ObserveRegion {
...
+                ref mut tracked_index,
+                ..
+            } => {
                 for log in change_logs {
                     match log {
-                        ChangeLog::Error(e) => return Err(Error::request(e.clone())),
-                        ChangeLog::Rows { rows, .. } => rows.iter().for_each(|row| match row {
-                            ChangeRow::Prewrite { key, start_ts, .. } => {
-                                locks.push(PendingLock::Track {
+                        ChangeLog::Error(e) => {
```
BusyJay, line 90: Why not return?

components/resolved_ts/src/endpoint.rs
```diff
@@ -78,55 +80,80 @@ impl ObserveRegion {
...
-                            }),
-                            // One pc command do not contains any lock, so just skip it
-                            ChangeRow::OnePc { .. } => {}
-                        }),
+                                    commit_ts: *commit_ts,
+                                }),
+                                // One pc command do not contains any lock, so just skip it
+                                ChangeRow::OnePc { .. } => {}
+                            });
+                            assert!(*tracked_index < *index);
```
BusyJay, line 119: Probably should show more information like region id and their actual index value.

components/resolved_ts/src/resolver.rs
```diff
@@ -98,11 +110,10 @@ impl Resolver {
         // Resolved ts never decrease.
         self.resolved_ts = cmp::max(self.resolved_ts, new_resolved_ts);
 
-        // Update region read progress
-        // TODO: should use `RegionReadProgress::update_safe_ts` to avoid direct
-        // write to `safe_ts`
-        self.read_progress
-            .fetch_max_safe_ts(self.resolved_ts.into_inner());
+        // Publish an `(apply index, safe ts)` item into the region read progress
+        if let Some(rrp) = &self.read_progress {
```
BusyJay, line 114: How can it prevent safe_ts being updated before apply index?

NingLin-P: `update_safe_ts` will not update `safe_ts` if `apply_index` is smaller than the passed `tracked_index` and instead putting the `(tracked_index, resolved_ts)` into the `RegionReadProgress::pending_items`, and update the `safe_ts` after `apply index` is updated by `RegionReadProgress::update_applied`

BusyJay: What if `resolved_ts` is updated before `tracked_index` is updated? Like `apply_index` is updated in apply worker, but resolved ts is updated before `apply_index` is sent.

components/resolved_ts/src/endpoint.rs
NingLin-P: Return an error will require to re-register the region, and I think it is unnecessary since an error (like `EpochNotMatch`) won't change the lock data that already write to kv engine at the apply thread. /cc @5kbpers

5kbpers: After region merge some locks could be merged, so better to return an error here or detect the type of region error.

components/resolved_ts/src/endpoint.rs
```diff
@@ -78,55 +80,86 @@ impl ObserveRegion {
...
-                            // One pc command do not contains any lock, so just skip it
-                            ChangeRow::OnePc { .. } => {}
-                        }),
+                        ChangeLog::Error(e) => {
+                            debug!(
+                                "skip change log error";
+                                "region" => self.meta.id,
+                                "error" => ?e,
+                            );
+                            continue;
```
5kbpers, line 140: ditto

5kbpers (approved):
LGTM

components/resolved_ts/src/endpoint.rs
```diff
@@ -78,78 +80,108 @@ impl ObserveRegion {
 
     fn track_change_log(&mut self, change_logs: &[ChangeLog]) -> Result<()> {
         match self.resolver_status {
-            ResolverStatus::Pending { ref mut locks, .. } => {
+            ResolverStatus::Pending {
+                ref mut locks,
```
BusyJay, line 84: If you match it using `&mut self.resolver_status` then you don't need to write `ref mut` everywhere.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

Yisaer (approved):
/lgtm

NingLin-P:
/merge cancel
Wait for CI back to life

PR #10096, opened 2021-04-29
Replace some RocksEngine with KvEngine type parameters in tikv::server

brson:

Making the tikv crate generic over storage engines.

Just replacing instances of RocksEngine with KvEngine-bound type parameters in the tikv::server module.

I discovered that KvEngines need to be Unpin here, so I went ahead and made all KvEngines Unpin so engine authors can find that out early.

### Release note 

- No release note.

### DIFF
--- a/components/engine_traits/src/engine.rs
+++ b/components/engine_traits/src/engine.rs
@@ -31,6 +31,7 @@ pub trait KvEngine:
     + Sync
     + Clone
     + Debug
+    + Unpin
     + 'static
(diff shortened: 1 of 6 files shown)


brson:
@Connor1996 I think you've wanted in the past to not add more engine type parameters here, but I don't know how to avoid them. Could you review?

Connor1996 (approved):
Have no idea😢 LGTM

PR #10097, opened 2021-04-29
speed up snapshot for ongoing online stores

hicqu:

For an ongoing online store it's unnecessary to use configured `snap_max_bytes_per_sync` for I/O speed limiter, because there aren't Region leaders on the store. It's OK to use an infinate speed limiter to speed up Region replicas moving.

Use an infinate I/O speed limiter for ongoing online stores.

### Check List 

Tests 

Side effects

### Release note 
* No release note.

### DIFF
--- a/components/raftstore/src/store/snap.rs
+++ b/components/raftstore/src/store/snap.rs
@@ -22,6 +22,7 @@ use futures::executor::block_on;
 use futures_util::io::{AllowStdIo, AsyncWriteExt};
 use kvproto::encryptionpb::EncryptionMethod;
 use kvproto::metapb::Region;
+use kvproto::pdpb::StoreStatus;
 use kvproto::raft_serverpb::RaftSnapshotData;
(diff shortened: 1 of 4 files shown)


nolouch:
cc @nolouch

components/raftstore/src/store/snap.rs
```diff
@@ -1168,6 +1171,16 @@ impl SnapManager {
         Ok(())
     }
 
+    pub fn update_store_status(&self, status: StoreStatus) {
+        let new_limit = match status {
+            StoreStatus::Normal => self.max_write_speed as f64,
+            StoreStatus::OngoingOnline => f64::INFINITY,
```
nolouch, line 1177: How to define the completion staus of the online?

hicqu: PD determines it based on expected region score and exact region score of the store. For example, if exact region score reaches 90% of the expected one, PD can think the store becomes normal.

Issue #10046, opened 2021-04-20
Sysbench oltp_point_select regression with new request-batch mechanism


youjiali1995:
## Description
[`enable-request-batch=true`](https://github.com/tikv/tikv/pull/7878) results in about 5% throughput regression in `oltp_point_select` and the throughput is more unstable than turning it off.

The informal benchmark result is
Version | threads 500*1 | CPU Usage | 500*3 | Usage
-- | -- | -- | -- | --
v4.0.12 TiDB + v4.0.12 TiKV | 137273 | 1520%(TiDB) + 3*350%(TiKV storage pool) | 345075 | 3\*1320% + 3\*740%
v4.0.12 TiDB + v5.0 TiKV | 132203 | 1520% + 3*350% | 322585 | 3\*1200% + 3\*600%
v4.0.12 TiDB + v5.0 TiKV disable request batch | 134747 | 1500% + 3*400% | 344224 | 3\*1300% + 3\*850%

## Topology
3 TiDB: c5.4xlarge
3 TiKV: i3.4xlarge
3 PD: m5.xlarge
1 haproxy: c5.4xlarge
1 sysbench: c5.9xlarge

### TiDB
version: 5.0.0

```yaml
log.level: “error”
performance.max-procs: 20
prepared-plan-cache.enabled: true
tikv-client.max-batch-wait-time: 2000000
```
### TiKV
version: 5.0.0

```yaml
storage.scheduler-worker-pool-size: 5
raftstore.store-pool-size: 3
raftstore.apply-pool-size: 3
rocksdb.max-background-jobs: 3
raftdb.max-background-jobs: 3
raftdb.allow-concurrent-memtable-write: true
server.request-batch-enable-cross-command: false
server.grpc-concurrency: 6
readpool.unified.min-thread-count: 5
readpool.unified.max-thread-count: 20
readpool.storage.normal-concurrency: 10
pessimistic-txn.pipelined: true
```

PR #10098, opened 2021-04-29
storage:  optimize strategy of batch commands

Little-Wallace:

Issue Number: close https://github.com/tikv/tikv/issues/10046

Problem Summary:
If the CPU load of TiKV read pool is not very high, batch get request together may cause latency increase. Because batchrequest process every sub request one by one in order, but if you put them into readpool, they can be processed in parallel.

What's Changed:
I judge whether the readpool is busy by readpool queue length. If the queue length is very long, TiKV will batch request together, otherwise it will put them into readpool separately.
We can see that this PR fixes QPS regression and makes lower P99 latency in TiKV than master without batch.

Result:
I tested sysbench-point-select with 150,300,600,1200,2000 threads. 

```
sysbench oltp_point_select --report-interval=1 --rand-type=uniform --mysql-db=sysbench --mysql-host=127.0.0.1 --mysql-port=3390 --mysql-user=root --time=300 --threads=$thread --tables=16 --table-size=10000000 run > /home/centos/aws-test/sysbench.oltp_point_select.$thread.output.txt
```

master with not enable batch
image
image

master with enable request batch
image
image

this PR with enable request batch
image
image

### Related changes

### Check List 

Tests 

Side effects

- Performance regression
    - Consumes more CPU. When the CPU resources of the TiKV process are relatively idle, this PR will use more CPU than master.

### Release note 
- Fix sysbench point-get performance regression when TiKV readpool is not busy.


src/read_pool.rs
```diff
@@ -144,6 +150,22 @@ impl ReadPoolHandle {
...
+                read_pool_normal, ..
+            } => {
+                read_pool_normal.get_running_task_count()
+                    > read_pool_normal.get_pool_size() * MAX_BATCH_GET_REQUEST_COUNT
+            }
+            ReadPoolHandle::Yatp {
+                running_tasks,
+                max_task_limit_to_busy,
+                ..
+            } => running_tasks.get() > *max_task_limit_to_busy,
```
sticnarf, line 166: Why do we have different strategies between the two kinds of pools?

src/read_pool.rs
```diff
@@ -144,6 +150,22 @@ impl ReadPoolHandle {
             rx.map_err(ReadPoolError::from).await
         }
     }
+
+    pub fn is_busy(&self) -> bool {
```
tabokie, line 154: Could we use something like "pending_tasks_per_worker" to avoid passing `MAX_BATCH_GET_REQUEST_COUNT` all the way in?

src/server/service/kv.rs
```diff
@@ -926,10 +922,17 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
...
+                req_len -= 1;
+                if enable_req_batch
+                    && batcher.is_none()
+                    && req_len >= MAX_BATCH_GET_REQUEST_COUNT
+                    && storage.is_readpool_busy()
+                {
+                    batcher = Some(ReqBatcher::new());
+                }
             }
             if let Some(mut batch) = batcher {
```
tabokie, line 934: Can the batcher be turned off once the read pressure is released?

src/read_pool.rs
```diff
@@ -144,6 +150,22 @@ impl ReadPoolHandle {
...
+        match self {
+            ReadPoolHandle::FuturePools {
+                read_pool_normal, ..
```
youjiali1995, line 157: I think it's better to differentiate different pools, i.e., `is_busy(&self, priority: CommandPri)`.

src/server/service/kv.rs
```diff
@@ -1058,7 +1061,7 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
     }
 }
 
-fn response_batch_commands_request<F>(
+pub fn response_batch_commands_request<F>(
```
youjiali1995, line 1064: Why make it public?

src/server/service/kv.rs
Little-Wallace: Yes... You can see that I just created one batcher for every BatchCommandsRequest

src/read_pool.rs
Little-Wallace: Yes, I have rewrite this code.

src/server/service/kv.rs
```diff
@@ -909,12 +904,13 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
         let request_handler = stream.try_for_each(move |mut req| {
             let request_ids = req.take_request_ids();
             let requests: Vec<_> = req.take_requests().into();
-            let mut batcher = if enable_req_batch && requests.len() > 2 {
+            let mut batcher = if enable_req_batch && storage.is_readpool_busy() {
```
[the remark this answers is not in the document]
Little-Wallace: Done

src/server/service/batch.rs
```diff
@@ -83,60 +93,124 @@ impl ReqBatcher {
...
+        }
+        if req_batch_size > self.pool_size * MIN_BATCH_GET_REQUEST_COUNT
+            && queue_per_worker >= MIN_BATCH_GET_REQUEST_COUNT
+        {
+            return Some(ReqBatcher::new(req_batch_size / self.pool_size));
+        }
+        if req_batch_size >= MIN_BATCH_GET_REQUEST_COUNT
+            && queue_per_worker >= MAX_QUEUE_SIZE_PER_WORKER
+        {
+            return Some(ReqBatcher::new(req_batch_size));
```
tabokie, line 120: `req_batch_size != get_req_batch_size`, I'm not sure how much the overhead though for delaying a few gets among lots of writes. If it's worth considering, maybe we could fix it by using `if filtered_request_count >= self.target_batch_size` as commit condition, where filtered requests aren't type sensitive.

src/server/service/batch.rs
```diff
@@ -83,60 +93,124 @@ impl ReqBatcher {
...
+        if !self.enable_batch {
+            return None;
+        }
+        if req_batch_size > self.pool_size * MIN_BATCH_GET_REQUEST_COUNT
+            && queue_per_worker >= MIN_BATCH_GET_REQUEST_COUNT
+        {
+            return Some(ReqBatcher::new(req_batch_size / self.pool_size));
+        }
+        if req_batch_size >= MIN_BATCH_GET_REQUEST_COUNT
+            && queue_per_worker >= MAX_QUEUE_SIZE_PER_WORKER
```
tabokie, line 118: I'm OK with this algorithm, but did you run benchmark using `req_batch_size / self.pool_size` without considering `MIN_BATCH_REQUEST_COUNT` or `queue_per_worker`? I just can't think of a losing case, assuming requests are evenly dispatched to multiple queues. E.g. 8 requests with 8 queues = no batch, 8 requests with 4 queues = batch 2.

tabokie (commented):
Looks good overall.

src/read_pool.rs
```diff
@@ -100,6 +105,7 @@ impl ReadPoolHandle {
                     return Err(ReadPoolError::UnifiedReadPoolFull);
                 }
 
+                running_tasks.inc();
```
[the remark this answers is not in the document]
Little-Wallace: Because this the right position. We shall count those tasks which are still in queue rather than only count them when they are executed.

src/server/metrics.rs
```diff
@@ -370,7 +370,7 @@ lazy_static! {
             "tikv_server_request_batch_size",
             "Size of request batch input",
             &["type"],
-            exponential_buckets(1f64, 5f64, 10).unwrap()
+            vec![1.0, 2.0, 4.0, 8.0, 12.0, 16.0, 20.0, 24.0, 28.0, 32.0, 64.0]
         )
         .unwrap();
     pub static ref REQUEST_BATCH_RATIO_HISTOGRAM_VEC: RequestBatchRatioHistogramVec =
```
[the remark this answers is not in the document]
Little-Wallace: No.... I use it to observe the batch size of `batch-get-command`

src/server/service/batch.rs
```diff
@@ -83,60 +93,124 @@ impl ReqBatcher {
...
     gets: Vec<GetRequest>,
     tx: Sender<(u64, batch_commands_response::Response)>,
 ) {
-    let begin_instant = Instant::now_coarse();
-    let ret = storage.batch_get_command(gets);
+    // let begin_instant = Instant::now_coarse();
+    REQUEST_BATCH_SIZE_HISTOGRAM_VEC
+        .point_get
+        .observe(gets.len() as f64);
+    let begin_instant = tikv_util::time::Instant::now_coarse();
```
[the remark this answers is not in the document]
Little-Wallace: OK.

src/server/service/batch.rs
Little-Wallace: Yes.... The batch may be slow than the best size.

tabokie (approved):
Update the benchmark results?

PR #4066, opened 2019-01-14
raftstore: make it scale

BusyJay:
This pr uses the batch/router system introduced in #3972 and #3916
to make raftstore scale.

Note that raft messages can be reordered in current implementation.
But raft-rs is tolerate with it, so this won't be a problem.

# Benchmark results
default
default

PR #10099, opened 2021-04-29
raftstore: remove applying snapshot count

### LINKED ISSUE #10094: Applying snapshot count is always 0
## Bug Report

Screen Shot 2021-04-28 at 7 09 11 PM
Screen Shot 2021-04-28 at 7 09 21 PM

I found `applying_snap_count` is never assigned so that the value is always 0.

master

N/A

### Steps to reproduce

Run a workload and make some scheduling.

### What did you expect?
Apply count should have a value sometimes.

### What did happened?
It is always 0.

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

Issue Number: close #10094 

Problem Summary: Since https://github.com/tikv/tikv/pull/4066, the `applying_snapshot_count` is no longer taking effect because it is never assigned a value.

What's Changed: This PR remove it since it has already been contained in receiving count. See https://github.com/tikv/tikv/blob/f81592b0f18b76e79b4f146124c0c1625867e222/components/raftstore/src/store/snap.rs#L1428

### Related changes

### Check List 

Tests 

- Unit test (existing tests are not affected)

### Release note 

- Fix the issue that the apply snapshot count is always zero.


hicqu (approved):

HunDunDM (approved):

BusyJay:
I don't get it. So this PR doesn't fix the problem by actually assigning the apply count, instead, relying on the wrong receiving count?

rleungx:
> I don't get it. So this PR doesn't fix the problem by actually assigning the apply count, instead, relying on the wrong receiving count?

IMO, the applying count isn't very useful.

BusyJay:
Why? It can hint whether the is a snapshot taking long time or whether jitters are caused by applying snapshots.

rleungx:
> Why? It can hint whether the is a snapshot taking long time or whether jitters are caused by applying snapshots.

Otherwise, we won't take a long time to find that it has a problem. Currently, we can get the applying time through another metrics.

BusyJay:
> > Why? It can hint whether the is a snapshot taking long time or whether jitters are caused by applying snapshots.
> 
> Otherwise, we won't take a long time to find that it has a problem. Currently, we can get the applying time through another metrics.

Applying time only shows up when it's finished. For example, a stuck apply won't show in the metrics.

PR #10102, opened 2021-04-29
config-template: add option 'raftdb.wal-dir'.

MrCroxx:

Problem Summary:

There is a config option `raftdb.wal-dir` which specifies path to wal of raftdb, but the option is not listed in `config-template/yaml`. 

What's Changed:

Add option `raftdb.wal-dir` and its doc to `config-template.yaml`.

Validate the wal dirs of kvdb and raftdb are not the same in `config.rs`. (with unit test for `validate`)

### Check List 

Tests 

- Unit test
- Integration test

### Release note

- No release note


etc/config-template.toml
```diff
@@ -785,6 +785,15 @@
 # enable-statistics = true
 # stats-dump-period = "10m"
 
+## RocksDB WAL directory.
```
tabokie, line 788: Rocksdb -> Raft RocksDB

etc/config-template.toml
```diff
@@ -785,6 +785,15 @@
...
+## This config specifies the absolute directory path for WAL.
+## If it is not set, the log files will be in the same directory as data. When you set the path to
+## RocksDB directory in memory like in `/dev/shm`, you may want to set`wal-dir` to a directory on a
+## persistent storage. See https://github.com/facebook/rocksdb/wiki/How-to-persist-in-memory-RocksDB-database .
+## If there are two disks on the machine, storing RocksDB data and WAL logs on different disks can
+## improve performance.
+# wal-dir = ""
```
tabokie, line 795: Mention the two WAL can't be put together.

tabokie: In fact, you should also add a check similar to https://github.com/tikv/tikv/blob/f81592b0f18b76e79b4f146124c0c1625867e222/src/config.rs#L2371

tabokie (commented):
This config is not often used, please check the codebase for any wrong assumption on wal dir being unspecified, and add tests for them if you see fit.

MrCroxx:
> This config is not often used, please check the codebase for any wrong assumption on wal dir being unspecified, and add tests for them if you see fit.

Got it! I will work on that.

etc/config-template.toml
yiwu-arbug, line 795:
```suggestion
## This config specifies the absolute directory path for WAL.
## If it is not set, the log files will be in the same directory as data. 
## If there are two disks on the machine, storing RocksDB data and WAL logs on different disks can
## improve performance.
# wal-dir = ""
```
We don't recommend putting the WAL in tempfs.

yiwu-arbug (approved):
rest LGTM

etc/config-template.toml
MrCroxx: > We don't recommend putting the WAL in tempfs.

It seems that the original comment means the same with you. It suggest that if the user set the rocksdb directory in tempfs, the user is supposed to set this config to a persistent storage.

yiwu-arbug: Indeed I misread the comment. But anyway we shouldn't mention tempfs. Both WAL and the SST files should be on persistent disk.

MrCroxx: I agree with you. I've deleted it in the latest commit.

MrCroxx:
@hunterlxt @tabokie PTAL

components/server/src/server.rs
```diff
@@ -258,13 +258,26 @@ impl<ER: RaftEngine> TiKVServer<ER> {
     /// - If the max open file descriptor limit is not high enough to support
     ///   the main database and the raft database.
     fn init_config(mut config: TiKvConfig) -> ConfigController {
+        validate_and_persist_config(&mut config, true);
+
+        // ensure critical directories exist
+        // - ensure kvdb dir exist
```
tabokie, line 264: Self-explanatory. Can remove these comments.

components/server/src/dump.rs
```diff
@@ -223,6 +278,50 @@ mod tests {
         assert(15, &raft_engine);
     }
 
+    #[test]
+    fn test_dump_with_separate_wal_without_config_when_dump() {
```
tabokie, line 282: You can merge these two tests, doing two `check_and_dump_raft_db` in one run.

etc/config-template.toml
```diff
@@ -785,6 +784,14 @@
 # enable-statistics = true
 # stats-dump-period = "10m"
 
+## Raft RocksDB WAL directory.
+## This config specifies the absolute directory path for WAL.
+## If it is not set, the log files will be in the same directory as data.
+## If there are two disks on the machine, storing RocksDB data and WAL logs on different disks can
+## improve performance.
+## Make surce NEVER set this config the same as `rocksdb.wal-dir`.
```
tabokie, line 792:
```suggestion
## Do not set this config the same as `rocksdb.wal-dir`.
```

components/server/src/dump.rs
```diff
@@ -214,6 +222,53 @@ mod tests {
...
+
+    #[test]
+    fn test_dump_with_separate_wal() {
+        let data_path = tempfile::Builder::new().tempdir().unwrap().into_path();
+        let mut raftdb_path = data_path.clone();
+        let mut raftengine_path = data_path;
+        raftdb_path.push("raft");
+        raftengine_path.push("raft-engine");
+        let mut raftdb_wal_path = raftdb_path.clone();
+        raftdb_wal_path.push("wal");
```
tabokie, line 243: Should probably use more unique names, like `test_dump_with_separate_wal_raft`

MrCroxx: I'm not sure if it is necessary here. Each test here uses separate temp directory. I think it is safe to be consistent with the origin test.

tabokie: Let's at least differentiate from possible default paths, `data/raft` is the default raft path right now.

MrCroxx: I got you. That will be better.

components/server/src/dump.rs
MrCroxx: Well I review here. I found it is not possible to merge theses tests. 'check_and_dump_raft_db' will rename the origin raftdb and delete it safely. To merge these test needs to generate the raftdb twice, too. So I think it's better to leave them alone here.

src/config.rs
```diff
@@ -2370,6 +2370,19 @@ impl TiKvConfig {
...
+            config::canonicalize_path(&kv_db_path)?
+        } else {
+            config::canonicalize_path(&self.rocksdb.wal_dir)?
+        };
+        let raft_db_wal_path = if self.raftdb.wal_dir.is_empty() {
+            config::canonicalize_path(&self.raft_store.raftdb_path)?
+        } else {
+            config::canonicalize_path(&self.raftdb.wal_dir)?
+        };
+        if kv_db_wal_path == raft_db_wal_path {
```
hunterlxt, line 2383: sometimes relative addr may be used and it will be sikpped in `canonicalize_sub_path()`. For example, `a!=xx/../a`

hunterlxt (approved):
rest LGTM

src/config.rs
MrCroxx: Yep, I reviewed this and found you're right. I'll restore the code here. @tabokie @hunterlxt

hunterlxt: seem that this issue existed before the PR 😅

MrCroxx: I'm sorry could you please show the issue below? I failed to found the issue you mentioned either associated with the PR or elsewhere.

hunterlxt: I mean issue not Github issue...

tabokie: I think restore the code won't solve the problem here. Say I have two paths `/symbolic_dir/non_exist_data` and `/nvme0n1/non_exist_data`, the equivalence check should still fails.

tabokie (approved):
Good work!

PR #10107, opened 2021-04-30
Support ingest multiple files in one raft command

Little-Wallace:

Issue Number: part of https://github.com/pingcap/tics/issues/1811

Problem Summary:
Tiflash need to transfer sst to tiflash-format file and if lightning just retried ingesting a write-cf file but it did not retry ingesting a default-cf, tiflash will go panic. 

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- Support ingesting default-cf and write-cf sst in one raft command.


src/import/sst_service.rs
```diff
@@ -94,6 +96,100 @@ where
...
+                cmd.mut_requests().push(ingest);
+            }
+            cmd.mut_header().set_term(resp_header.get_current_term());
+
+            // Here we shall check whether the file has been ingested before. This operation
+            // must execute after geting a snapshot from raftstore to make sure that the
+            // current leader has applied to current term.
+            for sst in ssts.iter() {
+                if !importer.exist(&sst) {
+                    warn!(
```
glorv, line 170: Is it better to return an error than just ignore it?

Little-Wallace: No.... If I returned an error, lightning may retry again and again......

glorv: Why, if the sst file does not exist, lightning will just write another, this shouldn't cause dead lock

3pointer (commented):
/lgtm

PR #10112, opened 2021-05-01
coprv2: enable tests for plugins

Jibbow:

Issue Number: #9747  

**[do not merge yet; PR currently only for experimenting with the CI]**

Problem Summary:

Previously, we commented out the tests because they always failed in CI. Why? Because they require `libexample_plugin.so`, which is being built by Cargo, but then not uploaded as an "artifact" for when the tests are executed.
We came to the conclusion, that we should somehow use `include_bytes!()` to include the library into the tests. This PR tries to fix this.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

 * No release note

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4,6 +4,22 @@ use std::env;
 use std::path::Path;
 
 fn main() {
+    locate_coprocessor_plugins(&["example-plugin"]);
+}
+
+/// Locates compiled coprocessor plugins and makes their paths available via environment variables.
+///
+/// This is a bit hacky, but can be removed as soon as RFC
+/// <https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html> is available.
+///
+/// Why do we need this?
+/// This is because we want to use some coprocessor tests during tests. We can't just add them as a
+/// `dev-dependency`, because of how the CI works. Thus, we want to include them in our tests with
+/// `include_bytes!()`. However, we don't really know where the artifacts are located, so we need to
+/// find them with this function and expose the locations via environment variables.
+///
+/// Coprocessor plugins need to be added as a `build-dependency` in `Cargo.toml` for TiKV.
+fn locate_coprocessor_plugins(plugin_names: &[&str]) {
     let out_dir = env::var_os("OUT_DIR").unwrap();
     let build_dir = Path::new(&out_dir)
         .parent()
@@ -12,11 +28,26 @@ fn main() {
         .unwrap()
         .parent()
         .unwrap();
-    let plugin_path = build_dir.join("libexample_plugin.so");
 
-    println!("cargo:rerun-if-changed={}", plugin_path.display());
-    println!(
-        "cargo:rustc-env=CARGO_DYLIB_FILE_EXAMPLE_PLUGIN={}",
-        plugin_path.display()
-    );
+    for plugin_name in plugin_names {
+        let plugin_path = build_dir.join(pkgname_to_libname(plugin_name));
+        println!("cargo:rerun-if-changed={}", plugin_path.display());
+        println!(
+            "cargo:rustc-env=CARGO_DYLIB_FILE_EXAMPLE_PLUGIN={}",
+            plugin_path.display()
+        );
+    }
+}
+
+/// Converts a Rust `dylib` crate name to the platform specific library name for the compiled
+/// artifact.
+pub fn pkgname_to_libname(pkgname: &str) -> String {
+    let pkgname = pkgname.to_string().replace("-", "_");
+    if cfg!(target_os = "windows") {
+        format!("{}.dll", pkgname)
+    } else if cfg!(target_os = "macos") {
+        format!("lib{}.dylib", pkgname)
+    } else {
+        format!("lib{}.so", pkgname)
+    }
 }
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,8 @@ fn main() {
         .parent()
         .unwrap();
     let plugin_path = build_dir.join("libexample_plugin.so");
+
+    println!("cargo:rerun-if-changed={}", plugin_path.display());
     println!(
         "cargo:rustc-env=CARGO_DYLIB_FILE_EXAMPLE_PLUGIN={}",
         plugin_path.display()
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,20 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::env;
+use std::path::Path;
+
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let build_dir = Path::new(&out_dir)
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap();
+    let plugin_path = build_dir.join("libexample_plugin.so");
+    println!(
+        "cargo:rustc-env=CARGO_DYLIB_FILE_EXAMPLE_PLUGIN={}",
+        plugin_path.display()
+    );
+}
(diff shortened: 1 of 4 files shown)


src/coprocessor_v2/plugin_registry.rs
```diff
@@ -443,129 +443,132 @@ fn is_library_file<P: AsRef<Path>>(path: P) -> bool {
...
+        std::fs::create_dir_all(&coprocessor_dir).unwrap();
+        std::fs::remove_dir_all(&coprocessor_dir).unwrap();
+        std::fs::create_dir_all(&coprocessor_dir).unwrap();
+
+        let mut registry = PluginRegistry::new();
+        registry.start_hot_reloading(&coprocessor_dir).unwrap();
+
+        // trigger loading
+        std::fs::copy(&original_library_path, &library_path).unwrap();
+        std::thread::sleep(Duration::from_secs(4));
```
andylokandy, line 550:
```suggestion
        // fs watcher detects changes in every 3 seconds, therefore wait 4 seconds so as to make sure the watcher is triggered.
        std::thread::sleep(Duration::from_secs(4));
```

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -443,129 +443,132 @@ fn is_library_file<P: AsRef<Path>>(path: P) -> bool {
...
+
+        assert!(registry.get_plugin(plugin_name).is_some());
+        assert_eq!(
+            &PathBuf::from(registry.get_path_for_plugin(plugin_name).unwrap()),
+            &library_path
+        );
+
+        // trigger rename
+        std::fs::rename(&library_path, &library_path_2).unwrap();
+        std::thread::sleep(Duration::from_secs(4));
```
andylokandy, line 560: ditto

andylokandy (approved):
rest LGTM

build.rs
```diff
@@ -0,0 +1,53 @@
...
+/// Why do we need this?
+/// This is because we want to use some coprocessor tests during tests. We can't just add them as a
+/// `dev-dependency`, because of how the CI works. Thus, we want to include them in our tests with
+/// `include_bytes!()`. However, we don't really know where the artifacts are located, so we need to
+/// find them with this function and expose the locations via environment variables.
+///
+/// Coprocessor plugins need to be added as a `build-dependency` in `Cargo.toml` for TiKV.
+fn locate_coprocessor_plugins(plugin_names: &[&str]) {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let build_dir = Path::new(&out_dir)
```
skyzh, line 24: If the behavior that `{OUT_DIR}/../../../`contains built libraries documented in cargo documentation? If not, I suggest just report a warning instead of panic with `unwrap` here.

build.rs
```diff
@@ -0,0 +1,53 @@
...
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap();
+
+    for plugin_name in plugin_names {
+        let plugin_path = build_dir.join(pkgname_to_libname(plugin_name));
+        println!("cargo:rerun-if-changed={}", plugin_path.display());
+        println!(
+            "cargo:rustc-env=CARGO_DYLIB_FILE_EXAMPLE_PLUGIN={}",
```
skyzh, line 36: It seems that we would set `CARGO_DYLIB_FILE_EXAMPLE_PLUGIN` multiple times. Should we export different env variables for different plugins?

skyzh (commented):
Rest LGTM

build.rs
Jibbow: hm good point! Unfortunately, I think it might not work when `--target-dir` or `--out-dir` is specified :( 
see https://doc.rust-lang.org/cargo/commands/cargo-build.html#output-options
The only thing I found and that should always work is to search in `LD_LIBRARY_PATH`, because Cargo will modify it accordingly (https://doc.rust-lang.org/cargo/reference/environment-variables.html#dynamic-library-paths).
So maybe we should go with `LD_LIBRARY_PATH` instead?

build.rs
Jibbow: hm no really? I tried to just print this line once per plugin. Not sure if I messed something up?

src/coprocessor_v2/plugin_registry.rs
Jibbow: Yes, thanks! :)

build.rs
skyzh: But the name of the environment variable seems always to be `CARGO_DYLIB_FILE_EXAMPLE_PLUGIN`. Not sure what will happen if there are multiple plugins...

skyzh: Here's an example:

`build.rs`

```rust
fn main() {
    println!(
        "cargo:rustc-env=TEST=1"
    );
    println!(
        "cargo:rustc-env=TEST=2"
    );
}
```

`main.rs`

```rust
fn main() {
    println!("{}", env!("TEST"));
}
```

The output is `2`.

Therefore, in this part, `CARGO_DYLIB_FILE_EXAMPLE_PLUGIN` will be set to the last plugin in the array. Is this expected in this case?

Jibbow: ah right, I got it! Thanks, good catch! Will fix this :)

Jibbow: Fixed it. PTAL. And thanks, totally missed this :blush:

skyzh (approved):
Good work!

BusyJay:
I don't think we should add build script back to TiKV again, we had taken a lot of effort to get rid of it in the past :( https://github.com/tikv/tikv/pull/3368

Jibbow:
> I don't think we should add build script back to TiKV again, we had taken a lot of effort to get rid of it in the past :( #3368

Yes I see your point and we discussed this internally before. It's not ideal, but I currently don't see a good other solution for this problem. On the upside, if https://github.com/rust-lang/cargo/issues/9096 is being implemented, it should work out-of-the-box, because the `build.rs` is currently doing exactly the same.

But if you have new ideas on solving this issue, I'm very much willing to implement it. In short, the issue is:
 * we want to load a `dylib` file to execute tests with it
 * the source for the `dylib` file is in `components/test_coprocessor_plugin/example_plugin`
 * because we need it for the tests, we need to compile it before the tests are executed
 * we can't simply declare `example_plugin` as a dependency. it does work locally, but not on Jenkins, because the build artifacts are uploaded after the build and then the tests are executed somewhere else. somehow, the `dylib` artifact is not uploaded, so it does not exist when executing the tests. (I hope I got this one correct)
 * thus, we settled with using `include_bytes!` to include the `dylib` directly in the binary
 * in order to use `include_bytes!`, the `dylib` needs to be compiled *before* the build of TiKV happens
 * our solution was to trigger the build in `build.rs`.

@andylokandy @skyzh are also involved and know maybe better how everything works :D

BusyJay:
> use include_bytes!, the dylib needs to be compiled before the build of TiKV happens
> our solution was to trigger the build in build.rs.

Isn't including it as a dev dependency will make cargo compile it before TiKV lib already?

skyzh:
> > use include_bytes!, the dylib needs to be compiled before the build of TiKV happens
> 
> > our solution was to trigger the build in build.rs.
> 
> 
> 
> Isn't including it as a dev dependency will make cargo compile it before TiKV lib already?

Then it will fail when running cargo clippy. When running clippy, the binary is not generated yet, and rustc will complain at include_bytes for file not found.

skyzh:
> On the upside, if rust-lang/cargo#9096 is being implemented, it should work out-of-the-box, because the build.rs is currently doing exactly the same.

After a quick thought, this won't help either. cargo clippy will still complain about missing files on `include_bytes`, as this command won't generate any binaries.

What about separating coprv2 plugin manager into a separate crate, so that `build.rs` only need to be applied on coprv2-plugin-manager crate instead of TiKV binary?

BusyJay:
I have been updating jenkins workflow recently. I can bundle the built library as test artifact.

skyzh:
> I have been updating jenkins workflow recently. I can bundle the built library as test artifact.

Alright! I'll send a PR removing `build.rs` and will surely fail on CI. But once the Jenkins workflow is fixed, that new PR should work.

skyzh:
I have sent a PR at https://github.com/tikv/tikv/pull/10281

Issue #10111, opened 2021-04-30
region may not trigger split check


gengliqi:
### What version of TiKV are you using?
v4.0.9-v4.0.12,  v5.0.0, v5.0.1

### What operating system and CPU are you using?
doesn't matter

### Steps to reproduce
doesn't need special operation

### What did you expect?
trigger split check when needed

### What did happened?
doesn't trigger split check sometimes

https://github.com/tikv/tikv/pull/9081 introduced a new task in the split checker whose name is `GetRegionApproximateSizeAndKeys`. https://github.com/tikv/tikv/blob/eb2ed7f64efa95f3b80890f795127c1d61a292a7/components/raftstore/src/store/fsm/peer.rs#L3563-L3573 However, the code above has an assumption that if the `approximate_size` is not `None`, the split check has been done before, which is not true now.
So if the region has no enough write cmd to trigger the threshold(`compaction_declined_bytes` or `size_diff_hint`), it may not be split no matter how large it is.
Furthermore, restarting TiKV can not solve this problem because the leader will call `heartbeat_pd` and update its `approximate_size` after a while then split check will be skipped.

ilovesoup:
This might block one of our critical PoC tests which needs to prove horizontal scalability. Encountering large regions, the workload won't be scatter among nodes well on TiFlash.

PR #10119, opened 2021-05-06
raftstore: fix not schedule split check

Little-Wallace:

Issue Number: close https://github.com/tikv/tikv/issues/10111

Problem Summary:
TiKV will schedule a task to get approximate size when the leader of TiKV tries to send a heartbeat message to PD but it finds that it has not gotten approximate size before. This task will reset the approximate size of peer and it will cause the condition of split check does not match. In these case, the regions will not split even if the TiKV restarts and ticks a split-check task again.

I just change the task type to split check task for heartbeat-pd. 

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No release note.

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -4,8 +4,6 @@ use std::borrow::Cow;
 use std::collections::Bound::{Excluded, Unbounded};
 use std::collections::VecDeque;
 use std::iter::Iterator;
-use std::sync::atomic::AtomicU64;
-use std::sync::Arc;
 use std::time::Instant;
 use std::{cmp, u64};
 
@@ -2242,8 +2240,6 @@ where
             if let Err(e) = self.ctx.split_check_scheduler.schedule(
                 SplitCheckTask::GetRegionApproximateSizeAndKeys {
                     region: self.fsm.peer.region().clone(),
-                    pending_tasks: Arc::new(AtomicU64::new(1)),
-                    cb: Box::new(move |_, _| {}),
                 },
             ) {
                 error!(
@@ -2401,11 +2397,7 @@ where
                 // The size and keys for new region may be far from the real value.
                 // So we let split checker to update it immediately.
                 if let Err(e) = self.ctx.split_check_scheduler.schedule(
-                    SplitCheckTask::GetRegionApproximateSizeAndKeys {
-                        region: new_region,
-                        pending_tasks: Arc::new(AtomicU64::new(1)),
-                        cb: Box::new(move |_, _| {}),
-                    },
+                    SplitCheckTask::GetRegionApproximateSizeAndKeys { region: new_region },
                 ) {
                     error!(
                         "failed to schedule split check task";
@@ -3586,18 +3578,7 @@ where
         }
         self.fsm.skip_split_count = 0;
 
-        let task =
-            SplitCheckTask::split_check(self.fsm.peer.region().clone(), true, CheckPolicy::Scan);
-        if let Err(e) = self.ctx.split_check_scheduler.schedule(task) {
-            error!(
-                "failed to schedule split check";
-                "region_id" => self.fsm.region_id(),
-                "peer_id" => self.fsm.peer_id(),
-                "err" => %e,
-            );
-        }
-        self.fsm.peer.size_diff_hint = 0;
-        self.fsm.peer.compaction_declined_bytes = 0;
+        self.fsm.peer.schedule_check_split(self.ctx);
         self.register_split_region_check_tick();
     }
 
@@ -3711,12 +3692,19 @@ where
 
     fn on_approximate_region_size(&mut self, size: u64) {
         self.fsm.peer.approximate_size = Some(size);
+        if self.fsm.peer.pending_pd_heartbeat_tasks && self.fsm.peer.approximate_keys.is_some() {
+            self.fsm.peer.heartbeat_pd(self.ctx);
+        }
         self.register_split_region_check_tick();
         self.register_pd_heartbeat_tick();
     }
 
     fn on_approximate_region_keys(&mut self, keys: u64) {
         self.fsm.peer.approximate_keys = Some(keys);
+        if self.fsm.peer.pending_pd_heartbeat_tasks && self.fsm.peer.approximate_size.is_some() {
+            self.fsm.peer.heartbeat_pd(self.ctx);
+        }
+
         self.register_split_region_check_tick();
         self.register_pd_heartbeat_tick();
     }
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -16,6 +16,7 @@ use fail::fail_point;
 use kvproto::errorpb;
 use kvproto::kvrpcpb::ExtraOp as TxnExtraOp;
 use kvproto::metapb::{self, PeerRole};
+use kvproto::pdpb::CheckPolicy;
 use kvproto::pdpb::PeerStats;
 use kvproto::raft_cmdpb::{
     self, AdminCmdType, AdminResponse, ChangePeerRequest, CmdType, CommitMergeRequest,
@@ -505,10 +506,7 @@ where
     /// The number of the last unpersisted ready.
     last_unpersisted_number: u64,
 
-    /// The number of pending pd heartbeat tasks. Pd heartbeat task may be blocked by
-    /// reading rocksdb. To avoid unnecessary io operations, we always let the later
-    /// task run when there are more than 1 pending tasks.
-    pub pending_pd_heartbeat_tasks: Arc<AtomicU64>,
+    pub pending_pd_heartbeat_tasks: bool,
 
     pub read_progress: Arc<RegionReadProgress>,
 }
@@ -603,7 +601,7 @@ where
             max_ts_sync_status: Arc::new(AtomicU64::new(0)),
             cmd_epoch_checker: Default::default(),
             last_unpersisted_number: 0,
-            pending_pd_heartbeat_tasks: Arc::new(AtomicU64::new(0)),
+            pending_pd_heartbeat_tasks: false,
             read_progress: Arc::new(RegionReadProgress::new(
                 applied_index,
                 REGION_READ_PROGRESS_CAP,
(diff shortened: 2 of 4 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2184,40 +2182,18 @@ where
...
-            self.register_split_region_check_tick();
-        }
-
         let mut meta = self.ctx.store_meta.lock().unwrap();
         meta.set_region(&self.ctx.coprocessor_host, derived, &mut self.fsm.peer);
         self.fsm.peer.post_split();
 
+        // If `estimated_size` or `estimated_keys` equals none, the leader will schedule a
+        // split-check task in
+        let need_check_split = estimated_size.is_some() && estimated_keys.is_some();
```
gozssky, line 2191: Maybe `estimated_size.is_none() || estimated_keys.is_none()` ? (To keep consistent with the above comment)

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2226,6 +2202,9 @@ where
                 "peer_id" => self.fsm.peer_id(),
                 "split_count" => regions.len(),
             );
+            if need_check_split {
+                self.fsm.peer.schedule_check_split(self.ctx);
```
gozssky, line 2206: The `estimated_size` may be far from the real size. Should we always call `schedule_check_split` here?

Little-Wallace: If `estimated_size ` is some, we will get approximate size in split-check thread again and decide whether to split.

components/raftstore/src/store/fsm/peer.rs
Little-Wallace: No. If `estimated_size ` is none, the leader will schedule a split-check task in method `heartbeat_pd`.

gozssky: I see. Maybe we should make the comment more clear? 😂

components/raftstore/src/store/fsm/peer.rs
gengliqi: If `approximate_size` is some and the threshold is not reached, there is no split check.

gengliqi (commented):
I think it's the better alternative that changing the condition mentioned here https://github.com/tikv/tikv/issues/10111. We should distinguish whether the split check has been done instead of judging whether `approximate_size` is some.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3978,8 +3907,11 @@ where
             size += sst.total_bytes;
             keys += sst.total_kvs;
         }
-        self.fsm.peer.approximate_size = Some(self.fsm.peer.approximate_size.unwrap_or(0) + size);
-        self.fsm.peer.approximate_keys = Some(self.fsm.peer.approximate_keys.unwrap_or(0) + keys);
+        self.fsm.peer.approximate_size = self.fsm.peer.approximate_size + size;
+        self.fsm.peer.approximate_keys = self.fsm.peer.approximate_keys + keys;
+        // The ingested file may be overlapped with the data in engine, so we need to check it
+        // again to get the actually value.
```
gengliqi, line 3916:
```suggestion
        // again to get the accurate value.
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2193,46 +2193,21 @@ where
         let region_id = derived.get_id();
         // Roughly estimate the size and keys for new regions.
         let new_region_count = regions.len() as u64;
-        let estimated_size = self.fsm.peer.approximate_size.map(|x| x / new_region_count);
-        let estimated_keys = self.fsm.peer.approximate_keys.map(|x| x / new_region_count);
+        let estimated_size = self.fsm.peer.approximate_size / new_region_count;
+        let estimated_keys = self.fsm.peer.approximate_keys / new_region_count;
         // It's not correct anymore, so set it to None to let split checker update it.
```
gengliqi, line 2198: The comment should be updated.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2193,46 +2193,21 @@ where
...
         self.fsm.peer.post_split();
 
+        self.fsm.peer.has_calculated_region_size = false;
+
+        let is_leader = self.fsm.peer.is_leader();
         if is_leader {
+            self.fsm.peer.approximate_size = estimated_size;
+            self.fsm.peer.approximate_keys = estimated_keys;
             self.fsm.peer.heartbeat_pd(self.ctx);
+            self.register_pd_heartbeat_tick();
```
gengliqi: Why register pd heartbeat tick here?

components/raftstore/src/store/peer.rs
```diff
@@ -431,9 +432,13 @@ where
...
     /// Approximate size of the region.
-    pub approximate_size: Option<u64>,
+    pub approximate_size: u64,
     /// Approximate keys of the region.
-    pub approximate_keys: Option<u64>,
+    pub approximate_keys: u64,
+    /// Whether this region has calculated region size by split-check thread. If we just splitted
+    ///  the region or ingested one file which may be overlapped with the existed data, the
+    /// `approximate_size` is not very accurate.
+    pub has_calculated_region_size: bool,
```
gengliqi, line 441: It depends on the fact that `CasualMessage::RegionApproximateSize` msg is only sent by split checker and split check is done.

Little-Wallace: No. It depends on that `CasualMessage::RegionApproximateSize` can give an accurate value.

components/raftstore/src/store/fsm/peer.rs
Little-Wallace: Because the value is not accurate and it will be calculated in split-check thread. So heartbeat pd just after soon can give an accurate value from split-check.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3978,8 +3910,11 @@ where
             size += sst.total_bytes;
             keys += sst.total_kvs;
         }
-        self.fsm.peer.approximate_size = Some(self.fsm.peer.approximate_size.unwrap_or(0) + size);
-        self.fsm.peer.approximate_keys = Some(self.fsm.peer.approximate_keys.unwrap_or(0) + keys);
+        self.fsm.peer.approximate_size = self.fsm.peer.approximate_size + size;
+        self.fsm.peer.approximate_keys = self.fsm.peer.approximate_keys + keys;
```
Connor1996, line 3914: If you want to check the size again, you don't need to see the approximate size here

Connor1996: And if that `has_calculated_region_size` is unnecessary, we can just keep the approximate be `Option`

Little-Wallace: Emmmm, but we will report to PD at once and TiKV will not schedule split-check during importing.

components/raftstore/src/store/peer.rs
```diff
@@ -3371,58 +3366,39 @@ where
...
+            );
             return;
         }
-
-        if self.pending_pd_heartbeat_tasks.load(Ordering::SeqCst) > 2 {
-            return;
+        fail_point!("schedule_check_split");
+        if !self.has_calculated_region_size {
+            if ctx.importer.get_mode() != SwitchMode::Import {
+                self.schedule_check_split(ctx);
```
gengliqi, line 3385: How about waiting for split check tick? If so, the function `schedule_check_split` can be removed and the logic in it can be returned back to `on_split_region_check_tick`

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3609,23 +3538,13 @@ where
...
-                "region_id" => self.fsm.region_id(),
-                "peer_id" => self.fsm.peer_id(),
-                "err" => %e,
-            );
-        }
-        self.fsm.peer.size_diff_hint = 0;
-        self.fsm.peer.compaction_declined_bytes = 0;
-        true
+        self.fsm.peer.schedule_check_split(self.ctx);
+        self.register_split_region_check_tick();
```
gengliqi, line 3547: I suggest moving `if self.fsm.peer.has_calculated_region_size...` after the `is_leader` check and after that call `self.register_split_region_check_tick();`.
The condition `if !self.ctx.cfg.hibernate_regions {..}` can be removed.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2193,45 +2179,19 @@ where
...
         let mut meta = self.ctx.store_meta.lock().unwrap();
         meta.set_region(&self.ctx.coprocessor_host, derived, &mut self.fsm.peer);
         self.fsm.peer.post_split();
 
+        self.fsm.peer.has_calculated_region_size = false;
+
+        let is_leader = self.fsm.peer.is_leader();
         if is_leader {
+            self.fsm.peer.approximate_size = estimated_size;
+            self.fsm.peer.approximate_keys = estimated_keys;
```
gengliqi, line 2194: `register_split_region_check_tick` should be called here.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2410,24 +2356,10 @@ where
...
-                        "failed to schedule split check task";
-                        "region_id" => new_region_id,
-                        "err" => ?e,
-                    );
-                }
-            }
+        }
+        drop(meta);
+        if is_leader {
+            self.register_pd_heartbeat_tick();
```
gengliqi, line 2362: Seems it is not necessary. After running split check and getting `approximate_size(key)`, the pd heartbeat tick will be registered here.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3520,11 +3509,19 @@ where
             return;
         }
 
+        self.register_split_region_check_tick();
```
Connor1996, line 3512: seems it should be ahead of the above check?

components/raftstore/src/store/fsm/peer.rs
Connor1996: why `if !self.ctx.cfg.hibernate_regions {..}` can be removed

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2193,45 +2179,19 @@ where
...
-                        "trigger split check immediately";
-                        "region_id" => self.fsm.region_id(),
-                        "peer_id" => self.fsm.peer_id(),
-                        "size" => estimated_size,
-                    );
-                    self.fsm.peer.approximate_size = None;
-                    // trigger a split check immediately when size is still large
-                    self.ctx
-                        .router
-                        .force_send(region_id, PeerMsg::Tick(PeerTicks::SPLIT_REGION_CHECK))
```
Connor1996, line 2220: why remove this?

Little-Wallace: We will register tick in the end of the method.

Connor1996: seems now it gets the approximate size every time it splits. I don't think it's really necessary.

components/raftstore/src/store/fsm/peer.rs
gengliqi: How about calling it here？It looks clearer than putting it at the end.

components/raftstore/src/store/fsm/peer.rs
gengliqi: Since the behavior can be the same whatever `hibernate_regions` is true or not. 
In the original code, the tick is registered each time even when this peer is a follower/learner which is very wasteful.

gengliqi (approved):
LGTM

components/raftstore/src/store/fsm/peer.rs
Little-Wallace: No. If we do not pass the check 
``` if self.fsm.peer.has_calculated_region_size
            && self.fsm.peer.compaction_declined_bytes < self.ctx.cfg.region_split_check_diff.0
            && self.fsm.peer.size_diff_hint < self.ctx.cfg.region_split_check_diff.0```, TiKV will never check again util more entries are appended.

Connor1996 (approved):
LGTM

Connor1996 (approved):
LGTM

components/raftstore/src/store/fsm/peer.rs
BusyJay: What if more entries are appended but committed or applied after next split check?

PR #10122, opened 2021-05-06
raftstore: support report hotspot in store heartbeat

Yisaer:

Problem Summary:

In order to support scheduling hot spot peer read stats in PD, we need to find a way to report hot spot read stats in peer level to PD.

see also: https://github.com/pingcap/kvproto/pull/760

collect read_keys and read_bytes for hot spot peer by `handle_store_heartbeat`

### Check List 

Tests 

- Unit test

### Release note 

* No release notes


components/raftstore/src/store/worker/pd.rs
```diff
@@ -659,6 +691,24 @@ where
...
 
+        for (region_id, region_peer) in &mut self.region_peers {
+            let read_bytes = region_peer.read_bytes - region_peer.last_store_report_read_bytes;
+            let read_keys = region_peer.read_keys - region_peer.last_store_report_read_keys;
+            // TODO: select hotspot peer by binaray heap in future
+            if read_bytes < hotspot_byte_report_threshold()
+                || read_keys < hotspot_key_report_threshold()
+            {
+                continue;
+            }
```
lhy1024, line 702: I think it's enough

components/raftstore/src/store/worker/pd.rs
```diff
@@ -659,6 +691,24 @@ where
...
+            let mut peer_stat = pdpb::PeerStat::default();
+            peer_stat.set_region_id(*region_id);
+            peer_stat.set_read_bytes(read_bytes);
+            peer_stat.set_read_keys(read_keys);
+            stats.peer_stats.push(peer_stat);
+            region_peer.last_store_report_read_bytes = region_peer.read_bytes;
+            region_peer.last_store_report_read_keys = region_peer.read_keys;
```
lhy1024, line 709: They also need to be updated although are less than threshold.

Yisaer: good catch.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1114,16 +1164,20 @@ where
...
+                    let written_bytes_delta =
+                        hb_task.written_bytes - peer_stat.last_region_report_written_bytes;
+                    let written_keys_delta =
+                        hb_task.written_keys - peer_stat.last_region_report_written_keys;
+                    let mut last_report_ts = peer_stat.last_region_report_ts;
+                    peer_stat.last_region_report_written_bytes = hb_task.written_bytes;
+                    peer_stat.last_region_report_written_keys = hb_task.written_keys;
+                    peer_stat.last_region_report_read_bytes = peer_stat.read_bytes;
+                    peer_stat.last_region_report_read_keys = peer_stat.read_keys;
+                    peer_stat.last_region_report_ts = UnixSecs::now();
```
lhy1024, line 1180: In fact, after this pr, it only collects the leader's read and write data？

components/raftstore/src/store/worker/pd.rs
```diff
@@ -191,11 +191,13 @@ impl Default for StoreStat {
...
-    pub last_read_keys: u64,
-    pub last_written_bytes: u64,
-    pub last_written_keys: u64,
-    pub last_report_ts: UnixSecs,
+    pub last_region_report_read_bytes: u64,
+    pub last_region_report_read_keys: u64,
+    pub last_region_report_written_bytes: u64,
+    pub last_region_report_written_keys: u64,
+    pub last_region_report_ts: UnixSecs,
+    pub last_store_report_read_bytes: u64,
```
lhy1024, line 199: IMO, difference between `last_region_report_read_bytes` and `last_store_report_read_bytes` is that their interval are different. Is it true? Would you like to add some comments?

components/raftstore/src/store/worker/pd.rs
```diff
@@ -659,6 +691,24 @@ where
...
+            // TODO: select hotspot peer by binaray heap in future
+            if read_bytes < hotspot_byte_report_threshold()
+                || read_keys < hotspot_key_report_threshold()
+            {
+                continue;
+            }
+            let mut peer_stat = pdpb::PeerStat::default();
+            peer_stat.set_region_id(*region_id);
+            peer_stat.set_read_bytes(read_bytes);
+            peer_stat.set_read_keys(read_keys);
```
lhy1024, line 708: Do we need report_ts?

components/raftstore/src/store/worker/pd.rs
Yisaer: updated.

components/raftstore/src/store/worker/pd.rs
Yisaer: I think this is unnecessary, otherwise it also should be recorded in the store stat instead of peer stat.

lhy1024 (approved):

components/raftstore/src/store/worker/pd.rs
```diff
@@ -659,6 +693,24 @@ where
             Ok(stats) => stats,
         };
 
+        for (region_id, region_peer) in &mut self.region_peers {
+            let read_bytes = region_peer.read_bytes - region_peer.last_store_report_read_bytes;
```
nolouch, line 697: how do you calculate the speed?

lhy1024: delta is enough

components/raftstore/src/store/worker/pd.rs
```diff
@@ -659,6 +693,24 @@ where
...
+            let read_keys = region_peer.read_keys - region_peer.last_store_report_read_keys;
+            region_peer.last_store_report_read_bytes = region_peer.read_bytes;
+            region_peer.last_store_report_read_keys = region_peer.read_keys;
+            // TODO: select hotspot peer by binaray heap in future
+            if read_bytes < hotspot_byte_report_threshold()
+                || read_keys < hotspot_key_report_threshold()
```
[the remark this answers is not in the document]
Yisaer: good catch, updated.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -431,28 +431,16 @@ const HOTSPOT_BYTE_RATE_THRESHOLD: u64 = 8 * 1024;
 // TODO: support dyamic configure threshold in future
 fn hotspot_key_report_threshold() -> u64 {
     #[cfg(feature = "failpoints")]
```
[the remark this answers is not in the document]
Yisaer: This is used to make clippy happy.

PR #10126, opened 2021-05-07
resolved_ts: delay matching default cf

5kbpers:

Problem Summary:
Filter out invalid default cf value by delaying matching it with write or lock.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A

### DIFF
--- a/components/batch-system/benches/batch-system.rs
+++ b/components/batch-system/benches/batch-system.rs
@@ -12,7 +12,7 @@ use std::sync::Arc;
 
 fn end_hook(tx: &std::sync::mpsc::Sender<()>) -> Message {
     let tx = tx.clone();
-    Message::Callback(Box::new(move |_| {
+    Message::Callback(Box::new(move |_, _| {
         tx.send(()).unwrap();
     }))
 }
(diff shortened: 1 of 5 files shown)


components/batch-system/benches/batch-system.rs
```diff
@@ -12,7 +12,7 @@ use std::sync::Arc;
 
 fn end_hook(tx: &std::sync::mpsc::Sender<()>) -> Message {
     let tx = tx.clone();
-    Message::Callback(Box::new(move |_| {
+    Message::Callback(Box::new(move |_, _| {
```
NingLin-P, line 15: Why change it?

components/resolved_ts/src/cmd.rs
```diff
@@ -209,6 +209,7 @@ struct RowChange {
 
 fn group_row_changes(requests: Vec<Request>) -> HashMap<Key, RowChange> {
     let mut changes: HashMap<Key, RowChange> = HashMap::default();
+    let mut unmatched_default = HashMap::default();
```
NingLin-P, line 212: Plz add comments about its use

NingLin-P (commented):
Rest LGTM

components/batch-system/benches/batch-system.rs
5kbpers: The signature was changed before and rust-analyzer reports a compilation error.

NingLin-P (approved):

nolouch (approved):

PR #10128, opened 2021-05-07
properly canonicalize non-existing path

tabokie:

What's Changed:

Normalize the path by lexical rules, then try iteratively canonicalize the result.

### Check List 

Tests 

- Unit test

### Release note 

- No release note

### DIFF
--- a/components/tikv_util/src/config.rs
+++ b/components/tikv_util/src/config.rs
@@ -470,22 +470,16 @@ fn canonicalize_fallback<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
             ret.push(c.as_os_str());
         }
 
-        let mut normal_nodes = 0;
         for component in components {
             match component {
                 Component::Prefix(..) | Component::RootDir => unreachable!(),
                 Component::CurDir => {}
                 c @ Component::ParentDir => {
-                    if normal_nodes > 0 {
-                        ret.pop();
-                    } else {
+                    if !ret.pop() {
                         ret.push(c.as_os_str());
                     }
                 }
-                Component::Normal(c) => {
-                    ret.push(c);
-                    normal_nodes += 1;
-                }
+                Component::Normal(c) => ret.push(c),
             }
         }
         ret
@@ -506,7 +500,7 @@ fn canonicalize_fallback<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
             components.next();
             ret.push(c.as_os_str());
         }
-        // Normalize will only preserve leading ParentDir.
+        // normalize() will only preserve leading ParentDir.
         while let Some(Component::ParentDir) = components.peek().cloned() {
             components.next();
             ret.pop();
--- a/components/tikv_util/src/config.rs
+++ b/components/tikv_util/src/config.rs
@@ -494,7 +494,9 @@ fn canonicalize_fallback<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
         } else {
             PathBuf::new()
         };
-        while let Some(c @ (Component::Prefix(..) | Component::RootDir)) = components.peek().cloned() {
+        while let Some(c @ (Component::Prefix(..) | Component::RootDir)) =
+            components.peek().cloned()
+        {
             components.next();
             ret.push(c.as_os_str());
         }
@@ -516,7 +518,10 @@ fn canonicalize_fallback<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
                         }
                     }
                 }
-                Component::Prefix(..) | Component::RootDir | Component::ParentDir | Component::CurDir => unreachable!(),
+                Component::Prefix(..)
+                | Component::RootDir
+                | Component::ParentDir
+                | Component::CurDir => unreachable!(),
             }
         }
         Ok(ret)
@@ -1446,7 +1451,12 @@ mod tests {
         }
         for first in &nodes {
             for second in &nodes {
-                let path = format!("{}/{}/../{}/non_existing", tmp_dir.to_str().unwrap(), first, second);
+                let path = format!(
+                    "{}/{}/../{}/non_existing",
+                    tmp_dir.to_str().unwrap(),
+                    first,
+                    second
+                );
                 let res_path = canonicalize_path(&path).unwrap();
                 // resolve to second/non_existing
                 if *second == "non_existing" {
--- a/components/tikv_util/src/config.rs
+++ b/components/tikv_util/src/config.rs
@@ -457,17 +457,63 @@ impl<'de> Deserialize<'de> for ReadableDuration {
     }
 }
 
+fn canonicalize_imp<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
+    fn canonicalize_non_existing_path<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
+        use std::path::Component;
+        let mut components = path.as_ref().components().peekable();
+        let mut last_canonicalized = None;
+        let mut should_canonicalize = true;
+        let mut ret = if let Some(c @ Component::Prefix(..)) = components.peek().cloned() {
+            components.next();
+            PathBuf::from(c.as_os_str())
+        } else {
+            PathBuf::new()
+        };
+        for component in components {
+            match component {
+                Component::Prefix(..) => unreachable!(),
+                Component::RootDir => {
+                    ret.push(component.as_os_str());
+                }
+                Component::CurDir => {}
+                Component::ParentDir => {
+                    ret.pop();
+                }
+                Component::Normal(c) => {
+                    // We try to canonicalize a longest path based on fs info.
+                    if should_canonicalize {
+                        match ret.as_path().canonicalize() {
+                            Ok(path) => {
+                                last_canonicalized.replace(path);
+                            }
+                            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+                                if last_canonicalized.is_some() {
+                                    ret = last_canonicalized.take().unwrap();
+                                }
+                                should_canonicalize = false;
+                            }
+                            other => return other,
+                        }
+                    }
+                    ret.push(c);
+                }
+            }
+        }
+        Ok(ret)
+    }
+    match path.as_ref().canonicalize() {
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => canonicalize_non_existing_path(path),
+        other => other,
+    }
+}
+
 pub fn canonicalize_path(path: &str) -> Result<String, Box<dyn Error>> {
     canonicalize_sub_path(path, "")
 }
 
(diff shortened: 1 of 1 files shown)


tabokie:
FYI partially based on a popular implement: https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61

components/tikv_util/src/config.rs
```diff
@@ -457,17 +457,63 @@ impl<'de> Deserialize<'de> for ReadableDuration {
...
+        };
+        for component in components {
+            match component {
+                Component::Prefix(..) => unreachable!(),
+                Component::RootDir => {
+                    ret.push(component.as_os_str());
+                }
+                Component::CurDir => {}
+                Component::ParentDir => {
+                    ret.pop();
```
skyzh, line 480: It is possible that we could still canonicalize after we pop the parent directory.

e.g.

```
realpath /folder1/folder2/non_exist/x/../../folder_3
```

where
```
/folder1 -> /folder2
/folder2/folder_3 -> /folder_4
```

should be resolved to `/folder_4`.

However, as `/folder1/folder2/non_exist` does not exist, this function won't do further canonicalize on `/folder1/folder2/folder_3`, if I understood this part correctly.

skyzh:
> FYI partially based on a popular implement: https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61

Normalize is different from canonicalize. Here's another example:
```
/folder1/folder2 -> /folder2
```

Normalize `/folder_1/folder2/..` would yield `/folder_1`. But canonicalize will make it to `/`.

MrCroxx:
As @skyzh mentioned, `canonicalize_imp` may lead to the following issues:

run

```rust
    #[test]
    fn test_canonicalize_imp() {
        let tmpath = tempfile::Builder::new().tempdir().unwrap().into_path();
        let sub = "../not-exists/not-exists-too";
        print_original_and_canonicalized_path(&tmpath, sub);
        let sub = "not-exists/not-exists-too";
        print_original_and_canonicalized_path(&PathBuf::from("."), sub);
    }

    fn print_original_and_canonicalized_path(base: &PathBuf, sub: &str) {
        let mut path = base.clone();
        path.push(sub);
        println!(
            "$BASE => {base:?}\n$BASE/{sub} => {ori:?}\n$BASE/{sub} => {can:?}\n",
            base = base.as_os_str(),
            sub = sub,
            ori = &path.as_os_str(),
            can = canonicalize_imp(&path).unwrap_or(PathBuf::from("/error"))
        );
    }
```

will get:

```
$BASE => "/tmp/.tmp1ebBeP"
$BASE/../not-exists/not-exists-too => "/tmp/.tmp1ebBeP/../not-exists/not-exists-too"
$BASE/../not-exists/not-exists-too => "/tmp/not-exists-too"

$BASE => "."
$BASE/not-exists/not-exists-too => "./not-exists/not-exists-too"
$BASE/not-exists/not-exists-too => "not-exists/not-exists-too"
```

For the first test case, it reach the case that @skyzh mentioned above.

For the second test case, `canonicalize_imp` does not canonicalize as much as possible.

I propose a PR [pull#10129](https://github.com/tikv/tikv/pull/10129) , I think that works better.

skyzh:
The `->` in my test case means symbolic link. And it seems that you have discovered another bug :)

MrCroxx:
> The `->` in my test case means symbolic link. And it seems that you have discovered another bug :)

Sorry, I forget the symbolic link. I need to reconsider it.

tabokie:
@MrCroxx @skyzh Updated to normalize then canonicalize, PTAL again

components/tikv_util/src/config.rs
```diff
@@ -457,17 +457,84 @@ impl<'de> Deserialize<'de> for ReadableDuration {
...
+    fn try_canonicalize_normalized_path<P: AsRef<Path>>(path: P) -> std::io::Result<PathBuf> {
+        use std::path::Component;
+        let mut components = path.as_ref().components().peekable();
+        let mut should_canonicalize = true;
+        let mut ret = if let Some(c @ Component::Prefix(..)) = components.peek().cloned() {
+            components.next();
+            PathBuf::from(c.as_os_str())
+        } else {
+            PathBuf::new()
+        };
```
MrCroxx, line 497: Should we canonicalize current directory first?

When we use a relative path and the first component of the path does not exist, this impl will not canonicalize as much as possible.

For example:

if current directory is '/folder1', and call `canonicalize_imp(PathBuf::from("./not-exists/not-exists-too")`, we will get output: `not-exists/not-exists-too`. It is supposed to give the output `/folder1/not-exists/not-exists-too`.

components/tikv_util/src/config.rs
```diff
@@ -457,17 +457,84 @@ impl<'de> Deserialize<'de> for ReadableDuration {
...
+
+        for component in components {
+            match component {
+                Component::Prefix(..) => unreachable!(),
+                Component::RootDir => {
+                    ret.push(component.as_os_str());
+                }
+                Component::CurDir => {}
+                Component::ParentDir => {
+                    ret.pop();
```
MrCroxx, line 479: There is another issue without symbolic links.

If some prefix of input not exists, the following `ParentDir` cannot be parsed properly.

For example:

the current directory is: `"/folder1/folder2"`, the input is `"./not-exists/../../.."`, the output is supposed to be `"/"`, but actually we get output `""` (which means a relative path to the current directory), because it just pop the path component without checking if `ParentDir` move out of the bound.

tabokie: I think `PathBuf` already handled this case: pop on root dir is ignored.

MrCroxx: Emm, my statement above is kind of misleading. I mean: because the method does not canonicalize path before the loop, PahtBuf `ret` can not determine if there is a parent directory of the current directory.

A clearer example: when the current directory is `"/folder1/folder2"`, the input is `"./not-exists/../.."`, the output is supposed to be "/folder1", but actually we got output "".

MrCroxx: [commit#9d2fdd9](https://github.com/tikv/tikv/pull/10128/commits/9d2fdd982fb89c004a65c383f2f4c0a06e3e996a) still get the `pop` problem.

```rust
        println!(
            "current: {:?}\noutput: {:?}",
            canonicalize_imp(PathBuf::from(".")).unwrap(),
            canonicalize_imp(PathBuf::from("./not-exists/../..")).unwrap()
        );
```

output:

```
current: "/home/mrcroxx/github/tikv/tikv/components/tikv_util"
output: "/home/mrcroxx/github/tikv/tikv/components/tikv_util"
```
(which is supposed to be `"/home/mrcroxx/github/tikv/tikv/components"`)

Maybe the proper way is canonicalizing current directory first in `normalize` if `path` is relative, but it is a little bit tricky I think. Maybe we need a future discuss.

tabokie: You are right, so it seems cargo's implementation isn't so solid ;/

components/tikv_util/src/config.rs
```diff
@@ -457,17 +457,104 @@ impl<'de> Deserialize<'de> for ReadableDuration {
...
+                                should_canonicalize = false;
+                            }
+                            other => return other,
+                        }
+                    }
+                }
+                Component::Prefix(..)
+                | Component::RootDir
+                | Component::ParentDir
+                | Component::CurDir => unreachable!(),
```
MrCroxx, line 535: `ParentDir` is no longer unreachable when `nomalize` needs to reserve the prefix `ParentDir`.
```suggestion
                Component::ParentDir => {
                    ret.pop();
                },
                Component::Prefix(..)
                | Component::RootDir
                | Component::CurDir => unreachable!(),
```

components/tikv_util/src/config.rs
```diff
@@ -457,17 +457,104 @@ impl<'de> Deserialize<'de> for ReadableDuration {
...
+        let mut ret = PathBuf::new();
+
+        while let Some(c @ (Component::Prefix(..) | Component::RootDir)) =
+            components.peek().cloned()
+        {
+            components.next();
+            ret.push(c.as_os_str());
+        }
+
+        let mut normal_nodes = 0;
```
MrCroxx, line 473:
```suggestion
```
no longer used

components/tikv_util/src/config.rs
```diff
@@ -457,17 +457,104 @@ impl<'de> Deserialize<'de> for ReadableDuration {
...
+                c @ Component::ParentDir => {
+                    if normal_nodes > 0 {
+                        ret.pop();
+                    } else {
+                        ret.push(c.as_os_str());
+                    }
+                }
+                Component::Normal(c) => {
+                    ret.push(c);
+                    normal_nodes += 1;
```
MrCroxx, line 487:
```suggestion
```
no longer used

MrCroxx (approved):
LGTM

components/tikv_util/src/config.rs
```diff
@@ -1353,15 +1434,51 @@ mod tests {
...
 
-        let path2 = format!("{}", tmp_dir.to_path_buf().join("test2").display());
-        assert!(canonicalize_path(&path2).is_ok());
-        ensure_dir_exist(&path2).unwrap();
-        let res_path2 = canonicalize_path(&path2).unwrap();
-        assert_eq!(
-            Path::new(&res_path2),
-            Path::new(&path2).canonicalize().unwrap()
-        );
+        let cases = vec![".", "/../../", "./../"];
```
skyzh, line 1437: What about the cases like `../wal_dir` when the `wal_dir` folder has not been created?

skyzh (approved):
I think things have gone too complicated there. From https://github.com/tikv/tikv/pull/10102, our original goal was just to check if the two paths are the same directory in the filesystem. From my perspective, the WAL / RocksDB directory should have been created upon this point, either by our user or by TiKV. Because of this, canonicalization from the operating system should always work, and we do not need to handle such complicated cases here.

tabokie:
> I think things have gone too complicated there. From #10102, our original goal was just to check if the two paths are the same directory in the filesystem. From my perspective, the WAL / RocksDB directory should have been created upon this point, either by our user or by TiKV. Because of this, canonicalization from the operating system should always work, and we do not need to handle such complicated cases here.

I agree this change is more complex than anticipated, but necessary in certain (rare) cases.

Semantics-wise, `config::validate()` should not have side effect. And this semantics is relied upon not only by tests, by production code too:
https://github.com/tikv/tikv/blob/2b7ae783b8b321f9d5a8029804b98199dd633551/cmd/tikv-server/src/main.rs#L174-L175

components/tikv_util/src/config.rs
tabokie: I think the tests below can cover it. We don't want to test it in current directory because it's hard to cross validate the result with std.

hunterlxt (approved):
Most LGTM

PR #10108, opened 2021-04-30
cdc: limit scan speed

overvenus:
### What problem does this PR solve?

There will be a series of PRs aims to fix the TiKV CDC OOM issue (too many pending send messages are buffered in memory).

To resolve the issue, we have two strategies,
1. Limit the speed of producing messages during the initial scan stage. #9948 
2. Set an upper bound of pending send messages. #9885 

The series of PRs adopts both strategies.

This PR cherry-picks #9948 to the master branch. Unfortunately, the master branch diverges release-5.0 and it is hard to pick, so I plan to revert flow control #9885 temporarily and will bring it back step by step in other PRs.

Plan:
- take1: cherry-pick speed limiter <-- we are here.
- take2: set an upper bound of internal channel used in CDC.
- take3: add memory quota to CDC and kill connection if it uses too much memory.
- take4: cherry-pick other bugfixes in #9885 

See full changes: https://github.com/tikv/tikv/compare/master...overvenus:cdc/flowcontrol/all

Cc #8168 #9996

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

- No release note.

Issue #9996, opened 2021-04-12
TiKV OOM due to CDC endpoint CPU 100% under 20k insert op/s


overvenus:
## Bug Report

![image](https://user-images.githubusercontent.com/2150711/114354756-39ff3a80-9ba1-11eb-8b85-b0a62a46b03f.png)
![image](https://user-images.githubusercontent.com/2150711/114354859-61560780-9ba1-11eb-9e28-1e546384a609.png)
![image](https://user-images.githubusercontent.com/2150711/114354696-27850100-9ba1-11eb-97cd-f94ffcb8dabe.png)
![image](https://user-images.githubusercontent.com/2150711/114354958-7cc11280-9ba1-11eb-86bd-3d9798859329.png)
![image](https://user-images.githubusercontent.com/2150711/114431686-5c6d7400-9bf2-11eb-8025-7f9e3eeba66d.png)

From the above graph, OOM is caused by endpoint CPU usage is high, TiKV buffers too many scan events in memory.

### What version of TiKV are you using?

TiKV 5.0.0
TiCDC 5.0.0

### What operating system and CPU are you using?

TiKV: 8 CPU, 64GB memory

### What did you expect?

No TiKV OOM

### What did happened?

TiKV OOM

Issue #8168, opened 2020-07-01
TiKV keeps OOM if there are too many change events need to be sent to CDC


overvenus:
## Bug Report

TiKV keeps OOM if there are too many change events need to be sent to CDC.

### What version of TiKV are you using?

v4.0.1

### Steps to reproduce

1. Create a changefeed 
  `cdc cli changefeed create --sink-uri="tidb://root@xxx:xxx/" --pd="http://yyy:yyy"`
2. Pause the change feed
  `cdc cli changefeed pause --changefeed-id xxx --pd="HTTP://yyy:yyy"`
3. Running sysbench for about 1 hour.
  `sysbench --db-driver=mysql --mysql-host=xxx --mysql-port=xxx --mysql-user=root --mysql-db=test --time=30000000 --report-interval=10 --threads=100 oltp_update_non_index --tables=100 --table-size=5000000 prepare && sysbench --db-driver=mysql --mysql-host=xxx --mysql-port=xxx --mysql-user=root --mysql-db=test --time=30000000 --report-interval=10 --threads=50 oltp_update_non_index --tables=100 --table-size=5000000 run`
4. Resume the changefeed
  `cdc cli changefeed resume --changefeed-id xxx --pd="HTTP://yyy:yyy"`

### What did you expect?

No OOM, everything works fine.

### What did happened?

OOM

overvenus:
This issue has been fixed by https://github.com/tikv/tikv/pull/10135

PR #10135, opened 2021-05-08
cdc: add sink memory quota

overvenus:

It limits the total memory used in the sink by setting a memory quota.

Plan:
- [x] take2: set an upper bound of an internal channel used in CDC #10142 
- [ ] take3: add memory quota to CDC and kill connection if it uses too much memory. <-- we are here.

Cc #8168 #9996

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

- Limit CDC sink memory consumption.


overvenus:
/cc @liuzix @amyangfei @ben1009

PTAL

components/cdc/src/channel.rs
```diff
@@ -19,17 +30,64 @@ const CDC_MSG_MAX_BATCH_SIZE: usize = 128;
...
+    }
+    pub fn cap(&self) -> usize {
+        self.capacity
+    }
+    fn alloc(&self, bytes: usize) -> bool {
+        let mut total_bytes = self.in_use.load(Ordering::Relaxed);
+        loop {
+            if total_bytes + bytes > self.capacity {
+                return false;
+            }
```
amyangfei, line 57: In theory this check could have false positive, when `total_bytes + bytes > self.capacity` in L55, but the `total_bytes` has been decreased actually.

components/cdc/src/channel.rs
```diff
@@ -125,13 +201,63 @@ impl Drain {
...
+    }
+}
+
+impl<S: Stream + Unpin> Drop for DrainGrpcMessage<S> {
+    // Must poll all pending messages to free memory quota.
+    fn drop(&mut self) {
+        // Receiver maybe dropped before senders, so set closed flag to prevent
+        // futher messages.
+        self.closed.store(true, Ordering::Release);
+        std::sync::atomic::compiler_fence(Ordering::SeqCst);
```
liuzix, line 238: Why is this necessary?

components/cdc/src/channel.rs
```diff
@@ -125,13 +201,63 @@ impl Drain {
...
+                let item = futures::future::poll_fn(|cx| {
+                    match Pin::new(&mut next).poll(cx) {
+                        // In case it is dropped before senders.
+                        Poll::Pending => Poll::Ready(None),
+                        t => t,
+                    }
+                })
+                .await;
+                if item.is_none() {
+                    break;
```
liuzix, line 252: If `next` yields a `Poll::Pending`, will the loop break here? If so, why is it acceptable?

components/cdc/src/channel.rs
```diff
@@ -125,13 +201,63 @@ impl Drain {
...
+        // Receiver maybe dropped before senders, so set closed flag to prevent
+        // futher messages.
+        self.closed.store(true, Ordering::Release);
+        std::sync::atomic::compiler_fence(Ordering::SeqCst);
+        let start = Instant::now();
+        let mut drain = Box::pin(async {
+            loop {
+                let mut next = self.stream.next();
+                let item = futures::future::poll_fn(|cx| {
+                    match Pin::new(&mut next).poll(cx) {
```
liuzix, line 244: Can we use `poll_unpin`?

components/cdc/src/service.rs
```diff
@@ -356,9 +360,9 @@ impl ChangeData for Service {
         let peer = ctx.peer();
         let scheduler = self.scheduler.clone();
 
-        let mut rx = event_drain.drain_grpc_message();
         ctx.spawn(async move {
-            let res = sink.send_all(&mut rx).await;
+            // let res = sink.send_all(&mut rx).await;
```
liuzix, line 364:
```suggestion
```

liuzix (approved):

amyangfei (approved):

overvenus:
Let's track take4 in #10374 .

PR #9702, opened 2021-02-24
raftstore: enable observing cmds for all regions

5kbpers:
Signed-off-by: 5kbpers <tangminghua@pingcap.com>

### What problem does this PR solve?

Problem Summary:
Separate from #9435

After introducing follower stale read, all commands executed by all leaders need to be observed to advance resolved ts.

### What is changed and how it works?

What's Changed:
This PR temporarily enable to observe commands for all regions.
There will be a mechanism to disable observing for follower peers after we complete merging all codes related to resolved ts.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* N/A

PR #10138, opened 2021-05-08
raftstore: reduce cost of observing applied cmd

NingLin-P:

Problem Summary:

https://github.com/tikv/tikv/pull/9702 enable to observe applied commands for all regions in all peer, which introduce costs:
- Copying all applied commands multiple times
- Send the commands to `cdc` worker and `resolved-ts` worker

Especially the latter one, in a heavy write workload, there may be a lot of messages sending to these two workers, and both `cdc` worker and `resolved-ts` worker are single thread, these unneeded messages could overwhelm the workers, and the pending message will also occupy a lot of memory.

What's Changed:

This PR try to observe lees data and filter data at the apply worker hence avoiding sending unneeded messages.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
- No release note


components/cdc/src/endpoint.rs
```diff
@@ -512,24 +516,26 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Endpoint<T> {
         if is_new_delegate {
             // The region has never been registered.
             // Subscribe the change events of the region.
-            let old_id = self.observer.subscribe_region(region_id, delegate.id);
+            let old_id = self
+                .observer
+                .subscribe_region(region_id, delegate.handle.id);
```
BusyJay, line 521: If assigning `handle.id` to temporary variable, you can save few lines in this file.

components/cdc/src/observer.rs
```diff
@@ -102,18 +103,25 @@ impl CdcObserver {
 impl Coprocessor for CdcObserver {}
 
 impl<E: KvEngine> CmdObserver<E> for CdcObserver {
-    fn on_prepare_for_apply(&self, cdc_id: ObserveID, rts_id: ObserveID, region_id: u64) {
+    fn on_prepare_for_apply(&self, cdc: &ObserveHandle, rts: &ObserveHandle, region_id: u64) {
+        *self.last_batch_observing.borrow_mut() = cdc.is_observing();
```
BusyJay, line 107: If assigning it to a temporary variable, you don't need to borrow it twice.

components/cdc/src/observer.rs
```diff
@@ -345,6 +356,28 @@ mod tests {
             _ => panic!("unexpected task"),
         };
 
+        // Stop observing cmd
+        observe_handle.stop_observing();
+        <CdcObserver as CmdObserver<RocksEngine>>::on_prepare_for_apply(
```
BusyJay, line 361: Why not `observer.on_prepare_for_apply(...)`?

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -175,6 +175,48 @@ impl Cmd {
...
+pub struct ObserveID(usize);
+
+impl ObserveID {
+    pub fn new() -> ObserveID {
+        ObserveID(OBSERVE_ID_ALLOC.fetch_add(1, Ordering::SeqCst))
+    }
+}
+
+#[derive(Clone, Default, Debug)]
+pub struct ObserveHandle {
```
BusyJay, line 191: How about adding some comment explaining what it's used for?

BusyJay (commented):
What's the improvement in the latest patch?

NingLin-P:
> What's the improvement in the latest patch?

On the master branch, all applied commands will clone twice and send to the `resolved-ts` and `cdc` worker thread, and most of them will be rejected by the workers. This PR only send these applied commands when the worker needed, normally send to `resolved-ts` if the peer is leader, send to `cdc` if the peer is leader and `ticdc` is subscribe to this region, and if not need to send the commands there will be no cloning.

components/cdc/src/observer.rs
```diff
@@ -103,9 +103,11 @@ impl CdcObserver {
 impl Coprocessor for CdcObserver {}
 
 impl<E: KvEngine> CmdObserver<E> for CdcObserver {
+    #[allow(unused_assignments)]
     fn on_prepare_for_apply(&self, cdc: &ObserveHandle, rts: &ObserveHandle, region_id: u64) {
-        *self.last_batch_observing.borrow_mut() = cdc.is_observing();
-        if !*self.last_batch_observing.borrow() {
+        let mut last_batch_observing = *self.last_batch_observing.borrow_mut();
```
BusyJay, line 108: This is wrong, `last_batch_observing` is a copy instead of a reference to `self.last_batch_observing`.

Why not
```
let is_observing = cdc.is_obsering();
*self.last_batch_observing.borrow_mut() = is_observing;
if !is_observing {
...
```
```

File: components/resolved_ts/src/observer.rs [hunk 6]
```diff
@@ -54,25 +55,36 @@ impl<E: KvEngine> Clone for Observer<E> {
...
 }
 
 impl<E: KvEngine> Coprocessor for Observer<E> {}
 
 impl<E: KvEngine> CmdObserver<E> for Observer<E> {
-    fn on_prepare_for_apply(&self, cdc_id: ObserveID, rts_id: ObserveID, region_id: u64) {
+    #[allow(unused_assignments)]
+    fn on_prepare_for_apply(&self, cdc: &ObserveHandle, rts: &ObserveHandle, region_id: u64) {
+        // TODO: Should not care about whether `cdc` is observing
+        let mut last_batch_observing = *self.last_batch_observing.borrow_mut();
```
5kbpers, line 69: ditto

5kbpers (approved):

components/cdc/src/observer.rs
```diff
@@ -102,18 +103,27 @@ impl CdcObserver {
 impl Coprocessor for CdcObserver {}
 
 impl<E: KvEngine> CmdObserver<E> for CdcObserver {
-    fn on_prepare_for_apply(&self, cdc_id: ObserveID, rts_id: ObserveID, region_id: u64) {
+    #[allow(unused_assignments)]
```
BusyJay, line 106: Why allow?

NingLin-P: A mistake, removed.

BusyJay (approved):

PR #10142, opened 2021-05-10
cdc: back pressure scan speed

overvenus:

This is part of #9885, it backpressure scan speed by replacing an unbounded channel with a bounded channel.

Plan:
- [ ] take2: set an upper bound of an internal channel used in CDC <-- we are here.
- [ ] take3: add memory quota to CDC and kill connection if it uses too much memory.

Cc #8168 #9996

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

- Support back pressure CDC scan speed.


overvenus:
Cc @liuzix

components/cdc/src/channel.rs
```diff
@@ -0,0 +1,233 @@
...
+use tikv_util::impl_display_as_debug;
+
+use crate::service::{CdcEvent, EventBatcher};
+
+const CDC_MSG_MAX_BATCH_SIZE: usize = 128;
+// Assume the average size of event is 1KB.
+// 2 = (CDC_MSG_MAX_BATCH_SIZE * 1KB / CDC_EVENT_MAX_BATCH_SIZE).ceil() + 1 /* reserve for ResolvedTs */;
+pub const CDC_EVENT_MAX_BATCH_SIZE: usize = 2;
+
+pub fn canal(buffer: usize) -> (Sink, Drain) {
```
5kbpers, line 22: typo

components/cdc/src/channel.rs
```diff
@@ -0,0 +1,233 @@
...
+            bounded_receiver,
+        },
+    )
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SendError {
+    Full,
+    Disconnected,
+    Congest,
```
5kbpers, line 41: Should be `Congested`?

components/cdc/src/channel.rs
```diff
@@ -0,0 +1,233 @@
...
+impl std::error::Error for SendError {}
+
+impl_display_as_debug!(SendError);
+
+macro_rules! impl_from_future_send_error {
+    ($($f:ty,)+) => {
+        $(
+            impl From<$f> for SendError {
+                fn from(e: $f) -> Self {
+                    if e.is_disconnected() {
```
5kbpers, line 53: How about using pattern matching like 
```rust
match e  {
    $f::Disconnected => {...},
}
```

components/cdc/src/delegate.rs
```diff
@@ -993,113 +1000,4 @@ mod tests {
         assert!(err.has_epoch_not_match());
         assert!(err.take_epoch_not_match().current_regions.is_empty());
     }
-
-    #[test]
-    fn test_scan() {
```
5kbpers, line 998: Should we add this test back later?

components/cdc/src/endpoint.rs
```diff
@@ -1132,13 +1149,16 @@ impl Initializer {
...
             debug!("cdc scan entries"; "len" => entries.len(), "region_id" => region_id);
             fail_point!("before_schedule_incremental_scan");
-            let scanned = Task::IncrementalScan {
-                region_id,
-                downstream_id,
-                entries,
-            };
-            if let Err(e) = self.sched.schedule(scanned) {
-                error!("cdc schedule cdc task failed"; "error" => ?e, "region_id" => region_id);
+            if let Err(e) = self.sink_scan_events(entries, done, require_barrier).await {
```
5kbpers, line 1152: If we send the scan event here, then ticdc could still receive scanned entries after a region had been deregistered. Is that OK?

5kbpers (commented):
rest LGTM

components/cdc/src/channel.rs
overvenus: `futures::channel::mpsc::{SendError, TrySendError}` are not enums. 😅

components/cdc/src/delegate.rs
overvenus: It tests `on_scan` but it is removed in this PR, so I think we can just remove the test too.

components/cdc/src/endpoint.rs
overvenus: @amyangfei PTAL

components/cdc/src/channel.rs
overvenus: See more: https://docs.rs/futures/0.3.14/futures/channel/mpsc/struct.SendError.html

PR #10145, opened 2021-05-10
cdc: back pressure scan speed (#10142)

overvenus:
This PR cherry-pick #10142

----------------

This is part of #9885, it backpressure scan speed by replacing an unbounded channel with a bounded channel.

Plan:
- [x] take2: set an upper bound of an internal channel used in CDC #10142 <-- we are here.
- [ ] take3: add memory quota to CDC and kill connection if it uses too much memory.

Cc #8168 #9996

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

- Support back pressure CDC scan speed.


components/cdc/src/endpoint.rs
```diff
@@ -886,13 +905,8 @@ impl Initializer {
...
             debug!("cdc scan entries"; "len" => entries.len(), "region_id" => region_id);
             fail_point!("before_schedule_incremental_scan");
-            let scanned = Task::IncrementalScan {
-                region_id,
-                downstream_id,
-                entries,
-            };
-            if let Err(e) = self.sched.schedule(scanned) {
-                error!("cdc schedule cdc task failed"; "error" => ?e, "region_id" => region_id);
+            if let Err(e) = self.sink_scan_events(entries, done, require_barrier).await {
```
5kbpers, line 908: Please confirm ticdc checks `RequestId` in release-4.0 too.

5kbpers (approved):

components/cdc/src/endpoint.rs
overvenus: Confirmed, release-4.0 can filter events by `RequestId`. https://github.com/pingcap/ticdc/blob/release-4.0/cdc/kv/client.go#L1243-L1257

PR #10155, opened 2021-05-12
statistics: resource metering

zhongzc:

Support CPU resource metering for TiKV RPC requests.

What's Changed:

- Support CPU time recording for requests
  - A request is with a `ResourceMeteringTag`.
  - During the execution of the request, its `ResourceMeteringTag` is attached to the current thread.
  - There is a global `CpuRecorder` running in a background thread. This `CpuRecorder` periodicity fetches stats of threads along with their attached tags, then it records CPU time of related requests.
- Introduce a reporter to periodicity send records to some agent.
- Support dynamically configuring the behavior of the recorder and the reporter.
   - `resource-metering.enabled`: Enable resource metering, for now, only CPU recorder involved. (default: **false**)
   - `resource-metering.agent-address`: Remote agent address (default: **""**)
   - `resource-metering.report-agent-interval-seconds` (default: **60**)
   - `resource-metering.max-resource-groups`: Control memory usage of the reporter. (default: **200**)
   - `resource-metering.precision-seconds`: Control the precision of recording. (default: **1**)
 
### Related changes

- Agent protocol: https://github.com/pingcap/kvproto/pull/773

### Check List

Tests 

- Unit test
- Integration test
- Manual test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- Support basic resource metering


components/req_cpu/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+#![feature(shrink_to)]
```
breeswish, line 3: Please add some basic doc comment to help others understand this module.

components/req_cpu/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+        Self {
+            record_interval_ms: 10.1,
+            collect_interval_ms: 1_000,
+            gc_interval_ms: 10 * 60 * 1_000,
+        }
+    }
+}
+
+#[derive(Debug, Default, Eq, PartialEq, Clone, Hash)]
+pub struct RequestTag {
```
breeswish, line 42: A better name is preferred. There are too many "tags". How about ResourceMeteringTag

components/req_cpu/src/dummy.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use crate::collector::{Collector, CollectorId};
+use crate::{Builder, RecorderConfig, RequestTag};
+
+use std::marker::PhantomData;
+use std::sync::{Arc, Mutex};
+
+impl RequestTag {
+    pub fn attach(self: &Arc<Self>) -> Guard {
```
breeswish, line 10: Please add some comments about what does this function do

src/storage/mod.rs
```diff
@@ -836,6 +849,10 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
                     Self::with_tls_engine(|engine| Self::snapshot(engine, snap_ctx)).await?;
                 {
                     let begin_instant = Instant::now_coarse();
+
+                    let req_tag = Arc::new(RequestTag::from_rpc_context(&ctx));
```
breeswish, line 853: Seems that there is always a Arc around. May be opt-in?

components/req_cpu/src/linux.rs
```diff
@@ -0,0 +1,690 @@
...
+        .send(CollectorRegistrationMsg::Register { collector, id })
+        .ok();
+    CollectorHandle { id }
+}
+
+impl RequestTag {
+    pub fn attach(self: &Arc<Self>) -> Guard {
+        CURRENT_REQ.with(|s| {
+            if s.is_set.get() {
+                panic!("Nested attachment is not allowed.")
```
breeswish, line 64: In what case will this panic happen?

components/req_cpu/src/linux.rs
```diff
@@ -0,0 +1,690 @@
...
+    }
+}
+
+struct LocalReqTag {
+    is_set: Cell<bool>,
+    shared_ptr: SharedReqTagPtr,
+}
+thread_local! {
+    static CURRENT_REQ: LocalReqTag = {
+        let thread_id = unsafe { libc::syscall(libc::SYS_gettid) as libc::pid_t };
```
breeswish, line 99: Could you extract real platform-dependent code into a standalone one, and leave the rest shared?

components/req_cpu/src/linux.rs
```diff
@@ -0,0 +1,690 @@
...
+                        ));
+                    }
+                })
+                .expect("Failed to create recorder thread");
+        };
+    }
+    *PHANTOM
+}
+
+pub fn register_collector(collector: Box<dyn Collector>) -> CollectorHandle {
```
breeswish, line 47: What is a collector?

components/req_cpu/src/linux.rs
```diff
@@ -0,0 +1,690 @@
...
+            current_window_records: RequestCpuRecords::default(),
+
+            collectors: HashMap::default(),
+        }
+    }
+
+    pub fn handle_collector_registration(&mut self) {
+        let mut should_reset = false;
+        loop {
+            while let Ok(msg) = COLLECTOR_REGISTRATION_CHANNEL.1.try_recv() {
```
breeswish, line 204: Do we need to add sleep in case of no collector is registered?

components/req_cpu/src/linux.rs
```diff
@@ -0,0 +1,690 @@
...
+                    if cur_tag.is_some() {
+                        thread_stat.prev_tag = cur_tag;
+                        thread_stat.prev_stat = stat;
+                    }
+                }
+            }
+        }
+    }
+
+    pub fn may_gc(&mut self) -> bool {
```
breeswish, line 285: Gc might not be a correct wording. Gc means to clean up some garbage, while we are not cleaning garbages..

components/req_cpu/src/linux.rs
zhongzc: In that case, the following  `COLLECTOR_REGISTRATION_CHANNEL.1.recv()` works like sleep.

components/req_cpu/src/linux.rs
zhongzc: Dead threads are garbages and this function will clean them.

components/req_cpu/src/linux.rs
zhongzc: For now, it can not happen. Tagged tasks are executed one by one; no nested tasks exist.

components/req_cpu/src/linux.rs
zhongzc: Collectors should handle records sent by the recorder per second.

components/resource_metering/src/agent.rs
```diff
@@ -0,0 +1,193 @@
...
+    // TODO: mock client for testing
+    client: Option<ResourceUsageAgentClient>,
+    cpu_records_collector: Option<CollectorHandle>,
+}
+
+impl ResourceMeteringAgent {
+    pub fn new(
+        config: Config,
+        env: Arc<Environment>,
+        security_mgr: Arc<SecurityManager>,
```
breeswish, line 151: Do we need some separated security config? Since as a collect agent it only need to accept connections from this TiKV and it doesn't need intra-cluster connections.

components/resource_metering/src/agent.rs
```diff
@@ -0,0 +1,193 @@
...
+            self.security_mgr.connect(cb, &self.config.agent_address)
+        };
+        self.client = Some(ResourceUsageAgentClient::new(channel));
+    }
+}
+
+impl Runnable for ResourceMeteringAgent {
+    type Task = Task;
+
+    fn run(&mut self, _task: Self::Task) {}
```
breeswish, line 182: Need to implement something to respond the config change?

components/resource_metering/src/cpu/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+mod linux;
+#[cfg(target_os = "linux")]
+pub use linux::{init_recorder, register_collector, CollectorHandle, Guard};
+
+#[cfg(not(target_os = "linux"))]
+mod dummy;
+#[cfg(not(target_os = "linux"))]
+pub use dummy::{init_recorder, register_collector, CollectorHandle, Guard};
+
+#[derive(Debug, Clone)]
```
breeswish, line 25: Is it possible to **not** derive the `Clone` for it? (This will help us find places where we mistakenly cloned the content, instead of cloning the Arc reference).

components/resource_metering/src/reporter.rs
```diff
@@ -0,0 +1,211 @@
...
+                    let tag = &tag.infos.extra_attachment;
+                    if tag.is_empty() {
+                        continue;
+                    }
+
+                    let ms = *ms as u32;
+                    match self.records.get_mut(tag) {
+                        Some((ts, cpu_time, total)) => {
+                            if *ts.last().unwrap() == timestamp_secs {
+                                *cpu_time.last_mut().unwrap() += ms;
```
breeswish, line 137: In what case will this happen?

components/resource_metering/src/reporter.rs
```diff
@@ -0,0 +1,211 @@
...
+                    for (_, _, total) in self.records.values() {
+                        self.find_top_k.push(*total);
+                    }
+                    pdqselect::select_by(
+                        &mut self.find_top_k,
+                        self.config.max_resource_groups,
+                        |a, b| b.cmp(a),
+                    );
+                    let kth = self.find_top_k[self.config.max_resource_groups];
+                    self.records.retain(|_, (_, _, total)| *total >= kth);
```
breeswish, line 162: Can we have a benchmark for this part?

components/resource_metering/src/reporter.rs
zhongzc: Records with the same `extra_attachment` but different ids in a batch

components/resource_metering/src/reporter.rs
zhongzc: It will cost under a microsecond and normally 2x faster than sort_unstable.

breeswish (approved):

andylokandy (approved):

PR #10164, opened 2021-05-13
*: Add `abort-on-panic` config

yiwu-arbug:

Problem Summary:
* Add a config `abort-on-panic` to allow TiKV generate a core dump. This is useful when core dump is desire to investigate issues.
* Adding message in panic_mark_file fatal log to warn users not to delete the panic_mark_file and force restart.
* Adding a warning when tikv-ctl encounter rocksdb LOCK file conflict, which warns user not to delete the LOCK file.

What's Changed: Add the above two configs.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`: will send PR later.
- Need to cherry-pick to the release branch: release-5.0

### Check List 

Tests 
Manually test the flows

### Release note 
* Add `abort-on-panic` config, which allow core dump to be generated when panic. Users still need to correctly config the environment to enable core dump.


components/server/src/server.rs
```diff
@@ -368,10 +368,18 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         self.lock_files.push(f);
 
         if tikv_util::panic_mark_file_exists(&self.config.storage.data_dir) {
-            fatal!(
-                "panic_mark_file {} exists, there must be something wrong with the db.",
-                tikv_util::panic_mark_file_path(&self.config.storage.data_dir).display()
-            );
+            if self.config.allow_restart_on_data_corruption {
```
BusyJay, line 371: This doesn't seem to be a good idea. Once there is such a configuration, it will be abused and we will lose the chance to find what's actually happening.

etc/config-template.toml
```diff
@@ -39,6 +39,15 @@
 ## It won't take effect when compiling without BCC_IOSNOOP=1. 
 # enable-io-snoop = true
 
+## Allow core dump being generated when TiKV panics. It is default to off to prevent
+## core dump from filling up disk space, especially when TiKV fall in crash loop.
+# allow-core-dump = false
```
BusyJay, line 44: How about `abort-on-panic`?

components/server/src/server.rs
yiwu-arbug: how about make it hidden and not mention in config template and docs? For a user they see intermittent corruption error where the physical file is not actually corrupted on retry. We need a temporary workaround until we get a better fix (i.e. auto-recovery on SST corruption).

etc/config-template.toml
yiwu-arbug: What are the other difference between abort and exit? I think core dump is more descriptive to what users can achieve with this config.

yiwu-arbug:
Piggybacking two changes:
* add content in panic_mark_file to warn user not to force restart TiKV
* add warning if tikv-ctl encounter rocksdb LOCK file conflict which warn user not to delete LOCK file.

components/tikv_util/src/lib.rs
```diff
@@ -62,14 +63,27 @@ pub fn panic_mark_is_on() -> bool {
 }
 
 pub const PANIC_MARK_FILE: &str = "panic_mark_file";
+pub const PANIC_MARK_FILE_CONTENT: &[u8] = b"\
+    The panic mark file is generated once TiKV encounter unrecoverable data corruption. \
+    The TiKV node will not restart. Do NOT delete the panic mark file to force TiKV \
+    restart. It would not fix the data corruption issue and could make the corruption \
+    more widespread. Please replace the TiKV node instead.\n";
```
yiwu-arbug, line 70: @Connor1996 I would like to link to a public document on how to replace the TiKV node. Any suggestion?

etc/config-template.toml
BusyJay: Actually TiKV can't control whether to create a coredump, it's determined by several factors, all TiKV can do is just sending an abort signal.

components/tikv_util/src/lib.rs
Connor1996: If there is a panic mark, it probably due to unknown bugs. I suggest to contact us to investigate instead of replacing the node by themselves.

components/server/src/server.rs
Connor1996: the temporary workaround is to delete the mark file manually.  I don't think this is really necessary.

etc/config-template.toml
yiwu-arbug: ok, will change the name.

components/server/src/server.rs
yiwu-arbug: discuss with @Connor1996 offline. It may not be a good idea to provide such an option. I will remove it.

components/server/src/server.rs
```diff
@@ -276,7 +276,7 @@ impl<ER: RaftEngine> TiKVServer<ER> {
 
         check_system_config(&config);
 
-        tikv_util::set_panic_hook(false, &config.storage.data_dir);
+        tikv_util::set_panic_hook(config.allow_core_dump, &config.storage.data_dir);
```
Connor1996, line 279: maybe generate core dump only for the critical error which sets panic mark. For other panics, it may abort again and again, causing disk full eventually.

yiwu-arbug: Can we leave for the user to decide what's the best? For example, for rare memory corruption or segfault, no panic mark will be created, but there won't be crash loop as well.

Connor1996: I believe the core dump is only used by us. When we want to investigate something, we let the user enable this config and wait for the core dump is generated.

yiwu-arbug: Since block cache is excluded from core dump, I think the core dump size would not be too large. There's still cases where we want to have core dump in cases other than panic mark. What do you think?

Connor1996: okay

Connor1996 (commented):
rest LGTM

components/engine_rocks/src/raw_util.rs
```diff
@@ -157,7 +157,7 @@ pub fn new_engine_opt(
         }
     }
     let cfds = cfs_v.into_iter().zip(cfs_opts_v).collect();
-    let mut db = DB::open_cf(db_opt, path, cfds).unwrap();
+    let mut db = DB::open_cf(db_opt, path, cfds)?;
```
BusyJay, line 160: Is it safe to return error?

cmd/tikv-ctl/src/main.rs
```diff
@@ -54,13 +54,29 @@ const METRICS_ROCKSDB_KV: &str = "rocksdb_kv";
...
+const LOCK_FILE_ERROR: &str = "IO error: While lock file";
+
 type MvccInfoStream = Pin<Box<dyn Stream<Item = Result<(Vec<u8>, MvccInfo), String>>>>;
 
 fn perror_and_exit<E: Error>(prefix: &str, e: E) -> ! {
     ve1!("{}: {}", prefix, e);
     process::exit(-1);
 }
 
+fn handle_engine_error(err: EngineError) {
```
BusyJay, line 66: You can make it return `!` then `unreachable!()` is unnecessary.

components/server/src/server.rs
```diff
@@ -369,7 +369,11 @@ impl<ER: RaftEngine> TiKVServer<ER> {
 
         if tikv_util::panic_mark_file_exists(&self.config.storage.data_dir) {
             fatal!(
-                "panic_mark_file {} exists, there must be something wrong with the db.",
+                "panic_mark_file {} exists, there must be something wrong with the db. \
+                     Do not remove the panic_mark_file and force the TiKV node to restart. \
```
BusyJay, line 373: The line seems redundant, people may not realize it can be deleted.

yiwu-arbug: We have seen several user who upon seeing the error just manually delete the panic mark and force TiKV restart. We should have clear message to prevent them doing so.

components/engine_rocks/src/raw_util.rs
yiwu-arbug: Caller should do the right thing when getting the error, I think?

components/tikv_util/src/lib.rs
yiwu-arbug: Some community user may maintain the cluster themselves and sometimes would like to recover the node as soon as possible.

components/engine_rocks/src/raw_util.rs
BusyJay: I'm not sure if the error is handled properly by panicking.

yiwu-arbug:
ping @BusyJay @Connor1996

components/engine_rocks/src/raw_util.rs
yiwu-arbug: RocksDB returning error on open is a clean state. No need to panic actually.

BusyJay: Then how to recover from the error?

BusyJay (approved):
Rest LGTM

Connor1996 (approved):
LGTM

PR #10171, opened 2021-05-14
importer: Detect duplicate data importing by Lightning.

Little-Wallace:

Issue Number: part of https://github.com/pingcap/br/issues/1110  

Problem Summary:
We need detect duplicate keys from source data when we use Lightning to import data to an initial TiKV cluster. The disk of Lightning is limited, so we can not keep all source data in one disk of Lightning.  

Proposal: [xxx](url) 
We will use TiKV cluster to store all data which may be imported by difference Lightning. Every lightning instance will use an  unique commit-ts for their request. When importing is end, we will collect the duplicate keys from TiKV, which have a multiple version in MVCC.

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- No release note.

### DIFF
--- a/components/encryption/export/Cargo.toml
+++ b/components/encryption/export/Cargo.toml
@@ -36,15 +36,7 @@ derive_more = "0.99.3"
 encryption = { path = "../", default-features = false }
 error_code = { path = "../../error_code", default-features = false }
 file_system = { path = "../../file_system", default-features = false }
-
-
-
-
 kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
-
-
-
-
 openssl = "0.10"
 protobuf = "2.8"
 slog = { version = "2.3", features = ["max_level_trace", "release_max_level_debug"] }
--- a/components/encryption/export/Cargo.toml
+++ b/components/encryption/export/Cargo.toml
@@ -36,7 +36,15 @@ derive_more = "0.99.3"
 encryption = { path = "../", default-features = false }
 error_code = { path = "../../error_code", default-features = false }
 file_system = { path = "../../file_system", default-features = false }
-kvproto = { rev = "dda0a102bc6ac7371e1f99cd2797825aa210b4a0", git = "https://github.com/pingcap/kvproto.git", default-features = false }
+
+
+
+
+kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
+
+
+
+
 openssl = "0.10"
 protobuf = "2.8"
 slog = { version = "2.3", features = ["max_level_trace", "release_max_level_debug"] }
--- a/src/import/duplicate_detect.rs
+++ b/src/import/duplicate_detect.rs
@@ -9,10 +9,12 @@ use tikv_util::keybuilder::KeyBuilder;
 use txn_types::{Key, TimeStamp, WriteRef, WriteType};
 
 #[cfg(test)]
-const MAX_SCAN_BATCH_SIZE: usize = 256;
+const MAX_SCAN_BATCH_COUNT: usize = 256;
 
 #[cfg(not(test))]
-const MAX_SCAN_BATCH_SIZE: usize = 4096;
+const MAX_SCAN_BATCH_COUNT: usize = 4096;
+
+const MAX_SCAN_BATCH_SIZE: usize = 1024 * 1024; // 1MB
 
 pub struct DuplicateDetector<S: Snapshot> {
     snapshot: S,
@@ -55,6 +57,7 @@ impl<S: Snapshot> DuplicateDetector<S> {
         let mut last_value = Vec::with_capacity(256);
         let mut last_commit_ts = TimeStamp::zero();
         let mut first_duplicate = true;
+        let mut batch_size = 0;
         while self.iter.valid().map_err(from_kv_error)? {
             let key = self.iter.key();
             let (user_key, commit_ts) = Key::split_on_ts_for(key)?;
@@ -72,13 +75,19 @@ impl<S: Snapshot> DuplicateDetector<S> {
                 }
                 WriteType::Put => {
                     if !last_key.is_empty() && user_key.eq(&last_key) {
+                        batch_size += user_key.len();
                         if first_duplicate {
-                            // println!("last value lenth: {}, ts: {:?}", last_value.len(), last_commit_ts);
+                            batch_size += last_key.len();
                             if self.key_only {
                                 ret.push(self.make_kv_pair(&last_key, None, last_commit_ts)?);
                             } else {
                                 let last_write =
                                     WriteRef::parse(&last_value).map_err(from_txn_types_error)?;
+                                batch_size += last_write
+                                    .short_value
+                                    .as_ref()
+                                    .map(|v| v.len())
+                                    .unwrap_or(0);
                                 ret.push(self.make_kv_pair(
                                     &last_key,
                                     Some(last_write),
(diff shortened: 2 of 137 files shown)


src/import/sst_service.rs
```diff
@@ -548,12 +573,76 @@ where
...
+                                "connection send message fail";
+                                "err" => %e
+                            );
+                        }
+                    }
+                    let _ = sink.close().await;
+                    return;
+                }
+            };
+            let detector = DuplicateDetector::new(snapshot, start_key, end_key, key_only).unwrap();
```
gozssky, line 629: It seem `min_commit_ts` in request is not used?

components/encryption/export/Cargo.toml
```diff
@@ -36,7 +36,15 @@ derive_more = "0.99.3"
...
-kvproto = { rev = "dda0a102bc6ac7371e1f99cd2797825aa210b4a0", git = "https://github.com/pingcap/kvproto.git", default-features = false }
+
+
+
+
+kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
+
+
+
+
```
gozssky, line 47: redundant blank lines?

src/import/sst_service.rs
Little-Wallace: Yes.... I just left it for incremental-import

src/import/duplicate_detect.rs
```diff
@@ -0,0 +1,306 @@
...
+                            if self.key_only {
+                                ret.push(self.make_kv_pair(&last_key, None, last_commit_ts)?);
+                            } else {
+                                let last_write =
+                                    WriteRef::parse(&last_value).map_err(from_txn_types_error)?;
+                                batch_size += last_write
+                                    .short_value
+                                    .as_ref()
+                                    .map(|v| v.len())
+                                    .unwrap_or(0);
```
MyonKeminta, line 90: Do  you need to calculate the values in default cf (non-short values)?

src/import/sst_service.rs
```diff
@@ -98,6 +106,37 @@ where
...
+        let mut cmd = RaftCmdRequest::default();
+        cmd.set_header(header);
+        cmd.set_requests(vec![req].into());
+        let (cb, future) = paired_future_callback();
+        if let Err(e) = router.send_command(cmd, Callback::Read(cb)) {
+            return Err(e.into());
+        }
+        let mut res = future.await.map_err(|_| {
+            let mut err = errorpb::Error::default();
+            let err_str = "too many sst files are ingesting";
```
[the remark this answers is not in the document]
Little-Wallace: Yes. This error only occurs when the thread is closed or full.

src/import/duplicate_detect.rs
```diff
@@ -0,0 +1,306 @@
...
+        if ret.is_empty() {
+            return Ok(None);
+        }
+        Ok(Some(ret))
+    }
+
+    fn make_kv_pair<'a>(
+        &self,
+        key: &[u8],
+        write: Option<WriteRef<'a>>,
```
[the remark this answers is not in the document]
Little-Wallace: Because writeref need a lifetime bound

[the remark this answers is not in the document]
MyonKeminta: Doesn't `'_` work here?

src/import/duplicate_detect.rs
```diff
@@ -56,110 +55,126 @@ impl<S: Snapshot> DuplicateDetector<S> {
...
+        Ok(None)
+    }
 
-            last_key.clear();
-            last_key.extend_from_slice(user_key);
-            last_commit_ts = commit_ts;
+    fn skip_all_version(&mut self, start_key: &[u8]) -> Result<()> {
+        while self.iter.valid().map_err(from_kv_error)? {
+            let (current_key, _) = Key::split_on_ts_for(self.iter.key())?;
+            if !current_key.eq(start_key) {
```
MyonKeminta, line 159: Why not use the `==` and `!=` operators when comparing keys?

src/import/duplicate_detect.rs
```diff
@@ -67,48 +67,42 @@ impl<S: Snapshot> DuplicateDetector<S> {
...
     }
 
-    fn read_next_user_key(&mut self, duplicate_pairs: &mut Vec<KvPair>) -> Result<()> {
-        let start_key = self.iter.key();
-        let (current_key, commit_ts) = Key::split_on_ts_for(start_key)?;
-        if commit_ts <= self.min_commit_ts {
+    fn move_to_next_import_key(&mut self) -> Result<Option<(Vec<u8>, TimeStamp)>> {
+        while self.iter.valid().map_err(from_kv_error)? {
+            let (current_key, commit_ts) = Key::split_on_ts_for(self.iter.key())?;
+            if commit_ts > self.min_commit_ts {
```
MyonKeminta, line 73: Is there any case that your `min_commit_ts` equals to any version's `commit_ts`?

src/import/duplicate_detect.rs
```diff
@@ -67,48 +67,42 @@ impl<S: Snapshot> DuplicateDetector<S> {
...
+
+    fn collect_current_key_duplicate(
+        &mut self,
+        start_key: Vec<u8>,
+        end_commit_ts: TimeStamp,
+        duplicate_pairs: &mut Vec<KvPair>,
+    ) -> Result<()> {
         let write = WriteRef::parse(self.iter.value()).map_err(from_txn_types_error)?;
-        match write.write_type {
+        let mut write_info = match write.write_type {
```
MyonKeminta, line 88: I'm afraid you may need to check the gc_fence to ignore the invalidated write records.

MyonKeminta (dismissed):

src/import/duplicate_detect.rs
Little-Wallace: No, it is an import key.

MyonKeminta (approved):
Special cases like records with gc fences and commit_ts == self.min_commit_ts needs test coverage. Rest LGTM.

src/import/duplicate_detect.rs
Little-Wallace: Because some language `==` just compare pointer address rather than the content of struct.

src/import/duplicate_detect.rs
```diff
@@ -0,0 +1,490 @@
...
+            }
+        }
+        Ok(())
+    }
+
+    fn skip_lock_and_rollback(&mut self, start_key: &[u8]) -> Result<Option<Write>> {
+        self.iter.next().map_err(from_kv_error)?;
+        while self.iter.valid().map_err(from_kv_error)? {
+            let (current_key, _) = Key::split_on_ts_for(self.iter.key())?;
+            if !current_key.eq(start_key) {
```
kennytm, line 144:
```suggestion
            if current_key != start_key {
```

src/import/duplicate_detect.rs
kennytm: @Little-Wallace I don't see how languages other than Rust matter here.

src/storage/mod.rs
```diff
@@ -120,7 +120,7 @@ pub type Callback<T> = Box<dyn FnOnce(Result<T>) + Send>;
 /// encoding and appending timestamp.
 pub struct Storage<E: Engine, L: LockManager> {
     // TODO: Too many Arcs, would be slow when clone.
-    engine: E,
+    pub engine: E,
```
kennytm, line 123: 🤔 

`storage.engine` is only used in tests. can you instead create a function like


```rust
#[cfg(test)]
pub fn xxxxx(&self) -> Result<E::Snap> {
    self.engine.snapshot(Default::default())
}
```

or investigate whether using `storage.get_engine().snapshot(Default::default())` in the tests is ok.

src/import/duplicate_detect.rs
```diff
@@ -0,0 +1,490 @@
...
+        write_data(&storage, data, 3);
+        let mut data = vec![];
+        let big_value = vec![1u8; 300];
+        for i in 0..1000 {
+            let key = format!("{}", i * 2);
+            data.push((key.as_bytes().to_vec(), big_value.clone()))
+        }
+        write_data(&storage, data, 5);
+        // We have to do the prewrite manually so that the mem locks don't get released.
+        let snapshot = storage.engine.snapshot(Default::default()).unwrap();
```
kennytm, line 368: (for reference)

src/storage/mod.rs
Little-Wallace: OK

kennytm (approved):

PR #10177, opened 2021-05-16
copr:  fix the wrong arguments type of json_unquote

### LINKED ISSUE #10176: json_unquote: Expect `Json`, received `Bytes`
5.0

AMD Ryzen 5 PRO 4650U with Radeon Graphics

Ubuntu 18.04

### Steps to reproduce
This is a **coprocessor bug**. And should enable push down of `json_unquote`, which is not by default. But still need to fix it.

    drop table if exists t1;
    create table t1(c1 varchar(100));
    insert into t1 values('abc'), ('"abc"');
    select c1 from t1 where json_unquote(c1) = 'abc';

### What did you expect?
    mysql>     select c1 from t1 where json_unquote(c1) = 'abc';
    +-------+
    | c1    |
    +-------+
    | abc   |
    | "abc" |
    +-------+
    2 rows in set (0.00 sec)
### What did happened?

    mysql> select c1 from t1 where json_unquote(c1) = 'abc';
    ERROR 1105 (HY000): other error: [components/tidb_query_expr/src/types/expr_builder.rs:295]: Invalid json_unquote (sig = JsonUnquoteSig) signature: Evaluate error: [components/tidb_query_expr/src/types/function.rs:272]: Expect `Json`, received `Bytes`
    m

guo-shaoge:

Issue Number: close #10176 

Problem Summary: `json_unquote` implementation use `JsonRef` as argument type, which is not correct. It causes the following SQL report errors.

    drop table if exists t1;
    create table t1(c1 varchar(100));
    insert into t1 values('abc'), ('"abc"');
    select c1 from t1 where json_unquote(c1) = 'abc';

    mysql> select c1 from t1 where json_unquote(c1) = 'abc';
    ERROR 1105 (HY000): other error: [components/tidb_query_expr/src/types/expr_builder.rs:295]: Invalid json_unquote (sig = JsonUnquoteSig) signature: Evaluate error: [components/tidb_query_expr/src/types/function.rs:272]: Expect `Json`, received `Bytes`
    m

What's Changed: change argument type from `JsonRef` to `BytesRef`.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 
- copr: fix the wrong arguments type of json_unquote


guo-shaoge:
@skyzh PTAL ~

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -226,6 +227,16 @@ fn valid_paths(expr: &tipb::Expr) -> Result<()> {
     Ok(())
 }
 
+fn unquote_string(s: &str) -> Result<String> {
```
skyzh, line 230: Why do we add a dedicated function to do this? Can’t we reuse the original logic?

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -204,9 +204,10 @@ fn quote(bytes: BytesRef) -> Result<Option<Bytes>> {
 
 #[rpn_fn(nullable)]
 #[inline]
-fn json_unquote(arg: Option<JsonRef>) -> Result<Option<Bytes>> {
+fn json_unquote(arg: Option<BytesRef>) -> Result<Option<Bytes>> {
```
skyzh, line 207: You may use non-nullable variant for this. Please remove nullable in rpn_fn macro and remove Option in both args and result, so that copr framework could automatically handle the case for null internally with better efficiency.

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -226,6 +227,16 @@ fn valid_paths(expr: &tipb::Expr) -> Result<()> {
     Ok(())
 }
 
+fn unquote_string(s: &str) -> Result<String> {
+    let first_char = s.chars().next();
+    let last_char = s.chars().nth(s.len() - 1);
```
skyzh, line 232: This seems to be not correct. len returns how many bytes are there in str, while chars is an iterator on UTF-8 characters. Please also considering adding test cases on CJK characters. After all, I suggest reusing previous logic instead of extracting quotes by ourselves.

components/tidb_query_expr/src/impl_json.rs
guo-shaoge: The original `unquote_json` doesn't remove first and last double quote, which is not compatible with MySQL.

    mysql> select json_unquote('"abc"');
    +-----------------------+
    | json_unquote('"abc"') |
    +-----------------------+
    | abc                   |
    +-----------------------+

So I add this one to remove.

components/tidb_query_expr/src/impl_json.rs
guo-shaoge: Yeah. I didn't notice the sematics of `len()`.

components/tidb_query_expr/src/impl_json.rs
guo-shaoge: Looks like all others functions in this file all use `nullable`. Maybe we open a new PR to fix all of them.

skyzh: As far as I could see, most functions in this file have non-nullable support, except for those accepting ScalarVlaueRef as parameter, which do not support non-nullable. Therefore, I suggest make this change in this PR.

skyzh:
Also I found this logic might be inconsistent with MySQL if the JSON is invalid. Consider this case:

```
SELECT  json_unquote('""a""');
```

where `""a""` is an invalid JSON object, and `json_unquote` might yield `a` after all. I think this might be a bug that should be fixed in both TiDB and TiKV in the future.

components/tidb_query_datatype/src/codec/mysql/json/mod.rs
```diff
@@ -74,7 +74,7 @@ mod json_merge;
 mod json_modify;
 mod json_remove;
 mod json_type;
-mod json_unquote;
+pub mod json_unquote;
```
Xuanwo, line 77: Why do we need to make `json_unquote` pub here?

components/tidb_query_expr/src/impl_json.rs
Xuanwo: I agreed with @skyzh.

components/tidb_query_datatype/src/codec/mysql/json/mod.rs
guo-shaoge: So we can use json_unquote::unquote_string in tidb_query_expr.

components/tidb_query_expr/src/impl_json.rs
guo-shaoge: done~

skyzh (approved):

Xuanwo (approved):

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -202,12 +202,11 @@ fn quote(bytes: BytesRef) -> Result<Option<Bytes>> {
     Ok(Some(result))
 }
 
-#[rpn_fn(nullable)]
+#[rpn_fn()]
```
Xuanwo, line 205:
```suggestion
#[rpn_fn]
```

guo-shaoge: done

PR #10180, opened 2021-05-17
copr: support enum type for in function

wshwsh12:
Signed-off-by: wshwsh12 <793703860@qq.com>

Issue Number: close https://github.com/pingcap/tidb/issues/24561 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:
Support extracting bytes/int from Enum type for IN function.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 
- No release note.


skyzh:
I wonder why we should handle `Enum` specifically in this case. We could already automatically cast enum to int, and therefore it is not necessary to implement enum variant for functions like `in`, `cast`, etc. This is something that should be handled in codegen.

So the first thing I would like to know is that, why can't we directly call `in` function with enum as input?

skyzh:
Okay, I got the point. When calling `extract`, the value is still some kind of raw value, which is not covered by what we have done before. I think the current approach is good enough.

By the way, why not directly compare int value of enum, which should be a lot more efficient?

wshwsh12:
Because tidb will push IfStringSig to TiKV..

skyzh:
Got it. This patch looks reasonable to me.

Xuanwo:
We can't implement Extract for Enum? It looks so wired for me to handle extra case in this trait.

skyzh:
> We can't implement Extract for Enum? It looks so wired for me to handle extra case in this trait.

Then we may need an extra signature for push down, and this adds extra engineering effort... But I can't come up with a better solution in the current framework.

skyzh (approved):
LGTM. Hope there are no more places where we need to handle the special cases for enum 😂😂😂

skyzh:
If it is possible to go a step further and write a new function for `in_enum`, I also suggest first mapping those strings to integer in enum and then compare enum by their integer values. This should be a lot more efficient.

Xuanwo (approved):
Looks not good to me, but there is no better choices.

PR #10157, opened 2021-05-12
*: add memory trace v2

5kbpers:
### What problem does this PR solve?

Problem Summary:

Add memory trace.

### What is changed and how it works?

What's Changed:

Base on https://github.com/tikv/tikv/pull/9019.
Make some changes to introduce a global memory counter in every traced module for implementing rate limiting more efficiently.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 
* N/A

Issue #10156, opened 2021-05-12
Implement tikv memory trace


5kbpers:
The tracking issue for https://github.com/tikv/rfcs/pull/59

Tasks:
* Trace memory usage
    - [ ] Batch system routers, record alive memory and leaked memory
    - [ ] Raftstore, include Fsms and PollerContext in store and apply.
    - [ ] Coprocessor, include intermediate data during aggregating and pending data in gRPC
    - [ ] CDC, include old value cache, pending data in the task channel, pending data in gRPC
    - [ ] BR, include generating SST
* Memory profile
    - [] Implement an interface to generate a memory profile

PR #10185, opened 2021-05-17
Add memory trace macros

5kbpers:

Problem Summary:
Introduce macros for tracing memory usages in TiKV. 
Separated from https://github.com/tikv/tikv/pull/10157
More details https://github.com/tikv/tikv/issues/10156

What's Changed:
* proc macro `MemoryTraceHelper` for implementing tracing struct
* macro `mem_trace` for defining the hierarchy of memory usage for a module

### Check List 

Tests 

- Unit test
- Integration test

### Release note 
* Add memory trace macros for memory trace


hicqu (approved):

components/memory_trace_macros/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+
+//! Procedural macros used for memory trace.
+
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DeriveInput, Fields};
+
+// MemoryTraceHelper adds two methods `reset` and `sum` to derived struct.
```
BusyJay, line 11: How about writing them as doc string?

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,200 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::num::NonZeroU64;
```
BusyJay, line 3: How about adding some comment explaining how it works?

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,216 @@
...
+        self.parent = Some(Arc::downgrade(parent));
+    }
+}
+
+impl MemoryTrace for MemoryTraceNode {
+    fn trace(&self, event: TraceEvent) {
+        match event {
+            TraceEvent::Add(val) => {
+                self.trace.fetch_add(val, Ordering::Relaxed);
+                self.parent_trace(event);
```
BusyJay, line 135: Tracing parent will make this operation more expansive. Every trace contains 3 atomic operations per node. Why can't it be updated during access?

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+    };
+    ($name: ident, [$($child:tt),+]) => {
+        {
+            use tikv_alloc::trace::MemoryTrace;
+
+            let mut node = mem_trace!($name);
+            $(
+                let mut child = mem_trace!($child);
+                unsafe {
```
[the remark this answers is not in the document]
5kbpers: We need to get a mutable reference of parent node to append child.

components/tikv_alloc/src/trace.rs
5kbpers: I think if we would use it for rate limiting, then it could be accessed very frequently. And surely we need to take care of the frequency of memory trace.

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,216 @@
...
+                    std::sync::Arc::get_mut_unchecked(&mut child).set_parent(&(node.clone() as Arc<dyn MemoryTrace + Send + Sync>));
+                    std::sync::Arc::get_mut_unchecked(&mut node).add_sub_trace(child.id, child);
+                }
+            )*
+            node
+        }
+    };
+    (($name: ident, [$($child:tt),+])) => {
```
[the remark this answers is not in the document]
5kbpers: To match `(b, [c])` in `mem_trace!(a, (b, [c]))`

components/tikv_alloc/src/trace.rs
BusyJay: /cc @Connor1996 How often can rate limiting be?

Tracing parent is not just expansive, it's also a waste. Not every updates need to be observed immediately. For example, if two queries allocate 1000 times and each allocates 10K. With tracing parent, the metrics will be updated 1000 times. But if update on read, then only read happen before every allocate will trigger calculation, which is as worse as tracing parent, otherwise, it performs better. If there is no read requests at all, then no calculation is necessary.

hicqu: I agree. We can sum it when necessary. So only atomics on leaf nodes needs to be updated.

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,220 @@
...
+pub enum Id {
+    Name(&'static str),
+    Number(u64),
+}
+
+impl Id {
+    pub fn readable_name(&self) -> String {
+        match self {
+            Id::Name(s) => {
+                let mut s = s.replace('_', "");
```
hicqu, line 39: Seems `_` is replaced, it conflicts with `split`? BTW it's better to add a test case for it.

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn snapshot(&self, parent: &mut MemoryTraceSnapshot) {
+        let mut snap = MemoryTraceSnapshot {
+            id: self.id,
+            trace: self.trace.load(Ordering::Relaxed),
+            children: vec![],
+        };
+        for child in self.children.values() {
+            child.snapshot(&mut snap);
+        }
+        parent.children.push(snap);
```
hicqu, line 138: Is it better to handle the root node specially?

5kbpers: Changed the parameter to an `Option`

hicqu: If it's `None`, how to return the snapshot?

5kbpers: 😬 Just was still sleepy, now I changed it to return a snapshot.

hicqu:
LGTM.

components/tikv_alloc/src/trace.rs
```diff
@@ -0,0 +1,220 @@
...
+        self.children.get(&id).cloned().unwrap()
+    }
+
+    fn add_sub_trace(&mut self, id: Id, trace: Arc<dyn MemoryTrace + Send + Sync>) {
+        self.children.insert(id, trace);
+    }
+
+    fn sum(&self) -> usize {
+        let sum: usize = self.children.values().map(|c| c.sum()).sum();
+        sum + self.trace.load(Ordering::Relaxed)
```
BusyJay, line 142: Perhaps we need a cache in the future.

BusyJay (approved):

components/tikv_alloc/src/trace.rs
5kbpers: Added a comment here.

Issue #10166, opened 2021-05-13
General RPC timeouts


sticnarf:
## Development Task

Except tidb_query DAG requests, which have a fixed 60s timeout, other TiKV RPCs don't have any timeout when they are executed in TiKV.

However, a typical TiKV client (like TiDB) should have timeouts for the RPC. When the the RPC expires at the client side, it will send a new request to retry, but the old expired request will continue executing on TiKV. This can lead to a vicious circle.

So I propose add a timeout to TiKV RPC's `Context`. All execution paths in TiKV should try to cancel an on-going request when it expires to try to waste less resource.

Related to https://github.com/tikv/tikv/issues/9113

cc @youjiali1995

PR #10188, opened 2021-05-17
storage: check deadline before executing scheduler command

sticnarf:

Part of #10166 

When `max_execution_duration_ms` in `Context` is set, we check whether the deadline is exceeded before acquiring the latches. So a scheduler command that waits in the scheduler for too long will not execute to waste resource.

### Check List 

Tests 

- Unit test

### Release note 

- Check deadline before executing scheduler command

### DIFF
--- a/src/storage/errors.rs
+++ b/src/storage/errors.rs
@@ -50,6 +50,9 @@ pub enum ErrorInner {
 
     #[error("ttl is not enabled, but get put request with ttl")]
     TTLNotEnabled,
+
+    #[error("deadline exceeded")]
+    DeadlineExceeded,
 }
 
 /// Errors for storage module. Wrapper type of `ErrorInner`.
@@ -85,6 +88,7 @@ impl ErrorCodeExt for Error {
             ErrorInner::KeyTooLarge { .. } => error_code::storage::KEY_TOO_LARGE,
             ErrorInner::InvalidCf(_) => error_code::storage::INVALID_CF,
             ErrorInner::TTLNotEnabled => error_code::storage::TTL_NOT_ENABLED,
+            ErrorInner::DeadlineExceeded => error_code::storage::DEADLINE_EXCEEDED,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/storage/txn/scheduler.rs
```diff
@@ -793,6 +835,17 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
             }
         }
     }
+
+    /// Returns `true` if the task has expired and needs to be dropped.
+    #[inline]
+    fn check_task_deadline(&self, task: &Task) -> bool {
```
ekexium, line 841:
```suggestion
    fn check_task_deadline_exceeded(&self, task: &Task) -> bool {
```

ekexium (commented):
/lgtm

src/storage/errors.rs
```diff
@@ -197,6 +208,11 @@ pub fn extract_region_error<T>(res: &Result<T>) -> Option<errorpb::Error> {
             err.set_message("TiKV is Closing".to_string());
             Some(err)
         }
+        Err(Error(box ErrorInner::DeadlineExceeded)) => {
```
youjiali1995, line 211: It's possible the client receives the error and it's a region error. TiDB will invalidate the cache and backoff. I think we should distinguish it.

youjiali1995: Never mind. TiDB can compare the error message for now.

youjiali1995 (approved):

src/storage/errors.rs
sticnarf: I think this PR does not affect current TiDB now because the field in `Context` is not set.

I want to change the TiDB handling of this later. The simplest handling is make client timeout a bit shorter than TiKV's timeout, then the timeout error will never be received by TiDB. The timeout in TiKV just prevents waste of resource.

youjiali1995: Agree with it.

PR #10189, opened 2021-05-17
coprocessor: check deadline while running requests

### LINKED ISSUE #10172: Limit the concurrency of analyze/checksum like DAG does
## Bug Report

TiKV coprocessor doesn’t limit the concurrency of analyze-table/index/checksum, so the request duration is proportional to the concurrent request count due to the yatp which results in endless query timeout and retry until backoff times out.

In our internal test, we use lightning to load data and lightning sets the `tidb_distsql_scan_concurrency` to 100. The request duration is far from the handle duration due to the scheduling of yatp.
image
The running tasks keep increasing due to the retry mechanism in the client and consumes all CPU of the read pool.
image

## Solution

Limit the concurrency of analyze/checksum like DAG does and cancel stale requests https://github.com/tikv/tikv/issues/10166.

sticnarf:

Issue Number: close #10172

A deadline checker interceptor is added. Everytime the executor polls, it checks whether the deadline is exceeded. So we can make sure all running coprocessor requests including analyze and checksum can check the deadline.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

-  Check deadline while running analyze or checksum requests.


src/coprocessor/interceptors/deadline.rs
```diff
@@ -0,0 +1,68 @@
...
+use pin_project::pin_project;
+use std::future::Future;
+use std::marker::PhantomData;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tikv_util::deadline::{Deadline, DeadlineError};
+
+/// Checks the deadline before every poll of the future. If the deadline is exceeded,
+/// `DeadlineError` is returned.
+pub fn check_deadline<'a, F: Future + 'a>(
```
zhongzc, line 12: I don't think we need the `'a`.

src/coprocessor/endpoint.rs
```diff
@@ -416,11 +418,13 @@ impl<E: Engine> Endpoint<E> {
 
         tracker.on_begin_all_items();
 
-        let handle_request_future = track(handler.handle_request(), &mut tracker);
+        let deadline = tracker.req_ctx.deadline;
+        let handle_request_future =
+            check_deadline(track(handler.handle_request(), &mut tracker), deadline);
         let result = if let Some(semaphore) = &semaphore {
-            limit_concurrency(handle_request_future, semaphore, LIGHT_TASK_THRESHOLD).await
+            limit_concurrency(handle_request_future, semaphore, LIGHT_TASK_THRESHOLD).await?
```
youjiali1995, line 425: Seems the logic is different from before.

sticnarf: Do you mean the `?`?
As `ConcurrencyLimiter` does not return an error, the `?` handles the result returned by `DeadlineCheck`. So it does not look different to me

src/coprocessor/interceptors/deadline.rs
sticnarf: Ah, right...

zhongzc (approved):

src/coprocessor/endpoint.rs
sticnarf: I map the error inside `handle_unary_request_impl`

src/coprocessor/endpoint.rs
```diff
@@ -416,12 +415,15 @@ impl<E: Engine> Endpoint<E> {
...
-        let result = if let Some(semaphore) = &semaphore {
+        let deadline = tracker.req_ctx.deadline;
+        let handle_request_future =
+            check_deadline(track(handler.handle_request(), &mut tracker), deadline);
+        let deadline_res = if let Some(semaphore) = &semaphore {
             limit_concurrency(handle_request_future, semaphore, LIGHT_TASK_THRESHOLD).await
         } else {
             handle_request_future.await
         };
+        let result = deadline_res.map_err(Error::from).and_then(|res| res);
```
youjiali1995, line 426: Why `and_then`?

sticnarf: If `deadline_res` is `Ok`, deadline is not exceeded, the result should just be the inner result.
If `deadline_res` is `Err`, the inner result is dropped and the result is the deadline error.

youjiali1995 (approved):

tests/failpoints/cases/test_coprocessor.rs
```diff
@@ -58,14 +58,16 @@ fn test_deadline_3() {
     fail::cfg("kv_cursor_seek", "sleep(2000)").unwrap();
     fail::cfg("copr_batch_initial_size", "return(1)").unwrap();
     let cop_resp = handle_request(&endpoint, req);
+    println!("{:?}", cop_resp);
```
Connor1996, line 61: should remove this

sticnarf: Sorry, I will remove it in my recent PR.

PR #10192, opened 2021-05-18
use bytes for raft Entry.data

hicqu:

When raftstore threads fetch entries from `EntryCache`, entries are cloned and then sent to apply threads. This involves a deep memory copy for `Entry.data`, which utilizes CPU and lots of memory.

Issue Number: close https://github.com/pingcap/tidb/issues/22964

Use copy-on-write `bytes` for `Entry.data`, so that deep momory copy for it will be replaced by a shallow copy.

### Related changes

### Check List 

Tests 

- No code

### Release note 
* No release note

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -1733,9 +1733,10 @@ where
             if !self.is_leader() {
                 fail_point!("raft_before_follower_send");
             }
-            for vec_msg in ready.take_messages() {
-                self.send(&mut ctx.trans, vec_msg, &mut ctx.raft_metrics.send_message);
-            }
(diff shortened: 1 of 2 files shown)


components/raftstore/src/store/peer.rs
```diff
@@ -1733,9 +1733,10 @@ where
             if !self.is_leader() {
                 fail_point!("raft_before_follower_send");
             }
-            for vec_msg in ready.take_messages() {
-                self.send(&mut ctx.trans, vec_msg, &mut ctx.raft_metrics.send_message);
-            }
+            let msgs = ready.take_persisted_messages();
```
gengliqi, line 1736: The `persisted_messages` should be sent after persisting data

PR #10194, opened 2021-05-18
support raft engine in tikv-ctl

hicqu:

Support raft engine in tikv-ctl. It contains such changes:
#### Replace `--db` and `--raft` with `--config` and `--data-dir`
We have at least 2 reasons to do this:
* after raft engine is introduced, it's necessary to specify more different options for tikv-ctl.
* some users can custom `wal-dir` for rocksdb or raftdb, in which case `--db` and `--raft` can't work.

Use `--config` and `--data-dir` can resolve both 2 problems.

**Error log for deprecated options**:
```
$ ./tikv-ctl --db ./data/tikv-60003/db raft region -r 2
error: Invalid value for '--db <db>': DEPRECATED!!! Use --data-dir and --config instead
```
I think it's clear enough to lead users use the new options.

#### Use slog to replace vlog
vlog is a nice logger. However the reason to replace it is we have already used `slog` in TiKV. It's better for `tikv-ctl` to use the same logger. And, many outputs in `tikv-ctl.rs` are not **log** exactly, they are just **result**. So use `println!` is better than `info!` or other somethings. 

### Check List 

Tests 

### Release note 
* No release note


cmd/tikv-ctl/src/main.rs
```diff
@@ -1118,6 +1101,42 @@ fn main() {
...
+                .help("TiKV data-dir, check <deploy-dir>/scripts/run.sh to get it"),
+        )
+        .arg(
+            Arg::with_name("skip-paranoid-checks")
+                .required(false)
+                .long("skip-paranoid-checks")
+                .takes_value(false)
+                .help("Skip paranoid checks when open rocksdb"),
+        )
+        // Deprecated arguments.
```
hunterlxt, line 1124: Is this just to remind users with old habits?

hunterlxt (approved):

cmd/tikv-ctl/src/main.rs
hicqu: Yes.

cmd/tikv-ctl/src/main.rs
```diff
@@ -165,15 +163,15 @@ fn new_debug_client(host: &str, mgr: Arc<SecurityManager>) -> DebugClient {
 trait DebugExecutor {
     fn dump_value(&self, cf: &str, key: Vec<u8>) {
         let value = self.get_value_by_key(cf, key);
-        v1!("value: {}", escape(&value));
+        println!("value: {}", escape(&value));
```
yiwu-arbug, line 166: what's the difference between `v1!` and `println!`?

yiwu-arbug (approved):
LGTM

cmd/tikv-ctl/src/main.rs
hicqu: `v1!` is from crate `vlog`. However as the PR descripted, here outputs are more like **result** instead of **log**, so I replace it with `println!`. And it's better to unify all logs to `slog` instead of keeping both `vlog` and `slog`.

BusyJay:
This PR changes the way tikv-ctl works a lot, but docs are not updated.

PR #10198, opened 2021-05-18
*: support parsing ReadableSize in scientific notations

tabokie:

Problem Summary:

IO data rate commonly uses decimal base (KB, MB, GB instead of KiB, MiB, GiB).

What's Changed:

~Legacy codes do not differentiate between decimal base and binary base (MB is treated as MiB). To avoid breaking backward compatibility, this PR introduces a separate `ReadableRate` config type to support parsing IO rate in decimal unit.~

Support parsing ReadableSize in scientific notations, in order to minimize inconvenience of using decimal unit to specify data rate.

### Related changes

Prepping for #9197

### Check List 

Tests

- Unit test

### Release note 

- No release note


components/tikv_util/src/config.rs
```diff
@@ -34,14 +34,18 @@ pub enum ConfigError {
...
 pub const MB: u64 = KB * DATA_MAGNITUDE;
 pub const GB: u64 = MB * DATA_MAGNITUDE;
-
-// Make sure it will not overflow.
-const TB: u64 = (GB as u64) * (DATA_MAGNITUDE as u64);
-const PB: u64 = (TB as u64) * (DATA_MAGNITUDE as u64);
+pub const TB: u64 = GB * DATA_MAGNITUDE;
+pub const PB: u64 = TB * DATA_MAGNITUDE;
+const BI_DATA_MAGNITUDE: u64 = 1024;
+pub const KI_B: u64 = UNIT * BI_DATA_MAGNITUDE;
```
BusyJay, line 44: Should be KIB, it's short for unit kibibyte.

src/config.rs
```diff
@@ -906,6 +906,7 @@ pub struct DbConfig {
     pub info_log_keep_log_file_num: u64,
     #[config(skip)]
     pub info_log_dir: String,
+    #[serde(with = "tikv_util::config::readable_size_in_decimal_base_serde")]
     pub rate_bytes_per_sec: ReadableSize,
```
BusyJay, line 910: Why it has to be 10 based?

components/tikv_util/src/config.rs
```diff
@@ -271,6 +276,114 @@ impl<'de> Deserialize<'de> for ReadableSize {
     }
 }
 
+pub mod readable_size_in_decimal_base_serde {
```
BusyJay, line 279: Can the code be merged with 2 based unit?

src/config.rs
tabokie: It's a common practice to measure data transfer rate with 10 base, e.g. the data rate unit in grafana uses (and can only uses) `MB/s, GB/s`.

components/tikv_util/src/config.rs
tabokie: I can refactor a bit, but they can't be merged as one implementation. That would means changing the default interpretation for `MB/GB`, which leads to 1) old configuration item will be interpreted differently in newer binary 2) in the future user has to explicitly use `MiB/GiB`

src/config.rs
BusyJay: But the EBS's throughput is calculated using MiB https://aws.amazon.com/ebs/features/
> ***Volume throughput is calculated as MB = 1024^2 bytes

Azure also uses the same unit https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types
> where MBps = 10^6 Bytes per second

Grafana can use 2 based unit too as long as "Data (ICE) / bytes" is selected.

BusyJay: GCE doesn't explicitly specify what unit it's using, but from calculation in the page https://cloud.google.com/compute/docs/disks/performance, it should also be MiB.
> 0.25 GB per second = 250 MB per second => 250 MB per second / 3.35

tabokie: Thanks for bringing up the info, but it seems both Azure and GCP uses 10 base unit? And for AWS, I think the footnotes are used to clarify special cases (they only use MiB for Max Throughput/Volume), other MBs are by default interpreted as 10 base.

BusyJay: > A case in point would be the pricing page

If you are saying the page https://aws.amazon.com/ebs/pricing/, there is also a note in the page:
> *on this page GB = 1024^3 bytes

"GB" or "MB/s" doesn't necessary mean it's 10 based.

tabokie: Oh I didn't see that... But it doesn't change two facts: 1) MB/GB is by default (maybe even by legal standard) interpreted as 10 base (we wouldn't need footnotes for standard interpretation) 2) Data transfer rate (which is priced in MB/s) is 10 based.

BusyJay: Can you show references about two "facts"? MB/GB can have different calculations in different context, I don't think there is a "default".

tabokie: I think right in this context (AWS/GCP/Azure), when document writers use `MB/GB` without footnotes, their underlying meaning is the default meaning. Or Wikipedia?

tabokie: And for the (2) fact, I think it's self-proving to assume they interpret `MB/s` as `10e6 Bytes/s`, since they don't use footnotes to specify otherwise. But I guess I can't really prove that unless consulting their tech support.

BusyJay: All your claim is based on the assumption "MB/GB is by default (maybe even by legal standard) interpreted as 10 base". However, even the page you quoted also mention that:
> By the end of 2009, the IEC Standard had been adopted by the IEEE, EU, ISO and NIST. Nevertheless, the term megabyte continues to be widely used with different meanings:
>
> Base 10
> ...
> Base 2
> ...
> Mixed
> ...

Connor1996: Seems AWS and Azure use different standards😢. But I agree with Jay, there is no default meaning for MB/s. Both are correct. It depends on the context.

tabokie: @BusyJay I don't think there is any ambiguity in a well-maintained, official, document's usage of `MB/GB`, if that's the case why don't we see AWS/GCP spelling that out in their documents? I admit in casual cases there can be mixed usage, that's why I didn't change the legacy implementation for interpreting `MB/GB`.

and @Connor1996 AWS and Azure both use 10 base for data rate, and AWS uses 2 base for *some* usage of `MB`, and all such usages are well notified.

Let's not forget the purpose of this PR, I'm trying to make tikv's configuration of disk IO rate in alignment with standard (more standard if you must insist) way of measuring. The most pressing issue is Grafana default uses `MB/s aka 10e6 B/s` as unit, even for Data(ICE) proposed here, the display unit is `MiB/s`. And even if we decide to change all occurrences of `Data Rate` to `Data (IDE) Rate` in grafana, we won't be able to change the cloud platforms' own cloud monitoring measuring.

tabokie: To add to the argument, I think data rate brings with itself the context to use `MB` as `1e6B`, as shown in [Data rate units](https://en.wikipedia.org/wiki/Data-rate_units), and any devices/services with a BW spec.

BusyJay: I object to interpret MB/GB and other units as 10 based for following reasons:
1. The unit itself is ambiguous, parse the same value with different rules based on which field is using will make things worse.
2. We have already parse MB/GB as 2 based in the past, there is no good reason to break the compatibility.

I think there can be two ways to solve the align problems:
1. Do nothing. It's quite easy to input 150MBps as 150,000,000Bps.
2. Accept other form like 150MBps as 150e6Bps. "150e6" is already accepted by Rust parse, all we need to do is also accept "e" as part of numbers.

tabokie: Discussed offline, going to use a separate data rate unit to avoid confusion.

yiwu-arbug (commented):
I feel it confusing when MBps is 10 based and MB is 2 based. We need to keep MB being 2 based for compatibility. If using MiB as 10 based is confusing, how about something like `MB_10`?

tabokie:
> I feel it confusing when MBps is 10 based and MB is 2 based. We need to keep MB being 2 based for compatibility. If using MiB as 10 based is confusing, how about something like `MB_10`?

We already know that: AWS uses MB/s as an alias for MiB/s, Azure ultra disk explicitly uses MBps for MB/s, and GCP implicitly uses MB/s for MB/s. So there is not consensus in the unit to describe cloud disk IO BW.

Based on this information, seems there is no ideal solution. Both `MiBps` and `MB_10` is not popularly known unit. And `MBps` could be confusing. And plain integer seems inconsistent with other configuration items.

cmd/tikv-ctl/src/main.rs
```diff
@@ -2274,21 +2274,21 @@ fn from_hex(key: &str) -> Result<Vec<u8>, hex::FromHexError> {
...
+    const MIB: u64 = 1024 * 1024;
+    if bytes < MIB {
         return format!("{} B", bytes);
     }
-    let mb = if bytes % GB == 0 {
+    let mb = if bytes % GIB == 0 {
         String::from("")
     } else {
-        format!("{:.3} MB", (bytes % GB) as f64 / MB as f64)
+        format!("{:.3} MiB", (bytes % GIB) as f64 / MIB as f64)
```
BusyJay, line 2285: Why is there a space? We usually write like "12MiB" instead of "12 MiB".

components/tikv_util/src/config.rs
```diff
@@ -34,14 +34,18 @@ pub enum ConfigError {
...
 const UNIT: u64 = 1;
-const DATA_MAGNITUDE: u64 = 1024;
-pub const KB: u64 = UNIT * DATA_MAGNITUDE;
-pub const MB: u64 = KB * DATA_MAGNITUDE;
-pub const GB: u64 = MB * DATA_MAGNITUDE;
-
-// Make sure it will not overflow.
-const TB: u64 = (GB as u64) * (DATA_MAGNITUDE as u64);
-const PB: u64 = (TB as u64) * (DATA_MAGNITUDE as u64);
+const DECIMAL_DATA_MAGNITUDE: u64 = 1000;
```
BusyJay, line 37: Is it being used? How about removing them to ensure no existing code is accidentally referencing the old name?

BusyJay (commented):
Rest LGTM

Connor1996 (commented):
So we parse both `MB` and `MiB` as the binary base. If someone want to pass decimal base, he should use `B` directly?

components/tikv_util/src/config.rs
tabokie: Good idea.

BusyJay (commented):
/lgtm

Connor1996 (approved):
LGTM

PR #10213, opened 2021-05-19
log: ensure panic output is flushed to the log

### LINKED ISSUE #8998: Panic log may be lost when using async logger

`187774ed9433af80d117b6576179515b2e24ab6a`

### Steps to reproduce

Start a runner on thread pool to panic as https://github.com/zhongzc/tikv/commit/bcda8dca653988169b96d3cb45e254594520019e.

### What did you expect?

The panic log can be guaranteed to see.

### What did happened?

I couldn't get the `boom!` panic output.

Connor1996:

Problem Summary: The panic message is still possibly missing from the log. The previous fix is not elegant https://github.com/tikv/tikv/pull/9955, e.g. for debug build, the wait duration is still too small.

Fix #8998

What's Changed: 
This PR solves the issue at root. The root cause is the panic log is not processed yet in the async log worker before `_exit` is called. So using [async guard](https://docs.rs/slog-async/2.6.0/slog_async/struct.AsyncGuard.html) to wait for the finish of the async log worker. 

### Check List 

Tests 

- Manual test (Use debug build)
- Unit test

### Release note 

- Ensure panic output is flushed to the log


components/tikv_util/src/lib.rs
```diff
@@ -487,6 +485,7 @@ pub fn set_panic_hook(panic_abort: bool, data_dir: &str) {
                 // static variables of RocksDB, which may cause other threads encounter
                 // pure virtual method call. So calling libc::_exit() instead to skip the
                 // cleanup process.
+                libc::fflush(std::ptr::null_mut());
```
BusyJay, line 488: Why? Cache are usually line buffered, no need to flush again.

components/tikv_util/src/logger/mod.rs
```diff
@@ -75,43 +75,57 @@ where
...
+    set_global_logger(level, init_stdlog, logger, guard)
+}
+
+// All Drains are reference-counted by every Logger that uses them. Async drain
+// runs a worker thread and sends a termination (and flushing) message only when
+// being dropped. Because of that it's actually quite easy to have a left-over
+// reference to a Async drain, when terminating: especially on panics or similar
+// unwinding event. Typically it's caused be a leftover reference like Logger in
+// thread-local variable, global variable, or a thread that is not being joined
+// on. So use AsyncGuard to send a flush and termination message to a Async
```
BusyJay, line 111: When it's sent?

components/tikv_util/src/lib.rs
yiwu-arbug, line 488: shall we flush also for the abort case?

gregwebs:
What is your test case?

components/tikv_util/src/lib.rs
Connor1996: I don't get what you mean. Do it in case the log is not flushed in time, see https://www.ibm.com/docs/en/zos/2.1.0?topic=functions-exit-end-process-bypass-cleanup

components/tikv_util/src/logger/mod.rs
Connor1996: When the async guard is dropped. In panic hook, we will replace the old global logger with a new terminal logger. At that time, the old async guard will be taken and dropped.

components/tikv_util/src/lib.rs
Connor1996: Seems yes. https://man7.org/linux/man-pages/man3/abort.3.html, since glibc 2.27, abort doesn't flush anymore.

Connor1996:
> What is your test case?

To reproduce that, it's highly related to if some thread-local variable held that reference. Not sure if a simple test case would reproduce it. Will have a try

Connor1996:
> What is your test case?

Anyway, if I run a debug-build tikv-server, it can be reproduced steadily in my cluster.

components/tikv_util/src/lib.rs
BusyJay: If logs are written to terminal, they are flush line by line, so no need to call `fflush`; if logs are written to file, content will be written to file immediately at write.

Connor1996: Oh, I see. The flush is called explicitly in `log` of `TikvFormat`

Connor1996: Seems it's indeed unnecessary. But maybe it's better to do it to make it close to the behavior of `exit()`?

BusyJay: > make it close to the behavior of `exit()`

Don't see any point in doing this. Better keep code clean.

Connor1996:
It is reproduced steadily by adding a test case, PTAL @gregwebs @BusyJay @yiwu-arbug

components/tikv_util/src/lib.rs
```diff
@@ -534,12 +532,85 @@ pub fn build_on_master_branch() -> bool {
...
+                std::process::exit(0);
+            }
+            Err(e) => {
+                return Err(format!("Fork failed: {}", e));
+            }
+        }
+    }
+
+    #[test]
+    fn test_panic_hook() {
```
BusyJay, line 583: Better make it only run on Unix like system.

components/tikv_util/Cargo.toml
```diff
@@ -77,3 +77,5 @@ regex = "1.0"
 tempfile = "3.0"
 toml = "0.5"
 utime = "0.2"
+gag = "0.1.10"
+nix = "0.11"
```
BusyJay, line 81: Should be unix-only dependency.

yiwu-arbug (approved):
LGTM

gregwebs:
In my testing an assert statement will not be logged. I tested this PR and master.

gregwebs:
In my testing of a panic I have been encountering this change does log one additional log statement that master does not.

gregwebs (approved):
This worked in my manual testing

Connor1996:
PTAL again @BusyJay

yiwu-arbug:
@Connor1996 there's lint error needs to be fixed.

Connor1996:
@hi-rustin  I push a new commit to my branch, but seems the commit is not shown up here. Have no idea if it is a bug of github or related to the CI bot?

tabokie:
@Connor1996 this will close #8998?

Connor1996:
> @Connor1996 this will close #8998?

Yes

hi-rustin:
> @hi-rustin I push a new commit to my branch, but seems the commit is not shown up here. Have no idea if it is a bug of github or related to the CI bot?

I think this is a GitHub bug that I've encountered on other repositories.

See also: https://github.com/diesel-rs/diesel/pull/2740#issuecomment-830836564

hi-rustin:
Now it looks like you only have 9 commits?

Connor1996:
> Now it looks like you only have 9 commits?

It's okay after reopen and push new commits

rleungx:
Does this PR need to be cherry-picked? @Connor1996

PR #10215, opened 2021-05-19
resolved_ts: reset `resolved_ts` after region merge

NingLin-P:

Issue Number: close #xxx 

Problem Summary:

`resolved_ts` is calculated by all locks on the region, after region merge, the target region's key range is extended hence there could be more locks, so the `resolved_ts` need to be re-calculate, but before re-calculate finished, the `resolved_ts` should not be used directly, since it could wrong.

For example, the source region has a lock at `ts_1` and the client getting a `ts_2` prepare to commit, and the target region has no lock and updates its `resolved_ts` to `ts3`,  then source region merged into target region. Now the `resolved_ts` of the target region is wrong to be `ts3` since it has a lock at `ts_1` which will be committed at `ts_2` and break definition of  `resolved_ts`: *ResolvedTS < ∀(unapplied CommitTS)*

What's Changed:

This PR handle region merge for `resolved_ts` by:
- Stop publish `(safe_ts, apply_index)` item until the `resolved_ts` is re-calculate based on the new key range
    - So `(safe_ts, apply_index)` item where `apply_index >= last_merge_index` won't be publish before the lock in new key range is re-scanned and `resolved_ts` is  re-calculate based on those locks
- Reset the `safe_ts` to `min(source_safe_ts, target_safe_ts)`
    - So replica can handle stale read request after region merge and before the `resolved_ts` is re-calculate and synced
- Reject stale `(safe_ts, apply_index)` item where `apply_index <= last_merge_index`
    - So a delay message won't wrongly increase the `resolved_ts`

### Related changes

### Check List 

Tests 

- Integration test

### Release note 
- No release note


5kbpers:
This case is acceptable for CDC since we always assume that CDC needs to subscribe to all regions of a table. After region merge completed, the resolved ts would be reset to `min(source region rts, target region rts)`. So the resolved-ts of the table should be correct. 
cc @overvenus

components/resolved_ts/src/endpoint.rs
```diff
@@ -126,7 +139,37 @@ impl ObserveRegion {
...
+                            | AdminCmdType::RollbackMerge
+                            | AdminCmdType::CommitMerge => {
+                                info!(
+                                    "region met split/merge command, stop tracking since key range changed, wait for re-register";
+                                    "req_type" => ?req_type,
+                                );
+                                // Stop tracking read progress so that `tracked_index` larger than the split/merge command index
+                                // won't be published untill `RegionUpdate` event trigger the region re-register and re-scan the
+                                // new key range
+                                self.resolver.stop_tracking_read_progress();
```
5kbpers, line 163: And resolved ts needs to stop tracking too.

components/resolved_ts/src/endpoint.rs
```diff
@@ -313,12 +356,11 @@ where
             }),
             before_start: None,
             on_error: Some(Box::new(move |observe_id, _region, e| {
-                let error = e.extract_error_header();
                 scheduler_error
-                    .schedule(Task::RegionError {
+                    .schedule(Task::ReRegisterRegion {
```
5kbpers, line 360: Do not re-register region if we met errors like `ReadIndexNotReady`.

components/cdc/src/endpoint.rs
```diff
@@ -1472,7 +1472,7 @@ mod tests {
             txn_extra_op: Arc::new(AtomicCell::new(TxnExtraOp::default())),
             max_ts_sync_status: Arc::new(AtomicU64::new(0)),
             track_ver: TrackVer::new(),
-            read_progress: Arc::new(RegionReadProgress::new(0, 0)),
+            read_progress: Arc::new(RegionReadProgress::new(0, 0, "".to_owned())),
```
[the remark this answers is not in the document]
NingLin-P: It is in test.

components/resolved_ts/src/endpoint.rs
NingLin-P: Yes, but we still need a way to retry though, perhaps we should do the retry in `get_snapshot`, and leave `on_error` as an interface for error that can't handle by retry.

components/resolved_ts/src/endpoint.rs
NingLin-P: Updated, PTAL

components/raftstore/src/store/util.rs
```diff
@@ -927,11 +936,36 @@ pub struct ReadState {
...
+        // Reset target region's `safe_ts`
+        let target_safe_ts = self.read_state.ts;
+        self.read_state.ts = cmp::min(source_safe_ts, target_safe_ts);
+        info!(
+            "reset safe_ts due to merge";
+            "tag" => &self.tag,
+            "source_safe_ts" => source_safe_ts,
+            "target_safe_ts" => target_safe_ts,
+            "safe_ts" => self.read_state.ts,
+        );
```
Yisaer, line 964: In fact, I think the log may not be helpful if we didn't print the region-id here.

components/raftstore/src/store/util.rs
```diff
@@ -961,6 +995,10 @@ impl RegionReadProgressCore {
 
     // Return the `safe_ts` if it is updated
     fn update_safe_ts(&mut self, idx: u64, ts: u64) -> Option<u64> {
+        // Discard stale item with `apply_index` before `last_merge_index`
```
Yisaer, line 998:
```suggestion
        // Discard stale item with `apply_index` before `last_merge_index` in order to prevent the stale item makes the `safe_ts` larger again
```

Yisaer (commented):
Rest LGTM

components/raftstore/src/store/util.rs
NingLin-P: The region id and peer id is contained in `self.tag`

components/raftstore/src/store/util.rs
NingLin-P: It is mentioned in the comment of `last_merge_index`

Yisaer: I think it's better to comment the reason here instead of what we are doing here.

Yisaer (approved):
/lgtm

components/raftstore/src/store/util.rs
NingLin-P: Updated!

BusyJay (approved):
LGTM for the safe_ts part.

PR #10218, opened 2021-05-20
always put engine info log into data dir

hicqu:

If `log-file` isn't specified, info log of rocksdb and raftdb will be print into `stderr`, just like TiKV's log. It makes the output hard to read.

Always print two engines' info log into `*.info` in data directory. It's reasonable because the data directory is always available whatever `log-file` is specified or not.

### Check List 

Tests 

- No code

### Release note 
* No release note.

### DIFF
--- a/components/server/src/setup.rs
+++ b/components/server/src/setup.rs
@@ -65,33 +65,111 @@ fn make_engine_log_path(path: &str, sub_path: &str, filename: &str) -> String {
 
 #[allow(dead_code)]
 pub fn initial_logger(config: &TiKvConfig) {
-    fn build_logger<D>(drainer: D, config: &TiKvConfig)
-    where
-        D: slog::Drain + Send + 'static,
-        <D as slog::Drain>::Err: std::fmt::Display,
-    {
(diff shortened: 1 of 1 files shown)


tabokie (commented):
Why is #10194 depending on this one? Is there a specific use case to not specifying log-file?

I'm just thinking maybe user don't specify log-file purposefully to prevent any log file landing on disk/outside stdout.

hicqu:
> Why is #10194 depending on this one?
tikv-ctl prints logs to stderr. Without the PR many rocksdb logs will be mixed.

 > Is there a specific use case to not specifying log-file?
In quick testing (start tikv-server at console) log-file generally will not be specified.

> user don't specify log-file purposefully to prevent any log file landing on disk/outside stdout
If it's not specified, tikv logs will be print to stderr mixed with rocksdb log, which is very hard to read.

tabokie (approved):
LGTM, I checked this is consistent with config template spec.

yiwu-arbug (approved):

PR #10220, opened 2021-05-20
batch-system,raftstore: add router memory trace

5kbpers:

Problem Summary:
Add router memory trace to detecting potential memory leaking.
Blocked by https://github.com/tikv/tikv/pull/10185

What's Changed:
* Define raftstore memory trace hierachy
* Add `state_cnt` and `alive_cnt` for counting routers state
* Update raftstore memory trace after registering and closing a router.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 
* Add router memory trace for raftstore

### DIFF
--- a/components/batch-system/src/router.rs
+++ b/components/batch-system/src/router.rs
@@ -324,6 +324,7 @@ where
         let alive = self.normals.lock().unwrap().alive_cnt.clone();
         let total = self.state_cnt.load(Ordering::Relaxed);
         let alive = alive.load(Ordering::Relaxed);
+        // 1 represents the control fsm.
         let leak = if total > alive + 1 {
             total - alive - 1
         } else {
--- a/components/batch-system/src/router.rs
+++ b/components/batch-system/src/router.rs
@@ -337,10 +337,10 @@ where
         // and every sender has at least one sender.
         let sender_block_unit = 31;
         RouterTrace {
-            alive: alive * mailbox_unit * 8 / 7 // hashmap uses 7/8 of allocated memory.
-                + state_unit * alive
-                + message_unit * sender_block_unit * alive,
-            leak: state_unit * leak + message_unit * sender_block_unit * leak,
+            alive: (mailbox_unit * 8 / 7 // hashmap uses 7/8 of allocated memory.
+                + state_unit + message_unit * sender_block_unit)
+                * alive,
+            leak: (state_unit + message_unit * sender_block_unit) * leak,
         }
     }
 }
--- a/components/batch-system/src/router.rs
+++ b/components/batch-system/src/router.rs
@@ -1,15 +1,29 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
-use crate::fsm::{Fsm, FsmScheduler};
+use crate::fsm::{Fsm, FsmScheduler, FsmState};
 use crate::mailbox::{BasicMailbox, Mailbox};
 use collections::HashMap;
 use crossbeam::channel::{SendError, TrySendError};
 use std::cell::Cell;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::mem;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use tikv_util::lru::LruCache;
 use tikv_util::Either;
 use tikv_util::{debug, info};
+
+/// A struct that traces the approximate memory usage of router.
+#[derive(Default)]
+pub struct RouterTrace {
+    pub alive: usize,
+    pub leak: usize,
+}
+
+struct NormalMailMap<N: Fsm> {
+    map: HashMap<u64, BasicMailbox<N>>,
+    /// Count of Mailbox that is stored in `map`.
+    alive_cnt: Arc<AtomicUsize>,
+}
 
 enum CheckDoResult<T> {
     NotExist,
@@ -29,7 +43,7 @@ enum CheckDoResult<T> {
 /// missing fsm for specified address. Normal fsm and control fsm can have
 /// different scheduler, but this is not required.
 pub struct Router<N: Fsm, C: Fsm, Ns, Cs> {
-    normals: Arc<Mutex<HashMap<u64, BasicMailbox<N>>>>,
+    normals: Arc<Mutex<NormalMailMap<N>>>,
     caches: Cell<LruCache<u64, BasicMailbox<N>>>,
     pub(super) control_box: BasicMailbox<C>,
     // TODO: These two schedulers should be unified as single one. However
@@ -38,6 +52,8 @@ pub struct Router<N: Fsm, C: Fsm, Ns, Cs> {
     pub(crate) normal_scheduler: Ns,
     control_scheduler: Cs,
 
+    /// Count of Mailbox that is not destroyed.
+    state_cnt: Arc<AtomicUsize>,
     // Indicates the router is shutdown down or not.
     shutdown: Arc<AtomicBool>,
 }
@@ -53,13 +69,18 @@ where
         control_box: BasicMailbox<C>,
         normal_scheduler: Ns,
         control_scheduler: Cs,
+        state_cnt: Arc<AtomicUsize>,
     ) -> Router<N, C, Ns, Cs> {
         Router {
-            normals: Arc::default(),
+            normals: Arc::new(Mutex::new(NormalMailMap {
+                map: HashMap::default(),
+                alive_cnt: Arc::default(),
+            })),
             caches: Cell::new(LruCache::with_capacity_and_sample(1024, 7)),
             control_box,
             normal_scheduler,
             control_scheduler,
+            state_cnt,
             shutdown: Arc::new(AtomicBool::new(false)),
         }
     }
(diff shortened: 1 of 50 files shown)


components/batch-system/src/router.rs
```diff
@@ -38,6 +52,8 @@ pub struct Router<N: Fsm, C: Fsm, Ns, Cs> {
     pub(crate) normal_scheduler: Ns,
     control_scheduler: Cs,
 
+    /// Count of Mailbox that is not destroyed.
+    state_cnt: Arc<AtomicUsize>,
```
hicqu, line 56: What's the different between `state_cnt` and `alive_cnt`?

components/batch-system/src/router.rs
```diff
@@ -132,19 +153,25 @@ where
     /// Register a mailbox with given address.
     pub fn register(&self, addr: u64, mailbox: BasicMailbox<N>) {
         let mut normals = self.normals.lock().unwrap();
-        if let Some(mailbox) = normals.insert(addr, mailbox) {
+        if let Some(mailbox) = normals.map.insert(addr, mailbox) {
             mailbox.close();
         }
+        normals
+            .alive_cnt
+            .store(normals.map.len(), Ordering::Relaxed);
```
hicqu, line 161: Is it possible to call `register` in threads concurrently? If so, `alive_cnt.fetch_add` is better.

components/batch-system/src/router.rs
```diff
@@ -272,14 +299,45 @@ where
         info!("[region {}] shutdown mailbox", addr);
         unsafe { &mut *self.caches.as_ptr() }.remove(&addr);
         let mut mailboxes = self.normals.lock().unwrap();
-        if let Some(mb) = mailboxes.remove(&addr) {
+        if let Some(mb) = mailboxes.map.remove(&addr) {
             mb.close();
         }
+        mailboxes
+            .alive_cnt
+            .store(mailboxes.map.len(), Ordering::Relaxed);
```
hicqu, line 307: As above, I prefer `fetch_sub`.

components/batch-system/src/router.rs
```diff
@@ -272,14 +299,45 @@ where
...
+        let leak = if total > alive + 1 {
+            total - alive - 1
+        } else {
+            0
+        };
+        let mailbox_unit = mem::size_of::<(u64, BasicMailbox<N>)>();
+        let state_unit = mem::size_of::<FsmState<N>>();
+        let message_unit = mem::size_of::<N::Message>();
+        RouterTrace {
+            alive: alive * mailbox_unit * 8 / 7
```
hicqu, line 335: Please comment `8`, `7` and `31`.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3676,6 +3688,22 @@ where
     pub router: BatchRouter<ApplyFsm<EK>, ControlFsm>,
 }
 
+impl<EK> Drop for ApplyRouter<EK>
+where
+    EK: KvEngine,
+{
+    fn drop(&mut self) {
+        RAFTSTORE_MEM_TRACE
```
hicqu, line 3696: Seems it's possible to reuse `update_router_trace`.

components/raftstore/src/store/memory.rs
```diff
@@ -0,0 +1,25 @@
...
+use lazy_static::lazy_static;
+use memory_trace_macros::MemoryTraceHelper;
+use std::sync::Arc;
+use tikv_alloc::{mem_trace, trace::MemoryTraceNode};
+
+lazy_static! {
+    pub static ref RAFTSTORE_MEM_TRACE: Arc<MemoryTraceNode> = mem_trace!(
+        raftstore,
+        [(raft_router, [alive, leak]), (apply_router, [alive, leak])]
+    );
```
hicqu, line 12: Something like this is possible:
```rust
lazy_static! {
    static ref X: HashMap<usize, usize> = hashmap!{1 => 1};
    static ref Y: usize = *X.get(&1).unwrap();
}
```
So I suggest to define all of `MEMTRACE_ROOT`, `MEMTRACE_RAFT_ROUTER_ALIVE` and `MEMTRACE_RAFT_ROUTER_LEAK` here.

components/batch-system/src/router.rs
rleungx, line 335: How about defining some variables for these numbers or add some comments?

components/batch-system/src/router.rs
```diff
@@ -272,14 +299,45 @@ where
...
+
+    pub fn alive_cnt(&self) -> Arc<AtomicUsize> {
+        self.normals.lock().unwrap().alive_cnt.clone()
+    }
+
+    pub fn trace(&self) -> RouterTrace {
+        let alive = self.normals.lock().unwrap().alive_cnt.clone();
+        let total = self.state_cnt.load(Ordering::Relaxed);
+        let alive = alive.load(Ordering::Relaxed);
+        let leak = if total > alive + 1 {
```
rleungx, line 326: Do we need an assertion to make sure the total won't be less than alive?

components/raftstore/src/store/memory.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+#[derive(MemoryTraceHelper, Default)]
+pub struct RaftContextTrace {
```
rleungx, line 16: Where do we use it?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3764,9 +3792,36 @@ where
...
+    }
+
+    pub fn close(&self, region_id: u64) {
+        self.router.close(region_id);
+        self.update_router_trace();
+    }
+
+    fn update_router_trace(&self) {
+        let router_trace = self.router.trace();
+        RAFTSTORE_MEM_TRACE
```
rleungx, line 3816: How about extracting a function so that we don't need to write it multiple times?

components/batch-system/src/router.rs
5kbpers: `state_cnt ` is modified when a mailbox created and destroyed.`alive_cnt` is modified when a mailbox poped and pushed from hashmap.
They could be different since we have a mailbox cache.

components/batch-system/src/router.rs
5kbpers: Which operation has better performance? 🤔

components/raftstore/src/store/memory.rs
5kbpers: In the next PR🤣, will remove it.

components/batch-system/src/router.rs
5kbpers: Should not because we get those two numbers asynchronously.

components/raftstore/src/store/fsm/apply.rs
rleungx: I think there are 4 times that we do the reset operation.

components/batch-system/src/router.rs
```diff
@@ -272,14 +300,49 @@ where
...
+        };
+        let mailbox_unit = mem::size_of::<(u64, BasicMailbox<N>)>();
+        let state_unit = mem::size_of::<FsmState<N>>();
+        // Every message in crossbeam sender needs 8 bytes to store state.
+        let message_unit = mem::size_of::<N::Message>() + 8;
+        // crossbeam unbounded channel sender has a list of blocks. Every block has 31 unit
+        // and every sender has at least one sender.
+        let sender_block_unit = 31;
+        RouterTrace {
+            alive: alive * mailbox_unit * 8 / 7 // hashmap uses 7/8 of allocated memory.
```
rleungx, line 340: 8/7 or 7/8?

components/batch-system/src/router.rs
nolouch: I think the problem is the atomic issue about the two operators `map insert` and `map len`.

components/batch-system/src/router.rs
5kbpers: It uses 7/8 memory so we need to multiple 8/7 to get actually allocated size.

components/raftstore/src/store/fsm/apply.rs
5kbpers: it's for different two routers.

rleungx: We still can provide a function and use a parameter to distinguish them. But It's ok for me with the current implementation.

5kbpers: Changed them to different static variables.

components/batch-system/src/router.rs
rleungx: got it. BTW, what not `(mailbox_unit * 8 / 7 + state_unit + message_unit * sender_block_unit) * alive`

components/raftstore/src/store/fsm/apply.rs
rleungx: much better

rleungx:
Do we need some tests?

5kbpers:
> Do we need some tests?

Added a unit test.

hicqu (approved):

rleungx (approved):

PR #10222, opened 2021-05-20
*: check request deadline before propose

sticnarf:

Part of #10166

A deadline is passed to `async_write_ext`. The deadline is checked before proposal and an error will be returned if the deadline exceeds.

TiKV clients cannot receive the response of expired requests. So these requests are actually meaningless to execute. This PR avoids wasting resource for requests that wait in TiKV too long.

### Check List 

Tests 

- Unit test

### Release note 

- Check deadline before proposing a write command

### DIFF
--- a/components/raftstore/src/router.rs
+++ b/components/raftstore/src/router.rs
@@ -48,19 +48,16 @@ where
 
     /// Sends RaftCmdRequest to local store.
     fn send_command(&self, req: RaftCmdRequest, cb: Callback<EK::Snapshot>) -> RaftStoreResult<()> {
-        self.send_command_with_deadline(req, cb, None)
+        send_command_impl::<EK, _>(self, req, cb, None)
     }
 
     fn send_command_with_deadline(
         &self,
         req: RaftCmdRequest,
         cb: Callback<EK::Snapshot>,
-        deadline: Option<Deadline>,
+        deadline: Deadline,
     ) -> RaftStoreResult<()> {
-        let region_id = req.get_header().get_region_id();
-        let cmd = RaftCommand::with_deadline(req, cb, deadline);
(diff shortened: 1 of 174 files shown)


components/raftstore/src/router.rs
```diff
@@ -48,8 +48,17 @@ where
...
     /// Sends RaftCmdRequest to local store.
     fn send_command(&self, req: RaftCmdRequest, cb: Callback<EK::Snapshot>) -> RaftStoreResult<()> {
+        self.send_command_with_deadline(req, cb, None)
+    }
+
+    fn send_command_with_deadline(
+        &self,
+        req: RaftCmdRequest,
+        cb: Callback<EK::Snapshot>,
+        deadline: Option<Deadline>,
```
hicqu, line 58: It names `with_deadline`, so is `Option` unnecessary?

tests/failpoints/cases/test_storage.rs
```diff
@@ -1104,3 +1104,44 @@ fn test_before_async_write_deadline() {
...
 }
+
+#[test]
+fn test_before_propose_deadline() {
+    let mut cluster = new_server_cluster(0, 1);
+    cluster.run();
+
+    let engine = cluster
+        .sim
+        .read()
```
hicqu, line 1115: Better to be `sim.rl().storage[&1].clone()`

zhangjinpeng1987:
How about the impact on performance?

sticnarf:
> How about the impact on performance?

The PR itself should have no visible performance change. It is very cheap to get a coarse time.

youjiali1995 (approved):

youjiali1995:
@hicqu PTAL

hicqu (approved):

PR #10226, opened 2021-05-21
raft_client: fix imbalance load of raft msg

gengliqi:

Problem Summary:

1. imbalance load of raft msg: raft client uses `region_id % grpc-raft-conn-num` to choose the specific connection. It's no problem when `grpc-raft-conn-num` is 1(default value). However, when `grpc-raft-conn-num` is larger, for example, 2, the distribution maybe not uniform because of the region id. If the num of TiKV is 3, the region id maybe always even because PD allocates region id with 3 peer id. To fix this problem, @5kbpers suggests using seahash. I test it with a parity check and find the distribution is very uniform. Even if the value is even, its hash value is odd or even. Also, the speed of seahash is good enough(benchmark result shows 11ns/iter on my laptop).
~~2. raft msg flush strategy: The raft msg is flushed only after handling all regions msg of this round. It may introduce some unnecessary latency. However, if the raft msg is flushed every time, the grpc CPU will be very high because of a large number of notifying. To alleviate this problem, the delay time is introduced. But I think this method is not very elegant. Maybe we should refactor this logic later.~~

What's Changed:
explained above.

### Related changes

### Check List 

Tests 

- No code

Side effects

- No

### Release note 
* No Release Note


BusyJay:
>  It may introduce some unnecessary latency.

Do you have any stats to show how large is the latency?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -681,6 +681,7 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> RaftPoller<EK, ER, T> {
                 PeerFsmDelegate::new(&mut peers[ready.batch_offset], &mut self.poll_ctx)
                     .post_raft_ready_append(ready);
             }
+            self.poll_ctx.trans.try_flush();
```
hicqu, line 684: Is it compatible with your origin design? @BusyJay

hicqu:
Maybe it's necessary to do a benchmark for the new flush mechanism. @gengliqi maybe your new mechanism can be used with https://github.com/tikv/grpc-rs/pull/469?

zhangjinpeng1987:
Better to split this PR into two, each one solves an individual issue.

src/server/raft_client.rs
```diff
@@ -844,7 +845,20 @@ where
...
+            if self.last_hash.0 == 0 || msg.region_id != self.last_hash.0 {
+                let mut hasher = SeaHasher::new();
+                hasher.write_u64(msg.region_id);
+                let h = hasher.finish();
+                self.last_hash = (
+                    msg.region_id,
+                    h % self.builder.cfg.grpc_raft_conn_num as u64,
+                );
+            };
+            self.last_hash.1 as usize
```
BusyJay, line 860: I'm afraid this will introduce out of order messages.

gengliqi: Why? I think it is the same as the former logic that the same `region_id` corresponds to the same `conn_id`.

tests/integrations/config/test-custom.toml
```diff
@@ -43,6 +43,10 @@ status-addr = "example.com:443"
 advertise-status-addr = "example.com:443"
 status-thread-pool-size = 1
 max-grpc-send-msg-len = 6291456
+raft-client-grpc-send-msg-buffer = 1263616
+raft-client-queue-size = 1234
+raft-msg-max-batch-size = 123
+raft-msg-flush-delay-us = 1234
```
BusyJay, line 49: Stale change.

src/server/raft_client.rs
```diff
@@ -844,7 +845,20 @@ where
     /// are sent out.
     pub fn send(&mut self, msg: RaftMessage) -> result::Result<(), DiscardReason> {
         let store_id = msg.get_to_peer().store_id;
-        let conn_id = (msg.region_id % self.builder.cfg.grpc_raft_conn_num as u64) as usize;
+        let conn_id = if self.builder.cfg.grpc_raft_conn_num == 1 {
+            0
+        } else {
+            if self.last_hash.0 == 0 || msg.region_id != self.last_hash.0 {
+                let mut hasher = SeaHasher::new();
```
BusyJay, line 852: You can use `seahash::hash(msg.region_id.as_ne_bytes())` to get better performance.

gengliqi: Done.

BusyJay (approved):

zhangjinpeng1987 (approved):
LGTM

PR #10227, opened 2021-05-21
config: add memory-usage-limit

hicqu:

Add configurations for memory usage limitation.

### Check List 

Tests 

- Unit test

### Release note 
* No release note.

### DIFF
--- a/etc/config-template.toml
+++ b/etc/config-template.toml
@@ -46,7 +46,7 @@
 # abort-on-panic = false
 
 ## Memory usage limit for the TiKV instance. If it's not set, the total of system memory
-# is avaiable to TiKV. Must be less than the total.
+## is avaiable to TiKV. Must be less than the total.
 # memory-usage-limit = "128GB"
 ## TiKV can oppress itself if memory usage ratio reaches `memory-usage-high-water`.
 # memory-usage-high-water = 0.8
--- a/etc/config-template.toml
+++ b/etc/config-template.toml
@@ -45,6 +45,12 @@
 ## being filled up.
 # abort-on-panic = false
 
+## Memory usage limit for the TiKV instance. If it's not set, the total of system memory
+# is avaiable to TiKV. Must be less than the total.
+# memory-usage-limit = "128GB"
+## TiKV can oppress itself if memory usage ratio reaches `memory-usage-high-water`.
+# memory-usage-high-water = 0.8
+
 ## Configurations for the single thread pool serving read requests.
 [readpool.unified]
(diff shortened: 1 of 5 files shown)


etc/config-template.toml
```diff
@@ -45,6 +45,12 @@
 ## being filled up.
 # abort-on-panic = false
 
+## Memory usage limit for the TiKV instance. Set it to 0 means the total of system memory
+## is avaiable to TiKV.
```
rleungx, line 49:
```suggestion
## is available to TiKV.
```

src/config.rs
```diff
@@ -2686,6 +2738,13 @@ impl TiKvConfig {
         cfg.cfg_path = tmp.path().join(LAST_CONFIG_FILE).display().to_string();
         Ok((cfg, tmp))
     }
+
+    fn default_memory_usage_limit() -> ReadableSize {
+        // TODO: is it necessary to reserve some space?
+        let reserve = 0;
```
rleungx, line 2744: What is it used for?

src/config.rs
```diff
@@ -2460,6 +2489,29 @@ impl TiKvConfig {
...
+                    .adjust_memory_usage_limit(self.memory_usage_limit);
+                self.raftdb
+                    .adjust_memory_usage_limit(self.memory_usage_limit);
+            }
+            cmp::Ordering::Greater => {
+                let msg = format!(
+                    "memory_usage_limit {:?} is greater than total {:?}",
+                    self.memory_usage_limit, default_memory_usage_limit
+                );
+                return Err(msg.into());
```
rleungx, line 2510: what if print a log then use the default_memory_usage_limit?

src/config.rs
```diff
@@ -2460,6 +2489,29 @@ impl TiKvConfig {
...
         self.pessimistic_txn.validate()?;
         self.gc.validate()?;
+
+        let default_memory_usage_limit = Self::default_memory_usage_limit();
+        if self.memory_usage_limit.0 == 0 {
+            self.memory_usage_limit = default_memory_usage_limit;
+            return Ok(());
+        }
+        match self.memory_usage_limit.0.cmp(&default_memory_usage_limit.0) {
+            cmp::Ordering::Less => {
```
rleungx, line 2499: Should we also limit the minimum memory to avoid setting the wrong value?

hicqu: It's a good idea but we can't determine the minimum value easily.

rleungx:
BTW, maybe we should support dynamically changing the high water.

nolouch (approved):
lgtm

PR #10229, opened 2021-05-21
copr_v2: Enhance raw coprocessor protocol and API

andylokandy:

Issue Number: #9747

- Update protobuf to https://github.com/pingcap/kvproto/pull/765
- Rename `coprocessor_v2` request to `raw_coprocessor`
- Unify `StorageError` and `PluginError`
- Feed plugin key ranges instead of `Region`
- Other refactor

### Release note 

- No release note

### DIFF
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -49,7 +49,7 @@ make_auto_flush_static_metric! {
         remove_lock_observer,
         coprocessor,
         coprocessor_stream,
-        coprocessor_v2,
+        raw_coprocessor,
         mvcc_get_by_key,
         mvcc_get_by_start_ts,
         split_region,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2217,7 +2217,7 @@ dependencies = [
 [[package]]
 name = "kvproto"
 version = "0.0.2"
-source = "git+https://github.com/pingcap/kvproto.git?rev=511a3225c29f9272af83d28efed92eba835cd34f#511a3225c29f9272af83d28efed92eba835cd34f"
+source = "git+https://github.com/pingcap/kvproto.git?rev=f11c6fb6b3a53b214ee94d54651b01eb99239334#f11c6fb6b3a53b214ee94d54651b01eb99239334"
 dependencies = [
  "futures 0.3.8",
  "grpcio",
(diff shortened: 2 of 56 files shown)


andylokandy:
PTAL @Jibbow

skyzh:
/retitle coprv2: Enhance raw coprocessor protcol and API

skyzh:
/title coprv2: Enhance raw coprocessor protocol and API

skyzh (commented):
LGTM generally. Will approve after kvproto PR is merged. By the way, why passing `Vec<Range<Key>>` instead of `&[Range<Key>]` to the plugin?

andylokandy:
> By the way, why passing Vec<Range<Key>> instead of &[Range<Key>] to the plugin?

Because it will invoke raw coprocessor once per request, so there is no need to keep the ownership.

src/server/metrics.rs
```diff
@@ -49,7 +49,7 @@ make_auto_flush_static_metric! {
         remove_lock_observer,
         coprocessor,
         coprocessor_stream,
-        coprocessor_v2,
+        raw_coprocessor,
```
Jibbow, line 52: In some lines, we now have `raw_coprocessor`, in others `copr_v2`. Is this confusing?

Jibbow (approved):
Thank you! LGTM

zhongzc (approved):

src/server/metrics.rs
andylokandy: `copr_v2` plugin is able to handle both `raw_coprocessor` and `txn_coprocessor`. So at the request level, I named it `RawCoprocessor`, and at the plugin level, it's called `copr_v2`.

skyzh (approved):

src/server/metrics.rs
Jibbow: makes sense! thanks!

skyzh:
Seems that the CI fails constantly. We should investigate more before merging.

skyzh:
💣💣💣

skyzh:
cc @zhongzc seems that you need to lgtm again :(

andylokandy:
@Jibbow Would you like to `/lgtm` again?

PR #10231, opened 2021-05-24
refresh global memory usage periodically

hicqu:

Refresh global memory usage periodically.

It introduces 2 functions `record_global_mem_usage` and `get_global_mem_usage`. The former is called periodically in the stat thread initialized in the server module, and the latter can be called anywhere to get the current memory usage.

BWT `SysQuota` is refactored to avoid lock the global `SYS_INFO` every times. 

### Check List 

Tests 

- No code

Side effects

### Release note 
* No release note

### DIFF
--- a/components/tikv_util/src/sys/mod.rs
+++ b/components/tikv_util/src/sys/mod.rs
@@ -6,108 +6,87 @@ pub mod cpu_time;
 mod cgroup;
 
 // re-export some traits for ease of use
+use crate::config::{ReadableSize, KB};
+use lazy_static::lazy_static;
+use std::sync::atomic::{AtomicU64, Ordering};
+use sysinfo::RefreshKind;
 pub use sysinfo::{DiskExt, NetworkExt, ProcessExt, ProcessorExt, SystemExt};
 
-use crate::config::ReadableSize;
-use lazy_static::lazy_static;
-use std::sync::Mutex;
+pub const HIGH_PRI: i32 = -1;
+const CPU_CORES_QUOTA_ENV_VAR_KEY: &str = "TIKV_CPU_CORES_QUOTA";
+
+static GLOBAL_MEMORY_USAGE: AtomicU64 = AtomicU64::new(0);
 
 lazy_static! {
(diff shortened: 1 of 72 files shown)


rleungx:
_"it’s much better to keep the same instance of System around instead of recreating it multiple times."_

So I am not sure if it's the right way to recreate the system multiple times.

Connor1996 (commented):
rest LGTM

components/tikv_util/src/sys/mod.rs
```diff
@@ -6,108 +6,87 @@ pub mod cpu_time;
...
 
-        pub fn log_quota(&self) {
-            info!(
-                "memory limit in bytes: {}, cpu cores quota: {}",
-                self.memory_limit_in_bytes(),
-                self.cpu_cores_quota()
-            );
-        }
+    fn sysinfo_memory_limit_in_bytes() -> u64 {
+        let system = sysinfo::System::new_with_specifics(RefreshKind::new().with_memory());
```
Connor1996, line 75: should it reuse the global sysinfo?

hicqu: Refresh memory is very light and these functions are only called when validating `TiKvConfig` or environments. `SYS_INFO` is only for diagnoise now.
So both this and `record_global_memory_usage` doesn't reuse `SYS_INFO`.

Connor1996 (approved):

components/tikv_util/src/sys/mod.rs
```diff
@@ -6,108 +6,87 @@ pub mod cpu_time;
...
-#[cfg(not(target_os = "linux"))]
-pub mod sys_quota {
-    use super::super::config::KIB;
-    use super::{SystemExt, SYS_INFO};
 
-    pub struct SysQuota {}
+    #[cfg(not(target_os = "linux"))]
+    pub fn cpu_cores_quota() -> f64 {
+        let cpu_num = num_cpus::get() as f64;
+        super::limit_cpu_cores_quota_by_env_var(cpu_num)
```
rleungx, line 47: no need super?

PR #10234, opened 2021-05-24
*: update deps

BusyJay:

Problem Summary:

Resolve RUSTSEC-2021-0055, RUSTSEC-2021-0057, RUSTSEC-2021-0058.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

    Some self-signed code may not work anymore.

### Release note 
- No release note.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2893,8 +2893,9 @@ checksum = "77af24da69f9d9341038eba93a073b1fdaaa1b788221b00a69bce9e762cb32de"
 
 [[package]]
 name = "openssl-src"
-version = "111.10.2+1.1.1g"
-source = "git+https://github.com/busyjay/openssl-src-rs?branch=patch-for-m1#f882f3e06bfcd89b5a27daaecf6ef562cb7ff231"
+version = "111.15.0+1.1.1k"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


zhouqiang-cl (approved):
LGTM

tisonkun:
@BusyJay according to https://github.com/pingcap/tidb-binlog/issues/1049#issuecomment-856081837 and the description here:

```
Breaking backward compatibility
Some self-signed code may not work anymore.
```

shall we add a release note field that declare the changes as well as known case like 

```
Invalid self signed certificates will not be accepted by latest openssl. For example, CA and host's CN name can not be the same.
```

It is surely change behavior and user will be confused if he saw "some self-signed code may not work anymore", what is "some"?

BusyJay:
> It is surely change behavior and user will be confused if he saw "some self-signed code may not work anymore", what is "some"?

Actually we don't know either. "CA and host's CN name can not be the same" is just one known cause. There is no release note from openssl that states the changes, we can't conclude what works and what doesn't.

tisonkun:
> > It is surely change behavior and user will be confused if he saw "some self-signed code may not work anymore", what is "some"?
> 
> Actually we don't know either. "CA and host's CN name can not be the same" is just one known cause. There is no release note from openssl that states the changes, we can't conclude what works and what doesn't.

Aha. Well then 😂

PR #10236, opened 2021-05-24
update protobuf and raft

hicqu:

Update some dependencies.

### Related changes

### Check List 

Tests 

- No code

### Release note 
* No release note.

### DIFF
--- a/components/raftstore/src/lib.rs
+++ b/components/raftstore/src/lib.rs
@@ -17,3 +17,8 @@ pub mod router;
 pub mod store;
 pub use self::coprocessor::{RegionInfo, RegionInfoAccessor, SeekRegionCallback};
 pub use self::errors::{DiscardReason, Error, Result};
+
+fn bytes_capacity(b: &bytes::Bytes) -> usize {
+    let range = b.as_ptr_range();
+    range.end as usize - range.start as usize
+}
--- a/components/raftstore/src/lib.rs
+++ b/components/raftstore/src/lib.rs
@@ -18,7 +18,16 @@ pub mod store;
 pub use self::coprocessor::{RegionInfo, RegionInfoAccessor, SeekRegionCallback};
 pub use self::errors::{DiscardReason, Error, Result};
 
+// With feature protobuf-codec, `bytes::Bytes` is generated for `bytes` in protobuf.
+#[cfg(feature = "protobuf-codec")]
 fn bytes_capacity(b: &bytes::Bytes) -> usize {
-    let range = b.as_ptr_range();
-    range.end as usize - range.start as usize
+    // NOTE: it's correct because in raftstore all `bytes::Bytes`s are immutable,
+    // which means `clear` or similar others will never be called for them.
+    b.len()
+}
+
+// Currently `bytes::Bytes` are not available for prost-codec.
+#[cfg(feature = "prost-codec")]
+fn bytes_capacity(b: &Vec<u8>) -> usize {
+    b.capacity()
 }
--- a/components/raftstore/src/lib.rs
+++ b/components/raftstore/src/lib.rs
@@ -21,8 +21,10 @@ pub use self::errors::{DiscardReason, Error, Result};
 // With feature protobuf-codec, `bytes::Bytes` is generated for `bytes` in protobuf.
 #[cfg(feature = "protobuf-codec")]
 fn bytes_capacity(b: &bytes::Bytes) -> usize {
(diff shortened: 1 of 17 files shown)


components/raftstore/src/store/peer.rs
```diff
@@ -1733,9 +1733,11 @@ where
             if !self.is_leader() {
                 fail_point!("raft_before_follower_send");
             }
-            for vec_msg in ready.take_messages() {
-                self.send(&mut ctx.trans, vec_msg, &mut ctx.raft_metrics.send_message);
-            }
+            // TODO: fix it are async raftstore is introduced.
```
hicqu, line 1736: PTAL @gengliqi

components/raftstore/src/lib.rs
```diff
@@ -17,3 +17,8 @@ pub mod router;
 pub mod store;
 pub use self::coprocessor::{RegionInfo, RegionInfoAccessor, SeekRegionCallback};
 pub use self::errors::{DiscardReason, Error, Result};
+
+fn bytes_capacity(b: &bytes::Bytes) -> usize {
```
BusyJay, line 21: I think it's the same as `b.len()`.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2553,7 +2555,7 @@ mod tests {
         // Sync if context is marked sync.
         let context = ProposalContext::SYNC_LOG.to_vec();
         let mut e = Entry::default();
-        e.set_context(context);
+        e.set_context(context.into());
```
BusyJay, line 2558: Perhaps static slice can be used.

components/raftstore/src/store/util.rs
```diff
@@ -785,7 +785,7 @@ impl<'a> ChangePeerI for &'a ChangePeerRequest {
         let mut cc = eraftpb::ConfChange::default();
         cc.set_change_type(self.get_change_type());
         cc.set_node_id(self.get_peer().get_id());
-        cc.set_context(ctx);
+        cc.set_context(ctx.into());
```
BusyJay, line 788: How about making `to_confchange` accept `Bytes`?

hicqu: Then `to_confchange` will be hard to implement for `prost-codec`.

Considering `ctx` is not copied in the caller of `to_confchange`(components/raftstore/src/store/peer.rs:3180), let's keep it.

components/raftstore/src/store/peer_storage.rs
hicqu: `&'static [ProposalContext::SYNC_LOG]` is not allowed. To use static slice we need to add
```rust
fn ProposalContxt::sync_log() -> &'static [u8] { ... }
```
 Considering it's only a test, I prefer to keep it.

hicqu:
@gengliqi PTAL at the latest changes about persisted messages.

components/raftstore/src/store/peer.rs
```diff
@@ -1801,6 +1808,15 @@ where
             };
         }
 
+        if !self.is_leader() {
```
BusyJay, line 1811: Should check whether there are persisted messages instead of whether it's leader.

components/raftstore/src/store/peer.rs
```diff
@@ -1801,6 +1808,15 @@ where
...
+            let msgs = ready.take_persisted_messages();
+            if self.is_applying_snapshot() {
```
BusyJay, line 1813: How about keeping using `apply_snap_result.is_some()` to avoid confusion.

components/raftstore/src/store/peer.rs
```diff
@@ -1631,6 +1634,10 @@ where
                 }
                 self.post_pending_read_index_on_replica(ctx);
                 self.read_progress.update_applied(self.last_applying_idx);
+                if !self.pending_messages.is_empty() {
```
BusyJay, line 1637: There seems to be a race condition between canceling snapshot and generating ready. If a snapshot is canceled, `check_applying_snap` will return Idle.

hicqu: Commented on `CheckApplyingSnapStatus::Idle`.

BusyJay (approved):

SchrodingerZhu:
Any update? gRPC upgrade is depend ending on this

5kbpers (approved):

PR #10245, opened 2021-05-26
apply: further reduce cmd clone of cmd observer

NingLin-P:

Issue Number: close #xxx 

Problem Summary:

Even after https://github.com/tikv/tikv/pull/10138, TiKV still needs to clone applied cmd 1 time if only `resolevd-ts` worker is observing and 2 times if the `cdc` worker is also observing, and the memory problem mentioned at https://github.com/tikv/tikv/pull/10138 is still unsolved.

What's Changed:

This PR further reduces the cmd cloning by share the cmd with `post_apply` observer, after this PR, TiKV does not need to clone cmd if only `resolevd-ts` worker is observing and clone once if the `cdc` worker is also observing. This PR also reduces the abovementioned memory problem by filter `default_cf` data when there is only `resolevd-ts` worker observing the region.

### Check List 

Tests 

- Integration test

### Release note 
- No release note


components/raftstore/src/coprocessor/mod.rs
```diff
@@ -75,7 +75,7 @@ pub trait AdminObserver: Coprocessor {
     fn pre_apply_admin(&self, _: &mut ObserverContext<'_>, _: &AdminRequest) {}
 
     /// Hook to call after applying admin request.
-    fn post_apply_admin(&self, _: &mut ObserverContext<'_>, _: &mut AdminResponse) {}
+    fn post_apply_admin(&self, _: &mut ObserverContext<'_>, _: &AdminResponse) {}
```
BusyJay, line 78: It's on purpose to allow modifying the response.

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -278,12 +304,10 @@ impl CmdBatch {
 }
 
 pub trait CmdObserver<E>: Coprocessor {
-    /// Hook to call after preparing for applying write requests.
-    fn on_prepare_for_apply(&self, cdc_id: &ObserveHandle, rts_id: &ObserveHandle, region_id: u64);
-    /// Hook to call after applying a write request.
-    fn on_apply_cmd(&self, cdc_id: ObserveID, rts_id: ObserveID, region_id: u64, cmd: &Cmd);
     /// Hook to call after flushing writes to db.
-    fn on_flush_apply(&self, engine: E);
+    fn on_flush_applied_cmd_batch(&self, cmd_batches: Vec<CmdBatch>, engine: E);
```
BusyJay, line 308: Why not just `&mut Vec<CmdBatch>`? So only one method is necessary.

BusyJay (commented):
Any benchmark result to show the improvement?

nolouch:
PTAL @JmPotato

components/raftstore/src/coprocessor/mod.rs
NingLin-P: There is no observer implement this method, and I think it is risky to take `&mut` since there is no mechanism to prevent two `Coprocessor` mistakenly modify the same field.

BusyJay: > prevent two Coprocessor mistakenly modify the same field.

Two coprocessors should be allow to do so. It's coprocessor and callers of registration's responsibility to maintain correct ordering. For other methods, a bool is returned to allow skipping following all coprocessors.

components/raftstore/src/coprocessor/mod.rs
BusyJay: >  it depends on the observer's priority, 

That's the exact design goal of `priority`. Using different methods is too verbose.

components/raftstore/src/coprocessor/mod.rs
NingLin-P: Yes, but I think it is unnecessary to use `&mut` here since there is no such use case.

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -219,33 +219,59 @@ impl ObserveHandle {
...
+    None,
+    // Only observe lock related data (i.e `lock_cf`, `write_cf`)
+    LockRelated,
+    // Observe all data
+    All,
+}
+
+impl ObserveLevel {
+    fn from(cdc: &ObserveHandle, rts: &ObserveHandle) -> ObserveLevel {
+        match (cdc.is_observing(), rts.is_observing()) {
```
BusyJay, line 235: It's quite misleading and easy to write the wrong code. How about let it return `ObserveLevel` directly and use the maximum one?

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -202,7 +202,8 @@ macro_rules! push {
         };
         let vec = &mut $vec;
         vec.push(e);
-        vec.sort_by(|l, r| l.priority.cmp(&r.priority));
+        // reverse sorting, so observer with higher priority will call first
+        vec.sort_by(|l, r| r.priority.cmp(&l.priority));
```
NingLin-P, line 206: /cc @BusyJay

BusyJay: I don't think it's a good idea, low value means higher priority is a common sense. For example, linux considers lower nice value as higher priority; Java considers least element is on the head of PriorityQueue.

NingLin-P: So this is a bug?
https://github.com/tikv/tikv/blob/70ee9c22e2c2485586d280dd0b1a39b353ffced6/components/raftstore/src/coprocessor/dispatcher.rs#L322-L342

BusyJay: Yes. But I don't think `TableCheckObserver` needs to have higher priority as every key triggered by observer is a valid split key.

components/resolved_ts/src/observer.rs
```diff
@@ -52,62 +49,46 @@ impl<E: KvEngine> Observer<E> {
...
-                info!("failed to schedule change log event"; "err" => ?e);
-            }
+        let region = {
+            let mut r = Region::default();
+            r.mut_peers().push(Peer::default());
+            r
+        };
+        // Create a snapshot here for preventing the old value was GC-ed.
+        // TODO: only need it after enabling old value, may add a flag to indicate whether to get it.
+        let snapshot = if self.need_old_value {
```
5kbpers, line 79: For now resolved ts need not snapshot so we can remove this.

components/cdc/src/observer.rs
```diff
@@ -99,48 +95,34 @@ impl CdcObserver {
...
-            };
-            if let Err(e) = self.sched.schedule(Task::MultiBatch {
-                multi: batches,
-                old_value_cb: Box::new(get_old_value),
-            }) {
-                warn!("cdc schedule task failed"; "error" => ?e);
-            }
+        let mut region = Region::default();
+        region.mut_peers().push(Peer::default());
+        // Create a snapshot here for preventing the old value was GC-ed.
```
5kbpers, line 115:
```suggestion
        // Create a snapshot here for preventing the old value was GC-ed.
        // TODO: only need it after enabling old value, may add a flag to indicate whether to get it.
```

components/cdc/src/observer.rs
```diff
@@ -99,48 +95,34 @@ impl CdcObserver {
...
 
 impl<E: KvEngine> CmdObserver<E> for CdcObserver {
-    fn on_prepare_for_apply(&self, cdc: &ObserveHandle, rts: &ObserveHandle, region_id: u64) {
-        let is_observing = cdc.is_observing();
-        *self.last_batch_observing.borrow_mut() = is_observing;
-        if !is_observing {
+    // `CdcObserver::on_flush_applied_cmd_batch` should only invoke if `cmd_batches` is not empty
+    fn on_flush_applied_cmd_batch(&self, cmd_batches: &mut Vec<CmdBatch>, engine: &E) {
+        assert!(!cmd_batches.is_empty());
+        if cmd_batches.iter().all(|cb| cb.level != ObserveLevel::All) {
```
5kbpers, line 101: This is redundant, we have a filter below.

5kbpers (commented):
rest LGTM

components/cdc/src/observer.rs
NingLin-P: It is a fast path to avoid following action  (i.e making an empty vec) for most peers, see https://github.com/tikv/tikv/pull/10245#discussion_r642294846

5kbpers: I don't think it is fast because it would iterate over all batches...

NingLin-P: Addressed, PTAL again.

5kbpers (approved):

components/resolved_ts/src/observer.rs
```diff
@@ -52,62 +36,35 @@ impl<E: KvEngine> Observer<E> {
...
-        let is_observing = cdc.is_observing() || rts.is_observing();
-        *self.last_batch_observing.borrow_mut() = is_observing;
-        if !is_observing {
+    fn on_flush_applied_cmd_batch(
+        &self,
+        max_level: ObserveLevel,
+        cmd_batches: &mut Vec<CmdBatch>,
+        _: &E,
+    ) {
+        if max_level < ObserveLevel::LockRelated {
```
nolouch, line 53: max_level == None?

BusyJay (approved):

PR #10247, opened 2021-05-27
raftstore: support limit the count of the peer stats reported in heartbeat

Yisaer:

Problem Summary:

The peer stats reported in one heartbeat is no limited now.

Use `TopN` to limit the count of the reported peer stat in single store heartbeat.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
No release note
```


components/raftstore/src/store/worker/pd.rs
```diff
@@ -214,6 +216,43 @@ pub struct PeerStat {
...
+    pub region_id: u64,
+    pub report_read_bytes: u64,
+    pub report_read_keys: u64,
+    pub report_query_stats: QueryStats,
+}
+
+impl Ord for PeerReportReadStat {
+    fn cmp(&self, other: &Self) -> CmpOrdering {
+        match self.report_read_bytes.cmp(&other.report_read_bytes) {
+            CmpOrdering::Equal => self.report_read_keys.cmp(&other.report_read_keys),
```
lhy1024, line 230: only cmp byte and key? if a region with high qps and low byte, it may be filtered.

Yisaer: What's your advice to implement `Ord` for `QueryStats`?

lhy1024: In the previous logic, the highest dimension of any one of the dimensions would definitely be selected and sent to pd, but the current logic requires that all three dimensions in order be relatively high before they are sent to pd, which will reduce the search space of pd.

Yisaer: yep, I think this is a trade off. We reduce the search space in order to stable the store heartbeat.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -214,6 +216,40 @@ pub struct PeerStat {
...
+    fn cmp(&self, other: &Self) -> CmpOrdering {
+        self.report_stat.cmp(&other.report_stat)
+    }
+}
+
+impl Eq for PeerCmpReadStat {}
+
+impl PartialEq for PeerCmpReadStat {
+    fn eq(&self, other: &Self) -> bool {
+        self.report_stat == other.report_stat
```
lhy1024, line 243: how about `region_id`

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1413,10 +1452,77 @@ fn send_destroy_peer_message<EK, ER>(
...
+        bytes_topn_report.push(byte_cmp_stat);
+        let mut query_cmp_stat = cmp_stat.clone();
+        query_cmp_stat.report_stat = read_stat.read_query_stats.get_read_query_num();
+        stats_topn_report.push(query_cmp_stat);
+    }
+
+    keys_topn_report.into_iter().for_each(|x| {
+        if let Some(report_stat) = report_read_stats.get(&x.region_id) {
+            peer_stats.insert(x.region_id, report_stat.clone());
+            {}
```
lhy1024, line 1492: how about removing it?

components/raftstore/src/store/worker/pd.rs
Yisaer: Why compare by region_id?

components/raftstore/src/store/worker/pd.rs
Yisaer: It will fail to compile if I removed it.

components/raftstore/src/store/worker/pd.rs
lhy1024: sorry, peer_id

lhy1024 (approved):

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1413,10 +1452,77 @@ fn send_destroy_peer_message<EK, ER>(
...
+        keys_topn_report.push(key_cmp_stat);
+        let mut byte_cmp_stat = cmp_stat.clone();
+        byte_cmp_stat.report_stat = read_stat.read_bytes;
+        bytes_topn_report.push(byte_cmp_stat);
+        let mut query_cmp_stat = cmp_stat.clone();
+        query_cmp_stat.report_stat = read_stat.read_query_stats.get_read_query_num();
+        stats_topn_report.push(query_cmp_stat);
+    }
+
+    keys_topn_report.into_iter().for_each(|x| {
```
NingLin-P, line 1489: How about just `for x in keys_top_report`?

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1413,10 +1452,77 @@ fn send_destroy_peer_message<EK, ER>(
...
+        if let Some(report_stat) = report_read_stats.get(&x.region_id) {
+            peer_stats.insert(x.region_id, report_stat.clone());
```
NingLin-P, line 1491:
```suggestion
        if let Some(report_stat) = report_read_stats.remove(&x.region_id) {
            peer_stats.insert(x.region_id, report_stat);
```
Use `remove` instead of `get` so we don't need to cloning.

components/tikv_util/src/topn.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    // this is added for rust-clippy#1013
+    #[allow(clippy::type_complexity)]
+    type IntoIter = iter::Map<binary_heap::IntoIter<Reverse<T>>, fn(Reverse<T>) -> T>;
+
+    // note: IntoIterator doesn't require the result in order, there is an `IntoIterSorted`, implement that if necessary
+    fn into_iter(self) -> Self::IntoIter {
+        self.heap.into_iter().map(|Reverse(x)| x)
+    }
+}
```
NingLin-P, line 63: How about just make `heap` public, so we can just `xxx_topn.heap.into_iter()`

Yisaer: I tend not to expose `heap` in order to prevent `topn` from unexpected usage.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -214,6 +216,40 @@ pub struct PeerStat {
     pub approximate_size: u64,
 }
 
+#[derive(Default, Clone)]
+struct PeerReportReadStat {
+    pub region_id: u64,
+    pub read_keys: u64,
+    pub read_bytes: u64,
+    pub read_query_stats: QueryStats,
+}
```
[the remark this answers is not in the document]
Yisaer: `PeerReportReadStat` is not exactly `pdpb::PeerStat` due to `QueryStats`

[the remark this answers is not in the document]
NingLin-P: I think it is the same:
https://github.com/pingcap/kvproto/blob/706fcaf286c8dd07ef59349c089f53289a32ce4c/proto/pdpb.proto#L475-L480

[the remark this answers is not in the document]
NingLin-P: I think it is the same:
https://github.com/pingcap/kvproto/blob/706fcaf286c8dd07ef59349c089f53289a32ce4c/proto/pdpb.proto#L475-L480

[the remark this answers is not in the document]
Yisaer: updated.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1413,10 +1444,73 @@ fn send_destroy_peer_message<EK, ER>(
...
+    mut stats: pdpb::StoreStats,
+) -> pdpb::StoreStats {
+    if report_read_stats.len() < capacity * 3 {
+        for (_, read_stat) in report_read_stats {
+            let mut peer_stat = pdpb::PeerStat::default();
+            peer_stat.set_region_id(read_stat.region_id);
+            peer_stat.set_read_bytes(read_stat.read_bytes);
+            peer_stat.set_read_keys(read_stat.read_keys);
+            peer_stat.set_query_stats(read_stat.get_query_stats().clone());
+            stats.peer_stats.push(peer_stat);
```
NingLin-P, line 1459:
```suggestion
            stats.peer_stats.push(read_stat);
```

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1413,10 +1444,73 @@ fn send_destroy_peer_message<EK, ER>(
...
+
+    for x in stats_topn_report {
+        if let Some(report_stat) = report_read_stats.remove(&x.region_id) {
+            peer_stats.insert(x.region_id, report_stat);
+        }
+    }
+
+    for (_, report_stat) in peer_stats {
+        stats.peer_stats.push(report_stat);
+    }
```
NingLin-P, line 1501:
```suggestion
    for x in stats_topn_report {
        if let Some(report_stat) = report_read_stats.remove(&x.region_id) {
            stats.peer_stats.push(report_stat);
        }
    }
```

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/worker/pd.rs
Yisaer: I think we should keep `peer_stats` in order to remove duplicated peer stat in topN.

NingLin-P: With `report_read_stats.remove` there is not duplicated peer stat

Yisaer: updated.

NingLin-P (approved):
LGTM

Issue #10137, opened 2021-05-08
Substitute rocksdb write stall


Connor1996:
## Development Task
When TiKV write stall occurs, it usaully blocks all the following write requests, which incurs a dramatic QPS drop. To mitigate the sudden spike, we try to disable the RocksDB write stall and add a dynamic write controller to control the write rate at scheduler level smoothly.

[design document](https://github.com/tikv/rfcs/pull/67)

### Task breakdown:

- [x] Find a case to reproduce write stall steadily
- [ ] Evaluate the design under single TiKV instance setup
  - [ ] Compare master with the case that write stall is disabled
  - [ ] Compare master with the case that higher delay write rate is set
  - [ ] Implement the design, compare it with the above two cases
- [ ] Test under multiple TiKV instances setup
  - [ ] The case that write stall only occurs on some TiKV instances, not all
  - [ ] The case that pending-compaction-bytes changes a lot after dropping table
  - [ ] The case that balance region causes too many L0

tisonkun:
@Connor1996 you can easily refer this issue from subtask by `ref #10137` which makes use of GitHub mechanism and avoid maintaining the checklist above. I think that checklist is extremely easy to be outdated.

tisonkun:
You can take a look at 

* https://github.com/pingcap/tidb/issues/24169
* https://github.com/tikv/community/issues/118

as examples.

PR #10249, opened 2021-05-27
Substitute write stall with flow control in scheduler level

Connor1996:

Issue Number: close https://github.com/tikv/tikv/issues/10137 

What's Changed:

Do flow control at scheduler level, and disable the write stall mechanism of RocksDB.

The main idea of this flow controller is to throttle at a steady write rate, so that the number of L0 keeps around the threshold. When it falls below the threshold, the throttle state wouldn't exit right away. Instead, it may keep or increase the throttle speed depending on some statistics.

### Related changes

### Check List 

Tests 

- Unit test

#### For single-node setup with heavy TPCC load
The result shows there is no more QPS drop-0 phenomenon, and QPS is relatively steady compared to write stall. Though the maximum QPS is lower, new flow control gain better throughput

write stall
image

new flow control
image

#### For multiple-node setup with low TPCC load
Write stall behavior well in this case. The new flow control is close to the write stall’s behavior but not that steady. But the throughput is the same.

write stall
image

new flow control
image

#### For multiple-node setup with heavy TPCC load
Flow control is better, as you can see write stall causes some QPS drop but flow control doesn't

write stall(not finish running a whole round caused by physical accident, but it's enough to illustrate the QPS drop issue of write stall)
image

new flow control
image

### Release note 

- Substitute write stall with flow control in scheduler level


src/storage/txn/scheduler.rs
```diff
@@ -744,17 +748,27 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
-                        engine_cb,
-                        proposed_cb,
-                        committed_cb,
-                    ) {
-                        SCHED_STAGE_COUNTER_VEC.get(tag).async_write_err.inc();
-
-                        info!("engine async_write failed"; "cid" => cid, "err" => ?e);
-                        scheduler.finish_with_err(cid, e.into());
+                    {
+                        let _guard = self.control_mutex.lock().await;
```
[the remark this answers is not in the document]
Connor1996: You're right, it's not necessarily needed

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,953 @@
...
+                // still on start, should not throttle now
+                return;
+            }
+        }
+
+        let pending_compaction_bytes = checker.long_term_pending_bytes.get_avg() / RATIO_PRECISION;
+        drop(checker);
+
+        for (_, checker) in &self.cf_checkers {
+            if num < (checker.long_term_pending_bytes.get_recent() as f64) / RATIO_PRECISION {
```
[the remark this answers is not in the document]
Connor1996: Make sure the CF of the biggest pending compaction bytes takes priority, otherwise the discard ratio would change up and down.

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
Connor1996: After adding a duration histogram of the delay, I found that the delay may exceed 1s. The speed limit is changed every second, so without the mutex, the write flow can't throttled strictly.

image

截屏2021-07-02 下午2 08 37

components/server/src/server.rs
```diff
@@ -966,6 +976,9 @@ impl TiKVServer<RocksEngine> {
         let mut kv_db_opts = self.config.rocksdb.build_opt();
         kv_db_opts.set_env(env);
         kv_db_opts.add_event_listener(self.create_raftstore_compaction_listener());
+        if self.config.storage.disable_write_stall {
```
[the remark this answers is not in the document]
Connor1996: Not now, but we can support it in later PR if needed

components/engine_rocks/src/flow_listener.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+}
+
+impl EventListener for FlowListener {
+    fn on_flush_completed(&self, info: &FlushJobInfo) {
+        let mut total = 0;
+        let p = info.table_properties();
+        total += p.data_size() + p.index_size() + p.filter_size();
+        let _ = self
```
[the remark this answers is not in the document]
Connor1996: no need, error indicates the store is gracefully shutting-down.

components/engine_rocks/src/flow_listener.rs
```diff
@@ -0,0 +1,94 @@
...
+        // we can regard ingestion in L0 as a flush
+        if info.picked_level() == 0 {
+            let mut total = 0;
+            let p = info.table_properties();
+            total += p.data_size() + p.index_size() + p.filter_size();
+            let _ = self
+                .flow_info_sender
+                .lock()
+                .unwrap()
+                .send(FlowInfo::Flush(info.cf_name().to_owned(), total));
```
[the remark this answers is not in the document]
Connor1996: `AddL0` seems more confusing for me...

components/engine_rocks/src/flow_listener.rs
```diff
@@ -0,0 +1,94 @@
...
+                let _ = self
+                    .flow_info_sender
+                    .lock()
+                    .unwrap()
+                    .send(FlowInfo::L0Intra(info.cf_name().to_owned()));
+            } else {
+                let index = info.input_file_count() - info.num_input_files_at_output_level();
+                let props = info.table_properties();
+                let mut read_bytes = 0;
+                for i in 0..index {
```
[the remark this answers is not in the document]
Connor1996: No, we can't. 😢

components/engine_rocks/src/flow_listener.rs
```diff
@@ -0,0 +1,94 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```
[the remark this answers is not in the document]
Connor1996: FlowListener implements `EventListener` which is a trait from RocksDB, so it should be in the component `engine_rocks`

components/engine_rocks/src/misc.rs
```diff
@@ -329,6 +329,20 @@ impl MiscExt for RocksEngine {
...
     }
 
+    fn get_cf_pending_compaction_bytes(&self, cf: &str) -> Result<Option<u64>> {
+        let handle = util::get_cf_handle(self.as_inner(), cf)?;
+        Ok(self
+            .as_inner()
+            .get_property_int_cf(handle, ROCKSDB_PENDING_COMPACTION_BYTES))
+    }
+
+    fn get_cf_num_memtables(&self, cf: &str) -> Result<Option<u64>> {
```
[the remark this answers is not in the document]
Connor1996: there is a metrics called `SCHED_MEMTABLE_GAUGE`

components/engine_rocks/src/flow_listener.rs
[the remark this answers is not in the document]
yiwu-arbug: make a hash map for files in `props`?

src/storage/config.rs
```diff
@@ -24,7 +24,7 @@ const MAX_SCHED_CONCURRENCY: usize = 2 * 1024 * 1024;
 // second, and it takes about 100ms to process the write requests
 // on average, in that situation the writing bytes estimated 10MB,
 // here we use 100MB as default value for tolerate 1s latency.
-const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
+const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 200;
```
yiwu-arbug, line 27: mention the change in release note?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+// ability of L0 wouldn't change dramatically corresponding to the ability of
+// hardware. So we can record the flush flow when reach the threshold as target
+// flow, and increase or decrease the throttle speed based on whether current
+// flush flow is smaller/larger than target flow.
+
+// For compaction pending bytes, we use discardable ratio...(TBD)
+pub struct FlowController {
+    discard_ratio: Arc<AtomicU64>,
+    limiter: Arc<Limiter>,
+    tx: Sender<bool>,
```
tabokie, line 65: Why is this needed? I thought `flow_receiver` will return error if server is shutdown.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+            limiter,
+            tx,
+            handle,
+        }
+    }
+
+    pub fn should_drop(&self) -> bool {
+        let ratio = self.discard_ratio.load(Ordering::Relaxed) as f64 / RATIO_PRECISION;
+        let mut rng = rand::thread_rng();
+        rng.gen::<f64>() < ratio
```
tabokie, line 114: Could we generate rand int in [0, RATIO_PRECISION] instead? BTW maybe rename it to `RATIO_SCALE_FACTOR`.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+            self.total -= v;
+        }
+
+        self.total += record;
+
+        self.records.push_back((record, Instant::now_coarse()));
+        self.clean_timeout();
+    }
+
+    fn clean_timeout(&mut self) {
```
tabokie, line 150:
```suggestion
    fn remove_stale_records(&mut self) {
```

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+                            }
+                            checker.check_long_term_l0_files(cf, l0_bytes)
+                        }
+                        Ok(FlowInfo::L0Intra(cf)) => {
+                            if let Some(throttle_cf) = checker.throttle_cf.as_ref() {
+                                if throttle_cf == &cf {
+                                    spare_ticks = 0;
+                                }
+                            }
+                            // do nothing for l0 intra compaction, it makes a bigger L0 file
```
tabokie, line 389: What if it's a ingested deletion file?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+    on_start_pending_bytes: bool,
+}
+
+impl Default for CFFlowChecker {
+    fn default() -> Self {
+        Self {
+            last_num_memtables: Smoother::default(),
+            long_term_pending_bytes: Smoother::default(),
+            long_term_num_l0_files: Smoother::default(),
+            last_num_l0_files: 0,
```
tabokie, line 297: `last_num_l0_files == long_term_num_l0_files.get_recent()`?

tabokie (commented):
will go over the core algorithm later ;_;

yiwu-arbug (commented):
I incline to move the whole new logic, including the flow controller and the event listener to a separate crate, which make the code structure more clear. The rocks event listener can definitely be exported from the engine_rocks crate.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+                    > self.cf_checkers[throttle_cf]
+                        .long_term_num_l0_files
+                        .get_max()
+                        + 4
+                {
+                    SCHED_THROTTLE_ACTION_COUNTER
+                        .with_label_values(&[&cf, "change_throttle_cf"])
+                        .inc();
+                    self.throttle_cf = Some(cf.clone());
+                    self.last_target_file = Some(num_l0_files);
```
yiwu-arbug, line 675: what does `last_target_file` mean to be? maybe it need a better name.

yiwu-arbug:
Some more general comments:
1. Please support dynamic change `disable_write_stall` before v5.2.0 code freeze. Could be in another PR. Let's also change `disable_write_stall` to a better name.
2. The code could have some refactoring. The core logic is not that complex as it currently looks like.
3. Disable raftdb flow control and write stall, if it is not the case. If raftdb compaction lags behind, it will only slowdown user writes, not the reads or other mechanism, so there's no point to impose any flow control on raftdb.

src/storage/config.rs
Connor1996: won't change it, revert it

components/engine_rocks/src/flow_listener.rs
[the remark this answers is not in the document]
Connor1996: IntroL0Compaction is recorded as `FlowInfo::L0Intra`, `Flush` only consists of flush and ingest L0

yiwu-arbug (approved):
LGTM in advance to unblock merging.

src/config.rs
```diff
@@ -2526,6 +2526,38 @@ impl TiKvConfig {
         self.backup.validate()?;
         self.pessimistic_txn.validate()?;
         self.gc.validate()?;
+
+        if self.storage.disable_write_stall {
+            self.raftdb.defaultcf.soft_pending_compaction_bytes_limit = ReadableSize::gb(500);
```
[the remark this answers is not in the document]
Connor1996: yes, so it does nothing for raftdb

src/storage/txn/flow_controller.rs
Connor1996: For dynamical configurable, it's needed

src/storage/txn/flow_controller.rs
Connor1996: It's hard to distinguish it, and it only takes a small part.

src/storage/txn/flow_controller.rs
Connor1996: I explained it in the latest comment. `last_num_l0_files` records last number of L0 files right after flush or L0 compaction, whereas `long_term_num_l0_files` only records L0 files right after L0 compaction.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+                left += r.0 as f64;
+            } else if i + 1 > half {
+                right += r.0 as f64;
+            } else {
+                if self.records.len() % 2 == 0 {
+                    left += r.0 as f64;
+                } else {
+                    continue;
+                }
+            }
```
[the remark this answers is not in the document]
Connor1996: the logic are same, ignoring or spliting the middle value doesn't make any difference for calculating slope. Though it's more neat.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+        } else {
+            INFINITY
+        };
+
+        self.update_speed_limit(throttle)
+    }
+
+    fn update_speed_limit(&mut self, mut throttle: f64) {
+        if throttle < MIN_THROTTLE_SPEED {
```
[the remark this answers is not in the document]
Connor1996: if the throttle speed is decreased to 0, then it can't be increased by multiplying a number.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+use rand::Rng;
+use tikv_util::time::{duration_to_sec, Consume, Instant, Limiter};
+
+use crate::storage::config::Config;
+use crate::storage::metrics::*;
+use crate::storage::Engine;
+
+const ADJUST_DURATION: u64 = 1000; // 1000ms
+const RATIO_PRECISION: f64 = 10000000.0;
+const EMA_FACTOR: f64 = 0.6;
```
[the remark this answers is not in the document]
Connor1996: Exponential Moving Average

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+            last_record_time: Instant::now_coarse(),
+        }
+    }
+
+    fn start(self, rx: Receiver<bool>, flow_info_receiver: Receiver<FlowInfo>) -> JoinHandle<()> {
+        Builder::new()
+            .name(thd_name!("flow-checker"))
+            .spawn(move || {
+                tikv_alloc::add_thread_memory_accessor();
+                let mut checker = self;
```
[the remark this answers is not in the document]
Connor1996: just a rename and move self into the closure.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+                // the write is accumulating, still need to throttle
+                checker.on_start_pending_bytes = false;
+            } else {
+                // still on start, should not throttle now
+                return;
+            }
+        }
+
+        let pending_compaction_bytes = checker.long_term_pending_bytes.get_avg() / RATIO_PRECISION;
+        drop(checker);
```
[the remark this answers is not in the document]
Connor1996: can't do it. some previous variables are needed in the below.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+use engine_traits::{CFNamesExt, MiscExt};
+use rand::Rng;
+use tikv_util::time::{duration_to_sec, Consume, Instant, Limiter};
+
+use crate::storage::config::Config;
+use crate::storage::metrics::*;
+use crate::storage::Engine;
+
+const ADJUST_DURATION: u64 = 1000; // 1000ms
+const RATIO_PRECISION: f64 = 10000000.0;
```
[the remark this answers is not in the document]
Connor1996: need it still cause f64 can't be atomic.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+
+struct FlowChecker<E: Engine> {
+    pending_compaction_bytes_soft_limit: u64,
+    pending_compaction_bytes_hard_limit: u64,
+    memtables_threshold: u64,
+    l0_files_threshold: u64,
+
+    cf_checkers: HashMap<String, CFFlowChecker>,
+    throttle_cf: Option<String>,
+    discard_ratio: Arc<AtomicU64>,
```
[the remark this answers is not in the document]
Connor1996: don't think so, `FlowController` needs to be `Sync`.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+        for (_, checker) in &self.cf_checkers {
+            if num < (checker.long_term_pending_bytes.get_recent() as f64) / RATIO_PRECISION {
+                return;
+            }
+        }
+
+        let ratio = if pending_compaction_bytes < soft {
+            0
+        } else {
+            let new_ratio = (pending_compaction_bytes - soft) / (hard - soft);
```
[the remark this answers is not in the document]
Connor1996: `Say if the pending bytes just spike for a short period of time, no matter how large the value gets, maybe we don't need to throttle a lot.`

That's why I use a large moving window to smooth pending bytes and logarithm the pending bytes. It would avoid the spike of it.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+                            checker.adjust_memtables(&cf);
+                            checker.check_l0_flow(cf, flush_bytes)
+                        }
+                        Ok(FlowInfo::Compaction(cf)) => {
+                            if let Some(throttle_cf) = checker.throttle_cf.as_ref() {
+                                if throttle_cf == &cf {
+                                    spare_ticks = 0;
+                                }
+                            }
+                            checker.adjust_pending_compaction_bytes(cf);
```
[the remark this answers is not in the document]
Connor1996: No, include L0-involved compaction.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+        } else {
+            for (cf, _) in &self.cf_checkers {
+                SCHED_THROTTLE_CF_GAUGE.with_label_values(&[cf]).set(0);
+            }
+        }
+
+        let rate =
+            self.limiter.total_bytes_consumed() as f64 / self.last_record_time.elapsed_secs();
+        if self.limiter.total_bytes_consumed() != 0 {
+            self.recorder.observe(rate as u64);
```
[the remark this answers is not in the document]
Connor1996: 1. for close loop system, if all the requests are delayed(assume > 1s) in the TiKV scheduler, then in the next second, the write rate would be 0. But it doesn't reflect the real write rate, so just ignore it.
2. use `write_flow_recorder` now

Connor1996:
1. Make `enable-flow-control` dynamical configurable (test will be added tomorrow)
2. Some comments are addressed, others are ongoing.
3. Paste test result in the PR description.

PTAL again @yiwu-arbug @tabokie

src/storage/txn/flow_controller.rs
tabokie: I think with #7794, this case can happen quite frequently.

src/storage/txn/flow_controller.rs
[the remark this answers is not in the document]
tabokie: My mistake, how about ingestions to Ln? Will that trigger pending bytes changes?

src/config.rs
```diff
@@ -2526,6 +2538,14 @@ impl TiKvConfig {
         self.backup.validate()?;
         self.pessimistic_txn.validate()?;
         self.gc.validate()?;
+
+        if self.storage.enable_flow_control {
+            self.raftdb.defaultcf.disable_write_stall = true;
```
tabokie, line 2543: Why disable raftdb?

src/storage/txn/scheduler.rs
```diff
@@ -744,17 +747,30 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
-                        to_be_write,
-                        engine_cb,
-                        proposed_cb,
-                        committed_cb,
-                    ) {
-                        SCHED_STAGE_COUNTER_VEC.get(tag).async_write_err.inc();
-
-                        info!("engine async_write failed"; "cid" => cid, "err" => ?e);
-                        scheduler.finish_with_err(cid, e.into());
+                    {
```
tabokie, line 750: `if (flow_controller.enabled())`?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1064 @@
...
+            limiter,
+            tx,
+            handle: Some(checker.start(rx, flow_info_receiver)),
+        }
+    }
+
+    pub fn should_drop(&self) -> bool {
+        let ratio = self.discard_ratio.load(Ordering::Relaxed);
+        let mut rng = rand::thread_rng();
+        rng.gen::<f64>(ratio, RATIO_SCALE_FACTOR)
```
tabokie, line 150: suppose to be `gen_ratio`?

src/config.rs
Connor1996: In the case where raftdb stalls, kvdb should be flow controlled first, cause kvdb usually has more compaction flow comparing to raftdb.

src/storage/txn/flow_controller.rs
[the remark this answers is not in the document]
Connor1996: Seems yes. Add the ingestion result as a compaction flow.

src/config.rs
```diff
@@ -546,6 +551,7 @@ impl Default for DefaultCfConfig {
             max_bytes_for_level_multiplier: 10,
             compaction_style: DBCompactionStyle::Level,
             disable_auto_compactions: false,
+            disable_write_stall: true,
```
sticnarf, line 554: To my understanding they should be `false` by default and controlled by `flow_control`?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1069 @@
...
+use collections::HashMap;
+use engine_rocks::FlowInfo;
+use engine_traits::KvEngine;
+use rand::Rng;
+use tikv_util::time::{duration_to_sec, Consume, Instant, Limiter};
+
+use crate::storage::config::FlowControlConfig;
+use crate::storage::metrics::*;
+
+const SPARE_TICK_DURATION: u64 = 1000; // 1000ms
```
sticnarf, line 21: Use `Duration` directly?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+        let checker = self.cf_checkers.get_mut(cf).unwrap();
+        SCHED_MEMTABLE_GAUGE
+            .with_label_values(&[cf])
+            .set(num_memtables as i64);
+        let prev = checker.last_num_memtables.get_recent();
+        checker.last_num_memtables.observe(num_memtables);
+        if checker.on_start_memtable {
+            if num_memtables < self.memtables_threshold {
+                checker.on_start_memtable = false;
+            } else if checker.long_term_pending_bytes.trend() == Trend::Increasing {
```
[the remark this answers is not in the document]
Connor1996: fix it

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
sticnarf: Could you add some comments here? I think this mutex is not very intuitive.

src/config.rs
Connor1996: you're right

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
sticnarf: If flow control is not triggered, I am a bit afraid this mutex may cause regression. Do you experience discoverable regression in your test?

src/config.rs
tabokie: I think we should keep it enabled for now. I came across some practical cases where raftdb has much more compaction than kvdb (can't recall the exact benchmark setting though), other than that, in some extreme cases (raftdb disk slow, cpu throttle) raftdb stall can be useful to avoid OOM.

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
tabokie: The limiter from `async-speed-limit` has an atomic flag [`is_unlimited`](https://github.com/tikv/async-speed-limit/blob/691439dd36a9dc2fe308964625793a198b877a96/src/limiter.rs#L301), maybe we can expose it and avoid mutex call until flow control kicks in.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1072 @@
...
+        let discard_ratio = Arc::new(AtomicU64::new(0));
+        let checker = FlowChecker::new(config, engine, discard_ratio.clone(), limiter.clone());
+        let (tx, rx) = mpsc::sync_channel(5);
+
+        tx.send(if config.enable {
+            Msg::Enable
+        } else {
+            Msg::Disable
+        })
+        .unwrap();
```
tabokie, line 139: can we reuse `fn enable(bool enable)`?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1072 @@
...
+                            enabled = true;
+                        }
+                        Err(_) => {}
+                    }
+
+                    if !enabled {
+                        // do nothing, just consume the flow info channel
+                        let _ = flow_info_receiver.recv();
+                        continue;
+                    }
```
tabokie, line 467: Since we are still receiving flow_info, would it be better to do the check as well? No much overhead I presume.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1072 @@
...
+                                if throttle_cf == &cf {
+                                    spare_ticks = 0;
+                                }
+                            }
+                            checker.adjust_pending_compaction_bytes(cf);
+                        }
+                        Err(RecvTimeoutError::Timeout) => {
+                            spare_ticks += 1;
+                            if spare_ticks == SPARE_TICKS_THRESHOLD {
+                                // there is no flush/compaction happens, we should speed up if throttled
```
tabokie, line 509: Could you explain again why this happens? Compaction is blocked somehow?

src/config.rs
Connor1996: OOM or disk full caused by raftdb's accumulation should be considered by another mechanism now.

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
Connor1996: Yes, better to avoid mutex call if not throttled. Working on it

src/storage/txn/flow_controller.rs
Connor1996: do the check, then what?

src/storage/txn/flow_controller.rs
Connor1996: the write is stopped, so there is nearly no flush and compaction happens.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+                break;
+            }
+        }
+    }
+
+    pub fn get_recent(&self) -> u64 {
+        if self.records.len() == 0 {
+            return 0;
+        }
+        self.records.back().unwrap().0
```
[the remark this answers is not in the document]
Connor1996: it's intentional, returning 0 may break the logic of the algorithm. Better use `unwrap` to reveal the bug.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+
+        let ratio = if pending_compaction_bytes < soft {
+            0
+        } else {
+            let new_ratio = (pending_compaction_bytes - soft) / (hard - soft);
+            let old_ratio = self.discard_ratio.load(Ordering::Relaxed);
+            (if old_ratio != 0 {
+                self.factor * (old_ratio as f64 / RATIO_PRECISION) + (1.0 - self.factor) * new_ratio
+            } else {
+                new_ratio
```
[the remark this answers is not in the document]
Connor1996: good idea

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+            } else {
+                // still on start, should not throttle now
+                return;
+            }
+        }
+        drop(checker);
+
+        for (_, c) in &self.cf_checkers {
+            if num_memtables < c.last_num_memtables.get_recent() {
+                return;
```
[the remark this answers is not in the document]
Connor1996: No, it is just to make sure the CF having most #memtable takes control of the speed limit.

src/config.rs
tabokie: I don't think the OOM flow control will consider RocksDB memory..

tabokie: Anyway I just don't see the downside of using raftdb write stall as a safety net.

Connor1996: But I can't see any circumstance that RaftDB stalls but KvDB doesn't. Consider the case that both KvDB and RaftDB has a lot of pending compaction bytes(no matter which has more), flow control begins to increase discardable ratio to control the accumulation, but it won't surpass it under threshold. Whereas, RaftDB triggers write stall now and affects raftstore.

tabokie: If you are so worried about pending compaction bytes, you can simply set its limit to 0. I'm more concerned about memtable.

src/storage/txn/flow_controller.rs
tabokie: Why is that? I haven't seen compaction paused because foreground stops. And even if that's the case we shouldn't allow foreground writes when stall condition is still there.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+use crate::storage::metrics::*;
+
+const SPARE_TICK_DURATION: Duration = Duration::from_millis(1000);
+const SPARE_TICKS_THRESHOLD: u64 = 10;
+const RATIO_SCALE_FACTOR: f64 = 10000000.0;
+const EMA_FACTOR: f64 = 0.6; // EMA stands for Exponential Moving Average
+const LIMIT_UP_PERCENT: f64 = 0.04; // 4%
+const LIMIT_DOWN_PERCENT: f64 = 0.02; // 2%
+const MIN_THROTTLE_SPEED: f64 = 16.0 * 1024.0; // 16KB
+const MAX_THROTTLE_SPEED: f64 = 200.0 * 1024.0 * 1024.0; // 200MB
```
tabokie, line 28: Not sure if this is discussed before, but 200MB seems pretty small. Will the limit be turned on and off repeatedly on fast disk?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+use rand::Rng;
+use tikv_util::time::{duration_to_sec, Consume, Instant, Limiter};
+
+use crate::storage::config::FlowControlConfig;
+use crate::storage::metrics::*;
+
+const SPARE_TICK_DURATION: Duration = Duration::from_millis(1000);
+const SPARE_TICKS_THRESHOLD: u64 = 10;
+const RATIO_SCALE_FACTOR: f64 = 10000000.0;
+const EMA_FACTOR: f64 = 0.6; // EMA stands for Exponential Moving Average
```
tabokie, line 24: Maybe move these algorithm factor below along with `PID_XX_FACTOR`?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+// a approximate value, usually, changes up and down dramatically, so it's unwise
+// to map compaction pending bytes to a specified throttle speed. Instead,
+// mapping it from soft limit to hard limit as 0% to 100% discardable ratio. With
+// this, there must be a point that foreground write rate is equal to the
+// background compaction pending bytes consuming rate so that compaction pending
+// bytes is kept around a steady level.
+//
+// Here is a brief flow showing where the mechanism works:
+// grpc -> check should drop(discardable ratio) -> limiter -> async write to raftstore
+//
```
tabokie, line 77: Comment is a little too long.. Maybe simplify a little? or add some more bullets.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+        }
+    }
+}
+
+// CFFlowChecker records some statistics and states related to one CF.
+// These statistics falls into five categories:
+//   * memtable
+//   * L0 files
+//   * flush flow
+//   * L0 flow (compaction read flow of L0)
```
tabokie, line 314: These two names are hard to understand without reading the code. Maybe L0 consumption/production, L0 increase/decrease? I think we can hide RocksDB concept from algorithm as much as possible.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+        let (tx, _rx) = mpsc::sync_channel(0);
+
+        Self {
+            discard_ratio: Arc::new(AtomicU64::new(0)),
+            limiter: Arc::new(Limiter::new(INFINITY)),
+            enabled: Arc::new(AtomicBool::new(false)),
+            tx,
+            handle: None,
+        }
+    }
```
tabokie, line 122: I lost track of where this is used. Only for test?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+            enabled: Arc::new(AtomicBool::new(config.enable)),
+            tx,
+            handle: Some(checker.start(rx, flow_info_receiver)),
+        }
+    }
+
+    pub fn should_drop(&self) -> bool {
+        let ratio = self.discard_ratio.load(Ordering::Relaxed);
+        let mut rng = rand::thread_rng();
+        rng.gen_ratio(ratio as u32, RATIO_SCALE_FACTOR as u32)
```
tabokie, line 153: Change `discard_ratio` to `AtomicU32`?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+            if i < half {
+                left += r.0;
+            } else if self.records.len() - i - 1 < half {
+                right += r.0;
+            }
+        }
+
+        // decide if there is a trend by the two averages
+        // adding 2 here is to give a tolerance
+        if right > left + 2 {
```
tabokie, line 299: Have you tried using `slope` for calculation trend?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+    init_speed: bool,
+
+    // L0 files related
+    // last number of l0 files right after flush or L0 compaction
+    last_num_l0_files: u64,
+    // last number of l0 files right after flush
+    last_num_l0_files_from_flush: u64,
+    // a few records of number of l0 files right after L0 compaction
+    // As we know, after flush the number of L0 files must increase by 1,
+    // whereas, after L0 compaction the number of L0 files must decrease a lot
```
tabokie, line 330: Memtable has the same pattern, why not use similar method?

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+        SCHED_THROTTLE_FLOW_GAUGE.set(if throttle == INFINITY {
+            0
+        } else {
+            throttle as i64
+        });
+        self.limiter.set_speed_limit(throttle)
+    }
+
+    // Check flush flow to compare with target flow to decide whether need to adjust throttle speed
+    fn check_flush_flow(&mut self, cf: String, flush_bytes: u64) {
```
tabokie, line 902:
```suggestion
    fn on_l0_incr(&mut self, cf: String, flush_bytes: u64) {
```

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+                    self.limiter.speed_limit() * (1.0 + 5.0 * LIMIT_UP_PERCENT)
+                };
+
+                self.update_speed_limit(throttle)
+            }
+        }
+    }
+
+    // Check the number of l0 files to decide whether need to adjust target flow
+    fn check_l0_files(&mut self, cf: String, l0_bytes: u64) {
```
tabokie, line 763:
```suggestion
    fn on_l0_decr(&mut self, cf: String, l0_bytes: u64) {
```

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+                    + (1.0 - EMA_FACTOR) * new_ratio
+            } else {
+                if new_ratio > 0.01 { 0.01 } else { new_ratio } // cap the initial discard ratio
+            } * RATIO_SCALE_FACTOR) as u64
+        };
+        SCHED_DISCARD_RATIO_GAUGE.set(ratio as i64);
+        self.discard_ratio.store(ratio, Ordering::Relaxed);
+    }
+
+    fn adjust_memtables(&mut self, cf: &str) {
```
tabokie, line 652:
```suggestion
    fn on_memtable_decr(&mut self, cf: &str) {
```

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+                std::cmp::Ordering::Less => {
+                    checker.memtable_debt -= 1.0;
+                    1.0
+                }
+                std::cmp::Ordering::Equal => {
+                    // keep, do nothing
+                    0.0
+                }
+            };
+            self.limiter.speed_limit() + diff * 1024.0 * 1024.0
```
tabokie, line 725: I find it a little weird to not consider memtable size in this algorithm.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+            }
+        } else if !should_throttle
+            || checker.last_num_memtables.get_recent() < self.memtables_threshold
+        {
+            // should not throttle memtable
+            checker.memtable_debt = 0.0;
+            if checker.init_speed {
+                INFINITY
+            } else {
+                self.limiter.speed_limit() + checker.memtable_debt
```
tabokie, line 707: This seems wrong, where is the 1MB factor? Looks like memtable throttle is not well tested.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+        if self.limiter.total_bytes_consumed() != 0 {
+            self.write_flow_recorder.observe(rate as u64);
+        }
+        SCHED_WRITE_FLOW_GAUGE.set(rate as i64);
+        self.last_record_time = Instant::now_coarse();
+
+        self.limiter.reset_statistics();
+    }
+
+    fn adjust_pending_compaction_bytes(&mut self, cf: String) {
```
tabokie, line 592:
```suggestion
    fn on_pending_compaction_bytes_change(&mut self, cf: String) {
```

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+                }
+            } else {
+                SCHED_THROTTLE_ACTION_COUNTER
+                    .with_label_values(&[&cf, "no_target_flow"])
+                    .inc();
+            }
+        }
+    }
+
+    fn up_flow(&mut self, cf: String) {
```
tabokie, line 1000:
```suggestion
    fn increase_limit_for_l0(&mut self, cf: String) {
```

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+                u = 0.0;
+            };
+            SCHED_UP_FLOW_GAUGE.set((u * RATIO_SCALE_FACTOR) as i64);
+
+            self.limiter.speed_limit() + u
+        };
+        self.update_speed_limit(throttle)
+    }
+
+    fn down_flow(&mut self, cf: String) {
```
tabokie, line 1023:
```suggestion
    fn decrease_limit_for_l0(&mut self, cf: String) {
```

src/storage/txn/flow_controller.rs
tabokie: And why L0 uses a much more sophisticated algorithm than memtable throttle?

src/storage/txn/flow_controller.rs
Connor1996: yes, only for test

src/storage/txn/flow_controller.rs
tabokie: For auto-tuned limiter though, should merge this one to avoid similar issue: https://github.com/tikv/rocksdb/pull/234

src/storage/txn/flow_controller.rs
sticnarf: Personally speaking, not a bad idea for a module introduction.

src/storage/txn/flow_controller.rs
sticnarf: Personally speaking, not a bad idea for a module introduction.

src/storage/txn/flow_controller.rs
sticnarf: `CachePadded<AtomicU32>`?

src/storage/txn/flow_controller.rs
Connor1996: The scheduler writing flow is 100MB, and I haven't seen any disk can reach the 200MB/s for foreground write flow of leader

src/config.rs
```diff
@@ -3503,6 +3523,12 @@ mod tests {
             )
             .unwrap(),
         ));
+        let (_tx, rx) = std::sync::mpsc::channel();
```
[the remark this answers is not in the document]
Connor1996: these tests are only related to config, so no need to care about flow info.

src/storage/txn/flow_controller.rs
Connor1996: Nope, `slope` takes duration into consideration.

src/storage/txn/flow_controller.rs
Connor1996: memtables can be consumed one by one, but all L0 files are involved in one compaction due to range overlap, so #L0 is of great unsteadiness which complicates the algorithm much.

src/storage/txn/flow_controller.rs
Connor1996: but how can you decide the rate by memtable size?

src/storage/txn/flow_controller.rs
Connor1996: the number of memtables is steady, didn't see any case that the number of memtable will change up and down dramatically.

src/storage/txn/flow_controller.rs
Connor1996: yes... hard to see memtable accumulates but L0 doesn't.

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1055 @@
...
+            return 0;
+        }
+        self.records.iter().max_by_key(|(k, _)| k).unwrap().0
+    }
+
+    pub fn get_percentile_90(&mut self) -> u64 {
+        if self.records.is_empty() {
+            return 0;
+        }
+        let mut v = self.records.make_contiguous().to_vec();
```
sticnarf, line 247: `make_contiguous` is heavier and unnecessary. I think `.iter().collect()` should work well.

sticnarf: Or you needn't call `to_vec`. Just sort the `&mut [T]` returned by `make_contiguous`.

src/storage/txn/flow_controller.rs
tabokie: Check this, the frequency is higher, but it's still zigzagging. {DD13BF97-3DB5-4255-B842-DE6FEA6CBC8D}
Anyway, no objection here, just curious.

src/storage/txn/flow_controller.rs
tabokie: Ok, it would make sense to move Smoother to util crate if it weren't for this magic number.

src/storage/txn/flow_controller.rs
tabokie: Probably, then make it doc comments?

src/storage/txn/flow_controller.rs
sticnarf: I see `long_term_pending_bytes` also calls the `trend`, is `2` is still appropriate for it? 
The recorded value is log2(bytes) * 1e7, so `2` looks too small here?

Connor1996: oh, forget that it is multiplied by 1e7

src/config.rs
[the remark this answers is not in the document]
tabokie: Then how about not returning a FlowControlller to avoid confusion?

src/storage/txn/flow_controller.rs
tabokie: No sure, consider when write buffer size is super small, the throttle should be stronger.

src/storage/txn/flow_controller.rs
tabokie: Maybe use OS injection tool to stop disk writes? @5kbpers has done that before.

tabokie: This seems like reproducible: #9649. I'm thinking, if memtable stall is rare, how about using a super simple algorithm? e.g. discard all requests when memtables breach threshold.

Connor1996: I think it's simple enough which closes to rocksdb impl. Will try to reproduce it later, if it doesn't fit well, change it then

src/storage/txn/flow_controller.rs
```diff
@@ -0,0 +1,1128 @@
...
+// Smoother is a sliding window used to provide steadier flow statistics.
+struct Smoother<
+    T: Default
+        + Add<Output = T>
+        + Sub<Output = T>
+        + AddAssign
+        + SubAssign
+        + PartialOrd
+        + AsPrimitive<f64>
+        + FromPrimitive,
```
tabokie, line 192: use where?

tabokie (approved):
Looks good generally, some non-urgent todo items in case forgotten:
- Merge tikv/rocksdb#234, without this the compaction will be suppressed as well when foreground flow is limited
- Paste a screenshot for the newly introduced metrics
- Test memtable stalls
- Simplify the algorithm of FlowController, I personally haven't read through every line of it

PR #10251, opened 2021-05-27
raftstore: add mem trace

5kbpers:

Problem Summary:

Trace memory usages for raftstore

Proposal: [xxx](url) 

What's Changed:

* Add a trait for get memory usage in the heap
* Trace memory usages
* Update memory trace after a batch ends

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 
* No release note

### DIFF
--- a/components/engine_traits/src/raft_engine.rs
+++ b/components/engine_traits/src/raft_engine.rs
@@ -90,8 +90,6 @@ pub trait RaftLogBatch: Send {
     fn put_raft_state(&mut self, raft_group_id: u64, state: &RaftLocalState) -> Result<()>;
 
     fn is_empty(&self) -> bool;
-
-    fn data_size(&self) -> usize;
 }
 
 #[derive(Clone, Copy, Default)]
--- a/components/engine_traits/src/raft_engine.rs
+++ b/components/engine_traits/src/raft_engine.rs
@@ -90,6 +90,8 @@ pub trait RaftLogBatch: Send {
     fn put_raft_state(&mut self, raft_group_id: u64, state: &RaftLocalState) -> Result<()>;
 
     fn is_empty(&self) -> bool;
+
+    fn data_size(&self) -> usize;
 }
 
 #[derive(Clone, Copy, Default)]
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -388,8 +388,6 @@ where
     priority: Priority,
     /// Whether to yield high-latency operation to low-priority handler.
     yield_high_latency_operation: bool,
-
-    trace: ApplyContextTrace,
 }
 
 impl<EK, W> ApplyContext<EK, W>
@@ -440,7 +438,6 @@ where
             pending_create_peers,
             priority,
             yield_high_latency_operation: cfg.apply_batch_system.low_priority_pool_size > 0,
-            trace: ApplyContextTrace::default(),
         }
     }
 
@@ -3678,30 +3675,6 @@ where
 
     fn get_priority(&self) -> Priority {
         self.apply_ctx.priority
-    }
-
-    fn pause(&mut self) {
-        let mut size = self.apply_ctx.apply_res.heap_size();
-        size += self.msg_buf.heap_size();
-        size += mem::size_of::<Self>();
-        let trace = ApplyContextTrace {
-            cbs_size: self.apply_ctx.cbs.heap_size(),
-            write_batch: self.apply_ctx.kv_wb.data_size(),
-            rest: size,
-        };
-        if let Some(event) = self.apply_ctx.trace.reset(trace) {
-            MEMTRACE_APPLY_CONTEXT.trace(event);
-        }
-    }
-}
-
-impl<EK, W> Drop for ApplyPoller<EK, W>
-where
-    EK: KvEngine,
-    W: WriteBatch<EK>,
-{
-    fn drop(&mut self) {
-        MEMTRACE_APPLY_CONTEXT.trace(TraceEvent::Sub(self.apply_ctx.trace.sum()));
     }
 }
 
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -755,7 +755,7 @@ where
     EK: KvEngine,
 {
     fn heap_size(&self) -> usize {
-        // TODO: impl HeapSize for Entry and
+        // TODO: impl HeapSize for Entry.
         let mut size = self.pending_entries.capacity() * mem::size_of::<Entry>()
             + self.pending_msgs.heap_size();
         for e in &self.pending_entries {
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -2835,6 +2835,15 @@ where
     pub must_pass_epoch_check: bool,
 }
 
+impl<S> HeapSize for Proposal<S>
+where
+    S: Snapshot,
+{
+    fn heap_size(&self) -> usize {
+        0
+    }
+}
+
 pub struct Destroy {
     region_id: u64,
     merge_from_snapshot: bool,
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -726,7 +726,7 @@ where
     /// the source peer has applied its logs and pending entries
     /// are all handled.
     pending_msgs: Vec<Msg<EK>>,
-
+    /// Approximate size of current buffer.
     approximate_size: usize,
 }
 
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -13,7 +13,7 @@ use std::sync::mpsc::SyncSender;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use std::vec::Drain;
-use std::{cmp, usize};
+use std::{cmp, mem, usize};
 
 use batch_system::{
     BasicMailbox, BatchRouter, BatchSystem, Fsm, HandlerBuilder, PollHandler, Priority,
@@ -44,6 +44,7 @@ use raft_proto::ConfChangeI;
 use sst_importer::SSTImporter;
 use tikv_alloc::trace::TraceEvent;
 use tikv_util::config::{Tracker, VersionTrack};
+use tikv_util::memory::HeapSize;
 use tikv_util::mpsc::{loose_bounded, LooseBoundedSender, Receiver};
 use tikv_util::time::{duration_to_sec, Instant};
 use tikv_util::worker::Scheduler;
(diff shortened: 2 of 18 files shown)


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3432,6 +3499,19 @@ where
...
             }
         }
+        let s = self.delegate.yield_state.as_mut().map_or(0, |s| {
+            if s.approximate_size == 0 {
+                s.approximate_size = s.heap_size();
+            }
+            s.approximate_size
+        });
+        let trace = ApplyMemoryTrace {
+            pending_cmds: self.delegate.pending_cmds.heap_size(),
```
nolouch, line 3509: Don’t need to track `pending_msgs` and `pending_entries`?

components/raftstore/src/store/peer.rs
```diff
@@ -1017,6 +1018,21 @@ where
         self.raft_group.snap()
     }
 
+    #[inline]
+    pub fn proposal_size(&self) -> usize {
+        self.proposals.queue.heap_size() + self.pending_reads.heap_size()
+    }
+
+    #[inline]
+    pub fn rest_size(&self) -> usize {
```
rleungx, line 1027: What about `pending_messages`?

components/raftstore/src/store/memory.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub static ref MEMTRACE_APPLYS: Arc<dyn MemoryTrace + Send + Sync> =
+        MEMTRACE_ROOT.sub_trace(Id::Name("applys"));
+    pub static ref MEMTRACE_RAFT_CONTEXT: Arc<dyn MemoryTrace + Send + Sync> =
+        MEMTRACE_ROOT.sub_trace(Id::Name("raft_context"));
+    pub static ref MEMTRACE_APPLY_CONTEXT: Arc<dyn MemoryTrace + Send + Sync> =
+        MEMTRACE_ROOT.sub_trace(Id::Name("apply_context"));
+}
+
+#[derive(MemoryTraceHelper, Default)]
+pub struct PeerMemoryTrace {
```
rleungx, line 25: `RouterTrace` is defined in router mod, should we keep the consistent?

components/tikv_util/src/memory.rs
```diff
@@ -0,0 +1,87 @@
...
+use kvproto::{
+    kvrpcpb::LockInfo,
+    metapb::{Region, RegionEpoch},
+    raft_cmdpb::ReadIndexRequest,
+};
+use std::collections::{BTreeMap, VecDeque};
+use std::hash::Hash;
+use std::mem;
+
+pub trait HeapSize {
```
rleungx, line 13: Can we directly use heapsize instead of implementing by ourselves

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3000,6 +3050,23 @@ where
...
 }
 
+impl<EK> HeapSize for Msg<EK>
+where
+    EK: KvEngine,
+{
+    #[inline]
+    fn heap_size(&self) -> usize {
+        match self {
+            Msg::Apply { apply, .. } => apply.entries_mem_size as usize,
```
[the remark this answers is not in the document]
5kbpers: Here we only count the memory in the heap

components/raftstore/src/store/fsm/apply.rs
5kbpers: They are included in `rest_size`

components/raftstore/src/store/peer_storage.rs
```diff
@@ -270,6 +272,12 @@ impl EntryCache {
     }
 
     fn flush_mem_size_change(&mut self) {
+        let event = if self.mem_size_change > 0 {
+            TraceEvent::Add(self.mem_size_change as usize)
+        } else {
+            TraceEvent::Sub(-self.mem_size_change as usize)
```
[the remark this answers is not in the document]
5kbpers: In `TraceEvent::Sub` we need to subtract a positive number.

components/tikv_util/src/memory.rs
5kbpers: Any suggestion?

components/raftstore/src/store/memory.rs
5kbpers: 🤔 I think it would be better to define `RouterTrace` and `Router` in the same place.

components/engine_traits/src/raft_engine.rs
```diff
@@ -90,6 +90,8 @@ pub trait RaftLogBatch: Send {
     fn put_raft_state(&mut self, raft_group_id: u64, state: &RaftLocalState) -> Result<()>;
 
     fn is_empty(&self) -> bool;
+
+    fn data_size(&self) -> usize;
```
hicqu, line 94: Is it necessary? There is only 1 write batch in one poller thread. And it is very small.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2300,6 +2339,7 @@ where
                     );
                 }
                 self.ctx.router.close(new_region_id);
+                MEMTRACE_PEERS.trace(TraceEvent::Sub(self.fsm.trace.sum()));
```
nolouch, line 2342: No need for this?

nolouch (commented):
rest LGTM

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -178,6 +179,16 @@ where
     }
 }
 
+impl<S> HeapSize for PendingCmdQueue<S>
+where
+    S: Snapshot,
+{
+    #[inline]
+    fn heap_size(&self) -> usize {
+        self.normals.capacity() * mem::size_of::<PendingCmd<S>>()
```
rleungx, line 188: I think it's better to sum of all fields' heap size instead of calculating here.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -178,6 +179,16 @@ where
     }
 }
 
+impl<S> HeapSize for PendingCmdQueue<S>
```
hicqu, line 182: Is it possible to move it into `memory.rs`? So that this file can keep clear.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2800,6 +2835,15 @@ where
     pub must_pass_epoch_check: bool,
 }
 
+impl<S> HeapSize for Proposal<S>
```
hicqu, line 2838: Possible to move into `memory.rs`?

components/raftstore/src/store/fsm/apply.rs
hicqu: We can do it later in another PR.

hicqu:
I think we can move all `heap_size` into `memory.rs`, so that all literals like `8`, `7` can be defined as constants. But let us do it later.

components/raftstore/src/store/fsm/apply.rs
hicqu: I think we can write `self.normals.capacity() * (mem::size_of::<PendingCmd<S>>() + 8 /*callback fn point*/)`.

hicqu (approved):

components/raftstore/src/store/fsm/apply.rs
5kbpers: Function pointer is not at the heap.

components/tikv_alloc/src/trace.rs
```diff
@@ -75,13 +76,34 @@ impl Display for Id {
...
+            (TraceEvent::Add(l), TraceEvent::Add(r)) => TraceEvent::Add(l + r),
+            (TraceEvent::Sub(l), TraceEvent::Sub(r)) => TraceEvent::Sub(l + r),
+            (TraceEvent::Add(l), TraceEvent::Sub(r)) | (TraceEvent::Sub(r), TraceEvent::Add(l)) => {
+                if l > r {
+                    TraceEvent::Add(l - r)
+                } else {
+                    TraceEvent::Sub(r - l)
+                }
+            }
+            (TraceEvent::Reset(v), TraceEvent::Sub(r)) => TraceEvent::Reset(v - r),
```
rleungx, line 100: Do we need to ensure v >= r?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -735,6 +750,21 @@ impl Debug for WaitSourceMergeState {
     }
 }
 
+impl<EK> HeapSize for YieldState<EK>
+where
+    EK: KvEngine,
+{
+    fn heap_size(&self) -> usize {
+        // TODO: impl HeapSize for Entry and
```
rleungx, line 758: and?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -713,6 +726,8 @@ where
     /// the source peer has applied its logs and pending entries
     /// are all handled.
     pending_msgs: Vec<Msg<EK>>,
+
+    approximate_size: usize,
```
rleungx, line 730: Please add a comment.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3002,6 +3051,23 @@ where
...
 
+impl<EK> HeapSize for Msg<EK>
+where
+    EK: KvEngine,
+{
+    #[inline]
+    fn heap_size(&self) -> usize {
+        match self {
+            Msg::Apply { apply, .. } => apply.entries_mem_size as usize,
+            Msg::Registration(r) => r.region.heap_size(),
```
rleungx, line 3062: Does `pending_request_snapshot_count` need to be counted?

components/tikv_alloc/src/trace.rs
5kbpers: No, it must be ensured by the caller.

components/raftstore/src/store/fsm/apply.rs
5kbpers: it is included in the trace of `Peer`

rleungx (approved):

hicqu:
I have done a benchmark `sysbench oltp_update_index` for the patch. Here is the result:
图片
图片
图片

With the same QPS results, seems the patch utilizes more CPU for raftstore and apply threads. I think it's a little expensive. Howabout do a large cut for the PR, only pay attation to:

* raft inflight: It's a fixed-length vector, so that memory usage won't change during a raft group's life time.
* entry cache: easily and cheap to update statistics.

And for other components (e.g. peers and applys), we can trigger memory usage statistics by events like https://github.com/tikv/tikv/pull/9019.

What's your opinions? @BusyJay

hicqu:
I have tested it again. Here is the result:
图片

Test 1 and 3 are for raftstore tracing enabled.
Test 2 and 4 are for raftstore tracing disabled.

There is no obvious differences. I think the old test result in https://github.com/tikv/tikv/pull/10251#issuecomment-852819869 is because we didn't run warmup before 2 tests.

PR #10258, opened 2021-05-27
resolved_ts: separate resolevd_ts config from cdc config

NingLin-P:

Problem Summary:

It is weird that the configs used by `resolved-ts` worker are resided in `CdcConfig`.

What's Changed:

Use a dedicated config `ResolvedTsConfig` for `resolved-ts` worker, and support dynamically change of the `resolevd-ts.advance-ts-interval` config.

Tests 

- Unit test

### Release note 
- Add resolved-ts config


components/resolved_ts/src/advance.rs
```diff
@@ -70,32 +59,27 @@ impl<T, E: KvEngine> AdvanceTsWorker<T, E> {
             pd_client,
             worker,
             timer: SteadyTimer::default(),
-            raft_router,
             store_meta,
             concurrency_manager,
-            advance_ts_interval,
-            hibernate_regions_compatible,
```
BusyJay, line 77: Why remove this?

BusyJay (commented):
The change seems break compatibility.

components/resolved_ts/src/advance.rs
NingLin-P: We need `CheckLeader` to sync `safe_ts` from leader to follower, the stale read feature can't work if `hibernate_regions_compatible` is `false`,  so we assume it is `true` if the feature is turned on by `resolved-ts.enable`

src/config.rs
```diff
@@ -2207,16 +2207,13 @@ impl Default for BackupConfig {
     }
 }
 
-// TODO: `CdcConfig` is used by both `cdc` worker and `resolved ts` worker
-// should separate it into two configs
-#[derive(Clone, Serialize, Deserialize, PartialEq, Debug, Configuration)]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct CdcConfig {
```
BusyJay, line 2213: Why keeps the struct?

BusyJay (commented):
Test LGTM as long as compatibility is accepted by cdc.

src/config.rs
NingLin-P: There are configs only use in cdc

BusyJay: Then why remove `Configuration` derive?

src/config.rs
```diff
@@ -2207,16 +2207,13 @@ impl Default for BackupConfig {
...
     pub min_ts_interval: ReadableDuration,
```
BusyJay, line 2214: Is it still a valid configuration given now `advance_ts_interval` is used?

src/config.rs
NingLin-P: `Configuration` trait is used for config that can be changed dynamically, `CdcConfig` is not supported to doing so currently.

src/config.rs
NingLin-P: `advance_ts_interval` and `min_ts_interval` is used by different config

Yisaer (approved):

components/test_raftstore/src/common-test.toml
```diff
@@ -110,5 +110,7 @@ num-threads = 1
 wait-for-lock-timeout = "3s"
 wake-up-delay-duration = "100ms"
 
-[cdc]
```
nolouch, line 113: Does this change break compatibility？

nolouch (commented):
rest LGTM

components/test_raftstore/src/common-test.toml
NingLin-P: No, this config file is used for test, and the `cdc` config is only used for stale read test

components/server/src/server.rs
```diff
@@ -611,16 +611,29 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
-        let rts_scheduler = rts_worker.scheduler();
-
         // Register cdc
         let cdc_ob = cdc::CdcObserver::new(cdc_scheduler.clone());
         cdc_ob.register_to(self.coprocessor_host.as_mut().unwrap());
-        // Register resolved ts
-        let resolved_ts_ob = resolved_ts::Observer::new(rts_scheduler.clone());
-        resolved_ts_ob.register_to(self.coprocessor_host.as_mut().unwrap());
+
+        // Create resolved ts worker
```
5kbpers, line 618: Could you leave a TODO for adding a cdc config manager here?

5kbpers (approved):

nolouch (commented):
LGTM. plz update the release note

PR #10260, opened 2021-05-28
server: Replace RocksEngine with KvEngine typaram

brson:

Issue Number: https://github.com/tikv/tikv/issues/6402

Problem Summary:

tikv::server is one of the last modules that is still concretely using RocksEngine in a way that prevents tikv from being started with another engine.

Just replacing more uses of RocksEngine with KvEngine typarams in tikv::server.

This unfortunately requires type-parameterizing GcTask, which in turn forces some new type parameters elsewhere. But it's the only way I know to proceed.

### Release note 

```release-note
None
```

### DIFF
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -195,16 +195,15 @@ struct TiKVEngines<ER: RaftEngine> {
 
 struct Servers<ER: RaftEngine> {
     lock_mgr: LockManager,
-    server: Server<
-        RaftRouter<RocksEngine, ER>,
-        resolve::PdStoreAddrResolver,
-        RaftKv<RocksEngine, ServerRaftStoreRouter<RocksEngine, ER>>,
-    >,
+    server: ServerType<ER>,
     node: Node<RpcClient, ER>,
     importer: Arc<SSTImporter>,
     cdc_scheduler: tikv_util::worker::Scheduler<cdc::Task>,
     cdc_memory_quota: MemoryQuota,
 }
+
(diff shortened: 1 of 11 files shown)


brson:
Rebased.

brson:
Rebased.

brson:
Rebased.

kennytm (approved):

brson:
Thanks for the review @kennytm

brson:
Rebased.

components/server/src/server.rs
```diff
@@ -195,13 +195,17 @@ struct TiKVEngines<ER: RaftEngine> {
...
     lock_mgr: LockManager,
-    server: Server<RaftRouter<RocksEngine, ER>, resolve::PdStoreAddrResolver>,
+    server: ServerType<ER>,
     node: Node<RpcClient, ER>,
     importer: Arc<SSTImporter>,
     cdc_scheduler: tikv_util::worker::Scheduler<cdc::Task>,
     cdc_memory_quota: MemoryQuota,
 }
 
+type ServerType<ER> =
```
tabokie, line 205: Use a different name than `xxType`? I think we usually use that for enum.

tabokie (approved):
LGTM

components/server/src/server.rs
brson: I pushed a change to rename them to LocalServer and LocalRaftKv. Does that work? Hard to come up with a good name for these since they are just here to reduce type complexity.

tabokie: Sounds good.

Issue #10114, opened 2021-05-06
CDC unlimited scan task concurrency may cause OOM


overvenus:
## Bug Report

CDC unlimited scan task concurrency may cause OOM. There is a speed limit of sending messages in TiKV, so scanned content is buffered in memory and causes memory spikes.

https://github.com/tikv/tikv/blob/e26389a278116b2f61addfa9f15ca25ecf38bc80/components/cdc/src/endpoint.rs#L1185-L1204

![image](https://user-images.githubusercontent.com/2150711/117233649-1fde1280-ae56-11eb-8e0e-f77041c9c634.png)
![image](https://user-images.githubusercontent.com/2150711/117233666-2c626b00-ae56-11eb-8b0a-d7cb5eb40d3e.png)

### What version of TiKV are you using?

TiKV v5.0.0

### Steps to reproduce

```
# create a changefeed for incremental scan large amount of data in 2 hours
sleep 7200 && tiup cdc:nightly cli --pd="http://172.16.4.55:47902" changefeed create --sink-uri="blackhole://" -c "bl2" --config cf-old-value.toml --start-ts=$(( $(pd-ctl -u http://172.16.4.55:47902 service-gc-safepoint | jq '.service_gc_safe_points[] | select(.service_id == "gc_worker") | .safe_point') + 1000 )) &

mysql -h 172.16.5.33 -P 47904 -u root -E -e 'update mysql.tidb set VARIABLE_VALUE="2400h" where VARIABLE_NAME="tikv_gc_life_time";' && \
sleep 60 && tiup cdc:nightly cli --pd="http://172.16.5.33:47902" changefeed create --sink-uri="blackhole://" -c "bl1" --config cf-old-value.toml && \
sleep 60 && go-ycsb load mysql -p workload=core -p mysql.host=172.16.5.33 -p mysql.port=47904 -p mysql.user=root -p recordcount=600000000 -p fieldcount=10 -p fieldlength=100 -p threadcount=8 -p table=usertable1 && \
sleep 60 && go-ycsb run  mysql -p workload=core -p mysql.host=172.16.5.33 -p mysql.port=47904 -p mysql.user=root -p recordcount=600000000 -p fieldcount=10 -p fieldlength=100 -p threadcount=8 -p table=usertable1 -p operationcount=1000000000 -p updateproportion=1 -p requestdistribution=uniform -p dotransactions=true
```

### What did you expect?

No memory spike.

lonng:
[...]

PR #10262, opened 2021-05-28
cdc: limit scan task concurrency

overvenus:

Close https://github.com/tikv/tikv/issues/10114
Close https://github.com/pingcap/ticdc/issues/1844
Problem Summary:

Limit CDC scan concurrency.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

Tested with TiCDC v4.0.13

<b style="font-weight:normal;" ><div dir="ltr" style="margin-left:0pt;" align="left">
v5.0.2 | This PR
-- | --
image  | image
image | image
image | image

</div></b>

### Release note 

```release-note
Fix OOM when TiCDC resumes a changefeed in a cluster that has lots of regions being captured. 
```


components/cdc/src/endpoint.rs
```diff
@@ -549,30 +552,38 @@ impl<T: 'static + RaftStoreRouter<RocksEngine>> Endpoint<T> {
...
-            },
-        ) {
-            warn!("cdc send capture change cmd failed"; "region_id" => region_id, "error" => ?e);
-            deregister_downstream(Error::request(e.into()));
-            return;
-        }
+        let scan_concurrency_semaphore = self.scan_concurrency_semaphore.clone();
         self.workers.spawn(async move {
+            let _permit = scan_concurrency_semaphore.acquire().await;
+
```
liuzix, line 560: Should we test for request cancellation here? If not, what would happen if the client sends 10000 requests and cancels the connection immediately, and then another client retries? Would the second client encounter a large latency to receiving data? Is this behavior risky?

components/cdc/src/endpoint.rs
```diff
@@ -1044,58 +1015,86 @@ struct Initializer {
...
-                }
-            } else {
-                Deregister::Downstream {
-                    region_id: self.region_id,
-                    downstream_id: self.downstream_id,
-                    conn_id: self.conn_id,
-                    err: Some(err),
-                }
-            }
+            self.async_incremental_scan(region_snapshot, region, require_barrier)
```
liuzix, line 1088: Does `require_barrier` have to be true here?

components/cdc/src/endpoint.rs
```diff
@@ -1253,6 +1252,28 @@ impl Initializer {
             error!("cdc schedule task failed"; "error" => ?e);
         }
     }
+
+    fn deregister_downstream(&self, err: Error) {
+        let deregister = if self.build_resolver || matches!(err, Error::Request(_)) {
```
liuzix, line 1257: Can you add some comments on the conditions used here?

components/cdc/src/endpoint.rs
```diff
@@ -1044,58 +1015,86 @@ struct Initializer {
...
+    async fn initialize<T: 'static + RaftStoreRouter<RocksEngine>>(
+        &mut self,
+        change_cmd: ChangeObserver,
+        raft_router: T,
+        concurrency_semaphore: Arc<Semaphore>,
+    ) -> Result<()> {
+        let _permit = concurrency_semaphore.acquire().await;
+
+        // The initialization maybe canceled after acquiring a permit.
+        if self.downstream_state.load() == DownstreamState::Stopped {
```
liuzix, line 1027: Can you add a comment on where `downstream_state` is written to?

components/cdc/src/endpoint.rs
overvenus: Do you mean comments in line 1258?

components/cdc/src/endpoint.rs
overvenus: Yes, in the current implementation it has be to true. It's false only in tests that do not care about the barrier. This parameter can be removed after the segmentation scan is supported.

components/cdc/src/endpoint.rs
overvenus: Yes, in the current implementation it has be to true. It's false only in tests that do not care about the barrier. This parameter can be removed after the segmentation scan is supported.

components/cdc/src/endpoint.rs
liuzix: It would be better if you could write something on how/when to use this function and how it is related to the conditions used here. I feel this could benefit those who would like to rewrite/modify this file in the future.

components/cdc/src/endpoint.rs
liuzix: > Yes, in the current implementation it has be to true. It's false only in tests that do not care about the barrier. This parameter can be removed after the segmentation scan is supported.

Ok. Can you add a TODO?

liuzix (approved):

src/config.rs
```diff
@@ -2234,6 +2240,30 @@ impl Default for CdcConfig {
...
+        }
+        if self.incremental_scan_concurrency == 0 {
+            return Err("cdc.incremental-scan-concurrency can't be 0".into());
+        }
+        if self.incremental_scan_concurrency < self.incremental_scan_threads {
+            return Err(
+                "cdc.incremental-scan-concurrency must be larger than cdc.incremental-scan-threads"
+                    .into(),
+            );
+        }
```
amyangfei, line 2262: No need to check ` self.incremental_scan_concurrency == 0` since `self.incremental_scan_concurrency < self.incremental_scan_threads` contains it

amyangfei (approved):

PR #10263, opened 2021-05-28
evict entry cache when memory usage is high

hicqu:

With https://github.com/tikv/tikv/pull/10236, appy threads and entry cache could share some `bytes::Bytes`. It makes memory usage incorrect. This PR corrects it.

And evict entry cache if memory usage reaches high-water mark.

Sending `CachedEntries` instead of `Vec<Entry>` to apply threads. So after entry cache is compacted, `CachedEntries` can be updated to release shared buffer.

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```


components/raftstore/src/store/memory.rs
```diff
@@ -51,3 +54,16 @@ pub struct ApplyMemoryTrace {
     pub pending_cmds: usize,
     pub rest: usize,
 }
+
+pub fn needs_evict_entry_cache() -> bool {
+    fail_point!("needs_evict_entry_cache", |_| true);
+    if memory_usage_reaches_high_water() {
+        let usage = get_global_memory_usage();
+        let ec_usage = MEMTRACE_ENTRY_CACHE.sum() as u64;
+        if ec_usage > GIB || ec_usage > usage / 3 {
```
rleungx, line 63: I think most of the time it will trigger the first condition. Is it GIB too small?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3175,13 +3158,24 @@ where
...
         }
 
+        let mut entries = apply.entries.take_entries();
+        if entries.is_empty() {
+            let rid = self.delegate.region_id();
+            let StdRange { start, end } = apply.entries.range;
+            entries = Vec::with_capacity((end - start) as usize);
+            self.delegate
+                .raft_engine
+                .fetch_entries_to(rid, start, end, None, &mut entries)
```
nolouch, line 3172: why the entries are empty but the range is not.

components/raftstore/src/store/config.rs
```diff
@@ -207,7 +207,7 @@ impl Default for Config {
             raft_max_size_per_msg: ReadableSize::mb(1),
             raft_max_inflight_msgs: 256,
             raft_entry_max_size: ReadableSize::mb(8),
-            raft_log_gc_tick_interval: ReadableDuration::secs(10),
+            raft_log_gc_tick_interval: ReadableDuration::secs(1),
```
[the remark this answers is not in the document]
hicqu: If it's not configured, it will become `1`.

components/raftstore/src/store/fsm/apply.rs
hicqu: When cached raft entries are deleted because of no memory, fetched committed entries will also be cleaned to release memory.

components/raftstore/src/store/config.rs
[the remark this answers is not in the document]
5kbpers: A smaller raftlog gc interval may cause more kvdb sync, could hurt performance.🤔

etc/config-template.toml
```diff
@@ -536,7 +536,7 @@
 ## default: false
 # enabled = false
 
-## Maximum number of threads of `Titan` background gc jobs.
+## Maximum number of threads of `Titan` background ac jobs.
```
rleungx, line 539:
```suggestion
## Maximum number of threads of `Titan` background gc jobs.
```

components/raftstore/src/store/fsm/apply.rs
rleungx: Will it affect the performance?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -202,86 +193,74 @@ impl EntryCache {
...
-            .map(|e| (bytes_capacity(&e.data) + bytes_capacity(&e.context)) as i64)
-            .sum::<i64>();
-    }
-
-    fn get_cache_vec_mem_size_change(&self, new_capacity: i64, old_capacity: i64) -> i64 {
-        ENTRY_MEM_SIZE as i64 * (new_capacity - old_capacity)
-    }
-
-    fn get_total_mem_size(&self) -> i64 {
+    fn get_total_mem_size(&self) -> u64 {
```
rleungx, line 243: Do we need to consider the allocated memory of `trace`?

nolouch: +1, you comment `It's already counted when building` in Apply Msg, but I cannot get the place.

components/raftstore/src/store/fsm/apply.rs
hicqu: Yes because it drops entry cache but load entries from storage. We can't avoid it.

hicqu:
图片
With the patch and set `memory-usage-limit` to `6GB`, entry cache will be limit to about `2GB` by the cache evict mechanism when 1 replica is shutted down.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -163,34 +164,30 @@ impl EntryCache {
...
-        if entries.is_empty() {
-            return;
+        if !entries.is_empty() {
+            let mut mem_size_change = 0;
+            let old_capacity = self.cache.capacity();
+            self.append_impl(tag, entries);
+            let new_capacity = self.cache.capacity();
+            mem_size_change += Self::get_cache_vec_mem_size_change(new_capacity, old_capacity);
+            mem_size_change += self.shrink_if_necessary();
+            Self::flush_mem_size_change(mem_size_change);
```
rleungx, line 174: flush two times in one append?

components/tikv_util/src/sys/mod.rs
```diff
@@ -78,15 +80,33 @@ impl SysQuota {
...
+    GLOBAL_MEMORY_USAGE.store(usage as u64, Ordering::Release);
+}
+
+#[cfg(not(target_os = "linux"))]
 pub fn record_global_memory_usage() {
-    let mut system = sysinfo::System::new();
-    system.refresh_memory();
-    let usage = system.get_used_memory() * KIB;
-    GLOBAL_MEMORY_USAGE.store(usage, Ordering::Release);
+    GLOBAL_MEMORY_USAGE.store(0, Ordering::Release);
```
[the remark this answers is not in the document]
hicqu: Yes. So that all logics about memory evicting won't make effects.

components/raftstore/src/store/peer_storage.rs
hicqu: It's OK because `flush_mem_size_change` only changes one atomic number.

rleungx: So what is the purpose of flushing in append_impl?

hicqu: However it's better to change it to `fn append_impl(...) -> i64`.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3559,6 +3569,20 @@ where
         PEER_GC_RAFT_LOG_COUNTER.inc_by(total_gc_logs as i64);
     }
 
+    fn register_entry_cache_evict_tick(&mut self) {
+        self.schedule_tick(PeerTicks::ENTRY_CACHE_EVICT)
+    }
+
+    fn on_entry_cache_evict_tick(&mut self) {
```
5kbpers, line 3576: Could you add some comments for this tick?

5kbpers (approved):
rest LGTM

rleungx (approved):
LGTM

PR #10264, opened 2021-05-28
*: support read only and recovery when disk full

tier-cap:
Solve the availability problem when the disk is full, no mather one or two or the majority or all.

1. Regularly monitor disk usage
2. Intercept business write traffic when the disk threshold is triggered
3. Space reclamation relies on PD for free space reclamation logic

### Check List
Test

### Release note
* not available


components/server/src/server.rs
```diff
@@ -937,6 +943,83 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
             });
     }
+    fn init_storage_stats_task(&self, engines: Engines<RocksEngine, ER>) {
+        let flag = self.write_permission.clone();
+        let config_disk_capacity: u64 = self.config.raft_store.capacity.0;
+        let store_path = self.store_path.clone();
+        let snap_path = self.snap_path.clone();
+        self.background_worker
+            .spawn_interval_task(DEFAULT_STORAGE_STATS_INTERVAL, move || {
+                let disk_stats = match fs2::statvfs(&store_path) {
```
hicqu, line 953: Why not reuse code in `components/raftstore/src/store/worker/pd.rs`?

components/raftstore/src/store/snap.rs
```diff
@@ -66,8 +66,8 @@ const DEL_RANGE_PREFIX: &str = "del_range";
 
 const TMP_FILE_SUFFIX: &str = ".tmp";
 const SST_FILE_SUFFIX: &str = ".sst";
-const CLONE_FILE_SUFFIX: &str = ".clone";
-const META_FILE_SUFFIX: &str = ".meta";
+pub const CLONE_FILE_SUFFIX: &str = ".clone";
```
hicqu, line 69: You can call `get_total_snap_size` directly.

components/server/src/server.rs
```diff
@@ -88,6 +90,8 @@ use tokio::runtime::Builder;
 use crate::raft_engine_switch::{check_and_dump_raft_db, check_and_dump_raft_engine};
 use crate::{setup::*, signal_handler};
 
+const GBSIZE: u64 = 1024 * 1024 * 1024;
```
hicqu, line 93: It's already defined as `GIB`.

components/server/src/server.rs
```diff
@@ -162,6 +170,7 @@ struct TiKVServer<ER: RaftEngine> {
     concurrency_manager: ConcurrencyManager,
     env: Arc<Environment>,
     background_worker: Worker,
+    write_permission: Arc<Mutex<bool>>,
```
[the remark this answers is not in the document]
hicqu: You can take a look at this: https://github.com/tikv/tikv/pull/10231

components/server/src/server.rs
```diff
@@ -937,6 +943,83 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+                            "kv path"=>store_path.to_str(),
+                            "err"=>?e
+                        );
+                        return;
+                    }
+                    Ok(stats) => stats,
+                };
+                let disk_cap = disk_stats.total_space();
+                let disk_available = disk_stats.available_space();
+                let snap_size = (|| -> Option<u64> {
```
hicqu, line 966: Why define som many closures?

components/server/src/server.rs
tier-cap: My first method is to reuse the logic and introduce a global variable to be updated, but my colleagues do not recommend this style of introducing an additional global variable.

tier-cap: first commit reuse pd.rs, but removed for some advice. now, it has been repushed.

components/raftstore/src/store/snap.rs
tier-cap: changed to reuse the pd.rs store heartbeat logic.

components/server/src/server.rs
tier-cap: removed

components/server/src/server.rs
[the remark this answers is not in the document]
tier-cap: The first version is to directly introduce global variables into the pd heartbeat logic, and it has been changed back.

components/raftstore/src/store/snap.rs
hicqu: Why not remove `pub`?

src/server/service/kv.rs
```diff
@@ -625,8 +627,27 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
                         my_store_id: store_id,
                     }))
                 } else {
-                    let ret = ch.send_raft_msg(msg).map_err(Error::from);
-                    future::ready(ret)
+                    let mut flag = true;
+                    if !disk::WRITE_PERMISSION.load(Ordering::Acquire) {
```
hicqu, line 631: It's better to move the check into `raftstore` before `step`.

hicqu:
store_heartbeat_tick_interval is `10s` by default. It's too large for disk stat. You can add a new tick type to do disk stat per second.

components/raftstore/src/store/snap.rs
tier-cap: The current implementation version no longer needs to use the global variable, just remove pub and keep the original non-pub mode

components/raftstore/src/store/peer.rs
```diff
@@ -2278,7 +2279,13 @@ where
                 return false;
             }
             Ok(RequestPolicy::ReadIndex) => return self.read_index(ctx, req, err_resp, cb),
-            Ok(RequestPolicy::ProposeNormal) => self.propose_normal(ctx, req),
+            Ok(RequestPolicy::ProposeNormal) => {
+                if disk::WRITE_PERMISSION.load(Ordering::Acquire) {
```
hicqu, line 2283: Better to allow configuration change on leader.

hicqu: And you need to write function
```rust
fn disk_almost_full() -> bool
```
then add failpoint in it.

components/server/src/server.rs
```diff
@@ -152,6 +156,7 @@ struct TiKVServer<ER: RaftEngine> {
     resolver: resolve::PdStoreAddrResolver,
     state: Arc<Mutex<GlobalReplicationState>>,
     store_path: PathBuf,
+    //raft_path: Pathbuf, currently not impl.
```
hicqu, line 159: Why not remove it?

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,203 @@
...
+    let right_value = String::from("77770000").into_bytes();
+    cluster.must_put(&right_key, &right_value);
+    cluster.must_get(&right_key);
+}
+
+fn fail_split_region(cluster: &mut Cluster<ServerCluster>) {
+    let split_key = String::from("0000").into_bytes();
+    let region = cluster.get_region(&split_key);
+    let split_count_before = cluster.pd_client.get_split_count();
+    cluster.split_region(&region, &split_key, Callback::None);
```
hicqu, line 45: Better to do tests in `Callback`. In the current implementation, we can't know whether split fails or it just is waitting for executing.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,203 @@
...
+    assert!(split_count_before == split_count_after);
+}
+
+fn fail_merge_region(cluster: &mut Cluster<ServerCluster>) {
+    let left_key = String::from("5555").into_bytes();
+    let right_key = String::from("7777").into_bytes();
+    let region1 = cluster.get_region(&left_key);
+    let region2 = cluster.get_region(&right_key);
+    let resp = cluster.try_merge(region1.get_id(), region2.get_id());
+    assert!(is_error_response(&resp));
```
hicqu, line 60: Better to extract and test the error code.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,203 @@
...
+
+[minority full]
+1. not allowed: log entry append
+2. allowed: config change,transfer leader,etc.
+
+[majority full]
+1. equal to minority full, no additional design.
+
+*/
+const DISK_FULL_LEADER: &str = "disk_full_leader";
```
hicqu, line 23: Better to rename `leader`, `follower_1` to `peer_1`, `peer_2`, ...

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,203 @@
...
+    fail::cfg(DISK_FULL_LEADER, "return").unwrap();
+
+    let leader = cluster.leader_of_region(1).unwrap();
+    let raft_stat = cluster.raft_local_state(1, leader.get_store_id());
+    let index_1 = raft_stat.get_last_index();
+
+    let key = String::from("a001").into_bytes();
+    let value = String::from("a001").into_bytes();
+    let rx = cluster.async_put(&key, &value).unwrap();
+    match rx.recv_timeout(Duration::from_secs(10)) {
```
hicqu, line 73: Why not call `unwrap` to simplify the code?

src/server/service/kv.rs
```diff
@@ -625,8 +626,31 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
                         my_store_id: store_id,
                     }))
                 } else {
-                    let ret = ch.send_raft_msg(msg).map_err(Error::from);
-                    future::ready(ret)
+                    let mut flag = true;
+                    let disk_full_precheck = || {
```
hicqu, line 630: Please write it as a function so that all places can reuse it.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,203 @@
...
+    let region1 = cluster.get_region(&left_key);
+    let region2 = cluster.get_region(&right_key);
+    let resp = cluster.try_merge(region1.get_id(), region2.get_id());
+    assert!(is_error_response(&resp));
+}
+
+fn fail_leader_full(cluster: &mut Cluster<ServerCluster>) {
+    fail::cfg(DISK_FULL_LEADER, "return").unwrap();
+
+    let leader = cluster.leader_of_region(1).unwrap();
```
hicqu, line 66: It's better to do a manual leader transferring to transfer the leader ship to store 1.

src/server/service/kv.rs
```diff
@@ -666,9 +690,31 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
                             my_store_id: store_id,
                         }));
                     }
-                    if let Err(e) = ch.send_raft_msg(msg) {
-                        return future::err(Error::from(e));
+                    let mut flag = true;
+                    let disk_full_precheck = || {
+                        if cfg!(feature = "failpoints") {
+                            fail_point!("disk_full_follower_1", store_id == 1, |_| {
```
hicqu, line 696: It's possible that store 1 is leader. Actually for lots of test cases, store 1 is the leader.

hicqu: And you need to move the check into raftstore, before `step`.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,203 @@
...
+        new_leader.get_store_id(),
+        new_leader.get_id()
+    );
+
+    let pd_client = cluster.pd_client.clone();
+    pd_client.must_remove_peer(region.get_id(), old_leader.clone());
+    let region = cluster.get_region(&key);
+    assert!(region.get_peers().len() == 2);
+    //TODO add peer case pass?
+    //let peer_3 = new_peer(old_leader.get_store_id(), 2000);
```
hicqu, line 132: There is already a peer on the old leader. The add must fail.

src/server/service/kv.rs
tier-cap: > It's possible that store 1 is leader. Actually for lots of test cases, store 1 is the leader.
1. I have set the leader not 1 and 2.

> And you need to move the check into raftstore, before `step`.
1. to filter in batch_raft will save all the msg process resourse.
2. So, my problem is why we should move to step?

components/raftstore/src/store/snap.rs
tier-cap: removed

tier-cap:
> store_heartbeat_tick_interval is `10s` by default. It's too large for disk stat. You can add a new tick type to do disk stat per second.

changed.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1166,6 +1167,13 @@ where
             "to_peer_id" => msg.get_to_peer().get_id(),
         );
 
+        let msg_type = msg.get_message().get_msg_type();
+        let store_id = self.ctx.store_id();
+        if disk::follower_disk_full_precheck(store_id) || disk::is_disk_full() {
```
hicqu, line 1172: s/follower_disk_full_precheck/disk_full_precheck/g

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1522,6 +1525,13 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
 
     fn on_raft_message(&mut self, mut msg: RaftMessage) -> Result<()> {
         let region_id = msg.get_region_id();
+        let msg_type = msg.get_message().get_msg_type();
```
hicqu, line 1528: This place is wrong. The correct place is `PeerFsmDelegate::on_raft_message`.

components/raftstore/src/store/peer.rs
```diff
@@ -2278,7 +2279,13 @@ where
                 return false;
             }
             Ok(RequestPolicy::ReadIndex) => return self.read_index(ctx, req, err_resp, cb),
-            Ok(RequestPolicy::ProposeNormal) => self.propose_normal(ctx, req),
+            Ok(RequestPolicy::ProposeNormal) => {
+                if disk::leader_disk_full_precheck() || disk::is_disk_full() {
```
hicqu, line 2283: Should be `disk_full_precheck`.

components/raftstore/src/store/peer.rs
```diff
@@ -2928,6 +2935,7 @@ where
         poll_ctx: &mut PollContext<EK, ER, T>,
         mut req: RaftCmdRequest,
     ) -> Result<Either<u64, u64>> {
+        //when disk full, propose normal will never be called.
```
hicqu, line 2938: It should be removed.

components/server/src/server.rs
```diff
@@ -938,6 +949,58 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+        let store_path = self.store_path.clone();
+        let snap_mgr = self.snap_mgr.clone().unwrap();
+        self.background_worker
+            .spawn_interval_task(DEFAULT_STORAGE_STATS_INTERVAL, move || {
+                let disk_stats = match fs2::statvfs(&store_path) {
+                    Err(e) => {
+                        error!(
+                            "get disk stat for kv store failed";
+                            "kv path"=>store_path.to_str(),
+                            "err"=>?e
```
hicqu, line 963: Needs to add spaces surround `=>`.

components/server/src/server.rs
```diff
@@ -938,6 +949,58 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+                // We only care about rocksdb SST file size, so we should check disk available here.
+                available = cmp::min(available, disk_stats.available_space());
+                if available <= disk::get_disk_reserved() {
+                    warn!(
+                        "disk full, available={},snap={},engine={},capacity={}",
+                        available, snap_size, kv_size, capacity
+                    );
+                    disk::set_disk_full();
+                } else {
+                    // warn!(
```
hicqu, line 995: It can be removed.

components/tikv_util/src/sys/disk.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+use fail::fail_point;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+pub static DISK_FULL: AtomicBool = AtomicBool::new(false);
```
hicqu, line 4: Shouldn't be `pub`.

components/tikv_util/src/sys/disk.rs
```diff
@@ -0,0 +1,52 @@
...
+pub fn clear_disk_full() {
+    DISK_FULL.store(false, Ordering::Release);
+}
+
+pub fn is_disk_full() -> bool {
+    DISK_FULL.load(Ordering::Acquire)
+}
+
+//underscore to disable unused warning
+pub fn follower_disk_full_precheck(_store_id: u64) -> bool {
```
hicqu, line 20: Should rename to `disk_full_precheck`. And s/disk_full_follower_1/disk_full_store_1

components/tikv_util/src/sys/disk.rs
```diff
@@ -0,0 +1,52 @@
...
+    if cfg!(feature = "failpoints") {
+        fail_point!("disk_full_follower_1", _store_id == 1, |_| {
+            return true;
+        });
+        return false;
+    } else {
+        return false;
```
hicqu, line 27: Suggest to s/return false;/false. Or I think clippy will warn it.

components/tikv_util/src/sys/disk.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+pub fn set_disk_reserved(v: u64) {
+    let mut value = v;
+    if v == 0 {
+        value = 5 * 1024 * 1024 * 1024;
```
hicqu, line 45: It's better to write the logic outside. For example, `Config::validate`. And I believe the logic exists already.

src/server/service/kv.rs
```diff
@@ -626,7 +625,7 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
                     }))
                 } else {
                     let ret = ch.send_raft_msg(msg).map_err(Error::from);
-                    future::ready(ret)
+                    return future::ready(ret);
```
hicqu, line 628: It can be reverted.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+1. not allowed: log entry append
+2. allowed: config change,transfer leader,etc.
+
+[majority full]
+1. equal to minority full, no additional design.
+*/
+const DISK_FULL_LEADER: &str = "disk_full_leader";
+const DISK_FULL_FOLLOWER_1: &str = "disk_full_follower_1";
+
+fn prepare_test_data(cluster: &mut Cluster<ServerCluster>) {
```
hicqu, line 24: It's unnecessary and can be removed.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    let leader = cluster.leader_of_region(1).unwrap();
+
+    {
+        let raft_stat = cluster.raft_local_state(1, leader.get_store_id());
+        let index_1 = raft_stat.get_last_index();
+
+        let key = String::from("8000").into_bytes();
+        let value = String::from("8000").into_bytes();
+        let rx = cluster.async_put(&key, &value).unwrap();
+        is_error_response(&rx.recv_timeout(Duration::from_secs(10)).unwrap());
```
hicqu, line 50: Should test the error code.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    {
+        let raft_stat = cluster.raft_local_state(1, leader.get_store_id());
+        let index_1 = raft_stat.get_last_index();
+
+        let split_key = String::from("0000").into_bytes();
+        let region = cluster.get_region(&split_key);
+        let split_count_before = cluster.pd_client.get_split_count();
+        cluster.split_region(&region, &split_key, Callback::None);
```
hicqu, line 64: Should test the error code in `Callback` instead of comparing region counts before and after the split.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+        let split_count_after = cluster.pd_client.get_split_count();
+        // println!(
```
hicqu, line 66: Can be removed.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+        assert!(index_1 == index_4);
+    }
+
+    fail::remove(DISK_FULL_LEADER);
+}
+
+fn success_leader_full(cluster: &mut Cluster<ServerCluster>) {
+    let key = String::from("5555").into_bytes();
+    fail::cfg(DISK_FULL_LEADER, "return").unwrap();
+    cluster.must_get(&key).unwrap();
```
hicqu, line 100: What is this for?

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    let old_leader = cluster.leader_of_region(region.get_id()).unwrap();
+    let peers = region.get_peers();
+
+    let target = peers
+        .into_iter()
+        .find(|x| x.get_store_id() != old_leader.get_store_id())
+        .unwrap();
+
+    cluster.must_transfer_leader(region.get_id(), (*target).clone());
+    let new_leader = cluster.leader_of_region(region.get_id()).unwrap();
```
hicqu, line 112: `must_transfer_leader` already contains the compare logic.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    assert!(
+        new_leader.get_store_id() != old_leader.get_store_id()
+            && new_leader.get_id() != old_leader.get_id()
+    );
+
+    let pd_client = cluster.pd_client.clone();
+    pd_client.must_remove_peer(region.get_id(), old_leader.clone());
+    let region = cluster.get_region(&key);
+    assert!(region.get_peers().len() == 2);
```
hicqu, line 121: `must_remove_peer` already contains the logic.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    let peer_3 = new_learner_peer(old_leader.get_store_id(), 20000);
+    pd_client.must_add_peer(region.get_id(), peer_3); //can not!
```
hicqu, line 124: Stale comments.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    // for i in region.get_peers() {
+    //     println!("peer info: {}-{}", i.get_store_id(), i.get_id());
+    // }
+    assert!(region.get_peers().len() == 3);
+    //failed to propose confchange, error: Other("[components/raftstore/src/store/peer.rs:2469]: [region 1000] 1002 unsafe to perform conf change [peer { id: 2000 store_id: 1 }], before: voters: 1004 voters: 1002 voters: 1003, after: voters: 2000 voters: 1002 voters: 1004 voters: 1003, truncated index 5, promoted commit index 0")
+    fail::remove(DISK_FULL_LEADER);
+}
+
+fn fail_follower_full(cluster: &mut Cluster<ServerCluster>) {
+    std::thread::sleep(Duration::from_secs(2));
```
hicqu, line 136: What is this line for?

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+        .find(|x| x.get_store_id() == 1)
+        .unwrap();
+    let follower2 = region
+        .get_peers()
+        .iter()
+        .find(|x| x.get_store_id() == 2)
+        .unwrap();
+
+    fail::cfg(DISK_FULL_FOLLOWER_1, "return").unwrap();
+    assert!(leader.get_store_id() == 3);
```
hicqu, line 163: Useless compares.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+        loop {
+            cluster.reset_leader_of_region(region.get_id());
+            let cur_leader = cluster.leader_of_region(region.get_id());
+            if let Some(ref cur_leader) = cur_leader {
+                assert!(
+                    !(cur_leader.get_id() == follower1.get_id()
+                        && cur_leader.get_store_id() == follower1.get_store_id())
+                );
+            }
+            if timer.elapsed() > Duration::from_secs(2) {
```
hicqu, line 186: So this part will keep 2 seconds? It's too long. We can write code like this:
```
cluster.transfer_leader(target);
cluster.async_command_on_peer(Put, ...);
assert(put_success);
```

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+            }
+            cluster.transfer_leader(region.get_id(), (*follower1).clone());
+        }
+    }
+    fail::remove(DISK_FULL_FOLLOWER_1);
+}
+
+#[test]
+fn test_disk_full() {
+    let reserve = 1024 * 1024 * 100;
```
hicqu, line 198: Unnecessary.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    let mut cluster = new_server_cluster(0, 3);
+    cluster.cfg.storage.reserve_space = ReadableSize(reserve);
+    cluster.cfg.raft_store.pd_store_heartbeat_tick_interval =
+        ReadableDuration(Duration::from_secs(3000)); //disable disk status update influence.
+    cluster.run();
```
hicqu, line 203: `run` will bootstrap a region with id `1`, and peers (1, 1), (2, 2) and `(3, 3)` in the cluster. You can use them directly instead of using `region.get_peers().find(...)`.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,209 @@
...
+    cluster.pd_client.disable_default_operator();
```
hicqu, line 204: It should be called before `run` or `run_conf_change`.

components/raftstore/src/store/peer.rs
tier-cap: done

components/raftstore/src/store/fsm/peer.rs
tier-cap: done

components/raftstore/src/store/peer.rs
tier-cap: done

components/raftstore/src/store/fsm/store.rs
tier-cap: PeerFsmDelegate has been dealed with. not sure now about store delegate, but will router.send()， ignore maybe ok

components/server/src/server.rs
tier-cap: done

components/tikv_util/src/sys/disk.rs
tier-cap: done

components/raftstore/src/store/peer.rs
```diff
@@ -2341,7 +2342,13 @@ where
                 return false;
             }
             Ok(RequestPolicy::ReadIndex) => return self.read_index(ctx, req, err_resp, cb),
-            Ok(RequestPolicy::ProposeNormal) => self.propose_normal(ctx, req),
+            Ok(RequestPolicy::ProposeNormal) => {
+                if disk::disk_full_precheck(0) || disk::is_disk_full() {
```
hicqu, line 2346: Why 0 instead of store_id?

components/server/src/server.rs
```diff
@@ -1004,6 +1019,58 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+                // We only care about rocksdb SST file size, so we should check disk available here.
+                available = cmp::min(available, disk_stats.available_space());
+                if available <= disk::get_disk_reserved() {
+                    warn!(
+                        "disk full, available={},snap={},engine={},capacity={}",
+                        available, snap_size, kv_size, capacity
+                    );
+                    disk::set_disk_full();
+                } else {
+                    warn!(
```
hicqu, line 1065: The log can be removed.

components/server/src/server.rs
```diff
@@ -1004,6 +1019,58 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+        let config_disk_capacity: u64 = self.config.raft_store.capacity.0;
+        let store_path = self.store_path.clone();
+        let snap_mgr = self.snap_mgr.clone().unwrap();
+        self.background_worker
+            .spawn_interval_task(DEFAULT_STORAGE_STATS_INTERVAL, move || {
+                let disk_stats = match fs2::statvfs(&store_path) {
+                    Err(e) => {
+                        error!(
+                            "get disk stat for kv store failed";
+                            "kv path"=>store_path.to_str(),
```
hicqu, line 1032: Please add spaces surround `=>`.

src/server/service/kv.rs
```diff
@@ -46,7 +46,6 @@ use tikv_util::future::{paired_future_callback, poll_future_notify};
 use tikv_util::mpsc::batch::{unbounded, BatchCollector, BatchReceiver, Sender};
 use tikv_util::worker::Scheduler;
 use txn_types::{self, Key};
-
```
hicqu, line 49: It's better to keep the line.

src/storage/config.rs
```diff
@@ -27,7 +27,7 @@ const MAX_SCHED_CONCURRENCY: usize = 2 * 1024 * 1024;
 // here we use 100MB as default value for tolerate 1s latency.
 const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
 
-const DEFAULT_RESERVED_SPACE_GB: u64 = 5;
+pub const DEFAULT_RESERVED_SPACE_GB: u64 = 5;
```
hicqu, line 30: Shouldn't be `pub`.

src/server/service/kv.rs
tier-cap: done

components/raftstore/src/store/peer.rs
tier-cap: done

tier-cap: > Better to allow configuration change on leader.

yes, Configure change allowed when leader disk full.

components/server/src/server.rs
```diff
@@ -238,6 +245,12 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         let latest_ts = block_on(pd_client.get_tso()).expect("failed to get timestamp from PD");
         let concurrency_manager = ConcurrencyManager::new(latest_ts);
 
+        let mut disk_reserved = config.storage.reserve_space.0;
+        if disk_reserved == 0 {
```
hicqu, line 249: If it's 0, `0` should be reserved.

components/server/src/server.rs
tier-cap: removed

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,257 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use kvproto::raft_cmdpb::*;
+use raftstore::store::msg::*;
+use std::time::Duration;
+use std::time::*;
+use test_raftstore::*;
+use tikv_util::config::*;
+/*
```
hicqu, line 9: How about remove this comment and comment on every case itself?

tier-cap: ok, I'll change

components/server/src/server.rs
tier-cap: ok, keep 0.

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

src/server/service/kv.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: should kept, because the merge region case need.

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: support read only.

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: Maybe not good, because we hope the transfer failed, and I've not found good way to check it.

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
tier-cap: ok

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

components/raftstore/src/store/peer.rs
tier-cap: It means here we no need to get the store id.

components/server/src/server.rs
tier-cap: done

components/server/src/server.rs
tier-cap: done

tier-cap: not auto format~

src/server/service/kv.rs
tier-cap: done

src/storage/config.rs
tier-cap: done

tier-cap: not reset from the earlier impl logic

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1231,6 +1231,13 @@ where
             "to_peer_id" => msg.get_to_peer().get_id(),
         );
 
+        let msg_type = msg.get_message().get_msg_type();
+        let store_id = self.ctx.store_id();
+        if (disk::disk_full_precheck(store_id) || disk::is_disk_full())
+            && (msg_type == MessageType::MsgAppend || msg_type == MessageType::MsgTransferLeader)
+        {
+            return Err(Error::Timeout(String::from("disk full")));
```
hicqu, line 1239: Why `Timeout`? Should we add a new error code `DiskFull`?

components/raftstore/src/store/peer.rs
```diff
@@ -2341,7 +2342,13 @@ where
...
+                    Err(box_err!("disk full, all the business data write forbiden"))
```
hicqu, line 2347: Should we add a new error code `DiskFull`?

hicqu (approved):

components/engine_traits/src/raft_engine.rs
```diff
@@ -80,6 +80,10 @@ pub trait RaftEngine: RaftEngineReadOnly + Clone + Sync + Send + 'static {
     fn stop(&self) {}
 
     fn dump_stats(&self) -> Result<String>;
+
+    fn get_engine_size(&self) -> Result<u64> {
```
hicqu, line 84: Shouldn't give a default implementation.

components/server/src/server.rs
```diff
@@ -1016,9 +1017,16 @@ impl<ER: RaftEngine> TiKVServer<ER> {
     }
 
     fn init_storage_stats_task(&self, engines: Engines<RocksEngine, ER>) {
+        // currently only for RocksEngine.
+        if self.config.raft_engine.enable {
```
hicqu, line 1020: It's wrong. raft_engine is only for replacing raftdb. rocksdb is still used even if raft_engine is enabled.

components/server/src/server.rs
```diff
@@ -1060,7 +1072,11 @@ impl<ER: RaftEngine> TiKVServer<ER> {
                         available, snap_size, kv_size, capacity
                     );
                     disk::set_disk_full();
-                } else {
+                } else if disk::is_disk_full() {
+                    warn!(
```
hicqu, line 1075: Should be `info`.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1233,15 +1233,40 @@ where
...
         let store_id = self.ctx.store_id();
-        if (disk::disk_full_precheck(store_id) || disk::is_disk_full())
-            && [MessageType::MsgAppend, MessageType::MsgTimeoutNow].contains(&msg_type)
-        {
-            debug!(
-                "skip {:?} because of disk full", msg_type;
-                "region_id" => self.region_id(), "peer_id" => self.fsm.peer_id()
-            );
-            return Err(Error::Timeout("disk full".to_owned()));
+        // if (disk::disk_full_precheck(store_id) || disk::is_disk_full())
```
hicqu, line 1227: Please clean it.

components/server/src/server.rs
tier-cap: right, forget to remove this after add the get_engine_size() fn.

components/raftstore/src/store/fsm/peer.rs
tier-cap: done

components/server/src/server.rs
tier-cap: right, done

components/engine_traits/src/raft_engine.rs
tier-cap: ok, changed.

tests/failpoints/cases/test_disk_full.rs
tier-cap: done

components/server/src/server.rs
```diff
@@ -1015,6 +1024,65 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+                let used_size = snap_size + kv_size + raft_size;
+                let capacity = if config_disk_capacity == 0 || disk_cap < config_disk_capacity {
+                    disk_cap
+                } else {
+                    config_disk_capacity
+                };
+
+                let mut available = capacity.checked_sub(used_size).unwrap_or_default();
+                available = cmp::min(available, disk_stats.available_space());
+                if available <= disk::get_disk_reserved() {
```
NingLin-P, line 1070: We have already reserved disk space by creating a placeholder file, adding this check along with the placeholder file will make the reserved size double /cc @hicqu , see: 
https://github.com/tikv/tikv/blob/a3da6cec14857de021da6ed14353ae461303047d/components/server/src/server.rs#L400-L420

components/server/src/server.rs
```diff
@@ -243,6 +247,9 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         let latest_ts = block_on(pd_client.get_tso()).expect("failed to get timestamp from PD");
         let concurrency_manager = ConcurrencyManager::new(latest_ts);
 
+        let disk_reserved = config.storage.reserve_space.0;
+        disk::set_disk_reserved(disk_reserved);
```
NingLin-P, line 251: The `config.storage.reserve_space` can't be changed online, maybe it is better to pass the config to the worker instead of setting & reading from the atomic value.

components/tikv_util/src/sys/disk.rs
```diff
@@ -0,0 +1,35 @@
...
+pub fn clear_disk_full() {
+    DISK_FULL.store(false, Ordering::Release);
+}
+
+pub fn is_disk_full() -> bool {
+    DISK_FULL.load(Ordering::Acquire)
+}
+
+pub fn disk_full_precheck(_store_id: u64) -> bool {
+    if cfg!(feature = "failpoints") {
```
NingLin-P, line 20: This check is unnecessary, the `fail_point` macro already includes this check.

components/server/src/server.rs
```diff
@@ -1015,6 +1024,65 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+    fn init_storage_stats_task(&self, engines: Engines<RocksEngine, ER>) {
+        let config_disk_capacity: u64 = self.config.raft_store.capacity.0;
+        let store_path = self.store_path.clone();
+        let snap_mgr = self.snap_mgr.clone().unwrap();
+        //TODO wal size ignore?
+        self.background_worker
+            .spawn_interval_task(DEFAULT_STORAGE_STATS_INTERVAL, move || {
+                if disk::get_disk_reserved() == 0 {
+                    return;
+                }
```
NingLin-P, line 1036: How about checking this before starting the worker, so if the reserve size is zero we don't need to start the worker.

components/engine_rocks/src/raft_engine.rs
```diff
@@ -221,6 +221,13 @@ impl RaftEngine for RocksEngine {
     fn dump_stats(&self) -> Result<String> {
         MiscExt::dump_stats(self)
     }
+
+    fn get_engine_size(&self) -> Result<u64> {
+        let handle = util::get_cf_handle(self.as_inner(), CF_DEFAULT)?;
```
NingLin-P, line 226: Why not checking other cf? or maybe we can reuse `get_engine_used_size`.

hicqu: `get_engine_used_size` is for kvdb instead of raftdb. Here `get_engine_size` is a trait method of `RaftEngine`, so only `CF_DEFAULT` needs to be considered.

components/server/src/server.rs
tier-cap: but, if users set it back to not zero, then it will not restast.
place here will let users set it online.

components/server/src/server.rs
tier-cap: we have resereved a replica value, to be used by the worker.
and after this pr, we will set up another pr to deal with the reserve_space and the place holder file.

components/server/src/server.rs
tier-cap: right, the place holder file we will deal with in another pr.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    // To ensure all replicas are not pending.
+    cluster.must_put(b"k1", b"v1");
+    must_get_equal(&cluster.get_engine(1), b"k1", b"v1");
+    must_get_equal(&cluster.get_engine(2), b"k1", b"v1");
+    must_get_equal(&cluster.get_engine(3), b"k1", b"v1");
+
+    test_unallowed_leader_behaviors(&mut cluster);
+    test_allowed_leader_behaviors(&mut cluster);
+    test_follower_behaviors(&mut cluster);
```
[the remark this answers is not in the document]
tier-cap: these cases have no much data race, maybe it's ok now.

components/tikv_util/src/sys/disk.rs
tier-cap: right，done!

components/server/src/server.rs
tier-cap: For this feature, a PR will be specially opened to handle the place holder file. The desired effect is to allow users to dynamically set the reserved size. So the worker is retained.

tests/failpoints/cases/test_disk_full.rs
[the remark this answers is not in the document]
NingLin-P: The later test may affect by the previous test, although we can clean up the environment before the next test but different test logic should not interfere with each other.

components/server/src/server.rs
NingLin-P: Plz leave a TODO for now.

components/server/src/server.rs
NingLin-P: We don't need to `set_disk_reserved` and `get_disk_reserved` since the `reserve_space` not support changing online, we can just pass the config value to the worker and use it directly.

components/server/src/server.rs
NingLin-P: Get it

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1232,6 +1232,33 @@ where
...
+        let msg_type = msg.get_message().get_msg_type();
+        let store_id = self.ctx.store_id();
+
+        if disk::disk_full_precheck(store_id) || disk::is_disk_full() {
+            let mut flag = false;
+            if MessageType::MsgAppend == msg_type {
+                let entries = msg.get_message().get_entries();
+                for i in entries {
+                    let entry_type = i.get_entry_type();
+                    if EntryType::EntryNormal == entry_type && !i.get_data().is_empty() {
```
NingLin-P, line 1244: `MsgAppend` may contain multiple entries, I think we should not drop the message if there are `confchange` entries. /cc @hicqu

components/server/src/server.rs
hicqu: > we can just pass the config value to the worker and use it directly

`disk_reserved` is also used in `fn disk_is_full`. If it's not stored in a global available place, we need to pass a parameter into `disk_is_full(reserved)`.

components/raftstore/src/store/fsm/peer.rs
hicqu: So if it contains lots of large normal entries, TiKV can still run out of disk space.

I think both 2 implementations are reasonable. We can make the decision later in another PR after the design document is posted and confirmed.

components/server/src/server.rs
tier-cap: ok. done

components/server/src/server.rs
tier-cap: good advise, I'll simplify it.

components/server/src/server.rs
tier-cap: When using tiup to change the configuration parameters, the load operation will restart one node after another. Therefore, this PR does not consider online matters for the time being. 
Done

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1232,6 +1232,33 @@ where
             "to_peer_id" => msg.get_to_peer().get_id(),
         );
 
+        let msg_type = msg.get_message().get_msg_type();
+        let store_id = self.ctx.store_id();
+
+        if disk::disk_full_precheck(store_id) || disk::is_disk_full() {
```
NingLin-P, line 1238: Checking an atomic for every raft message and proposal may bring extra cost. 

Maybe we can reduce the cost by only check the atomic once for each batch of peer, more specifically setting a flag like `poller_ctx.is_disk_full` at `PollHandler::begin` and then checking the `poller_ctx.is_disk_full`  for each raft message and proposal instead.

This may bring little delay to detect disk full but I think the delay will be much shorter than the disk stats monitor interval (1s).

NingLin-P (commented):
Rest LGTM

Also, I found this PR didn't cover snapshot generating and receiving, those could also consume disk space.

components/raftstore/src/store/fsm/peer.rs
tier-cap: > Also, I found this PR didn't cover snapshot generating and receiving, those could also consume disk space.
yes, currently, these are also not limited in this PR: snapshot, runninglog, rocksdb I/O, wal, etc.

tier-cap: > Checking an atomic for every raft message and proposal may bring extra cost.
> 
> Maybe we can reduce the cost by only check the atomic once for each batch of peer, more specifically setting a flag like `poller_ctx.is_disk_full` at `PollHandler::begin` and then checking the `poller_ctx.is_disk_full` for each raft message and proposal instead.
> 
> This may bring little delay to detect disk full but I think the delay will be much shorter than the disk stats monitor interval (1s).

The impact of access to atomic variables on performance should not be so big, we can add a performance test observation.

NingLin-P: Not just performance but also resource cost (like CPU), the impact may not big enough to be noticeable but I think it is an unnecessary cost though.

components/tikv_util/src/sys/disk.rs
```diff
@@ -0,0 +1,22 @@
...
+
+pub fn is_disk_full() -> bool {
+    DISK_FULL.load(Ordering::Acquire)
+}
+
+pub fn disk_full_precheck(_store_id: u64) -> bool {
+    fail_point!("disk_full_peer_1", _store_id == 1, |_| true);
+    fail_point!("disk_full_peer_2", _store_id == 2, |_| true);
+    false
+}
```
NingLin-P, line 22: Looks like `disk_full_precheck` is always use as ` disk_full_precheck || is_disk_full`, how about combining these two into one like:
```
pub fn is_disk_full() -> bool {
    fail_point!("disk_full_peer_1", _store_id == 1, |_| true);
    fail_point!("disk_full_peer_2", _store_id == 2, |_| true);
    DISK_FULL.load(Ordering::Acquire)
}
```

tier-cap: Both are ok.

tier-cap: Maybe not. Because disk_full_precheck only used to ut, if we merge both, then the pd store heartbeat will clear it.

tier-cap: And there will be a more obscure logic, that is, in other places where precheck is not needed, you have to pass a store_id to this merged function.

NingLin-P (approved):
LGTM

hicqu (approved):

PR #10273, opened 2021-06-01
raftstore: Add hibernate region metrics (#10152)

ti-srebot:
cherry-pick #10152 to release-5.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/10273
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/tikv.git pr/10273:release-5.0-0f5f66632b4d
```

---

### What problem does this PR solve?

Adding hibernate region metrics

Problem Summary:

### What is changed and how it works?

What's Changed:
Adding the following metrics
* region count in different hibernate state, breakdown by leader/non-leader
* leader count in different hibernate leader state (awaken/poll/hibernated)

example:
It is in the "Server" pannel.
Screen Shot 2021-05-26 at 11 27 55 PM

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 

- No release notes

PR #10277, opened 2021-06-01
tests: fix flaky test_read_index_retry_lock_checking

sticnarf:

https://github.com/tikv/tikv/pull/10273#issuecomment-851829389 reports that `test_read_index_retry_lock_checking` always fails.

The failure is on the first `async_read_index_on_peer`. Previously the election timeout is set to very short, then an election may happen before the read. So we receive a `Stale command` error instead of waiting.

This PR increases election timeout to solve the problem.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


5kbpers (approved):

NingLin-P (approved):

sticnarf:
The failed test `test_upload_sst` is not related to this PR :(

PR #10280, opened 2021-06-01
analyze: correct the offset used in sampling

winoros:

Problem Summary:

The offset was wrongly used. Then we would collect the wrong stats for the column group.

What's Changed:

Use the correct offsets and add tests for it.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


lzmhhh123 (approved):

sticnarf:
Please format your code :)

skyzh (commented):
The behavior seems inconsistent with https://github.com/pingcap/tidb/blob/master/statistics/row_sampler.go#L190, where `total_sizes` will still be modified if the column is null. (Does it always have length of 0?)

By the way, if we only need to use `offsets[j]`, why not iterate only on `offsets`?

```rust
for j in &offsets {
   // do something
}
```

skyzh:
Please also fix clippy warnings.

winoros:
@skyzh I'll fix TiDB's later. TIDB's is only used for test. NULL would always be zero when do calculate.

skyzh (approved):

PR #10281, opened 2021-06-01
*: remove build.rs

skyzh:
Signed-off-by: Alex Chi <iskyzh@gmail.com>

Problem Summary:

This PR removes `build.rs`.

What's Changed:

After `build.rs` has been removed, we must modify Jenkins to upload built artifacts to cache. Otherwise the tests would fail.

### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
None
```


BusyJay:
The test can't pass when running `env EXTRA_CARGO_ARGS=hot_reloading make test`.

skyzh:
> The test can't pass when running `env EXTRA_CARGO_ARGS=hot_reloading make test`.

I've fixed this and tested on my machine.

BusyJay:
example_plugin should be added as dev-dependency. There is no build script, so build-dependency is ignored.

skyzh:
> example_plugin should be added as dev-dependency. There is no build script, so build-dependency is ignored.

I’m no sure if defining example_plugin (which is not a library) as dev-dependency would produce a binary. I’ll have a try later.

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -450,17 +450,9 @@ fn is_library_file<P: AsRef<Path>>(path: P) -> bool {
...
-    static INIT: Once = Once::new();
-    static EXAMPLE_PLUGIN: &[u8] = include_bytes!(env!("CARGO_DYLIB_FILE_EXAMPLE_PLUGIN"));
 
     fn initialize_library() -> PathBuf {
-        let lib_path = std::env::temp_dir().join(&pkgname_to_libname("example-plugin"));
-        INIT.call_once(|| {
-            std::fs::write(&lib_path, EXAMPLE_PLUGIN).unwrap();
-        });
-        lib_path
+        Path::new("target/debug/deps").join(pkgname_to_libname("example-plugin"))
```
Jibbow, line 455: will this always be in `debug`? :see_no_evil:

skyzh: I don't think someone would run the tests in release mode... So I hardcoded debug here.

skyzh:
> example_plugin should be added as dev-dependency. There is no build script, so build-dependency is ignored.

After all, it works magically. I've pushed new commits to fix this.

src/coprocessor_v2/plugin_registry.rs
BusyJay: Actually, TiKV CI used to use release profile to speed up tests.

skyzh: Then from my perspective, the best way to do this without hard-code artifact path is to add a crate with `build.rs` which exports `OUT_DIR`, and we could use it here.

Another approach is to move all related logic and tests to a separate crate like `plugin_manager`, and we could create a `build.rs` that doesn't affect the whole TiKV crate.

skyzh: Now we just use `if cfg!(debug_assertions)` to identify which binary to use.

skyzh:
Looks cool! And now it's much easier to find which case fails

```
raftstore::test_region_heartbeat::test_server_down_peers_without_hibernate_regions
```

BusyJay (approved):

BusyJay:
> raftstore::test_region_heartbeat::test_server_down_peers_without_hibernate_regions

It's one of those unstable cases and unrelated to this PR.

andylokandy (approved):
Oh, good job! How did you resolve the CI problem?

skyzh:
> Oh, good job! How did you resolve the CI problem?

BusyJay has rewritten the CI workflow to speed up the tests while fixing this issue in the meantime.

PR #10284, opened 2021-06-01
build: test new workflow

BusyJay:

What's Changed:

This PR is a POC to show new workflow can speed up TiKV CI.

### Check List 

Tests 

- No code

### Release note 
- No release note
```release-note
Please add a release note.
```


BusyJay:
Old workflow takes 19m53s to finish, new workflow takes 10m27s to finish. There are two major improvements:
- Cache doesn't need network access anymore (11m55s -> 4m24s)
- Test job is split based on approximate cost time instead of hand writing, requires less concurrency and finish more quickly (7m15s -> 5m52s)

Time can still be further reduced by optimize test logic and better split policy.

BusyJay:
New flow is landed.

PR #10288, opened 2021-06-01
backup: do not recycle backup threads

### LINKED ISSUE #10287: Backup should not recycle threads since we set thread number via config
## Bug Report

Backup threads are recycled when there is no backup task in the past 30 minutes, it becomes unnecessary since https://github.com/tikv/tikv/pull/8193 sets threads number via config. Keeping a fixed-size threads pool is very cheap.

And for v4.0.x tikv, it leaks threads because we do not shutdown thread pools properly.

~~v4.0.0-v4.0.13~~
v4.0.0-v4.0.9

https://github.com/tikv/tikv/commit/1bdc5ffef8b8bb308ff3be69643a0d0c89aabc23 fixes the issue.

### Steps to reproduce

Run backup.

### What did you expect?

Do not recycle or spawn new threads.

overvenus:

Issue Number: close #10287

Do not recycle backup threads.

### Related changes

### Check List 

Tests 

Run backup for every 30 minutes. 

### Release note 

For release-5.0

```release-note
None
```

For release-4.0

```release-note
Fix backup threads leak
```


Leavrth:
The following code block may need to be also modified.
https://github.com/tikv/tikv/blob/master/components/backup/src/endpoint.rs#L576-L584

Leavrth:
Since function `check_active` is deleted, `self.size in struct Enpoint` isn't changed anymore. And `new_size` in function `adjust_with` is a const number. So The fuction adjust_with always this `if block` is always true.
https://github.com/tikv/tikv/blob/master/components/backup/src/endpoint.rs#L592

Therefore self.worker will not be replace, which means this code block can be update.
https://github.com/tikv/tikv/blob/master/components/backup/src/endpoint.rs#L576-L584

Leavrth (approved):

glorv (approved):

3pointer (approved):

PR #10289, opened 2021-06-02
raftstore: use asynchronous io

gengliqi:

Problem Summary:
The sync io in store thread introduces much extra latency. After detailed analysis, we found that it increases commit duration by about one IO+fsync time. After https://github.com/tikv/raft-rs/pull/403, we can use async io in store thread.

There are some things to pay attention to
1. snapshot process: I change it a lot in this PR. It is the most prone to bugs in my opinion. 
2. destroy process: for simplicity, the operation of writing kv db and raft db during destroy is still in store thread. We can move it to io thread in the future.
3. semantic of some functions in `PeerStorage`: the memory state is updated before the data is really persisted, which is different from the original code. Although the raft messages are delayed, maybe some other channels that I may not have considered may break correctness.

Furthermore, some things are not done yet
1. flow control: the unpersisted entries must in memory, so if the leader/follower receives a huge amount of entries, TiKV may OOM.
2. test: the unit tests and integration tests is important to some new logic.

What's Changed:
use asynchronous io in store thread.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
* Use asynchronous io in store thread to reduce commit latency


BusyJay:
There is another TODO, stall TiKV will not be auto recovered by elections from timeout followers.

gengliqi:
> There is another TODO, stall TiKV will not be auto recovered by elections from timeout followers.

Nice catch. Flow control should do its best to avoid this situation. On the other hand, tikv should report to PD if flow control is triggered and PD can transfer leader or move the region to other TiKVs. @nolouch What's your opinion?

src/server/config.rs
```diff
@@ -183,6 +193,10 @@ impl Default for Config {
             advertise_status_addr: DEFAULT_ADVERTISE_LISTENING_ADDR.to_owned(),
             status_thread_pool_size: 1,
             max_grpc_send_msg_len: DEFAULT_MAX_GRPC_SEND_MSG_LEN,
+            raft_client_grpc_send_msg_buffer: 256 * 1024,
+            raft_client_queue_size: 4096,
+            raft_msg_max_batch_size: 128,
+            raft_msg_flush_delay_us: 500,
```
innerr, line 199: This looks huge

gengliqi: It will be removed. See https://github.com/tikv/tikv/pull/10226

zhangjinpeng1987 (commented):
Such a huge PR is hard to merge. It will be delayed a lot by its size.

gengliqi:
> Such a huge PR is hard to merge. It will be delayed a lot by its size.

Yes, I agree. But the code which can be split into another PR is little. Many main changes are related to the async io, e.g. handle ready process, snapshot process. In my opinion, it is hard to change only one part without changing other parts.

BusyJay:
How about introducing async io first without changing any raftstore code?

gengliqi:
> How about introducing async io first without changing any raftstore code?

Good idea! I will propose this PR soon.

components/raftstore/src/store/config.rs
```diff
@@ -250,16 +265,21 @@ impl Default for Config {
             local_read_batch_size: 1024,
             apply_batch_system: BatchSystemConfig::default(),
             store_batch_system: BatchSystemConfig::default(),
+            store_io_pool_size: 1,
```
innerr, line 268: `1` as default looks too small

components/raftstore/src/store/metrics.rs
```diff
@@ -209,6 +208,155 @@ make_static_metric! {
 }
 
 lazy_static! {
+    pub static ref STORE_TIME_HISTOGRAM: Histogram =
+        register_histogram!(
+            "tikv_raftstore_store_duration_secs",
+            "TODO",
```
innerr, line 214: There are lots of "TODO" here

components/raftstore/src/store/fsm/store.rs
```diff
@@ -733,6 +647,7 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> RaftPoller<EK, ER, T> {
         if let Some(e) = trace_event {
             MEMTRACE_PEERS.trace(e);
         }
+        // TODO(ASYNC_IO): change the logic
```
innerr, line 650: Resolve this TODO?

components/raftstore/src/store/peer.rs
```diff
@@ -1747,7 +1898,26 @@ where
 
         let mut ready = self.raft_group.ready();
 
-        self.last_unpersisted_number = ready.number();
+        let mut proposal_times = vec![];
+
+        if ctx.raft_metrics.waterfall_metrics {
```
innerr, line 1903: Create another PR for new Grafana panels later?

gengliqi: Yes, I plan to introduce `waterfall_metrics` in another PR. By the way, https://github.com/tikv/tikv/pull/10326 is split from this PR for introducing `async_io.rs`.

components/raftstore/src/store/config.rs
gengliqi: I change it to 2 in #10326

zhangjinpeng1987:
Please keep the PR updated

zhangjinpeng1987:
please split this huge PR into several small PRs

purelind:
build

gengliqi:
FYI, the code of write router is split to https://github.com/tikv/tikv/pull/10667. The reschedule policy is not important because it's very hard to trigger. I think we can merge it quickly to reduce about 400 lines of code and then focus on this PR which is the most important one.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1298,6 +1312,9 @@ where
     }
 
     fn on_raft_message(&mut self, msg: InspectedRaftMessage) -> Result<()> {
+        if self.fsm.peer.pending_remove || self.fsm.stopped {
```
hicqu, line 1315: Please move it after `defer`, because some metrics need to be updated.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -782,14 +693,13 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
     fn begin(&mut self, _batch_size: usize) {
         self.previous_metrics = self.poll_ctx.raft_metrics.clone();
-        self.poll_ctx.processed_fsm_count = 0;
         self.poll_ctx.pending_count = 0;
+        self.poll_ctx.ready_count = 0;
         self.poll_ctx.sync_log = false;
         self.poll_ctx.has_ready = false;
         self.poll_ctx.disk_usage = disk::get_disk_status(self.poll_ctx.store.get_id());
-        self.timer = TiInstant::now_coarse();
+        self.timer = TiInstant::now();
```
hicqu, line 701: Is it expected to change it?

gengliqi: Yes. The store loop is faster than before and usually less than 1ms.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -140,7 +143,7 @@ where
     raft_entry_max_size: f64,
     batch_req_size: u32,
     request: Option<RaftCmdRequest>,
-    callbacks: Vec<(Callback<E::Snapshot>, usize, TiInstant)>,
```
hicqu, line 143: Why remove `TiInstant`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2053,13 +2057,32 @@ where
...
     }
 
-    fn destroy_peer(&mut self, merged_by_target: bool) {
+    fn destroy_peer(&mut self, merged_by_target: bool) -> bool {
         fail_point!("destroy_peer");
+        // Mark itself as pending_remove
+        self.fsm.peer.pending_remove = true;
+
+        if self.fsm.peer.has_unpersisted_ready() {
+            // The destroy must be destroyed if there are some unpersisted readies.
```
hicqu, line 2070: s/must be destroyed/must be delayed

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2112,13 +2132,16 @@ where
             );
         }
         let is_initialized = self.fsm.peer.is_initialized();
+        self.ctx.perf_context.start_observe();
```
hicqu, line 2135: How about moving `perf_context` modifications into `destroy`? I worry about someone can forget it again.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3179,14 +3216,19 @@ where
             "prev_region" => ?prev_region,
         );
 
+        meta.readers.insert(
```
hicqu, line 3219: Why register here? Is it a bug?

components/raftstore/src/store/peer.rs
```diff
@@ -110,6 +113,19 @@ impl<S: Snapshot> ProposalQueue<S> {
         }
     }
 
+    fn find_request_times(&self, index: u64) -> Option<(u64, &SmallVec<[TiInstant; 4]>)> {
+        self.queue
+            .binary_search_by_key(&index, |p: &Proposal<_>| (p.index))
```
hicqu, line 118: Seems redundant `(..)`.

components/raftstore/src/store/peer.rs
```diff
@@ -519,14 +561,19 @@ where
...
 
     pub read_progress: Arc<RegionReadProgress>,
 
     pub memtrace_raft_entries: usize,
+    /// Used for sending write msg.
+    write_router: WriteRouter<EK, ER>,
+    unpersisted_readies: VecDeque<UnpersistedReady>,
+    /// The message count in `unpersisted_readies` for memory caculation.
+    unpersisted_message_count: usize,
+    persisted_number: u64,
```
hicqu, line 574: It's better to document the field too.

components/raftstore/src/store/fsm/peer.rs
gengliqi: Because there is a bug in waterfall metrics here. If the cmd is not batched, the request time in its callback is not set.
So we don't need the `TiInstant` to set the request time in callback(it will be set during construction).

components/raftstore/src/store/fsm/peer.rs
gengliqi: Done. I realize that the `perf_context` here will record the kv db which is not right because it's raftstore's perf context.(We should add a new `perf_context` to record kv db metrics here.

components/raftstore/src/store/peer.rs
```diff
@@ -1767,57 +2005,127 @@ where
...
+                        region: snap_region,
+                        destroy_regions,
+                    }),
+                });
+                // Snapshot will be scheduled after persisting this ready
+                self.raft_group.advance_append_async(ready);
 
-            self.activate(ctx);
-            let mut meta = ctx.store_meta.lock().unwrap();
-            meta.readers
```
gengliqi, line 1813: The `meta.readers.insert` is moved from here @hicqu

components/raftstore/src/store/peer.rs
gengliqi: I don't get it.

components/raftstore/src/store/fsm/peer.rs
gengliqi: Seems comments below.

src/server/raft_client.rs
```diff
@@ -930,29 +936,44 @@ where
         }
     }
 
+    pub fn delay_flush(&mut self) {
+        if !self.need_flush() {
+            return;
+        }
+        let now = TiInstant::now_coarse();
+        if now > self.last_flush_time + self.delay_flush {
```
BusyJay, line 944: Why not use a timer in future executor instead?

components/raftstore/src/store/peer.rs
hicqu: I mean `s/(p.index)/p.index`.

gengliqi: Done.

src/server/raft_client.rs
gengliqi: Ok. I will propose another PR to fix this. I remove these codes.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -410,20 +415,15 @@ where
...
             self.batch_req_size = 0;
             if self.callbacks.len() == 1 {
-                let (mut cb, _, send_time) = self.callbacks.pop().unwrap();
-                if let Callback::Write { request_times, .. } = &mut cb {
-                    if metric.waterfall_metrics {
-                        metric
-                            .batch_wait
-                            .observe(send_time.saturating_elapsed_secs());
-                    }
-                    *request_times = smallvec![send_time];
```
BusyJay, line 424: Why delete?

components/raftstore/src/store/metrics.rs
```diff
@@ -284,6 +290,12 @@ lazy_static! {
             "Bucketed histogram of proposals' wait batch duration.",
             exponential_buckets(0.00001, 2.0, 26).unwrap()
         ).unwrap();
+    pub static ref STORE_TO_WRITE_QUEUE_DURATION_HISTOGRAM: Histogram =
+        register_histogram!(
+            "tikv_raftstore_store_to_write_queue_duration_seconds",
+            "Bucketed histogram of proposals' to write queue duration.",
```
BusyJay, line 296: What is "to write queue duration"?

components/raftstore/src/store/metrics.rs
```diff
@@ -302,6 +314,24 @@ lazy_static! {
             "Bucketed histogram of proposals' write db end duration.",
             exponential_buckets(0.00001, 2.0, 26).unwrap()
         ).unwrap();
+    pub static ref STORE_KNOW_PERSIST_DURATION_HISTOGRAM: Histogram =
```
BusyJay, line 317: How about removing the `know_` prefix? It seems redundant.

BusyJay (commented):
There seems to be no tests for async IO writes except delay destroying.

components/raftstore/src/store/fsm/peer.rs
gengliqi: Some cmd may not be batched so the `batch_wait` metric is moved to `propose_raft_command`.

components/raftstore/src/store/metrics.rs
gengliqi: Done.

components/raftstore/src/store/metrics.rs
gengliqi: Change the name to `send to queue`

components/raftstore/src/store/peer_storage.rs
```diff
@@ -205,18 +210,6 @@ impl EntryCache {
             }
         }
 
-        let (cache_len, mut entries_len) = (self.cache.len(), entries.len());
```
BusyJay, line 208: Why? If cache are not expected to drop entries, better add comment explains the constraint.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1238,10 +1199,10 @@ where
         &mut self,
         ctx: &mut InvokeContext,
         snap: &Snapshot,
-        kv_wb: &mut EK::WriteBatch,
-        raft_wb: &mut ER::LogBatch,
-        destroy_regions: &[metapb::Region],
-    ) -> Result<()> {
+        task: &mut WriteTask<EK, ER>,
+        msgs: Vec<RaftMessage>,
```
BusyJay, line 1203: It feels strange to pass `msgs` and `destroy_regions` only for building a result. How about changing it to
```
pub fn apply_snapshot(&mut self, ctx: &mut InvokeContext, snap: &Snapshot, task: &mut WriteTask<EK, ER>, destroy_regions: &[metapb::Region]) -> Result<metapb::Region>;
```

components/raftstore/src/store/peer_storage.rs
```diff
@@ -422,28 +406,28 @@ impl From<Error> for RaftError {
...
+    },
+    NoIOTask {
+        msgs: Vec<RaftMessage>,
+    },
 }
 
-/// Returned by `PeerStorage::handle_raft_ready`, used for recording changed status of
-/// `RaftLocalState` and `RaftApplyState`.
+/// Used for recording changed status of `RaftLocalState` and `RaftApplyState`.
 pub struct InvokeContext {
```
BusyJay, line 424: Now all states are updated in place without writing to disk, the struct seems can be deleted.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1499,71 +1438,81 @@ where
...
-        self.raft_state = ctx.raft_state;
-        self.apply_state = ctx.apply_state;
-        self.last_term = ctx.last_term;
-        // If we apply snapshot ok, we should update some infos like applied index too.
-        let snap_region = match ctx.snap_region {
-            Some(r) => r,
-            None => return None,
-        };
+    pub fn update_persist_index(&mut self, persisted: u64) {
+        self.cache.persisted = persisted;
```
BusyJay, line 1509: Should it clear cache immediately?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1594,24 +1543,8 @@ where
...
-                self.truncated_index()
-            );
-        }
 
-        Some(ApplySnapResult {
-            prev_region,
-            region: self.region().clone(),
-            destroyed_regions: ctx.destroyed_regions,
-        })
+        self.set_region(res.region.clone());
```
BusyJay, line 1547: This maybe confusing as all other states are updated in place.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2453,47 +2346,22 @@ mod tests {
...
 
-        // compact shrink
-        assert!(store.cache.cache.capacity() >= cap as usize);
-        store.compact_to(cap * 2);
-        exp_res = (cap * 2..cap * 2 + 7).map(|i| new_entry(i, 8)).collect();
-        validate_cache(&store, &exp_res);
-        assert!(store.cache.cache.capacity() < cap as usize);
-
-        // append shrink
-        entries = (0..=cap).map(|i| new_entry(i, 8)).collect();
```
BusyJay, line 2484: Should check append will not shrink instead of just removing the case.

components/raftstore/src/store/peer_storage.rs
gengliqi: The persisted entries may be used to apply soon. If the memory usage is high, `on_entry_cache_evict_tick` will try to evict entry cache next time.

components/raftstore/src/store/peer_storage.rs
gengliqi: Indeed, I think updating in place is more similar to the original code.

components/raftstore/src/store/peer_storage.rs
gengliqi: The `test_storage_cache_entry` and `test_storage_cache_size_change` can meet your requirement.

components/raftstore/src/store/peer_storage.rs
BusyJay: I'm afraid cache usage will be pushed to the limit as leak can happen from time to time. This is dangerous according to the past experience.

hicqu: We can also add fields `applied_index` and `majority_matched_index` in `PeerStorage`.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1499,71 +1444,82 @@ where
...
+        if prev_raft_state != self.raft_state || !ready.snapshot().is_empty() {
+            write_task.raft_state = Some(self.raft_state.clone());
+        }
+
+        if !ready.snapshot().is_empty() {
+            // In case of restart happens when we just write region state to Applying,
+            // but not write raft_local_state to raft db in time.
+            // We write raft state to kv db, with last index set to snap index,
+            // in case of recv raft log after snapshot.
+            if write_task.kv_wb.is_none() {
```
BusyJay, line 1496: It seems to be always `Some` as applying snapshot requires kv_wb be not empty.

components/raftstore/src/store/peer_storage.rs
BusyJay: Can you add a comment explaining why delay setting the field?

components/raftstore/src/store/peer_storage.rs
BusyJay: How? For example, if L208~L219 is added back to the file, will these cases fail?

components/raftstore/src/store/peer_storage.rs
gengliqi: Done.

components/raftstore/src/store/peer_storage.rs
gengliqi: Has added a comment.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1262,6 +1198,15 @@ where
             ));
         }
 
+        if task.raft_wb.is_none() {
+            task.raft_wb = Some(self.engines.raft.log_batch(64));
```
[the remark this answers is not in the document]
gengliqi: `raft_wb` is allocated only when this peer needs to apply snapshot. This case should not be very frequent.

components/raftstore/src/store/peer_storage.rs
gengliqi: It's because the number of appending entries is not big enough. The `MAX_CACHE_CAPACITY` has been removed now.
So what's the number of appending entries in this test? If we use 1024 in test, maybe someone will use 2048 as `MAX_CACHE_CAPACITY` and still can pass this test.

components/raftstore/src/store/peer.rs
```diff
@@ -802,7 +875,12 @@ where
     /// 1. Set the region to tombstone;
     /// 2. Clear data;
     /// 3. Notify all pending requests.
-    pub fn destroy<T>(&mut self, ctx: &PollContext<EK, ER, T>, keep_data: bool) -> Result<()> {
+    pub fn destroy(
+        &mut self,
+        engines: &Engines<EK, ER>,
```
BusyJay, line 880: Any reasons not to use `PollContext`?

components/raftstore/src/store/metrics.rs
```diff
@@ -278,30 +284,54 @@ lazy_static! {
...
             "Bucketed histogram of proposals' write db end duration.",
             exponential_buckets(0.00001, 2.0, 26).unwrap()
         ).unwrap();
+    pub static ref STORE_WF_PERSIST_LOG_DURATION_HISTOGRAM: Histogram =
+        register_histogram!(
+            "tikv_raftstore_store_persist_duration_seconds",
+            "Bucketed histogram of proposals' persist duration.",
+            exponential_buckets(0.00001, 2.0, 26).unwrap()
+        ).unwrap();
+    pub static ref STORE_WF_COMMIT_LOG_DURATION_HISTOGRAM: Histogram =
```
BusyJay, line 323: Is it duplicated with `tikv_raftstore_commit_log_duration_seconds`?

gengliqi: `tikv_raftstore_commit_log_duration_seconds` metrics doesn't contain all `proposal` and it's not very accurate.

BusyJay: Then how about only keeping the new metrics?

components/raftstore/src/store/peer.rs
```diff
@@ -390,6 +408,34 @@ impl<S: Snapshot> Drop for CmdEpochChecker<S> {
...
+    pub destroy_regions: Vec<metapb::Region>,
+}
+
+#[derive(Debug)]
+pub struct UnpersistedReady {
+    /// Number of ready.
+    pub number: u64,
+    /// Max number of following ready whose data to be persisted is empty.
+    pub max_number: u64,
+    pub raft_msgs: Vec<Vec<RaftMessage>>,
```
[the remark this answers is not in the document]
gengliqi: `unpersisted_message_count` is used for this purpose

components/raftstore/src/store/peer.rs
gengliqi: Because the `PerfContext` needs borrow as mutable and `PollContext` has been borrow as immutable for `StoreMeta`.

components/raftstore/src/store/peer.rs
```diff
@@ -108,6 +113,19 @@ impl<S: Snapshot> ProposalQueue<S> {
         }
     }
 
+    fn find_request_times(&self, index: u64) -> Option<(u64, &SmallVec<[TiInstant; 4]>)> {
```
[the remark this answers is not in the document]
gengliqi: It's used to reduce some read io. If `find_request_times` returns None, no need to get this entry's term from raft db.

components/raftstore/src/store/metrics.rs
gengliqi: I worry that it may affect a bit of performance.

components/raftstore/src/store/peer.rs
```diff
@@ -1592,39 +1777,63 @@ where
...
-                }
             }
             CheckApplyingSnapStatus::Idle => {
                 // FIXME: It's possible that the snapshot applying task is canceled.
                 // Although it only happens when shutting down the store or destroying
-                // the peer, it's still dengerous if continue to handle ready for the
+                // the peer, it's still dangerous if continue to handle ready for the
                 // peer. So it's better to revoke `JOB_STATUS_CANCELLING` to ensure all
                 // started tasks can get finished correctly.
+                self.snap_ctx = None;
```
BusyJay, line 1819: Doesn't seem correct to me. How about returning false here if `snap_ctx` is `Some`?

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
gengliqi: Yes, so the caller needs to check the term later.

components/raftstore/src/store/peer.rs
gengliqi: The `PerfContext` here still belongs to `PollContext `.

components/raftstore/src/store/peer.rs
```diff
@@ -4016,7 +4338,10 @@ mod memtrace {
             // 1 word for every item in `want_rollback_merge_peers`
             + 8 * self.want_rollback_merge_peers.capacity()
             // Ignore more heap content in `raft::eraftpb::Message`.
-            + self.pending_messages.len() * mem::size_of::<raft::eraftpb::Message>()
+            + (self.unpersisted_message_count
```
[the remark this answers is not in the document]
gengliqi: I change to use `capacity`.

components/raftstore/src/store/peer.rs
```diff
@@ -1080,11 +1158,52 @@ where
...
+    #[inline]
+    pub fn switch_to_raft_msg<T>(
+        &mut self,
+        ctx: &PollContext<EK, ER, T>,
+        msgs: Vec<eraftpb::Message>,
+    ) -> Vec<RaftMessage> {
+        let mut raft_msgs = Vec::with_capacity(msgs.len());
+        for msg in msgs {
+            if let Some(mut m) = self.fill_raft_message(msg) {
+                m.disk_usage = ctx.self_disk_usage;
```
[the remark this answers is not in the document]
gengliqi: I move it to `fill_raft_message `. Seems the original code use this style.

components/raftstore/src/store/peer.rs
```diff
@@ -1917,54 +2239,82 @@ where
...
+            panic!(
+                "{} no match of persisted number {}, unpersisted numbers {:?}",
+                self.tag, number, self.unpersisted_readies
+            );
+        }
+        self.persisted_number = persisted_number;
 
-        if !light_rd.messages().is_empty() {
-            if !self.is_leader() {
-                fail_point!("raft_before_follower_send");
```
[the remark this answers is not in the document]
gengliqi: This failpoint has been located in write thread. I added it to some other place in store thread.

components/raftstore/src/store/peer.rs
gengliqi: Done. 
Actually, it's the same as the original code, if the snapshot is canceled and `pending_remove` is false, we will continue to handle ready. 
If `snap_ctx` is `Some`, return false can fix part of this problem. Another case is the snapshot comes from restarting and the `snap_ctx` is `None`.

gengliqi:
> I can see there are a lot of `assert` in this PR that if panic no useful context will be shown. Better add its region_id and peer_id for diagnostics.

I have changed these `assert` to `panic`.

components/raftstore/src/store/peer.rs
```diff
@@ -1567,13 +1739,25 @@ where
...
+    /// 1.  Get the snapshot from the ready
+    /// 2.  Wait for the notify of persisting this ready through `Peer::on_persist_ready`
+    /// 3.  Schedule the snapshot task to region worker through `schedule_applying_snapshot`
+    /// 4.  Wait for applying snapshot to complete(`check_snap_status`)
+    /// Then it's valid to handle the next ready.
+    fn check_snap_status<T: Transport>(&mut self, ctx: &mut PollContext<EK, ER, T>) -> bool {
+        if let Some(snap_ctx) = self.snap_ctx.as_ref() {
+            if !snap_ctx.scheduled {
+                // There is a snapshot from ready but it is not scheduled because the ready has
+                // not been persisted yet. We should wait for the notify of persisting ready and
```
BusyJay, line 1755: s/notify/notification/

components/raftstore/src/store/peer.rs
```diff
@@ -1634,13 +1850,23 @@ where
...
+                    "applied_index" => self.get_store().applied_index(),
                     "last_applying_index" => self.last_applying_idx,
                     "pending_request_snapshot_count" => count,
                 );
                 return None;
             }
 
+            if !self.unpersisted_readies.is_empty() {
+                debug!(
+                    "not ready to apply snapshot because some unpersisted readies have not persisted yet";
```
BusyJay, line 1862:
```suggestion
                    "not ready to apply snapshot because there are some unpersisted readies";
```

components/raftstore/src/store/peer.rs
```diff
@@ -1703,7 +1929,26 @@ where
...
 
-        self.last_unpersisted_number = ready.number();
+        let mut request_times = vec![];
+
+        if ctx.raft_metrics.waterfall_metrics {
+            let mut now = None;
+            for entry in ready.entries() {
+                if let Some((term, times)) = self.proposals.find_request_times(entry.get_index()) {
+                    if entry.term == term {
+                        request_times.append(&mut times.clone().into_vec());
```
BusyJay, line 1939: Why not `request_times.extend_from_slice(&times)`?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -539,15 +525,22 @@ where
...
+            for (region_id, (peer_id, ready_number)) in &self.batch.readies {
+                self.notifier
+                    .notify_persisted(*region_id, *peer_id, *ready_number);
+            }
+            STORE_WRITE_CALLBACK_DURATION_HISTOGRAM
+                .observe(duration_to_sec(now2.saturating_elapsed()));
         }
-        STORE_WRITE_CALLBACK_DURATION_HISTOGRAM.observe(duration_to_sec(now2.saturating_elapsed()));
 
-        self.batch.clear();
```
BusyJay, line 548: It's very easy to make mistake if batch is not cleared after written. How about clearing the batch and returning all readiness in `write_to_db`?

gengliqi: I am thinking about completely removing the sync process.
I have added a new config `cmd-batch-concurrent-ready-max-count` (default 1), which means the proposal will be batched if the number of concurrent readies is greater than this config. By the way, I remove `raft-ready-size-limit` as well for better batching.
I do a test and the result is very good. The CPU is lower than previous while the QPS is better.
Although this batch pattern is not perfect, it's very similiar to original code. Maybe delaying a adaptive time is better but it's little hard to implement it efficiently.

BusyJay: The switch can still be useful in extreme case if async IO is not working as expected. Now that it's implemented, I suggest to keep it, though it can be set as async by default.

gengliqi: Got it.

components/raftstore/src/store/peer.rs
```diff
@@ -1764,57 +2019,156 @@ where
...
-
-            // The peer may change from learner to voter after snapshot applied.
-            let peer = self
-                .region()
-                .get_peers()
-                .iter()
-                .find(|p| p.get_id() == self.peer.get_id())
-                .unwrap()
-                .clone();
-            if peer != self.peer {
```
BusyJay, line 1797: If it's updated after persisting, then the peer can be slightly different from the one in raft node. And in the past, messages are sent with updated peer info, now the messages are sent with stale peer info.

Though I don't find problems with current code space.

components/raftstore/src/store/peer.rs
```diff
@@ -1764,57 +2019,156 @@ where
...
+                    }
+                    if !light_rd.messages().is_empty() {
+                        panic!(
+                            "{} advance ready that has no io task but message is not empty {:?}",
+                            self.tag,
+                            light_rd.messages()
+                        );
+                    }
+                    if !light_rd.committed_entries().is_empty() {
+                        self.handle_raft_committed_entries(ctx, light_rd.take_committed_entries());
```
BusyJay, line 2122: How can it happen?

components/raftstore/src/store/peer.rs
```diff
@@ -1764,57 +2019,156 @@ where
...
 
     pub fn post_raft_ready_append<T: Transport>(
         &mut self,
         ctx: &mut PollContext<EK, ER, T>,
-        invoke_ctx: InvokeContext,
-        ready: &mut Ready,
-    ) -> Option<ApplySnapResult> {
-        if invoke_ctx.has_snapshot() {
-            // When apply snapshot, there is no log applied and not compacted yet.
-            self.raft_log_size_hint = 0;
```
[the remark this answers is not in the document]
gengliqi: ok. Now it is updated immediately.

components/raftstore/src/store/peer.rs
```diff
@@ -1764,57 +2019,156 @@ where
...
+            .clone();
+        if peer != self.peer {
+            info!(
+                "meta changed in applying snapshot";
+                "region_id" => self.region_id,
+                "peer_id" => self.peer.get_id(),
+                "before" => ?self.peer,
+                "after" => ?peer,
+            );
+            self.peer = peer;
```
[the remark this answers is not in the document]
gengliqi: My mistake.

components/raftstore/src/store/peer.rs
gengliqi: You can see the comments a few lines above. (now I move this comment above this line for clarity)

components/raftstore/src/store/peer.rs
gengliqi: Make sense. I refine the code of `post_raft_ready_append` and build raft messages before really sending them when these messages are saved in `SnapshotContext` or `unpersisted_readies`.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -337,7 +327,7 @@ where
     receiver: Receiver<WriteMsg<EK, ER>>,
     notifier: N,
     trans: T,
-    batch: WriteTaskBatch<EK, ER>,
+    pub batch: WriteTaskBatch<EK, ER>,
```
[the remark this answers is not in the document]
gengliqi: It's used for sync io process.

components/raftstore/src/store/peer.rs
```diff
@@ -1752,69 +1997,171 @@ where
...
-            // When apply snapshot, there is no log applied and not compacted yet.
+        mut collected: CollectedReady<EK, ER>,
+    ) {
+        let ready_number = collected.ready.number();
+        let persisted_msgs = collected.ready.take_persisted_messages();
+
+        match &collected.res {
+            HandleReadyResult::SendIOTask | HandleReadyResult::Snapshot { .. } => {
+                if !persisted_msgs.is_empty() {
+                    collected.write_task.messages = self.build_raft_messages(ctx, persisted_msgs);
```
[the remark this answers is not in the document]
gengliqi: If the ready has a snapshot, the persisted msgs will be taken to `HandleReadyResult::Snapshot` in `PeerStorage::handle_raft_ready`.

components/raftstore/src/store/peer.rs
```diff
@@ -1917,52 +2264,83 @@ where
...
-            // ready immediately to avoid flow out of control, so use
-            // `on_persist_ready` instead of `advance_append`.
-            // We don't need to set `has_ready` to true, as snapshot is always
-            // checked when ticking.
-            self.raft_group
-                .on_persist_ready(self.last_unpersisted_number);
-            return;
+        number: u64,
+    ) -> Option<PersistSnapshotResult> {
+        if self.unpersisted_readies.is_empty()
```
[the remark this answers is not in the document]
gengliqi: Seems not possible. I remove two conditions and reserve the `self.persisted_number >= number` for robustness.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2053,25 +2062,46 @@ where
...
         }
     }
 
-    fn destroy_peer(&mut self, merged_by_target: bool) {
+    fn destroy_peer(&mut self, merged_by_target: bool) -> bool {
         fail_point!("destroy_peer");
+        // Mark itself as pending_remove
+        self.fsm.peer.pending_remove = true;
+
+        if self.fsm.peer.has_unpersisted_ready() {
```
BusyJay, line 2074: It seems possible `destroy_peer` is called twice. For example, it's suggested to destroy itself by gc message and raft entry conf remove apply.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3152,12 +3201,12 @@ where
         self.handle_destroy_peer(job);
     }
 
-    fn on_ready_apply_snapshot(&mut self, apply_result: ApplySnapResult) {
-        let prev_region = apply_result.prev_region;
-        let region = apply_result.region;
+    fn on_ready_apply_snapshot(&mut self, persist_res: PersistSnapshotResult) {
```
BusyJay, line 3204: Maybe we can change the name to `on_ready_persist_snapshot` to make it clear.

components/raftstore/src/store/peer.rs
```diff
@@ -1917,54 +2262,76 @@ where
...
-            // ready immediately to avoid flow out of control, so use
-            // `on_persist_ready` instead of `advance_append`.
-            // We don't need to set `has_ready` to true, as snapshot is always
-            // checked when ticking.
-            self.raft_group
-                .on_persist_ready(self.last_unpersisted_number);
-            return;
+        number: u64,
+    ) -> Option<PersistSnapshotResult> {
+        if self.persisted_number >= number {
```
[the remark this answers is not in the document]
gengliqi: It's not possible for now because write thread sends `Persisted Msg` immediately after writing to db.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -683,61 +647,36 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> RaftPoller<EK, ER, T> {
...
-        fail_point!("raft_after_save");
-        if ready_cnt != 0 {
-            let mut ready_res = mem::take(&mut self.poll_ctx.ready_res);
-            for ready in ready_res.drain(..) {
-                PeerFsmDelegate::new(&mut peers[ready.batch_offset], &mut self.poll_ctx)
-                    .post_raft_ready_append(ready);
+            let mut readies = write_worker.write_to_db(false);
+            self.poll_ctx.write_worker = Some(write_worker);
+
+            for fsm in peers {
```
BusyJay, line 657: It's not as efficient as past. In the past, only readiness are looped. For example, if there a lot of heartbeats, there can be a lot of waste.

components/raftstore/src/store/peer.rs
```diff
@@ -780,13 +833,33 @@ where
...
 
-        if self.is_applying_snapshot() && !self.mut_store().cancel_applying_snap() {
-            info!(
-                "stale peer is applying snapshot, will destroy next time";
-                "region_id" => self.region_id,
-                "peer_id" => self.peer.get_id(),
-            );
-            return None;
+        if let Some(snap_ctx) = self.snap_ctx.as_ref() {
+            if !snap_ctx.scheduled {
```
BusyJay, line 837: Seems not covered by any case.

components/raftstore/src/store/peer.rs
```diff
@@ -1567,13 +1739,25 @@ where
...
+    ///
+    /// The snapshot process order would be:
+    /// 1.  Get the snapshot from the ready
+    /// 2.  Wait for the notify of persisting this ready through `Peer::on_persist_ready`
+    /// 3.  Schedule the snapshot task to region worker through `schedule_applying_snapshot`
+    /// 4.  Wait for applying snapshot to complete(`check_snap_status`)
+    /// Then it's valid to handle the next ready.
+    fn check_snap_status<T: Transport>(&mut self, ctx: &mut PollContext<EK, ER, T>) -> bool {
+        if let Some(snap_ctx) = self.snap_ctx.as_ref() {
+            if !snap_ctx.scheduled {
```
BusyJay, line 1753: Seems not covered by any case.

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/fsm/store.rs
gengliqi: I added `pos` for `handle_normal` to fix this issue. I think it's clearer than `processed_fsm_count`.

BusyJay (commented):
Still missing a case to test if ready will not be handled when a snapshot is being persisted.

gengliqi:
> Still missing a case to test if ready will not be handled when a snapshot is being persisted.

The test is added.
Also, the sync io process is removed with some tiny changes. PTAL again.

BusyJay (approved):

PR #10290, opened 2021-06-02
storage: fix that some latches of expired commands are not cleaned

sticnarf:

When a request expires at acquiring latches, it will yield an error immediately and release its latches. But previously, in the `release` method of `Latch`, it requires an accurate `owned_count`. Then, some locks inserted into the lock are not cleared and then blocks subsequent requests.

If the deadline is exceeded, we first update the `owned_count` of the command. Then, calling `release` method can release all locks of the previous commands.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


src/storage/txn/scheduler.rs
```diff
@@ -250,7 +250,14 @@ impl<L: LockManager> SchedulerInner<L> {
         let tctx = task_slot.get_mut(&cid).unwrap();
         // Check deadline early during acquiring latches to avoid expired requests blocking
         // other requests.
-        tctx.task.as_ref().unwrap().deadline.check()?;
+        if let Err(e) = tctx.task.as_ref().unwrap().deadline.check() {
```
youjiali1995, line 253: The function name is misleading...

youjiali1995 (approved):

youjiali1995:
Build failure...

ekexium (approved):

PR #10300, opened 2021-06-03
engine_traits: Remove TableProperties / UserCollectedProperties abstractions

### LINKED ISSUE #10068: Remove TableProperties abstractions from engine_traits
## Development Task

These traits, in `engine_traits/src/table_properties.rs`, are very complex and rocksdb-specific. It's not reasonable to expect any other engine to implement them.

Instead we should have higher-level abstractions that ask, "what is the caller using table properties for", and providing those answers.

An example of this already exists in the `RangePropertiesExt` trait, which uses table properties to answer high level questions about key ranges.

Find remaining uses of table properties, and convert them to new high level abstractions that can feasibly be implemented by other engines, then remove the table properties abstractions.

brson:

Issue Number: close #10068

cc https://github.com/tikv/tikv/issues/6402

Problem Summary:

engine_traits contains a bunch of traits encapsulating rocksdb TableProperties and UserCollectedProperties. These are so rocks-specific there is no hope of other engines implementing them. The preferred way to write abstractions that rely on rocks table properties is to add high-level methods that answer questions that, in rocks, are answerable with table properties, letting other engines answer those questions however they can.

This PR removes these traits from engine_traits. It leaves in place the types inside engine_rocks that implemented those traits, just without the trait implementations, as there is still code inside engine_rocks that uses these types. Some support code from engine_traits::properties is moved into engine_rocks::properties_types.

There are only two places outside of engine_rocks that rely on these traits: a test case in sst_importer, and the debug module in tikv. This patch does not address those cases, and they continue to depend concretely on engine_rocks after this patch.

The test case can linger for a while, and I'll address the debug case soon.

### Release note 

```release-note
None
```


brson:
Rebased

brson:
Rebased.

PR #10315, opened 2021-06-07
statistics: fix resource report agent not supporting TLS

zhongzc:

As the resource report agent not supporting TLS, the unified security configuration may fail the reporting.

Problem Summary:

What's Changed:
- Connect to the agent directly instead of with TLS

Besides the changes above, there are several tiny changes out of topic: 
- Change duration related config in resource metering from integer to `ReadableDuration`
- Set channel capacity of report worker to `30`
- Limit count of in-flight reporting grpc to `1`
- Evicted records as others.
- Fix unstable tests due to heavy cpu environment. For the CPU workload at testing, measure CPU time using OS stat instead of inaccurate wall time.

### Check List 

Tests 

- Manual test

### Release note 
```release-note
fix resource report agent not supporting TLS
```


zhongzc:
PTAL @breeswish

components/server/src/server.rs
```diff
@@ -823,8 +823,12 @@ impl<ER: RaftEngine> TiKVServer<ER> {
 
         // Start resource metering.
         let resource_metering_cpu_recorder = resource_metering::cpu::recorder::init_recorder();
-        let mut resource_metering_reporter_worker =
-            Box::new(LazyWorker::new("resource-metering-reporter"));
+        let mut resource_metering_reporter_worker = Box::new(
+            WorkerBuilder::new("resource-metering-reporter")
+                .pending_capacity(30)
```
breeswish, line 828: Does this mean that we will have maximum 30 seconds' data to report when the report remote is blocked? And what will happen if the worker is full?

components/resource_metering/src/cpu/recorder/mod.rs
```diff
@@ -28,20 +28,20 @@ pub struct RecorderHandle {
 pub struct RecorderHandleInner {
     join_handle: JoinHandle<()>,
     pause: Arc<AtomicBool>,
-    precision_seconds: Arc<AtomicU64>,
```
breeswish, line 31: I guess you will find the change to be less (and better) if you accept `ReadableDuration` in config and accept milliseconds in `ResourceMetering/*`, which can be fit in an atomic variable instead of must use a Mutex.

components/server/src/server.rs
zhongzc: Basically, nothing will block the reporter worker except the runtime is busy, so it's hard to be full. If full, new coming per-second records are dropped.

The RPC call is spawned to grpc runtime at every tick (/minute). It's limited to be at most one RPC call is in-flight at this PR. If the report remote is blocked, it won't call RPC at later ticks.

components/resource_metering/src/cpu/recorder/mod.rs
zhongzc: addressed

BusyJay:
> As the resource report agent not supporting TLS, the unified security configuration may fail the reporting.

Why not make the agent accept TLS connections?

components/resource_metering/src/reporter.rs
```diff
@@ -192,6 +209,22 @@ impl RunnableWithTimer for ResourceMeteringReporter {
...
                         }
+
+                        // others
+                        let mut timestamp_list = others.keys().cloned().collect::<Vec<_>>();
+                        timestamp_list.sort_unstable();
+                        let cpu_time_ms_list = timestamp_list
+                            .iter()
+                            .map(|ts| others.get(ts).unwrap())
+                            .cloned()
+                            .collect::<Vec<_>>();
```
breeswish, line 220: Maybe no need to sort it!

components/resource_metering/src/reporter.rs
```diff
@@ -159,7 +156,16 @@ impl Runnable for ResourceMeteringReporter {
...
-                    self.records.retain(|_, (_, _, total)| *total >= kth);
+                    let others = &mut self.others;
+                    self.records.retain(|_, (secs_list, cpu_time_list, total)| {
+                        let retain = *total >= kth;
+                        if !retain {
+                            for (secs, cpu_time) in secs_list.iter().zip(cpu_time_list.iter()) {
+                                *others.entry(*secs).or_insert(0) += *cpu_time;
+                            }
+                        }
+                        retain
```
breeswish, line 167: How about using [`drain_filter`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter), which seems to be, more functional?

breeswish (approved):
Generally LGTM

breeswish:
> > As the resource report agent not supporting TLS, the unified security configuration may fail the reporting.
> 
> Why not make the agent accept TLS connections?

In Sprint3 the Agent will be spawn with the instance using a Daemonset (in TiDB Cloud) so that no TLS is needed. 

We plan to add TLS back in Sprint4 (in both TiKV and TiDB) --- The connection to the agent should be able to work using a dedicated TLS config, for example, using a different certificate compared to the intra-cluster certificate, considering that the connection only happens between a TiKV instance and an agent, instead of any arbitrary instances.

components/resource_metering/src/reporter.rs
zhongzc: How someone come up with such a marvellous API...

breeswish:
> > using a different certificate compared to the intra-cluster certificate
> 
> This is a bad idea comparing to the maintenance effort and no extra security guaranteed is added.

Actually using another certificate is safer, since it narrows the attack surface. When agent certificate is compromised it will only affect the connection between the TiKV and the agent, while keeping the rest of the connection safe.

BusyJay:
> When agent certificate is compromised it will only affect the connection between the TiKV and the agent, while keeping the rest of the connection safe.

This doesn't require to use another certificate. Only use a certificate with different CN for agent itself is enough. TiKV only needs to keep use its own same certificate with CN validation.

breeswish:
> > When agent certificate is compromised it will only affect the connection between the TiKV and the agent, while keeping the rest of the connection safe.
> 
> This doesn't require to use another certificate. Only use a certificate with different CN for agent itself is enough. TiKV only needs to keep use its own same certificate with CN validation.

Nice catch! This mean that we do need some different security verification configurations (e.g. verifying different CN) when connecting to an agent server, but it will not be necessary for us to use different certificates. We will consider this when we add TLS support back :) For now (in this PR), we are offering an option to not use TLS for connection with the agent while the intra-TiKV connection is TLS protected, as a MVP.

breeswish (approved):

breeswish:
@zhongzc Please find another reviewer for approvals.

andylokandy (approved):

PR #10326, opened 2021-06-08
raftstore: introduce asynchronous write io

gengliqi:

Problem Summary:
This PR is split from https://github.com/tikv/tikv/pull/10289.

What's Changed:
Introduce async io code first for review easily.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No

### Release note 

```release-note
None
```


Cargo.lock
```diff
@@ -2524,16 +2524,6 @@ dependencies = [
  "syn",
 ]
 
-[[package]]
-name = "mimalloc-sys"
```
BusyJay, line 2528: Why remove this?

components/engine_rocks/src/write_batch.rs
```diff
@@ -109,6 +113,10 @@ impl engine_traits::WriteBatch<RocksEngine> for RocksWriteBatch {
     fn rollback_to_save_point(&mut self) -> Result<()> {
         self.wb.rollback_to_save_point().map_err(Error::Engine)
     }
+
+    fn merge(&mut self, src: &Self) {
+        self.merge(src);
```
BusyJay, line 118: How about moving L68~L70 here? This implementation makes it look like indefinitely recursive calls.

components/engine_rocks/src/write_batch.rs
```diff
@@ -263,6 +271,10 @@ impl engine_traits::WriteBatch<RocksEngine> for RocksWriteBatchVec {
         }
         Err(Error::Engine("no save point".into()))
     }
+
+    fn merge(&mut self, _: &Self) {
+        panic!("merge is not implemented for RocksWriteBatchVec");
```
BusyJay, line 276: This can be easily implemented. /cc @Little-Wallace

components/engine_traits/src/raft_engine.rs
```diff
@@ -91,7 +91,11 @@ pub trait RaftLogBatch: Send {
 
     fn put_raft_state(&mut self, raft_group_id: u64, state: &RaftLocalState) -> Result<()>;
 
+    fn persist_size(&self) -> usize;
```
BusyJay, line 94: Please add some comments.

Cargo.lock
gengliqi: It's deleted automatically.

components/raftstore/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 #![feature(min_specialization)]
 #![feature(box_patterns)]
 #![feature(vecdeque_binary_search)]
+#![recursion_limit = "256"]
```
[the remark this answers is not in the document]
gengliqi: Because there are too many metrics in `store/metrics.rs`. The recursion limit is reached while expanding `lazy_static!`.

components/engine_rocks/src/write_batch.rs
gengliqi: I implemented it before https://github.com/gengliqi/tikv/blob/raftstore-async-io-tmp/components/engine_rocks/src/write_batch.rs#L174-L260. I think the change is not very small so I prefer to implement in later PR. Also, there is no side effect because the raft db can not use `RocksWriteBatchVec`.

components/raftstore/src/store/msg.rs
```diff
@@ -438,14 +438,17 @@ pub enum PeerMsg<EK: KvEngine> {
...
+        res: ApplyTaskRes<EK::Snapshot>,
+    },
     /// Message that can't be lost but rarely created. If they are lost, real bad
     /// things happen like some peers will be considered dead in the group.
     SignificantMsg(SignificantMsg<EK::Snapshot>),
     /// Start the FSM.
     Start,
     /// A message only used to notify a peer.
     Noop,
+    Persisted((u64, u64, Instant)),
```
BusyJay, line 451: How about using anonymous struct? So we can know what's each field.

components/raftstore/src/store/peer.rs
```diff
@@ -1072,46 +1072,19 @@ where
...
+            if msg_type == MessageType::MsgTimeoutNow && self.is_leader() {
+                // After a leader transfer procedure is triggered, the lease for
+                // the old leader may be expired earlier than usual, since a new leader
+                // may be elected and the old leader doesn't step down due to
+                // network partition from the new leader.
+                // For lease safety during leader transfer, transit `leader_lease`
+                // to suspect.
+                self.leader_lease.suspect(monotonic_raw_now());
+            }
+            if self.send_raft_message(msg, trans) {
```
BusyJay, line 1084: How about
```
let b = self.send_raft_message(msg, trans);
metrics.add(msg_type, b);
```

components/raftstore/src/store/peer.rs
```diff
@@ -1072,46 +1072,19 @@ where
     {
         for msg in msgs {
             let msg_type = msg.get_msg_type();
-            let snapshot_index = msg.get_request_snapshot();
-            let i = self.send_raft_message(msg, trans) as usize;
-            match msg_type {
-                MessageType::MsgAppend => metrics.append[i] += 1,
-                MessageType::MsgAppendResponse => {
-                    if snapshot_index != raft::INVALID_INDEX {
-                        metrics.request_snapshot[i] += 1;
```
BusyJay, line 1081: Why this is removed?

gengliqi: Because this msg is never used. If I keep it, the `add` function of `RaftSendMessageMetrics` must add one more argument. I think it's better to use a new msg type for this msg in the future instead of a special judge.

innerr:
This PR has a clean shot, I think it could be merged a lot easier than before

gengliqi:
PTAL

components/raftstore/src/store/async_io.rs
```diff
@@ -0,0 +1,949 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+// TODO: move it
+#![allow(dead_code)]
```
BusyJay, line 4: How about splitting the actual implementation and tests into two files?

components/raftstore/src/store/metrics.rs
```diff
@@ -315,8 +369,7 @@ lazy_static! {
         register_histogram!(
             "tikv_raftstore_propose_log_size",
             "Bucketed histogram of peer proposing log size.",
-            vec![256.0, 512.0, 1024.0, 4096.0, 65536.0, 262144.0, 524288.0, 1048576.0,
-                    2097152.0, 4194304.0, 8388608.0, 16777216.0]
+            exponential_buckets(0.5, 2.0, 26).unwrap()
```
BusyJay, line 372: Why start from 0.5?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -918,6 +923,19 @@ where
         }
     }
 
+    fn on_persisted_msg(&mut self, peer_id: u64, ready_number: u64, _send_time: Instant) {
+        if peer_id != self.fsm.peer_id() {
+            error!(
```
BusyJay, line 928: Can this happen?

components/raftstore/src/store/config.rs
```diff
@@ -172,6 +175,13 @@ pub struct Config {
     #[online_config(hidden)]
     pub apply_yield_duration: ReadableDuration,
 
+    #[serde(with = "engine_config::perf_level_serde")]
+    #[online_config(skip)]
+    pub perf_level: PerfLevel,
+
+    #[online_config(skip)]
+    pub trigger_write_size: ReadableSize,
```
BusyJay, line 183: Should add comment describing what it is. And etc/config-template.toml should be updated too.

components/raftstore/src/store/config.rs
```diff
@@ -250,16 +257,18 @@ impl Default for Config {
             local_read_batch_size: 1024,
             apply_batch_system: BatchSystemConfig::default(),
             store_batch_system: BatchSystemConfig::default(),
+            store_io_pool_size: 2,
             future_poll_size: 1,
             hibernate_regions: true,
             dev_assert: false,
             apply_yield_duration: ReadableDuration::millis(500),
+            perf_level: PerfLevel::Disable,
```
BusyJay, line 265: I think this can be enabled by default given that perf level is always enabled for read already. /cc @Little-Wallace

components/raftstore/src/store/fsm/peer.rs
gengliqi: It should not happen by design because the region should be destroyed after all io task has finished. It's used to defend against some unexpected bugs in my opinion.

components/raft_log_engine/src/engine.rs
```diff
@@ -74,9 +74,17 @@ impl RaftLogBatchTrait for RaftLogBatch {
...
+        panic!("persist_size is not implemented for raft engine");
+    }
+
     fn is_empty(&self) -> bool {
         self.0.items.is_empty()
     }
+
+    fn merge(&mut self, _: Self) {
+        panic!("merge is not implemented for raft engine");
+    }
```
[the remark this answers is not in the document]
gengliqi: @hicqu friendly ping.

components/raftstore/src/store/fsm/peer.rs
BusyJay: Then how about panicking?

components/raftstore/src/store/async_io.rs
gengliqi: I don't get it. The tests in this file are unit tests.

components/raftstore/src/store/metrics.rs
gengliqi: It's too small. I change it to 8.

components/raftstore/src/store/config.rs
gengliqi: Updated. I change its name to `raft_write_size_limit`.

components/raftstore/src/store/fsm/peer.rs
gengliqi: Maybe too strict? It seems no harm to ignore this msg.

BusyJay: I was thinking about mismatched messages between regions.

components/raftstore/src/store/async_io.rs
BusyJay: Yes, but they can still be split into other files. https://github.com/tikv/tikv/blob/master/src/storage/txn/actions/tests.rs is an example. The intention is to avoid too long files.

components/raftstore/src/store/config.rs
```diff
@@ -250,16 +258,18 @@ impl Default for Config {
             local_read_batch_size: 1024,
             apply_batch_system: BatchSystemConfig::default(),
             store_batch_system: BatchSystemConfig::default(),
+            store_io_pool_size: 2,
```
NingLin-P, line 261: With multiple io threads, is it possible that raft logs from the same region are appending concurrently? And cause log appending out of order like some logs (with larger index) are appended before the log (with smaller index) in front of them.

NingLin-P:
Instead of adding a new simple thread pool for log appending io, why not reuse the `BatchSystem` mechanism to append log just like how we apply log, there may have a lot of logic that we can resue.

components/raft_log_engine/src/engine.rs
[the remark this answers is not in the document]
hicqu: `persist_size` can be calculated with `protobuf::Message::compute_size`. For `Entry` it won't be a large expense

components/raftstore/src/store/async_io.rs
```diff
@@ -0,0 +1,949 @@
...
+where
+    EK: KvEngine,
+    ER: RaftEngine,
+{
+    pub fn new(router: RaftRouter<EK, ER>) -> Self {
+        Self { router }
+    }
+}
+
+impl<EK, ER> Notifier for RegionNotifier<EK, ER>
```
hicqu, line 57: Why not implement `Notifier` for `RaftRouter` directly?

components/raftstore/src/store/async_io.rs
```diff
@@ -0,0 +1,949 @@
...
+        !(self.raft_state.is_none()
+            && self.entries.is_empty()
+            && self.cut_logs.is_none()
+            && self.kv_wb.as_ref().map_or(true, |wb| wb.is_empty())
+            && self.raft_wb.as_ref().map_or(true, |wb| wb.is_empty()))
+    }
+
+    /// Sanity check for robustness.
+    pub fn valid(&self) -> Result<()> {
+        if self.region_id == 0 || self.peer_id == 0 || self.ready_number == 0 {
```
hicqu, line 157: Why not `assert` or `debug_assert`? I think it can't recover from this error.

components/raftstore/src/store/async_io.rs
```diff
@@ -0,0 +1,949 @@
...
+            return Err(box_err!(
+                "invalid id, region_id {}, peer_id {}, ready_number {}",
+                self.region_id,
+                self.peer_id,
+                self.ready_number
+            ));
+        }
+        let last_index = self.entries.last().map_or(0, |e| e.get_index());
+        if let Some((from, _)) = self.cut_logs {
+            if from <= last_index {
```
hicqu, line 167: Should it be `assert_eq!(from, last_index + 1)`?

components/raftstore/src/store/async_io.rs
```diff
@@ -0,0 +1,949 @@
...
+{
+    pub kv_wb: EK::WriteBatch,
+    pub raft_wb: ER::LogBatch,
+    // region_id -> (peer_id, ready_number)
+    pub readies: HashMap<u64, (u64, u64)>,
+    // Write raft state once for a region everytime writing to disk
+    pub raft_states: HashMap<u64, RaftLocalState>,
+    pub state_size: usize,
+    pub tasks: Vec<AsyncWriteTask<EK, ER>>,
+    _phantom: PhantomData<EK>,
```
hicqu, line 216: Is it necessary? `tasks` already has the generic arguments.

components/raftstore/src/store/config.rs
gengliqi: It's not possible. The io tasks must be handled in sequence.

gengliqi:
> Instead of adding a new simple thread pool for log appending io, why not reuse the `BatchSystem` mechanism to append log just like how we apply log, there may have a lot of logic that we can resue.

I don't reuse the `BatchSystem` because it may introduce a reorder of io tasks which may increase tail latency.

components/raftstore/src/store/async_io.rs
gengliqi: Good catch. These codes here have changed many times. I forget to remove it.

components/raftstore/src/store/async_io.rs
gengliqi: The caller will panic if this function returns an error.

components/raftstore/src/store/async_io.rs
gengliqi: Makes sense🤣.

BusyJay:
How to handle write imbalance? For example, two hot spot are scheduled to the same thread.

NingLin-P:
> > Instead of adding a new simple thread pool for log appending io, why not reuse the `BatchSystem` mechanism to append log just like how we apply log, there may have a lot of logic that we can resue.
> 
> I don't reuse the `BatchSystem` because it may introduce a reorder of io tasks which may increase tail latency.

Binding a region's io task to a specified thread could also introduce reordering. For example, a later coming io task could be handled immediately while an early coming io task must wait for all previous tasks on the same thread to finish before start.

gengliqi:
> How to handle write imbalance? For example, two hot spot are scheduled to the same thread.

For now, it depends on random to decrease the probability of imbalance. Also, the CPU work of io thread is small so the imbalance should not have a great influence.
We can optimize it in the future if it really has this problem. For example, if a region has no pending io work, it can choose a thread that has the least amount of tasks.

gengliqi:
> > > Instead of adding a new simple thread pool for log appending io, why not reuse the `BatchSystem` mechanism to append log just like how we apply log, there may have a lot of logic that we can resue.
> > 
> > 
> > I don't reuse the `BatchSystem` because it may introduce a reorder of io tasks which may increase tail latency.
> 
> Binding a region's io task to a specified thread could also introduce reordering. For example, a later coming io task could be handled immediately while an early coming io task must wait for all previous tasks on the same thread to finish before start.

As long as there are many queues, reorder events will always exist. However, in my opinion, if the number of queues is smaller, the reorder events should be smaller too. Here the number of queues is the thread number of io thread. It is the number of regions when using batch system. Obviously, the latter is much greater than the former.

BusyJay:
> Also, the CPU work of io thread should be small so the imbalance should not greatly influence performance.

What about IO balance? Any tests to show the potential delay? And anyway to get around the problem?

gengliqi:
> > Also, the CPU work of io thread should be small so the imbalance should not greatly influence performance.
> 
> What about IO balance? Any tests to show the potential delay? And anyway to get around the problem?

There is only one thread calling write syscall no matter in rocksdb or raft engine.

BusyJay:
Then why bother using two threads? There should be something that can utilize threaded pool and hence can cause imbalance. What I'm concerned is how it performs when most writes are scheduled to one thread.

gengliqi:
> Then why bother using two threads? There should be something that can utilize threaded pool and hence can cause imbalance. What I'm concerned is how it performs when most writes are scheduled to one thread.

I believe the probability of this corner case is low enough because of random distribution.

BusyJay:
Why? Let's say if not less than 2/3 of hotspots are scheduled to one thread is imbalance, then the probability is 0.5 for 2 hotspots, and  0.25 for 3 hotspots, 0.5 for 4 hotspots.

gengliqi:
PTAL

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+    pub fn valid(&self) -> Result<()> {
+        if self.region_id == 0 || self.peer_id == 0 || self.ready_number == 0 {
+            return Err(box_err!(
+                "invalid id, region_id {}, peer_id {}, ready_number {}",
+                self.region_id,
+                self.peer_id,
+                self.ready_number
+            ));
+        }
+        let last_index = self.entries.last().map_or(0, |e| e.get_index());
```
5kbpers, line 150: Should we skip succeeding checks if no entry?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+                for t in &task.request_times {
+                    metrics
+                        .kvdb_end
+                        .observe(duration_to_sec(now.saturating_duration_since(*t)));
+                }
+            }
+        }
+    }
+
+    fn after_write_to_db(&mut self, metrics: &StoreWriteMetrics) {
```
5kbpers, line 314:
```suggestion
    fn after_write_to_raft_db(&mut self, metrics: &StoreWriteMetrics) {
```

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+                                .observe(duration_to_sec(loop_begin.saturating_elapsed()));
+                            handle_begin = Instant::now();
+                            msg
+                        }
+                        Err(_) => return,
+                    }
+                } else {
+                    match self.receiver.try_recv() {
+                        Ok(msg) => msg,
+                        Err(TryRecvError::Empty) => break,
```
5kbpers, line 428: Should we need to block the worker if there is not any pending write.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+                    );
+                });
+            self.perf_context.report_metrics();
+            STORE_WRITE_RAFTDB_DURATION_HISTOGRAM
+                .observe(duration_to_sec(now.saturating_elapsed()) as f64);
+        }
+
+        self.batch.after_write_to_db(&self.metrics);
+
+        self.last_unpersisted_id += 1;
```
5kbpers, line 530: Why plus one here? `self.last_persisted_id < self.global_persisted_id.load(Ordering::Acquire)` needs to wait for global_persisted_id was increased by 2?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+                let msg = if first_time {
+                    first_time = false;
+                    match self.receiver.recv() {
+                        Ok(msg) => msg,
+                        Err(_) => return,
+                    }
+                } else {
+                    match self.receiver.try_recv() {
+                        Ok(msg) => msg,
+                        Err(TryRecvError::Empty) => break,
```
5kbpers, line 679: (ditto)

BusyJay:
Can this PR be split into smaller ones?

components/raftstore/src/store/async_io/write.rs
gengliqi: It will. Next time it will block on the `recv` function if no new msg.

components/raftstore/src/store/async_io/write.rs
gengliqi: This is `last_unpersisted_id` not `last_persisted_id`. `last_unpersisted_id` is just a monotonic increasing counter and is used for recording the last unpersisted id.

components/raftstore/src/store/async_io/write.rs
gengliqi: (ditto)

gengliqi:
> Can this PR be split into smaller ones?

Some new metrics can be split into another PR but I think these codes are simple and can be read quickly.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -407,14 +408,22 @@ where
...
             if self.callbacks.len() == 1 {
-                let (cb, _) = self.callbacks.pop().unwrap();
+                let (mut cb, _, send_time) = self.callbacks.pop().unwrap();
+                if let Callback::Write { request_times, .. } = &mut cb {
+                    if metric.waterfall_metrics {
+                        metric
+                            .batch_wait
+                            .observe(send_time.saturating_elapsed_secs());
+                    }
+                    *request_times = vec![send_time];
```
BusyJay, line 422: There seems to be always only one time.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -547,9 +554,18 @@ where
         self.host
             .on_flush_applied_cmd_batch(batch_max_level, cmd_batch, &self.engine);
         // Invoke callbacks
+        let now = Instant::now();
         for (cb, resp) in cb_batch.drain(..) {
-            cb.invoke_with_response(resp)
+            if let Some(times) = cb.get_request_times() {
```
BusyJay, line 559: Shouldn't that be controlled by waterfall metrics?

components/raftstore/src/store/fsm/peer.rs
gengliqi: If some requests are batched into one cmd, `request_times` should have many times.

components/raftstore/src/store/fsm/apply.rs
gengliqi: No. The `store_time` and `apply_time` metrics are open by default. The sum of their avg value should be equal to `async_write_time`.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
```
BusyJay, line 2: Can you add module docs to explain how data are processed in different structure of this file?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+            write_senders,
+            global_persisted_ids,
+        }
+    }
+
+    fn run(&mut self) {
+        let mut persist_vec = vec![];
+        for _ in 0..self.write_senders.len() {
+            persist_vec.push(None);
+        }
```
NingLin-P, line 665:
```suggestion
        let mut persist_vec = vec![None; self.write_senders.len()];
```

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+                cfg,
+            );
+            let t = thread::Builder::new().name(thd_name!(tag)).spawn(move || {
+                worker.run();
+            })?;
+            global_persisted_ids.push(global_persisted_id);
+            self.writers.push(tx);
+            self.handlers.push(t);
+        }
+        let mut syncer = SyncWorker::new(
```
NingLin-P, line 775: What are the benefits to add an extra `SyncWorker`? IMO there is no difference to queueing write task in `channel` or  `Worker:: pending_tasks` and having a single sync worker and multiple write worker may cause a task need to wait for all previous tasks from all worker to finish before start.

components/raftstore/src/store/msg.rs
```diff
@@ -76,6 +76,7 @@ pub enum Callback<S: Snapshot> {
         /// `committed_cb` is called after a request is committed and before it's being applied, and
         /// it's guaranteed that the request will be successfully applied soon.
         committed_cb: Option<ExtCallback>,
+        request_times: Vec<Instant>,
```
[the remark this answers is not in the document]
gengliqi: Done.

components/raftstore/src/store/config.rs
```diff
@@ -203,6 +194,29 @@ pub struct Config {
     // * system=16G, memory_usage_limit=12G, evict=2.4G
     // * system=32G, memory_usage_limit=24G, evict=4.8G
     pub evict_cache_on_memory_ratio: f64,
+
+    pub cmd_batch: bool,
```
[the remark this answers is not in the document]
gengliqi: Yes. But this code change is very small. It's just used for tests and it should be always true in production.

components/raftstore/src/store/config.rs
```diff
@@ -203,6 +194,29 @@ pub struct Config {
...
+
+    // When the size of raft db writebatch exceeds this value, write will be triggered.
+    pub raft_write_size_limit: ReadableSize,
+
+    pub store_waterfall_metrics: bool,
```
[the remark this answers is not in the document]
gengliqi: Because the waterfall metrics are only located in store threads and store io threads for now. Remove `store` prefix is also LGTM. I have removed it.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,839 @@
...
+use tikv_util::time::{duration_to_sec, Instant};
+use tikv_util::{box_err, debug, thd_name, warn};
+
+const KV_WB_SHRINK_SIZE: usize = 1024 * 1024;
+const KV_WB_DEFAULT_SIZE: usize = 16 * 1024;
+const RAFT_WB_SHRINK_SIZE: usize = 10 * 1024 * 1024;
+const RAFT_WB_DEFAULT_SIZE: usize = 256 * 1024;
+
+/// Notify the event to the specified region.
+pub trait Notifier: Clone + Send + 'static {
```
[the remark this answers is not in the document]
gengliqi: These traits and message structs are related to the implementation of `Worker` and `SyncWorker`.

components/raftstore/src/store/async_io/write.rs
gengliqi: There is only one `fsync` at the same time, which is the same as no `SyncWorker`. So there is no extra wait time. 
The benefit is that the multiple `Worker` can do more cpu workers like `batch WriteTask`, `send callback to peer` and `send msg` instead of waiting for `fsync`.

NingLin-P: What I mean is that it may trigger more `fsync` than before because the `WriteWorker` has less time of blocking io, the batch size may reduce hence trigger more `SyncMsg::Sync` and wait more `fsync`.

Also, we may wait for more blocking io operations to send messages to other peers than before:
- Without `SyncWorker`: `prev_write` + `prev_fsync` + `write` + `fsync` -> send msg
- Wirth `SyncWorker`: `prev_write` + `write` + `prev_fsync` + `fsync` + `later_coming_write` -> send msg

gengliqi:
I remove the sync worker for now. Remove about 400 lines of codes. PTAL again, thanks!

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,640 @@
...
+            message_metrics: Default::default(),
+            perf_context,
+        }
+    }
+
+    fn run(&mut self) {
+        loop {
+            let loop_begin = Instant::now();
+            let mut handle_begin = loop_begin;
+
```
NingLin-P, line 389:
```suggestion
```
Looks unneeded.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,640 @@
...
+                    }
+                } else {
+                    match self.receiver.try_recv() {
+                        Ok(msg) => msg,
+                        Err(TryRecvError::Empty) => break,
+                        Err(TryRecvError::Disconnected) => return,
+                    }
+                };
+                match msg {
+                    WriteMsg::Shutdown => return,
```
NingLin-P, line 409: Should we add some graceful shutdown mechanisms? like handle all received tasks before shutdown.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,640 @@
...
+    EK: KvEngine,
+    ER: RaftEngine,
+{
+    region_id: u64,
+    peer_id: u64,
+    ready_number: u64,
+    pub send_time: Instant,
+    pub kv_wb: Option<EK::WriteBatch>,
+    pub raft_wb: Option<ER::LogBatch>,
+    pub entries: Vec<Entry>,
```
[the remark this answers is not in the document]
gengliqi: If so, it will introduce an extra copy cost.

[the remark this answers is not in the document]
NingLin-P: How comes?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,640 @@
...
+        self.state_size + self.raft_wb.persist_size()
+    }
+
+    fn before_write_to_db(&mut self, metrics: &StoreWriteMetrics) {
+        // Put raft state to raft writebatch
+        let raft_states = std::mem::take(&mut self.raft_states);
+        for (region_id, state) in raft_states {
+            self.raft_wb.put_raft_state(region_id, &state).unwrap();
+        }
+        self.state_size = 0;
```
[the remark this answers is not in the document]
gengliqi: I think it's not necessary to make these functions all about metrics? This logic is all about `WriteTaskBatch` so it is reasonable to locate in its member function.

components/raftstore/src/store/async_io/write.rs
gengliqi: Make sense.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,649 @@
...
+        }
+        let now2 = Instant::now();
+        STORE_WRITE_SEND_DURATION_HISTOGRAM
+            .observe(duration_to_sec(now2.saturating_duration_since(now)));
+
+        for (region_id, (peer_id, ready_number)) in &self.batch.readies {
+            self.notifier
+                .notify_persisted(*region_id, *peer_id, *ready_number, now);
+        }
+        STORE_WRITE_CALLBACK_DURATION_HISTOGRAM.observe(duration_to_sec(now2.saturating_elapsed()));
```
BusyJay, line 545: It seems to only send several messages, is it necessary?

components/raftstore/src/store/async_io/write.rs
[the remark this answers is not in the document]
gengliqi: Copy the `WriteTask::raft_wb` to the `WriteTaskBatch::raft_wb`.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,649 @@
...
+                        "target_store_id" => to_store_id,
+                        "err" => ?e,
+                        "error_code" => %e.error_code(),
+                    );
+                    self.message_metrics.add(msg_type, false);
+                    // Send unreachable to this peer.
+                    // If this msg is snapshot, it is unnecessary to send snapshot
+                    // status to this peer because it has already become follower.
+                    // (otherwise the snapshot msg should be sent in store thread other than here)
+                    unreachable_peers.insert((task.region_id, to_peer_id));
```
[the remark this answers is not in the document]
gengliqi: There is another case raft client doesn't handle: If the queue is full, the msg is dropped. In original codes, the peer will call `raft_group.report_unreachable`. If this case is not necessary to report unreachable, these codes can be removed and the send msg part in store thread should be updated too.

components/raftstore/src/store/async_io/write.rs
gengliqi: Sending persisted msgs is not very fast. In sysbench insert workload tests, the time it takes is about half of handling write tasks, 1.5 times of sending raft msgs.

components/raftstore/src/store/async_io/write.rs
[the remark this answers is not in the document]
NingLin-P: `WriteBatch::merge` take `raft_wb` not `&raft_wb`, I think it should be a `move` semantics.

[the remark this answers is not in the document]
gengliqi: No, it's not. The underlying data structure of rocksdb's `WriteBatch` is a string.

components/raftstore/src/store/async_io/write.rs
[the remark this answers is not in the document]
BusyJay: I think the code in store thread is still necessary. Only leader can adjust flow control, and leader will always send messages before writing.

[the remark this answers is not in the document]
gengliqi: Addressed.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,635 @@
...
+                store_id,
+                tag.clone(),
+                engines.clone(),
+                rx,
+                notifier.clone(),
+                trans.clone(),
+                cfg,
+            );
+            let t = thread::Builder::new().name(thd_name!(tag)).spawn(move || {
+                worker.run();
```
NingLin-P, line 586: How about adding some logs when a worker is started or stopped.

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,635 @@
...
+        if let Some(raft_state) = task.raft_state.take() {
+            if self
+                .raft_states
+                .insert(task.region_id, raft_state)
+                .is_none()
+            {
+                self.state_size += std::mem::size_of::<RaftLocalState>();
+            }
+        }
+        self.readies
```
BusyJay, line 239: What if different writes from different peers are batched together?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,635 @@
...
+
+    #[inline]
+    fn get_raft_size(&self) -> usize {
+        self.state_size + self.raft_wb.persist_size()
+    }
+
+    fn before_write_to_db(&mut self, metrics: &StoreWriteMetrics) {
+        // Put raft state to raft writebatch
+        let raft_states = std::mem::take(&mut self.raft_states);
+        for (region_id, state) in raft_states {
```
BusyJay, line 266: Why not use `self.raft_states.drain()`?

components/raftstore/src/store/async_io/write_tests.rs
```diff
@@ -0,0 +1,400 @@
...
+use crossbeam::channel::unbounded;
+use engine_rocks::RocksWriteBatch;
+use engine_test::kv::{new_engine, KvTestEngine};
+use engine_traits::{Mutable, Peekable, WriteBatchExt, ALL_CFS};
+use kvproto::raft_serverpb::RaftMessage;
+use tempfile::Builder;
+
+use super::*;
+
+fn create_engines(path: &str) -> Engines<KvTestEngine, KvTestEngine> {
```
BusyJay, line 18: Why not use `engine_test::new_temp_engine()`?

components/raftstore/src/store/async_io/write.rs
gengliqi: Done

components/raftstore/src/store/async_io/write.rs
gengliqi: It should not happen because the peer must be destroyed after all writes are done.

BusyJay: How about checking the returned value to ensure this?

components/raftstore/src/store/async_io/write_tests.rs
```diff
@@ -0,0 +1,387 @@
...
+use engine_test::kv::KvTestEngine;
+use engine_test::new_temp_engine;
+use engine_traits::{Mutable, Peekable, WriteBatchExt};
+use kvproto::raft_serverpb::RaftMessage;
+use tempfile::Builder;
+
+use super::*;
+
+fn create_engines(path: &str) -> Engines<KvTestEngine, KvTestEngine> {
+    let path = Builder::new().prefix(path).tempdir().unwrap();
```
BusyJay, line 20: Directory will be removed when `path` is dropped.

components/raftstore/src/store/async_io/write_tests.rs
```diff
@@ -0,0 +1,387 @@
...
+    task_3.raft_state = Some(new_raft_state(6, 345, 6, 6));
+    task_3
+        .messages
+        .append(&mut vec![RaftMessage::default(), RaftMessage::default()]);
+
+    t.write_sender(0).send(WriteMsg::WriteTask(task_3)).unwrap();
+
+    must_wait_same_notifies(vec![(1, (1, 15)), (2, (2, 20))], &t.notify_rx);
+
+    must_have_entries_and_state(
```
BusyJay, line 368: Should also check if kv engine is updated correctly.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,636 @@
...
+    fn write_to_db(&mut self) {
+        self.batch.before_write_to_db(&self.metrics);
+
+        fail_point!("raft_before_save");
+
+        if !self.batch.kv_wb.is_empty() {
+            let now = Instant::now();
+            let mut write_opts = WriteOptions::new();
+            write_opts.set_sync(true);
+            self.batch.kv_wb.write_opt(&write_opts).unwrap_or_else(|e| {
```
BusyJay, line 445: Should also observe perf context.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -0,0 +1,636 @@
...
+            perf_context,
+        }
+    }
+
+    fn run(&mut self) {
+        let mut stopped = false;
+        while !stopped {
+            let mut handle_begin = None;
+
+            let mut first_time = true;
```
BusyJay, line 374: A more clear way is calling `recv` and then loop `try_recv`.

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/async_io/write.rs
gengliqi: In the original code, these perf metrics are mixed with raft db which is very strange. Maybe we need another `PerfContextKind` to fix it(BTW, destroy process writes both kv db and raft db so it should also be fixed).
I think it's not very urgent. We can fix it in future PR.

components/raftstore/src/store/async_io/write.rs
gengliqi: If so, the logic of handling msg will be written twice?

components/raftstore/src/store/async_io/write.rs
```diff
@@ -236,8 +236,17 @@ where
...
-            .insert(task.region_id, (task.peer_id, task.ready_number));
+        if let Some(prev_readies) = self
+            .readies
+            .insert(task.region_id, (task.peer_id, task.ready_number))
+        {
+            // The peer id must be same if they belong to the same region because
+            // the peer must be destroyed after all write tasks have been finished.
+            assert_eq!(prev_readies.0, task.peer_id);
+            // The ready number should be monotonically increasing.
+            assert!(prev_readies.1 < task.ready_number);
```
BusyJay, line 247: You probably also want region_id or peer_id if it does happen.

components/raftstore/src/store/async_io/write.rs
BusyJay: Handling msg can be a method.

components/raftstore/src/store/async_io/write.rs
BusyJay: Then should add TODO.

components/raftstore/src/store/async_io/write.rs
gengliqi: Done

NingLin-P (approved):

BusyJay (approved):

PR #10333, opened 2021-06-09
raftdb: add support for raft log gc on compaction.

MrCroxx:

Add support for comaction-time raft log gc.

This will relieve write stall problem.

More details on [unistore:pull#529](https://github.com/ngaut/unistore/pull/529)

Use `RaftLogCompactionFilter` to drop old raft log entries.

Add config `raftdb.gc_on_compaction`. When enabled, use compaction filter to gc raft log instead of inserting tombstones to rocksdb.

### Benchmark

- TODO

### Related changes

### Check List 

Tests

- [x] Unit test
- [x] Integration test

Side effects

- No obvious affects on CPU or MEM

### Release note 

```release-note
Add configurable feature: gc raftdb log on compaction.
```


src/server/gc_worker/gc_raftdb.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl RaftLogCompactionFilterFactory {}
+
+impl CompactionFilterFactory for RaftLogCompactionFilterFactory {
+    fn create_compaction_filter(
+        &self,
+        _: &engine_rocks::raw::CompactionFilterContext,
+    ) -> *mut engine_rocks::raw::DBCompactionFilter {
+        let ctx = RAFT_LOG_GC_CONTEXT.read().unwrap();
+        let filter = Box::new(RaftLogCompactionFilter::new(ctx.apply_idxs.clone()));
```
tabokie, line 17: This map could be pretty big, maybe instead of copying it, use a callback?

components/engine_rocks/src/raft_engine.rs
```diff
@@ -233,6 +257,13 @@ impl RaftLogBatch for RocksWriteBatch {
     }
 
     fn cut_logs(&mut self, raft_group_id: u64, from: u64, to: u64) {
+        {
+            let mut ctx = RAFT_LOG_GC_CONTEXT.write().unwrap();
+            if ctx.gc_on_compaction {
```
tabokie, line 262: Seems to me we can avoid a lock when accessing read-only `gc_on_compaction`.

src/server/gc_worker/gc_raftdb.rs
tabokie: Or just implement `CompactionFilter` for the context (`Arc<Context>`), I see the code logic here is quite simple.

MrCroxx: > This map could be pretty big, maybe instead of copying it, use a callback?

I'm afraid I did not get what 'callback' refers here. Do you mean use the ref of the map instead of copying it? I copy the map here to avoid the compaction blocking gc thread to update the map.

tabokie: By callback I means rlock the map during compaction, instead of copying the full volume every time. If blocking GC thread is an issue, you can move the map into filter factory, and keeping only the delta changes in context, which is merged back every time a new filter is created.

MrCroxx: OK, I will consider a better way to record the indexes.

components/engine_rocks/src/raft_engine.rs
```diff
@@ -233,6 +257,13 @@ impl RaftLogBatch for RocksWriteBatch {
...
+                ctx.apply_idxs.insert(raft_group_id, to);
```
Connor1996, line 263: `cut_logs` just delete the entries in the write batch, not the db. Seems shouldn't update gc context here.

Connor1996 (commented):
This will make the log index discontinuous, please make sure there is no logic depending on the assumption.

And need a benchmark to see the effect of this PR.

MrCroxx:
> This will make the log index discontinuous, please make sure there is no logic depending on the assumption.
> 
> And need a benchmark to see the effect of this PR.

@Connor1996  I have checked the code earlier. It's okay to leave gaps in the raft log before where function `gc` gives. 

When `raftstore` implementing `Storage` trait for `PeerStorage`, it use `apply_state.index` as the first remaining raft log instead of the first log in the underlying storage. 

```rust
    fn check_range(&self, low: u64, high: u64) -> raft::Result<()> {
        if low > high {
            return Err(storage_error(format!(
                "low: {} is greater that high: {}",
                low, high
            )));
        } else if low <= self.truncated_index() {
            return Err(RaftError::Store(StorageError::Compacted));
        } else if high > self.last_index() + 1 {
            return Err(storage_error(format!(
                "entries' high {} is out of bound lastindex {}",
                high,
                self.last_index()
            )));
        }
        Ok(())
    }

    #[inline]
    pub fn truncated_index(&self) -> u64 {
        self.apply_state.get_truncated_state().get_index()
    }
```

When applying raft log, `kvdb` will buffer write `ApplyState`, and periodically sync it. If and only if `kvdb` has synced the `ApplyState`, raftstore will run GC task to make sure it's safe to delete any logs before `apply_state.index`.

```rust
fn flush(&mut self) {
        // Sync wal of kv_db to make sure the data before apply_index has been persisted to disk.
        self.engines.kv.sync().unwrap_or_else(|e| {
            panic!("failed to sync kv_engine in raft_log_gc: {:?}", e);
        });
        let tasks = std::mem::take(&mut self.tasks);
        for t in tasks {
            match t {
                Task::Gc {
                    region_id,
                    start_idx,
                    end_idx,
                } => {
                    debug!("gc raft log"; "region_id" => region_id, "end_index" => end_idx);
                    match self.gc_raft_log(region_id, start_idx, end_idx) {
                        Err(e) => {
                            error!("failed to gc"; "region_id" => region_id, "err" => %e);
                            self.report_collected(0);
                        }
                        Ok(n) => {
                            debug!("gc log entries"; "region_id" => region_id, "entry_count" => n);
                            self.report_collected(n);
                        }
                    }
                }
        // ... ...
```

@Connor1996  PTAL, I'm not sure if I missed something.

MrCroxx:
And I'm working on reducing the copy of `map` and the benchmark.

components/engine_rocks/src/raft_engine.rs
MrCroxx: `gc_on_compaction` is a mutable static value, although it will not be changed after parsing the config, a lock is needed to make the compiler happy.

What about make `gc_on_compaction` an AtomicBool to avoid locking?

src/server/gc_worker/gc_raftdb.rs
```diff
@@ -0,0 +1,78 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```
Connor1996, line 1: I don't think it should be put into the gc_worker dir. GC worker is at the level of MVCC which is quite unrelated with raft log.

MrCroxx: I agree with you on it. Considering the option only works when using rocksdb as raft log storage, I think engine_rocks dir is the best place for it.

components/engine_rocks/src/raft_compaction_filter.rs
```diff
@@ -0,0 +1,152 @@
...
+        _context: &CompactionFilterContext,
+    ) -> *mut DBCompactionFilter {
+        let filter = Box::new(RaftLogCompactionFilter::new());
+        let name = CString::new("raft_log_gc_compaction_filter").unwrap();
+        unsafe { new_compaction_filter_raw(name, filter) }
+    }
+}
+
+struct RaftLogCompactionFilter {
+    map: HashMap<u64, u64>,
```
hicqu, line 27: Regions are met one by one, they are sorted. So use a `current_region_id` is enough.

MrCroxx: Get it!

src/config.rs
```diff
@@ -1174,14 +1174,23 @@ impl Default for RaftDefaultCfConfig {
...
-    pub fn build_opt(&self, cache: &Option<Cache>) -> ColumnFamilyOptions {
+    pub fn build_opt(&self, cache: &Option<Cache>, gc_on_compaction: bool) -> ColumnFamilyOptions {
         let no_region_info_accessor: Option<&RegionInfoAccessor> = None;
         let mut cf_opts = build_cf_opt!(self, CF_DEFAULT, cache, no_region_info_accessor);
         let f = Box::new(FixedPrefixSliceTransform::new(region_raft_prefix_len()));
         cf_opts
             .set_memtable_insert_hint_prefix_extractor("RaftPrefixSliceTransform", f)
             .unwrap();
         cf_opts.set_titandb_options(&self.titan.build_opts());
+        if gc_on_compaction {
```
hicqu, line 1185: I prefer to always install the factory. And then examine the `GC_IN_COMPACTION_FILTER` in `create_compaction_filter`. Then this feature can be enabled/disabled dynamically.

src/config.rs
```diff
@@ -1239,6 +1248,8 @@ pub struct RaftDbConfig {
     pub allow_concurrent_memtable_write: bool,
     pub bytes_per_sync: ReadableSize,
     pub wal_bytes_per_sync: ReadableSize,
+    #[online_config(skip)]
+    pub gc_on_compaction: bool,
```
hicqu, line 1252: How about enable dynamic configuration change for it?

components/engine_rocks/src/raft_engine.rs
```diff
@@ -13,6 +18,12 @@ use tikv_util::{box_err, box_try};
 
 const RAFT_LOG_MULTI_GET_CNT: u64 = 8;
 
+lazy_static! {
+    pub static ref RAFT_LOG_GC_ON_COMPACTION: AtomicBool = AtomicBool::new(false);
+    pub static ref RAFT_LOG_GC_INDEXES: Arc<RwLock<HashMap<u64, u64>>> =
```
hicqu, line 23: I'm not sure is there any performance issue here. But I prefer this:
```
impl AdminObserver {
    fn on_create_region(&self) {
        let truncate_index = Arc::new(AtomicU64::new(0));
        RAFT_LOG_GC_INDEXES.write().unwrap().insert(region_id, truncate_index.clone());
        peer.peer_storage.truncate_index = truncate_index;
    }
    fn on_remove_region(&self) {
         ......
    }
}
```
So that the the write lock of the hashmap is only required when creating/removing a Region on a store. Update truncate index for a Region requires no lock.

src/config.rs
MrCroxx: That's a good idea, but some extra work needed. It's safe to dynamically enable/disable the config (not break consistency), but dynamically disabling it may leave some log entries without tombstone and not deleted. An extra machism needed to handle it.

hicqu: If the region is compacted again, those log will be deleted. Seems it's OK?

MrCroxx: Enabling `gc-on-compaction` will stop inserting tombstone when deleting. If the previous config enabled `gc-on-comaction` and then disable it before all log entries between `from` and `to` is deleted by compaction filter, these log entries will not be able to be deleted again (raftstore will track index `to`), expect enable `gc-on-compaction` again.

components/engine_rocks/src/raft_engine.rs
MrCroxx: That's a good idea. I will check `AdminObserver` and implement it.

MrCroxx:
@hicqu I think the main problem with compaction filter is that it's hard to decide when the old log entries are fully deleted for a region. With the fully deleted time, we can make config changing dynamically, and delete the region info from the gc point tracking map when the region is destroyed to save space.

MrCroxx:
@hicqu @tabokie Should we record the range of log entires to delete instead of the gc point only? With that we can have a clear view of which log entries have been deleted / not yet deleted. With the range info we can easily turn on/off `gc-on-compaction` dynamically and remove region gc tracking after the region is deleted and all log entires are deleted. But maintaining the range may cost more mem and cpu (mostly mem).

components/engine_rocks/src/raft_compaction_filter.rs
```diff
@@ -0,0 +1,169 @@
...
+        };
+
+        if current_region != self.last_region {
+            self.last_region = current_region;
+            let indexes = RAFT_LOG_GC_INDEXES.read().unwrap();
+            self.gc_point = indexes.get(&current_region).unwrap_or(&0).to_owned();
+        }
+
+        if current_index < self.gc_point {
+            return CompactionFilterDecision::Remove;
```
hicqu, line 78: It still leave tombstones. Use `Decision::RemoveAndSkipUntil` instead can avoid tombstones.

MrCroxx:
Close. RaftDB compaction filter doesn't work well enough.

PR #10334, opened 2021-06-09
raftstore: unify memory trace and count dangle entries

hicqu:

This PR add a submodule `memtrace` for `peer.rs` and `apply.rs`. All code about memory tracing are put it that submodule so that the main logic becomes clear.

And for entry cache, size of dangle entries is also counted.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note
None
```


components/raftstore/src/store/peer_storage.rs
```diff
@@ -306,11 +307,40 @@ impl EntryCache {
...
+            size
+        };
+
         let old_capacity = self.trace.capacity();
         self.trace.push_back(entries);
         let new_capacity = self.trace.capacity();
         let diff = Self::get_trace_vec_mem_size_change(new_capacity, old_capacity);
-        self.flush_mem_size_change(diff);
+
+        self.flush_mem_size_change(diff + dangle_size as i64);
```
5kbpers, line 343: Should we need to subtract `dangle_size` here? As I understand it is actually dropped.

components/raftstore/src/store/peer.rs
```diff
@@ -3851,6 +3833,39 @@ pub trait AbstractPeer {
...
+    {
+        pub fn proposal_size(&self) -> usize {
+            let mut heap_size = self.pending_reads.heap_size();
+            for prop in &self.proposals.queue {
+                heap_size += prop.heap_size();
+            }
+            heap_size
+        }
+
+        pub fn rest_size(&self) -> usize {
```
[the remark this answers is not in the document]
hicqu: Yes. And lot of recursive heap content is omited. The result is very coarse.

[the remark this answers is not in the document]
hicqu: Do you have any idea? Or just remove it?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -771,8 +762,8 @@ where
     /// are all handled.
     pending_msgs: Vec<Msg<EK>>,
 
-    /// Approximate size of current buffer.
-    approximate_size: usize,
+    /// Cache heap size for itself.
+    heap_size: Option<usize>,
```
rleungx, line 766: Why do we need an option here?

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
rleungx: No. I don't know either.

components/raftstore/src/store/fsm/apply.rs
hicqu: Indicates whether it has been already calculated or not. It's possible that it's `Some(0)`, for which case we don't need to do the calculation again.

components/raftstore/src/store/peer_storage.rs
hicqu: Entries won't be put into `EntryCache` when they are fetched from `Storage`. That's the reason for calling it `dangle`. So here we trace the size of dangle entries in `EntryCache`. And then if they are evicted later, we can substract the size.

components/raftstore/src/store/peer.rs
```diff
@@ -1943,11 +1925,11 @@ where
                 committed_entries,
                 cbs,
             );
+            self.mut_store().trace_cached_entries(apply.entries.clone());
```
rleungx, line 1928: Why do we need to change it?

hicqu: `trace_cached_entries` must be called before `evict_cache`. Otherwise `evict` can't release dangle entries (entries not in the entry cache).

rleungx (approved):

5kbpers (approved):
LGTM

components/raftstore/src/store/peer_storage.rs
hhwyt: Why should the entry cache size account for the dangle size? Why not separately track the size of entries sent to apply, and treating apply entries as including the dangle size?

hhwyt: @hicqu

PR #10336, opened 2021-06-09
statitics: send original string value for sampling data

winoros:

Compabiliable change with https://github.com/pingcap/tidb/pull/25311

Problem Summary:

TiDB still needs the original data instead of the collation key.

Proposal: [xxx](url) 

What's Changed:

This pr sends the original string value instead of its collation key to TiDB.
This also saves the encoding CPU costs.

### Related changes

### Check List 

Tests 

- Integration test

### Release Note
```
None
```


src/coprocessor/statistics/analyze.rs
```diff
@@ -461,19 +470,33 @@ impl RowSampleCollector {
...
             }
-            self.fm_sketches[i].insert(&columns_val[i]);
+            if columns_info[i].as_accessor().is_string_like() {
+                self.fm_sketches[i].insert(&collation_keys_val[i]);
+            } else {
+                self.fm_sketches[i].insert(&columns_val[i]);
+            }
             self.total_sizes[i] += columns_val[i].len() as i64;
         }
         self.sampling(columns_val);
```
skyzh, line 502: So we are sampling on `columns_val` and generating other statistics on `collation_keys_val`?

skyzh (commented):
I doubt whether it is necessary to use two separate vectors for `column_vals` and `collation_key_vals`. As they are both of type `Vec<Vec<u8>>`, can we combine them into one?

By the way, please fix clippy errors. See Jenkins for details.

src/coprocessor/statistics/analyze.rs
time-and-fate: Yes, now we always sample on the original values, calculate the total size using the original values, and build FM Sketch using the collate key (if it has).

skyzh: Looks good. Please re-request my review after lint passes.

skyzh (approved):

lzmhhh123 (approved):

Issue #10341, opened 2021-06-10
raftstore: stale peer may cause the store-wide safe-ts can't be updated


NingLin-P:
## Bug Report

Stale peers are peers that already removed from the peer list but the removed peer itself doesn't know that. When stale peer election timeout (take 10s~20s) it will send `RequestVote` message to other peers, it will receive a GC message and then destroy itself. But with hibernate region enabled the stale peer may take minutes to trigger election timeout.

Usually, stale peer has no harms but occupy a few resources, but with the stale read feature, the stale peer's `safe-ts` won't be updated by the leader, and may cause the store-wide `safe-ts` (which is used by the time-bound staleness transaction) can't update

NingLin-P:
Closing this PR since https://github.com/tikv/tikv/pull/10352 fixed the most common case

NingLin-P:
Closing this PR since #10352 fixed the most common case

PR #10352, opened 2021-06-11
raftstore: reply to stale peer with tombstone message everytime

gengliqi:

Problem Summary:
If a leader receives the message from a stale peer(follower or learner), it may not reply to this peer a tombstone msg which may cause a stale region can not be destroyed for a long time. 
Also, https://github.com/tikv/tikv/issues/10341 the problem described here can be alleviated.

What's Changed:
Since the number of msg from stale peers should be small, this PR totally removes the condition of whether to send tombstone msg or not. 

### Related changes

### Check List 

Tests 

- Integration test

Side effects

- No

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1481,11 +1481,8 @@ where
         if util::is_epoch_stale(from_epoch, self.fsm.peer.region().get_region_epoch())
             && util::find_peer(self.fsm.peer.region(), from_store_id).is_none()
         {
-            self.ctx.handle_stale_msg(
-                msg,
-                self.fsm.peer.region().get_region_epoch().clone(),
-                None,
-            );
+            self.ctx
+                .handle_stale_msg(msg, self.fsm.peer.region().get_region_epoch().clone(), None);
             return true;
         }
 
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1481,20 +1481,9 @@ where
         if util::is_epoch_stale(from_epoch, self.fsm.peer.region().get_region_epoch())
             && util::find_peer(self.fsm.peer.region(), from_store_id).is_none()
         {
-            let mut need_gc_msg = util::is_vote_msg(msg.get_message());
-            if msg.has_extra_msg() {
-                // A learner can't vote so it sends the check-stale-peer msg to others to find out whether
-                // it is removed due to conf change or merge.
-                need_gc_msg |=
-                    msg.get_extra_msg().get_type() == ExtraMessageType::MsgCheckStalePeer;
-                // For backward compatibility
-                need_gc_msg |= msg.get_extra_msg().get_type() == ExtraMessageType::MsgRegionWakeUp;
-            }
-            // The message is stale and not in current region.
             self.ctx.handle_stale_msg(
                 msg,
                 self.fsm.peer.region().get_region_epoch().clone(),
-                need_gc_msg,
                 None,
             );
             return true;
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -572,6 +572,16 @@ where
                 }
                 PeerMsg::Noop => {}
                 PeerMsg::UpdateReplicationMode => self.on_update_replication_mode(),
+                PeerMsg::Destroy(peer_id) => {
+                    if self.fsm.peer.peer_id() == peer_id {
+                        match self.fsm.peer.maybe_destroy(&self.ctx) {
+                            None => self.ctx.raft_metrics.message_dropped.applying_snap += 1,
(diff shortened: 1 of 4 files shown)


solotzg:
lgtm

NingLin-P (approved):

nolouch (approved):
lgtm

BusyJay (changes requested):
I don't think it's a good idea. If a peer is not isolated, it should remove it self in the end by applying logs.

NingLin-P:
> I don't think it's a good idea. If a peer is not isolated, it should remove it self in the end by applying logs.

Which could take minutes, and even longer for the learner. And with this PR, the peer only destroys itself by gc msg when it can't destroy itself by applying logs.

BusyJay:
> ...when it can't destroy itself by applying logs.

How? The PR only checks epoch, so even `MsgAppendResponse` can also trigger self-destruction.

NingLin-P:
> > ...when it can't destroy itself by applying logs.
> 
> How? The PR only checks epoch, so even `MsgAppendResponse` can also trigger self-destruction.

The leader won't send any msg to a removed peer. So in this case, the peer can only destroy itself if it has the `confchange` log and knows the log is committed. And when it receives a gc msg, it will send msg to destroy the apply fsm first, so the committed `confchange` log (if any) will be applied before destroy.

BusyJay:
> The leader won't send any msg to a removed peer. So in this case, the peer can only destroy itself if it has the `confchange` log and knows the log is committed. And when it receives a gc msg, it will send msg to destroy the apply fsm first, so the committed `confchange` log (if any) will be applied before destroy.

For example, leader sends two messages to follower, the first one is MsgAppend with latest logs and commit index, the second one is gc message and follower handles two messages during a ready. So it will not wait applying entries, instead remove it self directly.

NingLin-P:
> For example, leader sends two messages to follower, the first one is MsgAppend with latest logs and commit index, the second one is gc message and follower handles two messages during a ready. So it will not wait applying entries, instead remove it self directly.

It seems to be a rare case that the raft msg (carry the commitment of the `confchange`) and the gc msg are handled in the same `Ready`, and I think it is totally valid to destroy a peer by gc msg even the peer is not isolated, since we already allow to doing so.

BusyJay:
> ...since we already allow to doing so.

This only happens after some time, and follower is unlikely to catch up logs.

I suggest to only handle non-response message instead.

NingLin-P:
> This only happens after some time, and follower is unlikely to catch up logs.

This could happen as long as the leader applies the `confchange` log before the follower's `MsgAppendResponse` of the `confchange`arrived. And in cases I encountered there is no network isolation.

gengliqi:
> > For example, leader sends two messages to follower, the first one is MsgAppend with latest logs and commit index, the second one is gc message and follower handles two messages during a ready. So it will not wait applying entries, instead remove it self directly.
> 
> It seems to be a rare case that the raft msg (carry the commitment of the `confchange`) and the gc msg are handled in the same `Ready`, and I think it is totally valid to destroy a peer by gc msg even the peer is not isolated (and has the `confchange` log), since we already allow to doing so.

I agree it's a rare case. If leader has applied the conf change log, it won't send any append msg to this removed peer. Only after that, leader will send a tombstone msg to this removed peer when receiving any msg from this peer. In other words, the peer can not catch up when receiving tombstone msg. 
If we want to try our best to make peer removed after applying conf change, it's better to make the peer apply all the committed raft log before removing rather than not sending tombstone msg in some cases on leader side.

BusyJay:
> ...it's better to make the peer apply all the committed raft log before removing...

This looks good to me too.

NingLin-P:
> > ...it's better to make the peer apply all the committed raft log before removing...
> 
> This looks good to me too.

Besides the abovementioned case (assume msg arrive in order), the peer will apply all known committed logs before destroy. And for that case, while handling gc msg, I think we can get a `Ready` and send the committed log to apply worker first and then send another msg to destroy apply fsm.

NingLin-P:
Or we can handle the gc msg on the next round, so all know committed logs will be handled and apply before the gc msg remove destroy the `ApplyFsm` and `PeerFsm`. We can do it by keeping the gc msg in `poll_ctx` and resend the msg at `PollHandler::end` (also add some flag to distinguish the gc msg that can handle without delay).

gengliqi:
> Or we can handle the gc msg on the next round, so all known committed logs will be handled and applied before the gc msg destroy the `ApplyFsm` and `PeerFsm`. We can do it by keeping the gc msg in `poll_ctx` and resend the msg at `PollHandler::end` (also add some flag to distinguish the gc msg that can be handled without delay).
I make a small change that is similar to this method.
Add a new `PeerMsg::Destroy` and send it after handling `tombstone msg`. Thus we can ensure the committed entries are applied as many as possible.

BusyJay (approved):

NingLin-P (approved):

PR #10354, opened 2021-06-11
resolved_ts: fix high tail latency cause by time-consuming operation in raftstore 

### LINKED ISSUE #10342: raftstore: with large number of region the CheckLeader rpc may increase the tail latency of propose wait/append log
## Bug Report

The `CheckLeader` rpc also sync the `(apply index, safe ts)` from leader to follower, it is done by `RegionReadProgress:: update_safe_ts` which invoke lock acquiring, although the lock is usually not contended, but when the number of regions is large, it may acquiring lot of lock and being time-consuming. As result, the tail latency of propose wait and append log duration may increase:

image

image

NingLin-P:

Issue Number: close #10342 

Problem Summary:

Currently, the `CheckLeader` rpc is handled in `raftstore`, when the number of regions info in the `CheckLeader` request is large, handling the request could be time-consuming, and may increase the tail latency of `Append log duration`/`Propose wait duration`.

What's Changed:

This PR fix the issue by:
- move the handling of `CheckLeader` request to the `resolved-ts` worker
- instead of `StoreMeta`, using another fine-grained mutex for `region_read_progress`

### Related changes

### Check List 

- Existing integration test

### Release note 

```release-note
None
```


BusyJay:
How does it perform comparing to no check-leaders at all?

NingLin-P:
> How does it perform comparing to no check-leaders at all?

image

截屏2021-06-17 下午11 41 22

The `CheckLeader` is disabled by dynamically changing the `resolved-ts.advance-ts-interval` to 1h at about 23:15

components/raftstore/src/store/util.rs
```diff
@@ -905,6 +994,60 @@ impl RegionReadProgress {
...
+                    if !core.pause {
+                        self.safe_ts.store(ts, AtomicOrdering::Release);
+                    }
+                }
+            }
+        }
+        // whether the provided `LeaderInfo` is same as ours
+        core.leader_info.leader_term == leader_info.term
+            && core.leader_info.leader_id == leader_info.peer_id
+            && region_epoch_equal(&core.leader_info.epoch, leader_info.get_region_epoch())
```
Yisaer, line 1017: Why not check leaderInfo first?

Yisaer (commented):
Rest LGTM

components/raftstore/src/store/util.rs
NingLin-P: So that we can return the result directly instead of assigning the result to a temporary variable and return it at the end.

Yisaer (approved):

components/raftstore/src/store/worker/check_leader.rs
```diff
@@ -0,0 +1,164 @@
+// Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: Should be 2021.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -614,9 +614,23 @@ impl<'a, EK: KvEngine + 'static, ER: RaftEngine + 'static, T: Transport>
                     self.on_store_unreachable(store_id);
                 }
                 StoreMsg::Start { store } => self.start(store),
-                StoreMsg::CheckLeader { leaders, cb } => self.on_check_leader(leaders, cb),
+                StoreMsg::CheckLeader { leaders, cb } => {
```
BusyJay, line 617: Why have to go through raftstore?

NingLin-P: I have tried to redirect the request at the server layer through the `resolved-ts` worker scheduler but failed due to cycle dependency of `resolved-ts` and `tikv` crate.

BusyJay: I don't get it. Seems what's needed is just a channel sender, why need to include all the dependency?

NingLin-P: It needs the worker's `Task`, but since we move the `CheckLeader` handler to another worker, I think we can use that worker's scheduler to redirect requests now.

NingLin-P:
I have done more benchmark about this PR, the result looks good:

image

The first three workloads are running on the master branch with the `CheckLeader` interval set to `1s`, `5s` and `1h` (which can consider as disable `CheckLeader`) respectively, we can see the `.999` and `.99` latency is highly affected by the CheckLeader interval and with this PR the effect is not more exist.

components/raftstore/src/store/worker/check_leader.rs
```diff
@@ -0,0 +1,166 @@
...
+                .filter(|ts| *ts != 0) // ts == 0 means the peer is uninitialized
+                .min()
+                .unwrap_or(0)
+            })
+        } else {
+            // `store_safe_ts` won't be accessed frequently (like per-request or per-transaction),
+            // also this branch won't entry because the request key range is empty currently (in v5.1)
+            // keep this branch for robustness and future use, so it is okay getting `store_safe_ts`
+            // from `store_meta` (behide a mutex)
+            let meta = self.store_meta.lock().unwrap();
```
BusyJay, line 73: This can be moved to L77.

components/raftstore/src/store/worker/check_leader.rs
```diff
@@ -0,0 +1,166 @@
...
+            let (start_key, end_key) = (
+                data_key(key_range.get_start_key()),
+                data_end_key(key_range.get_end_key()),
+            );
+            meta.region_read_progress.map(|registry| {
+                meta.region_ranges
+                // get overlapped regions
+                .range((Excluded(start_key.clone()), Unbounded))
```
BusyJay, line 81: Why clone?

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +844,90 @@ impl Display for MsgType<'_> {
...
+            .unwrap()
+            .get(region_id)
+            .map(|rp| rp.safe_ts())
+    }
+
+    // Update `safe_ts` with the provided `LeaderInfo` and return the regions that have the
+    // same `LeaderInfo`
+    pub fn handle_check_leaders(&self, leaders: Vec<LeaderInfo>) -> Vec<u64> {
+        let mut regions = Vec::with_capacity(leaders.len());
+        let registry = self.registry.lock().unwrap();
```
BusyJay, line 890: Can be dropped at L898.

components/raftstore/src/store/util.rs
BusyJay: Then why it's checked first in the original implementation?

components/raftstore/src/store/peer.rs
```diff
@@ -982,6 +983,10 @@ where
         // becoming a leader.
         self.maybe_update_read_progress(reader, progress);
 
+        // Update leader info
+        self.read_progress
```
BusyJay, line 987: I find it's confusing. There are `maybe_update_read_progress` and `read_progress`, and they refer to different things.

components/raftstore/src/store/util.rs
NingLin-P: Just for robustness I think, the check is unnecessary since even it passes in this peer doesn't mean it can pass among majority peers. The `read_state` item should be promised valid when it is published by the leader instead of letting the receiver validated it with the receiver's limited information.

components/raftstore/src/store/peer.rs
NingLin-P: How about rename the `maybe_update_read_progress` to `maybe_update_leader_read_progress` or `maybe_update_lease_read_progress`?

components/raftstore/src/store/worker/check_leader.rs
NingLin-P: Because `map` takes `FnMut`, I change it to `FnOnce` and remove the `clone` now.

components/raftstore/src/store/util.rs
BusyJay: /cc @5kbpers

components/raftstore/src/store/peer.rs
BusyJay: How about renaming `RegionReadProgress` to `SafeReadTracker`?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -126,8 +124,8 @@ pub struct StoreMeta {
     /// source_region_id -> need_atomic
     /// Used for reminding the source peer to switch to ready in `atomic_snap_regions`.
     pub destroyed_region_for_snap: HashMap<u64, bool>,
-
-    pub region_read_progress: HashMap<u64, Arc<RegionReadProgress>>,
+    /// region_id -> `RegionReadProgress`
+    pub region_read_progress: RegionReadProgressRegistry,
```
5kbpers, line 128: Should we remove `leaders` of `StoreMeta`?

components/resolved_ts/src/advance.rs
```diff
@@ -136,68 +143,54 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
-                                .push(leader_info);
-                        }
-                        region_map.insert(region_id, region.clone());
+        // region_id -> `(Vec<Peer>, LeaderInfo)`
+        let info_map = region_read_progress.dump_leader_infos(&regions);
+
+        for (region_id, (peer_list, leader_info)) in info_map {
+            let leader_id = leader_info.get_peer_id();
+            // Check if the leader in this store
+            match find_store_id(&peer_list, leader_id) {
```
BusyJay, line 163: You mean `find_store_id(&peer_list, leader_id) != Some(store_id)`?

components/resolved_ts/src/advance.rs
```diff
@@ -136,68 +143,54 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
+                        continue;
                     }
                 }
             }
+            for peer in &peer_list {
+                if peer.store_id == store_id && peer.id == leader_id {
+                    resp_map.entry(region_id).or_default().push(store_id);
+                    continue;
+                }
+                store_map
```
BusyJay, line 176: Why learner doesn't need to be filtered anymore?

components/resolved_ts/src/advance.rs
NingLin-P: Yes

components/resolved_ts/src/advance.rs
NingLin-P: The `CheckLeader` handling is port from `cdc`, and it is original for just checking leadership so filter learner to reduce request size, but now with `resolved-ts` it is also used to sync `read_state` from leader to follower/learner, so we don't filter learner anymore.

components/raftstore/src/store/peer.rs
NingLin-P: Renaming `RegionReadProgress` would be a huge change, if just to resolve the naming conflict I think it is better to change the less used one.

components/resolved_ts/src/advance.rs
BusyJay: Seems a bug, any test case?

components/resolved_ts/src/advance.rs
```diff
@@ -127,6 +133,7 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
     async fn region_resolved_ts_store(
```
5kbpers, line 133: Could you update the related code in cdc or add a TODO comment there?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -126,8 +124,8 @@ pub struct StoreMeta {
     /// source_region_id -> need_atomic
     /// Used for reminding the source peer to switch to ready in `atomic_snap_regions`.
     pub destroyed_region_for_snap: HashMap<u64, bool>,
-
-    pub region_read_progress: HashMap<u64, Arc<RegionReadProgress>>,
+    /// region_id -> `RegionReadProgress`
```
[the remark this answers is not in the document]
NingLin-P: It is kept for `cdc` usage, we can remove it after `cdc` adopt the same change as `resolved-ts`

components/raftstore/src/store/util.rs
```diff
@@ -836,6 +844,92 @@ impl Display for MsgType<'_> {
...
+        for leader_info in leaders {
+            let region_id = leader_info.get_region_id();
+            if let Some(rp) = registry.get(&region_id) {
+                if rp.consume_leader_info(leader_info) {
```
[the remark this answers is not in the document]
NingLin-P: `consume_leader_info` not just for checking leader info but also updating `safe_ts`, it do two jobs at once to reduce one `Mutex::lock` calling, `is_leader_info_valid` may not best fit.

components/resolved_ts/src/advance.rs
NingLin-P: Done

components/resolved_ts/src/advance.rs
NingLin-P: added, PTAL

components/raftstore/src/store/peer.rs
NingLin-P: Leave a TODO for now, will change it later.

components/raftstore/src/store/peer.rs
```diff
@@ -1498,6 +1503,10 @@ where
             "term" => term,
             "peer_id" => self.peer_id(),
         );
+
+        self.read_progress
+            .update_leader_info(leader_id, term, self.region());
```
BusyJay, line 1508: `region` may not be correct here. If it's an uninitialized peer, region's peer list can be empty.

NingLin-P: Yes, but we only use the leader's peer list. The `leader_info` will also be updated after snapshot applied and epoch changed, so the peer's `leader_info` will be corrected as it catches up with more data.

BusyJay: I don't think so. `update_leader_info` is only called when leadership changes, so even for leader, the peer list can become wrong after applying configuration change.

NingLin-P: Actaully,  `update_leader_info` is called inside `Peer::set_config`, `Peer::on_leader_changed` and `PeerFsmDelegate::on_ready_apply_snapshot`, it has cover all cases where the peer list, region epoch and leader id will change.

BusyJay (approved):

5kbpers (approved):

PR #10364, opened 2021-06-15
raftstore: skip clearing callback when shutdown

### LINKED ISSUE #10307: Suspicious warnings during shutdown
v4.0.11

Doesn't matter.

### Steps to reproduce
Shutdown the TiKV

### What did you expect?
Everything work fine

### What did happened?
```
[2021/05/26 22:55:43.967 +08:00] [WARN] [mod.rs:396] ["async snapshot callback is dropped"] [backtrace="stack backtrace:\n   0: tikv::storage::kv::drop_snapsh
ot_callback\n             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/src/storage/kv/mod.rs:395\n      core::ops::function::FnOnce::
call_once\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ops/function.rs:232\n   1: <tikv_util::callback::MustCall<T,C,A> as cor
e::ops::drop::Drop>::drop\n             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/components/tikv_util/src/callback.rs:48\n      c
ore::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n      core::ptr::real_drop_in_place\
n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n   2: core::ptr::real_drop_in_place\n             at /rustc/0de96
d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n      core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817
669bc8/src/libcore/ptr/mod.rs:182\n   3: core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs
:182\n      core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n   4: core::ptr::real_d
rop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n      core::ptr::real_drop_in_place\n             at
 /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n   5: core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c
18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n      core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libco
re/ptr/mod.rs:182\n      core::ptr::drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:172\n      alloc::ve
c::Vec<T>::truncate\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/liballoc/vec.rs:746\n      alloc::vec::Vec<T>::clear\n             at
 /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/liballoc/vec.rs:1318\n      raftstore::store::read_queue::ReadIndexQueue::clear_all\n             at comp
onents/raftstore/src/store/read_queue.rs:120\n   6: <raftstore::store::fsm::peer::PeerFsm<E> as core::ops::drop::Drop>::drop\n             at components/rafts
tore/src/store/peer.rs:2779\n      core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n
   7: core::ptr::real_drop_in_place\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ptr/mod.rs:182\n      batch_system::fsm::FsmS
tate<N>::release\n             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/components/batch-system/src/fsm.rs:114\n      batch_syste
m::mailbox::BasicMailbox<Owner>::release\n             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/components/batch-system/src/mailb
ox.rs:39\n      batch_system::batch::Batch<N,C>::release\n             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/components/batch-
system/src/batch.rs:124\n      batch_system::batch::Poller<N,C,Handler>::poll\n             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/t
ikv/components/batch-system/src/batch.rs:340\n   8: batch_system::batch::BatchSystem<N,C>::spawn::{{closure}}\n             at /home/jenkins/agent/workspace/b
uild_tikv_multi_branch_v4.0.11/tikv/components/batch-system/src/batch.rs:398\n      std::sys_common::backtrace::__rust_begin_short_backtrace\n             at 
/rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/sys_common/backtrace.rs:136\n   9: std::thread::Builder::spawn_unchecked::{{closure}}::{{closure}}\
n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/thread/mod.rs:469\n      <std::panic::AssertUnwindSafe<F> as core::ops::function::
FnOnce<()>>::call_once\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/panic.rs:318\n      std::panicking::try::do_call\n         
    at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/panicking.rs:292\n      std::panicking::try\n             at /rustc/0de96d37fbcc54978458c18f
5067cd9817669bc8//src/libpanic_unwind/lib.rs:78\n      std::panic::catch_unwind\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/pa
nic.rs:394\n      std::thread::Builder::spawn_unchecked::{{closure}}\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libstd/thread/mod.rs
:468\n      core::ops::function::FnOnce::call_once{{vtable.shim}}\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/libcore/ops/function.rs
:232\n  10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/liballoc/
boxed.rs:1022\n  11: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once\n             at /rustc/0de96d37fbcc54978458c18f5067cd9817669bc8/src/
liballoc/boxed.rs:1022\n      std::sys_common::thread::start_thread\n             at src/libstd/sys_common/thread.rs:13\n      std::sys::unix::thread::Thread:
:new::thread_start\n             at src/libstd/sys/unix/thread.rs:80\n  12: start_thread\n  13: clone\n"]
```

### LINKED ISSUE #10353: response from raftstore is not deterministic
v4.0.11, but every version after v3.0.0 is affected.

Doesn't matter.

### Steps to reproduce
Produce highly conflict requests and slow down the shutdown process

### What did you expect?
It shutdown successfully, and ACID is still hold.

### What did happened?
Transaction is corrupted by either missing data from default CF or some part of transaction are rollback.

When a node is shutdown without evicting leader, then leader can't know whether a log is committed or not, so to clear callback, it will respond with a stale response to hint the client to retry. StaleCommand error doesn't mean the request will fail eventually or not, it just mean the leader can't handle the request at the moment.

But latch assumes all responses from raftstore are deterministic that if an error is returned, then the write should never be written unless retry. Hence during shutdown, if there are more than two requests in latch queue, then the first two requests will be processed one by one. If the first request is to commit a transaction and the second is to rollback, then the key can be both committed and rollback. If data is large enough to be stored in default cf, error like `DefaultNotFound` will be reported; in all conditions, ACID is broken.

`StaleCommand` can also be called when a peer is about to be destroyed. But we don't allow to remove leader eagerly, so when the error is reported, the leader should either be removed by leader at last term and it won't process any read before being destroy. If the leader is paused, and then being destroyed by a new leader, all following commands won't be committed as it's not leader anymore.

So to fix the problem, we can either make raftstore's response deterministic or make latch to handle undeterministic error.

For the former, an easy fix is to stop invoking any callbacks during shutdown, this can have the side effects raw kv may not know the requests are stopped, instead, it may timeout before next try.

For the latter, latch should not release slot when an undeterministic error is returned. I think it is also necessary after implementing timeout inside raftstore.

There is a similar discussion in #9113, which is discussing deterministic of transaction.

BusyJay:

Issue Number:
close #10353 
close #10307 

Problem Summary:

If callbacks are cleared during shutdown, latch may not work as expected
as it expects deterministic result. So this PR drops all callbacks
instead of invoking.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
skip clearing callback during gracefully shutdown to avoid breaking ACID in some cases
```

### DIFF
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -3577,7 +3577,7 @@ where
     EK: KvEngine,
 {
     fn drop(&mut self) {
-        if tikv_util::thread_group::is_shutdown().unwrap_or(false) {
+        if tikv_util::thread_group::is_shutdown(!cfg!(test)) {
             self.delegate.clear_all_commands_silently()
         } else {
             self.delegate.clear_all_commands_as_stale();
@@ -3836,7 +3836,7 @@ where
                         "target region is not found, drop proposals";
                         "region_id" => region_id
                     );
-                    if !tikv_util::thread_group::is_shutdown().unwrap_or(false) {
+                    if !tikv_util::thread_group::is_shutdown(!cfg!(test)) {
                         for p in apply.cbs.drain(..) {
                             let cmd = PendingCmd::<EK::Snapshot>::new(p.index, p.term, p.cb);
                             notify_region_removed(apply.region_id, apply.peer_id, cmd);
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -3836,6 +3836,17 @@ where
                         "target region is not found, drop proposals";
                         "region_id" => region_id
                     );
+                    // Invoking callback can release txn latch, if it's still leader, following
+                    // command may not read the writes of previous commands and break ACID. If
+                    // it's still leader, there are two possibility that mailbox is closed:
+                    // 1. The process is shutting down.
+                    // 2. The leader is destroyed. A leader won't propose to destroy itself, so
+                    //     it should either destroyed by older leaders or newer leaders. Leader
+                    //     won't respond to read until it has applied to current term, so no
+                    //     command will be proposed until command from older leaders have applied,
+                    //     which will then stop it from accepting proposals. If the command is
+                    //     proposed by new leader, then it won't be able to propose new proposals.
+                    // So only shutdown needs to be checked here.
                     if !tikv_util::thread_group::is_shutdown(!cfg!(test)) {
                         for p in apply.cbs.drain(..) {
                             let cmd = PendingCmd::<EK::Snapshot>::new(p.index, p.term, p.cb);
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -1341,6 +1341,15 @@ where
         }
     }
 
+    fn clear_all_commands_silently(&mut self) {
+        for mut cmd in self.pending_cmds.normals.drain(..) {
+            cmd.cb.take();
+        }
+        if let Some(mut cmd) = self.pending_cmds.conf_change.take() {
+            cmd.cb.take();
+        }
+    }
+
     fn new_ctx(&self, index: u64, term: u64) -> ExecContext {
         ExecContext::new(self.apply_state.clone(), index, term)
     }
@@ -3587,7 +3596,11 @@ where
     EK: KvEngine,
 {
     fn drop(&mut self) {
-        self.delegate.clear_all_commands_as_stale();
+        if tikv_util::thread_group::is_shutdown().unwrap_or(false) {
+            self.delegate.clear_all_commands_silently()
+        } else {
+            self.delegate.clear_all_commands_as_stale();
+        }
     }
 }
 
@@ -3840,9 +3853,11 @@ where
                         "target region is not found, drop proposals";
                         "region_id" => region_id
                     );
-                    for p in apply.cbs.drain(..) {
-                        let cmd = PendingCmd::<EK::Snapshot>::new(p.index, p.term, p.cb);
-                        notify_region_removed(apply.region_id, apply.peer_id, cmd);
+                    if !tikv_util::thread_group::is_shutdown().unwrap_or(false) {
+                        for p in apply.cbs.drain(..) {
+                            let cmd = PendingCmd::<EK::Snapshot>::new(p.index, p.term, p.cb);
+                            notify_region_removed(apply.region_id, apply.peer_id, cmd);
+                        }
                     }
                     return;
                 }
--- a/components/tikv_util/src/thread_group.rs
+++ b/components/tikv_util/src/thread_group.rs
@@ -1,0 +1,55 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+//! Provides util functions to manage share properties across threads.
+
+use std::cell::RefCell;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
+#[derive(Default)]
+struct GroupPropertiesInner {
+    pub shutdown: AtomicBool,
+}
+
+#[derive(Default, Clone)]
+pub struct GroupProperties {
+    inner: Arc<GroupPropertiesInner>,
+}
+
+impl GroupProperties {
+    #[inline]
+    pub fn mark_shutdown(&self) {
+        self.inner.shutdown.store(true, Ordering::SeqCst);
+    }
+}
+
+thread_local! {
+    static PROPERTIES: RefCell<Option<GroupProperties>> = RefCell::new(None);
+}
+
+pub fn current_properties() -> Option<GroupProperties> {
+    PROPERTIES.with(|p| p.borrow().clone())
+}
+
+pub fn set_properties(props: Option<GroupProperties>) {
+    PROPERTIES.with(move |p| {
+        p.replace(props);
+    })
+}
+
+/// Checks if the system is shutdown.
+pub fn is_shutdown() -> Option<bool> {
+    PROPERTIES.with(|p| {
+        p.borrow()
+            .as_ref()
+            .map(|props| props.inner.shutdown.load(Ordering::SeqCst))
+    })
+}
+
+pub fn mark_shutdown() {
+    PROPERTIES.with(|p| {
+        if let Some(props) = &mut *p.borrow_mut() {
+            props.mark_shutdown();
+        }
+    })
+}
--- a/components/tikv_util/src/thread_group.rs
+++ b/components/tikv_util/src/thread_group.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 
 #[derive(Default)]
 struct GroupPropertiesInner {
-    pub shutdown: AtomicBool,
+    shutdown: AtomicBool,
 }
 
 #[derive(Default, Clone)]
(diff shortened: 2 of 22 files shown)


sticnarf (commented):
Adding properties to threads looks error-prone. The `unwrap_or(false)` in reading shutdown property also don't panic if the property is not set.
I'm a bit afraid that in the future some threads can forget to add the properties... Can we change `unwrap_or(false)` to `unwrap`? (if no unit test depends it....)

BusyJay:
Now property is always checked, including integration tests, unless running unit tests.

BusyJay:
Failed case is related to #10372

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3568,7 +3577,11 @@ where
     EK: KvEngine,
 {
     fn drop(&mut self) {
-        self.delegate.clear_all_commands_as_stale();
+        if tikv_util::thread_group::is_shutdown(!cfg!(test)) {
+            self.delegate.clear_all_commands_silently()
```
Connor1996, line 3581: Maybe add some comments here?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3823,9 +3836,11 @@ where
...
                         "region_id" => region_id
                     );
-                    for p in apply.cbs.drain(..) {
-                        let cmd = PendingCmd::<EK::Snapshot>::new(p.index, p.term, p.cb);
-                        notify_region_removed(apply.region_id, apply.peer_id, cmd);
+                    if !tikv_util::thread_group::is_shutdown(!cfg!(test)) {
+                        for p in apply.cbs.drain(..) {
+                            let cmd = PendingCmd::<EK::Snapshot>::new(p.index, p.term, p.cb);
+                            notify_region_removed(apply.region_id, apply.peer_id, cmd);
+                        }
```
NingLin-P, line 3843: If a leader is removed by applying `confchange` (perhaps leader changed after the `confchange` proposed), invoking the callback while the raft log is not applied and `PeerFsm` still working may cause the same issue.

components/tikv_util/src/thread_group.rs
```diff
@@ -0,0 +1,59 @@
...
+impl GroupProperties {
+    #[inline]
+    pub fn mark_shutdown(&self) {
+        self.inner.shutdown.store(true, Ordering::SeqCst);
+    }
+}
+
+thread_local! {
+    static PROPERTIES: RefCell<Option<GroupProperties>> = RefCell::new(None);
+}
```
NingLin-P, line 28: Could we use a global variable instead? So we don't need to pass it around threads.

components/raftstore/src/store/fsm/apply.rs
NingLin-P, line 3581: Looks like a no-op, the callback should be droped while `ApplyFsm` is dropped.

components/raftstore/src/store/fsm/apply.rs
BusyJay: First of all, a leader will not commit entries to delete itself, so the entry has to come from last term. And a leader will not propose new writes until it has applied to current term, so same issue will not happen.

components/tikv_util/src/thread_group.rs
BusyJay: The problem of global variable is that it affects all threads, which can pollute tests as it will start multiple instance using threads.

components/raftstore/src/store/fsm/apply.rs
NingLin-P: Oh, the `apply_res` will stop the `PeerFsm` first before the incoming write.

components/raftstore/src/store/fsm/apply.rs
BusyJay: We have assert to ensure every pending command is called properly to avoid latch leak.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3836,6 +3836,17 @@ where
                         "target region is not found, drop proposals";
                         "region_id" => region_id
                     );
+                    // Invoking callback can release txn latch, if it's still leader, following
+                    // command may not read the writes of previous commands and break ACID. If
+                    // it's still leader, there are two possibility that mailbox is closed:
+                    // 1. The process is shutting down.
+                    // 2. The leader is destroyed. A leader won't propose to destroy itself, so
```
NingLin-P, line 3843: Looks like the `raftstore.allow-remove-leader` config is not remove yet

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3836,6 +3836,17 @@ where
...
+                    //     it should either destroyed by older leaders or newer leaders. Leader
+                    //     won't respond to read until it has applied to current term, so no
+                    //     command will be proposed until command from older leaders have applied,
```
NingLin-P, line 3846: Is there any command from txnkv will propose write directly without reading?

BusyJay: /cc @sticnarf 

The statement is true for all versions. And for latest version, leader also won't propose any command before applying to current term to build correct pipeline.

components/raftstore/src/store/fsm/apply.rs
BusyJay: I will remove the config in another PR.

BusyJay: See #10384.

sticnarf (approved):
LGTM

NingLin-P (approved):

youjiali1995 (approved):

PR #10384, opened 2021-06-17
raftstore: always reject remove leader

BusyJay:

What's Changed:

Correctness relies on the assumption that leader won't propose to remove
itself. We used to have a hidden configuration to control the behavior,
this PR make the rejection permanent.

See also #10353.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

    `raftstore.allow-remove-leader` is removed and ignored.

### Release note 

```release-note
Always reject proposing to remove leader
```


components/raftstore/src/store/config.rs
```diff
@@ -138,8 +138,6 @@ pub struct Config {
     #[config(hidden)]
     pub right_derive_when_split: bool,
 
-    pub allow_remove_leader: bool,
```
NingLin-P, line 141: Should also remove `allow-remove-leader`

NingLin-P (commented):
Rest LGTM

components/raftstore/src/store/config.rs
BusyJay: Do you mean also search for `allow-remove-leader`? I did, and it's a hidden configuration, so nothing shows up.

NingLin-P: It is in some test config file like `components/test_raftstore/src/common-test.toml`

hicqu (approved):

gengliqi (approved):

BusyJay:
@gengliqi @hicqu  PTAL, the new change still keeps the configuration for tests and remove for production.

hicqu (approved):

gengliqi (approved):

PR #10404, opened 2021-06-18
raftstore: Refactor SplitObserver.adjust_key()

andylokandy:

The `SplitObserver.adjust_key()` is unclear. So renamed the function and move the checks of keys to the `on_split()`.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/coprocessor/split_observer.rs
+++ b/components/raftstore/src/coprocessor/split_observer.rs
@@ -49,7 +49,7 @@ fn is_valid_split_key(key: &[u8], index: usize, region: &Region) -> bool {
 }
 
 /// `SplitObserver` adjusts the split key so that it won't separate
-/// multiple MVCC versions of a key into two region.
+/// multiple MVCC versions of a key into two regions.
 #[derive(Clone)]
 pub struct SplitObserver;
 
@@ -81,7 +81,7 @@ impl SplitObserver {
         if ajusted_splits.is_empty() {
             Err("no valid key found for split.".to_owned())
         } else {
-            // Rewrite the splites.
+            // Rewrite the splits.
             std::mem::swap(splits, &mut ajusted_splits);
             Ok(())
         }
--- a/components/raftstore/src/coprocessor/split_observer.rs
+++ b/components/raftstore/src/coprocessor/split_observer.rs
@@ -61,7 +61,7 @@ impl SplitObserver {
         ctx: &mut ObserverContext<'_>,
         splits: &mut Vec<SplitRequest>,
     ) -> Result<(), String> {
-        let mut ajusted_splits = std::mem::take(splits)
+        let ajusted_splits = std::mem::take(splits)
             .into_iter()
             .enumerate()
             .filter_map(|(i, mut split)| {
@@ -94,7 +94,7 @@ impl SplitObserver {
             Err("no valid key found for split.".to_owned())
         } else {
             // Rewrite the splits.
-            std::mem::swap(splits, &mut ajusted_splits);
+            *splits = ajusted_splits;
             Ok(())
         }
     }
--- a/components/raftstore/src/coprocessor/split_observer.rs
+++ b/components/raftstore/src/coprocessor/split_observer.rs
@@ -59,44 +59,26 @@ impl SplitObserver {
         ctx: &mut ObserverContext<'_>,
         splits: &mut Vec<SplitRequest>,
     ) -> Result<(), String> {
-        let (mut i, mut j) = (0, 0);
-        let mut last_valid_key: Option<Vec<u8>> = None;
-        let region = ctx.region();
-        while i < splits.len() {
-            let k = i;
-            i += 1;
-            {
-                let split = &mut splits[k];
+        let mut ajusted_splits = std::mem::replace(splits, Vec::new())
+            .into_iter()
+            .enumerate()
+            .filter_map(|(i, mut split)| {
                 let key = split.take_split_key();
                 let key = strip_timestamp_if_exists(key);
-
-                if !is_valid_split_key(&key, k, region) {
-                    continue;
+                if is_valid_split_key(&key, i, ctx.region()) {
+                    split.set_split_key(key);
+                    Some(split)
+                } else {
+                    None
                 }
+            })
+            .collect::<Vec<_>>();
 
-                if last_valid_key.as_ref().map_or(false, |k| *k >= key) {
-                    warn!(
-                        "skip invalid split key: key should not larger than previous";
-                        "region_id" => region.get_id(),
-                        "key" => log_wrappers::Value::key(&key),
-                        "previous" => log_wrappers::Value::key(last_valid_key.as_ref().unwrap()),
-                        "index" => k,
-                    );
-                    continue;
-                }
-                last_valid_key = Some(key.clone());
+        // Make sure that the split keys are sorted.
+        ajusted_splits.sort_unstable_by(|l, r| l.get_split_key().cmp(r.get_split_key()));
 
-                split.set_split_key(key)
-            }
-            if k != j {
-                splits.swap(k, j);
-            }
-            j += 1;
-        }
-        if j == 0 {
-            return Err("no valid key found for split.".to_owned());
-        }
-        splits.truncate(j);
+        // Rewrite the splitest
+        std::mem::swap(splits, &mut ajusted_splits);
         Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


BusyJay:
You may misunderstand `adjust_key`, test cases should show you what it does.

andylokandy:
@BusyJay I see, `encode_bytes(decode_bytes(key))` does not always equal to `key` because it may strip timestamp for the key. So I added the side effect back.

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -236,7 +235,7 @@ mod tests {
         assert!(resp.is_ok());
         assert!(!req.has_split(), "only split req should be handle.");
 
-        req = new_split_request(b"test");
```
andylokandy, line 239: The previous test case is invalid because the raw key is not in the region.

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,85 +1,83 @@
...
+                if key.is_empty() {
+                    warn!(
+                        "key is empty";
+                        "region_id" => region_id,
+                        "index" => k,
+                    );
+                    continue;
                 }
+
+                if let Err(Error::KeyNotInRegion(..)) =
```
BusyJay, line 54: Can you make it a minimum change by keeping `adjust_keys` and move theses checks into `adjust_keys`?

components/raftstore/src/coprocessor/split_observer.rs
BusyJay: Does the following line have relations to the key?

andylokandy: nope.

components/raftstore/src/coprocessor/split_observer.rs
andylokandy: I think the new name more precisely described what it does. Until 1 year ago, there are other adjustment logic, but then removed, so only strip timestamp now. And I think the check is more suitable in the `on_split()`.

BusyJay: It's harder to read than before. I don't think the name matters, it can be named as `adjust_keys` or `check_keys`, the point is put those validations into a function and let outer function to do error handling.

andylokandy: done.

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,85 +1,92 @@
...
+        Err(_) => 0,
+    };
+    key.truncate(key.len() - strip_len);
+    key
+}
+
+fn is_valid_split_key(key: &Vec<u8>, index: usize, region: &Region) -> bool {
+    if key.is_empty() {
+        warn!(
+            "key is empty";
```
Connor1996, line 29:
```suggestion
            "skip invalid split key: key is empty";
```

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,85 +1,92 @@
...
+            "region_id" => region.get_id(),
+            "index" => index,
+        );
+        return false;
+    }
+
+    if let Err(Error::KeyNotInRegion(..)) = util::check_key_in_region_exclusive(&key, region) {
+        warn!(
+            "key is not in region";
```
Connor1996, line 38:
```suggestion
            "skip invalid split key: key is not in region";
```

Connor1996 (commented):
rest LGTM

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,85 +1,92 @@
...
+    };
+    key.truncate(key.len() - strip_len);
+    key
+}
+
+fn is_valid_split_key(key: &Vec<u8>, index: usize, region: &Region) -> bool {
+    if key.is_empty() {
+        warn!(
+            "key is empty";
+            "region_id" => region.get_id(),
```
BusyJay, line 30: Why not return error and let outer function log context.

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,85 +1,92 @@
...
 use crate::store::util;
+use crate::Error;
 use kvproto::metapb::Region;
 use kvproto::raft_cmdpb::{AdminCmdType, AdminRequest, SplitRequest};
-use std::result::Result as StdResult;
+
+fn strip_timestamp_if_exists(mut key: Vec<u8>) -> Vec<u8> {
+    let mut slice = key.as_slice();
+    let strip_len = match bytes::decode_bytes(&mut slice, false) {
+        // It is a encoded key and the slice points to the remaining unparsable
```
BusyJay, line 15:
```suggestion
        // It is an encoded key and the slice points to the remaining unparsable
```

BusyJay (commented):
Rest LGTM

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,85 +1,92 @@
...
-                        );
-                        continue;
-                    }
+                let key = strip_timestamp_if_exists(key);
+
+                if !is_valid_split_key(&key, k, region) {
+                    continue;
                 }
+
+                if last_valid_key.as_ref().map_or(false, |k| *k >= key) {
```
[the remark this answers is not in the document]
andylokandy: it uses the variable `last_valid_key` in the loop, which  `is_valid_split_key` doesn't know.

Connor1996 (approved):
LGTM

components/raftstore/src/coprocessor/split_observer.rs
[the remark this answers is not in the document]
BusyJay: Isn't it just a matter of an additional parameter?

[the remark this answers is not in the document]
andylokandy: I don't think it's appropriate for a function to check the caller loop state...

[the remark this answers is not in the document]
andylokandy: I removed the check by sorting the vector in the end. Let's wait for the CI.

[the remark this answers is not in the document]
BusyJay: Isn't `key` itself is also a part of caller loop state? Sorting may not work as expected. For example, same keys in different positions of the array can pass the check and produce wrong proposal.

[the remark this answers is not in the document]
andylokandy: good catch. fixed.

> Isn't key itself is also a part of caller loop state?
The key is obviously the objective of `is_valid_sort_key()`, while `last_split_key` checks the relation between keys. Anyway, I abstracted away the `j` `k` `i` stuff and used iterator instead. They should be equivalent in performance.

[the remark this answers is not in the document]
BusyJay: > They should be equivalent in performance.

How can they be equivalent giving that original implementation only loop once and no allocation at all?

Now both duplicated keys and wrong order keys will be handled silently.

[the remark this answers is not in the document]
andylokandy: Before, every key will be cloned into last_split_key, so in the happy path, it'll be even faster.

Btw, is the duplicates or wrong-order matters. They are not skipped.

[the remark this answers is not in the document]
BusyJay: > every key will be cloned into last_split_key, so in the happy path, it'll be even slower.

I thought we were talking about the performance of loop and iterator.

> do the duplicates or wrong-order keys matter? 

Yes. In normal case, there should be no duplicated or out of order keys. It's a hint for potential bugs or wrong user inputs.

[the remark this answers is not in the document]
andylokandy: First, split keys are usually in small amount. Second, ajusting keys is insignificant comparing to other split process. Third, it loops two more times while reduces an allocation for every key, gain and pain. So I said it should be fairly equal (without messurable difference)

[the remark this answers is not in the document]
andylokandy: @BusyJay Your concern is fixed, and there is no clone for `the_last_key`.

components/raftstore/src/coprocessor/split_observer.rs
```diff
@@ -1,96 +1,102 @@
...
+                    Ok(prev)
+                }
+            })
+            .collect::<Vec<_>>();
+
+        if ajusted_splits.is_empty() {
+            Err("no valid key found for split.".to_owned())
+        } else {
+            // Rewrite the splits.
+            std::mem::swap(splits, &mut ajusted_splits);
```
BusyJay, line 97: You mean `*splits = adjusted_splits`?

BusyJay (approved):

andylokandy:
/run-all-tests

PR #10409, opened 2021-06-20
README.md: replace TiKV roadmap url

hongrubb:
The previous file is inaccessible. so I changed the url.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ You can view the list of [TiKV Adopters](https://tikv.org/adopters/).
 
 ## TiKV roadmap
 
-You can see the [TiKV Roadmap](docs/ROADMAP.md).
+You can see the [TiKV Roadmap](https://tikv.org/docs/4.0/roadmap/).
 
 ## TiKV software stack
 


skyzh (commented):
Roadmap for TiKV 4.0 is not applicable to current version. @tisonkun do we still have our roadmap tracked on GitHub? If not, I suggest removing this section or replace this URL with TiDB Internals.

tisonkun:
@skyzh agree. At least we should remove misleading information.

skyzh:
@hongrubb would you please help remove that section in this PR? By the way, it seems that you are using different email for git command line and GitHub, which makes your avatar missing from that commit.

hongrubb:
> @hongrubb would you please help remove that section in this PR? By the way, it seems that you are using different email for git command line and GitHub, which makes your avatar missing from that commit.

sure, I will try to fix this.

hongrubb:
> @hongrubb would you please help remove that section in this PR? By the way, it seems that you are using different email for git command line and GitHub, which makes your avatar missing from that commit.

Very thanks for pointing out the problem with the avatar. 

But unfortunately, I don't know why, I couldn't update the commit to this PR... 
I can get the commit log in my own repo, but they are not connected. 
As you can see, on this page, the link in the title `hongrubb:master` is grey.

skyzh:
I have no idea either. You may open a new PR, attach a link to this PR, and cc me and tisonkun for review.

PR #10414, opened 2021-06-21
Support dynamically configuring IO priority

tabokie:

Issue Number: close #xxx 

Problem Summary:

Off-line import/export and on-line import/export have different requirements.

What's Changed:

Support online configuration.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


yiwu-arbug (approved):
LGTM

hicqu (approved):

components/file_system/src/lib.rs
```diff
@@ -107,6 +126,7 @@ impl std::ops::Sub for IOBytes {
     }
 }
 
+#[repr(u32)]
```
hicqu, line 129: Seems u8 is enough. How about use u8 instead?

tabokie: I think we can't really gain from using u8, the vector is wrapped in `CachePadded`, and it's possible that u8 arithmetic is slower, considering most architectures are at least 32bit aligned.

Connor1996 (approved):

PR #10415, opened 2021-06-21
statistic: return user-visible scanned bytes

zhongzc:

Lack of some statistics for calculating SLI/SLO.

What's Changed:

Collect processed data size in `PointGetter` and `Scanner`.

### Related changes

- kvproto: https://github.com/pingcap/kvproto/pull/778

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/components/tikv_util/src/codec/bytes.rs
+++ b/components/tikv_util/src/codec/bytes.rs
@@ -278,20 +278,6 @@ pub fn decode_bytes_in_place(data: &mut Vec<u8>, desc: bool) -> Result<()> {
     }
 }
 
-// Returns the decoded key length of the given encoded key which is assumed valid without actually
-// decoding it.
-pub fn unchecked_decoded_len(encoded: &[u8], desc: bool) -> usize {
-    let group_cnt = encoded.len() / (ENC_GROUP_SIZE + 1);
-    let last_group_marker = encoded[(group_cnt * (ENC_GROUP_SIZE + 1)) - 1];
-    let last_group_size = ENC_GROUP_SIZE
-        - if desc {
-            last_group_marker as usize
-        } else {
-            (ENC_MARKER - last_group_marker) as usize
-        };
-    (group_cnt - 1) * ENC_GROUP_SIZE + last_group_size
-}
-
 /// Returns whether `encoded` bytes is encoded from `raw`. Returns `false` if `encoded` is invalid.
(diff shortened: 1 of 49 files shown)


components/tikv_util/src/codec/bytes.rs
```diff
@@ -278,6 +278,20 @@ pub fn decode_bytes_in_place(data: &mut Vec<u8>, desc: bool) -> Result<()> {
     }
 }
 
+// Returns the decoded key length of the given encoded key which is assumed valid without actually
+// decoding it.
+pub fn unchecked_decoded_len(encoded: &[u8], desc: bool) -> usize {
```
breeswish, line 283: Maybe it's fine to just use the length, since we will at most have a 1/8 * length + 7 inaccuracy.

zhongzc:
Updated, cc @breeswish

breeswish (approved):

src/storage/mvcc/reader/point_getter.rs
```diff
@@ -554,39 +557,47 @@ mod tests {
...
         let s = getter.take_statistics();
         assert_seek_next_prev(&s.write, 1, 0, 0);
+        assert_eq!(s.processed_size, 0);
 
         // Get a key that exists
         must_get_value(&mut getter, b"foo2", b"foo2v");
         let s = getter.take_statistics();
         // We have to check every version
         assert_seek_next_prev(&s.write, 1, 40, 0);
+        assert_eq!(s.processed_size, 269);
```
iosmanthus, line 572: Could we describe the key structure here? `269` seems too magic and difficult to maintain if the next contributor doesn't familiar with key content.

src/storage/mvcc/reader/point_getter.rs
```diff
@@ -554,39 +557,47 @@ mod tests {
...
         // Get again
         must_get_value(&mut getter, b"foo2", b"foo2v");
         let s = getter.take_statistics();
         assert_seek_next_prev(&s.write, 1, 40, 0);
+        assert_eq!(s.processed_size, 269);
```
iosmanthus, line 577: Ditto. We could use a variable instead. Remember to add some comment to its definition.

zhongzc:
@iosmanthus Updated

iosmanthus (approved):
LGTM

PR #10418, opened 2021-06-21
copr: make CM Sketch built with the same encoding as what TiDB assumes

time-and-fate:

Issue Number: close https://github.com/pingcap/tidb/issues/25638

Problem Summary:

When TiKV inserting bytes into CM Sketch, the encoding used is different from what TiDB assumes.

When querying values from CM Sketch in TiDB, it uses `tablecodec.EncodeValue()` to encode a datum into bytes.
In TiKV, we insert raw bytes from TableScanExecutor into CM Sketch. This is different from the results of `tablecodec.EncodeValue()` after we switch to row format v2.
The difference can be recognized if we compare `tablecodec.EncodeValue()` with `V1CompatibleEncoder::write_v2_as_datum()`.

I added some special logic to handle the differences.
An integration test is added to TiDB in https://github.com/pingcap/tidb/pull/25654.

(It may be more graceful to specially add a encode method for this usage. But what we get here are raw bytes from the table scan executor, and we can't decode them because decoding for some types is still unimplemented, so we can only do all the transformations in the "flattened" state.)

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
None
```


skyzh (approved):
LGTM

lzmhhh123 (approved):

lzmhhh123 (approved):

PR #10429, opened 2021-06-22
service: normal errors shouldn't break store connections

hicqu:

After receiving raft messages in the service layer, it's possible to meet errors when dispatching them to raftstore. Potential reasons could be the target Region is removed, or similar something.

We have met such a error log:
Lark20210622-174611

Those errors shouldn't break store connections.

### Related changes

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/src/server/service/kv.rs
+++ b/src/server/service/kv.rs
@@ -626,6 +626,10 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
                 let to_store_id = msg.get_to_peer().get_store_id();
                 if to_store_id != store_id {
                     warn!("store not match"; "self" => store_id, "given" => to_store_id);
+                    return future::err(Error::from(RaftStoreError::StoreNotMatch {
+                        to_store_id,
+                        my_store_id: store_id,
+                    }));
                 } else if let Err(e) = ch.send_raft_msg(msg) {
                     warn!("dispatch raft msg from gRPC to raftstore fail"; "err" => ?e);
                 }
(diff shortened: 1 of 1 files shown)


src/server/service/kv.rs
```diff
@@ -625,20 +625,16 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
                 RAFT_MESSAGE_RECV_COUNTER.inc();
                 let to_store_id = msg.get_to_peer().get_store_id();
                 if to_store_id != store_id {
-                    future::err(Error::from(RaftStoreError::StoreNotMatch {
```
BusyJay, line 628: I think this is intended. Otherwise no way to recover from a wrong resolved address.

hunterlxt (approved):
LGTM

src/server/service/kv.rs
```diff
@@ -625,20 +625,20 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
...
+                    warn!("store not match"; "self" => store_id, "given" => to_store_id);
```
BusyJay, line 628: Isn't it a redundant log?

src/server/service/kv.rs
```diff
@@ -625,20 +625,20 @@ impl<T: RaftStoreRouter<RocksEngine> + 'static, E: Engine, L: LockManager> Tikv
...
+                    return future::err(Error::from(RaftStoreError::StoreNotMatch {
                         to_store_id,
                         my_store_id: store_id,
-                    }))
-                } else {
-                    let ret = ch.send_raft_msg(msg).map_err(Error::from);
-                    future::ready(ret)
+                    }));
+                } else if let Err(e) = ch.send_raft_msg(msg) {
+                    warn!("dispatch raft msg from gRPC to raftstore fail"; "err" => ?e);
```
BusyJay, line 634: I'm afraid it can produce too many logs. And the flow control side effect is removed by this way, how to stop remote from sending a lot of messages when downstream is overloaded?

JmPotato:
Any updates?

hicqu:
It's fixed by #10555.

PR #10365, opened 2021-06-16
statistic: introduce resource metering integration test

zhongzc:
Signed-off-by: Zhenchi <zhongzc_arch@outlook.com>

### What is changed and how it works?

What's Changed:

- Introduce test suite
- Add test cases

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```

PR #10436, opened 2021-06-23
statistic: alter evict policy of resource metering reporter

zhongzc:

Problem Summary:

Previously, the LFU evict policy is based on accumulated CPU time in a minute. In case that the buffer comes full in the first few seconds, the coming records with other tags will be always dropped in the rest of the minute.

What's Changed:

Evict policy is based on CPU time in a second. Default `max_resource_groups` is changed to `200`.

### Related changes

- Based on https://github.com/tikv/tikv/pull/10365

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```


andylokandy (approved):

zhongzc:
cc @crazycs520 @breeswish

components/resource_metering/src/reporter.rs
```diff
@@ -119,54 +125,84 @@ impl Runnable for ResourceMeteringReporter {
...
-                                cpu_time.push(ms);
-                            }
-                            *total += ms;
-                        }
-                        None => {
-                            self.records
-                                .insert(tag.clone(), (vec![timestamp_secs], vec![ms], ms));
-                        }
-                    }
+                    *tmp_group_map.entry(tag.clone()).or_insert(0) += *ms as u32;
```
breeswish, line 141: How about using an Arc to wrap the resource group to reduce memory allocation?

components/resource_metering/src/reporter.rs
```diff
@@ -119,54 +125,84 @@ impl Runnable for ResourceMeteringReporter {
...
+                //   +----------+----------+-------+-------+-------+-------+-------+
+                //   | Threshold           |   30                      ^           |
+                //   +----------+----------+-------+-------+-------+-------+-------+
+                //   | Will be Saved       |   Y   |   Y   |   Y   |   N   |   N   |
+                //   +---------------------+-------+-------+-------+-------+-------+
+                let threshold = (tmp_group_map.len() > max_resource_groups)
+                    .then(|| {
+                        tmp_top_vec.extend(tmp_group_map.values());
+                        let (_, threshold, _) = tmp_top_vec
+                            .select_nth_unstable_by(max_resource_groups, |a, b| b.cmp(a));
```
breeswish, line 187: 👍

components/resource_metering/src/reporter.rs
```diff
@@ -119,54 +125,84 @@ impl Runnable for ResourceMeteringReporter {
...
 
-                if self.records.len() > self.config.max_resource_groups {
-                    self.find_top_k.clear();
-                    for (_, _, total) in self.records.values() {
-                        self.find_top_k.push(*total);
+                let timestamp_secs = records.begin_unix_time_secs;
+                let mut other = 0;
+                for (tag, ms) in tmp_group_map.drain() {
+                    if ms > threshold {
+                        let (ts, cpu) = self.records.entry(tag).or_insert((vec![], vec![]));
```
breeswish, line 196: Can we use `SmallVec` to reduce allocations?

breeswish (commented):
I think there is one additional work:

Previously we keep 200 records for each seconds, and finally just use the records for reporting.

Now we will have more and more records for each seconds (there may be +200 for each second). And just before the reporting we should evict and report only top 200.

components/resource_metering/src/reporter.rs
zhongzc: If so, it still needs to allocate later. 

https://github.com/tikv/tikv/blob/3044802f1171ca9edbd16f49214988e3a1a33b3c/components/resource_metering/src/reporter.rs#L241-L249

components/resource_metering/src/reporter.rs
zhongzc: It sounds to be complicated. 

It's a good option to use references, but we may either:
1. not to reuse `tmp_group_map` and `tmp_group_vec` or
2. to introduce some `unsafe`s

I prefer `1`: two allocations for eliminating others

zhongzc:
> I think there is one additional work:
> 
> Previously we keep 200 records for each seconds, and finally just use the records for reporting.
> 
> Now we will have more and more records for each seconds (there may be +200 for each second). And just before the reporting we should evict and report only top 200.

Do you mean evict 60+1 times every minute?

breeswish:
> > I think there is one additional work:
> > Previously we keep 200 records for each seconds, and finally just use the records for reporting.
> > Now we will have more and more records for each seconds (there may be +200 for each second). And just before the reporting we should evict and report only top 200.
> 
> Do you mean evict 60+1 times every minute?

@crazycs520 Please take a look about the evict behaviour. What kind of eviction do we have? Please make sure that the TiKV's data evict and TiDB's data evict works identically.

PR #10443, opened 2021-06-23
*: the `start_ts` of stale read request should not used to update `max_ts`

NingLin-P:

Issue Number: close #xxx 

Problem Summary:

Currently, if we use a future ts as the `start_ts` of stale read transaction, the stale read request will be rejected by TiKV with the `DataIsNotReady` error until the `resolved_ts` is catch up with the future ts. 

But before the request is rejected, we use its `starte_ts` to update the `max_ts`, because the `starte_ts` is a future ts which will result in the commit request reject by `commit_ts_expired` error and keep retrying, also 1pc transaction will fallback to 2pc.

Proposal: [xxx](url) 

What's Changed:

Don't update the `max_ts` with the `start_ts` from a stale read transaction, it is okay because TiKV will reject stale read request with `starts_ts` < `resolved_ts`, and the `resolved_ts` have already used to update the `max_ts`.

### Related changes

### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
Not use the stale read request's `start_ts` to update `max_ts` to avoid commit request keep retrying
```

### DIFF
--- a/src/coprocessor/endpoint.rs
+++ b/src/coprocessor/endpoint.rs
@@ -106,7 +106,9 @@ impl<E: Engine> Endpoint<E> {
 
     fn check_memory_locks(&self, req_ctx: &ReqContext) -> Result<()> {
         let start_ts = req_ctx.txn_start_ts;
-        self.concurrency_manager.update_max_ts(start_ts);
+        if !req_ctx.context.get_stale_read() {
+            self.concurrency_manager.update_max_ts(start_ts);
+        }
         if req_ctx.context.get_isolation_level() == IsolationLevel::Si {
             let begin_instant = Instant::now();
             for range in &req_ctx.ranges {
--- a/components/test_raftstore/src/util.rs
+++ b/components/test_raftstore/src/util.rs
@@ -862,12 +862,14 @@ pub fn kv_read(client: &TikvClient, ctx: Context, key: Vec<u8>, ts: u64) -> GetR
     client.kv_get(&get_req).unwrap()
 }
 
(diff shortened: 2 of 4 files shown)


nolouch:
PTAL @youjiali1995

sticnarf (approved):
LGTM

src/coprocessor/endpoint.rs
```diff
@@ -106,7 +106,9 @@ impl<E: Engine> Endpoint<E> {
 
     fn check_memory_locks(&self, req_ctx: &ReqContext) -> Result<()> {
         let start_ts = req_ctx.txn_start_ts;
-        self.concurrency_manager.update_max_ts(start_ts);
+        if !req_ctx.context.get_stale_read() {
```
youjiali1995, line 109: Need stale-read check memory locks? If `resolved-ts` reaches the `stale-read-ts`, there is no lock before the ts.

NingLin-P: There may have locks. For example,  right after the `resolved-ts` is advanced with a ts from PD, a prewrite request with a smaller `start_ts` come, in this case, `resolved-ts` won't decrease. 

Although `resolved-ts` guarantee that it must larger than any incoming `commit_ts`, so it is okay to ignore locks that less than `resolved-ts`. But I think it is too risky and not much benefit since reading historical data will encounter less locks, so currently, stale read request will check both the memory locks and persisted locks.

youjiali1995 (approved):

PR #10452, opened 2021-06-24
ignore some unstable CI about thread stats

hicqu:

Ignore some unstable CI about thread stats.

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```


hicqu:
The new added lines are commented out for now. Let's take a look at whether it can pass or not.

hicqu:
test_thread_io_statistics and test_thread_stat_io fail as expected.

PR #10454, opened 2021-06-25
allow to build one snapshot twice

hicqu:

There is a bug about deleting a snapshot: the associated item in `DataKeyManager` is not cleaned. So that if the one snapshot is built twice, it will fail at this line:
```rust
rename_tmp_cf_file_for_send(...) {
    mgr.link(tmp_file_path, cf_file_path);
}
```
because `cf_file_path` is already in the key manager.

Clean the item in `DataKeyManager` when deleting a cf file.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Avoid panic when building a snapshot twice if encryption enabled
```

### DIFF
--- a/components/raftstore/src/store/snap.rs
+++ b/components/raftstore/src/store/snap.rs
@@ -716,6 +716,10 @@ impl Snapshot {
                 self.mgr.rename_tmp_cf_file_for_send(cf_file)?;
             } else {
                 delete_file_if_exist(&cf_file.tmp_path).unwrap();
+                if let Some(ref mgr) = self.mgr.encryption_key_manager {
+                    let src = cf_file.tmp_path.to_str().unwrap();
+                    mgr.delete_file(src)?;
+                }
             }
 
             SNAPSHOT_CF_KV_COUNT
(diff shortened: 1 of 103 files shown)


hunterlxt (commented):
What is the reason for directly operating the key manager? Why not delete the previous file normally

hunterlxt (approved):

components/raftstore/src/store/snap.rs
```diff
@@ -1463,7 +1465,13 @@ impl SnapManagerCore {
...
             // It's ok that the cf file is moved but machine fails before `mgr.rename_file`
             // because without metadata file, saved cf files are nothing.
-            mgr.link_file(src, dst)?;
+            while let Err(e) = mgr.link_file(src, dst) {
+                if e.kind() == ErrorKind::AlreadyExists {
+                    mgr.delete_file(dst)?;
+                    continue;
+                }
+                return Err(e.into());
+            }
```
gengliqi, line 1474: It seems not necessary. These snapshot files can be deleted in the `drop` function.

gengliqi (commented):
Rest LGTM

gengliqi (approved):

PR #10457, opened 2021-06-25
add memory trace for raft messages and unstable entries

hicqu:

Add memory trace for raft messages and unstable entries.

### Check List 

Tests 

- Manual test
  
图片

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -148,10 +149,15 @@ where
 {
     fn drop(&mut self) {
         self.peer.stop();
+        let mut raft_messages_size = 0;
         while let Ok(msg) = self.receiver.try_recv() {
             let callback = match msg {
                 PeerMsg::RaftCommand(cmd) => cmd.callback,
                 PeerMsg::CasualMessage(CasualMessage::SplitRegion { callback, .. }) => callback,
+                PeerMsg::RaftMessage(im) => {
+                    raft_messages_size += im.heap_size;
```
5kbpers, line 158: I am afraid that sometimes here we would miss some messages. Could we use `Drop` to trace dropped messages?

hicqu: If `Drop` is implemented for `InspectedRaftMessage` we can't move things out from the message like this:
```rust
let InspectedRaftMessage { heap_size, mut msg } = msg;
```
And Implementing `Drop` for it requires to update `MEMTRACE_RAFT_MESSAGES` when it's constructed. To ensure this will make code more complex.

hicqu: Here the logic only called after a Region is destroyed. After this no messages can be sent to its channel. And all messages in its channel can be fetched in the `while` loop. So no messages can be missed theoretically.

components/server/src/server.rs
```diff
@@ -159,7 +159,7 @@ pub fn run_tikv(config: TiKvConfig) {
 
 const RESERVED_OPEN_FDS: u64 = 1000;
 
-const DEFAULT_METRICS_FLUSH_INTERVAL: Duration = Duration::from_millis(10_000);
+const DEFAULT_METRICS_FLUSH_INTERVAL: Duration = Duration::from_millis(1_000);
```
5kbpers, line 162: Why change this?

5kbpers (approved):
rest LGTM

components/server/src/server.rs
hicqu: `10s` is too long for memory usage increasing. For example, with a heavy write load TiKV can OOM in a less duration than `10s`, so metrics can't record it if the flush interval is `10s`.

components/raftstore/src/store/memory.rs
```diff
@@ -18,7 +18,9 @@ lazy_static! {
             applys,
             entry_cache,
             (raft_router, [alive, leak]),
-            (apply_router, [alive, leak])
+            (apply_router, [alive, leak]),
+            raft_message,
```
rleungx, line 22:
```suggestion
            raft_messages,
```

rleungx (approved):

PR #10459, opened 2021-06-25
raftstore: adjust max_down_peer_duration based on hibernate timeout

### LINKED ISSUE #10017: raftstore: down peer collection did not consider hibernate region
## Bug Report

When the leader heartbeat to PD, it collects down peers by checking whether a peer did not send raft message for `raftstore.max_peer_down_duration`. When hibernate region is in use, inactive regions will go to hibernate and wake up every 5m (config through `raftstore.peer_stale_state_check_interval`) and will be not raft message when the region is hibernated. So when the leader heartbeat to PD, it collects down peers and will take the time of hibernating into count (since the last time received raft message is when the region awake), and may wrongly report normal peers as down peers to PD and this could affect PD's schedule decision

Usually, the leader won't report to PD frequently, but there are some cases the leader will heartbeat to PD even the leader is hibernated, for example when the region size change due to compaction, the leader will report the region size to PD immediately.

yiwu-arbug:
Signed-off-by: yiwu <yiwu@pingcap.com>

Issue Number: close #10017

Problem Summary: Adjust `max_down_peer_duration` to be at least 2 * `peer_stale_state_check_interval`, so that peers in hibernated state would not be reported as down peer by mistake.

What's Changed: 
Adjust `max_down_peer_duration` to be at least 2 * `peer_stale_state_check_interval` when hibernated regions is enabled.

### Related changes

N/A

### Check List

Tests 

CI

Side effects

### Release note 

```release-note
Adjust max_down_peer_duration to 10 minutes. This is to avoid hibernated peers being reported as down peer by mistake.
```

### DIFF
--- a/components/raftstore/src/store/config.rs
+++ b/components/raftstore/src/store/config.rs
@@ -228,7 +228,7 @@ impl Default for Config {
             snap_mgr_gc_tick_interval: ReadableDuration::minutes(1),
             snap_gc_timeout: ReadableDuration::hours(4),
             messages_per_tick: 4096,
-            max_peer_down_duration: ReadableDuration::minutes(5),
+            max_peer_down_duration: ReadableDuration::minutes(10),
             max_leader_missing_duration: ReadableDuration::hours(2),
             abnormal_leader_missing_duration: ReadableDuration::minutes(10),
             peer_stale_state_check_interval: ReadableDuration::minutes(5),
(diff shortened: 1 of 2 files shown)


NingLin-P:
Should also change the value on `etc/config-template.toml`

NingLin-P (approved):

BusyJay (approved):

BusyJay:
Should also send PR to change docs.

nolouch (commented):
lgtm

yiwu-arbug:
> Should also send PR to change docs.

updating in https://github.com/pingcap/docs-cn/pull/6557

HunDunDM:
Will it cherry-pick?

PR #10478, opened 2021-06-29
Remove commit of kvproto for every Cargo.toml

Little-Wallace:

Issue Number: close #xxx 

Problem Summary:
The package manager Cargo could keep the dependency library in one commit version, so we do not need to set it in every Cargo.toml, or we must change a lot of files every time we need to update kvproto.

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 
- None

### DIFF
--- a/src/import/sst_service.rs
+++ b/src/import/sst_service.rs
@@ -589,12 +589,3 @@ fn pb_error_inc(type_: &str, e: &errorpb::Error) {
 
     IMPORTER_ERROR_VEC.with_label_values(&[type_, label]).inc();
 }
-
-fn make_request_header(mut context: Context) -> RaftRequestHeader {
-    let region_id = context.get_region_id();
-    let mut header = RaftRequestHeader::default();
-    header.set_peer(context.take_peer());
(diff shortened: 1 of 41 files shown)


src/import/sst_service.rs
```diff
@@ -554,6 +554,15 @@ where
         self.threads.spawn_ok(buf_driver);
         self.threads.spawn_ok(handle_task);
     }
+
+    fn duplicate_detect(
```
BusyJay, line 558: I don't think it's necessary as new grpcio will provide a default implementation.

BusyJay (approved):
Rest LGTM

src/import/sst_service.rs
Little-Wallace: I have removed it.

MyonKeminta (approved):

Little-Wallace:
cc @gregwebs

Issue #10460, opened 2021-06-26
5.1.0  thread 'sched-worker-pool-0' has overflowed its stack fatal runtime error: stack overflow


darren:
### What version of TiKV are you using?

```
TiKV 
Release Version:   5.1.0
Edition:           Community
Git Commit Hash:   5a2d13fe14a31edee3c98715ac575b1663c0df47
Git Commit Branch: release-5.1
UTC Build Time:    2021-06-25 03:02:28
Rust Version:      rustc 1.53.0-nightly (16bf626a3 2021-04-14)
Enable Features:   jemalloc mem-profiling portable sse protobuf-codec test-engines-rocksdb cloud-aws cloud-gcp
Profile:           release
```

config:

```toml
[server]
log-level = "info"
[metric]
interval = "15s"
address = ""
job = "tikv"
[raftstore]
sync-log = true
raft-base-tick-interval = "2s"
[storage.block-cache]
shared = true
capacity = "2GB"
[pd]
[rocksdb]
compaction-readahead-size = "8MB"
[rocksdb.defaultcf]
compression-per-level = ["no", "snappy", "lz4", "lz4", "lz4", "zstd", "zstd"]
max-write-buffer-number = 15
[storage]
scheduler-worker-pool-size = 8
reserve-space = "0"
```
### What operating system and CPU are you using?

```
processor	: 31
vendor_id	: GenuineIntel
cpu family	: 6
model		: 45
model name	: Intel(R) Xeon(R) CPU E5-2680 0 @ 2.70GHz
stepping	: 7
microcode	: 0x710
cpu MHz		: 2201.343
cache size	: 20480 KB
physical id	: 1
siblings	: 16
core id		: 7
cpu cores	: 8
apicid		: 47
initial apicid	: 47
fpu		: yes
fpu_exception	: yes
cpuid level	: 13
wp		: yes
[...]

PR #10499, opened 2021-07-01
storage: avoid sched_pool stack overflow when commands fail successively

sticnarf:

Potentially #10460.

If many queuing tasks are expired, they will be handled in a recursion like `try_to_wake_up -> finish_with_err -> release_lock -> try_to_wake_up`.

In the `try_to_wake_up` function, if the command fails because the deadline is exceeded, we change to spawn a task to the pool to handling the error. So the stack depth will not accumulate.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU (but only in an extreme case)

### Release note 

```release-note
Prevent the risk of stack overflow when handling many expired commands.
```

### DIFF
--- a/src/storage/txn/scheduler.rs
+++ b/src/storage/txn/scheduler.rs
@@ -389,9 +389,13 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
                 // Spawn the finish task to the pool to avoid stack overflow
                 // when many queuing tasks fail successively.
                 let this = self.clone();
-                self.inner.worker_pool.pool.spawn_force(async move {
-                    this.finish_with_err(cid, err);
-                });
+                self.inner
+                    .worker_pool
+                    .pool
+                    .spawn(async move {
+                        this.finish_with_err(cid, err);
+                    })
+                    .unwrap();
             }
         }
     }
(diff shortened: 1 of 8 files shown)


src/storage/txn/scheduler.rs
```diff
@@ -609,49 +613,53 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
-                    tls_collect_read_duration(tag.get_str(), elapsed);
-                }
-                .in_resource_metering_tag(resource_tag),
-            )
-            .unwrap();
+                tls_collect_read_duration(tag.get_str(), elapsed);
+            }
+            .in_resource_metering_tag(resource_tag),
+        );
+        if let Err(_full) = res {
```
youjiali1995, line 658: Seems it's an infinite pool.

src/storage/txn/scheduler.rs
```diff
@@ -1104,4 +1112,49 @@ mod tests {
...
+            Arc::new(AtomicBool::new(true)),
+            false,
+        );
+
+        let mut lock = Lock::new(&[Key::from_raw(b"b")]);
+        let cid = scheduler.inner.gen_id();
+        assert!(scheduler.inner.latches.acquire(&mut lock, cid));
+
+        // Push lots of requests in the queue.
+        for _ in 0..65536 {
```
youjiali1995, line 1135: Can it reproduce the stack overflow?

src/storage/txn/scheduler.rs
sticnarf: Ah, right... I misread the code.

src/storage/txn/scheduler.rs
sticnarf: Yes. Without re-spawning the task, running this test will cause a stack overflow.

youjiali1995 (approved):

MyonKeminta (approved):

PR #10501, opened 2021-07-01
don't clean snapshot when applying

hicqu:

Fix a panic bug about "file doesn't exist" when applying a snapshot.
Here is a log sector of tiflash:
```
[2021/06/24 10:49:12.737 +08:00] [INFO] [raft.rs:1064] ["became follower at term 9"] [term=9] [raft_id=175749763] [region_id=157802790]
[2021/06/24 10:49:12.997 +08:00] [INFO] [snap.rs:239] ["saving snapshot file"] [file=/data1/tidb-data/tiflash-9000/flash/snap/rev_157802790_9_656810_(default|lock|write).ss
[2021/06/24 10:49:13.008 +08:00] [INFO] [raft_log.rs:531] ["log [committed=0, persisted=0, applied=0, unstable.offset=1, unstable.entries.len()=0] starts to restore snapsho
[2021/06/24 10:49:13.008 +08:00] [INFO] [raft.rs:2443] ["switched to configuration"] [config="Configuration { voters: Configuration { incoming: Configuration { voters: {157[2021/06/24 10:49:13.008 +08:00] [INFO] [raft.rs:2427] ["restored snapshot"] [snapshot_term=9] [snapshot_index=656810] [last_term=9] [last_index=656810] [commit=656810] [ra[2021/06/24 10:49:13.008 +08:00] [INFO] [raft.rs:2308] ["[commit: 656810, term: 9] restored snapshot [index: 656810, term: 9]"] [snapshot_term=9] [snapshot_index=656810] [c
[2021/06/24 10:49:13.008 +08:00] [INFO] [peer_storage.rs:1130] ["begin to apply snapshot"] [peer_id=175749763] [region_id=157802790]
[2021/06/24 10:49:13.008 +08:00] [INFO] [peer_storage.rs:1173] ["apply snapshot with state ok"] [state="applied_index: 656810 truncated_state { index: 656810 term: 9 }"] [r[2021/06/24 10:49:13.008 +08:00] [INFO] [peer.rs:2968] ["snapshot is applied"] [region="id: 157802790 start_key: 7480000000000000FF9A5F7292000000E0FF6C9C880000000000FA end_
[2021/06/24 11:13:40.813 +08:00] [INFO] [peer.rs:1456] ["target peer id is larger, destroying self"] [target_peer="id: 175755143 store_id: 148534374 role: Learner"] [peer_i[2021/06/24 11:14:24.752 +08:00] [INFO] [peer.rs:735] ["deleting applied snap file"] [snap_file=157802790_9_656810] [peer_id=175749763] [region_id=157802790]
[2021/06/24 13:33:35.553 +08:00] [INFO] [store.rs:976] ["region is applying snapshot"] [store_id=148534374] [region="id: 157802790 start_key: 7480000000000000FF9A5F72920000
[2021/06/24 13:33:35.553 +08:00] [INFO] [peer.rs:189] ["create peer"] [peer_id=175749763] [region_id=157802790]
[2021/06/24 13:33:35.553 +08:00] [INFO] [raft.rs:2443] ["switched to configuration"] [config="Configuration { voters: Configuration { incoming: Configuration { voters: {157
[2021/06/24 13:33:35.553 +08:00] [INFO] [raft.rs:1064] ["became follower at term 9"] [term=9] [raft_id=175749763] [region_id=157802790]
[2021/06/24 13:33:35.553 +08:00] [INFO] [raft.rs:375] [newRaft] [peers="Configuration { incoming: Configuration { voters: {157802793, 157802794, 157802795} }, outgoing: Con
[2021/06/24 13:33:35.553 +08:00] [INFO] [raw_node.rs:285] ["RawNode created with id 175749763."] [id=175749763] [raft_id=175749763] [region_id=157802790]
[2021/06/24 13:33:35.638 +08:00] [INFO] [region.rs:377] ["begin apply snap data"] [region_id=157802790]
[2021/06/24 13:33:35.639 +08:00] [INFO] [region.rs:411] ["apply data to engine-store"] [region_id=157802790]
[2021/06/24 13:33:35.639 +08:00] [ERROR] [region.rs:469] ["failed to apply snap!!!"] [err_code=KV:Raftstore:SnapUnknown] [err="Other(\"[components/raftstore/src/store/worke

```

The root cause is the snapshot is deleted when the peer is deleted. However the deletion of the peer can be not persisted, so after restart the instance, the peer will continue to apply the snapshot.

This PR ensures that snapshot won't be deleted during applying.

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -742,7 +742,8 @@ where
         let s = self.fsm.peer.get_store();
         let compacted_idx = s.truncated_index();
         let compacted_term = s.truncated_term();
-        let is_applying_snap = s.is_applying_snapshot();
+        let is_applying_snap =
+            s.is_applying_snapshot() || s.peer_state() == Some(PeerState::Applying);
         for (key, is_sending) in snaps {
             if is_sending {
                 let s = match self.ctx.snap_mgr.get_snapshot_for_gc(&key, is_sending) {
@@ -792,7 +793,7 @@ where
                     }
                 }
             } else if key.term <= compacted_term
-                && (key.idx < compacted_idx || key.idx == compacted_idx && !is_applying_snap)
+                && (key.idx < compacted_idx || (key.idx == compacted_idx && !is_applying_snap))
             {
                 info!(
                     "deleting applied snap file";
@@ -1933,6 +1934,11 @@ where
         // The initialized flag implicitly means whether apply fsm exists or not.
         if job.initialized {
             // Destroy the apply fsm first, wait for the reply msg from apply fsm
+            info!(
+                "schedule ApplyTask::destroy";
+                "region_id" => job.region_id,
+                "peer_id" => self.fsm.peer_id(),
+            );
             self.ctx
                 .apply_router
                 .schedule_task(job.region_id, ApplyTask::destroy(job.region_id, false));
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -3272,6 +3272,7 @@ where
 
     /// Handles peer destroy. When a peer is destroyed, the corresponding apply delegate should be removed too.
     fn handle_destroy<W: WriteBatch<EK>>(&mut self, ctx: &mut ApplyContext<EK, W>, d: Destroy) {
+        fail_point!("before_apply_handle_destroy", |_| {});
         assert_eq!(d.region_id, self.delegate.region_id());
         if d.merge_from_snapshot {
             assert_eq!(self.delegate.stopped, false);
--- a/components/raftstore/src/store/fsm/store.rs
+++ b/components/raftstore/src/store/fsm/store.rs
@@ -2061,6 +2061,7 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
                 "schedule snap gc";
                 "store_id" => self.fsm.store.id,
                 "region_id" => region_id,
+                "snaps" => ?snaps,
             );
             let gc_snap = PeerMsg::CasualMessage(CasualMessage::GcSnap { snaps });
             match self.ctx.router.send(region_id, gc_snap) {
--- a/components/raftstore/src/store/peer_storage.rs
+++ b/components/raftstore/src/store/peer_storage.rs
@@ -965,6 +965,15 @@ where
         self.region = region;
     }
 
+    pub fn peer_state(&self) -> Option<PeerState> {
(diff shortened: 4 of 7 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -742,7 +742,8 @@ where
         let s = self.fsm.peer.get_store();
         let compacted_idx = s.truncated_index();
         let compacted_term = s.truncated_term();
-        let is_applying_snap = s.is_applying_snapshot();
+        let is_applying_snap =
+            s.is_applying_snapshot() || s.peer_state() == Some(PeerState::Applying);
```
BusyJay, line 746: How about checking whether `applied_index == truncated_index` instead?

BusyJay:
I'm afraid other usage of `is_applying_snapshot` can also go wrong.

NingLin-P:
I had checked some code and found that if the apply snapshot task is canceled before the following code:

https://github.com/tikv/tikv/blob/3587fe8ce5c1894a06f1c0d256380d1d3190e2a0/components/raftstore/src/store/worker/region.rs#L380-L386

the peer's state won't recover from `PeerState::Applying` to `PeerState::Normal`, although the apply snapshot task usually canceled because the peer is about to destroy and its state will set to `PeerState::Tombstone` when it is destroying, but I'm not sure is it okay.

Also, the above `writebatch` is written with the default option where the `sync` is `false`, it may possible that we found the peer is not applying snapshot so delete the snapshot files but after restart the peer's state is still `PeerState::Applying` and we still get panic because snapshot files not found.

JaySon-Huang:
Hi @hicqu @BusyJay, could we continue to address comments and fix this bug?

flowbehappy:
Hi, can we push this pr forward? @hicqu @BusyJay @NingLin-P

you06:
Does this bug still exist in master and other release branches?

hicqu:
> Does this bug still exist in master and other release branches?

Yes, I guess so.

you06:
> Yes, I guess so.

If I understand @NingLin-P's comment correctly, even with this patch, the snapshot may still be cleaned up and there is the same error. Is there any plan to fix it?

BusyJay:
Please also send it as an issue so others can pick it up.

CalvinNeo:
@hicqu Hi, can we have this pr merge now?

flowbehappy:
@BusyJay @tonyxuqqi Ping

tonyxuqqi:
It looks that master already have similar logic more than 6 months ago. And #10501 just add some enhancement on it. This PR is not needed.

Connor1996:
> 

You referred to the same PR as this

PR #10505, opened 2021-07-01
Set write_global_seqno to false while ingesting SST files.

### LINKED ISSUE #10430: rocksdb: add sst_ingestion_write_global_seqno option and set it to false by default
## Development Task
Older versions of RocksDB write global seqno (the sequence number of the ingested SST file) in the ingested SST file, which incur an IO while holding db_mutex (per RocksDB instance global mutex), which could cause latency jitter for RocksDB writes. Newer versions of RocksDB provide the `write_global_seqno` (see https://github.com/tikv/rocksdb/blob/6.4.tikv/include/rocksdb/options.h#L1500) as an option and allow disabling it. When disabled, RocksDB instead write the global seqno into manifest file, which is without holding db_mutex. We need to add the corresponding option in rust-rocksdb (see https://github.com/tikv/rust-rocksdb/blob/master/src/rocksdb_options.rs#L2029) then in TiKV (to the `DbConfig` struct, see https://github.com/tikv/tikv/blob/master/src/config.rs#L889). Let's name this option `sst_ingestion_write_global_seqno` and default it to false. Related interfaces in engine trait (components/engine_traits) also need to be updated. Then whenever ingest SST is called, set the corresponding option to RocksDB.

We need to provide it as an option instead of just hard-code the option value to false. This is to provide some backward compatibility. TiKV < v3.0 (e.g. v2.1) was on RocksDB v5.15, which does not support this option and wouldn't be able to read the DB after using this option. But the need to downgrade to v2.1 should be rare.

v01dstar:

Issue Number: close #10430

Problem Summary: See details in issue #10430 

What's Changed:  A new configurable option write_global_seqno is added and set to false by default to reduce lock contentions and remain backward compatibility at the same time.

### Related changes

- PR that adds this option to Rocks DB rust client in `tikv/rust-rocksdb`: https://github.com/tikv/rust-rocksdb/pull/631

### Check List 

Tests 

- Unit test

Side effects

N/A

### Release note 

```release-note
* rocksdb: set write_global_seqno=false while ingesting file, which remove an IO while holding rocksdb lock, potentially improve tail latency for writes. Note with the change, the data format is incompatible with TiKV < 3.0.
```


v01dstar:
This implementation adds the configurable option into `tikv/components/sst_importer/src/config.rs` rather than `tikv/src/config.rs` which was proposed in #10430 because adding the option into `tikv` while using it in `sst_importer` brings circular dependency.

components/engine_panic/src/import.rs
```diff
@@ -27,7 +27,16 @@ impl IngestExternalFileOptions for PanicIngestExternalFileOptions {
     fn new() -> Self {
         panic!()
     }
+
     fn move_files(&mut self, f: bool) {
         panic!()
     }
+
+    fn get_write_global_seqno(&mut self) -> bool {
```
yiwu-arbug, line 35:
```suggestion
    fn get_write_global_seqno(self) -> bool {
```

same for other call sites.

components/sst_importer/src/config.rs
```diff
@@ -14,6 +14,7 @@ pub struct Config {
     ///
     /// Default is 10m.
     pub import_mode_timeout: ReadableDuration,
+    pub write_global_seqno: bool,
```
yiwu-arbug, line 17: `SstImporter` is for data import (BR & lightning) and not the right place to have the `write_global_seqno` config. Let's do the following:
* still have the config in `DbConfig` and store the flag with `RocksEngine` struct
* remove `IngestExternalFileOptions::new` method, instead construct the ingestion options in `Engine::ingest_external_file_cf` internal.

yiwu-arbug:
`cases::test_import_service::test_ingest_reentrant` is failing and probably related.

components/engine_rocks/src/import.rs
```diff
@@ -55,4 +55,26 @@ impl IngestExternalFileOptions for RocksIngestExternalFileOptions {
...
+        self.0.set_write_global_seqno(f);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_ingest_external_file_options() {
```
Connor1996, line 73: we don't need to test it again

components/engine_panic/src/import.rs
v01dstar: Will apply the change once https://github.com/tikv/rust-rocksdb/pull/633 is merged.

v01dstar:
PTAL

src/config.rs
```diff
@@ -954,6 +954,8 @@ pub struct DbConfig {
     pub raftcf: RaftCfConfig,
     #[online_config(skip)]
     pub titan: TitanDBConfig,
+    #[online_config(skip)]
+    pub sst_ingestion_write_global_seqno: bool,
```
Connor1996, line 958: Do we necessarily need this config? @yiwu-arbug

yiwu-arbug:
@v01dstar Can you do me a favor, adding the inline comment per https://github.com/tikv/tikv/issues/10430#issuecomment-873318183?

src/config.rs
yiwu-arbug: I was thinking, in case user upgrade from <=2.1 version to >=5.2 version directly, they can set this option to true on upgrade to make sure they can downgrade back to <=2.1 version. But I guess such an upgrade will meet with a ton of other compatibility issues. What do you think?

yiwu-arbug:
Also please fix build.

tests/failpoints/cases/test_import_service.rs
```diff
@@ -126,8 +126,10 @@ fn test_ingest_reentrant() {
     assert!(!resp.has_error());
 
     let checksum2 = calc_crc32(save_path).unwrap();
-    // Checksums are different because ingest changed global seqno in sst file.
-    assert_ne!(checksum1, checksum2);
+    // TODO: Remove this once write_global_seqno is deprecated.
+    // Checksums are the same since the global seqno in the SST file no longer gets updated with the
+    // default setting, which is write_global_seqno=false.
+    assert_eq!(checksum1, checksum2);
```
yiwu-arbug, line 132: Is it possible to explicitly set write_global_seqno=false for the test?

yiwu-arbug:
Please fix test_serde_custom_tikv_config test. The other failing test looks like just being flaky.

tests/failpoints/cases/test_import_service.rs
v01dstar: Since we are not making this option configurable as discussed offline, this comment is no longer applicable.

yiwu-arbug:
Per our offline discussion, is it possible to also remove the `reset_global_seq` logic when it is not necessary?

components/raftstore/src/store/snap.rs
```diff
@@ -597,7 +597,11 @@ impl Snapshot {
             }
 
             if !plain_file_used(cf_file.cf) {
-                // Reset global seq number.
+                // TODO: remove reset_global_seq() once write_global_seqno is only allowed to be set
```
yiwu-arbug, line 600: Do you decide to keep the `reset_global_seq` logic for backward compatiblity? If so the inline comment need some update. And shall we add a test for the backward compatiblity?

v01dstar:
I tried to ingests sst files that haves not-null global seqno and it failed [here](https://github.com/facebook/rocksdb/blob/ca89ac2ba997dfa0e135bd75d4ccf6f5774a7eff/table/block_based_table_reader.cc#L730), w/ or w/o `write_global_seqno`.

It seems to me that RocksDB only allows ingesting files that have global seqno = 0, since the `largest_seqno` is passed all the way from TableReaderOption [here](https://github.com/facebook/rocksdb/blob/c8665611bcfae9ac1c3f7accbfc02f2b762ce088/table/block_based/block_based_table_factory.cc#L498), and the instance of TableReaderOption used by ingestion job [here](https://github.com/facebook/rocksdb/blob/0b75b223212ca19a07ad6259b37450c287ff34ae/db/external_sst_file_ingestion_job.cc#L559) does not set it (means it is 0).

So, my guess is, as long as we still want to be able to ingest files that have non-zero global_seqno,  we have to keep the `reset_global_seq`. We can only deprecate it if we do not care about backward compatibility (ingesting sst files that have non-zero global_seqno) or the validation logic in RockDB has been changed.

Correct me if I am wrong.

components/raftstore/src/store/snap.rs
```diff
@@ -597,7 +597,9 @@ impl Snapshot {
             }
 
             if !plain_file_used(cf_file.cf) {
-                // Reset global seq number.
+                // TODO: remove reset_global_seq() if possible. TiKV does not write global_seqno
+                // into to-be-ingested SST files by default, this is only kept for backward
```
Connor1996, line 601: `reset_global_seq` is to make sure the ingest can be retryable. now, it doesn't write global seqno anymore, so there is no need to reset global seq.

components/raftstore/src/store/snap.rs
```diff
@@ -597,7 +597,9 @@ impl Snapshot {
...
+                // compatibility.
```
yiwu-arbug, line 602:
```suggestion
                // Prior to v5.2.0, TiKV use 
```

yiwu-arbug:
> I tried to ingests sst files that haves not-null global seqno and it failed [here](https://github.com/facebook/rocksdb/blob/ca89ac2ba997dfa0e135bd75d4ccf6f5774a7eff/table/block_based_table_reader.cc#L730), w/ or w/o `write_global_seqno`.
> 
> It seems to me that RocksDB only allows ingesting files that have global seqno = 0, since the `largest_seqno` is passed all the way from TableReaderOption [here](https://github.com/facebook/rocksdb/blob/c8665611bcfae9ac1c3f7accbfc02f2b762ce088/table/block_based/block_based_table_factory.cc#L498), and the instance of TableReaderOption used by ingestion job [here](https://github.com/facebook/rocksdb/blob/0b75b223212ca19a07ad6259b37450c287ff34ae/db/external_sst_file_ingestion_job.cc#L559) does not set it (means it is 0).
> 
> So, my guess is, as long as we still want to be able to ingest files that have non-zero global_seqno, we have to keep the `reset_global_seq`. We can only deprecate it if we do not care about backward compatibility (ingesting sst files that have non-zero global_seqno) or the validation logic in RockDB has been changed.
> 
> Correct me if I am wrong.

It all make sense. Let's keep the `reset_global_seq` logic, but more it inside of `ingest_external_file_cf`, and update inline comment to the following
```
// Prior to v5.2.0, TiKV use `write_global_seqno=true` for ingestion. For backward compatibility, in case TiKV is retrying an ingestion job generated by older version, it needs to reset the global seqno to 0.
```

components/raftstore/src/store/snap.rs
yiwu-arbug: This is for backward compatibility. See discussion above.

components/raftstore/src/store/snap.rs
yiwu-arbug: > Could you please be more specific on what type of test you want see?
> 
> As for the compatibility issue, I listed my observations below.

basically the `test_ingest_reentrant` test, but use `write_global_seqno=true` in the first ingestion.

Connor1996 (approved):
LGTM

yiwu-arbug (approved):
LGTM

PR #10535, opened 2021-07-08
Remove all request snapshot code

### LINKED ISSUE #10526: Remove all request snapshot code
Request snapshot is used to request a snapshot explicitly from follower to fix problems like corruption. But the API has never been used or tested. Technically, it's an unreleased feature. I propose to remove all related code for now to make code base clean.

jiangyuzhao:

Issue Number: close #10526

Problem Summary: 

Request snapshot is used to request a snapshot explicitly from follower to fix problems like corruption. But the API has never been used or tested. Technically, it's an unreleased feature. So we can remove these code to make code clean.

Just remove the code about request snapshot

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.
```


components/error_code/src/raft.rs
```diff
@@ -31,8 +30,8 @@ impl ErrorCodeExt for Error {
             Error::CodecError(_) => CODEC_ERROR,
             Error::Exists { .. } => EXISTS,
             Error::NotExists { .. } => NOT_EXISTS,
-            Error::RequestSnapshotDropped => REQUEST_SNAPSHOT_DROPPED,
             Error::ConfChangeError(_) => CONF_CHANGE_ERROR,
+            _ => unreachable!(),
```
BusyJay, line 34: I suggest to write the type explicitly to avoid accidentally missing new definitions in the future.

BusyJay (approved):
/cc @overvenus @NingLin-P

NingLin-P (approved):
LGTM

PR #10536, opened 2021-07-09
support both cgroup v1 and v2

hicqu:

Refactor `tikv_util::sys::cgroup` to support both v1 and v2.
And things about `mount point` is removed because they are not used actually.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```


components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,236 @@
...
+            assert_eq!(limit, expect);
+        }
     }
 
     #[test]
-    fn test_cgroup_sys() {
-        let temp_dir = TempDir::new().unwrap();
-        let path1 = temp_dir.path().join("cgroup");
-        let mut f1 = File::create(path1.clone()).unwrap();
-        f1.write_all(b"4:memory:/kubepods/burstable/poda2ebe2cd-64c7-11ea-8799-eeeeeeeeeeee/a026c487f1168b7f5442444ac8e35161dfcde87c175ef27d9a806270e267a575\n").unwrap();
```
BusyJay, line 486: I think this can be reserved. It's important that the code can parse files in real life.

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,236 @@
...
 
-    let mut sep_pos = MountInfoFieldPart1::OptionalFields as usize;
-    let mut found_sep = false;
-    for field in &fields[MountInfoFieldPart1::OptionalFields as usize..] {
-        if field == OPTIONAL_FIELDS_SEP {
-            found_sep = true;
-            break;
+    /// -1 means no limit.
+    pub fn memory_limit_in_bytes(&self) -> i64 {
+        if self.is_v2 {
```
BusyJay, line 31: How about adding some comments explaining the format of files and difference between versions?

BusyJay (commented):
Much cleaner! It's always difficult for me to figure out how original code works.

components/tikv_util/src/sys/cgroup.rs
hicqu: Now it's added into `test_parse_proc_cgroup_v1`.

components/tikv_util/src/sys/cgroup.rs
hicqu: Now I add a comment sector.

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,348 @@
...
-    // the host file system path in the mount namespace the *MountPoint belongs to.
-    pub fn translate(&self, abs_path: &str) -> Result<String, Error> {
-        let abs = Path::new(abs_path);
-        let rel = abs.strip_prefix(&self.root)?;
-        let path = Path::new(&self.mount_point).join(rel);
-        if let Some(path) = path.to_str() {
-            Ok(String::from(path))
+impl CGroupSys {
+    pub fn new() -> Self {
+        let mut f = File::open(&format!("/proc/{}/cgroup", process::id())).unwrap();
```
sticnarf, line 39: A better way is to just use `std::fs::read_to_string`.

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,348 @@
...
-        for seg in content.split(',') {
-            if let Some((start, end)) = seg.split_once('-') {
-                cpuset.extend(start.parse::<usize>()?..=end.parse()?)
-            } else if !seg.is_empty() {
-                cpuset.insert(seg.parse()?);
-            }
+fn parse_proc_cgroup_v1(lines: &str) -> HashMap<String, String> {
+    let mut subsystems = HashMap::new();
+    let lines = lines
+        .split('\n')
```
sticnarf, line 133: There is a provided function `str::lines`

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,341 @@
...
-            found_sep = true;
-            break;
+    /// -1 means no limit.
+    pub fn memory_limit_in_bytes(&self) -> i64 {
+        if self.is_v2 {
+            let group = self.cgroups.get("").unwrap();
+            let path = format!("/sys/fs/cgroup/{}/memory.max", group);
+            if let Ok(mut f) = File::open(&path) {
+                let mut buffer = String::new();
+                f.read_to_string(&mut buffer).unwrap();
```
sticnarf, line 56: Lots of patterns like this can be simplified by using `fs::read_to_string`.

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,341 @@
...
+    for v in value.split(',') {
+        if v.contains('-') {
+            let mut v = v.split('-');
+            let s = v.next().unwrap().parse::<usize>().unwrap();
+            let e = v.next().unwrap().parse::<usize>().unwrap();
+            for x in s..=e {
+                cores.insert(x);
             }
+        } else {
+            let s = v.parse::<usize>().unwrap();
```
sticnarf, line 165: I test this PR on my own machine and it panics here. It is possible that `cpuset.cpus` is just an empty file.

sticnarf (approved):

sticnarf:
/lgtm cancel
Sorry I mis-clicked the approve...

sticnarf (changes requested):
Sorry I mis-clicked the approve...

sticnarf (approved):
LGTM

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,322 @@
...
-        sep_pos += 1;
-    }
-    if !found_sep {
-        return Err(
-            "mount point format invalid, doesn't found optional field separator".to_string(),
-        );
+    /// -1 means no limit.
+    pub fn memory_limit_in_bytes(&self) -> i64 {
+        let path = if self.is_v2 {
+            let group = self.cgroups.get("").unwrap();
```
BusyJay, line 51: How about log a warning instead? I worry if TiKV can't parse some special system and then cause panic.

components/tikv_util/src/sys/cgroup.rs
```diff
@@ -1,520 +1,322 @@
...
+    subsystems
 }
 
-#[derive(Debug)]
-pub struct CGroupSys {
-    // subsystem name -> CGroup
-    pub cgroups: HashMap<String, CGroup>,
+fn parse_proc_cgroup_v2(lines: &str) -> HashMap<String, String> {
+    let subsystems = parse_proc_cgroup_v1(lines);
+    assert_eq!(subsystems.len(), 1);
```
BusyJay, line 117: What if it doesn't equal to 1?

hicqu: It's descripted by the spec:
> “/proc/$PID/cgroup” lists a process’s cgroup membership. If legacy cgroup is in use in the system, this file may contain multiple lines, one for each hierarchy. The entry for cgroup v2 is always in the format “0::$PATH”

components/tikv_util/src/sys/cgroup.rs
hicqu: According to the spec:
> “/proc/$PID/cgroup” lists a process’s cgroup membership. If legacy cgroup is in use in the system, this file may contain multiple lines, one for each hierarchy. The entry for cgroup v2 is always in the format “0::$PATH”
For cgroup-v2 there is only one subsystem named `""`, so here just `unwrap` is OK.

I will attach the spec link to the PR and comment here.

BusyJay (approved):

DanielZhangQD (approved):
LGTM

PR #10538, opened 2021-07-09
storage: fix that reader statistics are not collected on errors

sticnarf:

In many commands, reader statistics are collected after everything is done. But during the execution, if there is an error, the function returns so we forget to collect reader statistics.

`ReaderWithStats` is a wrapper of `SnapshotReader` but with a reference to the outer statistics. When `ReaderWithStats` is dropped, the statistics in the inner reader are added to the outer statistics.

Most commands are modified to replace `SnapshotReader` with `ReaderWithStats` to become consistent, even if there were no problems before.

### Check List 

Tests 

### Release note 

```release-note
Fix the bug that reader statistics are not collected on errors.
```


longfangsong (approved):

MyonKeminta:
🤔 Do we need to separate the statistics data of successful requests and failed requests?

sticnarf:
> thinking Do we need to separate the statistics data of successful requests and failed requests?

Emm, seems not so necessary to me.

Do you mean that it may be useful to investigate the proportion of cursor operations of failed requests?

MyonKeminta:
I'm thinking if mixing statistics of successful and failed requests will make it inconvenient to investigate problems. I think it might be ok, but I'm not sure.

sticnarf:
> I'm thinking if mixing statistics of successful and failed requests will make it inconvenient to investigate problems. I think it might be ok, but I'm not sure.

Okay... I can have a try to separate the statistics for successful and failed requests.

MyonKeminta (approved):

PR #10544, opened 2021-07-10
Avoid duration calculation panic when clock jumps back

tabokie:

Problem Summary:

Duration utilities for standard clock can panic, see https://github.com/rust-lang/rust/issues/86470#issuecomment-864544441

A relate internal tracking issue: https://internal.pingcap.net/jira/browse/ONCALL-3295

For most of our use cases, it isn't necessary to enforce a strict monotonic timer at userland, like rust does:
https://github.com/rust-lang/rust/blob/80184183ba0a53aa4f491753de9502acd3d6920c/library/std/src/time.rs#L223-L250

Saturating subtraction is enough.

What's Changed:

- `std::time::Instant::duration_since` -> `std::time::Instant::saturating_duration_since`
- `std::time::Instant::elapsed` -> `tikv_util::time::Instant::saturating_elapsed`
- `std::time::Instant::elapsed` -> `InstantExt::saturating_elapsed` when `std::time::Instant` can't be replaced:
  - 3rd-party functions that take `std::time::Instant` as input
  - Critical logic that uses `std::time::Instant`, such as `fsm::Store`

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Fix duration calculation panics on certain platforms
```


yiwu-arbug (approved):
LGTM except for raftstore changes. @gengliqi do you mind confirm raftstore changes are good?

components/tikv_util/src/time.rs
```diff
@@ -274,13 +285,30 @@ impl Instant {
...
+                Instant::saturating_elapsed_duration(now, t)
+            }
+            Instant::MonotonicCoarse(t) => {
+                let now = monotonic_coarse_now();
+                Instant::saturating_elapsed_duration_coarse(now, t)
             }
         }
     }
 
     pub fn elapsed_secs(&self) -> f64 {
```
Connor1996, line 306: remove non-saturating functions?

Connor1996 (commented):
rest LGTM

components/tikv_util/src/time.rs
Connor1996: or just handle the jump back silently instead of adding a new function.

components/tikv_util/src/time.rs
```diff
@@ -321,12 +363,28 @@ impl Instant {
         }
     }
 
+    pub(crate) fn saturating_elapsed_duration(later: Timespec, earlier: Timespec) -> Duration {
+        if later >= earlier {
+            (later - earlier).to_std().unwrap()
+        } else {
+            debug!(
```
gengliqi, line 370: Since this case should happen rarely, the log here should be an error or warn?

gengliqi (commented):
Rest LGTM. BTW, I think the test parts don't need to be changed because these tests don't run in the production environment.

components/tikv_util/src/time.rs
tabokie: I think I'll mark them as `[test]` to avoid misuse, but keep the interface here.

components/tikv_util/src/time.rs
tabokie: Good idea.

tabokie:
> Rest LGTM. BTW, I think the test parts don't need to be changed because these tests don't run in the production environment.

@gengliqi I'll revert uses of `saturating_duration_since` from tests, which can indeed hide real errors.

Connor1996 (approved):
LGTM

PR #10553, opened 2021-07-12
copr: fix json_unquote bug when argument is like `""a""`

wzru:

Problem Summary: port https://github.com/pingcap/tidb/pull/25875 to TiKV. 

json_unquote() bug when input is like `""a""`
expected output:
```bash
mysql> SELECT  json_unquote('""a""');
ERROR 3141 (22032): Invalid JSON text in argument 1 to function json_unquote: "The document root must not be followed by other values." at position 2.
```
wrong output:
```
mysql> SELECT  json_unquote('""a""');
+-----------------------+
| json_unquote('""a""') |
+-----------------------+
| "a"                   |
+-----------------------+
1 row in set (0.01 sec)
```

What's Changed:

### Related changes

- https://github.com/pingcap/tidb/pull/25875

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/tidb_query_expr/src/impl_json.rs
```diff
@@ -9,6 +9,9 @@ use tidb_query_common::Result;
 use tidb_query_datatype::codec::data_type::*;
 use tidb_query_datatype::codec::mysql::json::*;
 
+use serde::de::*;
+use serde_json::*;
```
skyzh, line 13: Why do we need to use all items from `serde_json`? Seems that `from_str` and `IgnoredAny` are enough.

skyzh (commented):
Mostly LGTM.

I've seen PR title like `fix json_unquote` dozens of times. I would suggest using a more meaningful title, and revisit the body of your PR, so that people would easily know what have done.

By the way, if we need to enable `json_unquote` pushdown for TiKV coprocessor, it's better to enable pushdown and run integration test. I would help you trigger it later.

components/tidb_query_expr/src/impl_json.rs
sticnarf, line 13: Please do not import all if only a few functions in it are used.

components/tidb_query_expr/src/impl_json.rs
```diff
@@ -206,6 +209,11 @@ fn quote(bytes: BytesRef) -> Result<Option<Bytes>> {
 #[inline]
 fn json_unquote(arg: BytesRef) -> Result<Option<Bytes>> {
     let tmp_str = std::str::from_utf8(arg)?;
+    let first_char = tmp_str.chars().next();
+    let last_char = tmp_str.chars().last();
+    if tmp_str.len() >= 2 && first_char == Some('"') && last_char == Some('"') {
+        let _: IgnoredAny = from_str(&tmp_str)?;
```
sticnarf, line 215: And it is a good idea to use `serde_json::from_str` to reduce ambiguity.

skyzh (approved):

sticnarf (approved):

PR #10555, opened 2021-07-12
reject raft messages if they take too much memory

hicqu:

Under heavy write load, received raft messages and unstable raft entries can take much memory to cause OOM finally.

Reject raft messages if they take too much memory.

### Related changes

### Check List 

Tests 

- Integration test
- Manual test
图片
With the patch, if a store intstance stops and restarts, received raft messages size is limited and the instance won't be killed by OOM-killer.

### Release note 

```release-note
None
```


src/server/service/kv.rs
```diff
@@ -617,20 +619,36 @@ impl<T: RaftStoreRouter<E::Local> + 'static, E: Engine, L: LockManager> Tikv for
...
                         my_store_id: store_id,
-                    }))
-                } else {
-                    let ret = ch.send_raft_msg(msg).map_err(Error::from);
-                    future::ready(ret)
+                    }));
                 }
-            });
+                while needs_delay_raft_append() {
+                    let to = std::time::Instant::now() + std::time::Duration::from_millis(100);
```
rleungx, line 639: Will it be too long? Or should it be dynamically changed?

src/server/service/kv.rs
```diff
@@ -1940,6 +1966,22 @@ fn raftstore_error_to_region_error(e: RaftStoreError, region_id: u64) -> RegionE
     e.into()
 }
 
+fn needs_delay_raft_append() -> bool {
```
rleungx, line 1969:
```suggestion
fn needs_reject_raft_append() -> bool {
```

rleungx (commented):
prefer to find a better name for `test_memory_full.rs`, the rest LGTM.

rleungx (approved):

src/server/service/kv.rs
```diff
@@ -1940,6 +1966,22 @@ fn raftstore_error_to_region_error(e: RaftStoreError, region_id: u64) -> RegionE
     e.into()
 }
 
+fn needs_reject_raft_append() -> bool {
+    fail_point!("needs_reject_raft_append", |_| true);
+    let mut usage = 0;
+    if memory_usage_reaches_high_water(&mut usage) {
+        let raft_msg_usage = (MEMTRACE_RAFT_ENTRIES.sum() + MEMTRACE_RAFT_MESSAGES.sum()) as u64;
+        if raft_msg_usage > usage / 5 {
```
5kbpers, line 1974: Should we add a config item for it?

hicqu: Good idea.

5kbpers (approved):

PR #10557, opened 2021-07-12
pd_client: use stream RPC for TSO requests

sticnarf:

The mechanism that TiKV gets timestamps from PD was not well implemented. Every time it needs to get a timestamp, it creates a stream RPC to PD and just send a request.

After async commit is implemented on TiKV, when a region becomes a new leader, TiKV needs to get a timestamp from PD before the new region leader starts to serve writing requests. So if a large number of regions become leaders in a short time, it is really inefficient to get so many timestamp one by one stream RPC.

A user with millions of regions tried to run `tiup cluster restart` and many regions cannot work after quite some while because of `MaxTimestampNotSynced` errors. This is probably because the problem described above. It is also a heavy burden for PD to handle millions of TSO requests in a short time without any batch mechanism.

This PR re-implements the `get_tso` method. The code is mostly ported from https://github.com/tikv/client-rust/blob/4404c7e1f0c2128d1947e0c8a27f4771c9f9221f/tikv-client-pd/src/timestamp.rs plus some bug fixes. Of course the bug fixes will be submitted to the client-rust later.

After TiKV connects to PD, it creates only one stream TSO RPC. A `tso-worker` thread is created to handle requests and bundle them into batch requests to PD. 

### Related changes

- Need to cherry-pick to the release branch (?)

### Check List 

Tests 

### Release note 

```release-note
None
```


components/pd_client/src/tso.rs
```diff
@@ -0,0 +1,236 @@
...
+        .ok_or_else(|| -> Error { box_err!("No timestamp in TsoResponse") })?;
+
+    let mut offset = resp.count;
+    if let Some(RequestGroup {
+        tso_request,
+        requests,
+    }) = pending_requests.pop_front()
+    {
+        if tso_request.count != offset {
+            return Err(box_err!("PD gives expected number of timestamps"));
```
JmPotato, line 222: The error message seems not right.

components/pd_client/src/tso.rs
```diff
@@ -0,0 +1,236 @@
...
+    let tail_ts = resp
+        .timestamp
+        .as_ref()
+        .ok_or_else(|| -> Error { box_err!("No timestamp in TsoResponse") })?;
+
+    let mut offset = resp.count;
+    if let Some(RequestGroup {
+        tso_request,
+        requests,
+    }) = pending_requests.pop_front()
```
JmPotato, line 219: Can we guarantee the order of `resp` matches the `pending_requests`?

sticnarf: I believe PD handles requests in a single stream RPC serially. So the order should be the same.

JmPotato (approved):

components/pd_client/src/tso.rs
```diff
@@ -0,0 +1,236 @@
...
+        Ok(())
+    };
+
+    let receive_and_handle_responses = async move {
+        while let Some(Ok(resp)) = rpc_receiver.next().await {
+            let mut pending_requests = pending_requests.borrow_mut();
+
+            // Wake up the sending future blocked by too many pending requests as we are consuming
+            // some of them here.
+            if pending_requests.len() == MAX_PENDING_COUNT {
```
Connor1996, line 134:
```suggestion
            if pending_requests.len() >= MAX_PENDING_COUNT {
```

components/pd_client/src/tso.rs
```diff
@@ -0,0 +1,236 @@
...
+
+async fn run_tso(
+    cluster_id: u64,
+    mut rpc_sender: impl Sink<(TsoRequest, WriteFlags), Error = Error> + Unpin,
+    mut rpc_receiver: impl Stream<Item = Result<TsoResponse>> + Unpin,
+    mut request_rx: mpsc::Receiver<TimestampRequest>,
+    close_tx: watch::Sender<()>,
+) {
+    // The `TimestampRequest`s which are waiting for the responses from the PD server
+    let pending_requests = Rc::new(RefCell::new(VecDeque::with_capacity(MAX_PENDING_COUNT)));
```
Connor1996, line 107: do we have a metrics for this?

sticnarf: Not yet. Nice to have, let me add it.

MyonKeminta (approved):

components/pd_client/src/tso.rs
sticnarf: Now I add a gauge and a grafana panel for it. (In normal cases, the value is always 0...)

sticnarf: @Connor1996 PTAL again when you're free.

Connor1996 (approved):
LGTM

Issue #10537, opened 2021-07-09
Tracking issue of TiKV disk full


tier-cap:
## Development Task

This issue describes a new feature that makes TiKV can avoid panicking from disk full failure, and provide space reclamation support at the same time. The spec:
* TiKV shouldn't panic even if its disk usage is almost full
* Even if most TiKV instances in a cluster are almost full, the cluster should be still readable.
* Even if most TiKV instances in a cluste are almost full, users should still can release space by `drop table` or `truncate table`.

## Task list

* [x] #10597
* [x] https://github.com/pingcap/tidb/issues/27665
* [x] https://github.com/tikv/client-go/issues/286
* [x] https://github.com/pingcap/kvproto/issues/803
## Bug List
* [x] https://github.com/tikv/tikv/issues/10960
* [x] https://github.com/tikv/tikv/issues/10959
* [x] https://github.com/tikv/tikv/issues/10926
* [x] https://github.com/tikv/tikv/issues/10925
* [x] https://github.com/tikv/tikv/issues/10895
* [x] https://github.com/tikv/tikv/issues/10894
* [x] https://github.com/tikv/tikv/issues/10868
* [x] https://github.com/tikv/tikv/issues/10840
* [x] https://github.com/tikv/tikv/issues/10839
* [x] https://github.com/tikv/tikv/issues/10819
* [x] https://github.com/tikv/tikv/issues/10801

## Left & Low Priority
* [ ] https://github.com/tikv/tikv/issues/11010
* [ ] https://github.com/tikv/tikv/issues/10665
* [ ] https://github.com/tikv/tikv/issues/11109
* [ ] https://github.com/tikv/tikv/issues/11110
* [ ] https://github.com/tikv/tikv/issues/11232
* [ ] https://github.com/tikv/tikv/issues/11236
* [ ] https://github.com/pingcap/tidb/issues/29557

tisonkun:
Hi @tier-cap ! Thanks for creating this tracking issue. You may create issue for each subtask dedicatedly, see also https://github.com/pingcap/tidb/issues/26022 . The rationale of tracing issues instead of pull requests can be found [here](https://tisonkun.github.io/ziggurat/f379bf788b20/#Issus-or-Pull-Request).

For example, you may create an issue that gets closed once #10562 resolved.

tisonkun:
we are tracking issue by issues, why do you update the description redundant to show PRs?

DanielZhangQD:
[...]

PR #10562, opened 2021-07-13
*: solve cloud tikv disk full problem

tier-cap:
Issue Number：#10537
When the disk is full, no mather one or two or the majority or all, what we expected will be:
1. TiKV not crash directly.
2. Support restoring storage through drop table/delete from table, expansion is also allowed.
3. Provide slightly better reading ability.
1. Introduce a two-level threshold defense mechanism
2. Different flow control algorithms are used under each threshold to provide different business support capabilities and spatial recovery capabilities

Check List
Test

UserDoc

### Release note 

```release-note
None
```


components/raftstore/src/store/peer.rs
```diff
@@ -3118,8 +3144,6 @@ where
         if self.is_applying_snapshot()
             || self.has_pending_snapshot()
             || msg.get_from() != self.leader_id()
-            // For followers whose disk is full.
-            || disk::disk_full_precheck(ctx.store_id()) || ctx.is_disk_full
```
hicqu, line 3122: I think this strategy is better:
1. don't transfer leadership to a disk almost full target, if the current leader is far from disk full;
2. transfer leadership to a disk almost full target, if the current leader is also almost full or already full
3. never transfer leadership to a target which is already disk full.

components/raftstore/src/store/peer.rs
```diff
@@ -1072,9 +1073,10 @@ where
         T: Transport,
         I: IntoIterator<Item = eraftpb::Message>,
     {
+        let disk_status = disk::get_disk_status();
         for msg in msgs {
             let msg_type = msg.get_msg_type();
-            let i = self.send_raft_message(msg, trans) as usize;
+            let i = self.send_raft_message(msg, trans, disk_status) as usize;
```
hicqu, line 1079: It's wrong. It should be the `DiskFullOpt` of the carried commands instead of the peer's disk status. And please consider this situation: proposing a drop table operation, and then proposing a large normal entry on the same region. `RaftMessage` should contains the information about `Vec<(log_index, DiskFullOpt)>`.

components/raftstore/src/errors.rs
```diff
@@ -55,6 +55,9 @@ pub enum Error {
     #[error("peer is not leader for region {0}, leader may {1:?}")]
     NotLeader(u64, Option<metapb::Peer>),
 
+    #[error("{0:?}")]
```
hicqu, line 58: Why not also print field the another field?

components/raftstore/src/router.rs
```diff
@@ -48,16 +49,26 @@ where
...
-        send_command_impl::<EK, _>(self, req, cb, None)
+        send_command_impl::<EK, _>(self, req, cb, None, DiskFullOpt::default())
     }
 
     fn send_command_with_deadline(
         &self,
         req: RaftCmdRequest,
         cb: Callback<EK::Snapshot>,
         deadline: Deadline,
+        disk_full_opt: DiskFullOpt,
```
hicqu, line 60: How about add a new `Options` to carry `deadline` and `disk_full_opt`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -546,6 +546,7 @@ where
                     }
                 }
                 PeerMsg::RaftCommand(cmd) => {
+                    self.ctx.disk_full_opt = cmd.disk_full_opt;
```
hicqu, line 549: `disk_full_opt` works on every commnad instead of a batch of commands. So it shouldn't be put into `PollContext`.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1267,29 +1268,34 @@ where
...
             }
 
-            if flag {
+            if !allowed {
                 debug!(
                     "skip {:?} because of disk full", msg_type;
                     "region_id" => self.region_id(), "peer_id" => self.fsm.peer_id()
                 );
-                return Err(Error::Timeout("disk full".to_owned()));
+                return Err(Error::DiskFull(
```
hicqu, line 1295: So the message is dropped. Please update the metrics `STORE_RAFT_DROPPED_MESSAGE_COUNTER`.

components/raftstore/src/store/fsm/peer.rs
tier-cap: Done!

components/raftstore/src/store/peer.rs
```diff
@@ -2336,11 +2338,32 @@ where
             }
             Ok(RequestPolicy::ReadIndex) => return self.read_index(ctx, req, err_resp, cb),
             Ok(RequestPolicy::ProposeNormal) => {
-                let store_id = ctx.store_id();
-                if disk::disk_full_precheck(store_id) || ctx.is_disk_full {
-                    Err(Error::Timeout("disk full".to_owned()))
-                } else {
+                let mut allowed = true;
+                let mut err_msg = String::from("propose failed: disk ");
```
hicqu, line 2342: It can be allocated when necessary. By the way, the error message seems not complete.

hicqu: It's not necessary to make 2 different error message. Just `propose failed: disk full` is OK for them.

components/raftstore/src/store/peer.rs
```diff
@@ -2336,11 +2338,32 @@ where
...
+                match disk_full_opt {
+                    DiskFullOpt::AllowedOnAlreadyFull => allowed = true,
+                    DiskFullOpt::AllowedOnAlmostFull => {
+                        if disk::is_disk_already_full(ctx.disk_status, store_id) {
+                            allowed = false;
+                            err_msg.push_str("already full");
+                        }
+                    }
+                    DiskFullOpt::NotAllowedOnFull => {
+                        if disk::is_disk_already_full(ctx.disk_status, store_id) {
```
hicqu, line 2353: It's better to use `fn disk::usage_status() -> UsageStatus`.

components/server/src/server.rs
```diff
@@ -412,20 +412,17 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
         }
+        let mut reserve_space = self.config.storage.reserve_space.0;
+        if self.config.storage.reserve_space.0 != 0 {
+            reserve_space = cmp::max(
+                (capacity as f64 * 0.05) as u64,
+                self.config.storage.reserve_space.0,
+            );
+        }
+        disk::set_disk_reserved_space(reserve_space);
         //TODO after disk full readonly impl, such file should be removed.
```
hicqu, line 423: The log can be removed.

components/test_raftstore/src/cluster.rs
```diff
@@ -1483,6 +1490,19 @@ impl<T: Simulator> Cluster<T> {
     pub fn partition(&self, s1: Vec<u64>, s2: Vec<u64>) {
         self.add_send_filter(PartitionFilterFactory::new(s1, s2));
     }
+
+    // set operation allowed level when tikv disk full happens.
+    pub fn set_disk_full_opt(&mut self, level: DiskFullOpt) {
```
hicqu, line 1495: I think it's wrong. `DiskFullOpt` should be carried in `Context`. It shouldn't be a global variable for `Cluster`.

components/test_raftstore/src/util.rs
```diff
@@ -908,6 +908,37 @@ pub fn must_kv_prewrite_with(
     );
 }
 
+pub fn try_kv_prewrite_with(
+    client: &TikvClient,
+    ctx: Context,
```
hicqu, line 913: Please set `DiskFullOpt` in `Context`.

components/raftstore/src/router.rs
tier-cap: Done!

components/raftstore/src/store/peer.rs
tier-cap: err_msg left will be filled in under each threshold.

components/tikv_kv/src/lib.rs
```diff
@@ -111,19 +111,48 @@ pub struct WriteData {
     pub modifies: Vec<Modify>,
     pub extra: TxnExtra,
     pub deadline: Option<Deadline>,
+    pub disk_full_opt: DiskFullOpt,
 }
 
 impl WriteData {
     pub fn new(modifies: Vec<Modify>, extra: TxnExtra) -> Self {
+        Self::new_ext(modifies, extra, None, DiskFullOpt::NotAllowedOnFull)
```
hicqu, line 119: These changes about constructors of `WriteData` are not necessary. Just keep the old code, and add a new method `set_disk_full_opt` is fine enough.

components/tikv_util/src/sys/disk.rs
```diff
@@ -1,22 +1,59 @@
...
-static DISK_FULL: AtomicBool = AtomicBool::new(false);
+use std::sync::atomic::{AtomicI32, AtomicU64, Ordering};
 
-pub fn set_disk_full() {
-    DISK_FULL.store(true, Ordering::Release);
+static DISK_RESERVED_SPACE: AtomicU64 = AtomicU64::new(0);
+
+#[derive(Copy, Clone)]
+pub enum DiskStatus {
+    DiskNormal = 0,
```
hicqu, line 9: `Disk` can removed for them.

components/raftstore/src/store/peer.rs
tier-cap: Better to let fail_point only appear in disk.rs?

src/storage/txn/commands/mod.rs
```diff
@@ -89,6 +89,28 @@ pub enum Command {
     RawAtomicStore(RawAtomicStore),
 }
 
+/// Store Transaction scheduler commands type used by raftstore
+pub enum CommandType {
```
hicqu, line 93: Seems it can be removed?

components/server/src/server.rs
tier-cap: done

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -81,7 +94,29 @@ fn test_disk_full_allowed_leader_behaviors() {
     cluster.pd_client.must_add_peer(1, new_peer(3, 3));
     must_get_equal(&cluster.get_engine(3), b"k1", b"v1");
 
-    fail::remove("disk_full_peer_1");
+    // Test operations with special flag is allowed.
```
hicqu, line 97: It's better to add a new case `test_transactions_on_full_leader` to test changes about transaction layer.

components/test_raftstore/src/cluster.rs
tier-cap: Maybe this logic should be removed.

components/test_raftstore/src/util.rs
tier-cap: This option should be set in test case. When not set, it will be the default value.

tier-cap: Done

components/raftstore/src/store/fsm/store.rs
```diff
@@ -391,7 +391,7 @@ where
     pub perf_context: EK::PerfContext,
     pub tick_batch: Vec<PeerTickBatch>,
     pub node_start_time: Option<TiInstant>,
-    pub is_disk_full: bool,
+    pub disk_status: disk::DiskUsageStatus,
```
hicqu, line 394: It's better: s/disk_status/disk_usage.

components/server/src/server.rs
```diff
@@ -1051,12 +1047,14 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
         let snap_mgr = self.snap_mgr.clone().unwrap();
-        let disk_reserved = self.config.storage.reserve_space.0;
-        if disk_reserved == 0 {
+        let reserve_space = disk::get_disk_reserved_space();
+        if reserve_space == 0 {
             info!("disk space checker not enabled");
             return;
         }
-        //TODO wal size ignore?
+        //threshold not exposed to users.
```
hicqu, line 1055: This comment is stale. Let's remove it.

components/raftstore/src/store/peer.rs
hicqu: We can move failpoints into `get_disk_status`.

components/raftstore/src/router.rs
```diff
@@ -47,17 +48,13 @@ where
     }
 
     /// Sends RaftCmdRequest to local store.
-    fn send_command(&self, req: RaftCmdRequest, cb: Callback<EK::Snapshot>) -> RaftStoreResult<()> {
```
hicqu, line 50: Why not keep it? So most places are necessary to change. We just need to change `send_command_with_deadline` to `send_command_with_options`.

components/raftstore/src/router.rs
```diff
@@ -47,17 +48,13 @@ where
...
-        send_command_impl::<EK, _>(self, req, cb, None)
-    }
-
-    fn send_command_with_deadline(
+    fn send_command(
         &self,
         req: RaftCmdRequest,
         cb: Callback<EK::Snapshot>,
-        deadline: Deadline,
+        extra_opt: RaftCmdExtraOpt,
```
hicqu, line 55: Should be `RaftCmdExtraOpts` or `RaftCmdOptions`. There are several configurable options in it.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1267,29 +1275,34 @@ where
         );
 
         let msg_type = msg.get_message().get_msg_type();
-        let store_id = self.ctx.store_id();
 
-        if disk::disk_full_precheck(store_id) || self.ctx.is_disk_full {
-            let mut flag = false;
+        if disk::is_disk_almost_full(self.ctx.disk_status, self.ctx.store.get_id())
```
hicqu, line 1279: is_disk_almost_full and is_disk_already_full can be removed. We can match `disk_status` directly.

components/raftstore/src/router.rs
tier-cap: Done!

components/server/src/server.rs
tier-cap: Done

components/raftstore/src/store/peer.rs
tier-cap: Done

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -81,7 +94,38 @@ fn test_disk_full_allowed_leader_behaviors() {
     cluster.pd_client.must_add_peer(1, new_peer(3, 3));
     must_get_equal(&cluster.get_engine(3), b"k1", b"v1");
 
-    fail::remove("disk_full_peer_1");
+    // Test Txn commit/rollback is also allowed.
```
hicqu, line 97: Seems rollback is not covered. And please extract test code about transaction to another test case. This case only cares raft layrer instead of transaction layer.

hicqu: And, we also need a cases that prewrite reuqests with `DislFullOpt::Allowed` should be able to be proposed.

src/storage/txn/commands/mod.rs
tier-cap: Right,Done!

components/raftstore/src/router.rs
```diff
@@ -101,15 +98,15 @@ fn send_command_impl<EK, PR>(
     router: &PR,
     req: RaftCmdRequest,
     cb: Callback<EK::Snapshot>,
-    deadline: Option<Deadline>,
+    extra_opt: RaftCmdExtraOpts,
```
hicqu, line 101: should s/extra_opt/extra_opts.

components/server/src/server.rs
```diff
@@ -1092,18 +1090,35 @@ impl<ER: RaftEngine> TiKVServer<ER> {
 
                 let mut available = capacity.checked_sub(used_size).unwrap_or_default();
                 available = cmp::min(available, disk_stats.available_space());
-                if available <= disk_reserved {
+                let disk_status = disk::get_disk_status(0);
```
hicqu, line 1093: How about `prev_disk_status` and `curr_disk_status`, and change log to `warn!("disk usage {} -> {}, ...")`?

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -63,15 +76,15 @@ fn test_disk_full_allowed_leader_behaviors() {
     must_get_equal(&cluster.get_engine(3), b"k1", b"v1");
 
     cluster.must_transfer_leader(1, new_peer(1, 1));
-    fail::cfg("disk_full_peer_1", "return").unwrap();
+    fail::cfg("disk_almost_full_peer_1", "return").unwrap();
 
-    // Test transfer leader should be allowed.
+    // // Test transfer leader should be allowed.
```
hicqu, line 81: Redudant `//`.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -47,11 +49,22 @@ fn test_disk_full_unallowed_leader_behaviors() {
     let new_last_index = cluster.raft_local_state(1, 1).last_index;
     assert_eq!(old_last_index, new_last_index);
 
-    fail::remove("disk_full_peer_1");
+    // Test Txn Prewrite is not allowed.
```
hicqu, line 52: It can be moved into `test_disk_almost_full_allowed_leader_txn_behaviors`. And we can rename `test_disk_almost_full_allowed_leader_txn_behaviors` to `test_disk_almost_full_leader_txn_behaviors`.

tests/failpoints/cases/test_disk_full.rs
```diff
@@ -81,7 +94,97 @@ fn test_disk_full_allowed_leader_behaviors() {
...
+    fail::cfg("disk_almost_full_peer_1", "return").unwrap();
+
+    // Test Txn commit/rollback is also allowed.
+    fail::remove("disk_almost_full_peer_1");
+    let pd_client = cluster.pd_client.clone();
+    let lead_client = PeerClient::new(&cluster, 1, new_peer(1, 1));
+    let prewrite_ts = get_tso(&pd_client);
+
+    lead_client.must_kv_prewrite(
+        vec![new_mutation(Op::Put, &b"k4"[..], &b"v4"[..])],
```
hicqu, line 122: `[..]` is not necessary.

tests/failpoints/cases/test_disk_full.rs
tier-cap: Right, these are already covered, like "Test Txn Prewrite is allowed with AllowedOnAlmostFull flag."

components/raftstore/src/router.rs
tier-cap: this is an one struct obj, should it be extra_opt?

tests/failpoints/cases/test_disk_full.rs
tier-cap: Done!

tests/failpoints/cases/test_disk_full.rs
tier-cap: Done!

components/tikv_kv/src/lib.rs
```diff
@@ -119,12 +120,25 @@ impl WriteData {
...
             deadline: None,
+            disk_full_opt: DiskFullOpt::NotAllowedOnFull,
         }
     }
 
     pub fn from_modifies(modifies: Vec<Modify>) -> Self {
         Self::new(modifies, TxnExtra::default())
     }
+
+    pub fn set_allowed_on_almost_full(&mut self) {
```
hicqu, line 131: Just `set_disk_full_opt` is fine. This function name is bad, because it doesn't show **what** fulll. Should point out **disk** if you would like to keep it.

tests/failpoints/cases/test_disk_full.rs
tier-cap: Right,Done

src/storage/txn/scheduler.rs
```diff
@@ -726,6 +726,17 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
                 response_policy,
             }) => {
                 SCHED_STAGE_COUNTER_VEC.get(tag).write.inc();
+                match ctx.get_disk_full_opt() {
+                    DiskFullOpt::AllowedOnAlreadyFull => {
+                        to_be_write.disk_full_opt = DiskFullOpt::AllowedOnAlreadyFull
+                    }
+                    DiskFullOpt::AllowedOnAlmostFull => {
+                        if to_be_write.disk_full_opt != DiskFullOpt::AllowedOnAlreadyFull {
```
hicqu, line 734: How can it happen? If it can't happen theoretically, please use `debug_assert` or `assert`.

hicqu:
Rest LGTM. Good job!

src/storage/txn/scheduler.rs
tier-cap: Like Delete from table? TiDB set it to AllowedOnAlmostFull, but TiKV trait it as normal,.Then here we need change it.

hicqu: Got. Please comment it.

tier-cap: Done

components/tikv_kv/src/lib.rs
tier-cap: Done

hicqu (approved):

5kbpers (approved):

PR #10572, opened 2021-07-14
Avoid duration calculation panic when clock jumps back (#10544)

tabokie:
Cherry-pick #10544

### Release note 

```release-note
Fix duration calculation panics on certain platforms
```

### DIFF
--- a/components/tikv_util/src/time.rs
+++ b/components/tikv_util/src/time.rs
@@ -285,7 +285,7 @@ impl Instant {
             }
             Instant::MonotonicCoarse(t) => {
                 let now = monotonic_coarse_now();
-                Instant::saturating_elapsed_duration_coarse(now, t)
+                Instant::elapsed_duration_coarse(now, t)
             }
         }
     }
@@ -304,7 +304,7 @@ impl Instant {
                 Instant::elapsed_duration(later, earlier)
             }
             (Instant::MonotonicCoarse(later), Instant::MonotonicCoarse(earlier)) => {
-                Instant::saturating_elapsed_duration_coarse(later, earlier)
+                Instant::elapsed_duration_coarse(later, earlier)
             }
             _ => {
                 panic!("duration between different types of Instants");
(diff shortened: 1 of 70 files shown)


yiwu-arbug (approved):

components/tikv_util/src/time.rs
```diff
@@ -405,8 +454,10 @@ impl SubAssign<Duration> for Instant {
 impl Sub<Instant> for Instant {
     type Output = Duration;
 
+    // TODO: For safety in production code, `sub` actually does saturating_sub.
+    // We should remove this operator from public scope.
```
BusyJay, line 458: What places is using `sub`? Is it safe to do `saturating_duration_since` for all of them?

components/tikv_util/src/time.rs
```diff
@@ -317,12 +350,28 @@ impl Instant {
...
+        }
+    }
+
     // It is different from `elapsed_duration`, the resolution here is millisecond.
     // The processors in an SMP system do not start all at exactly the same time
     // and therefore the timer registers are typically running at an offset.
     // Use millisecond resolution for ignoring the error.
     // See more: https://linux.die.net/man/2/clock_gettime
-    fn elapsed_duration_coarse(later: Timespec, earlier: Timespec) -> Duration {
+    pub(crate) fn saturating_elapsed_duration_coarse(
```
BusyJay, line 371: I don't think the rename makes sense as there is no accurate time when using coarse timer, so there is no such "non-saturating` version at all.

BusyJay (commented):
I think we have a daemon thread to monitor jumping back of time. I can't say whether to panic is good or not as when monotonic timer is dumping back, there are many places are also affected including unify thread pool scheduling or hot fsm scheduling.

5kbpers (approved):

components/tikv_util/src/time.rs
tabokie: make sense.

components/tikv_util/src/time.rs
tabokie: I replaced them in the last commit, PTAL.

tabokie:
> I think we have a daemon thread to monitor jumping back of time. I can't say whether to panic is good or not as when monotonic timer is dumping back, there are many places are also affected including unify thread pool scheduling or hot fsm scheduling.

But I presume this won't affect service ability? Just quality. In this patch we will print error log to alert the admin to investigate if this can turn into a real problem.

components/tikv_util/src/time.rs
BusyJay: Seems the trait implementation can be deleted.

tabokie: We are using it in "async_speed_limit::Limiter", I can make the op private though.

Issue #10539, opened 2021-07-09
Introduce slow node detecting mechanism


5kbpers:
Tracking issue for slow node detection

# Background 
As the scale of the tikv cluster becomes larger, the frequency of failures that slow down a single tikv (known as fail-slow) due to various reasons will become higher (according to the 1.02% annual failure rate[1], as a rough estimate, if the total number of nodes running tikv for all customers is 5k, then we will encounter this problem almost every week). Therefore, we need a more timely mechanism to discover and automatically cope with those problems as much as possible to improve the stability of TiKV clusters.

# Task
- [x] Collect how tikv performs after some components become slow
- [x] Design and implement a mechanism to measure the performance of a tikv
- [x] Introduce a mechanism to mitigate fail-slow

tisonkun:
@5kbpers on the [monthly meeting](https://internals.tidb.io/t/topic/363/2) @BusyJay said this feature is released in 5.2, but it seems there is still

> Introduce a mechanism to mitigate fail-slow

under development. Is there anything PR actually implement it?

5kbpers:
> @5kbpers on the [monthly meeting](https://internals.tidb.io/t/topic/363/2) @BusyJay said this feature is released in 5.2, but it seems there is still
> 
> > Introduce a mechanism to mitigate fail-slow
> 
> under development. Is there anything PR actually implement it?

Yes, you can check those PRs refer to this issue.

5kbpers:
Close since this work has done.

PR #10589, opened 2021-07-19
raftstore/pd: replace FutureWorker with Worker for pd worker

5kbpers:

Problem Summary:

For https://github.com/tikv/tikv/issues/10539 we need to implement a tick for pd worker to check the status of tikv. But `FutureWorker` doesn't support tick.

Replace `FutureWorker` with `Worker` for pd worker.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
N/A
```


CodingPoeta:
lalaland

CodingPoeta:
hey

CodingPoeta:
hahaha

CodingPoeta:
lalaland

CodingPoeta:
test

CodingPoeta:
life is a drug

CodingPoeta:
life is a drug

CodingPoeta:
truetrue

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1408,9 +1409,6 @@ impl<EK: KvEngine, ER: RaftEngine> RaftBatchSystem<EK, ER> {
         MEMTRACE_RAFT_ROUTER_ALIVE.trace(TraceEvent::Reset(0));
         MEMTRACE_RAFT_ROUTER_LEAK.trace(TraceEvent::Reset(0));
 
-        if let Some(h) = handle {
```
sticnarf, line 1411: Does this change the behavior of `shutdown`?

BusyJay (commented):
As far as I know, Lazy worker shares a thread pool with other workers, and we have seem an overloaded worker cause starvation on other workers and lead to deadlock in some case. Is it safe for PD worker?

5kbpers:
> As far as I know, Lazy worker shares a thread pool with other workers, and we have seem an overloaded worker cause starvation on other workers and lead to deadlock in some case. Is it safe for PD worker?

It will create a new thread pool, lazy worker here is just for creating the runner laziliy.

components/raftstore/src/store/fsm/store.rs
5kbpers: It does change the shutdown of pd worker from asynchronously to synchronously, but I think this doesn't have other side effects.

BusyJay (approved):

Little-Wallace (approved):
LGTM. Maybe you shall run it in a real TiKV cluster to avoid some problems we do not find.

5kbpers:
> LGTM. Maybe you shall run it in a real TiKV cluster to avoid some problems we do not find.

Thanks! Have been running for a few days.

PR #10591, opened 2021-07-19
raftstore: Add hot fsm log

Connor1996:

Issue Number: close #xxx 
When encounter a high apply wait duration, it may caused by some hot regions with lots of requests. But it's hard to find which region it is, so add a log here.

Problem Summary:

Proposal: [xxx](url) 

What's Changed:
Add a log 

### Check List 

- No code

Side effects

- None

### Release note 

```release-note
None
```

### DIFF
--- a/components/batch-system/src/batch.rs
+++ b/components/batch-system/src/batch.rs
@@ -17,7 +17,7 @@ use std::sync::Arc;
 use std::thread::{self, JoinHandle};
 use std::time::Duration;
 use tikv_util::mpsc;
-use tikv_util::time::Instant;
+use tikv_util::time::{duration_to_sec, Instant};
 use tikv_util::{debug, error, info, safe_panic, thd_name, warn};
 
 /// A unify type for FSMs so that they can be sent to channel easily.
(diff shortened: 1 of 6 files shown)


components/batch-system/src/batch.rs
```diff
@@ -308,12 +308,19 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
...
                 let len = self.handler.handle_normal(p);
+                let region_id = p.region_id();
                 if p.is_stopped() {
                     reschedule_fsms.push((i, ReschedulePolicy::Remove));
                 } else if p.get_priority() != self.handler.get_priority() {
                     reschedule_fsms.push((i, ReschedulePolicy::Schedule));
                 } else {
-                    if batch.timers[i].saturating_elapsed() >= self.reschedule_duration {
+                    let d = batch.timers[i].saturating_elapsed();
+                    if d >= self.reschedule_duration {
```
Little-Wallace, line 318: Will this log be too frequently?

components/batch-system/src/fsm.rs
```diff
@@ -55,6 +55,10 @@ pub trait Fsm {
     fn get_priority(&self) -> Priority {
         Priority::Normal
     }
+
+    fn region_id(&self) -> u64 {
```
5kbpers, line 59: Prefer `fsm_id` or `id`

PR #10592, opened 2021-07-19
raftstore: separate read write ready

### LINKED ISSUE #10475: Separate read/write ready
Let's define read ready as ready that won't modify write batch, write ready as ready that modifies write batch. When read/write ready are fetched into one batch, read ready will wait for write ready to finish.

Considering following situation, a full ranges scan will notify all idle regions, which will just propose a read index request to extend read lease. There is no need to do any writes. Internal network latency is usually hundreds of us, but write latency can vary from several to hundreds of milliseconds. So leader can take longer than expected to process a read and cause high tail latency.

We should separate read/write readiness. #10289 solve the problem by using a separate thread to write raft logs and states. But the change may not be back ported to 4.0/5.0 as changes are large and complicated. For old codebase, we can still optimize the tail latency by adding a new method `light_end` to batch system. `light_end` will be used to release fsms that produce read only readiness. So the procedure of handling ready becomes:
```
begin
fetch fsms
handle normal
    fetch ready
light end
    post_ready for read ready
release read fsms
end
   write disk
   post_ready for write ready
release write fsms
```

I wrote a prototype and the result is promising:

image

The left part is with optimization patch, the right part is v5.0.2. My patch can be see [here](https://github.com/BusyJay/tikv/commit/4b65f35f64f94c500ddf36336871c258ef447f8c)

Do nothing for other branch and wait for #10289.

But the change in my prototype is quite small, I think it's better accept the change and optimize for all versions.

BusyJay:

Issue Number: close #10475

Problem Summary:

Read ready can be processed early to reduce read latency.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
- separate read write ready to reduce read latency
```


gengliqi (approved):

components/raftstore/src/store/fsm/store.rs
```diff
@@ -660,8 +664,27 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> RaftPoller<EK, ER, T> {
...
+            let mut readonly_ready_res = mem::take(&mut self.poll_ctx.readonly_ready_res);
+            for ready in readonly_ready_res.drain(..) {
+                to_skip_end.push(ready.batch_offset);
+                PeerFsmDelegate::new(
+                    peers[ready.batch_offset].as_mut().unwrap(),
+                    &mut self.poll_ctx,
+                )
+                .post_raft_ready_append(ready);
+            }
+            self.poll_ctx.readonly_ready_res = readonly_ready_res;
```
NingLin-P, line 678:
```suggestion
```
`mem::take` will replace `self.poll_ctx.readonly_ready_res` with `Vec::default`

components/raftstore/src/store/fsm/store.rs
```diff
@@ -852,27 +880,46 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
         delegate.handle_msgs(&mut self.peer_msg_buf);
-        delegate.collect_ready();
-        self.poll_ctx.processed_fsm_count += 1;
-        expected_msg_count
+        // No readiness is generated, skipping calling ready and release early.
+        if !delegate.collect_ready(offset) {
+            if let HandleResult::StopAt { skip_end, .. } = &mut handle_result {
+                *skip_end = true;
+            }
+        }
```
NingLin-P, line 901: Should we also `skip_end` for those peers that `handle_result  == HandleResult::KeepProcessing`?

components/raftstore/src/store/fsm/store.rs
BusyJay: Yes, I reassign to avoid future allocations.

components/raftstore/src/store/fsm/store.rs
BusyJay: No, `KeepProcessing` means it has further messages to be processed. It may and may not have ready. Even if it doesn't have ready, keep processing it in the same thread can reduce context switch and friendly to caches.

NingLin-P (approved):

PR #10600, opened 2021-07-21
storage: prevent repeated non-pessimistic lock prewrite from writing locks

sticnarf:

Issue Number: close https://github.com/pingcap/tidb/issues/26359

Considering a pessimistic async-commit transaction, both row keys and index keys are prewritten successfully but the prewrite responses of the index keys are lost. Then, some other transactions commit the transaction by resolving locks because all locks exist. After that, the original committer resends the prewrite request of the index keys.

In the previous implementation, prewriting index keys of a pessimistic transaction will just write the locks again. The committer receives success and starts committing the transaction. Now the commit TS is different from the commit TS used by resolving locks. So the index keys will be committed twice, making the transaction have two different commit TS.

This PR just adds an additional check when prewriting the non-pessimistic lock keys. If the commit record already exists, we should not write the lock again.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
Fix the bug that index keys in a pessimistic transaction may be repeatedly committed.
```

### DIFF
--- a/src/storage/txn/actions/prewrite.rs
+++ b/src/storage/txn/actions/prewrite.rs
@@ -79,7 +79,7 @@ pub fn prewrite<S: Snapshot>(
     // we need to check the key has not been committed before if it is a retry request.
     //
     // It is to prevent the following case:
-    // The key was prewrited successfully before, but the response is lost. The client resend
+    // The key was prewritten successfully before, but the response is lost. The client resends
     // the same request after the transaction is resolved (to become committed). If we still
     // write the lock, the client might commit these keys more than once.
     //
@@ -97,7 +97,9 @@ pub fn prewrite<S: Snapshot>(
                 if write.write_type != WriteType::Rollback =>
             {
                 info!("prewrited transaction has been committed";
-                    "start_ts" => txn_props.start_ts, "commit_ts" => commit_ts);
+                    "start_ts" => txn_props.start_ts, "commit_ts" => commit_ts,
+                    "key" => ?mutation.key, "mutation_type" => ?mutation.mutation_type,
+                    "write_type" => ?write.write_type);
                 return Ok((commit_ts, OldValue::Unspecified));
             }
             _ => {}
--- a/src/storage/txn/actions/prewrite.rs
+++ b/src/storage/txn/actions/prewrite.rs
@@ -5,7 +5,7 @@ use crate::storage::{
             CONCURRENCY_MANAGER_LOCK_DURATION_HISTOGRAM, MVCC_CONFLICT_COUNTER,
             MVCC_DUPLICATE_CMD_COUNTER_VEC,
         },
-        Error, ErrorInner, Lock, LockType, MvccTxn, Result, SnapshotReader,
+        Error, ErrorInner, Lock, LockType, MvccTxn, Result, SnapshotReader, TxnCommitRecord,
     },
     txn::actions::check_data_constraint::check_data_constraint,
     txn::LockInfo,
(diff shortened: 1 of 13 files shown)


sticnarf:
The fix seems to introduce too much overhead. I propose to add a flag in `Context` to indicate this is a retry.

src/storage/txn/actions/prewrite.rs
```diff
@@ -75,6 +75,35 @@ pub fn prewrite<S: Snapshot>(
...
+        && !is_pessimistic_lock
+        && txn_props.retry_request
+        && matches!(txn_props.commit_kind, CommitKind::Async(..))
+    {
+        match reader.get_txn_commit_record(&mutation.key)? {
+            TxnCommitRecord::SingleRecord { commit_ts, write }
+                if write.write_type != WriteType::Rollback =>
+            {
+                info!("prewrited transaction has been committed";
+                    "start_ts" => txn_props.start_ts, "commit_ts" => commit_ts);
```
cfzjywxk, line 100: Maybe print some infomation about this mutation in the log.

cfzjywxk (approved):
/lgtm

src/storage/txn/actions/prewrite.rs
```diff
@@ -75,6 +75,37 @@ pub fn prewrite<S: Snapshot>(
...
+    // write the lock, the client might commit these keys more than once.
+    //
+    // If the commit record is Rollback, it is still safe for us to write the Lock because the primary
+    // key of the transaction must also be rolled back, and this lock will be eventually rolled back.
+    // For simplicity, we don't handle this case specially, making it the same behavior as the Rollback
+    // record is collapsed.
+    if txn_props.is_pessimistic()
+        && !is_pessimistic_lock
+        && txn_props.retry_request
+        && matches!(txn_props.commit_kind, CommitKind::Async(..))
```
MyonKeminta, line 93: Does 1PC have any related problem?

MyonKeminta: emm....  seems not

MyonKeminta (approved):

src/storage/txn/actions/prewrite.rs
sticnarf: Emm, I find something wrong. When encountering `PessimisticLockNotFound`, the buffered `modifies` should be cleared. Otherwise, the problem can happen on 1PC.

sticnarf: Ah, it won't make any problem because the final commit ts is correct, the only problem is that the same record is rewritten once...
Though it does not affect correctness, I still clear the `modifies` and add one more test for it. Please take a look.

MyonKeminta: Emmm.. The change looks fine to me, but I think I didn't fully understand it. In which case is it possible that a replayed 1PC request buffered some modifies?

sticnarf: For example, the first prewritten key does not need a pessimistic lock, so it saves its 1PC memory lock. Then the second key is a row key which needs a pessimistic lock, it then encounters PessimisticLockNotFound error. But later, we find the transaction is actually committed so no error is returned. We continue the 1PC process and convert the previous 1PC lock to a Write mutation.

cfzjywxk: Do we need to call the `txn.clear()` is not called here?

sticnarf: > Do we need to call the txn.clear() here?
Not necessary, but good to have. Now I added it.

PR #10609, opened 2021-07-22
coprocessor_plugin_api: Use atomic for allocator functions

plenkl:

Problem Summary:

1. The current implementation causes a segmentation fault whenever an allocation/deallocation occurs.
2. The coprocessor plugin version check matches any version except the one requested.

`AtomicPtr<AllocFn>` stores a pointer to a function pointer, so the deref on line 50 and 53 is incorrect as the types don't match (`*mut AllocFn` vs `AllocFn`).

Also inverts the version check for the coprocessor plugins.

### Check List 

Tests 

*Unit test*
The unit tests unfortunately do not catch this issue, since the allocator function pointers are not used during the tests: https://github.com/tikv/tikv/blob/master/components/coprocessor_plugin_api/src/util.rs#L90

Since `example_plugin` gets included as a dev-dependency we cannot set `#[global_allocator]` since it would conflict with https://github.com/tikv/tikv/blob/master/components/tikv_alloc/src/lib.rs#L129-L130.

There is some discussion about it here: https://github.com/tikv/tikv/pull/10112

*Manual test (add detailed scripts or steps below)*
I have done manual testing by loading and using a coprocessor plugin successfully in tikv.

Side effects

### Release note 

```release-note
Use atomic crate for coprocessor_plugin_api allocator functions
```

### DIFF
--- a/components/coprocessor_plugin_api/src/allocator.rs
+++ b/components/coprocessor_plugin_api/src/allocator.rs
@@ -1,7 +1,7 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
+use atomic::{Atomic, Ordering};
 use std::alloc::{GlobalAlloc, Layout};
-use std::sync::atomic::{AtomicPtr, Ordering};
 
 type AllocFn = unsafe fn(Layout) -> *mut u8;
 type DeallocFn = unsafe fn(*mut u8, Layout);
@@ -18,19 +18,19 @@ pub struct HostAllocatorPtr {
 
 /// An allocator that forwards invocations to the host (TiKV) of the plugin.
 pub struct HostAllocator {
-    alloc_fn: AtomicPtr<AllocFn>,
-    dealloc_fn: AtomicPtr<DeallocFn>,
+    alloc_fn: Atomic<Option<AllocFn>>,
+    dealloc_fn: Atomic<Option<DeallocFn>>,
 }
 
 impl HostAllocator {
     /// Creates a new [`HostAllocator`].
     ///
-    /// The internal function pointers are set to `nullptr`, so any attempt to allocate memory
-    /// before a call to [`set_allocator()`] will result in a segmentation fault.
+    /// The internal function pointers are initially `None`, so any attempt to allocate memory
+    /// before a call to [`set_allocator()`] will result in a panic.
     pub const fn new() -> Self {
         HostAllocator {
-            alloc_fn: AtomicPtr::new(std::ptr::null_mut()),
-            dealloc_fn: AtomicPtr::new(std::ptr::null_mut()),
+            alloc_fn: Atomic::new(None),
+            dealloc_fn: Atomic::new(None),
         }
     }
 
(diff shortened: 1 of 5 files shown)


andylokandy:
Is it possible to remove the `#[cfg(test)]` in`declare_plugin!()`?

skyzh:
Hi, you may need to sign DCO in your commit. You may run the following commands:

```
git reset HEAD~2 # You only have two commits, so reset to master
git add -A # Add your files
git commit -m "coprocessor_plugin_api: Use atomic for allocator functions" -s # Use -s to attach DCO in your commit
git push -f
```

plenkl:
> Is it possible to remove the `#[cfg(test)]` in`declare_plugin!()`?

Yes, I can remove it. But it still doesn't trigger the segmentation fault as long as `example_plugin` is a dependency of `tikv`. I think it is because the two `#[global_allocator]` statements conflict. If I remove `example_plugin` and compile it manually, I can trigger the segmentation fault (on the `master` branch) in the tikv tests
```
$ cargo test -p tikv -- registry
    Finished test [unoptimized + debuginfo] target(s) in 0.27s
     Running unittests (target/debug/deps/tikv-e3e5fec05f3ffa46)

running 5 tests
error: test failed, to rerun pass '-p tikv --lib'

Caused by:
  process didn't exit successfully: `/home/user/workspace/tikv/target/debug/deps/tikv-e3e5fec05f3ffa46 registry` (signal: 11, SIGSEGV: invalid memory reference)
```

However, then this will require another build step (to build `example_plugin` before running the tests). Related discussion here: https://github.com/tikv/tikv/pull/10112

Jibbow:
> Yes, I can remove it. But it still doesn't trigger the segmentation fault as long as `example_plugin` is a dependency of `tikv`. I think it is because the two `#[global_allocator]` statements conflict. If I remove `example_plugin` from the `tikv` dependencies and compile it manually, I can trigger the segmentation fault (on the `master` branch) in the tikv tests

+1
Also ran in this issue

skyzh:
> /lgtm

This command is disabled, so you may need to use GitHub review functionality to approve this PR :(

Jibbow:
> > /lgtm
> 
> This command is disabled, so you may need to use GitHub review functionality to approve this PR :(

ah whoops, thanks!

components/coprocessor_plugin_api/src/allocator.rs
```diff
@@ -39,17 +39,27 @@ impl HostAllocator {
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn atomic_is_lock_free() {
+        assert!(Atomic::<AllocFn>::is_lock_free());
```
skyzh, line 63: Should be `Atomic<Option<AllocFn>>` instead of `Atomic<AllocFn>`.

Jibbow (approved):

skyzh (commented):
Good work! Rest LGTM.

skyzh (approved):

PR #10610, opened 2021-07-22
engine_rocks: do not panic if disk is full when flushing or compacting.

MrCroxx:

Issue Number: close #xxx 

Problem Summary:

If there is no space for compaction or flushing, TiKV shoud still serve read-only requests, and should also allow smaller compactions.

To achieve that, incomplete sst files created by flushing or compactions shoud be deleted.

What's Changed:

Log an error instead of panic if disk is full when flushing or compacting, and auto recycle the incomplete sst files.

### Check List 

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)

Side effects

This pr needs to be carefully tested, in case critical data loss.

### Release note 

```release-note
If the disk is full when rocksdb flushing or compacting, TiKV will not panic, but log errors instead.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,9 +34,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.6"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


tabokie:
Let's extend the case to encountering any I/O error during flush or compaction.

tabokie:
Closing this in favor of #11225

PR #10615, opened 2021-07-22
fix incomplete sub_path canonicalization

tabokie:

Issue Number:

Problem Summary:

Sub path is not canonicalized.

What's Changed:

Concatenate sub path before canonicalize.

### Check List 

Tests
- Unit test

### Release note 

```release-note
None
```


yiwu-arbug (approved):

yiwu-arbug:
@tabokie why this one is closed?

tabokie:
@yiwu-arbug This turns out not to be the exact cause for tikv/raft-engine#75:
By default we use `data-dir`/raft-engine as raft engine path, `data-dir`/raft as raftdb path. The user manually links `data-dir`/raft to another disk and expects TiKV to automatically create a sibling directory for storing raft engine.

But it's related if we decide to use existing raftdb directory as raft engine path. I might reopen this later.

tabokie:
Ping @Connor1996

Connor1996 (approved):
rest LGTM

Issue #10514, opened 2021-07-04
Unstable test ``test_split_region::test_node_auto_split_region`


zhouqiang-cl:
## Bug Report
in ci https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/3515/pipeline

zhouqiang-cl:
```
[2021-07-02T11:21:00.837Z] test raftstore::test_split_region::test_node_auto_split_region ... thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Other("[components/test_raftstore/src/pd.rs:1304]: no region contains key ")', tests/integrations/raftstore/test_split_region.rs:181:42
[2021-07-02T11:21:00.837Z] stack backtrace:
[2021-07-02T11:21:00.837Z]    0: rust_begin_unwind
[2021-07-02T11:21:00.837Z]              at /rustc/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/library/std/src/panicking.rs:493:5
[2021-07-02T11:21:00.837Z]    1: core::panicking::panic_fmt
[2021-07-02T11:21:00.837Z]              at /rustc/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/library/core/src/panicking.rs:92:14
[2021-07-02T11:21:00.837Z]    2: core::result::unwrap_failed
[2021-07-02T11:21:00.837Z]              at /rustc/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/library/core/src/result.rs:1355:5
[2021-07-02T11:21:00.837Z]    3: core::result::Result<T,E>::unwrap
[2021-07-02T11:21:00.837Z]              at /rustc/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/library/core/src/result.rs:1037:23
[2021-07-02T11:21:00.837Z]    4: integrations::raftstore::test_split_region::test_auto_split_region
[2021-07-02T11:21:00.837Z]              at /home/jenkins/tikv-src/tests/integrations/raftstore/test_split_region.rs:181:16
[2021-07-02T11:21:00.837Z]    5: integrations::raftstore::test_split_region::test_node_auto_split_region
[2021-07-02T11:21:00.837Z]              at /home/jenkins/tikv-src/tests/integrations/raftstore/test_split_region.rs:227:5
[...]

Issue #10448, opened 2021-06-23
Unstable test `raftstore::test_split_region::test_server_auto_split_region`


zhouqiang-cl:
## Bug Report
in ci https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/2994/pipeline

### What version of TiKV are you using?

master

zhouqiang-cl:
happends again in https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/4832/pipeline

Issue #10446, opened 2021-06-23
Unstable test `raftstore::test_split_region::test_incompatible_node_auto_split_region`


zhouqiang-cl:
## Bug Report

```
[2021-06-23T11:51:05.196Z] + awk '/^failures/{flag=1}/^test result:/{flag=0}flag' tests.out

[2021-06-23T11:51:05.196Z] failures:

[2021-06-23T11:51:05.196Z] 

[2021-06-23T11:51:05.196Z] failures:

[2021-06-23T11:51:05.196Z]     raftstore::test_split_region::test_incompatible_node_auto_split_region
```

### What version of TiKV are you using?

master bd0d265369bf796bcdf6a35d0021f9c484c663d6

### Steps to reproduce

in ci https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/2985/pipeline/101

PR #10622, opened 2021-07-23
test: retry on get_region() to stable related test

Connor1996:

Issue Number: close https://github.com/tikv/tikv/issues/10514 https://github.com/tikv/tikv/issues/10448 https://github.com/tikv/tikv/issues/10446

Problem Summary:
This test is failed because the two after-split regions' heartbeats are reported separately and there is a time that a hole in the PD's region cache. So if tests call `get_region` at that time exactly right, then it may find no region for the key.

What's Changed: 
Retry on `get_region()`, the hole in the PD's region cache will be filled up in the final.

### Related changes

- None

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -758,10 +758,11 @@ dependencies = [
 [[package]]
 name = "crossbeam-epoch"
 version = "0.9.3"
-source = "git+https://github.com/tikv/crossbeam.git?branch=tikv-5.0#e0e083d062649484188b7337fe388fd12f2c8d94"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2584f639eb95fea8c798496315b297cf81b9b58b6d30ab066a75455333cf4b12"
 dependencies = [
  "cfg-if 1.0.0",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -758,10 +758,11 @@ dependencies = [
 [[package]]
 name = "crossbeam-epoch"
 version = "0.9.3"
-source = "git+https://github.com/tikv/crossbeam.git?branch=tikv-5.0#e0e083d062649484188b7337fe388fd12f2c8d94"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2584f639eb95fea8c798496315b297cf81b9b58b6d30ab066a75455333cf4b12"
```
BusyJay, line 762: Why changes?

Connor1996: Have no idea, it's changed automatically

BusyJay (approved):

gengliqi (approved):

PR #10625, opened 2021-07-26
resource_metering: tiny fix for linux cpu collector 

crazycs520:

Before this PR, the resource_metering collected CPU doesn't match the real thread CPU usage.

### Check List 

Tests 

- Manual test

Side effects

- N/A

### Release note 

```release-note
None
```


crazycs520:
@zhongzc PTAL

zhongzc:
dup of #10668

PR #10636, opened 2021-07-28
resource_metering: fix the enabled reporter doesn't initialize client during starting server

crazycs520:
Signed-off-by: crazycs <crazycs520@gmail.com>

As the title said. If the config `resource-metering` is enabled when the server started, the `resource-metering` won't  work since it's client doesn't init.

### Check List 

Tests 

- Manual test

### Release note 

```release-note
None
```


crazycs520:
@zhongzc PTAL

components/resource_metering/src/reporter.rs
```diff
@@ -110,13 +114,13 @@ impl Runnable for ResourceMeteringReporter {
                 if !new_config.should_report() {
                     self.client.take();
                     self.cpu_records_collector.take();
+                    self.config = new_config;
```
zhongzc, line 117: why moving the outer to here?

crazycs520: since the `if` and `else` branch both need this, but the order is different, so I move to here.

zhongzc:
I have an idea. How about lazily initializing the client at checking `should_report` so we don't need to do that during config change?

zhongzc (approved):

zhongzc (changes requested):
test failed

zhongzc (approved):

breeswish (approved):

crazycs520:
should cherry-pick? @zhongzc @breeswish

crazycs520:
> cherry pick to release-5.2 failed

release-5.2 already has this fix.

PR #10652, opened 2021-08-01
txn: skip resolve commit fail for left pessimistic locks

cfzjywxk:

Issue Number: 
https://github.com/pingcap/tidb/issues/26404

Problem Summary:
The read query execution fails because the resolve lock reports error. In the situation the read and write conflicts with each other, there is possibility that the pessimistic locks are lefted, these lefted locks could be fetched by the resolve read phase but pessimistically rolled back before the resolve write phase by the async pessimistic lock rollback.
 

What's Changed:
Do not report error if the resolve commit lock is a pessimistic lock type, then the read queries will not fail and report error to the client.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
Fix the resolve failures caused by the left pessimisic locks.
```

### DIFF
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -364,6 +364,10 @@ impl Lock {
     pub fn is_pessimistic_txn(&self) -> bool {
         !self.for_update_ts.is_zero()
     }
+
+    pub fn is_pessimistic_lock_type(&self) -> bool {
+        self.lock_type == LockType::Pessimistic
+    }
 }
 
 #[cfg(test)]
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -365,7 +365,7 @@ impl Lock {
         !self.for_update_ts.is_zero()
     }
 
-    pub fn is_pessimistic_lock_type(&self) -> bool {
+    pub fn is_pessimistic_lock(&self) -> bool {
         self.lock_type == LockType::Pessimistic
     }
 }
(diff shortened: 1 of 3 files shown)


components/txn_types/src/lock.rs
```diff
@@ -364,6 +364,10 @@ impl Lock {
     pub fn is_pessimistic_txn(&self) -> bool {
         !self.for_update_ts.is_zero()
     }
+
+    pub fn is_pessimistic_lock_type(&self) -> bool {
```
youjiali1995, line 368:
```suggestion
    pub fn is_pessimistic_lock(&self) -> bool {
```

src/storage/txn/commands/resolve_lock.rs
```diff
@@ -93,7 +95,20 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for ResolveLock {
...
-                commit(&mut txn, &mut reader, current_key.clone(), commit_ts)?
+                // Continue to resolve locks if the not found committed locks are pessimistic type.
+                // They could be left if the transaction is finally committed and pessimistic conflict
+                // retry happens during execution.
+                match commit(&mut txn, &mut reader, current_key.clone(), commit_ts) {
+                    Ok(res) => res,
+                    Err(MvccError(box MvccErrorInner::TxnLockNotFound { .. }))
+                        if current_lock.is_pessimistic_lock_type() =>
+                    {
+                        info!("skip resolve commit pessimistic lock error as it's not found";
```
youjiali1995, line 106: I don't think this log is necessary since we already log once in commit, and the debug fmt of `Lock` is very verbose.😂

youjiali1995 (commented):
Rest LGTM

MyonKeminta (approved):

youjiali1995 (approved):

PR #10656, opened 2021-08-02
raftstore: add message for inspecting latency

5kbpers:

Problem Summary:
Raftstore is the component with the responsibility to handle user writes in TiKV, thus the latency of Raftstore could be considered to be a representative metric of the speed of TiKV.
This PR adds a new message type for inspecting the latency of Raftstore to measure whether the store was slow down (related to #10539)

TODO: need a patch after async-io gets merged

* Add a new message type for measuring the performance of Raftstore, includes most of the stages of writing a command(store-wait, store-process, apply-wait, apply-process)

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
N/A, next PR that detects the latency will include release note.
```


yiwu-arbug (approved):
LGTM

hicqu (approved):

PR #10661, opened 2021-08-04
alertmanager: no GC cannot work alert if no compaction happens

sticnarf:

After TiDB 5.0, compaction filter is the default GC approach. If the cluster is idle thus no compaction is triggered, the GC compaction filter will not be triggered. Then, there will be a false "GC can not work" alert which is really annoying.

This PR modifies the alertmanager rule of "GC can not work". It raises the alert only if there are compaction operations while no GC compaction filter runs.

### Related changes

### Check List 

Tests 

### Release note 

```release-note
Avoid false "GC can not work" alert under low write flow.
```

### DIFF
--- a/metrics/alertmanager/tikv.rules.yml
+++ b/metrics/alertmanager/tikv.rules.yml
@@ -14,7 +14,7 @@ groups:
       summary: TiKV memory used too fast
 
   - alert: TiKV_GC_can_not_work
-    expr: sum(increase(tikv_gcworker_gc_tasks_vec{task="gc"}[1d])) < 1 and sum(increase(tikv_gc_compaction_filter_perform[1d])) < 1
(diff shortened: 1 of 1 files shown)


MyonKeminta:
Is it that when compaction is performed on WRITE_CF, there must be some records in `tikv_gc_compaction_filter_perform` (if enabled)?

sticnarf:
> Is it that when compaction is performed on WRITE_CF, there must be some records in `tikv_gc_compaction_filter_perform` (if enabled)?

It seems so, unless there is something wrong such as no safepoint is set or the database is stall.

MyonKeminta (approved):
🤔 LGTM

hicqu (approved):

PR #10666, opened 2021-08-04
cdc: reduce resolved ts message size

overvenus:

Issue Number: close https://github.com/pingcap/ticdc/issues/2448

Problem Summary:

Reduce resolved ts message size to save network bandwidth.

What's Changed:

Only attaching region ids that are captured by connections.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Reduce resolved ts message size to save network bandwidth.
```


liuzix:
Is it possible to maintain the list of all subscribed regions as a `hashset`, and pass the resolved regions in as a `vector`? This way we can avoid having to build a hashset whenever we want to send a resolved event, saving some CPU.

overvenus:
> Is it possible to maintain the list of all subscribed regions as a `hashset`, and pass the resolved regions in as a `vector`? This way we can avoid having to build a hashset whenever we want to send a resolved event, saving some CPU.

It can save some CPU, although I don't think it the bottleneck for now. In tests, capturing 10k region takes ~1.5% CPU.

hi-rustin:
/cc

I will help look at it tonight or tomorrow.

liuzix (approved):

5kbpers (approved):

PR #10671, opened 2021-08-05
handle proposals correctly if majority peers are disk full

hicqu:

This PR handles proposals correctly for majority peers are disk full. It contains:
* Raft peers report disk usage to each other. Please also take a look at [#kvproto/791](https://github.com/pingcap/kvproto/pull/791).
* Leader won't append entries to peers which are disk full any more. It's implemented with `Raft::adjust_max_inflight_msgs`.
* If a proposal has a special `DiskFullOpt` flag, leader can broadcast it in order to commit it.

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -127,8 +127,6 @@ where
     // Batch raft command which has the same header into an entry
     batch_req_builder: BatchRaftCmdRequestBuilder<EK>,
 
-    max_inflight_msgs: usize,
-
     trace: PeerMemoryTrace,
 }
 
@@ -236,7 +234,6 @@ where
                 batch_req_builder: BatchRaftCmdRequestBuilder::new(
                     cfg.raft_entry_max_size.0 as f64,
                 ),
-                max_inflight_msgs: cfg.raft_max_inflight_msgs,
                 trace: PeerMemoryTrace::default(),
             }),
         ))
@@ -281,7 +278,6 @@ where
                 batch_req_builder: BatchRaftCmdRequestBuilder::new(
                     cfg.raft_entry_max_size.0 as f64,
                 ),
-                max_inflight_msgs: cfg.raft_max_inflight_msgs,
                 trace: PeerMemoryTrace::default(),
             }),
         ))
@@ -317,6 +313,9 @@ where
 
     pub fn reset_hibernate_state(&mut self, state: GroupState) {
         self.hibernate_state.reset(state);
+        if state == GroupState::Idle {
+            self.peer.raft_group.raft.maybe_free_inflight_buffers();
+        }
     }
 
     pub fn maybe_hibernate(&mut self) -> bool {
@@ -4343,8 +4342,8 @@ mod memtrace {
 
         pub fn raft_progress_size(&self) -> usize {
             let peer_cnt = self.peer.region().get_peers().len();
-            let inflight_size = self.max_inflight_msgs * mem::size_of::<u64>();
-            mem::size_of::<Progress>() * peer_cnt * 6 / 5 + inflight_size * peer_cnt
+            mem::size_of::<Progress>() * peer_cnt * 6 / 5
+                + self.peer.raft_group.raft.inflight_buffers_size()
         }
     }
 }
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1333,10 +1333,17 @@ where
             "to_peer_id" => msg.get_to_peer().get_id(),
         );
 
-        let msg_type = msg.get_message().get_msg_type();
         self.handle_reported_disk_usage(&msg);
 
-        if matches!(self.ctx.self_disk_usage, DiskUsage::AlreadyFull) {
+        let msg_type = msg.get_message().get_msg_type();
+        if matches!(self.ctx.self_disk_usage, DiskUsage::AlreadyFull)
+            && [
+                MessageType::MsgSnapshot,
+                MessageType::MsgAppend,
+                MessageType::MsgTimeoutNow,
+            ]
+            .contains(&msg_type)
+        {
             debug!(
                 "skip {:?} because of disk full", msg_type;
                 "region_id" => self.region_id(), "peer_id" => self.fsm.peer_id()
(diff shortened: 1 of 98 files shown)


tier-cap:
One Advice: If we can change the intercept condition of processing the received raft message from "if not disk normal" to "if disk is already full", then ti will work fine, and no need to get disk usage info from followers. One advantage is Logically decoupling.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -590,8 +591,7 @@ where
                     }
 
                     let req_size = cmd.request.compute_size();
-                    if self.ctx.cfg.cmd_batch
-                        && cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull
+                    if cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull
```
5kbpers, line 594: Why change this?

5kbpers (approved):

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -591,8 +592,8 @@ where
 
                     let req_size = cmd.request.compute_size();
                     if self.ctx.cfg.cmd_batch
-                        && cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull
                         && self.fsm.batch_req_builder.can_batch(&cmd.request, req_size)
+                        && cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull
```
gengliqi, line 596: Seems the `disk_full_op` has nothing to do with cmd batch?

hicqu: Normal requests will be rejected in `propsoe` if the store's disk is full. So here we avoid to merge them, so that only some special operations will be allowed.

gengliqi: Got it. Would you mind to add comments?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -393,7 +393,12 @@ where
     pub tick_batch: Vec<PeerTickBatch>,
     pub node_start_time: Option<TiInstant>,
     pub pending_latency_inspect: Vec<util::LatencyInspector>,
-    pub disk_usage: disk::DiskUsage,
+
+    /// Disk usage for the store itself.
+    pub self_disk_usage: DiskUsage,
+    /// Disk usage for other stores. The store itself is not included.
+    /// Only contains items which is not `DiskUsage::Normal`.
+    pub store_disk_usages: HashMap<u64, DiskUsage>,
```
gengliqi, line 401: If a peer is removed by conf change, when does it remove from `store_disk_usages`?
It seems just doing it in `handle_reported_disk_usage` is not enough?

hicqu: Nice catch. The hash map is in `PollContext` and only contains `store_id` instead of `peer_id`, so the only problem is how to handle an offline store. I have no idea now. How about left a todo here?

gengliqi: LGTM.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1297,6 +1298,38 @@ where
...
+                        1
+                    }
+                } else {
+                    return;
+                }
+            }
+        };
+        if self.fsm.peer.max_inflight_msgs != max_inflight_msgs {
+            self.fsm.peer.max_inflight_msgs = max_inflight_msgs;
+            raft.adjust_max_inflight_msgs(id, 1);
```
gengliqi, line 1329:
```suggestion
            raft.adjust_max_inflight_msgs(id, max_inflight_msgs);
```

components/raftstore/src/store/peer.rs
```diff
@@ -3344,6 +3332,87 @@ where
...
+        // Then if majority peers for the raft group meets the disk full problem,
+        // reject the propose.
+        let mut check_majority_disk_full = || {
+            let self_peer_id = self.peer_id();
+            let peers_len = self.get_store().region().get_peers().len();
+
+            let mut disk_full_peers = HashSet::default();
+            let mut next_idxs = Vec::with_capacity(peers_len);
+            for peer in self.get_store().region().get_peers() {
+                let (peer_id, store_id) = (peer.get_id(), peer.get_store_id());
```
gengliqi, line 3362: Should only consider voter?

components/raftstore/src/store/peer.rs
```diff
@@ -3344,6 +3332,87 @@ where
...
+    fn check_disk_usages_before_propose_normal<T>(
+        &mut self,
+        ctx: &mut PollContext<EK, ER, T>,
+        disk_full_opt: DiskFullOpt,
+        disk_full_stores: &mut Vec<u64>,
+    ) -> bool {
+        // Propose can be rejected if of the store's disk is full.
+        let self_allowed = match disk_full_opt {
+            DiskFullOpt::NotAllowedOnFull => matches!(ctx.self_disk_usage, DiskUsage::Normal),
+            _ => !matches!(ctx.self_disk_usage, DiskUsage::AlreadyFull),
```
gengliqi, line 3347: If it's `AllowedOnAlreadyFull`, is it can be proposed?

hicqu: `AllowedOnAlreadyFull` is not used exactly. We will remove it from `kvproto` later.

components/raftstore/src/store/peer.rs
```diff
@@ -3344,6 +3332,87 @@ where
         assert!(self.pending_merge_state.is_some());
         self.want_rollback_merge_peers.insert(peer_id);
     }
+
+    // Check disk usages for the peer itself and other peers in the raft group.
+    // The return value indicates whether the propose is allowed or not.
```
gengliqi, line 3337: The comments here and below should use the `proposal` instead of `propose`.

components/raftstore/src/store/peer.rs
hicqu: It's unnecessary to handle learners specially because we use `has_quorum` to judge a quorum.

hicqu:
> One Advice: If we can change the intercept condition of processing the received raft message from "if not disk normal" to "if disk is already full", then ti will work fine, and no need to get disk usage info from followers. One advantage is Logically decoupling.

Of course it's a potential solution. However maintaining a Vec<(index, DiskFullOpt)> correctly in raftstore is not easy. You can take try if you have time.

tier-cap:
> > One Advice: If we can change the intercept condition of processing the received raft message from "if not disk normal" to "if disk is already full", then ti will work fine, and no need to get disk usage info from followers. One advantage is Logically decoupling.
> 
> Of course it's a potential solution. However maintaining a Vec<(index, DiskFullOpt)> correctly in raftstore is not easy. You can take try if you have time.
Set all the sended raft msg with the leader disk usage status flag, and let follower's first threshold passed and second threshold rejected. Then it will work well.

gengliqi (approved):

PR #10672, opened 2021-08-05
raftstore: always enable entry cache for async write io

gengliqi:

Problem Summary:
This PR is split from https://github.com/tikv/tikv/pull/10289.

What's Changed:
Always enable entry cache and optimize `term` function a bit.

### Related changes

### Check List 

Tests 

- No code

Side effects

- No

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/peer_storage.rs
+++ b/components/raftstore/src/store/peer_storage.rs
@@ -225,10 +225,13 @@ impl EntryCache {
         mem_size_change
     }
 
-    pub fn term(&self, idx: u64) -> u64 {
-        let cache_low = self.cache.front().unwrap().get_index();
-        let start_idx = idx.checked_sub(cache_low).unwrap() as usize;
-        self.cache[start_idx].get_term()
+    pub fn entry(&self, idx: u64) -> Option<&Entry> {
+        let cache_low = self.cache.front()?.get_index();
+        if idx >= cache_low {
+            Some(&self.cache[(idx - cache_low) as usize])
+        } else {
+            None
+        }
     }
 
     pub fn compact_to(&mut self, mut idx: u64) -> u64 {
@@ -876,9 +879,8 @@ where
         if self.truncated_term() == self.last_term || idx == self.last_index() {
             return Ok(self.last_term);
         }
-        let cache_low = self.cache.first_index().unwrap_or(u64::MAX);
-        if idx >= cache_low {
-            Ok(self.cache.term(idx))
+        if let Some(e) = self.cache.entry(idx) {
+            Ok(e.get_term())
         } else {
             let mut entries = vec![];
             self.engines.raft.fetch_entries_to(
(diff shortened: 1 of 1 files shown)


components/raftstore/src/store/peer_storage.rs
```diff
@@ -1164,64 +1163,50 @@ where
...
     }
 
     pub fn compact_cache_to(&mut self, idx: u64) {
-        if let Some(ref mut cache) = self.cache {
-            cache.compact_to(idx);
-        } else {
-            let rid = self.get_region_id();
+        self.cache.compact_to(idx);
+        let rid = self.get_region_id();
+        if self.engines.raft.has_builtin_entry_cache() {
```
BusyJay, line 1174: Can it be removed completely?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -887,8 +876,20 @@ where
         if self.truncated_term() == self.last_term || idx == self.last_index() {
             return Ok(self.last_term);
         }
-        let entries = self.entries(idx, idx + 1, raft::NO_LIMIT)?;
-        Ok(entries[0].get_term())
+        let cache_low = self.cache.first_index().unwrap_or(u64::MAX);
```
BusyJay, line 879: How about adding a `entry` method that returns a single entry instead?

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/peer_storage.rs
gengliqi: I don't know the future plan for raft engine. /cc @hicqu @tabokie

components/raftstore/src/store/peer_storage.rs
```diff
@@ -887,8 +879,19 @@ where
         if self.truncated_term() == self.last_term || idx == self.last_index() {
             return Ok(self.last_term);
         }
-        let entries = self.entries(idx, idx + 1, raft::NO_LIMIT)?;
-        Ok(entries[0].get_term())
+        if let Some(e) = self.cache.entry(idx) {
+            Ok(e.get_term())
+        } else {
+            let mut entries = vec![];
```
BusyJay, line 885: You mean `self.engines.raft.get_entry(self.get_region_id(), idx)`?

BusyJay (approved):

components/raftstore/src/store/peer_storage.rs
gengliqi: Yep. It's the original code. I change to use `get_entry`.

components/raftstore/src/store/peer_storage.rs
hicqu: Entry cache will be removed from RaftEngine. However for now let's keep the code.

hicqu (approved):

PR #10675, opened 2021-08-06
Update raft engine

tabokie:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

Mainly to include metrics and interfaces: tikv/raft-engine#82, tikv/raft-engine#76.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raft_log_engine/src/engine.rs
+++ b/components/raft_log_engine/src/engine.rs
@@ -8,25 +8,31 @@ use engine_traits::{
 };
 use kvproto::raft_serverpb::RaftLocalState;
 use raft::eraftpb::Entry;
-use raft_engine::{EntryExt, Error as RaftEngineError, LogBatch, RaftLogEngine as RawRaftEngine};
+use raft_engine::{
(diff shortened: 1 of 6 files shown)


components/raft_log_engine/src/engine.rs
```diff
@@ -173,25 +173,17 @@ impl RaftEngine for RaftLogEngine {
...
-            miss: stat.miss,
-            cache_size: stat.cache_size,
-        })
+        None
     }
 
-    fn stop(&self) {
-        self.0.stop();
-    }
+    fn stop(&self) {}
```
MrCroxx, line 186: Shall we support stop RaftLogEngine in future?

MrCroxx (approved):

components/raft_log_engine/src/engine.rs
tabokie: Raft engine itself doesn't manage any threads anymore, so stop is simply exiting the client process.

Connor1996 (approved):
LGTM

hicqu (approved):

PR #10681, opened 2021-08-08
Substitute write stall with flow control in scheduler level (#10249)

ti-srebot:
cherry-pick #10249 to release-5.2
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/10681
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/10681:release-5.2-b32ee4afcb29
```

---

Issue Number: close https://github.com/tikv/tikv/issues/10137 

What's Changed:

Do flow control at scheduler level, and disable the write stall mechanism of RocksDB.

The main idea of this flow controller is to throttle at a steady write rate, so that the number of L0 keeps around the threshold. When it falls below the threshold, the throttle state wouldn't exit right away. Instead, it may keep or increase the throttle speed depending on some statistics.

### Related changes

### Check List 

Tests 

- Unit test

#### For single-node setup with heavy TPCC load
The result shows there is no more QPS drop-0 phenomenon, and QPS is relatively steady compared to write stall. Though the maximum QPS is lower, new flow control gain better throughput

write stall
image

new flow control
image

#### For multiple-node setup with low TPCC load
Write stall behavior well in this case. The new flow control is close to the write stall’s behavior but not that steady. But the throughput is the same.

write stall
image

new flow control
image

#### For multiple-node setup with heavy TPCC load
Flow control is better, as you can see write stall causes some QPS drop but flow control doesn't

write stall(not finish running a whole round caused by physical accident, but it's enough to illustrate the QPS drop issue of write stall)
image

new flow control
image

### Release note 

- Substitute write stall with flow control in scheduler level


Connor1996 (approved):
LGTM

zhouqiang-cl (approved):
LGTM

PR #10690, opened 2021-08-09
raft_client: fix imbalance load of raft msg (#10226) (#10495)

ti-srebot:
cherry-pick #10495 to release-5.0
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/10690
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/10690:release-5.0-7e2a44fd022c
```

---

cherry-pick #10226 to release-5.1
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/10495
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/10495:release-5.1-a41ae4597455
```

---

Problem Summary:

1. imbalance load of raft msg: raft client uses `region_id % grpc-raft-conn-num` to choose the specific connection. It's no problem when `grpc-raft-conn-num` is 1(default value). However, when `grpc-raft-conn-num` is larger, for example, 2, the distribution maybe not uniform because of the region id. If the num of TiKV is 3, the region id maybe always even because PD allocates region id with 3 peer id. To fix this problem, @5kbpers suggests using seahash. I test it with a parity check and find the distribution is very uniform. Even if the value is even, its hash value is odd or even. Also, the speed of seahash is good enough(benchmark result shows 11ns/iter on my laptop).
~~2. raft msg flush strategy: The raft msg is flushed only after handling all regions msg of this round. It may introduce some unnecessary latency. However, if the raft msg is flushed every time, the grpc CPU will be very high because of a large number of notifying. To alleviate this problem, the delay time is introduced. But I think this method is not very elegant. Maybe we should refactor this logic later.~~

What's Changed:
explained above.

### Related changes

### Check List 

Tests 

- No code

Side effects

- No

### Release note 
* No Release Note


BusyJay:
Duplicated with #10496.

PR #10693, opened 2021-08-09
*:fix one bug about drop table/truncate table failed when TiKV is almost disk full

tier-cap:

Fix drop table failed when TiKV is almost disk full.

Problem Summary:
1. Push data until TiKV is disk full.
2. Exec Drop Table, then it will be blocked.

1. Proposal will be accepted by leader if TiKV is not disk full, except with special flag.
2. Newest PR change the judge condition only AllowedOnAlmostFull, but left deal with the AllowedOnAlreadyFull.
3. This PR unifies these differences.

### Related changes

### Release note

```
None
```

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -2319,7 +2319,10 @@ where
                 if self.check_disk_usages_before_propose(ctx, disk_full_opt, &mut stores) {
                     self.propose_normal(ctx, req)
                 } else {
-                    let errmsg = String::from("propose failed: disk full");
+                    let errmsg = format!(
+                        "propose failed: tikv disk full, cmd-disk_full_opt={:?}, leader-diskUsage={:?}",
+                        disk_full_opt, ctx.self_disk_usage
+                    );
                     Err(Error::DiskFull(stores, errmsg))
                 }
             }
@@ -2329,7 +2332,10 @@ where
                 if self.check_disk_usages_before_propose(ctx, disk_full_opt, &mut stores) {
                     self.propose_conf_change(ctx, &req)
                 } else {
-                    let errmsg = String::from("propose failed: disk full");
+                    let errmsg = format!(
+                        "propose failed: tikv disk full, cmd-disk_full_opt={:?}, leader-diskUsage={:?}",
+                        disk_full_opt, ctx.self_disk_usage
+                    );
                     Err(Error::DiskFull(stores, errmsg))
                 }
             }
--- a/components/raftstore/src/store/worker/pd.rs
+++ b/components/raftstore/src/store/worker/pd.rs
@@ -46,7 +46,6 @@ use futures::FutureExt;
 use pd_client::metrics::*;
 use pd_client::{Error, PdClient, RegionStat};
 use tikv_util::metrics::ThreadInfoStatistics;
-use tikv_util::sys::disk;
 use tikv_util::time::UnixSecs;
 use tikv_util::timer::GLOBAL_TIMER_HANDLE;
 use tikv_util::topn::TopN;
@@ -895,13 +894,6 @@ where
         let mut available = capacity.checked_sub(used_size).unwrap_or_default();
         // We only care about rocksdb SST file size, so we should check disk available here.
         available = cmp::min(available, disk_stats.available_space());
-        available = available
-            .checked_sub(disk::get_disk_reserved_space())
-            .unwrap_or_default();
-
-        if disk::is_disk_full() {
-            available = 0;
-        }
 
         if available == 0 {
             warn!("no available space");
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -17,6 +17,7 @@ use std::{
     fs::{self, File},
     net::SocketAddr,
     path::{Path, PathBuf},
+    str::FromStr,
     sync::{
         atomic::{AtomicU32, AtomicU64, Ordering},
         mpsc, Arc, Mutex,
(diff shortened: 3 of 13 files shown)


components/raftstore/src/store/peer.rs
```diff
@@ -2319,7 +2319,11 @@ where
                 if self.check_disk_usages_before_propose(ctx, disk_full_opt, &mut stores) {
                     self.propose_normal(ctx, req)
                 } else {
-                    let errmsg = String::from("propose failed: disk full");
+                    // let errmsg = String::from("propose failed: disk full");
+                    let errmsg = format!(
+                        "propose failed: disk full, cmd::disk_full_opt={:?}, diskUsage={:?}",
```
hicqu, line 2324: Maybe it's not caused by `self_disk_usage`. It's better to show disk full stores instead.

components/raftstore/src/store/peer.rs
```diff
@@ -2329,7 +2333,11 @@ where
                 if self.check_disk_usages_before_propose(ctx, disk_full_opt, &mut stores) {
                     self.propose_conf_change(ctx, &req)
                 } else {
-                    let errmsg = String::from("propose failed: disk full");
+                    // let errmsg = String::from("propose failed: disk full");
+                    let errmsg = format!(
+                        "propose failed: disk full, cmd::disk_full_opt={:?}, diskUsage={:?}",
```
hicqu, line 2338: Same as above.

hicqu:
I think it can't fix the problem in the PR description, although it can be merged.

components/raftstore/src/store/peer.rs
tier-cap: It's caused by the judge condition, showed at check_disk_usages_before_propose, as what the changes describe.

components/raftstore/src/store/peer.rs
tier-cap: This is only a log,, not the drop table failed logic

tier-cap:
> I think it can't fix the problem in the PR description, although it can be merged. So please update the PR description.

Pls just take a test, and review comment is not get focused what this fix changes.
That comments are only a error msg log, not logically.

hicqu:
> 
> 
> > I think it can't fix the problem in the PR description, although it can be merged. So please update the PR description.
> 
> Pls just take a test, and review comment is not get focused what this fix changes.
> That comments are only a error msg log, not logically.

If you think it's really helpful to fix the descripted bug, you can add integration tests.

src/storage/txn/scheduler.rs
```diff
@@ -729,15 +729,8 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
             }) => {
                 SCHED_STAGE_COUNTER_VEC.get(tag).write.inc();
                 match ctx.get_disk_full_opt() {
```
hicqu, line 731: The match seems not necessary. Why not `to_be_write.disk_full_opt = ctx.get_disk_full_opt()`?

components/raftstore/src/store/worker/pd.rs
```diff
@@ -889,19 +889,13 @@ where
         stats.set_capacity(capacity);
 
         let used_size = self.snap_mgr.get_total_snap_size().unwrap()
-            + store_info.engine.get_engine_used_size().expect("cf");
+            + store_info.engine.get_engine_used_size().expect("cf")
+            + disk::get_disk_reserved_space() / 5;
```
hicqu, line 893: The file may doesn't exist. `fstate` the file is better.

components/server/src/server.rs
```diff
@@ -1107,7 +1107,7 @@ impl<ER: RaftEngine> TiKVServer<ER> {
                     .get_engine_size()
                     .expect("get raft engine size");
 
-                let used_size = snap_size + kv_size + raft_size;
+                let used_size = snap_size + kv_size + raft_size + reserve_space / 5;
```
hicqu, line 1110: Also needs to `fstat` the place holder file.

components/raftstore/src/store/worker/pd.rs
tier-cap: Removed, to keep Pd store heartbeat used size as the original

components/server/src/server.rs
tier-cap: Ok, add stats.

src/storage/txn/scheduler.rs
tier-cap: No, the to_be_write setting will be overwriten by client setting.

components/server/src/server.rs
```diff
@@ -1107,7 +1109,16 @@ impl<ER: RaftEngine> TiKVServer<ER> {
                     .get_engine_size()
                     .expect("get raft engine size");
 
-                let used_size = snap_size + kv_size + raft_size;
+                let placeholer_file_path = PathBuf::from_str(&data_dir)
+                    .unwrap()
+                    .join(Path::new(file_system::SPACE_PLACEHOLDER_FILE));
+                let placeholder_size: u64 = if file_system::file_exists(&placeholer_file_path) {
```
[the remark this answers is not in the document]
tier-cap: Ok, fix with unwrap_or_else

hicqu (approved):

5kbpers (approved):

5kbpers:
Don't forget to push this fix to master.

Mini256:
@tier-cap Hi, for removing the `do-not-merge/release-note-label-needed` label, You need to specify the language of the code block as 'release-note'.

like this: 

    ```release-note
    None
    ```

not like:

    ```
    None
    ```

Issue #10716, opened 2021-08-12
Performance regression due to disk full option for cmd batch


gengliqi:
### What version of TiKV are you using?
master

### What operating system and CPU are you using?
doesn't matter

### Steps to reproduce
run sysbench insert workload

### What did you expect?
no performance regression

### What did happened?
performance regression

https://github.com/tikv/tikv/blob/834ec1108d85623b2b19b00df2fa8e1359c726b6/components/raftstore/src/store/fsm/peer.rs#L596-L599
I remove this line of code above and compare it to the original one by running sysbench insert.
(The left is the original code.)
![image](https://user-images.githubusercontent.com/21253569/129192852-8bb3f354-6c8a-44d8-a4a9-3571f89817c9.png)
The code above leads to the cmd batch's poor effect, producing more msgs and more CPU overhead.
![image](https://user-images.githubusercontent.com/21253569/129192911-f9416e16-744b-4678-ab3a-8f9fa5b04e7d.png)
As a result, the QPS of the original code is much lower than the modified one.

gengliqi:
@hicqu @tier-cap PTAL, this issue is very urgent.

PR #10717, opened 2021-08-12
*: Fix disk full feature leads to batch performance back-off

tier-cap:
Issue Number:: https://github.com/tikv/tikv/issues/10716
Solve disk full option setting leads to raft batch performance back off.

Problem Summary:
When raftcmd is set with special flag about disk full, then the batch system will be drained, which leads to the batch system not work in more time.

Set the special flag back to delegate, and let all the batch be set with special flag if current raftcmd is set with special flag.

### Related changes

### Check List
Performance regression.

Tests

Side effects

- Performance regression

### Release note

```
None
```


tier-cap:
target branch

PR #10719, opened 2021-08-12
*: fix disk full leads to batch perf back off

tier-cap:
Signed-off-by: tier-cap <zhengxiaojin@pingcap.com>

Issue Number: https://github.com/tikv/tikv/issues/10716
Problem Summary:
Raft batch performance down when tikv disk full opt setting blocks batch.

Let all the batch be set with the same flag if some raftcmd is set.

### Related changes

### Check List
Tests
- Performance regression

### Release note
```
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -553,6 +553,7 @@ where
 {
     fsm: &'a mut PeerFsm<EK, ER>,
     ctx: &'a mut PollContext<EK, ER, T>,
+    disk_full_opt: DiskFullOpt,
 }
 
 impl<'a, EK, ER, T: Transport> PeerFsmDelegate<'a, EK, ER, T>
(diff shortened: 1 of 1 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -585,6 +590,11 @@ where
                         .propose
                         .request_wait_time
                         .observe(duration_to_sec(cmd.send_time.saturating_elapsed()) as f64);
+                    // If cur cmd is set with special flag, then we need all the batch to be set with it.
+                    // Because, if not, then it will lead to ddl or other ops failure when disk is almost full.
+                    if cmd.extra_opts.disk_full_opt != DiskFullOpt::NotAllowedOnFull {
+                        self.disk_full_opt = cmd.extra_opts.disk_full_opt;
```
hicqu, line 596: Where can the flag be cleared? Seems it will be never cleared, right?

tier-cap:
PR to release-5.2 to performance test.

PR #10722, opened 2021-08-12
raftstore: fix compact_to for entry cache

gengliqi:

Problem Summary:
In the original code, the `idx` of `compact_to` function means all entries whose indexes are less than `idx` should be compacted.
However, the code below incorrectly uses `idx`.
https://github.com/tikv/tikv/blob/834ec1108d85623b2b19b00df2fa8e1359c726b6/components/raftstore/src/store/peer_storage.rs#L245
For example, if there are some `cached_entries` whose indexes are [0, 10), [10, 20). After this peer has applied the first `cached_entries` and call `compact_to` with `applied_index + 1`, which is 10. The second one will be removed because of the above code, leading to the unnecessary read io of raft db.

What's Changed:
Fix it and add comments for this function.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No

### Release note 

```release-note
None
```


hicqu (approved):

BusyJay:
There is a test `test_storage_cache_update` already. Can you update the case to cover this?

gengliqi:
> There is a test `test_storage_cache_update` already. Can you update the case to cover this?

I updated `test_storage_cache_size_change` test which is more suitable for the bug here. PTAL.

BusyJay (approved):

hicqu (approved):

PR #10724, opened 2021-08-13
raft_client: Lift the default value of max_grpc_send_msg_len to 64MB

### LINKED ISSUE #9714: batch messages can be too large in new raft client implementation
## Bug Report

New raft client implementation tends to batch more messages than before and we found it exceeded the max grpc messages in internal tests.

```
[2021/02/25 17:42:42.658 +08:00] [ERROR] [raft_client.rs:431] ["connection aborted"] [addr=] [receiver_err="Some(RpcFailure(RpcStatus { stat
us: 8-RESOURCE_EXHAUSTED, details: Some(\"Sent message larger than max (10705556 vs. 10485760)\") }))"] [sink_error=Some(RemoteStopped)] [store_id=7]
```

Calculating the message size using `Message::compute_size` should make it correct at the cost with extra CPU usage. We may also check what fields can be missed as even max batch size (128) is reached, there was about 1K extra size per message.

tonyxuqqi:
Issue Number: close #9714

### Problem Summary:

The RaftClient's way of estimating size is not accurate, partly because it does not count the field name into the size, also it does not fully check all the fields of RaftMessage to estimate the size. And thus the estimated size is not accurate in some scenarios and that leads to exceeding Grpc's own check which uses compute_size.
We prefer not call compute_size here simply because we tested in some cases it could lead to P99.99 latency goes up more than 10%.

### What's Changed:
Lift the default size of max_grpc_send_msg_len to 64MB. 

### Check List
Tests
    - Unittest
    - Manual test

### Release note 
None

/cc @BusyJay @gengliqi

### DIFF
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -43,7 +43,11 @@ const MIN_ENDPOINT_MAX_CONCURRENCY: usize = 4;
 
 const DEFAULT_SNAP_MAX_BYTES_PER_SEC: u64 = 100 * 1024 * 1024;
 
-const DEFAULT_MAX_GRPC_SEND_MSG_LEN: i32 = 10 * 1024 * 1024;
+// Lift default value to 64MB.
+// In fact the hard limit of sent message size may not be necessary.
+// Because we have the batch size as soft limit
+// https://github.com/tikv/tikv/issues/9714
+const DEFAULT_MAX_GRPC_SEND_MSG_LEN: i32 = 64 * 1024 * 1024;
 
 /// A clone of `grpc::CompressionAlgorithms` with serde supports.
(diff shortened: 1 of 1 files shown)


BusyJay (changes requested):
I would like to see how it performs when message size always exceeds the limits and metrics to show the cause if experiencing latency jitters.

BusyJay:
As new comments in #9714, I think the source of error is found. I suggest to also count all missing bytes length and leave size limit untouched in this PR.

In the long term, I agree we should set the message size to unlimited. But before that, we need a way to observe the roundtrip latency and the size of raft messages. As when the message size is unlimited, similar problem can cause mysterious high commit duration that nobody can reason about.

PR #10735, opened 2021-08-15
*: fix Clippy warnings about rule new_without_default

### LINKED ISSUE #10734: Fix Clippy warnings about rule new_without_default
## Development Task
This is a subtask of [Fix Clippy warnings ](https://github.com/tikv/tikv/issues/8336) and this PR only care about the new_without_default issues, more about rule new_without_default see https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default

yuqi1129:

Issue Number: close #10734  

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note

```release-note
None
```


tisonkun:
@yuqi1129 thanks for your contribution @yuqi1129 ! I'd like to know how can we prevent further development break this rule? I don't see a new ci checker introduced/enabled in this PR.

yuqi1129:
@yuqi1129  Nice point, I would like to take some steps to introduce ci checker to complete this PR, thanks

yuqi1129:
@tisonkun Nice point, I would like to take some steps to introduce ci checker to complete this PR, thanks

sticnarf:
This lint warning is explicit suppressed by https://github.com/tikv/tikv/blob/master/scripts/clippy#L20.

Personally, this style recommendation is not always necessary. Anyway, it is nice to have them all fixed and the line https://github.com/tikv/tikv/blob/master/scripts/clippy#L20 can be removed now.

yuqi1129:
@sticnarf , Thanks for your suggestions, i will take your advice~

skyzh:
It seems that clippy is not taking effect on CI, as found by https://github.com/tikv/tikv/pull/10638#discussion_r689405265

skyzh:
> This lint warning is explicitly suppressed by https://github.com/tikv/tikv/blob/master/scripts/clippy#L20.
> 
> Personally, this style recommendation is not always necessary. Anyway, it is nice to have them all fixed and the line https://github.com/tikv/tikv/blob/master/scripts/clippy#L20 can be removed now.

+1 for this. While I don't have a strong opinion on enable or disable this check, I think we could retain `XX::new()` instead of changing them to `XX::default()`.

yuqi1129:
@skyzh , This PR only adds the XX::default() and do not replace `new()` with `default()`, so the method of `default()` has not been called so far,  `XX:default` has no use but only fix the clippy warnings so is it necessary to continue working on this issue ?

components/raftstore/src/store/worker/read.rs
```diff
@@ -390,7 +396,7 @@ where
     E: KvEngine,
 {
     pub fn new(kv_engine: E, store_meta: Arc<Mutex<StoreMeta>>, router: C) -> Self {
-        let cache_read_id = ThreadReadId::new();
+        let cache_read_id = ThreadReadId::default();
```
skyzh, line 399: I mean this part. Does clippy requires the use of “default” instead of “new”?

yuqi1129: No, this have nothing to do with clippy requirement, If there is no need to make any changes here , i will restore the logic

sticnarf:
@yuqi1129 Please fix the same problem for `metrics::threads_linux::ThreadInfoStatistics` as the CI instructs: https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/6072/pipeline

Since this is Linux-specific code, if you are developing on macOS, you might not encounter this after running `make clippy`.

yuqi1129:
@sticnarf OK，i just merge the master branch to this branch and do not expected there exists the this problem within 10 commits

skyzh (commented):
Rest LGTM. Please fix all clippy errors, see CI log for more information.

components/test_pd/src/mocker/retry.rs
```diff
@@ -36,6 +36,12 @@ impl Retry {
     }
 }
 
+impl Default for Retry {
+    fn default() -> Self {
+        Self::new(usize::default())
```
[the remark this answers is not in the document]
yuqi1129: OK

tisonkun:
/cc @skyzh @sticnarf 

could you give it another look?

skyzh (approved):
Good work!

components/test_pd/src/mocker/retry.rs
```diff
@@ -36,6 +36,12 @@ impl Retry {
     }
 }
 
+impl Default for Retry {
+    fn default() -> Self {
+        Self::new(0)
+    }
+}
+
```
sticnarf, line 44: This `Deafult` implementation looks questionable to me. If the user does not need a retry, they should use `NotRetry`. "Retry 0 times" seems not a good default...

And because `Retry::new` takes an argument, clippy should not complain about it. I prefer to just remove this implementation.

yuqi1129: > And because Retry::new takes an argument, clippy should not complain about it. I prefer to just remove this implementation.

You mean we can just let clippy omit this implementation?

> This Deafult implementation looks questionable to me. If the user does not need a retry, they should use NotRetry. "Retry 0 times" seems not a good default
Shall we change 0 to a reasonable value ?

skyzh: Will clippy complain if we don't implement `Default` for `Retry`?

sticnarf: > You mean we can just let clippy omit this implementation?

Were you complained by clippy about missing default for `Retry`? I just remove this implementation and clippy does not give a warning.

yuqi1129: Yes, Clippy will output warning information. i wander if exists some way to make this method as an exception.

skyzh:
```
#[allow(clippy::xxx)]
```

If clippy still gives a warning, use the above marker to ignore check for a single line / struct.

sticnarf (approved):
LGTM

PR #10743, opened 2021-08-16
[DNM] hotfix

ekexium:
Signed-off-by: ekexium <ekexium@gmail.com>

Issue Number: close #xxx 

Problem Summary:

Add a field in the `ScanResponse` to indicate whether there are locks to improve TiSpark's performance.

Proposal: [xxx](url) 

What's Changed:

### Related changes

- [ ] https://github.com/pingcap/kvproto/pull/792

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/src/server/service/kv.rs
+++ b/src/server/service/kv.rs
@@ -1392,11 +1392,15 @@ fn future_scan<E: Engine, L: LockMgr>(
             if let Some(err) = extract_region_error(&v) {
                 resp.set_region_error(err);
             } else {
-                resp.set_has_locks(
-                    v.as_ref()
-                        .map(|v| v.iter().any(|r| r.is_err()))
-                        .unwrap_or(true),
-                );
+                let first_lock = v.as_ref().map(|v| v.iter().find(|&r| r.is_err()));
+                let first_lock = match first_lock {
+                    Ok(o) => o.and_then(|r| r.as_ref().err()),
+                    Err(e) => Some(e),
+                };
+                if let Some(l) = first_lock {
+                    resp.set_error(extract_key_error(l));
+                }
+
                 resp.set_pairs(RepeatedField::from_vec(extract_kv_pairs(v)));
             }
             Ok(resp)
--- a/src/server/service/kv.rs
+++ b/src/server/service/kv.rs
@@ -1392,6 +1392,11 @@ fn future_scan<E: Engine, L: LockMgr>(
             if let Some(err) = extract_region_error(&v) {
                 resp.set_region_error(err);
             } else {
+                resp.set_has_locks(
+                    v.as_ref()
+                        .map(|v| v.iter().any(|r| r.is_err()))
+                        .unwrap_or(true),
+                );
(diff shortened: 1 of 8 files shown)


src/server/service/kv.rs
```diff
@@ -1392,6 +1392,11 @@ fn future_scan<E: Engine, L: LockMgr>(
             if let Some(err) = extract_region_error(&v) {
                 resp.set_region_error(err);
             } else {
+                resp.set_has_locks(
```
cfzjywxk, line 1395: This may introduce the cost scanning the result vector another time which the `tispark` client wants to avoid. 

The current `tispark` scan process is like this:
- 1.Send the `ScanRequest` to tikv.
- 2.Get the kv pairs in the `ScanResponse`.
- 3.Loop through all the kv pairs, collect the locks if there is.
- 4.Batch resolve all the encountered locks, that is send `BatchResolveRequest` to tikv.
- 5.Retry the `ScanRequest` until there is no lock error.

The step 3 may introduce much cost for every scan request which hurts the performance a lot. So we may need to collect all the locks in the response, so if the `len(locks) > 0` the tispark could try to batch resolve these locks directly using the information from the `ScanResponse` and step 3 could be avoided.

If the `tispark` server supports triggering resolve the locks for a whole region like `tidb`, we could break the tikv scanner if there is a lock encountered directly and return the lock error in the response. But seems tispark only supports batch resolve by now.

birdstorm: > This may introduce the cost scanning the result vector another time which the `tispark` client wants to avoid. 
> 
> The current `tispark` scan process is like this:
> - 1.Send the `ScanRequest` to tikv.
> - 2.Get the kv pairs in the `ScanResponse`.
> - 3.Loop through all the kv pairs, collect the locks if there is.
> - 4.Batch resolve all the encountered locks, that is send `BatchResolveRequest` to tikv.
> - 5.Retry the `ScanRequest` until there is no lock error.
> 
> The step 3 may introduce much cost for every scan request which hurts the performance a lot. So we may need to collect all the locks in the response, so if the `len(locks) > 0` the tispark could try to batch resolve these locks directly using the information from the `ScanResponse` and step 3 could be avoided.
> 
> If the `tispark` server supports triggering resolve the locks for a whole region like `tidb`, we could break the tikv scanner if there is a lock encountered directly and return the lock error in the response. But seems tispark only supports batch resolve by now.

thx for your explanation. I believe it is a right way not to break the tikv scanner here. It will not break the resolve lock logic for elder versions of tidb and tispark.

On the other hand, while returning all locks in response, it's still possible to implement the tispark logic to trigger a region-wise lock resolver if necessary.

ekexium: > This may introduce the cost scanning the result vector another time which the tispark client wants to avoid.

Returning the first lock from the scanner is also O(N), I think it won't make much difference?

PR #10746, opened 2021-08-16
raftstore: separate read write ready (#10592)

### LINKED ISSUE #10475: Separate read/write ready

BusyJay:

Close #10475.

Read ready can be processed early to reduce read latency.

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
- separate read write ready to reduce read latency
```

### DIFF
--- a/components/raft_log_engine/src/engine.rs
+++ b/components/raft_log_engine/src/engine.rs
@@ -74,8 +74,12 @@ impl RaftLogBatchTrait for RaftLogBatch {
         Ok(())
     }
 
+    #[inline]
     fn persist_size(&self) -> usize {
-        panic!("persist_size is not implemented for raft engine");
+        // TODO: This is not data size, but sufficient for current usage.
+        // Data size is useful when controlling bytes per sync, so it's better
+        // to implement it correctly.
+        self.0.items.len()
     }
 
     fn is_empty(&self) -> bool {
--- a/components/raftstore/src/store/fsm/store.rs
+++ b/components/raftstore/src/store/fsm/store.rs
@@ -11,7 +11,8 @@ use std::time::{Duration, Instant};
 use std::{mem, u64};
 
 use batch_system::{
-    BasicMailbox, BatchRouter, BatchSystem, Fsm, HandlerBuilder, PollHandler, Priority,
+    BasicMailbox, BatchRouter, BatchSystem, Fsm, HandleResult, HandlerBuilder, PollHandler,
+    Priority, TrackedFsm,
 };
 use crossbeam::channel::{TryRecvError, TrySendError};
 use engine_traits::PerfContext;
@@ -347,8 +348,6 @@ where
     EK: KvEngine,
     ER: RaftEngine,
 {
-    /// The count of processed normal Fsm.
-    pub processed_fsm_count: usize,
     pub cfg: Config,
     pub store: metapb::Store,
     pub pd_scheduler: Scheduler<PdTask<EK>>,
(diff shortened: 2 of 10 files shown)


components/raftstore/src/store/fsm/store.rs
```diff
@@ -877,27 +905,46 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
+            peer.update_memory_trace(&mut self.trace_event);
+        }
+        MEMTRACE_PEERS.trace(mem::take(&mut self.trace_event));
         if self.poll_ctx.has_ready {
-            self.handle_raft_ready(peers);
+            self.handle_raft_ready(peers, to_skip_end);
+        }
+    }
+
+    fn end(&mut self, peers: &mut [Option<impl TrackedFsm<Target = PeerFsm<EK, ER>>>]) {
```
BusyJay, line 945: @5kbpers PTAL, I'm not sure if end is the correct place for pending latency inspector after introducing `light_end`.

gengliqi (approved):

components/raft_log_engine/src/engine.rs
```diff
@@ -74,8 +74,12 @@ impl RaftLogBatchTrait for RaftLogBatch {
         Ok(())
     }
 
+    #[inline]
     fn persist_size(&self) -> usize {
-        panic!("persist_size is not implemented for raft engine");
+        // TODO: This is not data size, but sufficient for current usage.
```
BusyJay, line 79: /cc @tabokie as a reminder.

NingLin-P (approved):

components/raftstore/src/store/fsm/store.rs
5kbpers: The purpose of inspectors is to inspect write latency so I think it's ok.

components/raft_log_engine/src/engine.rs
tabokie: It was implemented in https://github.com/tikv/raft-engine/pull/82 a few days ago.

NingLin-P (approved):

PR #10768, opened 2021-08-17
storage: Flow controller consumes quota only when write succeeds

Connor1996:

Issue Number: close #xxx 

Problem Summary:
When there are some errors(e.g. EpochNotMatch) returned in raftstore, the failed write requests consume the flow controller quota as usual. So normal writes maybe delayed for a long time leading to QPS drop.

image

image

Proposal: [xxx](url) 

What's Changed:
Consume flow controller quota and delay in the callback of write finished.

### Related changes

### Check List 

Tests 

- Manual test

### Release note 

```release-note
None
```


src/storage/txn/scheduler.rs
```diff
@@ -845,25 +847,39 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
-                        }
+                    if self.inner.flow_controller.enabled()
+                        && !self.inner.flow_controller.is_unlimited()
+                    {
+                        let start = Instant::now_coarse();
+                        // Control mutex is used to ensure there is only one request consuming the budget.
+                        // The delay may exceed 1s, and the speed limit is changed every second.
+                        // If the speed of next second is larger than the one of first second,
+                        // without the mutex, the write flow can't throttled strictly.
+                        let _guard = self.control_mutex.lock().await;
```
yiwu-arbug, line 881: style: shall we just wrapping `flow_controller` with a mutex, so it is more clear the mutex is for the flow controller, and the logic being more clear that we are adding delay here for the sake of flow control?

src/storage/txn/scheduler.rs
```diff
@@ -845,25 +847,39 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+                                            // no need to delay if unthrottled, just call consume to record write flow
+                                            let _ = sched.inner.flow_controller.consume(write_size);
+                                        } else {
+                                            // Control mutex is used to ensure there is only one request consuming the budget.
+                                            // The delay may exceed 1s, and the speed limit is changed every second.
+                                            // If the speed of next second is larger than the one of first second,
+                                            // without the mutex, the write flow can't throttled strictly.
+                                            let _guard = sched.control_mutex.lock().await;
+                                            let delay =
+                                                sched.inner.flow_controller.consume(write_size);
```
yiwu-arbug, line 864: comment why we need to consume budget in the finish callback, and relay the delay to the other writes via the mutex lock.

yiwu-arbug (approved):

src/storage/txn/scheduler.rs
tabokie: That's would introduce overhead when flow control is disabled, plus we don't need the mutex for discarding request. Maybe put the `control_mutex` inside flow_controller?

tabokie: Actually, why still lock the mutex here?

tabokie (approved):
LGTM

src/storage/txn/scheduler.rs
Connor1996: I want to move `control_mutex ` inside flow_controller too. But `Arc<SchedulerInner>` is not `Send`, to solve it may need some changes on other structs. 
```
error: future cannot be sent between threads safely
   --> src/storage/txn/scheduler.rs:832:30
    |
832 | ...                   .spawn(async move {
    |                        ^^^^^ future created by async block is not `Send`
    |
note: future is not `Send` as this value is used across an await
   --> src/storage/txn/scheduler.rs:861:58
    |
861 | ...                   let _guard = sched.inner.flow_controller.control_mutex.lock().await;
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ first, await occurs here, with `sched.inner` maybe used later...
note: `sched.inner` is later dropped here
   --> src/storage/txn/scheduler.rs:861:112
    |
861 | ...                   let _guard = sched.inner.flow_controller.control_mutex.lock().await;
    |                                    -----------                                           ^
    |                                    |
    |                                    has type `std::sync::Arc<SchedulerInner<L>>` which is not `Send`
help: consider further restricting this bound
    |
287 | impl<E: Engine, L: LockManager + std::marker::Sync> Scheduler<E, L> {
```

PR #10771, opened 2021-08-18
storage: spawn a background task to return deadline error

sticnarf:

It is important to respond to the client in time. Otherwise, the client is hard to distinguish a down TiKV from a busy TiKV. And lost response may cause more "undetermined errors" at the client.

Although we have added deadline checks at many places after latches are all acquired, there are still various possibilities for a command to stuck at the scheduler, such as blocked by the latch of another throttled command.

So this PR hopes to reduce the chance of client recv timeout as much as possible.

When scheduling a command, spawn a task that calls the callback of the command if it expires. An extra flag `owned` is added to make sure the callback is not called twice.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Respond with deadline error even if the command is waiting in the scheduler.
```


youjiali1995 (commented):
I wonder its performance impact.

sticnarf:
> I wonder its performance impact.

It depends on how many conflicts there are. If all latches are acquired successfully at first, the background task is not started.

For conflicted scenarios, I test sysbench zipfian (exp=0.99) as an example:

photo_2021-08-23_15-59-20

The global timer increases about 0.4% CPU and spawning the background task increases 0.4% CPU.

The final QPS and latency impact is hardly noticeable.

src/storage/txn/scheduler.rs
```diff
@@ -874,9 +937,19 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
                     if self.inner.flow_controller.enabled()
                         && !self.inner.flow_controller.is_unlimited()
                     {
+                        let mut deadline_fut = GLOBAL_TIMER_HANDLE
+                            .delay(deadline.to_std_instant())
+                            .compat()
+                            .fuse();
```
youjiali1995, line 943: Just curious, why need `fuse` here?

youjiali1995 (approved):

src/storage/txn/scheduler.rs
sticnarf: It's a requirement of the `select!` macro.
https://docs.rs/futures/0.3.16/futures/macro.select.html

youjiali1995 (commented):
It may not be able to cancel `ResolveLock`, because the deadline is refreshed for each sub command. It's better to add comments for it.

Issue #10736, opened 2021-08-16
Report CPU usage to PD with the heartbeat


nolouch:
## Backgroup
After https://github.com/tikv/tikv/pull/10155, we can collect the CPU usage with specific a tag. here is required to record the CPU usage for each region and report to PD.  It's helpful for PD schedule strategy, and also we can add a CPU dimension
for key visual.

## Development Task

- [ ] collect CPU usage for each region
- [ ] report to PD with the heartbeat, add CPU dimension for hot scheduler

nolouch:
@JmPotato would you like to take a look?

JmPotato:
> @JmPotato would you like to take a look?

I will take this task. /cc JmPotato

PR #10789, opened 2021-08-20
raftstore: collect and report the CPU time usage of regions to PD

JmPotato:

Part of https://github.com/tikv/tikv/issues/10736. Collect the CPU usage of regions to support reporting related information to the PD in the future.

* Implement `Collector` for the PD runner.
* Calculate `cpu_usage` since the last heartbeat for each region.

`cpu_usage` is the CPU time usage of the leader region since the last heartbeat, which is calculated by cpu_time_delta/heartbeat_reported_interval, it reflects the CPU usage of the region from the workload during the heartbeat interval.

### Related changes

- **[Need Review First]** kvproto change: https://github.com/pingcap/kvproto/pull/798
- Pre PR: https://github.com/tikv/tikv/pull/10155

### Check List 

Tests 

- Unit test
- Manual test

image

I modified a version of PD to log out the CPU time usage info from every region heartbeat. The region `cpu_usage` is increasing after 17:48:12 while I was running a sysbench insert workload.

### Release note 

```release-note
None.
```

### DIFF
--- a/components/pd_client/src/client.rs
+++ b/components/pd_client/src/client.rs
@@ -530,6 +530,7 @@ impl PdClient for RpcClient {
         req.set_query_stats(region_stat.query_stats);
         req.set_approximate_size(region_stat.approximate_size);
         req.set_approximate_keys(region_stat.approximate_keys);
+        req.set_cpu_time_ms(region_stat.cpu_time_ms);
         if let Some(s) = replication_status {
             req.set_replication_status(s);
         }
--- a/components/pd_client/src/client.rs
+++ b/components/pd_client/src/client.rs
@@ -530,7 +530,7 @@ impl PdClient for RpcClient {
         req.set_query_stats(region_stat.query_stats);
         req.set_approximate_size(region_stat.approximate_size);
         req.set_approximate_keys(region_stat.approximate_keys);
-        req.set_cpu_time_ms(region_stat.cpu_time_ms);
+        req.set_cpu_usage(region_stat.cpu_usage);
         if let Some(s) = replication_status {
             req.set_replication_status(s);
         }
--- a/components/raftstore/src/store/worker/pd.rs
+++ b/components/raftstore/src/store/worker/pd.rs
@@ -1347,8 +1347,7 @@ fn calculate_region_cpu_records(
 ) {
     for (tag, ms) in &records.records {
         let record_store_id = tag.infos.store_id;
-        let extra_attachment = &tag.infos.extra_attachment;
-        if record_store_id != store_id || extra_attachment.is_empty() {
+        if record_store_id != store_id {
             continue;
         }
 
@@ -1485,10 +1484,7 @@ where
                     peer_stat.last_region_report_read_query_stats =
                         peer_stat.read_query_stats.clone();
                     // Take out the region CPU record.
-                    let cpu_time_ms = match self.region_cpu_records.remove(&region_id) {
-                        Some(total_cpu_time_ms) => total_cpu_time_ms,
-                        None => 0,
-                    };
+                    let cpu_time_ms = self.region_cpu_records.remove(&region_id).unwrap_or(0);
 
                     let mut last_report_ts = peer_stat.last_region_report_ts;
                     peer_stat.last_region_report_ts = UnixSecs::now();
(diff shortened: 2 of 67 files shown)


components/raftstore/src/store/worker/pd.rs
```diff
@@ -1290,6 +1334,43 @@ where
...
+            continue;
+        }
+
+        let ms = *ms as u32;
+        match region_cpu_records.get_mut(region_id) {
+            Some((ts, cpu_time, total)) => {
+                if *ts.last().unwrap() == timestamp_secs {
+                    *cpu_time.last_mut().unwrap() += ms;
+                } else {
+                    ts.push(timestamp_secs);
```
nolouch, line 1364: Does this will have too many entries and cost high memory?

JmPotato: It will be cleaned up every time we report it through the region heartbeat. Maybe GC logic is needed also? WDYT?

JmPotato: > Does this will have too many entries and cost high memory?

I think we can maintain a top-k map here rather than saving them all.

nolouch: Seems every second will have an entry. We only care about the tags of `region_id`, so we can sum them up in the heartbeat interval instead of storing them all.

JmPotato: Now it will only store the total CPU time since the last region heartbeat. PTAL.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1290,6 +1334,35 @@ where
...
+
+fn calculate_region_cpu_records(
+    store_id: u64,
+    records: Arc<CpuRecords>,
+    region_cpu_records: &mut HashMap<u64, u32>,
+) {
+    for (tag, ms) in &records.records {
+        let record_store_id = tag.infos.store_id;
+        let extra_attachment = &tag.infos.extra_attachment;
+        if record_store_id != store_id || extra_attachment.is_empty() {
```
zhongzc, line 1351: `extra_attachment` is used by TiDB to embed SQL information.

Should PD care about whether the CPU load comes from SQL requests? If not, I think it doesn't need to skip these records.

nolouch:
Could you test it in a real environment?

JmPotato:
> Could you test it in a real environment?

OK. I will post some test results later.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1290,6 +1334,34 @@ where
...
+        let region_id = &tag.infos.region_id;
+        let ms = *ms as u32;
+        match region_cpu_records.get_mut(region_id) {
+            Some(total) => {
+                *total += ms;
+            }
+            None => {
+                region_cpu_records.insert(*region_id, ms);
+            }
+        }
```
zhongzc, line 1363: Can be simplified as

`*map.entry(key).or_insert(0) += value;`

JmPotato:
@nolouch I added a simple test result to the description. PTAL.

components/pd_client/src/client.rs
```diff
@@ -530,6 +530,7 @@ impl PdClient for RpcClient {
         req.set_query_stats(region_stat.query_stats);
         req.set_approximate_size(region_stat.approximate_size);
         req.set_approximate_keys(region_stat.approximate_keys);
+        req.set_cpu_time_ms(region_stat.cpu_time_ms);
```
nolouch, line 533: I wonder the unit `cpu time ms` is different with the unit of the store heartbeat `CpuUsages`. better to unify them.

JmPotato: Done. PTAL.

nolouch (commented):
rest LGTM

nolouch (approved):
lgtm

zhongzc:
The recording of cpu time depends on the [resource-metering.enabled] configuration. What do you think? Will it be enabled by default in the future?

nolouch:
No, I think once it is enabled in the feature, resource-metering should be enabled in default.

nolouch:
BTW, If `resource-metering.enable=false`, will it report with nothing?

zhongzc:
> BTW, If `resource-metering.enable=false`, will it report with nothing?

Yes. Now it is false, means you will see `cpu_usage` always be `0` unless you turn it on via config.

zhongzc (approved):

PR #10800, opened 2021-08-23
storage: ban txn cmds when TTL is enabled.

youjiali1995:

Issue Number: close https://github.com/pingcap/tidb/issues/27303

Problem Summary:

The TTL feature can break transaction atomicity so we can't run transaction commands with TTL enabled.

What's Changed:

Returns an error if runs a transaction command with TTL enabled.

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -938,6 +938,10 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
 
         let cmd: Command = cmd.into();
 
+        if self.enable_ttl && !cmd.support_ttl() {
+            return Err(box_err!("can't sched cmd({}) with TTL enabled", cmd.tag()));
+        }
+
         match &cmd {
             Command::Prewrite(Prewrite { mutations, .. }) => {
                 check_key_size!(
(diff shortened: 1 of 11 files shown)


src/storage/mod.rs
```diff
@@ -6909,4 +6913,70 @@ mod tests {
...
+    fn test_txn_cmd_with_ttl_enabled() {
+        let storage = TestStorageBuilder::new(DummyLockManager {}, true)
+            .build()
+            .unwrap();
+        let (tx, rx) = channel();
+
+        // TTL commands can pass the check.
+        storage
+            .raw_batch_put(
+                Context::default(),
```
sticnarf, line 6927: The `for_cas` field should be enabled if you want it to be a scheduler command.

src/storage/mod.rs
```diff
@@ -6909,4 +6913,70 @@ mod tests {
...
+                "".to_string(),
+                vec![(b"k".to_vec(), b"v".to_vec())],
+                10,
+                expect_ok_callback(tx.clone(), 0),
+            )
+            .unwrap();
+        rx.recv().unwrap();
+        storage
+            .raw_batch_delete(
+                Context::default(),
```
sticnarf, line 6937: ditto

src/storage/mod.rs
youjiali1995: I missed  the `_atomic` suffix...

sticnarf (approved):
LGTM

andylokandy:
I'd rather like to ban txn requests on the service layer because the [API V2](https://github.com/tikv/rfcs/pull/69) is going to do so.

youjiali1995:
> I'd rather like to ban txn requests on the service layer because the [API V2](https://github.com/tikv/rfcs/pull/69) is going to do so.

It's a long term goal. I think we can forbid it in this way for now. @andylokandy

andylokandy:
@youjiali1995 You may get me wrong. I mean we can ban the txn requests in `service.rs` instead of in storage.

cfzjywxk (approved):
LGTM

youjiali1995:
> @youjiali1995 You may get me wrong. I mean we can ban the txn requests in `service.rs` instead of in storage.

@andylokandy Do you mean in `service/kv.rs`, moreover, the `txn_command_future` macro? It's verbose to do it because different response types have different error fields, part of them have `errors` field, other have `error` field.

andylokandy:
@youjiali1995 You can add a light wrapper(a function or a closure) to check the mode before executing the storage command. Mixing the mode check into one specific mode make (txn) will make later life harder.

youjiali1995:
@andylokandy @sticnarf PTAL again.

sticnarf (commented):
LGTM

andylokandy (approved):

PR #10825, opened 2021-08-25
*: update backtrace-rs to 0.3.61

sticnarf:

Previously backtrace-rs uses https://github.com/ianlancetaylor/libbacktrace for resolving symbols from debug info. But libbacktrace had poor support for DWARF 5.

Although building TiKV itself does not generate DWARF 5 by default, if the dynamic libraries provided by the system like libpthread are built with DWARF 5, libbacktrace will fail to resolve correct symbols.

This may affect generating backtrace when TiKV panics and profiling with pprof-rs.

Case 1: Fail to print function symbols in backtrace:
图片

Case 2: Fail to profile with pprof-rs
图片

This PR upgrades backtrace-rs. The new version of backtrace-rs uses gimli for resolving symbols. It supports DWARF 5 well. After this upgradation, the above two issues are resolved. 

### Check List 

Tests 

### Release note 

```release-note
Support dynamic libraries built with DWARF 5.
```


sticnarf:
cc @YangKeao

YangKeao (approved):
LGTM

BusyJay:
Any side effects?

sticnarf:
> Any side effects?

Gimli has been used by rustc for one year. And backtrace-rs has switched from libbacktrace to gimli by default for more than one year. So I think symbolizing with gimli is stable.

And I also test the performance impact of profiling with 99Hz is too small to see (< 1%). My workload is sysbench select_random_ranges, which can make TiKV run out my CPU resource.

5kbpers (approved):

PR #10831, opened 2021-08-25
raftstore: continue GC  snapshot files when it encounters a corrupted snapshot

### LINKED ISSUE #10813: corrupted snapshot can prevent snapshot files gc
v3.0.8

Doesn't matter.

### Steps to reproduce
Some of snapshot files get deleted after received, so the snapshot is corrupted. And the peer is deleted before applying snapshot.

### What did you expect?
Corrupted snapshot will be deleted in the end.

### What did happened?
Store fails to delete the corrupted snapshot files. And all other stale snapshot files are not gc anymore.

The problem is it returns early here:

https://github.com/tikv/tikv/blob/master/components/raftstore/src/store/fsm/store.rs#L2129.

tonyxuqqi:

Issue Number: close #10813 

Problem Summary:

When snapshot GC runs into problem (such as file corrupted), today it stopped processing other GC-able snapshot files.
This PR will make it continue processing rest of the snapshot files.

### Related changes

Tests 
- Unit test

Side effects
No.

### Release note 

```release-note
RaftStore Snapshot GC fix: fix the issue that snapshot GC missed GC snapshot files when there's one snapshot file failed to be GC-ed. 
```
/cc @BusyJay

### DIFF
--- a/tests/failpoints/cases/test_snap.rs
+++ b/tests/failpoints/cases/test_snap.rs
@@ -6,6 +6,9 @@ use std::time::Duration;
 use std::{fs, io, thread};
 
 use raft::eraftpb::MessageType;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::PathBuf;
 use test_raftstore::*;
 use tikv_util::config::*;
 use tikv_util::time::Instant;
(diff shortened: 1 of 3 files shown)


tests/failpoints/cases/test_snap.rs
```diff
@@ -467,3 +470,73 @@ fn test_cancel_snapshot_generating() {
...
+    loop {
+        let snap_keys = cluster.get_snap_mgr(3).list_idle_snap().unwrap();
+        if snap_keys.is_empty() {
+            panic!("no snapshot file is found");
+        }
+
+        let mut found_unexpected_file = false;
+        let mut found_expected_file = false;
+        for (snap_key, _is_sending) in snap_keys {
+            if snap_key.region_id == 2 && snap_key.idx == 1 {
```
hicqu, line 518: It's so strange that the case can pass. I thought `run_conf_change` creates a region with ID 1 instead of 2.

hicqu:
Rest LGTM.

hicqu (approved):

hicqu (approved):

tests/failpoints/cases/test_snap.rs
tonyxuqqi: The reason to have region_id 2 is because we want to make "match self.ctx.router.send(region_id, gc_snap)" fail. And the way to make it fail is to have a non-exist region_id. That's why we have region_id 2 here instead of 1.

BusyJay (approved):

PR #10837, opened 2021-08-26
tikv-ctl: Migrate clap to structopt

### LINKED ISSUE #10793: Use StructOpt in tikv-ctl
## Development Task

Refactor `https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs` by using [`structopt`](https://docs.rs/structopt/0.3.22/structopt/#subcommands).

peng1999:

Issue Number: close #10793

What's Changed:

Add `structopt` as a dependency, for parsing command line options of `tikv-ctl`.

### Check List 

Tests 

  - I tried my best to ensure CLI behaviour keep the same after the migration. Most codes are transformed literally.
  - The `--help` output for every subcommand is printed and compared between clap version and structopt version, so that we can ensure we have the same number of arguments/flags with flag types unchanged.
  - I tested most command in the [`tikv-ctl` part of TiKV documention](https://tikv.org/docs/dev/reference/tools/tikv-ctl/) manually. But supported flags are too many. I am not able to test every possible flag/argument combination to ensure correctness.

### Release note 

```release-note
None
```

### DIFF
--- a/cmd/tikv-ctl/src/main.rs
+++ b/cmd/tikv-ctl/src/main.rs
@@ -1123,7 +1123,7 @@ static VERSION_INFO: SyncLazy<String> = SyncLazy::new(|| {
     tikv::tikv_version_info(build_timestamp)
 });
 
-const RAW_KEY_HINT: &'static str = "Raw key (generally starts with \"z\") in escaped form";
+const RAW_KEY_HINT: &str = "Raw key (generally starts with \"z\") in escaped form";
 
 #[derive(StructOpt)]
 #[structopt(
--- a/cmd/tikv-ctl/src/main.rs
+++ b/cmd/tikv-ctl/src/main.rs
@@ -1210,7 +1210,7 @@ struct Opt {
     encode: Option<String>,
 
     #[structopt(subcommand)]
-    cmd: Cmd,
+    cmd: Option<Cmd>,
 }
 
 #[derive(StructOpt)]
@@ -1825,8 +1825,30 @@ fn main() {
     );
     let mgr = new_security_mgr(&opt);
 
+    let cmd = match opt.cmd {
+        Some(cmd) => cmd,
+        None => {
+            // Deal with arguments about key utils.
+            if let Some(hex) = opt.hex_to_escaped.as_deref() {
+                println!("{}", escape(&from_hex(hex).unwrap()));
+            } else if let Some(escaped) = opt.escaped_to_hex.as_deref() {
+                println!("{}", log_wrappers::hex_encode_upper(unescape(escaped)));
+            } else if let Some(encoded) = opt.decode.as_deref() {
+                match Key::from_encoded(unescape(encoded)).into_raw() {
+                    Ok(k) => println!("{}", escape(&k)),
+                    Err(e) => println!("decode meets error: {}", e),
+                };
+            } else if let Some(decoded) = opt.encode.as_deref() {
+                println!("{}", Key::from_raw(&unescape(decoded)));
+            } else {
+                Opt::clap().print_help().ok();
+            }
+            return;
+        }
+    };
+
     // Bypass the ldb and sst dump command to RocksDB.
-    if let Cmd::External(args) = opt.cmd {
+    if let Cmd::External(args) = cmd {
         match args[0].as_str() {
             "ldb" => run_ldb_command(args, &cfg),
             "sst_dump" => run_sst_dump_command(args, &cfg),
@@ -1835,7 +1857,7 @@ fn main() {
         return;
     }
 
-    if let Cmd::BadSsts { db, manifest, pd } = opt.cmd {
+    if let Cmd::BadSsts { db, manifest, pd } = cmd {
         let pd_client = get_pd_rpc_client(&pd, Arc::clone(&mgr));
         print_bad_ssts(&db, manifest.as_deref(), pd_client, &cfg);
         return;
@@ -1843,7 +1865,7 @@ fn main() {
 
     // Deal with subcommand dump-snap-meta. This subcommand doesn't require other args, so process
     // it before checking args.
-    if let Cmd::DumpSnapMeta { file } = opt.cmd {
+    if let Cmd::DumpSnapMeta { file } = cmd {
         let path = file.as_ref();
         return dump_snap_meta_file(path);
     }
@@ -1866,7 +1888,7 @@ fn main() {
         return;
     }
 
-    if let Cmd::DecryptFile { file, out_file } = opt.cmd {
+    if let Cmd::DecryptFile { file, out_file } = cmd {
         let message = "This action will expose sensitive data as plaintext on persistent storage";
         if !warning_prompt(message) {
             return;
@@ -1916,7 +1938,7 @@ fn main() {
         return;
     }
 
-    if let Cmd::EncryptionMeta { cmd: subcmd } = opt.cmd {
+    if let Cmd::EncryptionMeta { cmd: subcmd } = cmd {
         match subcmd {
             EncryptionMetaCmd::DumpKey { ids } => {
                 let message = "This action will expose encryption key(s) as plaintext. Do not output the \
@@ -1951,7 +1973,7 @@ fn main() {
             to,
             threads,
             bottommost,
-        } = opt.cmd
+        } = cmd
         {
             let db_type = if db == "kv" { DBType::Kv } else { DBType::Raft };
             let cfs = cf.iter().map(|s| s.as_ref()).collect();
@@ -1965,7 +1987,7 @@ fn main() {
         if let Cmd::SplitRegion {
             region: region_id,
             key,
-        } = opt.cmd
+        } = cmd
         {
             let key = unescape(&key);
             return split_region(&pd_client, mgr, region_id, key);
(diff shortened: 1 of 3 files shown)


cmd/tikv-ctl/src/main.rs
```diff
@@ -1050,867 +1051,661 @@ fn warning_prompt(message: &str) -> bool {
     }
 }
 
-fn main() {
-    let raw_key_hint: &'static str = "Raw key (generally starts with \"z\") in escaped form";
+static VERSION_INFO: SyncLazy<String> = SyncLazy::new(|| {
```
skyzh, line 1054: Is it required to use `SyncLazy`? Can we do something like `version = get_version_info()`?

cmd/tikv-ctl/src/main.rs
```diff
@@ -1050,867 +1051,661 @@ fn warning_prompt(message: &str) -> bool {
...
+    name = "TiKV Control (tikv-ctl)",
+    about = "A tool for interacting with TiKV deployments.",
+    author = crate_authors!(),
+    version = &**VERSION_INFO,
+    long_version = &**VERSION_INFO,
+    setting = AppSettings::DontCollapseArgsInUsage,
+)]
+struct Opt {
+    #[structopt(long)]
+    /// Set the address of pd
```
skyzh, line 1072: Maybe we should use `#[structopt(short, help = "xxx")]`, otherwise the help message won't show in command line.

cmd/tikv-ctl/src/main.rs
```diff
@@ -1050,867 +1051,661 @@ fn warning_prompt(message: &str) -> bool {
...
+    )]
+    /// Encode a key in escaped format
+    encode: Option<String>,
+
+    #[structopt(subcommand)]
+    cmd: Option<Cmd>,
+}
+
+#[derive(StructOpt)]
+enum Cmd {
```
skyzh, line 1150: Reviewed to this line so far, will continue later.

cmd/tikv-ctl/src/main.rs
peng1999: Yes it’s required. `App::version` takes a `&str`, but `get_version_info()` returns a `String` because of the use of `format!` and `str::trim`. It must have a static variable to hold ownership of the version string.

cmd/tikv-ctl/src/main.rs
peng1999: Doc strings and `help = xxx` are equivalent. See [related chapter](https://docs.rs/structopt/0.3.22/structopt/#help-messages) of its document.

cmd/tikv-ctl/src/main.rs
```diff
@@ -1050,867 +1051,661 @@ fn warning_prompt(message: &str) -> bool {
...
+        /// db directory.
+        db: String,
+
+        #[structopt(long)]
+        /// specify manifest, if not set, it will look up manifest file in db path
+        manifest: Option<String>,
+
+        #[structopt(long, value_delimiter = ",")]
+        /// PD endpoints
+        pd: String,
```
skyzh, line 1590: Should this be `Vec<String>` instead?

cmd/tikv-ctl/src/main.rs
```diff
@@ -1050,867 +1051,661 @@ fn warning_prompt(message: &str) -> bool {
...
+    },
+    #[structopt(external_subcommand)]
+    External(Vec<String>),
+}
+
+#[derive(StructOpt)]
+enum RaftCmd {
```
skyzh, line 1597: Reviewed to this point, will continue later.

cmd/tikv-ctl/src/main.rs
peng1999: The current code in master is confusing.

On one side, `value_delimiter` is [presented](https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs#L1902), and the [help text](https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs#L1903) indicates that it accepts multiple arguments. On the other side, `.multiple(true)` is *not* presented, and the argument is [used](https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs#L1903) as a single value. Thus in master, this leads a strange behaviour that it do accept multiple arguments, but only the first one is used. And this PR have the same help information and usage logic as master, with only difference that multiple arguments will results a panic, the behaviour which I think is not worse.

This PR is indented to have as few behaviour change as possible. So my suggestion is to open a new issue to address this, and keep the status quo in this PR.

peng1999: The current code in master is confusing.

On one side, `value_delimiter` is [presented](https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs#L1902), and the [help text](https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs#L1903) indicates that it accepts multiple arguments. On the other side, `.multiple(true)` is *not* presented, and the argument is [used](https://github.com/tikv/tikv/blob/master/cmd/tikv-ctl/src/main.rs#L1903) as a single value. Thus in master, this leads a strange behaviour that it do accept multiple arguments, but only the first one is used. And this PR have the same help information and usage logic as master, with only difference that multiple arguments will results a panic, the behaviour which I think is not worse.

This PR is indented to have as few behaviour change as possible. So my suggestion is to open a new issue to address this, and keep the status quo in this PR.

peng1999:
@skyzh Do you still have time to review this PR?

skyzh:
> @skyzh Do you still have time to review this PR?

Will look into the PR this weekend :)

cmd/tikv-ctl/src/main.rs
```diff
@@ -2660,12 +2439,7 @@ fn run_ldb_command(cmd: &ArgMatches<'_>, cfg: &TiKvConfig) {
     engine_rocks::raw::run_ldb_tool(&args, &opts);
 }
 
-fn run_sst_dump_command(cmd: &ArgMatches<'_>, cfg: &TiKvConfig) {
-    let mut args: Vec<String> = match cmd.values_of("") {
-        Some(v) => v.map(ToOwned::to_owned).collect(),
-        None => Vec::new(),
-    };
-    args.insert(0, "sst_dump".to_owned());
```
skyzh, line 2668: Why remove this line? Would you please test if `sst_dump` works?

cmd/tikv-ctl/src/main.rs
```diff
@@ -2035,67 +1823,78 @@ fn main() {
...
-    if let Some(cmd) = matches.subcommand_matches("ldb") {
-        run_ldb_command(cmd, &cfg);
-        return;
-    }
+    let mgr = new_security_mgr(&opt);
+
+    let cmd = match opt.cmd {
+        Some(cmd) => cmd,
+        None => {
+            // Deal with arguments about key utils.
```
skyzh, line 1831: It seems this part is duplicated with L1873?

skyzh (commented):
Rest LGTM

cmd/tikv-ctl/src/main.rs
peng1999: Yes it's tested and works.

Unlike clap's `ArgMatches`, the `external_subcommand` args in structopt contains the subcommand itself. See https://docs.rs/structopt/0.3.23/structopt/#external-subcommands and L1852. Thus there is no need to insert a `"sst_dump"`.
https://github.com/tikv/tikv/blob/3e53d3b342aba694b11686afb1f6964ee61f89b1/cmd/tikv-ctl/src/main.rs#L1852-L1856

cmd/tikv-ctl/src/main.rs
peng1999: Yes they are duplicate code. This is due to a merge mistake. Fixed.

skyzh (approved):
LGTM. Would you please check if there are any configuration changes during this PR is being reviewed? If there is no more code to change or rebase, ping me for a merge. Thanks for your contribution!

peng1999:
@skyzh There is only one change (which is fd1385f1f2a) in `tikv-ctl/src/main.rs` during the review of the PR. I have already doing a rebase two weeks ago. I think no more rebase is needed.

cmd/tikv-ctl/src/main.rs
andylokandy: is it possible that the tikv-ctl defines these two commands and therefore proxy it to rocks with the same args?

peng1999: Currently the arg proxy mechanism is done my `external_command` in `clap` as well as `structopt`. Otherwise you cannot get arbitrary args and flags as-is without a `--` seperator. That is, if we define these two sub-commands, we must write a `--` between `sst-dump` and the first flag: `tikv-ctl sst-dump -- --help`. See also https://github.com/clap-rs/clap/issues/484.

andylokandy (approved):

andylokandy:
@peng1999 Thank you for your contribution. @skyzh also appreciated your hard work on reviewing!

cmd/tikv-ctl/src/main.rs
```diff
@@ -1117,913 +1118,700 @@ fn warning_prompt(message: &str) -> bool {
...
+        // so we left `all_regions` unused.
+        #[allow(dead_code)]
+        #[structopt(required_unless = "regions", conflicts_with = "regions", long)]
+        /// Do the command for all regions
+        all_regions: bool,
+    },
+}
+
+fn main() {
+    let opt = Opt::from_args();
```
JayiceZ, line 1808: Does it look like this line does not return the value we want?

ref https://github.com/tikv/tikv/issues/11514

Issue #10841, opened 2021-08-27
*:Logger thread can be bottleneck and slow down the whole TiKV instance


sunxiaoguang:
### What version of TiKV are you using?

4.0.14

### What operating system and CPU are you using?

Linux x86_64

### Steps to reproduce

Send extreme coprocessor load to a single region to trigger enormous  amount of slow log that can overload slogger thread. Once slogger thread is overloaded, log message can accumulate and fill up pending message queue quickly. All code path that generates log are affected and can't make progress as usual.   

### What did you expect?
Try to send less slow log to slogger at first to make it less likely to collapse. And having a mechanism to short circuit not critical logs when slogger thread is heavy loaded.      

### What did happened?
Slogger thread uses 100% CPU and slow down everything. Especially coprocessor.

BusyJay:
I propose to reduce the slow log count and enlarge the log buffer to twice the current. Also @sticnarf suggested to replace mpsc channel in slog-async as crossbeam channel, which should be faster.

zhangjinpeng1987:
The bottleneck is the formatting part of slogger, I prefer
1) using OverflowStrategy::Drop instead of OverflowStrategy::Block for short term solution, since print log become the bottleneck is not a frequently thing. So the first priority is keep the service is OK under any throughput.
2) simplify and reduce the total number of log when the throughput is high, we need to sample the same type logs to guarantee key messages are logged. Also I prefer using OverflowStrategy::Drop when the slogger is overflow.

sticnarf:
## Short-term solution

I agree to reduce the slow log count first by changing the `req_lifetime` to `total_process_time`: https://github.com/tikv/tikv/blob/19e4d00ad7965681b5139cd71c7eaf8290cc22ce/src/coprocessor/tracker.rs#L249 

And not all errors need to be logged here https://github.com/tikv/tikv/blob/19e4d00ad7965681b5139cd71c7eaf8290cc22ce/src/coprocessor/endpoint.rs#L635

I'm afraid it is not useful to enlarge the log buffer. If the bottleneck is the slogger thread, however large the buffer is, the buffer will be finally filled full.

## Mid-term solution

We fork slog project and add `try_log` macros which allows dropping the log when the buffer is full. I think the feature is useful for other users and may be potentially accepted by the upstream.

sunxiaoguang:
[...]

PR #10858, opened 2021-08-30
coprocessor: reduce the slow log count by changing the req_lifetime to total_process_time

L3T:

#10841 *:Logger thread can be bottleneck and slow down the whole TiKV instance 

Problem Summary:

Proposal: #10841

What's Changed:

### Related changes

release-3.0
release-3.1
release-4.0
release-5.0
release-5.1
release-5.2

### Check List 

Tests 

- No code

Side effects

- Slow log will only consider time spent on processing the request. 
- Drop log instead of blocking threads when slogger thread is overloaded and queue is filled up. 

### Release note 

```release-note
- TiKV coprocessor slow log will only consider time spent on processing the request. 
- Drop log instead of blocking threads when slogger thread is overloaded and queue is filled up. 
```

### DIFF
--- a/components/tikv_util/src/logger/mod.rs
+++ b/components/tikv_util/src/logger/mod.rs
@@ -31,7 +31,7 @@ pub const DATETIME_ROTATE_SUFFIX: &str = "%Y-%m-%d-%H:%M:%S%.f";
 const SLOG_CHANNEL_SIZE: usize = 10240;
 // Default is DropAndReport.
 // It is not desirable to have dropped logs in our use case.
-const SLOG_CHANNEL_OVERFLOW_STRATEGY: OverflowStrategy = OverflowStrategy::Block;
+const SLOG_CHANNEL_OVERFLOW_STRATEGY: OverflowStrategy = OverflowStrategy::Drop;
 const TIMESTAMP_FORMAT: &str = "%Y/%m/%d %H:%M:%S%.3f %:z";
 
 static LOG_LEVEL: AtomicUsize = AtomicUsize::new(usize::max_value());
--- a/src/coprocessor/tracker.rs
+++ b/src/coprocessor/tracker.rs
@@ -246,7 +246,7 @@ impl Tracker {
 
         let total_storage_stats = std::mem::take(&mut self.total_storage_stats);
 
-        if self.req_lifetime > self.slow_log_threshold {
+        if self.total_process_time > self.slow_log_threshold {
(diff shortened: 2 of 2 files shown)


sunxiaoguang (approved):
LGTM

skyzh (approved):
LGTM. Are you sure to cherry-pick this commit to so many releases?

skyzh:
Ping us when you are ready to merge. Feel free to add or remove releases you want to cherry-pick.

L3T:
I`m Ready to Merge.

components/tikv_util/src/logger/mod.rs
```diff
@@ -31,7 +31,7 @@ pub const DATETIME_ROTATE_SUFFIX: &str = "%Y-%m-%d-%H:%M:%S%.f";
 const SLOG_CHANNEL_SIZE: usize = 10240;
 // Default is DropAndReport.
 // It is not desirable to have dropped logs in our use case.
-const SLOG_CHANNEL_OVERFLOW_STRATEGY: OverflowStrategy = OverflowStrategy::Block;
+const SLOG_CHANNEL_OVERFLOW_STRATEGY: OverflowStrategy = OverflowStrategy::Drop;
```
zhongzc, line 34: Comments need to be updated.

zhongzc: Would `DropAndReport` be better?

sunxiaoguang: @L3T please take a look at comment. You can file a new PR to address comment. And make sure all the cherrypick PRs have comments updated, thanks.

sunxiaoguang: > Would `DropAndReport` be better?

Good question, we kind of discussed this strategy as well. The agreement was overflowing message isn't much useful to both users and developers therefore `Drop` is an appropriate choice. However it is definitely open to discussion before our long term solution is implemented.

zhongzc: Thanks for your reply.

A drop report log can inform us the logger is busy in a cheap way. Although we are working on letting this information be shown on grafana, I feel better with an error can bring many different forms of warnings.

sunxiaoguang: DropAndReport inform us by a line of log message which is for human being only. We can neither know anything about it from metrics nor the context of dropped log was about. And that's why we thought it was not useful at the moment and used Drop instead.

PR #10860, opened 2021-08-30
trace memory usage for analyze

hicqu:

Trace memory usage for analyze.

### Check List 

Tests 

- Unit test
- Manual test
  
图片

Side effects

### Release note 

```release-note
None
```


src/coprocessor/statistics/analyze.rs
```diff
@@ -31,11 +30,13 @@ use super::cmsketch::CmSketch;
 use super::fmsketch::FmSketch;
 use super::histogram::Histogram;
 use crate::coprocessor::dag::TiKVStorage;
+use crate::coprocessor::MEMTRACE_ANALYZE;
 use crate::coprocessor::*;
 use crate::storage::{Snapshot, SnapshotStore, Statistics};
 
 const ANALYZE_VERSION_V1: i32 = 1;
 const ANALYZE_VERSION_V2: i32 = 2;
+const ANALYZE_MAX_TIME_SLICE: Duration = Duration::from_secs(1);
```
hicqu, line 39: @sticnarf here I think analyze task can run long time, so I change the time slice from `1ms` to `1s`. Please take a look whether it's ok/necessary or not.

sticnarf: Time slice should not be that long. Otherwise it may occupy the thread for too long and short tasks will be affected.
Short time slice does not mean the task is only allowed to run that short

components/tikv_alloc/src/trace.rs
```diff
@@ -237,6 +237,77 @@ macro_rules! mem_trace {
...
+    {
+        let item = std::mem::take(&mut self.item);
+        MemTraced {
+            item: f(item),
+            size: self.size,
+            node: self.node.take(),
+        }
+    }
+
+    pub fn consume(&mut self) -> T {
```
zhongzc, line 264: How about using `fn into_inner(self) -> T` as function signature just like [Box::into_inner](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_inner) and [Cell::into_inner](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.into_inner)?

hicqu: I prefer this too but It conflicts with `fn drop`. In Rust a struct with explicitly `drop` can't have another method which moves itself.

components/tikv_alloc/src/trace.rs
```diff
@@ -237,6 +237,77 @@ macro_rules! mem_trace {
     }
 }
 
+pub struct MemTraced<T: Default> {
+    item: T,
+    size: usize,
+    node: Option<Arc<dyn MemoryTrace + Send + Sync>>,
```
zhongzc, line 243: How about extracting these fields into a `Guard`?

My thought: `Guard` is built via `MemoryTrace::guard(size)`. We can do `MemoryTrace::trace(Add)` inside `MemoryTrace::guard` and `MemoryTrace::trace(Sub)` inside `Guard::drop`.

andylokandy (approved):

PR #10875, opened 2021-08-31
coprocessor: Prevent reloading from unloaded plugins

### LINKED ISSUE #10854: copr_plugin: should disallow unregistering plugins
## Bug Report

Currently, TiKV loads plugin from `./coprocessors` on startup by default, and if the plugin directory is explicitly defined in the config file, TiKV will additionally watch that folder and load new plugins when file added and remove from TiKV when the corresponding file is removed.

But now we've found a [problem](https://github.com/andylokandy/rfcs/blob/plugin/text/0063-coprocessor-plugin.md#auto-loading-1) that prevents us to remove plugins. So we decided to disallow unregistering plugins. In detail:

1. Don't watch file change / file remove events anymore.
2. Use [fs2](https://crates.io/crates/fs2) to lock the plugin file when the plugin is loaded into TiKV.

peng1999:

Issue Number: close #10854 

Problem Summary:

It solves the coprocessor hot reloading correctness issue by simply forbid reloading.

What's Changed:

It [turns out](https://github.com/andylokandy/rfcs/blob/plugin/text/0063-coprocessor-plugin.md#auto-loading-1) that due to TLS related problems, the coprocessor plugin cannot hot reload correctly. The original plan is to prevent deletion and modification of the loaded plugin using file locks. But that's not possible because there's no reliable way to acquire a mandatory file lock. To solve this problem, this PR simply ~~unloads the  plugin whenever it removed or changed, and forbid future reloading from the same path~~ forbid unloading from loaded plugins regardless of file status.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/src/coprocessor_v2/plugin_registry.rs
+++ b/src/coprocessor_v2/plugin_registry.rs
@@ -279,7 +279,7 @@ struct PluginRegistryInner {
     /// Provides a mapping from the plugin's name to the actual instance.
     loaded_plugins: HashMap<String, (OsString, Arc<LoadedPlugin>)>,
 
-    /// Paths of loaded *and* unloaded plugins.
+    /// Original paths that plugins loaded from.
     /// Files in this list should not be loaded again.
     library_paths: HashSet<OsString>,
 }
--- a/src/coprocessor_v2/plugin_registry.rs
+++ b/src/coprocessor_v2/plugin_registry.rs
@@ -584,7 +584,7 @@ mod tests {
         // fs watcher detects changes in every 3 seconds, therefore, wait 4 seconds so as to make sure the watcher is triggered.
         std::thread::sleep(Duration::from_secs(4));
 
-        // plugin will not unloading
+        // plugin will not be unloadad
         assert!(registry.get_plugin(plugin_name).is_some());
     }
 }
--- a/src/coprocessor_v2/plugin_registry.rs
+++ b/src/coprocessor_v2/plugin_registry.rs
@@ -580,11 +580,11 @@ mod tests {
             &library_path_2
         );
 
-        // trigger unloading
         std::fs::remove_file(&library_path_2).unwrap();
         // fs watcher detects changes in every 3 seconds, therefore, wait 4 seconds so as to make sure the watcher is triggered.
         std::thread::sleep(Duration::from_secs(4));
 
-        assert!(registry.get_plugin(plugin_name).is_none());
+        // plugin will not unloading
+        assert!(registry.get_plugin(plugin_name).is_some());
     }
 }
--- a/src/coprocessor_v2/plugin_registry.rs
+++ b/src/coprocessor_v2/plugin_registry.rs
@@ -132,12 +132,6 @@ impl PluginRegistry {
                         hot_reload_registry.load_plugin(file).ok();
                     }
                 };
-                let unload = |file: &PathBuf| {
-                    let mut hot_reload_registry = hot_reload_registry.write().unwrap();
-                    if let Some(plugin) = hot_reload_registry.get_plugin_by_path(file) {
-                        hot_reload_registry.unload_plugin(plugin.name());
-                    }
-                };
                 let rename = |old_path: &PathBuf, new_path: &PathBuf| {
                     let mut hot_reload_registry = hot_reload_registry.write().unwrap();
                     if let Some(plugin) = hot_reload_registry.get_plugin_by_path(old_path) {
@@ -151,7 +145,8 @@ impl PluginRegistry {
                             maybe_load(&file);
                         }
                         Ok(DebouncedEvent::Remove(file)) => {
-                            unload(&file);
+                            // We cannot do much when the file is deleted. See issue #10854
+                            warn!("a loaded coprocessor plugin is removed. Be aware that original plugin is still running"; "plugin_path" => ?file);
                         }
                         Ok(DebouncedEvent::Rename(old_file, new_file)) => {
                             // If the file is renamed with a different parent directory, we will receive a `Remove` instead.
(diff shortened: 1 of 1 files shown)


src/coprocessor_v2/plugin_registry.rs
```diff
@@ -306,12 +315,22 @@ impl PluginRegistryInner {
         &mut self,
         filename: P,
     ) -> Result<String, PluginLoadingError> {
+        if self.library_paths.contains(filename.as_ref()) {
+            let err = Err(PluginLoadingError::ReloadError {
+                path: filename.as_ref().to_owned(),
+            });
+            let filename = filename.as_ref().to_string_lossy();
```
skyzh, line 322: `OsStr` should have implemented `Debug`, so we do not need to convert that to `String` here. Does the compiler complains if we remove `.to_string_lossy()`?

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -153,9 +159,8 @@ impl PluginRegistry {
                             rename(&old_file, &new_file);
                         }
                         Ok(DebouncedEvent::Write(file)) => {
-                            warn!("another process is overwriting a coprocessor plugin while the plugin is loaded. This can lead to severe issues!"; "plugin_path" => ?file);
+                            warn!("another process is overwriting a coprocessor plugin while the plugin is loaded. Unloading the plugin..."; "plugin_path" => ?file);
                             unload(&file);
```
skyzh, line 163: After this change, do we really need to unload the plugin?

> To solve this problem, this PR simply unloads the plugin whenever it removed or changed, and forbid future reloading from the same path.

Unloading a plugin may cause problems. Imagine that there is a tokio future running in the runtime, and the plugin disappears. Then the program would crash. Since a plugin cannot be released for now, what about also removing `unload` here?

src/coprocessor_v2/plugin_registry.rs
peng1999: Yes `to_string_lossy()` is optional. But this line is copied from [the same function](https://github.com/tikv/tikv/blob/ef5a703dc671e9eb0583f90be7f8b5b1552b6ab2/src/coprocessor_v2/plugin_registry.rs#L328). Is there a reason not to be consistent?

src/coprocessor_v2/plugin_registry.rs
peng1999: I concerned following behaviours:

||file renamed| file changed | file deleted |
|--|----|---|---|
|behaviour 1|unload|unload|unload|
|behaviour 2|do nothing|unload|unload|
|behaviour 3|do nothing|do nothing|unload|
|behaviour 4|do nothing|do nothing|do nothing|

In this PR we have behaviour 2. I think it will be useful if users can unload a plugin, and "file changed" and "file deleted" both implies the original file is not needed.

Do you mean we should use behaviour 3? If `unload` is a dangerous operation, should we use behaviour 4? I'm not sure which is more sensible.

skyzh: I would prefer behavior 4 indeed 🤣🤣 what do you think @andylokandy

andylokandy: Given that we can neither lock the plugin file nor reload when the plugin file is recreated, we can not do much. I prefer behavior 4 because it's simpler for the user, but we can print a warning instead of doing nothing.

peng1999: Ok. Then I will remove all unload related code.

skyzh (approved):

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -566,11 +580,11 @@ mod tests {
             &library_path_2
         );
 
-        // trigger unloading
         std::fs::remove_file(&library_path_2).unwrap();
         // fs watcher detects changes in every 3 seconds, therefore, wait 4 seconds so as to make sure the watcher is triggered.
         std::thread::sleep(Duration::from_secs(4));
 
-        assert!(registry.get_plugin(plugin_name).is_none());
+        // plugin will not unloading
```
andylokandy, line 587:
```suggestion
        // plugin will not be unloadad
```

src/coprocessor_v2/plugin_registry.rs
```diff
@@ -278,6 +278,10 @@ struct PluginRegistryInner {
     /// Plugins that are currently loaded.
     /// Provides a mapping from the plugin's name to the actual instance.
     loaded_plugins: HashMap<String, (OsString, Arc<LoadedPlugin>)>,
+
+    /// Paths of loaded *and* unloaded plugins.
+    /// Files in this list should not be loaded again.
+    library_paths: HashSet<OsString>,
```
andylokandy, line 284: Since no plugin will be unloaded, what does this differ from the `loaded_plugins`?

peng1999: Plugins will not be unloaded but they can be renamed when the corresponding `so` file being renamed. But I believe that it is the *original* path that should not be reload again.  (I can't reproduce the TLS related behaviour in my computer though)

peng1999: The doc comment is a bit outdated, I will update it soon.

andylokandy: Maybe we can make rename a noop either.

andylokandy: fair enough

andylokandy (approved):

peng1999:
@andylokandy I rebased the branch to the current master, and the merge is canceled. Can you please re-trigger the merge?

PR #10876, opened 2021-08-31
resrc_mtr: extract endpoint trait

zhongzc:

Make code smell good

What's Changed:

- Extract a trait `Endpoint`, and the gRPC endpoint implements it.
- Decouple some stuff

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/resource_metering/src/cpu/reporter/endpoint/grpc.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl GRPCEndpoint {
+    fn init_client(&mut self) {
+        let channel = {
+            let cb = ChannelBuilder::new(self.env.clone())
+                .keepalive_time(Duration::from_secs(10))
+                .keepalive_timeout(Duration::from_secs(3));
+            cb.connect(&self.address)
```
mornyx, line 94: Can this tcp connection be reconnected when the server(agent) down or restarts? If not, what I understand is that all gRPC requests will fail after a restart from the server side, because the connection has been broken. So do we need to add a complex connection management mechanism for this?

zhongzc: I think it will reconnect. The [integration test](https://github.com/tikv/tikv/blob/5c887790bfbe2487f479a5fa964577f415f247e2/tests/integrations/resource_metering/test_agent.rs#L136-L199) simulated the situation you described.

PR #4232, opened 2019-02-19
raftstore: remove message box/unbox

BusyJay:
## What have you changed? (mandatory)

This PR removes the redundant enum Msg introduced in #4066, which aimed to maintain a compatible layer and made refactor more smoothly. Now that the PR is merged, it's time to clean up. This PR also simplifies the signature of simulate transport by merge `SendFilter` and `RecvFilter` as a unify `Filter`.

## What are the type of the changes? (mandatory)

- Improvement

## How has this PR been tested? (mandatory)

unit tests

## Does this PR affect documentation (docs) update? (mandatory)

No.

## Does this PR affect tidb-ansible update? (mandatory)

No.

## Refer to a related PR or issue link (optional)

#4066

PR #10901, opened 2021-09-03
Add tikv_channel_full metric

### LINKED ISSUE #10377: Channel full metric needs to be updated
\>= v3.0.0

doesn't matter

### Steps to reproduce
Just see the channel full metric

### What did you expect?
The channel full metric should be updated when some channels are full.

### What did happened?
The channel full metric is always empty even when some peer's channel are full.

image
I find the `tikv_channel_full` metric has been removed for a long time since v3.0.0. It should be updated.

lsampras:
Signed-off-by: sampras lopes <sampraslop995@gmail.com>

Issue Number: close #10377

Problem Summary:

What's Changed:
- Add `tikv_channel_full` metric for monitoring channel full events

### Check List 

Tests 

- Manual test

Monitor the `tikv_channel_full_total` metric on grafana

Side effects
No

### Release note 

```release-note
None
```

### DIFF
--- a/components/batch-system/src/metrics.rs
+++ b/components/batch-system/src/metrics.rs
@@ -1,0 +1,13 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use lazy_static::lazy_static;
+use prometheus::*;
+
+lazy_static! {
+    pub static ref CHANNEL_FULL_COUNTER_VEC: IntCounterVec = register_int_counter_vec!(
+        "tikv_channel_full_total",
+        "Total number of channel full errors.",
+        &["type"]
+    )
+    .unwrap();
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,6 +271,8 @@ dependencies = [
  "crossbeam",
  "derive_more",
  "file_system",
+ "lazy_static",
+ "prometheus",
  "serde",
  "serde_derive",
  "slog",
--- a/components/batch-system/Cargo.toml
+++ b/components/batch-system/Cargo.toml
@@ -26,6 +26,8 @@ serde_derive = "1.0"
 slog = { version = "2.3", features = ["max_level_trace", "release_max_level_debug"] }
 slog-global = { version = "0.1", git = "https://github.com/breeswish/slog-global.git", rev = "d592f88e4dbba5eb439998463054f1a44fbf17b9" }
 derive_more = { version = "0.99", optional = true }
+prometheus = { version = "0.12", default-features = false, features = ["nightly"] }
+lazy_static = "1.3"
 
 [dev-dependencies]
 criterion = "0.3"
--- a/components/batch-system/src/lib.rs
+++ b/components/batch-system/src/lib.rs
@@ -4,6 +4,7 @@ mod batch;
 mod config;
 mod fsm;
 mod mailbox;
+mod metrics;
 mod router;
 
 #[cfg(feature = "test-runner")]
--- a/components/batch-system/src/router.rs
+++ b/components/batch-system/src/router.rs
@@ -2,6 +2,7 @@
 
 use crate::fsm::{Fsm, FsmScheduler, FsmState};
 use crate::mailbox::{BasicMailbox, Mailbox};
+use crate::metrics::CHANNEL_FULL_COUNTER_VEC;
 use collections::HashMap;
 use crossbeam::channel::{SendError, TrySendError};
 use std::cell::Cell;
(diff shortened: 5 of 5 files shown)


src/server/raft_client.rs
```diff
@@ -892,6 +892,9 @@ where
                         if !s.full {
                             s.full = true;
                             self.full_stores.push((store_id, conn_id));
+                            CHANNEL_FULL_COUNTER_VEC
```
BusyJay, line 895: It should be recorded in `kv.rs`, this is a different case.

lsampras: Thanks,
I have now moved this to the `handle_raft_message` function in `kv.rs`.

src/server/service/kv.rs
```diff
@@ -153,7 +153,11 @@ impl<T: RaftStoreRouter<E::Local> + 'static, E: Engine, L: LockManager> Service<
             let id = msg.get_region_id();
             let peer_id = msg.get_message().get_from();
             let m = CasualMessage::RejectRaftAppend { peer_id };
-            let _ = ch.send_casual_msg(id, m);
+            if let Err(RaftStoreError::Transport(DiscardReason::Full)) = ch.send_casual_msg(id, m) {
```
BusyJay, line 156: It should be checked at L166.

src/server/service/kv.rs
```diff
@@ -160,6 +160,12 @@ impl<T: RaftStoreRouter<E::Local> + 'static, E: Engine, L: LockManager> Service<
         match ch.send_raft_msg(msg) {
             Ok(()) => Ok(()),
             Err(RaftStoreError::RegionNotFound(_)) => Ok(()),
+            Err(RaftStoreError::Transport(DiscardReason::Full)) => {
+                CHANNEL_FULL_COUNTER_VEC
+                    .with_label_values(&[&store_id.to_string()])
+                    .inc();
+                Ok(())
```
BusyJay, line 167: Should return `Err(Error::from(e))`.

lsampras: do you mean something like `Err(Error::from(RaftStoreError::Transport(DiscardReason::Full)))` ?

Also confused about why the `RegionNotFound` error is being silenced & Discard is not.

BusyJay: `RegionNotFound` can happen in normal condition, ignore it to avoid building new connections.

BusyJay (approved):

src/server/service/kv.rs
```diff
@@ -160,6 +160,12 @@ impl<T: RaftStoreRouter<E::Local> + 'static, E: Engine, L: LockManager> Service<
         match ch.send_raft_msg(msg) {
             Ok(()) => Ok(()),
             Err(RaftStoreError::RegionNotFound(_)) => Ok(()),
+            Err(RaftStoreError::Transport(DiscardReason::Full)) => {
+                CHANNEL_FULL_COUNTER_VEC
+                    .with_label_values(&[&store_id.to_string()])
+                    .inc();
```
gengliqi, line 166: Seems this metric can be updated in `handle_send_error` in `router.rs` in order to detect the channel full errors in `send_raft_msg` and `send_command` from all callers.

lsampras: Ok any labels to be added to provide info about the source of channel full errors if we are detecting from all_callers?

Also what would be a good replacement for the existing label value (we used channel name [earlier](https://github.com/tikv/tikv/pull/4232/files#diff-2d7955eb02fa5282d80c7917a02c45bb5a669a7e5fea181d176e38e664297723L122) which is no longer available).

gengliqi: We don't care about the `store_id` because it's always the local id. But we care about which type of channel is full, namely, normal channel or control channel.
You can the code(it's better to add the metric here)https://github.com/tikv/tikv/blob/master/components/raftstore/src/store/fsm/store.rs#L226-L285.

lsampras: I've updated the metrics in `components/batch-system/src/router.rs` which is being used by the code you referenced.
Using `control` & `normal` as the values of the type label.

gengliqi (approved):
LGTM，thanks!

components/batch-system/src/metrics.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
```
gengliqi, line 1:
```suggestion
// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```

BusyJay:
Receiving snapshot too slow lead to failure case, it's unrelated to the PR.

/run-test

PR #10922, opened 2021-09-08
coprocessor: disable coproccesor when dir config is not presented.

### LINKED ISSUE #10855: copr_plugin: Disable coprocessor plugin if the plugin folder is not explictly sepcified in config
## Feature Request

Currently, TiKV loads all plugins from `./coprocessors` if the plugin folder is not explicitly specified in config, which may not be a good thing for security. And as stated in config:

https://github.com/tikv/tikv/blob/19e4d00ad7965681b5139cd71c7eaf8290cc22ce/etc/config-template.toml#L445-L450

Whether specifying the plugin folder or not enables or disenables the auto-loading, which seems puzzling. So I suggest changing the config as so:

```toml
[coprocessor-v2]
# Path to the directory where compiled coprocessor plugins are located.
# Plugins in this directory will be automatically loaded by TiKV.
# If the config value is not set, the coprocessor plugin will be disabled.
coprocessor-plugin-directory = "./coprocessors"
```

peng1999:
Signed-off-by: Peng Guanwen <pg999w@outlook.com>

Issue Number: close #10855 

Disable plugin when `coprocessor-plugin-directory` is not presented.
coprocessor request will result an error when it  is disabled.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Adjust the behaviour of  `coprocessor-v2.coprocessor-plugin-directory` config item.
```

### DIFF
--- a/src/coprocessor_v2/endpoint.rs
+++ b/src/coprocessor_v2/endpoint.rs
@@ -19,28 +19,30 @@ enum CoprocessorError {
 /// A pool to build and run Coprocessor request handlers.
 #[derive(Clone)]
 pub struct Endpoint {
-    plugin_registry: Arc<PluginRegistry>,
+    plugin_registry: Option<Arc<PluginRegistry>>,
 }
 
 impl tikv_util::AssertSend for Endpoint {}
 
 impl Endpoint {
     pub fn new(copr_cfg: &Config) -> Self {
-        let mut plugin_registry = PluginRegistry::new();
+        let plugin_registry =
+            copr_cfg
+                .coprocessor_plugin_directory
+                .as_ref()
+                .map(|plugin_directory| {
+                    let mut plugin_registry = PluginRegistry::new();
 
-        // Enable hot-reloading of plugins if the user configured a directory.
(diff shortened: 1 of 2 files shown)


src/coprocessor_v2/endpoint.rs
```diff
@@ -19,28 +19,30 @@ enum CoprocessorError {
...
+            copr_cfg
+                .coprocessor_plugin_directory
+                .as_ref()
+                .map(|plugin_directory| {
+                    let mut plugin_registry = PluginRegistry::new();
+
+                    if let Err(err) = plugin_registry.start_hot_reloading(plugin_directory) {
+                        warn!("unable to start hot-reloading for coprocessor plugins.";
                     "coprocessor_directory" => plugin_directory.display(),
                     "error" => ?err);
```
andylokandy, line 39: the lines are not aligned.

andylokandy (approved):
rest LGTM

src/coprocessor_v2/endpoint.rs
peng1999: Code formatted.

andylokandy:
cc @zhongzc

skyzh (approved):

PR #10929, opened 2021-09-10
encryption: support seek for encryption writer

tabokie:

Problem Summary:

Turns out we still need that `Seek` for encryption to work with raft engine.

What's Changed:

Unify implementation for both encryption reader and writer. `CrypterCore` now encapsulate all the encryption core algorithm, and it's seek-able. Implementation details are mostly the same, read/write still uses different code paths:

* reader calls `do_crypter_in_place`, writer calls `do_crypter`. Only the latter will return a slice of internal buffer as encryption result.
* reader use `MAX_INPLACE_CRYPTION_SIZE`(1M) as a upper bound for internal buffer, writer impose no such limit

### Check List 

Tests

- Unit test

### Release note 

```release-note
None
```


components/encryption/src/io.rs
```diff
@@ -17,70 +17,162 @@ use file_system::File;
 use tikv_util::box_err;
 
 const AES_BLOCK_SIZE: usize = 16;
-const MAX_CRYPTER_READER_BUFFER_SIZE: usize = 1024;
+const MAX_INPLACE_CRYPTION_SIZE: usize = 1024 * 1024;
 
-/// Encrypt content as data being read.
-pub struct EncrypterReader<R>(CrypterReader<R>);
+/// Encrypt or decrypt content as data being read.
+pub struct CrypterReader<R> {
```
Connor1996, line 23: can we still separate `Encrypter` and `Decrypter`?

tabokie: Make sense, it's separated now.

components/encryption/src/io.rs
```diff
@@ -17,73 +17,303 @@ use file_system::File;
...
+        Ok(EncryptWriter(CrypterWriter::new(
+            writer,
+            method,
+            key,
+            Mode::Encrypt,
+            iv,
+        )?))
+    }
+
+    pub fn finalize(self) -> IoResult<W> {
```
Connor1996, line 117: why need to return the inner writer

Connor1996: should we call flush explicitly?

components/encryption/src/io.rs
```diff
@@ -17,73 +17,303 @@ use file_system::File;
 use tikv_util::box_err;
 
 const AES_BLOCK_SIZE: usize = 16;
-const MAX_CRYPTER_READER_BUFFER_SIZE: usize = 1024;
```
Connor1996, line 20: why remove

components/encryption/src/io.rs
```diff
@@ -106,63 +336,50 @@ pub fn create_aes_ctr_crypter(
...
-    fn reserve_buffer(&mut self, size: usize) {
+    #[inline]
+    fn reset_buffer(&mut self, size: usize) {
         // OCrypter require the output buffer to have block_size extra bytes, or it will panic.
-        let target = std::cmp::min(size, MAX_CRYPTER_READER_BUFFER_SIZE) + self.block_size;
-        if target > self.crypter_buffer.len() {
-            self.crypter_buffer.resize(target, 0);
+        self.buffer.reserve(size + self.block_size);
+        unsafe {
+            self.buffer.set_len(size + self.block_size);
```
Connor1996, line 369: not sure if it's okay to not fill the padding of 0. /cc @yiwu-arbug

components/encryption/src/io.rs
tabokie: It's just an idiomatic to deal with chained streams, e.g. https://docs.rs/lz4/1.15.131/lz4/struct.Encoder.html#method.finish or https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.into_inner.
This function usually only guarantees bytes are written, no need to be flushed.

components/encryption/src/io.rs
tabokie: It's renamed.

components/encryption/src/io.rs
tabokie: It's just an output buffer, the behavior is stated in document: https://docs.rs/openssl/0.9.1/openssl/symm/struct.Crypter.html#method.update

components/encryption/src/io.rs
Connor1996: but with different value

tabokie: @yiwu-arbug suggested in last PR

components/encryption/src/io.rs
```diff
@@ -422,25 +553,69 @@ mod tests {
...
+        OsRng.fill_bytes(&mut plaintext);
+        let offsets = [1024, 1024 + 1, 10240 - 1];
+        let sizes = [1024, 8000];
+        let written = vec![0; 10240];
+        for method in methods {
+            let key = generate_data_key(method);
+            let writable_text = std::io::Cursor::new(written.clone());
+            let iv = Iv::new_ctr();
+            let encrypter = EncrypterWriter::new(writable_text, method, &key, iv).unwrap();
+            let mut decrypter = DecrypterWriter::new(encrypter, method, &key, iv).unwrap();
```
hunterlxt, line 594: Maybe we need to prove here that the previous encryption was successful

tabokie: You mean encrypted != plaintext?

components/encryption/src/io.rs
hunterlxt: @tabokie can you post that

hunterlxt (approved):
LGTM

tabokie:
Pin @Connor1996

5kbpers (approved):

PR #10937, opened 2021-09-13
engine: provide customized file builder to raft engine

tabokie:

Issue Number: part of https://github.com/tikv/raft-engine/issues/95

What's Changed:

Update raft engine and opt in new interface.

### Related changes

Stacked on https://github.com/tikv/tikv/pull/10929.

### Check List 

Tests
- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raft_log_engine/src/engine.rs
+++ b/components/raft_log_engine/src/engine.rs
@@ -1,15 +1,19 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
 use std::fs;
+use std::io::{Read, Result as IoResult, Seek, SeekFrom, Write};
 use std::path::Path;
+use std::sync::Arc;
 
+use encryption::{DataKeyManager, DecrypterReader, EncrypterWriter};
 use engine_traits::{
     CacheStats, RaftEngine, RaftEngineReadOnly, RaftLogBatch as RaftLogBatchTrait, Result,
 };
+use file_system::{IOOp, IORateLimiter, IOType};
 use kvproto::raft_serverpb::RaftLocalState;
 use raft::eraftpb::Entry;
 use raft_engine::{
-    Command, Error as RaftEngineError, LogBatch, MessageExt, RaftLogEngine as RawRaftEngine,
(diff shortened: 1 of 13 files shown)


components/raft_log_engine/src/engine.rs
```diff
@@ -25,14 +29,143 @@ impl MessageExt for MessageExtTyped {
     }
 }
 
+struct ManagedReader<R: Seek + Read> {
+    raw: Option<R>,
```
hicqu, line 33: How about `tikv_util::Either` instead of `Option`?

components/raft_log_engine/src/engine.rs
```diff
@@ -25,14 +29,143 @@ impl MessageExt for MessageExtTyped {
...
+    fn read(&mut self, buf: &mut [u8]) -> IoResult<usize> {
+        let mut size = buf.len();
+        if let Some(ref mut limiter) = self.rate_limiter {
+            size = limiter.request(IOType::ForegroundRead, IOOp::Read, size);
+        }
+        match (&mut self.raw, &mut self.decrypter) {
+            (Some(ref mut reader), None) => reader.read(&mut buf[..size]),
+            (None, Some(ref mut reader)) => reader.read(&mut buf[..size]),
+            _ => unreachable!(),
+        }
```
hicqu, line 58: how about
```
let read_size = match {
};
limiter.request(read_size);
```

hicqu:
Rest LGTM.

components/raft_log_engine/src/engine.rs
tabokie: Good idea.

components/raft_log_engine/src/engine.rs
tabokie: We put the `request` before actual IOs in other call sites too (std::File, rocksdb::Env), right now let's keep it consistent. But I agree the ideal placement is an open question, will look into it when we work on rate limiting later.

Connor1996 (approved):
LGTM

hicqu (approved):

PR #10529, opened 2021-07-06
Make it possible to launch tikv-server with non-rocks KV engines

brson:
### What problem does this PR solve?

Issue Number: close https://github.com/tikv/tikv/issues/10528

Part of https://github.com/tikv/tikv/issues/6402

Problem Summary:

Make it possible to typecheck tikv-server with non-rocks KV engines, and to launch tikv-server with PanicEngine.

### What is changed and how it works?

What's Changed:

This is the last set of steps necessary to run tikv-server with a non-rocks KV engine. After this patch set it is possible to launch tikv-server with PanicEngine. This engine doesn't do anything, but it does ensure that tikv-server typechecks with alternate engines.

The big changes here are in components/server/src/server.rs, which is responsible for constructing the server.

These changes are kinda ugly, so this needs some good review.

There are still some limitations. Primarily, non-rocks engines are not able to launch the debug service or the backup service. These two still have hard rocks dependencies for now. This patch works around these limitations by introducing a `CreateKvEngine` trait that is implemented for all KV engines, and that contains some empty methods for PanicEngine.

The CreateKvEngine trait should be able to be deleted with time and refactoring.

Note that this patch series contains commits that temporarily introduce some ugly dynamic casts; those dynamic casts are later removed before the end of the series.

I have an out of tree engine, SimpleEngine, that is _almost_ able to complete the startup sequence and reach a steady-state where it doesn't crash. I am focused on getting that engine to more-or-less work, while reworking engine_traits to remove rocksisms, before attempting to add a "real" alternate engine. For others though, it should be feasible to work toward alternate KV engines right now.

### Release note 

```release-note
None
```

Issue #10528, opened 2021-07-06
Make it possible to create a tikv-server with a non-rocks KV engine


brson:
## Development Task

We're one or two steps away from having a tikv-server that can launch with a non-rocks KV engine. This is an issue to track that milestone.

Note that even once this is done, we won't actually be close to having a working alternative KV engine. We'll just be able to typecheck tikv-server with PanicEngine.

PR #10945, opened 2021-09-13
Make tikv-server fully generic over kv engines

brson:

https://github.com/tikv/tikv/issues/10528

Problem Summary:

Make tikv-server startup generic over engines.

This is a rewrite of previous pr https://github.com/tikv/tikv/pull/10529. This iteration has a cleaner history, but ends at basically the same place.

The previous patch bitrotted badly after the addition of FlowController, which is rocks-specific. This patch makes it so that FlowController should work if not set up. It does this by using the FlowController::empty ctor, which is documented as "test only", but would seem to create a basically-working FlowController.

The rest of the series is moving routines that are engine-specific into a new trait, CreateKvEngine.

After this patch, a second engine can be added in a single commit. I have not done this yet to avoid the question of whether and how to launch PanicEngine, which as a feature is useless, but is the only other in-tree implementation of KvEngine.

### Release note 

```release-note
None
```


brson:
I intend a future patch to convert the raft engine setup code to use a CreateRaftEngine trait, like the CreateKvEngine crate in this PR. I think it will be a simplification.

brson:
I went ahead and pushed the commits to add the CreateRaftEngine trait. This makes the setup routine a generic function instead of a macro. So after this patch, kv engines need to implement CreateKvEngine and raft engines CreateRaftEngine.

brson:
With the addition of CreateRaftEngine, the patch to enable PanicEngine for both KvEngine and RaftEngine is not so ugly, so I pushed that as well.

After this patch it is possible to launch with PanicEngine as either the KV engine or the raft engine.

components/server/src/server.rs
```diff
@@ -1218,38 +1207,76 @@ impl<ER: RaftEngine> TiKVServer<ER> {
...
+trait CreateKvEngine<ER>: KvEngine
+where
+    ER: RaftEngine,
+{
+    fn create_kv_engine(
+        config: &TiKvConfig,
+        region_info_accessor: &RegionInfoAccessor,
+        store_path: &Path,
+        router: &RaftRouter<Self, ER>,
+        env: Arc<engine_rocks::raw::Env>,
```
Connor1996, line 1222: it's still specific to rocksDB

brson: Yes, that's right. There are several features that remain rocks-specific for now, including the debug and backup services and this encryption configuration. This patch enables non-rocks engines to ignore those features and still boot.

brson: It's not in this series, but a patch that introduces other engines implements some of these CreateKvEngine methods as no-ops.

Connor1996 (approved):
LGTM

brson:
/test

please test?

zhouqiang-cl:
@Mini256 please take a look at the invaid-commit-message or give a guide how to handle with invalid commit message

Mini256:
> [FORMAT CHECKER NOTIFICATION]
> 
> Please provide the associated issue number in the commit messages, for example: `close #12345`, or `ref #12345`.

@brson, Hi, because the [vote-0150](https://github.com/tikv/community/blob/master/votes/0150-pr-issue-association.md) passed, the process of submitting PR has changed some changes: We need to provide the issue number associated with PR in the commit message.

To pass the format checker, you can attach the issue number in the previous commit message through the `git commit --amend` command, If the associated issue will be closed by merged PR, you can use `close #123`, if you just reference the related issue, you can use `ref #456`, for example:

```
server: some message
close #123
```

The format checker will ensure that at least one commit‘s message carries the issue number before the Pull Request is merged. 

Before a new Pull Request is submitted, we can fill in the `close # <issue_number>` or `ref # <issue_number>` by `git commit -s -a` command in advance. For example:

image

kennytm:
/test

https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/11840/pipeline

`cases::test_gc_worker::test_collect_applying_locks` (#8691?)

brson:
Thanks @zhouqiang-cl . I'll look into it.

PR #10946, opened 2021-09-14
Tikv: add comment [PerformanceCriticalPath] on these functions in critical path

tonyxuqqi:

Identifies these functions in critical path of read/write, makes developer be aware of the performance impact when changing code. 

Problem Summary:
 Sometimes developers may not be aware that a code change could lead to huge performance regression in specific scenarios. Totally rely on performance test costs a lots of effort and sometimes may not work due to incomplete test scenarios.

What's Changed:
Provide the best-effort hint on functions involved in critical path of read/write.

Tests 

- No code

### DIFF
--- a/PERFORMANCE_CRITICAL_PATH.md
+++ b/PERFORMANCE_CRITICAL_PATH.md
@@ -1,0 +1,9 @@
+# Performance Critical Path of read or write requests
+There're some functions that are in the critical path of read or write requests. They're so important to the overall performance that any regression will directly impact user experience. We added a comment [PerformanceCriticalPath] to highlight these critical functions. Please note that not all functions in critical path are marked, for example many short but very common functions are not marked. So please pay special attention when you change code that is itself marked as [PerformanceCriticalPath] or called by functions marked as [PerformanceCriticalPath]. 
+Here're some typical mistakes that should be avoided in these functions.
+* Unnecessary synchronous I/O.
+* Verbose logging with infor or above log level.
+* Global lock.
+* Long tasks that do not have to be synchronous.(Could be done in background thread instead.)
+
+Here's the [performance map](https://github.com/pingcap/tidb-map/blob/master/maps/performance-map.png) that would help you to understand how read/write requests are executed inside Tikv. 
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -567,6 +567,7 @@ where
         PeerFsmDelegate { fsm, ctx }
     }
 
+    // [PerformanceCriticalPath]
     pub fn handle_msgs(&mut self, msgs: &mut Vec<PeerMsg<EK>>) {
         for m in msgs.drain(..) {
             match m {
@@ -637,6 +638,7 @@ where
                         match self.fsm.peer.maybe_destroy(self.ctx) {
                             None => self.ctx.raft_metrics.message_dropped.applying_snap += 1,
                             Some(job) => {
+                                // [PerformanceCriticalPath]?? could it be done async?
                                 self.handle_destroy_peer(job);
                             }
                         }
@@ -664,6 +666,7 @@ where
         }
     }
 
+    // [PerformanceCriticalPath]
     fn on_casual_msg(&mut self, msg: CasualMessage<EK>) {
         match msg {
             CasualMessage::SplitRegion {
@@ -698,6 +701,7 @@ where
                 self.on_schedule_half_split_region(&region_epoch, policy, source);
             }
             CasualMessage::GcSnap { snaps } => {
+                // [PerformanceCriticalPath]?? should it be done in background thread?
                 self.on_gc_snap(snaps);
             }
             CasualMessage::ClearRegionSize => {
@@ -717,11 +721,13 @@ where
                 }
             }
             CasualMessage::SnapshotGenerated => {
+                // [PerformanceCriticalPath]?? should it be done in background thread? 
                 // Resume snapshot handling again to avoid waiting another heartbeat.
                 self.fsm.peer.ping();
                 self.fsm.has_ready = true;
             }
             CasualMessage::ForceCompactRaftLogs => {
+                // [PerformanceCriticalPath]?? should it be done in background thread? 
                 self.on_raft_gc_log_tick(true);
             }
             CasualMessage::AccessPeer(cb) => cb(self.fsm as &mut dyn AbstractPeer),
@@ -919,6 +925,7 @@ where
         );
     }
 
+    // [PerformanceCriticalPath]
     fn on_significant_msg(&mut self, msg: SignificantMsg<EK::Snapshot>) {
         match msg {
             SignificantMsg::SnapshotStatus {
@@ -1041,6 +1048,7 @@ where
         }
     }
 
+    // [PerformanceCriticalPath]
     pub fn collect_ready(&mut self) {
         let has_ready = self.fsm.has_ready;
         self.fsm.has_ready = false;
@@ -1063,6 +1071,7 @@ where
         }
     }
 
+    // [PerformanceCriticalPath]
     pub fn post_raft_ready_append(&mut self, ready: CollectedReady) {
         if ready.ctx.region_id != self.fsm.region_id() {
             panic!(
@@ -1251,6 +1260,7 @@ where
         }
     }
 
+    // [PerformanceCriticalPath]
     fn on_apply_res(&mut self, res: ApplyTaskRes<EK::Snapshot>) {
         fail_point!("on_apply_res", |_| {});
         match res {
@@ -1285,6 +1295,7 @@ where
             } => {
                 assert_eq!(peer_id, self.fsm.peer.peer_id());
                 if !merge_from_snapshot {
+                    // [PerformanceCriticalPath]?? should it be done async? 
                     self.destroy_peer(false);
                 } else {
                     // Wait for its target peer to apply snapshot and then send `MergeResult` back
@@ -1306,6 +1317,7 @@ where
         }
     }
 
+    // [PerformanceCriticalPath]
     fn handle_reported_disk_usage(&mut self, msg: &RaftMessage) {
         // Mocked
         if matches!(msg.disk_usage, DiskUsage::Normal) {
@@ -1346,6 +1358,7 @@ where
         }
     }
 
+    // [PerformanceCriticalPath]
     fn on_raft_message(&mut self, msg: InspectedRaftMessage) -> Result<()> {
         let InspectedRaftMessage { heap_size, mut msg } = msg;
         let stepped = Cell::new(false);
@@ -1420,6 +1433,9 @@ where
         }
 
         let is_snapshot = msg.get_message().has_snapshot();
+
+        // [PerformanceCriticalPath]??
+        // Can delete_snapshot be put in background thread?
         let regions_to_destroy = match self.check_snapshot(&msg)? {
             Either::Left(key) => {
                 // If the snapshot file is not used again, then it's OK to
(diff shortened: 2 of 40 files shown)


yiwu-arbug:
@BusyJay @Connor1996 mind take a look?

PERFORMANCE_CRITICAL_PATH.md
```diff
@@ -0,0 +1,9 @@
+# Performance Critical Path of read or write requests
+There're some functions that are in the critical path of read or write requests. They're so important to the overall performance that any regression will directly impact user experience. We added a comment [PerformanceCriticalPath] to highlight these critical functions. Please note that not all functions in critical path are marked, for example many short but very common functions are not marked. So please pay special attention when you change code that is itself marked as [PerformanceCriticalPath] or called by functions marked as [PerformanceCriticalPath]. 
+Here're some typical mistakes that should be avoided in these functions.
+* Unnecessary synchronous I/O.
+* Verbose logging with infor or above log level.
```
BusyJay, line 5:
```suggestion
* Verbose logging with info or above log level.
```

BusyJay (commented):
These comments may not be easy to maintain with future refactoring. Every constraints depends on human work is unreliable unfortunately.

PERFORMANCE_CRITICAL_PATH.md
```diff
@@ -0,0 +1,9 @@
+# Performance Critical Path of read or write requests
+There're some functions that are in the critical path of read or write requests. They're so important to the overall performance that any regression will directly impact user experience. We added a comment [PerformanceCriticalPath] to highlight these critical functions. Please note that not all functions in critical path are marked, for example many short but very common functions are not marked. So please pay special attention when you change code that is itself marked as [PerformanceCriticalPath] or called by functions marked as [PerformanceCriticalPath].
```
Connor1996, line 2:
```suggestion
There're some functions that are in the critical path of read or write requests. They're so important to the overall performance that any regression will directly impact user experience. We add a comment [PerformanceCriticalPath] to highlight these critical functions. Please note that not all functions in the critical path are marked, for example, many short but very common functions are not marked. So please pay special attention when you change code that is itself marked as [PerformanceCriticalPath] or called by functions marked as [PerformanceCriticalPath]. 
```

Connor1996 (commented):
Seems almost all the functions called in raftstore and apply loop should be marked as `PerformanceCriticalPath`. Though we mark the upper-level functions, it has a weak effort to remind someone writing the subfunctions of the function.  Also, as Jay said, it's hard to maintain.

tonyxuqqi:
> These comments may not be easy to maintain with future refactoring. Every constraints depends on human work is unreliable unfortunately.

That’s true it’s hard to maintain. But it will just make us to think harder regarding if it’s on the critical path. It’s an best effort thing for now before we come up an automatic test to find out problematic code. I think it’s still better to have them.

PERFORMANCE_CRITICAL_PATH.md
```diff
@@ -0,0 +1,10 @@
+# Performance Critical Path of user requests
+There're many files whose functions are in the critical path of read or write requests. They're so important to the overall performance that any regression will directly impact user experience. We add a comment #[PerformanceCriticalPath] to highlight these critical files. Please note that this is the best-effort work and some files in critical path may not be marked. But if a file is marked,  please pay special attention when you change its code.
+
+Here're some typical mistakes that should be avoided in the #[PerformanceCriticalPath] files.
+* Unnecessary synchronous I/O. Here 'unnecessary' means it's not a MUST for serving the current user request. For example, on_gc_snap() in peers.rs should not be in Fsm message handling.
```
BusyJay, line 5: > on_gc_snap() in peers.rs should not be in Fsm message handling

This can be controversial. Some safety check cannot be done without accessing peer's field.

PERFORMANCE_CRITICAL_PATH.md
```diff
@@ -0,0 +1,10 @@
...
+* Verbose logging with info or above log level.
+* Global lock.
+* Long tasks that do not have to be synchronous (Could be done in background thread instead).
+
+Here's the [performance map](https://github.com/pingcap/tidb-map/blob/master/maps/performance-map.png) that would help you to understand how read/write requests are executed inside Tikv.
```
BusyJay, line 10: Can TiKV has its own version and hosted in TiKV repository?

components/raftstore/src/coprocessor/consistency_check.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
+// #[PerformanceCriticalPath]
```
BusyJay, line 3: It's not.

components/raftstore/src/store/async_io/write_router.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
+// #[PerformanceCriticalPath]
```
BusyJay, line 3: I believe functions in this file are expected to block.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -637,6 +638,7 @@ where
                         match self.fsm.peer.maybe_destroy(self.ctx) {
                             None => self.ctx.raft_metrics.message_dropped.applying_snap += 1,
                             Some(job) => {
+                                // [PerformanceCriticalPath]?? could it be done async?
```
BusyJay, line 641: How about leaving a TODO instead since it's stated at the header of the file. And it should be added at the definition of `destroy_peer` instead of spreading everywhere.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -698,6 +700,7 @@ where
                 self.on_schedule_half_split_region(&region_epoch, policy, source);
             }
             CasualMessage::GcSnap { snaps } => {
+                // [PerformanceCriticalPath]?? should it be done in background thread?
```
BusyJay, line 703: No.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -717,11 +720,13 @@ where
                 }
             }
             CasualMessage::SnapshotGenerated => {
+                // [PerformanceCriticalPath]?? should it be done in background thread?
```
BusyJay, line 723: It doesn't block.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -717,11 +720,13 @@ where
...
                 // Resume snapshot handling again to avoid waiting another heartbeat.
                 self.fsm.peer.ping();
                 self.fsm.has_ready = true;
             }
             CasualMessage::ForceCompactRaftLogs => {
+                // [PerformanceCriticalPath]?? should it be done in background thread?
```
BusyJay, line 729: It doesn't block.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1285,6 +1290,7 @@ where
             } => {
                 assert_eq!(peer_id, self.fsm.peer.peer_id());
                 if !merge_from_snapshot {
+                    // [PerformanceCriticalPath]?? should it be done async?
```
BusyJay, line 1293: How about a `TODO`?

components/raftstore/src/store/peer.rs
```diff
@@ -2158,6 +2159,7 @@ where
         );
 
         if !self.is_leader() {
+            // [PerformanceCriticalPath]?? should it be done async?
```
BusyJay, line 2162: It doesn't block.

components/raftstore/src/store/worker/consistency_check.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
 
+// #[PerformanceCriticalPath] called by raftstore/coprocessor
```
BusyJay, line 3: `worker` is background threads that critical path offload work to.

PERFORMANCE_CRITICAL_PATH.md
tonyxuqqi: refined the wording.

PERFORMANCE_CRITICAL_PATH.md
tonyxuqqi: good point.

components/raftstore/src/coprocessor/consistency_check.rs
tonyxuqqi: Actually the whole file is to serve its compute_hash function, which is called eventually by PeerMsg::ApplyRes handler, which is on the critical path.

components/raftstore/src/store/async_io/write_router.rs
tonyxuqqi: Yes, they should block, but the point here is to keep us aware do not add unnecessary code here.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: sounds good.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: You're right. ping is not a blocking API. Thanks.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: ok

components/raftstore/src/store/peer.rs
tonyxuqqi: ok

components/raftstore/src/coprocessor/consistency_check.rs
tonyxuqqi: ok, looks that it's indeed run in background.

PERFORMANCE_CRITICAL_PATH.md
```diff
@@ -0,0 +1,10 @@
+# Performance Critical Path of user requests
+There're many files whose functions are in the critical path of read or write requests. They're so important to the overall performance that any regression will directly impact user experience. We add a comment #[PerformanceCriticalPath] to highlight these critical files. Please note that this is the best-effort work and some files in critical path may not be marked. But if a file is marked,  please pay special attention when you change its code.
+
+Here're some typical mistakes that should be avoided in the #[PerformanceCriticalPath] files.
+* Unnecessary synchronous I/O. Here 'unnecessary' means it's not a MUST for serving the current user request. For example, on_gc_snap() in peers.rs should spin off its I/O related work to background thread.
+* Verbose logging with info or above log level.
+* Global lock.
+* Long tasks that do not have to be synchronous (Could be done in background thread instead).
+
+Here's the [performance map](Tikv_Map.png) that would help you to understand how read/write requests are executed inside Tikv.
```
BusyJay, line 10: Better put it under images directory.

src/storage/mvcc/consistency_check.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
+// #[PerformanceCriticalPath] called by raftstore
```
BusyJay, line 3: Isn't it called in worker?

src/storage/mvcc/reader/scanner/backward.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
+// #[PerformanceCriticalPath
```
BusyJay, line 3: Missing `]`.

Issue #10947, opened 2021-09-15
TiKV panic during import data because of RocksDB Corruption.


Little-Wallace:
### What version of TiKV are you using?

master

### Steps to reproduce

![rocksdb](https://user-images.githubusercontent.com/16376959/133368509-46cad931-9f5c-4414-89f5-5bbea4b64219.png)

### What did happened?

TiKV panic and can not restart again

PR #10948, opened 2021-09-15
rocksdb: fix panic caused by ingest multiple files

Little-Wallace:

Issue Number: close https://github.com/tikv/tikv/issues/10947

Problem Summary:
TiKV will panic after ingest two sst files in once call. Because these two files are both in L0.

Proposal: [xxx](url) 

What's Changed:

### Related changes

- PR to update `tikv/rocksdb`: https://github.com/tikv/rocksdb/pull/257

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Fix rocksdb panic after ingesting two files in once call.
```


Connor1996 (approved):
LGTM

Little-Wallace:
I check the code again and find that this bug will not occur when the binary tikv-server  is compile in release mode.

hunterlxt (approved):

PR #10951, opened 2021-09-15
status_server: add pprof flamegraph response header

shhdgit:
Signed-off-by: Suhaha <jklopsdfw@gmail.com>

Problem Summary:

No related PR. The response of svg returned by the status server's pprof does not have a corresponding Content-Type header. And the correct return header makes it easier for different clients to handle the mime type.

What's Changed:

Add the correct Content-Type to response header. 

### Related changes

### Check List 

Tests 

- Manual test
```
curl -v http://127.0.0.1:20180/debug/pprof/profile\?
```
will see the response header with Content-Type
```
< HTTP/1.1 200 OK
< content-type: image/svg+xml
< content-length: 187695
< date: Wed, 15 Sep 2021 15:49:13 GMT
```

Side effects

### Release note 

```release-note
Add the correct Content-Type to `debug/pprof/profile` response header. 
```


sticnarf (approved):
Looks great!

It would be better if you add a test for it in `tests/integrations/server/status_server.rs`.

shhdgit:
Sure, let me try.

shhdgit:
I found a test case `test_pprof_profile_service` for pprof below, can I just add it here.

sticnarf:
Please take another look @hunterlxt

5kbpers (approved):

PR #10953, opened 2021-09-15
raftstore: add rate limit for snapshot generation when using plain cf file.

### LINKED ISSUE #10930: If there are too many snapshots, node can run out of space under low load

v5.1.1

linux

### Steps to reproduce

Delete a node. And make receiving snapshot very slow, which can cause a lot of operator in PD timeout. So there will a lot of on-going snapshots.

When there are a lot of snapshot, sending will be aborted and snapshot files will be left in disk. It will be deleted 4 hours later by default or when leader has compacted to a larger index. When under low load, leader can keep proposing requests from time to time, so new snapshots will always be regenerated. And leader may not trigger log compactions in a long time. Because the rate limit of generating snapshot is about 100MiB/s, so 4 hours can generate about 1.4TiB snapshots in theory.

In practice, I have observed about more than 10k snapshots generated.

### What did you expect?
Snapshot files count in the directory should be reasonably small.

### What did happened?
A lot of pending snapshot files that starts with `gen_` prefix and different index suffix.

tonyxuqqi:
Also changes the default snap_max_total_size from u64 MAX to 50GB

Issue Number: close #10930

Problem Summary:
1) Today we have the snapshot send/receive RateLimiter, and we also have rate limiter for SST snapshot file generation , but not for plain cf file.
2) The default value of snap_max_total_size is 0, which translates to u64 Max in the logic. And that means under default value, the rate limit for snapshot file generation does not take effect at all. That could lead to problem 10930.

What's Changed:
1) Add the rate limiter for plain cf snapshot file generation as well. 
2) Change the default value of snap_max_total_size from 0 to 50GIB.

- Unit test

```release-note
Default value of snap_max_total_size is changed from 0 to 50GIB.
```

### DIFF
--- a/components/raftstore/src/store/snap.rs
+++ b/components/raftstore/src/store/snap.rs
@@ -691,7 +691,13 @@ impl Snapshot {
             let cf_stat = if plain_file_used(cf_file.cf) {
                 let key_mgr = self.mgr.encryption_key_manager.as_ref();
                 snap_io::build_plain_cf_file::<EK>(
-                    path, key_mgr, kv_snap, cf_file.cf, &begin_key, &end_key,
+                    path,
+                    key_mgr,
+                    kv_snap,
+                    cf_file.cf,
+                    &begin_key,
+                    &end_key,
+                    &self.mgr.limiter,
                 )?
             } else {
                 snap_io::build_sst_cf_file::<EK>(
--- a/components/raftstore/src/store/snap.rs
+++ b/components/raftstore/src/store/snap.rs
@@ -691,13 +691,7 @@ impl Snapshot {
             let cf_stat = if plain_file_used(cf_file.cf) {
                 let key_mgr = self.mgr.encryption_key_manager.as_ref();
                 snap_io::build_plain_cf_file::<EK>(
-                    path,
-                    key_mgr,
-                    kv_snap,
-                    cf_file.cf,
-                    &begin_key,
-                    &end_key,
-                    &self.mgr.limiter,
+                    path, key_mgr, kv_snap, cf_file.cf, &begin_key, &end_key,
                 )?
             } else {
                 snap_io::build_sst_cf_file::<EK>(
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -9,7 +9,7 @@ use regex::Regex;
 
 use collections::HashMap;
 use online_config::{ConfigChange, ConfigManager, OnlineConfig};
-use tikv_util::config::{self, ReadableDuration, ReadableSize, VersionTrack};
+use tikv_util::config::{self, ReadableDuration, ReadableSize, VersionTrack, GIB};
 use tikv_util::sys::SysQuota;
 use tikv_util::worker::Scheduler;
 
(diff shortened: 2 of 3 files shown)


components/raftstore/src/store/snap.rs
```diff
@@ -691,7 +691,13 @@ impl Snapshot {
...
                 let key_mgr = self.mgr.encryption_key_manager.as_ref();
                 snap_io::build_plain_cf_file::<EK>(
-                    path, key_mgr, kv_snap, cf_file.cf, &begin_key, &end_key,
+                    path,
+                    key_mgr,
+                    kv_snap,
+                    cf_file.cf,
+                    &begin_key,
+                    &end_key,
+                    &self.mgr.limiter,
```
BusyJay, line 700: Only lock cf will use plain cf. And lock cf is usually very small, like several bytes to several kilobytes.

src/server/config.rs
```diff
@@ -232,7 +232,7 @@ impl Default for Config {
             ),
             end_point_max_concurrency: cmp::max(cpu_num as usize, MIN_ENDPOINT_MAX_CONCURRENCY),
             snap_max_write_bytes_per_sec: ReadableSize(DEFAULT_SNAP_MAX_BYTES_PER_SEC),
-            snap_max_total_size: ReadableSize(0),
+            snap_max_total_size: ReadableSize(50 * GIB),
```
BusyJay, line 235: Can you explain why it's `50GiB`?

components/raftstore/src/store/snap.rs
tonyxuqqi: But technically plain cf can be configured for write cf as well right?
I think it's better to have a complete coverage. It does not hurt at least.

src/server/config.rs
tonyxuqqi: The size consideration is that:
1) It should be reasonable big enough so that it can have a few snapshot files being generated at parallel. 
2) But it should not too big otherwise it would waste disk space when snapshots fail to be sent.

src/server/config.rs
```diff
@@ -232,7 +236,7 @@ impl Default for Config {
             ),
             end_point_max_concurrency: cmp::max(cpu_num as usize, MIN_ENDPOINT_MAX_CONCURRENCY),
             snap_max_write_bytes_per_sec: ReadableSize(DEFAULT_SNAP_MAX_BYTES_PER_SEC),
-            snap_max_total_size: ReadableSize(0),
+            snap_max_total_size: ReadableSize(DEFAULT_SNAP_MAX_TOTAL_SIZE),
```
BusyJay, line 239: Can it be adjusted according to the region size? For example, if we want to allow 8 concurrent snapshots but each snapshot can take 10 GiB, then the maximum size should be 80GiB.

BusyJay (commented):
Is `snap_max_total_size` covered by any test case?

src/server/config.rs
tonyxuqqi: mm, but for smaller region, we probably want to have bigger concurrent snapshots to speed up the process. 
I think use a size is a more clear mechanism and easier to understand.

tonyxuqqi:
> Is `snap_max_total_size` covered by any test case?

Yes, for example, test_snapshot_max_total_size,

BusyJay:
As I remember, TiKV will panic if it fails to retrieve snapshot more than 5 times. If a failed node return after a long time, all peers will have to accept snapshots to capture up logs. Will it cause other TiKV to panic with 50GiB?

PR #10961, opened 2021-09-17
storage: support automatic sst corruption check and deletion

### LINKED ISSUE #10578: storage: support automatic sst corruption recovery
## Development Task
When SST files are corrupted or inaccessible, TiKV would panic and cannot start normally which would block the user’s business. This work is to implement automatic sst corruption recovery while tikv server doesn't need be shutdown.

hunterlxt:

Issue Number: ref #10578

Problem Summary:

When SST files are corrupted or inaccessible, TiKV would panic and cannot start normally which would block the user’s business. This work is part of #10578. This is TiKV part which means that TiKV don't panic immediately but alive a period of time. During this period, PD will get the stats through store heartbeat to add evict leader scheduler. Users need to fix the sst error and manually remove evict-leader-scheduler with pd-ctl.

Proposal: #10578

Here is a test case to show you how does this feature work:

image

image

image

image

image

And PD can remove this peer after receiving store heartbeat. Like this example, after remove region 180 on store 1. You can see damaged files is 0 after auto deleteion.

```
[2022/04/27 12:04:49.381 +08:00] [ERROR] [debug.rs:35] ["debug_compact failed: RemoteStopped"]
[2022/04/27 12:04:54.467 +08:00] [WARN] [store.rs:211] ["detected damaged regions overlapping damaged file ranges"] [id={180}]
[2022/04/27 12:04:54.471 +08:00] [INFO] [pd.rs:1398] ["try to transfer leader"] [to_peers="[]"] [to_peer="id: 183 store_id: 3"] [from_peer="id: 181 store_id: 1"] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [peer.rs:3351] ["transfer leader"] [peer="id: 183 store_id: 3"] [peer_id=181] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [raft.rs:1904] ["[term 6] starts to transfer leadership to 183"] [lead_transferee=183] [term=6] [raft_id=181] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [raft.rs:1917] ["sends MsgTimeoutNow to 183 immediately as 183 already has up-to-date log"] [lead_transferee=183] [raft_id=181] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [raft.rs:1364] ["received a message with higher term from 183"] ["msg type"=MsgRequestVote] [message_term=7] [term=6] [from=183] [raft_id=181] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [raft.rs:1120] ["became follower at term 7"] [term=7] [raft_id=181] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [raft.rs:1565] ["[logterm: 6, index: 6, vote: 0] cast vote for 183 [logterm: 6, index: 6] at term 7"] ["msg type"=MsgRequestVote] [term=7] [msg_index=6] [msg_term=6] [from=183] [vote=0] [log_index=6] [log_term=6] [raft_id=181] [region_id=180]
[2022/04/27 12:04:54.472 +08:00] [INFO] [endpoint.rs:388] ["deregister observe region"] [observe_id=ObserveID(86)] [region_id=180] [store_id=Some(1)]
[2022/04/27 12:04:54.474 +08:00] [INFO] [apply.rs:1385] ["execute admin command"] [command="cmd_type: ChangePeer change_peer { change_type: RemoveNode peer { id: 181 store_id: 1 } }"] [index=8] [term=7] [peer_id=181] [region_id=180]
[2022/04/27 12:04:54.474 +08:00] [INFO] [apply.rs:1759] ["exec ConfChange"] [epoch="conf_ver: 5 version: 45"] [type=RemoveNode] [peer_id=181] [region_id=180]
[2022/04/27 12:04:54.474 +08:00] [INFO] [apply.rs:1867] ["remove peer successfully"] [region="id: 180 start_key: 7480000000000000FF515F728000000000FF059F200000000000FA end_key: 7480000000000000FF515F728000000000FF093E270000000000FA region_epoch { conf_ver: 5 version: 45 } peers { id: 181 store_id: 1 } peers { id: 182 store_id: 7 } peers { id: 183 store_id: 3 }"] [peer="id: 181 store_id: 1"] [peer_id=181] [region_id=180]

[2022/04/27 12:04:56.473 +08:00] [WARN] [store.rs:1991] ["Damaged region overlapped and reject to create peer"] [region_id=180] [peer_id="id: 324 store_id: 1 role: Learner"]
[2022/04/27 12:04:58.475 +08:00] [WARN] [store.rs:1991] ["Damaged region overlapped and reject to create peer"] [region_id=180] [peer_id="id: 324 store_id: 1 role: Learner"]
[2022/04/27 12:05:00.002 +08:00] [INFO] [profiler.rs:299] ["starting cpu profiler"]
[2022/04/27 12:05:00.477 +08:00] [WARN] [store.rs:1991] ["Damaged region overlapped and reject to create peer"] [region_id=180] [peer_id="id: 324 store_id: 1 role: Learner"]
[2022/04/27 12:05:02.479 +08:00] [WARN] [store.rs:1991] ["Damaged region overlapped and reject to create peer"] [region_id=180] [peer_id="id: 324 store_id: 1 role: Learner"]
[2022/04/27 12:05:03.342 +08:00] [WARN] [sst_recovery.rs:178] ["damaged file has been deleted"] [largest_key="[122, 116, 128, 0, 0, 0, 0, 0, 0, 255, 81, 95, 114, 128, 0, 0, 0, 0, 255, 9, 62, 38, 0, 0, 0, 0, 0, 250, 249, 254, 91, 86, 19, 139, 255, 252]"] [smallest_key="[122, 116, 128, 0, 0, 0, 0, 0, 0, 255, 81, 95, 114, 128, 0, 0, 0, 0, 255, 5, 159, 32, 0, 0, 0, 0, 0, 250, 249, 254, 91, 86, 81, 67, 255, 253]"] [file=/000383.sst]
[2022/04/27 12:05:03.477 +08:00] [INFO] [peer.rs:289] ["replicate peer"] [peer_id=324] [region_id=180]
[2022/04/27 12:05:03.477 +08:00] [INFO] [raft.rs:2646] ["switched to configuration"] [config="Configuration { voters: Configuration { incoming: Configuration { voters: {} }, outgoing: Configuration { voters: {} } }, learners: {}, learners_next: {}, auto_leave: false }"] [raft_id=324] [region_id=180]
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

None

### Release note 

```release-note
Support the checking of some rocksdb background errors and reporting of damaged regions. Such errors will not panic TiKV immediately.
```


components/engine_rocks/src/error_listener.rs
```diff
@@ -0,0 +1,83 @@
...
+                if err.starts_with("Corruption") {
+                    set_panic_mark();
+                }
+                panic!(
+                    "rocksdb background error. db: {}, reason: {}, error: {}",
+                    self.db_name, r, err
+                );
+            } else {
+                // enable auto sst recovery mode.
+                if err.starts_with("Corruption") || err.starts_with("IO error") {
```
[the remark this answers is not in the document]
hunterlxt: only detect "Corruption" for now

components/engine_rocks/Cargo.toml
```diff
@@ -69,9 +69,11 @@ kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = f
 raft = { version = "0.6.0-alpha", default-features = false }
 protobuf = "2"
 fail = "0.4"
+regex = "1"
 
 [dependencies.rocksdb]
-git = "https://github.com/tikv/rust-rocksdb.git"
+git = "https://github.com/hunterlxt/rust-rocksdb.git"
```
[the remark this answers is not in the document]
hunterlxt: updated to a new branch

components/engine_rocks/src/error_listener.rs
```diff
@@ -0,0 +1,83 @@
...
+                    if let Ok(sst) = resolve_sst_number_from_err(&err) {
+                        if self.recovery_tx.as_ref().unwrap().schedule(sst).is_ok() {
+                            // todo: metric to record
```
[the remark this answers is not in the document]
hunterlxt: done

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,138 @@
...
+            set_panic_mark_and_panic(sst, "Can't find overlap regions");
+        }
+
+        for id in regions {
+            let region_meta = block_on(self.pd_client.get_region_by_id(id)).unwrap();
+            self.scheduling_regions.insert(id, Instant::now());
+        }
+    }
+
+    fn check_scheduling_tasks(&mut self) {}
```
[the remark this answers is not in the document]
hunterlxt: done

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,138 @@
...
+            let exist_region = &meta.regions[id];
+            if keys::enc_end_key(exist_region) <= end_key.to_owned() {
+                regions.push(*id);
+            }
+        }
+        regions
+    }
+}
+
+fn run() {}
```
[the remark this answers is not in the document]
hunterlxt: redundant removed

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,138 @@
...
+                    self.check_scheduling_tasks();
+                }
+            }
+        }
+    }
+}
+
+impl RecoveryJob {
+    fn generate_task(&mut self, sst: u64) {
+        self.db.resume().unwrap();
```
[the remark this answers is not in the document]
hunterlxt: Repeated errors will occur again I guess, but eventually they will all converge to `scheduling_tasks` and check. Maybe I need add a global timer to the entire recovery job to prevent repeated errors from happening, so that tikv will never panic

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,156 @@
...
+const CHECK_DURATION: Duration = Duration::from_secs(30);
+
+struct RegionInfo {
+    name: String,
+    smallestkey: Vec<u8>,
+    largestkey: Vec<u8>,
+    instant: Instant,
+}
+
+// pub fn create_sst_recovery_worker(
```
Connor1996, line 29: clean it

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    fn check_and_finish_scheduling_tasks(&mut self) -> bool {
+        let meta = self.meta.lock().unwrap();
+        let flag= true;
+        for r in self.scheduling_regions {
+            if r.instant.elapsed() > CHECK_DURATION {
+                set_panic_mark_and_panic(r.name, "recovery job exceeded CHECK_DURATION");
+            }
+            let overlap = self.check_and_notify_damaged_regions(small, large);
```
Connor1996, line 129: see no place to drop the damaged SST.

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,156 @@
...
+        let mut flag = false;
+        for (_, id) in meta
+            .region_ranges
+            .range((Included(start_key.to_owned()), Unbounded::<Vec<u8>>))
+        {
+            let exist_region = &meta.regions[id];
+            if keys::enc_end_key(exist_region) <= end_key.to_owned() {
+                meta.damaged_regions_id.insert(*id);
+                flag = true
+            }
```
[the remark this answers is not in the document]
hunterlxt: don't we need check all regions in store_meta?

components/engine_rocks/src/error_listener.rs
```diff
@@ -0,0 +1,93 @@
...
+                // Ignore NoSpace error and let RocksDB automatically recover.
+                return;
+            }
+
+            let r = match reason {
+                DBBackgroundErrorReason::Flush => "flush",
+                DBBackgroundErrorReason::Compaction => "compaction",
+                DBBackgroundErrorReason::WriteCallback => "write_callback",
+                DBBackgroundErrorReason::MemTable => "memtable",
+            };
```
MrCroxx, line 47: What about move these to the impl block of `DBBackgroundErrorReason`?

hunterlxt:
[PD part](https://github.com/tikv/pd/pull/4709) has been merged before, but it will not be triggered if this PR not merged @Connor1996

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,248 @@
...
+                }
+
+                return;
+            }
+        }
+    }
+
+    // `sst_path` has been processed and is still in a damaged state.
+    fn exist_scheduling_regions(&self, sst_path: &str) -> bool {
+        for region in &self.damaged_files {
```
Connor1996, line 102: self.damaged_files.iter().any(|f| f.name == sst_path)

components/engine_rocks/src/error_listener.rs
```diff
@@ -0,0 +1,95 @@
...
+use tikv_util::set_panic_mark;
+use tikv_util::worker::Scheduler;
+use tikv_util::{error, warn};
+
+// Message for RocksDB status subcode kNoSpace.
+const NO_SPACE_ERROR: &str = "IO error: No space left on device";
+
+// `ErrorListener` is responsible for implementing the function of `on_background_error`.
+// In order to avoid redundant panic, all rocksdb error processing should be implemented here.
+pub struct ErrorListener {
```
Connor1996, line 15: why not integrate in `EventListener`?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,248 @@
...
+
+use std::collections::Bound::{Excluded, Unbounded};
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use engine_rocks::raw::*;
+use raftstore::store::fsm::StoreMeta;
+use tikv_util::{self, set_panic_mark, worker::*};
+
+pub const CHECK_DURATION: Duration = Duration::from_secs(10);
```
tabokie, line 11:
```suggestion
pub const CHECK_INTERVAL: Duration = Duration::from_secs(10);
```

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,248 @@
...
+
+#[derive(Debug, Clone)]
+struct FileInfo {
+    // Corrupted file name. example: /000033.sst
+    name: String,
+    smallestkey: Vec<u8>,
+    largestkey: Vec<u8>,
+    overlap_region_ids: Vec<u64>,
+    // Time to generate recovery task, be used to record whether the timeout
+    instant: Instant,
```
tabokie, line 29:
```suggestion
    start_time: Instant,
```

etc/config-template.toml
```diff
@@ -242,6 +242,13 @@
 ## Set it to 0 will cause no space is reserved at all. It's generally used for tests.
 # reserve-space = "5GB"
 
+## The maximum hang time after rocksdb detects a background error. When some data belonging to the
+## data range is damaged, it will be reported to PD, and PD will add `evict-leader-scheduler` in
+## this store to minimize the impact on user workload as much as possible.
+##
+## Set to 0 to disable this feature if you want to panic immediately when encountering such an error.
+# max-background-error-hang-time = "24h"
```
tabokie, line 250: I think the explanation is not clear enough. What's the behavior during "hang", and what's the behavior after it?

components/engine_rocks_helper/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```
Connor1996, line 1: 2022

components/server/src/server.rs
```diff
@@ -90,6 +91,7 @@ use tikv::{
         txn::flow_controller::FlowController, Engine,
     },
 };
+use tikv_util::worker::Scheduler;
```
Connor1996, line 94: move to L103

components/engine_rocks/src/error_listener.rs
```diff
@@ -0,0 +1,105 @@
...
+    db_name: String,
+    sst_recovery_tx: Option<Scheduler<String>>,
+}
+
+impl ErrorListener {
```
Connor1996, line 20: should be removed now?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+    // Considering that files will not be too much, it is enough to use `Vec`.
+    damaged_files: Vec<FileInfo>,
+    max_damage_duration: Duration,
+}
+
+#[derive(Debug, Clone)]
+struct FileInfo {
+    // Corrupted file name. example: /000033.sst
+    name: String,
+    smallestkey: Vec<u8>,
```
Connor1996, line 25: smallest_key

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+    largestkey: Vec<u8>,
```
Connor1996, line 26: largest_key

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+
+                // only support recovering data key for now.
+                if !keys::validate_data_key(&f.smallestkey)
+                    || (!keys::validate_data_key(&f.largestkey)
+                        && f.largestkey.as_slice() != keys::DATA_MAX_KEY)
+                {
+                    set_panic_mark_and_panic(
+                        path,
+                        &format!(
+                            "key range mismatch, small key:{:?}, large key:{:?}",
```
Connor1996, line 84: small -> smallest, large -> largest

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+    }
+
+    // Check whether the StoreMeta contains the region range, if it contains,
+    // recorded fault region ids to report to PD and add file info into `damaged_files`.
+    //
+    // Acquire meta lock.
+    fn check_overlap_damaged_regions(&mut self, mut file: FileInfo) -> bool {
+        let mut meta = match self.store_meta.lock() {
```
Connor1996, line 130: self.store_meta.lock().unwrap() is enough

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+
+        let mut overlap = false;
+
+        // Find overlapping region ids.
+        // Condition:
+        // end_key > file.smallestkey
+        // start_key <= file.largestkey
+        let mut ids = vec![];
+        for (_, id) in meta
+            .region_ranges
```
Connor1996, line 146: Does the keys of `region_ranges` have the data prefix?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn check_overlap_damaged_regions(&mut self, mut file: FileInfo) -> bool {
+        let mut meta = match self.store_meta.lock() {
+            Ok(meta) => meta,
+            Err(e) => {
+                set_panic_mark_and_panic(&file.name, &format!("{}", e));
+                return false;
+            }
+        };
+
+        let mut overlap = false;
```
Connor1996, line 138: no need, `ids.is_empty()` implies `overlap == true`

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+        for (_, id) in meta
+            .region_ranges
+            .range((Excluded(file.smallestkey.clone()), Unbounded::<Vec<u8>>))
+        {
+            let region = &meta.regions[id];
+            if keys::enc_start_key(region) <= file.largestkey {
+                ids.push(*id);
+                file.overlap_region_ids.push(*id);
+                overlap = true;
+            }
```
Connor1996, line 154: else break?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        if !overlap {
+            // TODO: User trigger to delete sst file here. `delete_file` only allow
+            // delete files in the last level, consider `delete_files_in_range`.
+            for id in ids {
+                meta.damaged_regions_id.remove(&id);
```
Connor1996, line 161: why?

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: yes, for example, we can found `meta.region_ranges.insert(enc_end_key(region), region_id);` and `enc_end_key` is implemented here
```
pub fn enc_end_key(region: &Region) -> Vec<u8> {
    // only initialized region's end_key can be encoded, otherwise there must be bugs
    // somewhere.
    assert!(!region.get_peers().is_empty());
    data_end_key(region.get_end_key())
}
```

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: This requires implicit ordering guarantees and there doesn't seem to be any performance impact without breaking early here(handling damaged regions is not a normal operation)

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: Did you have question on 「`delete_file` only allow delete files in the last level, consider `delete_files_in_range`.」?

hunterlxt:
I have a question, in the failpoint test, you will see that I generate a sst file with the same start and end key, and it will be broken to test the feature. But I found that when `delete_files_in_range` does not contain the ending key, the file cannot be deleted. Is this as expected? Do we need to set it to true? cc@Connor1996

Connor1996:
> I have a question, in the failpoint test, you will see that I generate a sst file with the same start and end key, and it will be broken to test the feature. But I found that when `delete_files_in_range` does not contain the ending key, the file cannot be deleted. Is this as expected? Do we need to set it to true? cc @Connor1996 @tabokie

Not sure, you can check the code of RocksDB. And set what to true?

hunterlxt:
> > I have a question, in the failpoint test, you will see that I generate a sst file with the same start and end key, and it will be broken to test the feature. But I found that when `delete_files_in_range` does not contain the ending key, the file cannot be deleted. Is this as expected? Do we need to set it to true? cc @Connor1996 @tabokie
> 
> Not sure, you can check the code of RocksDB. And set what to true?

set `include_end` (which is a parameter passed into `delete_files_in_range`) to true

components/engine_rocks_helper/src/sst_recovery.rs
Connor1996: What's the ordering guarantee? The `range()` is a guarantee

Connor1996:
> > > I have a question, in the failpoint test, you will see that I generate a sst file with the same start and end key, and it will be broken to test the feature. But I found that when `delete_files_in_range` does not contain the ending key, the file cannot be deleted. Is this as expected? Do we need to set it to true? cc @Connor1996 @tabokie
> > 
> > 
> > Not sure, you can check the code of RocksDB. And set what to true?
> 
> set `include_end` (which is a parameter passed into `delete_files_in_range`) to true

I think yes

hunterlxt:
> > > > I have a question, in the failpoint test, you will see that I generate a sst file with the same start and end key, and it will be broken to test the feature. But I found that when `delete_files_in_range` does not contain the ending key, the file cannot be deleted. Is this as expected? Do we need to set it to true? cc @Connor1996 @tabokie
> > > 
> > > 
> > > Not sure, you can check the code of RocksDB. And set what to true?
> > 
> > 
> > set `include_end` (which is a parameter passed into `delete_files_in_range`) to true
> 
> I think yes

in rocksdb, it will check the largest key of the file to be deleted. So same with you, i thinks the answer is yes.

```
        for (uint32_t j = 0; j < level_files.size(); j++) {
          level_file = level_files[j];
          if (level_file->being_compacted) {
            continue;
          }
          if (deleted_files.find(level_file) != deleted_files.end()) {
            continue;
          }
          if (!include_end && end != nullptr &&
              cfd->user_comparator()->Compare(level_file->largest.user_key(),
                                              *end) == 0) {
            continue;
          }
          edit.SetColumnFamily(cfd->GetID());
          edit.DeleteFile(i, level_file->fd.GetNumber());
          deleted_files.insert(level_file);
          level_file->being_compacted = true;
        }
```

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,256 @@
...
+                        &format!(
+                            "key range mismatch, smallest key:{:?}, largest key:{:?}",
+                            &f.smallest_key, &f.largest_key
+                        ),
+                    );
+                }
+
+                // lock the store_meta and check if the range exist.
+                if !self.check_overlap_damaged_regions(f) {
+                    set_panic_mark_and_panic(path, "Can't find overlap regions");
```
Connor1996, line 100: it maybe possible

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,256 @@
...
+
+        let overlap = !ids.is_empty();
+        if overlap && !self.exist_scheduling_regions(&file.name) {
+            // This sst file was detected for the first time so store meta and
+            // file context should be updated.
+            for id in ids {
+                meta.damaged_regions_id.insert(id);
+                file.overlap_region_ids.push(id);
+            }
+            self.damaged_files.push(file);
```
Connor1996, line 171: Add a metric for the number of damage files?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,256 @@
...
+                );
+            }
+            let need_retain = self.check_overlap_damaged_regions(f.clone());
+            if !need_retain {
+                println!(
+                    "ready delete_file {} {:?} {:?} (this line must be deleted before merge PR)",
+                    f.name, &f.smallest_key, &f.largest_key
+                );
+                // set `include_end` to `true` to ensure that sst with the same start key and end key.
+                self.db
```
Connor1996, line 132: Add some metrics for the check?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,256 @@
...
+                    .delete_files_in_range(&f.smallest_key, &f.largest_key, true)
```
Connor1996, line 133: What if there is a add-peer within the range. You should make sure the meta lock is held when `delete_files_in_range` is executing. You can check the original design for the validity dicussion. And please add a test case for this.

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,256 @@
...
+            .range((Excluded(file.smallest_key.clone()), Unbounded::<Vec<u8>>))
+        {
+            let region = &meta.regions[id];
+            if keys::enc_start_key(region) <= file.largest_key {
+                ids.push(*id);
+            }
+        }
+
+        let overlap = !ids.is_empty();
+        if overlap && !self.exist_scheduling_regions(&file.name) {
```
Connor1996, line 164: if the file is detected before, but one region within the range is split. Then you should check the ids again to push the newly created one to `damage_regions_id`.

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,256 @@
...
+            // This sst file was detected for the first time so store meta and
+            // file context should be updated.
+            for id in ids {
+                meta.damaged_regions_id.insert(id);
+                file.overlap_region_ids.push(id);
```
Connor1996, line 169: seems `overlap_region_ids` is used in nowhere

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: fine,  none overlap regions existed seems that delete sst operations can be executed directly

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: what check?

Connor1996: `check_damaged_files`

etc/config-template.toml
```diff
@@ -244,6 +244,18 @@
...
+## to the data range is damaged, it will be reported to PD, and PD will add `evict-leader-scheduler`
+## in this store to minimize the impact on user workload as much as possible. When hung, the corruption
+## SST file will remain, and the KV storage engine can still write new content normally, but it will
+## still return error when reading corrupt data range. 
+##
+## If after this time, the peer where the corrupted data range is located has not been removed from the
+## current store, TiKV will panic.
+##
+## Set to 0 to disable this feature if you want to panic immediately when encountering such an error.
+# max-background-error-hang-time = "24h"
```
tabokie, line 257: Maybe background-error-recovery-window?

components/engine_rocks/src/event_listener.rs
```diff
@@ -3,23 +3,29 @@
...
 use tikv_util::set_panic_mark;
+use tikv_util::worker::Scheduler;
+use tikv_util::{error, warn};
 
 // Message for RocksDB status subcode kNoSpace.
 const NO_SPACE_ERROR: &str = "IO error: No space left on device";
 
 pub struct RocksEventListener {
     db_name: String,
+    sst_recovery_tx: Option<Scheduler<String>>,
```
tabokie, line 21:
```suggestion
    sst_recovery_scheduler: Option<Scheduler<String>>,
```

tabokie: And why exactly are we doing the recovery in another thread?

hunterlxt: rocksdb event listener thread should not be blocked.

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: only add guage to stat damaged files, what do you mean here? add check ops?

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: implemented this test by failpoint

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,265 @@
...
+            }
+        }
+    }
+
+    // `sst_path` has been processed and is still in a damaged state.
+    fn exist_scheduling_regions(&self, sst_path: &str) -> bool {
+        self.damaged_files.iter().any(|f| f.name == sst_path)
+    }
+
+    // Periodically check for recorded damaged files。
```
tabokie, line 110:
```suggestion
    // Cleans up obsolete damaged files and panics if some files are not handled in time.
```

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,265 @@
...
+    //
+    // Acquire meta lock.
+    fn check_damaged_files(&mut self) {
+        let mut new_damaged_files = self.damaged_files.clone();
```
tabokie, line 114: Why clone?

tests/integrations/config/mod.rs
```diff
@@ -631,6 +631,7 @@ fn test_serde_custom_tikv_config() {
     raft_engine_config.recovery_read_block_size.0 = ReadableSize::kb(1).0;
     raft_engine_config.recovery_threads = 2;
     value.storage = StorageConfig {
+        background_error_recovery_window: ReadableDuration::hours(24),
```
tabokie, line 634: Should be placed correctly, i.e. at the back.

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,265 @@
...
+            if !self.exist_scheduling_regions(&file.name) {
+                self.damaged_files.push(file);
+                TIKV_ROCKSDB_DAMAGED_FILES.inc();
+            }
+        } else {
+            fail_point!("sst_recovery_before_delete_files");
+            // The sst file can be deleted safely.
+            // set `include_end` to `true` otherwise the file with the same largest key will be skipped.
+            self.db
+                .delete_files_in_range(&file.smallest_key, &file.largest_key, true)
```
tabokie, line 172: Not good to do this under lock?

Connor1996: it's necessary to keep correctness considering concurrent add-peer. Should add comment here to explain

components/engine_rocks/src/event_listener.rs
```diff
@@ -3,23 +3,32 @@
...
 use tikv_util::set_panic_mark;
+use tikv_util::worker::Scheduler;
+use tikv_util::{error, warn};
 
 // Message for RocksDB status subcode kNoSpace.
 const NO_SPACE_ERROR: &str = "IO error: No space left on device";
 
 pub struct RocksEventListener {
     db_name: String,
+    sst_recovery_scheduler: Option<Scheduler<String>>,
```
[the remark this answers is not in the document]
hunterlxt: If wrap this a new task, the task definition will be placed in `engine_rocks` to avoid cycle dependency, which make it separate from the sst recovery main logic.

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: Added comment

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,265 @@
...
+            if path == live_files.get_name(i as i32) {
+                let f = FileInfo {
+                    name: live_files.get_name(i as i32),
+                    smallest_key: live_files.get_smallestkey(i as i32).to_owned(),
+                    largest_key: live_files.get_largestkey(i as i32).to_owned(),
+                    start_time: Instant::now(),
+                };
+
+                // only support recovering data key for now.
+                if !keys::validate_data_key(&f.smallest_key)
```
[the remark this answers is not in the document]
hunterlxt: range mismatch indicates that the corruption cannot be recovered. So I think just `set_panic_mark_and_panic` is ok?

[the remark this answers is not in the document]
tabokie: Oh, I misread. The error message "key range mismatch" is a bit confusing.

components/engine_rocks_helper/src/sst_recovery.rs
tabokie: I guess you could make `check_overlap_damaged_regions` immutable, by pushing to `damaged_files` outside of it.

hunterlxt: yes if we make `check_overlap_damaged_regions` immutable, it still can't be called while `self` is mutable borrowed. of course we could pass struct fields instead of self into `check_overlap_damaged_regions`, but damaged_files are referenced internally to log

components/engine_rocks_helper/src/sst_recovery.rs
tabokie: So a peer cannot be added when the data is missing? What is the exact row of data that determines this logic?

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,280 @@
...
+            }
+        } else {
+            fail_point!("sst_recovery_before_delete_files");
+            // The sst file can be deleted safely and set `include_end` to `true` otherwise the
+            // file with the same largest key will be skipped.
+            // Here store meta lock should be held to prevent peers from being added back.
+            self.db
+                .delete_files_in_range(&file.smallest_key, &file.largest_key, true)
+                .unwrap();
+            if self.db.get(&file.smallest_key).unwrap().is_some() {
```
tabokie, line 181: Why? I thought `delete_files_in_range` won't delete data in memtable or L0.

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: a peer with same range can be added back when the data is missing but cannot when cleanup logic is executing.

tabokie: So this lock is to prevent newly added peer mistakenly deleted here?

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: `test_sst_recovery_overlap_range_sst_exist` may exactly tested this case？

components/engine_rocks_helper/src/sst_recovery.rs
Connor1996: Yes, it's about to clean up the range. If there is an add-peer command add peer on this TiKV and apply a snapshot concurrently. This clean-up may delete the data of the newly added peer mistakenly.

components/engine_rocks_helper/src/sst_recovery.rs
tabokie: Could you move the part of code that checks file against region ranges into a static function? (`fn get_overlapped_region(meta: &MutexGuard<..>, f: FileInfo) -> Vec<u64>`)

tabokie: Could you move the part of code that checks file against region ranges into a static method? (`get_overlapped_region(meta: &MutexGuard, f: FileInto) -> Vec<u64>`). You can then clone the meta to get around the immutable reference.

components/engine_rocks_helper/src/sst_recovery.rs
Connor1996: "So this lock is to prevent newly added peer mistakenly deleted here?" 
No, the peer can't be added due to lock is hold.

components/engine_rocks_helper/src/sst_recovery.rs
hunterlxt: At first I implemented exactly the same as you said, but there are still some places that needs to use `damaged_regions`, such as `set_panic_mark_and_panic`, which needs to print out all file information when panicking.
Also there shouldn't be any performance penalty for clone here as it is impossible that thousands of sst files exist at the same time.

tabokie: The code I mentioned (L140-L153) doesn't call `set_panic_mark_and_panic`. And yes, I agree the overhead is not a problem, I'm just not too comfortable with the code structure. Here I have to think hard to understand how some modifications inside `check_overlap_damaged_regions` is not effective (because the file list is cloned).

hunterlxt: > Here I have to think hard to understand how some modifications inside check_overlap_damaged_regions is not effective (because the file list is cloned).
Only modify when first time detection. Other times, you can think of the `damaged_files` as a snapshot?

> The code I mentioned (L140-L153) doesn't call set_panic_mark_and_panic. 
L122~L125 could access the file list early when timeout, if we retain `self.damaged_files` directly, maybe more flag vars are included and we may move L122~L125 to back.

hunterlxt: In [0458b9c](https://github.com/tikv/tikv/pull/10961/commits/0458b9c8397e2d3e9a518e77c49b6963a7323549) move out the modifications to `damged_files` to make the code clear.

tabokie: You are right, it's messier than I think. How about this:
```
let old_damaged_files = self.damaged_files.clone();
self.damaged_files.clear();
old_damaged_files.into_iter().for_each(|f| {
  if f.start_time... { panic }
  check_overlap_damaged_regions(f)
});
```

hunterlxt: If implement like this, files may be pushed again in `check_overlap_damaged_regions`. I think just make `check_overlap_damaged_regions` immutable and this also matches the naming of this function (only do check and don't modify `damaged_files` everywhere)

tabokie: Okay

components/raftstore/src/store/fsm/store.rs
```diff
@@ -129,6 +129,8 @@ pub struct StoreMeta {
     pub destroyed_region_for_snap: HashMap<u64, bool>,
     /// region_id -> `RegionReadProgress`
     pub region_read_progress: RegionReadProgressRegistry,
+    /// record which regions are damaged.
+    pub damaged_regions_id: HashSet<u64>,
```
tabokie, line 133: This field will never be cleaned up?

hunterlxt:
Tested multiple times with https://github.com/tikv/pd/pull/4851 and https://github.com/tikv/tikv/pull/10961, I confirmed that corrupted sst can always be recoverd automatically.

But there are some additional notes:
1. Damaged injection by writing empty sst file and only writing L1~L6 files. This is done by parsing the manifest dump file. Not every time the background merge triggers. (after some attempts, like clear page cache, disable block cache, diable auto compaction, etc)
2. Sometimes some behaviors trigger panic before recovery succeeds. But it's reasonable like,as it used to be 

 @Connor1996  @tabokie @rleungx

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,276 @@
...
+            "Failed to recover sst file: {}, error: {}, damaged_files:{:?}",
+            sst, err, self.damaged_files
+        );
+    }
+
+    fn must_file_not_exist(&self, fname: &str) {
+        let live_files = self.db.get_live_files();
+        for i in 0..live_files.get_files_count() {
+            if live_files.get_name(i as i32) == fname {
+                self.set_panic_mark_and_panic(fname, "file still exists");
```
Connor1996, line 198: add comment that if the file is on the level 0, it can't be deleted.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -129,6 +129,8 @@ pub struct StoreMeta {
     pub destroyed_region_for_snap: HashMap<u64, bool>,
     /// region_id -> `RegionReadProgress`
     pub region_read_progress: RegionReadProgressRegistry,
+    /// record which regions are damaged.
```
Connor1996, line 132: add comment `sst_file_name -> (sst_smallest_key, sst_largest_key)`

components/raftstore/src/store/fsm/store.rs
```diff
@@ -163,6 +166,64 @@ impl StoreMeta {
...
+    /// end_key > file.smallestkey
+    /// start_key <= file.largestkey
+    pub fn update_overlap_damaged_ranges(&mut self, fname: &str, start: &[u8], end: &[u8]) -> bool {
+        for (_, id) in self
+            .region_ranges
+            .range((Excluded(start.to_owned()), Unbounded::<Vec<u8>>))
+        {
+            let region = &self.regions[id];
+            if keys::enc_start_key(region).as_slice() <= end {
+                match self.damaged_ranges.entry(fname.to_owned()) {
```
Connor1996, line 182: use `try_insert`

components/engine_rocks_helper/src/sst_recovery.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    // `sst_path` has been processed and is still in a damaged state.
+    fn exist_scheduling_regions(&self, sst_path: &str) -> bool {
+        self.damaged_files.iter().any(|f| f.name == sst_path)
+    }
+
+    // Cleans up obsolete damaged files and panics if some files are not handled in time.
+    //
+    // Acquire meta lock.
```
Connor1996, line 135: remove it

components/raftstore/src/store/fsm/store.rs
hunterlxt: seem that it's a unstable feature and can you explain why it's better than `.entry()`?

etc/config-template.toml
```diff
@@ -244,6 +244,18 @@
 ## Set it to 0 will cause no space is reserved at all. It's generally used for tests.
 # reserve-space = "5GB"
 
+## The maximum recovery time after rocksdb detects restorable background errors. When the data belonging
+## to the data range is damaged, it will be reported to PD, and PD will add `evict-leader-scheduler`
```
rleungx, line 248: need to update

Connor1996 (approved):
LGTM

components/raftstore/src/store/fsm/store.rs
tabokie: Can use `if let`

tabokie (approved):

PR #10973, opened 2021-09-22
resource_metering: add summary module with a general resource metering framework

mornyx:

Added support for the collection and reporting of summary data recording the number of scanned keys. In addition, a general resource metering framework is abstracted, and cpu/summary related codes are used as sub-modules to work under the new framework.

What's Changed:

- Add a new public abstraction layer, including public models such as `Recorder`, `Reporter` and `Collector`.
- Support the processing of summary statistics under the public abstraction layer.
- Migrate the processing of cpu statistics to the public abstraction layer.
- The only background thread supports collecting multiple resources at the same time.
- Insert the code to collect scanned keys in the main logic.
- Add a lot of unit tests.

### Related changes

- PR to update `pingcap/kvproto`: [pingcap/kvproto#814](https://github.com/pingcap/kvproto/pull/814)

### Check List

Tests

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note

```release-note
Support the statistics of scanned keys in resource metering.
```


Cargo.toml
```diff
@@ -162,7 +162,9 @@ hyper-openssl = "0.9"
 http = "0"
 into_other = { path = "components/into_other", default-features = false }
 keys = { path = "components/keys", default-features = false }
-kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
+# TODO(mornyx): Migrate when kvproto merges.
+# kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = false }
+kvproto = { git = "https://github.com/mornyx/kvproto.git", branch = "resource-metering", default-features = false }
```
zhongzc, line 167: Ref to L261-L265, using patch then you don't need to replace kvproto dependencies everywhere. You can consider patch as `go mod replace` for rust.

components/resource_metering/Cargo.toml
```diff
@@ -35,6 +29,7 @@ slog-global = { version = "0.1", git = "https://github.com/breeswish/slog-global
 futures = "0.3"
 pdqselect = "0.1"
 fail = "0.4"
+thread-id = "4.0.0"
```
zhongzc, line 32: dependency leak?

components/resource_metering/src/collector.rs
```diff
@@ -0,0 +1,23 @@
...
+/// to the `Scheduler` for processing.
+///
+/// `Reporter` implements [Runnable] and [RunnableWithTimer], aggregates the
+/// data sent by the `Collector` internally, and reports it regularly through RPC.
+///
+/// [Recorder]: crate::recorder::Recorder
+/// [Reporter]: crate::reporter::Reporter
+/// [Scheduler]: tikv_util::worker::Scheduler
+/// [Runnable]: tikv_util::worker::Runnable
+/// [RunnableWithTimer]: tikv_util::worker::RunnableWithTimer
```
zhongzc, line 20: 👍

Cargo.toml
mornyx: done.

components/resource_metering/Cargo.toml
mornyx: I use crate `thread-id` in summary module (non-test) to get the unique id of the thread, which is cross-platform by default. In this case we only need a unique id, without any semantics.

mornyx:
This PR was split into two: [10998](https://github.com/tikv/tikv/pull/10998), [11085](https://github.com/tikv/tikv/pull/11085)

PR #10976, opened 2021-09-23
resolved_ts: fix coroutine leaking

### LINKED ISSUE #10965: Coroutine leaking after one TiKV got stuck
### Affected versions
v5.1, v5.2, master

### Description
For Advancing the Resolved TS, we need to verify the leadership of Raft Leaders by calling a `check_leader` RPC every __1s__ to all the other TiKVs (Here we did not filter out learners so there will call the RPC to TiFlash).
However, [here](https://github.com/tikv/tikv/blob/master/components/resolved_ts/src/advance.rs#L216) we use `futures::join_all` to wait for all responses to get received. Then once one TiKV got stuck and the timeout is too long, some coroutines would be accumulated and probably lead to OOM.

5kbpers:

Issue Number: close #10965

Problem Summary:
See #10965

Proposal: [xxx](url) 

What's Changed:

Add a timeout for the coroutine executing.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Please add a release note.
```
* resolved_ts: fix coroutine leaking

### DIFF
--- a/components/cdc/Cargo.toml
+++ b/components/cdc/Cargo.toml
@@ -70,7 +70,7 @@ thiserror = "1.0"
 tikv = { path = "../..", default-features = false }
 tikv_util = { path = "../tikv_util", default-features = false }
 collections = { path = "../collections" }
-tokio = { version = "1.5", features = ["rt-multi-thread"]}
+tokio = { version = "1.5", features = ["rt-multi-thread", "time"]}
 txn_types = { path = "../txn_types", default-features = false }
 concurrency_manager = { path = "../concurrency_manager", default-features = false }
 fail = "0.4"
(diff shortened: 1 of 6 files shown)


BusyJay:
Can it be disabled if not used at all?

NingLin-P (approved):
LGTM

NingLin-P:
> Can it be disabled if not used at all?

The whole `resolved-ts` worker can be disabled by `resolved-ts.enable = false`.

overvenus (commented):
Please add a unit test to cover timeout, rest LGTM

overvenus (approved):
LGTM

NingLin-P (approved):

NingLin-P:
Perhaps we should also cherry-pick this PR to branches that CDC have used the `CheckLeader` mechanism.

PR #10977, opened 2021-09-23
txn: Do assertions on data for mutations to reduce potential data inconsistency

MyonKeminta:

Issue Number: ref https://github.com/pingcap/tidb/issues/26833
Problem Summary: This PR does checks data on each mutation during prewrite, according to the assertion set to the mutation. This expected to avoid some kinds of data inconsistencies.

What's Changed:
When handling prewrites  and reports error if checks fail.

Requires https://github.com/pingcap/kvproto/pull/815

### Related changes

### Check List 

Tests 

- Unit test (TODO)
- Integration test (TODO)

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None.
```


src/storage/txn/commands/acquire_pessimistic_lock.rs
```diff
@@ -87,6 +88,10 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for AcquirePessimisticLock
         let rows = keys.len();
         let mut res = if self.return_values {
             Ok(PessimisticLockRes::Values(vec![]))
+        } else if self.check_existence {
+            // If return_value is set, the existence status is implicitly included in the result.
+            // So check_existence only need to be explicitly handled if `return_values` is not set.
+            Ok(PessimisticLockRes::Existence(vec![]))
```
cfzjywxk, line 94: Do we have cases to cover this new path for `AcquirePessimisticLock`?

MyonKeminta: It should be covered by `test_pessimistic_lock` in storage/mod.rs. You can see if it's enough.

MyonKeminta: And I'll split the pessimistic lock part to another PR.

cfzjywxk: Got it, maybe we could add some explicit cases for it in another PR.

src/storage/txn/actions/prewrite.rs
```diff
@@ -412,6 +437,78 @@ impl<'a> PrewriteMutation<'a> {
...
+
+        // Load the most recent version if prev write is not loaded yet, or the prev write is not
+        // a data version (`Put` or `Delete`)
+        let need_reload = !write_loaded
+            || write.as_ref().map_or(false, |(w, _)| {
+                w.write_type != WriteType::Put && w.write_type != WriteType::Delete
+            });
+        if need_reload {
+            let reload_ts = write.as_ref().map_or(TimeStamp::max(), |(_, ts)| *ts);
+            reloaded_write = reader.get_write_with_commit_ts(&self.key, reload_ts)?;
```
cfzjywxk, line 476: How could we measure the reload or check cost quickly? We may need some utilities maybe some counter?

MyonKeminta: Do  you mean add a metric here? I think that's a good idea

src/storage/txn/actions/prewrite.rs
```diff
@@ -412,6 +437,78 @@ impl<'a> PrewriteMutation<'a> {
...
+                |(w, _)| matches!(w.gc_fence, Some(gc_fence_ts) if !gc_fence_ts.is_zero()),
+            )
+        {
+            // The previously-loaded write record has an invalid gc_fence. Regard it as none.
+            write = &None;
+        }
+
+        // Load the most recent version if prev write is not loaded yet, or the prev write is not
+        // a data version (`Put` or `Delete`)
+        let need_reload = !write_loaded
```
sticnarf, line 470: Do we need to reload if it is an invalid write due to existing gc_fence? It seems to me that it should be equivalent to a delete?🤔
Then, it looks in this case, the assertion would be incorrect after reload, because we treat a deleted (but gc-ed) entry as a existing one.

MyonKeminta: If the current write has an invalid gc_fence, I expect `write_loaded` is true and `need_reload` is false. Does this make sense or is anything wrong in the code?

sticnarf: Ah, the problem here is I thought `seek_write` is used later to reload the write. But actually, `get_write_with_commit_ts` is used, so the invalid gc_fence is ignored.

src/storage/txn/actions/prewrite.rs
```diff
@@ -412,6 +438,98 @@ impl<'a> PrewriteMutation<'a> {
...
+            || write.as_ref().map_or(false, |(w, _)| {
```
ekexium, line 480: Is it possible to reload if the level is not `STRICT`?

MyonKeminta: Yes but only when it has already loaded a write but it's a Rollback or a Lock. In this case reload might happen, but it's expected to be relatively cheap since it only need to do `next`s, so I think that should be ok

ekexium (approved):

youjiali1995 (approved):

sticnarf (approved):

PR #10985, opened 2021-09-26
server: remove register PD process

rleungx:

For now, when TiKV starts, it will register its status address to PD which is useless because the component interface is not used.

Remove the register process.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


NingLin-P (approved):
LGTM, thanks!

overvenus (approved):

PR #10996, opened 2021-09-27
coprocessor: add gbk_bin support in coprocessor

xiongjiwei:

Issue Number: close https://github.com/pingcap/tidb/issues/26822

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/tidb_query_datatype/src/codec/collation/collator/gbk_bin.rs
```diff
@@ -0,0 +1,327 @@
...
+            ch.hash(state);
+        }
+        Ok(())
+    }
+}
+
+// GBK_BIN_TABLE are the encoding tables from Unicode to GBK code, it is totally the same with golang's GBK encoding.
+// If there is no mapping code in GBK, use 0x3F(?) instead. It should not happened.
+#[rustfmt::skip]
+static GBK_BIN_TABLE: [u16; 0xffff + 1] = [
```
skyzh, line 68: Consider using `include_bytes!` macro, or using an existing library like https://docs.rs/encoding_rs/0.8.28/encoding_rs/static.GBK.html

I don't quite like embedding a such a large table in Rust code.

xiongjiwei: - `include_bytes!` macro return `[u8; N]`, the table we need is `[u16; N]`, we need a special process to use it.
- we need to make sure the result is exactly the same as tidb which uses the golang GBK library, and I test the rust GBK library there is some difference that makes us cannot use it.

components/tidb_query_datatype/src/codec/collation/collator/gbk_bin.rs
```diff
@@ -0,0 +1,327 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use super::*;
+
+/// Collator for `gbk_bin` collation with padding behavior (trims right spaces).
+#[derive(Debug)]
+pub struct CollatorGBKBin;
```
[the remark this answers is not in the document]
xiongjiwei: I prefer renaming `CollatorUtf8Mb4UnicodeCi` to `CollatorUTF8Mb4UnicodeCI` 😅

components/tidb_query_datatype/src/codec/collation/collator/gbk_bin.rs
skyzh: ... and writing such big array in Rust this will slower compile speed https://github.com/rust-lang/rust/issues/39352

skyzh: When I paste the full table in my local playground, the first thing to happen is that, my VSCode is stuck. No-one else will be willing to modify this part if it slows down the editor. So please at least create a separate file for that.

Regarding the compile speed,

```bash
cargo build
   Compiling rust-playground v0.1.0 (/home/skyzh/Work/rust-playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.84s
```

it seems not to be a big overhead.

components/tidb_query_datatype/src/codec/collation/collator/gbk_bin.rs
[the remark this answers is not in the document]
xiongjiwei: ok, I find there are many places that use `Utf`, I will use `Gbk` to match it.

components/tidb_query_datatype/src/codec/collation/collator/gbk_bin.rs
xiongjiwei: yes, large table makes editor stocks. I decided to use `include_bytes!`, PTAL

Xuanwo (approved):

skyzh (commented):
Rest LGTM

skyzh (approved):

components/tidb_query_datatype/src/codec/collation/collator/mod.rs
```diff
@@ -185,9 +208,10 @@ mod tests {
             (Collation::Utf8Mb4GeneralCi, 2),
             (Collation::Utf8Mb4UnicodeCi, 3),
             (Collation::Latin1Bin, 4),
+            (Collation::GbkBin, 5),
         ];
         let cases = vec![
-            // (str, [Utf8Mb4Bin, Utf8Mb4BinNoPadding, Utf8Mb4GeneralCi, Utf8Mb4UnicodeCi])
+            // (str, [Utf8Mb4Bin, Utf8Mb4BinNoPadding, Utf8Mb4GeneralCi, Utf8Mb4UnicodeCi, GBKbin])
```
andylokandy, line 214:
```suggestion
            // (str, [Utf8Mb4Bin, Utf8Mb4BinNoPadding, Utf8Mb4GeneralCi, Utf8Mb4UnicodeCi, Latin1Bin, GBKbin])
```

PR #11006, opened 2021-09-29
external-storage: add HdfsStorage upload

peng1999:
Signed-off-by: Peng Guanwen <pg999w@outlook.com>

Issue Number: TBD

Problem Summary: add a storage backend to support hdfs upload.

Proposal: TBD

What's Changed: a `HdfsStorage` type in `external_storage` is added.

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
`components/external_storage/export/examples/scli.rs` can be used to test locally
```console
$ cargo run --example scli -- --file Cargo.toml --name test1.txt --storage hdfs --path hdfs://172.16.5.82:49000/ save
error: Custom { kind: Other, error: "Cannot fount hdfs command, please specify HADOOP_HOME or HDFS_CMD" }
$ JAVA_HOME=/usr/lib/jvm/java-1.8.0/ HADOOP_HOME=/root/peng1999/hadoop-3.3.1 cargo run --example scli -- --file test1.txt --name test1.txt --storage hdfs --path hdfs://172.16.5.82:49000/ save
done
```

### Release note 

```release-note
Support RawKV backup to HDFS.
```

### DIFF
--- a/components/external_storage/export/examples/scli.rs
+++ b/components/external_storage/export/examples/scli.rs
@@ -6,10 +6,7 @@ use std::{
     path::Path,
 };
 
-use external_storage_export::{
-    create_storage, make_cloud_backend, make_gcs_backend, make_local_backend, make_noop_backend,
-    make_s3_backend, ExternalStorage,
-};
+use external_storage_export::{ExternalStorage, create_storage, make_cloud_backend, make_gcs_backend, make_hdfs_backend, make_local_backend, make_noop_backend, make_s3_backend};
 use futures_util::io::{copy, AllowStdIo};
 use ini::ini::Ini;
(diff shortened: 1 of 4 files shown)


andylokandy:
Since hadoop is only available on linux, what will happen when if we use it on mac?

peng1999:
> Since hadoop support is only available on linux, what will happen when if we use it on mac?

The hadoop support of tikv here only relies on an external hadoop installation and the `sudo` command, which I think also supports mac. So hadoop support should run on mac as expected. I haven't use mac and don't know much about it though.

andylokandy:
@peng1999 I got it. But can we avoid using static variables but instead pass the config struct to the hdfs backend?

PR #11012, opened 2021-09-30
tidb_query_expr: handle overflow case for degrees()

kennytm:

Issue Number: ref pingcap/tidb#28488

Problem Summary:

The RPN function `degrees()` is not supposed to return ±infinity result. However, if the input is close to `f64::MAX`, the computation `x * 180/π` may exceed the maximum and becomes infinity. 

Since `Real` is just an alias of `NotNan<f64>` it won't prevent using the ±infinity, and thus a non-finite value got returned as Ok.

What's Changed:

We copy the behavior of `exp()` and returns an Err if the result is infinite.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
When degrees() overflows, it now raises "DOUBLE value is out of range" error instead of silently returning infinity.
```


overvenus (approved):

kennytm:
#17853

PR #11039, opened 2021-10-12
load base split: more accurate sample for large key ranges batch

lhy1024:

To load base split, we sample all requests. For some batch requests, we use `key ranges` for batch processing. As we have previously used read qps directly as the current total, longer `key ranges` have a lower probability of being sampled

For `key ranges` within `key ranges`, we calculate separate totals for each of them.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
 more accurate sample for large key ranges batch
```


components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -339,10 +340,10 @@ impl ReadStats {
             .entry(region_id)
             .or_insert_with(|| RegionInfo::new(sample_num));
         region_info.update_peer(peer);
-        region_info.add_query_num(kind, query_num);
         if is_read_query(kind) {
             region_info.add_key_ranges(key_ranges);
         }
+        region_info.add_query_num(kind, query_num);
```
rleungx, line 346: Why reorder it?

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -131,13 +131,14 @@ impl RegionInfo {
     }
 
     fn add_key_ranges(&mut self, key_ranges: Vec<KeyRange>) {
-        for key_range in key_ranges {
-            if self.key_ranges.len() < self.sample_num || self.get_read_qps() == 0 {
+        for (i, key_range) in key_ranges.into_iter().enumerate() {
+            let n = self.get_read_qps() + i;
+            if n == 0 || self.key_ranges.len() < self.sample_num {
```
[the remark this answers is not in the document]
lhy1024: only avoid panic in `gen_range`

rleungx (approved):

HunDunDM (approved):

disksing (approved):
LGTM on behalf of HundumDM.

components/raftstore/src/store/worker/split_controller.rs
lhy1024: to ensure qps is correct when add key range

PR #11048, opened 2021-10-13
server: fix channel full could break the raft connection

### LINKED ISSUE #11047: Channel full could break the raft connection
## Bug Report

Inside the implementation of the `raft` and `batch_raft`:
https://github.com/tikv/tikv/blob/6d0edd7e690d376e651aecf6f0bbefdeec22c390/src/server/service/kv.rs#L697-L709

If L705 returns error `RaftStoreError::Transport(DiscardReason::Full)`, the `res` future will resolve to that error immediately and the `RequestStream` will be dropped hence the RPC stream is broken.

Breaking the raft connection when the channel full occur doesn't help much about the situation and even bring more problems instead:
- Raft messages that have been received but not process yet will be dropped silently
- Breaking the connection could also cause hibernated followers to wake up to start a new election, and potentially could cause region unavailable due to split vote (usually happen when there are more than 3 voters)

NingLin-P:

Issue Number: close #11047 

Problem Summary:

See #11047 

What's Changed:

Fix #11047 by ignoring the `RaftStoreError::Transport(DiscardReason::Full)` error.

### Related changes

### Check List 

Tests 

- Integration test

### Release note 

```release-note
fix channel full could break the raft connection
```


BusyJay:
> I think there is such flow control mechanism in `raft-rs` along with the config `raftstore.raft-max-inflight-msgs` (default 256).

If a message is dropped, leader will keep sending 256MiB data before pause. I suggest to change the raft stream into duplex stream and report error on full.

PR #11059, opened 2021-10-14
[WIP]*: fix the high cost in destroying peers

### LINKED ISSUE #10210: Cleaning raft log costs too much time during destroying a peer

v4.0.6

### What did happened?
```
```

image

w41ter:
Reading the first index from TruncatedState instead of performing
a seeking operations.

Issue Number: close #10210  

Problem Summary: Cleaning raft log costs too much time during destroying a peer and cause much high jitter of latency and leaders to drop.

What's Changed: read the first index from ApplyState::TruncatedState (if it's exists).

### Related changes

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -69,7 +69,7 @@ use crate::store::worker::{
 use crate::store::PdTask;
 use crate::store::{
     util, AbstractPeer, CasualMessage, Config, MergeResultKind, PeerMsg, PeerTicks,
-    RaftCmdExtraOpts, RaftCommand, SignificantMsg, SnapKey, StoreMsg,
+    RaftCmdExtraOpts, RaftCommand, SignificantMsg, SnapKey,
 };
 use crate::{Error, Result};
 
@@ -621,6 +621,7 @@ where
                 PeerMsg::UpdateRegionForUnsafeRecover(region) => {
                     self.on_update_region_for_unsafe_recover(region)
                 }
+                PeerMsg::Cleaned(merge_by_target) => self.on_meta_cleaned(merge_by_target),
             }
         }
         // Propose batch request which may be still waiting for more raft-command
@@ -1827,21 +1828,7 @@ where
                             "peer_id" => self.fsm.peer_id(),
                             "target_peer" => ?target,
                         );
-                        if self.handle_destroy_peer(job) {
-                            // It's not frequent, so use 0 as `heap_size` is ok.
-                            let store_msg = StoreMsg::RaftMessage(InspectedRaftMessage {
-                                heap_size: 0,
-                                msg: msg.clone(),
-                            });
-                            if let Err(e) = self.ctx.router.send_control(store_msg) {
-                                info!(
-                                    "failed to send back store message, are we shutting down?";
-                                    "region_id" => self.fsm.region_id(),
-                                    "peer_id" => self.fsm.peer_id(),
-                                    "err" => %e,
-                                );
-                            }
-                        }
+                        self.handle_destroy_peer(job);
                     }
                     None => self.ctx.raft_metrics.message_dropped.applying_snap += 1,
                 }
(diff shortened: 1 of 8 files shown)


zhangjinpeng1987:
Any quantified comparison result?

components/server/src/raft_engine_switch.rs
```diff
@@ -45,8 +45,11 @@ fn rename_to_tmp_dir<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) {
 fn clear_raft_engine(engine: &RaftLogEngine) -> Result<(), EngineError> {
     let mut batch_to_clean = engine.log_batch(0);
     for id in engine.raft_groups() {
-        let state = engine.get_raft_state(id)?.unwrap();
-        engine.clean(id, &state, &mut batch_to_clean)?;
+        let raft_state = engine.get_raft_state(id)?.unwrap();
+        let first_index = engine.first_index(id).unwrap_or(0);
+        let mut apply_state = RaftApplyState::default();
+        apply_state.mut_truncated_state().set_index(first_index);
```
5kbpers, line 51: `truncated_state.index` was set as `first_index` and added 1 [here](https://github.com/tikv/tikv/pull/11059/files#diff-455d4f08e0767d97cc39ae6cde4ce31e163cde02e52e91a700ae9d01065f9b69R159), is it right?

components/engine_rocks/src/raft_engine.rs
```diff
@@ -137,25 +148,26 @@ impl RaftEngine for RocksEngine {
...
     ) -> Result<()> {
         batch.delete(&keys::raft_state_key(raft_group_id))?;
-        let seek_key = keys::raft_log_key(raft_group_id, 0);
-        let prefix = keys::raft_log_prefix(raft_group_id);
-        if let Some((key, _)) = self.seek(&seek_key)? {
-            if !key.starts_with(&prefix) {
+        let first_index = if apply_state.has_truncated_state()
+            && apply_state.get_truncated_state().get_index() != 0
+        {
+            apply_state.get_truncated_state().get_index() + 1
```
BusyJay, line 159: This is not trustworthy. When compacting logs, the truncated state is updated immediately, and a background job is started to do clean up. If TiKV is shutdown before the job finishes, there will be logs never be deleted until next log compact job.

PatrickNicholas:
@BusyJay @gengliqi 

Hi, I have submitted a draft, which seperating io operations from `destroy_peer`. Please take a look and give some suggestion.

components/raftstore/src/store/worker/region.rs
```diff
@@ -77,6 +86,16 @@ pub enum Task<S> {
...
         end_key: Vec<u8>,
     },
+    ClearMeta {
+        peer_id: u64,
+        tag: String,
+        region: Region,
+        raft_state: RaftLocalState,
+        merge_state: Option<MergeState>,
+        #[derivative(Debug = "ignore")]
+        notifier: Box<dyn CleanedNotifier>,
```
gengliqi, line 96: How about adding a `RaftRouter` to `Runner`? `Clone` seems not necessary.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2341,24 +2332,39 @@ where
...
-            panic!("{} destroy err {:?}", self.fsm.peer.tag, e);
+        };
+
+        if let Err(e) = self.ctx.region_scheduler.must_schedule(task) {
+            panic!(
+                "failed to schedule clear meta task, region_id={}, peer_id={}, err={}",
+                self.fsm.region_id(),
+                self.fsm.peer_id(),
+                e
+            );
```
gengliqi, line 2354: I worry about the change that some metadata are changed before clearing meta and the others are changed after clearing meta.
You may take some time to figure out the correctness dependency of atomicity changing, which is the most difficult one in this PR.
My suggestion is to write kv db/raft db first, then change the memory state(like `StoreMeta`). I think it may introduce fewer bugs than the method used here.
However, it's still needed to check whether some other places read the `kv db` or `raft db` and the correctness isn't broken. (E.g. the create peer process will check the `RegionLocalState` in kv db)

PR #11063, opened 2021-10-15
raftstore: TransferLeader support multiple target peers

### LINKED ISSUE #10602: Speed up evict leader scheduler

When restart/offline a TiKV instance, PD need to evict leader from that instance in case the region is unavailable until the new  leader elected. Evict leaders could take a long time if there are a large number of leaders in the store and hence slow down the restart/offline process.

When transfer leadership, after some checks passed, the follower needs to append all lagging logs before start election by `MsgTimeoutNow`, although the number of lagging logs is limited by `raftstore.leader-transfer-max-log-lag`, it can still cost time. And since PD just randomly pick the transferee for the evict leader scheduler, we can let TiKV decide to transfer leadership to the peer that has the least lagging logs to speed up the process.

So PD sends the `TransferLeader` command with the store/peer list (to filter some store that we can't transfer because violate restraints like placement rules) that TiKV can choose to transfer, and TiKV decide to transfer leadership based on the store/peer list and the lagging logs.

MrCroxx:

Close #10602 .

Problem Summary:

PD evict leader scheduler picks all valid stores, let TiKV decide the best store to transfer leader to. 

Raftstore choose one of the valid stores by comparing their matched index in its local raft ProgressTracker status.

What's Changed:

### Related changes

- [kvproto#822](https://github.com/pingcap/kvproto/pull/822)
- [pd#4214](https://github.com/tikv/pd/pull/4214)

### Check List 

Tests 

- [x] Unit test
- [x] Integration test

Side effects

- Evict leader may lead to leader unbalanced when some store is slow.

### Release note 

```release-note
Evict leader automatically pick the fastest store to transfer leader to.
```


MrCroxx:
/cc @NingLin-P @BusyJay  PTAL

BTW, is there any practice to check matched_idx in integration test?

components/raftstore/src/store/peer.rs
```diff
@@ -3595,9 +3595,32 @@ where
         ctx.raft_metrics.propose.transfer_leader += 1;
 
         let transfer_leader = get_transfer_leader_cmd(&req).unwrap();
-        let peer = transfer_leader.get_peer();
+        let peers = transfer_leader.get_peers();
+
+        let transferred = match peers.is_empty() {
+            true => self.pre_transfer_leader(transfer_leader.get_peer()),
```
BusyJay, line 3601: You should use `if/else` instead of matching a bool value.

components/raftstore/src/store/peer.rs
```diff
@@ -3595,9 +3595,32 @@ where
...
+            false => {
+                let prs = self.raft_group.raft.prs();
+                let mut peer = metapb::Peer::default();
+                let mut max_matched = 0;
+
+                for candidate in peers {
```
BusyJay, line 3607: You mean `peers.iter().max_by_key(|c| prs.get(c.id).map_or(0, |pr| pr.matched))`?

BusyJay:
> BTW, is there any practice to check matched_idx in integration test?

I'm afraid no.

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -108,6 +109,46 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
...
+        );
+
+        for _ in 0..100 {
+            // reset leader and wait transfer successfully.
+            cluster.reset_leader_of_region(1);
+
+            sleep_ms(20);
+
+            if let Some(leader) = cluster.leader_of_region(1) {
+                if leader.get_id() == id || leader.get_id() == (id % 3) + 1 {
```
NingLin-P, line 130: We should also test that tikv will choose the peer with the highest matched index as the next leader, we could do that by adding an `IsolationFilter` to a store then write some data and test that leadership will transfer to another store.

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -108,6 +109,46 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
         assert!(!resp.get_header().has_error(), "{:?}", resp);
         assert_eq!(resp.get_responses()[0].get_get().get_value(), b"v");
     }
+
+    // with multi targets
+    for id in 1..4 {
+        // select multi target leaders to transfer
+        // set `peer` 0 to make sure the new leader must come from `peers`
+        pd_client.transfer_leader(
```
NingLin-P, line 117: Perhaps we should introduce another function like `transfer_leader_multi_target`

components/test_raftstore/src/pd.rs
```diff
@@ -239,8 +244,15 @@ impl Operator {
...
             Operator::TransferLeader {
                 ref peer,
+                ref peers,
                 ref mut policy,
-            } => leader == peer || !policy.schedule(),
+            } => {
+                leader == peer
+                    || peers
+                        .iter()
+                        .fold(false, |finished, peer| finished | (leader == peer))
```
NingLin-P, line 253: How about using `Iterator::any` instead?

tests/integrations/raftstore/test_transfer_leader.rs
MrCroxx: @NingLin-P done. PTAL

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -110,12 +110,73 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
...
+fn test_pd_transfer_leader_multi_target<T: Simulator>(cluster: &mut Cluster<T>) {
+    let pd_client = Arc::clone(&cluster.pd_client);
+    pd_client.disable_default_operator();
+
+    cluster.run();
+
+    cluster.must_put(b"k1", b"v1");
+    cluster.add_send_filter(IsolationFilterFactory::new(3));
+    cluster.must_put(b"k2", b"v2");
+    // append index of the cluster will finally be 2, 2, 1
```
NingLin-P, line 122: How about using `must_get_equal` and `must_get_none` ensure that before transfer leader.

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -110,12 +110,73 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
...
+
+    // call command on this leader directly, must successfully.
+    let mut region = cluster.get_region(b"");
+    let mut req = new_request(
```
NingLin-P, line 126: Better move `req` close to where it actually used.

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -110,12 +110,73 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
...
+        // reset leader and wait transfer successfully.
+        cluster.reset_leader_of_region(1);
+
+        sleep_ms(20);
+
+        if let Some(leader) = cluster.leader_of_region(1) {
+            if leader.get_id() == 2 {
+                // make sure new leader apply an entry on its term
+                // so we can use its local reader safely
+                cluster.must_put(b"k2", b"v2");
```
NingLin-P, line 151: It is unnecessary, the newly elected leader will append a noop entry on its term, and only serve read requests after the noop entry is applied.

NingLin-P:
Plz resolve the conflict

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1274,9 +1274,9 @@ where
                         "try to transfer leader";
                         "region_id" => region_id,
                         "from_peer" => ?peer,
-                        "to_peer" => ?transfer_leader.get_peer()
+                        "to_peers" => ?transfer_leader.get_peers()
```
NingLin-P, line 1277: `transfer_leader.get_peer()` should also be logged.

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -112,12 +112,73 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
...
+    let mut req = new_request(
+        region.get_id(),
+        region.take_region_epoch(),
+        vec![new_get_cmd(b"k1")],
+        false,
+    );
+    assert!(cluster.leader_of_region(1) == Some(new_peer(2, 2)));
+    let leader_id = cluster.leader_of_region(1).unwrap().id;
+    req.mut_header().set_peer(new_peer(leader_id, leader_id));
+    debug!("requesting {:?}", req);
```
NingLin-P, line 162: Should remove this line

tests/integrations/raftstore/test_transfer_leader.rs
```diff
@@ -112,12 +112,73 @@ fn test_pd_transfer_leader<T: Simulator>(cluster: &mut Cluster<T>) {
...
+    let resp = cluster
+        .call_command(req.clone(), Duration::from_secs(5))
```
NingLin-P, line 164:
```suggestion
        .call_command(req, Duration::from_secs(5))
```

components/raftstore/src/store/peer.rs
```diff
@@ -3722,12 +3722,19 @@ where
...
-        } else {
-            self.pre_transfer_leader(peer)
+        let peer = transfer_leader
+            .get_peers()
+            .iter()
+            .chain([transfer_leader.get_peer()])
+            .max_by_key(|c| prs.get(c.id).map_or(0, |pr| pr.matched));
+
+        let transferred = match peer {
+            Some(peer) if peer.id != self.peer.id && *peer != metapb::Peer::default() => {
```
NingLin-P, line 3734: Actually there may exist `peer` equal to `metapb::Peer::default()` especially in test.

MrCroxx: What is the expected behavior for the default peer, 0.0? Still call `pre_transfer_leader` with it?

NingLin-P: What I mean is we should not rely on `peer == Peer::default()` to determine whether the `peer` field is set, because the field may exactly be set to `Peer::default()`. Should use `xxx.has_peer()` and `xxx.has_peers()` instead.

MrCroxx:
@NingLin-P PTAL

components/raftstore/src/store/peer.rs
```diff
@@ -3780,7 +3780,13 @@ where
         ctx.raft_metrics.propose.transfer_leader += 1;
 
         let transfer_leader = get_transfer_leader_cmd(&req).unwrap();
-        let peer = transfer_leader.get_peer();
+        let prs = self.raft_group.raft.prs();
+
+        let peer = transfer_leader
+            .get_peers()
+            .iter()
+            .max_by_key(|c| prs.get(c.id).map_or(0, |pr| pr.matched))
```
BusyJay, line 3788: If the preferred node is the first one and all their matcheds are the same, then the first one should be chose. `max_by_key` seems to use the last one.

NingLin-P (approved):

components/raftstore/src/store/peer.rs
NingLin-P: When using the `peers` field, we preferred a random pick node if all `pr.matched` are the same.

BusyJay: Always using the last one is not random apparently. As I recall, PD will put the preferred one to the first of the list.

NingLin-P: > Always using the last one is not random apparently

The position in the peer list is fixed but the node id may be random

> PD will put the preferred one to the first of the list

Using the first one is okay to me if there are such API promise

MrCroxx: > As I recall, PD will put the preferred one to the first of the list.

I'm not sure where you are referring to. In [pd#4401](https://github.com/tikv/pd/pull/4401/files) I just return and copy the full candidates list, without sorting or shuffling. So maybe a random pick is better here?

BusyJay: Maybe check `transfer_leader` first? If `transfer_leader` doesn't have enough progress, then a random pick.

components/raftstore/src/store/peer.rs
```diff
@@ -3782,11 +3783,28 @@ where
...
 
-        let peer = transfer_leader
-            .get_peers()
-            .iter()
-            .max_by_key(|c| prs.get(c.id).map_or(0, |pr| pr.matched))
-            .unwrap_or_else(|| transfer_leader.get_peer());
+        let (_, peers) = transfer_leader.get_peers().iter().fold(
+            (0, vec![]),
+            |(max_matched, mut chosen), peer| {
+                if let Some(pr) = prs.get(peer.id) {
```
BusyJay, line 3789: Should probably filter `self_id`.

BusyJay (approved):

PR #11080, opened 2021-10-16
github/workflow: adopt new DCO validator

### LINKED ISSUE #11078: Adopt new DCO validator
According to https://github.com/probot/dco/issues/162, DCO App is possibly offline in weeks.

Even if there is new maintainer volunteering, it can be considered to adopt new stable DCO validator.

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

This closes #11078 

```release-note
None
```

A drawback is that we may pick this workflow to almost every active branch so that the check can be apply on them.

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -1,0 +1,9 @@
+name: DCO Check
+
+on: [pull_request]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ziggurat-project/actions-dco@v1.0


tisonkun:
And after we migrate to this checker (if it goes well), we can change the required status from one to the other and remove DCO App integration.

tisonkun:
@hi-rustin I check https://book.prow.tidb.io/ but don't find the dco bot, you can comment in the issue #11078 about other resolution and we can evaluate it.

The original issue doesn't specify a solution and open to any possible one.

PR #11087, opened 2021-10-18
resource_metering: introduce pub/sub mode

zhongzc:

Make the diagnostic node dependent on TiKV instead of TiKV dependent on the diagnostic node to reverse the dependency to a normal form.

Depends on: https://github.com/pingcap/kvproto/pull/827
Related: https://github.com/pingcap/tidb/pull/29020

- Deprecate config `resource-metering.enabled`
   - If exists external subscribers, the recorder will be enabled. If no external subscribers, the recorder will be paused. The config `resource-metering.enabled` takes no effect anymore.
- Provide gRPC service `kvproto::resource_usage_agent_grpc::ResourceMeteringPubSub`
   - Clients can subscribe to resource metering records through this service.
- Modify trait `resource_metering::Client`
   - Add methods `is_pending` and `is_closed`. Reporter depends on them to enable/disable the recorder.
   - Add a new implementation `PubClient` for connecting the reporter and external subscribers.

### Check List 

Tests 

- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Add resource metering pub/sub gRPC service for publishing records
```


breeswish:
Maybe we can hold the change of this PR until the change of [Pub/Sub at TiDB side](https://github.com/pingcap/tidb/pull/29020) comes to an agreement since mostly they are similar designs.

Issue #10905, opened 2021-09-06
Update prost in TiKV projects to 0.8


sticnarf:
## Development Task

Prost 0.7 is vulnerable to [RUSTSEC-2021-0073](https://rustsec.org/advisories/RUSTSEC-2021-0073.html). This security issue does not affect TiKV itself. But many TiKV projects are widely used by the community. The users of these projects may hope to get the issue fixed without having multiple prost versions.

Prost 0.7 does not include [this patch](https://github.com/tokio-rs/prost/pull/499) so it also blocks us from updating to future Rust toolchains.

### Projects that use prost

- [ ] https://github.com/tikv/protobuf-build
- [ ] https://github.com/tikv/grpc-rs
- [ ] https://github.com/tikv/raft-rs
- [x] https://github.com/tikv/pprof-rs
- [ ] https://github.com/tikv/client-rust
- [x] https://github.com/tikv/agatedb
- [ ] https://github.com/pingcap/kvproto
- [ ] https://github.com/pingcap/tipb

Xuanwo:
raft-rs looks like don't have direct depends on prost.

sticnarf:
> 
> 
> raft-rs looks like don't have direct depends on prost.

The proto of raft-rs depends on protobuf-build and prost is one of the options.

Xuanwo:
> The proto of raft-rs depends on protobuf-build and prost is one of the options.
> 
> https://github.com/tikv/raft-rs/blob/master/proto/Cargo.toml

OK，I will start a PR tomorrow.

Xuanwo:
raft-rs fixed in https://github.com/tikv/raft-rs/pull/456

sticnarf:
Prost 0.9 is released, I think the target should be changed to 0.9.
Also cc pprof-rs maintainer @YangKeao and agatedb maintainer @skyzh.

sticnarf:
If we want to totally remove prost from TiKV itself, pprof-rs also needs to provide an option to not use prost.

YangKeao:
> 
> 
> If we want to totally remove prost from TiKV itself, pprof-rs also needs to provide an option to not use prost.

There is a PR https://github.com/tikv/pprof-rs/pull/23 several months ago 😢 . I will take a look to this PR again.

PR #11099, opened 2021-10-19
*: remove prost support

BusyJay:

Problem Summary:
Since it's not used in production, remove it to reduce dependencies and
codes.

To fully remove prost, we still need pprof to support protobuf-codec.

See also #10905.

### Check List 

Tests 

### Release note 

```release-note
None
```


src/storage/txn/scheduler.rs
```diff
@@ -931,7 +931,11 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
                                     SCHED_THROTTLE_TIME.observe(start.saturating_elapsed_secs());
                                     return;
                                 }
-                                async_std::task::sleep(Duration::from_millis(1)).await;
+                                GLOBAL_TIMER_HANDLE
+                                    .delay(std::time::Instant::now() + Duration::from_millis(1))
```
Connor1996, line 935: LGTM.  Better to reuse the `now` but seems not easy

Connor1996 (approved):

PR #11100, opened 2021-10-19
storage: fix unstable flow controller test

Connor1996:

Issue Number: close #xxx 

Problem Summary:
Fix unstable test:
- `test_flow_controller_l0`
- `test_flow_controller_memtable`

Proposal: [xxx](url) 

What's Changed:
Add special zero durtaion check when calculating the write rate.

### Related changes

- None

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/flow_controller.rs
+++ b/src/storage/txn/flow_controller.rs
@@ -595,8 +595,11 @@ impl<E: CFNamesExt + FlowControlFactorsExt + Send + 'static> FlowChecker<E> {
         }
 
         // calculate foreground write flow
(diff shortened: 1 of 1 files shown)


Connor1996:
> I noticed here the deadline should be now() + TICK:
> 
> https://github.com/tikv/tikv/blob/2d29d7c0c1fdb3af7e6ce9e8bd33a2fc54fc5ae6/src/storage/txn/flow_controller.rs#L460

It's on purpose, to init the gauge at first. Though actually doesn't make any difference😂

PR #11107, opened 2021-10-20
metrics: use process_cpu_seconds_total to display process CPU

overvenus:

Use process_cpu_seconds_total display process CPU usage, because it's faster than sum(tikv_thread_cpu).

tikv_thread_cpu is slow because it has a larger number of cardinality, eg., name and tid.

### Related changes

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/metrics/grafana/tikv_details.json
+++ b/metrics/grafana/tikv_details.json
@@ -456,7 +456,7 @@
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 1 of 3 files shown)


jackysp:
> `process_cpu_seconds_total{instance="1.1.1.1:20181", job="tikv"}`.
> 
> No `tidb_cluster=\"$tidb_cluster\"` type

I think its the request from DBaas.

PR #11111, opened 2021-10-20
raftstore: delay some events to reduce store and grpc cpu

gengliqi:

Problem Summary:
After using async write io(https://github.com/tikv/tikv/pull/10289), the store loop will be very fast than before.
Thus some events may lead to high CPU cost of store and grpc thread. For example, the high frequency of calling `Transport::flush` in `PollHandler::end` will generate many send system calls. 

What's Changed:
Add `store_events_flush_interval` and `raft_msg_flush_interval_us` to control the frequency of these events.

### Related changes

- No.

### Check List 

Tests 

Side effects

- No

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/store.rs
```diff
@@ -826,28 +797,69 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
+        let now = TiInstant::now();
+        if self.poll_ctx.trans.need_flush()
+            && now.saturating_duration_since(self.last_flush_msg_time)
+                >= Duration::from_micros(self.poll_ctx.cfg.raft_msg_flush_interval_us)
+        {
+            self.last_flush_msg_time = now;
+            self.poll_ctx.trans.flush();
+        }
+
+        for peer in peers {
```
BusyJay, line 809: Can this be moved to `flush_events`?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -826,28 +797,69 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
     }
 
     fn end(&mut self, peers: &mut [Box<PeerFsm<EK, ER>>]) {
-        self.flush_ticks();
+        let now = TiInstant::now();
+        if self.poll_ctx.trans.need_flush()
+            && now.saturating_duration_since(self.last_flush_msg_time)
```
BusyJay, line 802: What if it fails the check and then the system becomes quiet for a long time? When will the trans be flushed?

components/raftstore/src/store/local_metrics.rs
```diff
@@ -385,7 +385,6 @@ pub struct RaftMetrics {
     pub message_dropped: RaftMessageDropMetrics,
     pub propose: RaftProposeMetrics,
     pub process_ready: LocalHistogram,
-    pub append_log: LocalHistogram,
```
BusyJay, line 388: How about reusing the metrics for writing raft logs in async write?

components/raftstore/src/store/fsm/store.rs
gengliqi: If so, `pause` should be called and the `flush` will be called in here.

components/raftstore/src/store/fsm/store.rs
gengliqi: Seems impossible because `flush_events` is also called in `pause`.

components/raftstore/src/store/local_metrics.rs
gengliqi: Done.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -828,28 +799,69 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
     }
 
     fn end(&mut self, peers: &mut [Box<PeerFsm<EK, ER>>]) {
-        self.flush_ticks();
+        let now = TiInstant::now();
```
BusyJay, line 802: If these is put to the end of the function, the flush of metrics may be reduce by at lease one time. Otherwise L856 will always mutate the metrics.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -828,28 +799,69 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
+        }
+
+        let dur = self.timer.saturating_elapsed();
         if self.poll_ctx.has_ready {
-            self.handle_raft_ready(peers);
+            // Only enable the fail point when the store id is equal to 3, which is
+            // the id of slow store in tests.
+            fail_point!("on_raft_ready", self.poll_ctx.store_id() == 3, |_| {});
+
+            let dur = self.timer.saturating_elapsed();
```
BusyJay, line 828: Can it be only calculated once?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -828,28 +799,69 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
     }
 
     fn end(&mut self, peers: &mut [Box<PeerFsm<EK, ER>>]) {
-        self.flush_ticks();
+        let now = TiInstant::now();
+        if self.poll_ctx.trans.need_flush()
+            && now.saturating_duration_since(self.last_flush_msg_time)
+                >= Duration::from_micros(self.poll_ctx.cfg.raft_msg_flush_interval_us)
+        {
+            self.last_flush_msg_time = now;
```
BusyJay, line 807: In my opinion, timer should count the time between now and the first time a msg is pushed to buffer.

BusyJay (commented):
Rest LGTM.

components/raftstore/src/store/local_metrics.rs
5kbpers: I think it may introduce some confusion.🤔

components/raftstore/src/store/fsm/store.rs
gengliqi: It doesn't matter. I use it just to reduce the frequency of `flush`.

BusyJay (approved):

5kbpers (approved):

components/raftstore/src/store/fsm/store.rs
```diff
@@ -828,28 +800,72 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
             .process_ready
-            .observe(duration_to_sec(self.timer.saturating_elapsed()) as f64);
-        self.poll_ctx.raft_metrics.flush();
-        self.poll_ctx.store_stat.flush();
+            .observe(duration_to_sec(dur));
 
-        for peer in peers {
-            peer.update_memory_trace(&mut self.trace_event);
+        if now.saturating_duration_since(self.last_flush_time)
+            >= self.poll_ctx.cfg.store_events_flush_interval.0
```
Connor1996, line 852: add a config check to make sure `store_events_flush_interval` is smaller than all ticks in `flush_ticks`?

gengliqi: I remove this config and change it to 1ms.

PR #11122, opened 2021-10-22
server: fix IOType configure for recving snapshot

### LINKED ISSUE #10659: IOType is not set correctly for receving snapshot

master

Unrelated 

### Steps to reproduce

Unrelated 

### What did you expect?
`IOType` is set correctly

### What did happened?

the `IOType` thread local is set and unset at the begin and end of lifetime of `RecvSnapContext`, whereas, the context is used in a future. So the process of receiving snapshots across multiple threads but these thread's local `IOType` is not correctly set.

MuZhou233:

Issue Number: close #10659 

Problem Summary: 

`RecvSnapContext` is used in a future. So the process of receiving snapshots across multiple threads but these thread's local IOType is not correctly set.

What's Changed:

Now `IOType` is set and unset at begin and end of each method of `RecvSnapContext`

### Check List 

Tests 

### Release note 

```release-note
None
```


Connor1996 (approved):
LGTM

tabokie (approved):

Issue #11121, opened 2021-10-21
async-io `cmd-batch-concurrent-ready-max-count` influnence  kv_pessimistic_lock


nolouch:
## Question

I test TPCC  1000 warehouse with 1000 clients in a cloud environment. Due to disk performance issues, we want to see the effect of async io. In this process, we found that after opening async io, adjusting `cmd-batch-concurrent-ready-max-count ` can improve some performance, and the main benefit is that the latency of kv_pessimistic_lock 0.9 is reduced:
kv:
![image](https://user-images.githubusercontent.com/6428910/138328580-38c13a15-5b05-4621-85ac-0ef432b160c8.png)
tidb:
![image](https://user-images.githubusercontent.com/6428910/138328759-bd281672-2c5b-4a84-934b-47eff851beea.png)

As the picture shows:
- 0.9 kv_pessimistic_lock is reduced after increasing the cmd-batch-concurrent-ready-max-count from 1 to 10, but 0.9 kv_prewrite is increase.
- The total TPCC QPS in TiDB is increased, `kv_pessimistic_lock` is the main factor.

nolouch:
Under @gengliqi investigation, there may exist an improvement related to pipelined pessimistic lock handling.

PR #8499, opened 2020-08-24
raftstore&txn: add a proposed callback

gengliqi:
Signed-off-by: Liqi Geng <gengliqiii@gmail.com>

### What problem does this PR solve?

Problem Summary:

Add a proposed callback. It's invoked after the write request is proposed to the engine successfully which means the raftstore does all the checking and can guarantee that if there is no abnormal leader transfer, this write operation must succeed in the near future.

It's used for improving the success rate of pipelined pessimistic lock for now. 
This work is part of #7980.

### What is changed and how it works?

What's Changed:

Described above.

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit Test
- Intergration Test
Test by https://github.com/pingcap/tipocket/pull/301. The failure rate drops from 0.05%(master) to 0(PR).
```
2020/08/28 01:48:26 pipeline.go:258: [info] [1h39m0s] tps: 1725.33 err/s: 0.00 rate: 0.00% errors/total: 0/11112480(0.00%)
```

Side effects

No.

### Release note 
* Improve the success rate of pipelined pessimistic locking and enable it by default.

PR #11147, opened 2021-10-26
raftstore: call proposed callback in advance

gengliqi:

Issue Number: close https://github.com/tikv/tikv/issues/11121

Problem Summary:
This PR should be merged after https://github.com/tikv/tikv/pull/11111.

Like https://github.com/tikv/tikv/pull/11111 said, the store loop is much faster than before.
The `cmd-batch-concurrent-ready-max-count` is used to enhance the effect of batch cmd and reduce CPU overhead.
The proposals will be postponed if the number of concurrent ready is greater than `cmd-batch-concurrent-ready-max-count`.
However, this method introduces a problem that the proposed callback is postponed because it is called only when proposing the cmd.
There is an optimization named `pipelined pessimistic lock` that depends on proposed callback. So this PR can reduce the duration of `pipelined pessimistic lock`.

What's Changed:

This PR tries to call proposed callback in advance to solve this problem.
In order to maintain the same success rate as before, the batched proposals will be proposed if the leader is about to transfer or a admin command is about to propose.

### Related changes

- No

### Check List 

Tests 

I run tipocket `pipeline` test like https://github.com/tikv/tikv/pull/8499 did.
```
2021/10/29 14:47:42 pipeline.go:354: [info] [5h59m50s] tps: 49.00 lock_err/s: 0.00 commit_err/s: 0.00 lock_err/commit_err/total: 0/0/1023440(0.00%)
```
The result shows this PR can provide the same guarantee as before, which is if there is no abnormal leader transfer, the write operation must succeed in the near future.

Side effects

- No

### Release note 

```release-note
None.
```


zhangjinpeng1987:
Any test code for this changing?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3653,6 +3721,33 @@ where
     }
 
     fn propose_raft_command(
+        &mut self,
+        msg: RaftCmdRequest,
+        cb: Callback<EK::Snapshot>,
+        diskfullopt: DiskFullOpt,
+    ) {
+        if msg.has_admin_request()
+            && self.fsm.batch_req_builder.request.is_some()
```
BusyJay, line 3730: How can this happen?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -637,11 +650,47 @@ where
...
+            || self.fsm.batch_req_builder.header_checked.is_some()
+            || self.fsm.peer.pending_remove
+        {
+            return;
         }
+        let cmd = self.fsm.batch_req_builder.request.take().unwrap();
+        self.fsm.batch_req_builder.header_checked = Some(false);
+        if let Ok(None) = self.pre_propose_raft_command(&cmd) {
+            let term = self.fsm.peer.term();
+            if self.fsm.peer.pending_merge_state.is_none()
```
BusyJay, line 677: Better make a method in peer to check these conditions.

components/raftstore/src/store/fsm/peer.rs
gengliqi: It's a defense to avoid a case that a admin cmd is proposed and it can change the epoch version but there are some requests in `batch_req_builder ` and their proposed callback are called.
In most of the case, the admin request are proposed through `PeerMsg::RaftCommand` and the `propose_batch_raft_command(true)` will be called so the code here is useless.
The `RollbackMerge` cmd is proposed through calling `propose_raft_command` directly but the `header_checked` in `batch_req_builder` should be false.
In summary, the code here can not be triggered for now.

components/raftstore/src/store/peer.rs
```diff
@@ -3541,24 +3538,24 @@ where
         Ok(Either::Left(propose_index))
     }
 
-    fn execute_transfer_leader<T>(
+    pub fn execute_transfer_leader<T>(
```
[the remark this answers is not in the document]
gengliqi: I change this logic. Leader should call `ready_to_transfer_leader` while follower should call `execute_transfer_leader `.

youjiali1995 (commented):
Changes to guarantee the success rate of proposed callbacks LGTM. But codes become more obscure because the check of ongoing admin request is scattered.

I'm curious about the necessity of blocking proposals due to concurrent readies. Seems if the request batch is big enough or there is an admin request, requests will be proposed regardless of concurrent readies. Only the batch is not big enough, it will be blocked.

youjiali1995 (approved):

components/raftstore/src/store/fsm/peer.rs
BusyJay: I suggest to remove the code and make `RollbackMerge` to use `propose_batch_raft_command` instead. Or just rename `propose_batch_raft_command` to `propose_raft_command`. Useless code is the source of confusion.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -583,7 +595,6 @@ where
                             self.propose_batch_raft_command(true);
                         }
                     } else {
-                        self.propose_batch_raft_command(true);
```
BusyJay, line 586: I don't think it's correct. For example, delete range request may be reordered to front of other puts/deletes as `propose_raft_command` may not propose any commands.

components/raftstore/src/store/peer.rs
```diff
@@ -564,7 +564,7 @@ where
     pub max_ts_sync_status: Arc<AtomicU64>,
 
     /// Check whether this proposal can be proposed based on its epoch.
-    cmd_epoch_checker: CmdEpochChecker<EK::Snapshot>,
+    pub cmd_epoch_checker: CmdEpochChecker<EK::Snapshot>,
```
BusyJay, line 567: Seems can be private now.

components/raftstore/src/store/peer.rs
```diff
@@ -3541,43 +3538,12 @@ where
...
-    fn execute_transfer_leader<T>(
+    pub fn execute_transfer_leader<T>(
         &mut self,
         ctx: &mut PollContext<EK, ER, T>,
         msg: &eraftpb::Message,
         peer_disk_usage: DiskUsage,
     ) {
-        // log_term is set by original leader, represents the term last log is written
-        // in, which should be equal to the original leader's term.
-        if msg.get_log_term() != self.term() {
```
BusyJay, line 3552: This is still necessary.

components/raftstore/src/store/peer.rs
```diff
@@ -4005,6 +3971,19 @@ where
         }
         false
     }
+
+    /// Check if proposed callback can be called immediately.
+    pub fn check_for_proposed_cb(&mut self, cmd: &RaftCmdRequest) -> bool {
```
BusyJay, line 3976: How about `will_likely_propose`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3652,7 +3736,25 @@ where
         }
     }
 
+    /// Propose batched raft commands(if any) first, then propose the given raft command.
     fn propose_raft_command(
```
BusyJay, line 3740: Seems the method is unnecessary at all. The only place it's used is https://github.com/tikv/tikv/pull/11147/files#diff-7a4718b9cddd27a4ef7e563024d5283cf14b4ba3abf26edc33b974b87680a5fbL587, which can be prepend batch propose call.

gengliqi: Another place is `RollbackMerge`. This method is intended to be called when the order between this command and batched normal commands is necessary.

components/raftstore/src/store/peer.rs
gengliqi: This check is moved to `on_transfer_leader_msg`.

BusyJay: I mean it's also necessary for follower.

gengliqi: Yes. Follower should also call `on_transfer_leader_msg` when receiving `TransferLeader` msg.

components/raftstore/src/store/peer.rs
```diff
@@ -267,7 +267,7 @@ impl<S: Snapshot> ProposedAdminCmd<S> {
     }
 }
 
-struct CmdEpochChecker<S: Snapshot> {
+pub struct CmdEpochChecker<S: Snapshot> {
```
BusyJay, line 270: Can be private now.

BusyJay (approved):
Rest LGTM

PR #11164, opened 2021-10-27
security, server: Implement SAN-based client certificate verification

LINKIWI:

Problem Summary:
Currently, TiKV supports `cert-allowed-cn` for application-layer validation of connecting peers for both the gRPC server and HTTP status server by common name. There does not currently exist support for peer validation on any other X509 certificate fields.

Additional context: https://pingcap.slack.com/archives/C01AXTD2EHJ/p1635350511014300

What's Changed:
This change adds a new configuration field `cert-allowed-san` which allows the specification of X509 subject alternative names (both DNS names and URI names, to support standards like [SPIFFE](https://spiffe.io/)) against which to validate client certificates. It is implemented as an additional RPC checker routine for the gRPC code path, and as another logic branch in the `check_cert` routine for the status server.

Note that the change proposal allows CN and SAN checks to operate simultaneously. 

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`: Will create a PR to update TiKV configuration file documentation after getting some feedback on whether we should proceed with this change.

### Check List 

Tests 

Manual test:

Provision two certificates signed by `ca.pem`

- `foo.pem` with URI `spiffe://example.com/workload`
- `bar.pem` with URI `spiffe://example.com/invalid`

Populate configuration as follows:

```
[security]
ca-path = "/path/to/ca.pem"
cert-path = "/path/to/cert.pem"
key-path = "/path/to/key.pem"
cert-allowed-san = ["spiffe://example.com/workload", "spiffe://example.com/unrelated"]
```

Run TiKV.

Attempt connection from a TiDB process with `foo.pem`: TiDB starts successfully and `select tidb_version();` completes successfully.

Attempt connection from a TiDB process with `bar.pem`: observe in TiDB logs:

```
...
{"level":"INFO","time":"2021/10/27 20:34:41.075 +00:00","caller":"client_batch.go:583","message":"batchRecvLoop fails when receiving, needs to reconnect","target":"scrubbed-hostname:20160","forwardedHost":"","error":"rpc error: code = Unauthenticated desc = subject alternative name check fail, reason: no matching x509_subject_alternative_name or doesn't exist, cert_allowed_san: {\"spiffe://example.com/workload\",\"spiffe://example.com/unrelated\"}"}
{"level":"INFO","time":"2021/10/27 20:34:41.075 +00:00","caller":"client_batch.go:547","message":"batchRecvLoop re-create streaming success","target":"scrubbed-hostname:20160","forwardedHost":""}
...
```

There are no side effects and the change is backwards compatible.

### Release note 

```release-note
security, server: Implement SAN-based client certificate verification
```

### DIFF
--- a/components/security/src/lib.rs
+++ b/components/security/src/lib.rs
@@ -289,7 +289,10 @@ fn check_common_name(cert_allowed_cn: &HashSet<String>, ctx: &RpcContext) -> Res
 /// Check peer SAN with cert-allowed-san field.
 /// Similar to check_common_name, but allows for a match on any available certificate SAN (there
 /// may be multiple).
-fn check_subject_alternative_name(cert_allowed_san: &HashSet<String>, ctx: &RpcContext) -> Result<(), String> {
+fn check_subject_alternative_name(
+    cert_allowed_san: &HashSet<String>,
+    ctx: &RpcContext,
+) -> Result<(), String> {
     if let Some(auth_ctx) = ctx.auth_context() {
         let auth_property_sans = auth_ctx
             .into_iter()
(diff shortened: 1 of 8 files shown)


nicolaslopezbravo:
Hello, is there any plans to add SAN to TiKV? if this is still valid can we merge it?

PR #11170, opened 2021-10-28
raftstore: distinguish empty and size uninitialized region

### LINKED ISSUE #11114: PD to perform a lot of scheduling caused by the reported region size error

master 

no matter

### Steps to reproduce

just restart a tikv node.

### What did you expect?

When we try to restart a tikv node, the region size reported by tikv's region heartbeat shouldn't suddenly changed to 0.

### What did happened?

In the first few reports, the region size was 0, which caused the PD to perform a lot of scheduling.

Effect versions:
- >= 4.0.14
- >= 5.0.2

PatrickNicholas:

Issue Number: close #11114  

Problem Summary:

Use zero to identify size uninitialized regions and 1 to identify empty regions.  The PR to adapt to PD in here: https://github.com/tikv/pd/pull/4243.

What's Changed:

### Related changes

### Check List 

Tests 

- No code

Side effects

### Release note 

```release-note
None
```


PatrickNicholas:
/cc @5kbpers @BusyJay 
PTAL

5kbpers (approved):

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1464,6 +1464,16 @@ where
...
+                // HACK! In order to keep the compatible of protos, we use 0 to identify
+                // the size uninitialized regions, and use 1 to identify the empty regions.
+                //
+                // See tikv/tikv#11114 for details.
+                let approximate_size = match hb_task.approximate_size {
+                    Some(0) => 1,
+                    Some(v) => v,
+                    None => 0, // size uninitialized
+                };
+                let approximate_keys = hb_task.approximate_keys.unwrap_or_default();
```
BusyJay, line 1476: Is it necessary to do the similar hack for keys?

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/worker/pd.rs
PatrickNicholas: the approximate keys is used in:

- /regions/check/hist-keys
- merge checker
- region stats

It seems that this sudden change in value will not affect scheduling, so there is no need to hack this value.

BusyJay (approved):

PR #11196, opened 2021-11-01
Fix label leaking of thread metrics

### LINKED ISSUE #11195: Label leaking of thread metrics
## Bug Report

Here we buffered all labels (thread IDs), then labels can not be dropped even if threads have been exited.

https://github.com/tikv/tikv/blob/13db52e63a746f546cdf6fa3c779378f228d907c/components/tikv_util/src/metrics/threads_linux.rs#L25-L40

5kbpers:

Issue Number: close #11195 

Problem Summary:
Here we buffered all labels (thread IDs), then labels can not be dropped even if threads have been exited.

https://github.com/tikv/tikv/blob/13db52e63a746f546cdf6fa3c779378f228d907c/components/tikv_util/src/metrics/threads_linux.rs#L25-L40

What's Changed:
* Replace `Counter` with `Gauge`
* Reset metrics after thread IDs updated

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
Fix label leaking of thread metrics
```


breeswish (approved):

sticnarf:
@5kbpers tells me that about 200k tokio runtime threads are created in about 1 months. This is really often. 

I think we should also make the tokio blocking threads live longer (https://docs.rs/tokio/1.13.0/tokio/runtime/struct.Builder.html#method.thread_keep_alive). Otherwise, we might meet the problem https://github.com/tikv/tikv/pull/7306 again. cc @breeswish

sticnarf (approved):
LGTM as it should relieve the burden on prometheus...

BusyJay:
> about 200k tokio runtime threads are created in about 1 months.

Amazing!

breeswish:
@sticnarf  Indeed. Nice catch! The current thread behavior will break the fix of #7306. I have created a new issue to track it: https://github.com/tikv/tikv/issues/11204

@kennytm What do you think?

PR #11205, opened 2021-11-02
status_server: add glibc, pthread, libgcc to the blocklist

### LINKED ISSUE #11108: Memory leak when continuous profiling
nightly, debug build

Linux 5.4.0-80-generic x86_64

### Steps to reproduce
```
make build

tiup playground nightly --tiflash 0 --kv.binpath target/debug/tikv-server

while true; do curl -sSL http://127.0.0.1:20180/debug/pprof/profile 2>&1 >/dev/null; done
```

### What did happened?
Increasing memory usage. Won't come down.
image

### What did you expect?
Reasonable memory footprint.

YangKeao:

close #11108 

Problem Summary:

If the `pprof-rs` sample arrives and the current program is fetching a backtrace, the program will panic. The reason is that there is a lock inside the `dl_iterate_phdr`, which is called while unwinding the backtrace.

A more detailed description about this problem can be found in https://github.com/tikv/pprof-rs/pull/85#issuecomment-954446008 . In tikv, when an `Error` is generated, it's likely to get a backtrace and cause this problem.

After adding these libraries to the blocklist, it will also be able to turn on the heap profiling and `pprof-rs` cpu profiling at the same time, but I haven't modified (and tested) it yet.

What's Changed:

Skip all samples inside the `glibc`, `pthread` and `libgcc`.

### Related changes

### Check List 

Side effects

The sample in the `libc`, `libgcc` and `pthread` will not be collected in the profiling result.

### Release note 

```release-note
status_server: skip profiling sample in glibc, pthread, libgcc to avoid possible deadlock 
status_server: upgrade pprof-rs to fix memory leak
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,9 +49,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.6"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "58fb5e95d83b38284460a5fda7d6470aa0b8844d283a0b614b8535e880800d2d"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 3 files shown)


BusyJay (commented):
Please fix release note, the PR is clearly fixing a bug.

BusyJay:
Have you confirmed profiling works after blocking these symbols?

YangKeao:
> Have you confirmed profiling works after blocking these symbols?

Yes. It actually works.

BusyJay (approved):

YangKeao:
@sticnarf PTAL

sticnarf (approved):
LGTM

PR #11207, opened 2021-11-02
raftstore: add PoolState to save the info needed to build a poller (#11159)

### LINKED ISSUE #11159: Support adjusting the pool size of raftstore poller threads dynamically
A dynamic config modification is performed from TiKV-Control command line ([documentation](https://docs.pingcap.com/tidb/dev/tikv-control#modify-the-tikv-configuration-dynamically)). So expect to adjust the thread pool size in this way:
```sql
set config tikv `raftstore.apply-pool-size`=3
set config tikv `raftstore.store-pool-size`=3
```

### Technical proposal
- Add a `PoolStateBuilder` to `BatchSystem` to create a `PoolState` object to save what is needed to build a `Poller`；
- Define a `RefreshConfigRunner` to handle dynamic configuration updates;
- Define a `PoolController` to save the `PoolState` object and handle dynamically adjusting the pool size of the `Poller`;

ethercflow:

Issue Number: close #11159

Proposal: #11159 

What's Changed: #11159 

### Related changes

- No

### Check List 

Tests 

- No code

Side effects

- No 

### Release note 

```release-note
None
```


ethercflow:
@NingLin-P @5kbpers PTAL

components/batch-system/src/batch.rs
```diff
@@ -10,19 +10,24 @@ use crate::config::Config;
...
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+use std::thread::{self, current, JoinHandle, ThreadId};
 use std::time::Duration;
 use tikv_util::mpsc;
 use tikv_util::time::Instant;
 use tikv_util::{debug, error, info, safe_panic, thd_name, warn};
 
+lazy_static! {
+    pub static ref SCALE_FINISHED_CHANNEL: (Sender<()>, Receiver<()>) = channel::bounded(1);
```
NingLin-P, line 26: It looks too hack...

components/batch-system/src/batch.rs
```diff
@@ -383,9 +416,13 @@ pub struct BatchSystem<N: Fsm, C: Fsm> {
     low_receiver: channel::Receiver<FsmTypes<N, C>>,
     pool_size: usize,
     max_batch_size: usize,
-    workers: Vec<JoinHandle<()>>,
+    workers: Arc<Mutex<Vec<JoinHandle<()>>>>,
     reschedule_duration: Duration,
     low_priority_pool_size: usize,
+    alive_workers: Arc<AtomicUsize>,
```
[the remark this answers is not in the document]
ethercflow: > What `alive_workers` are used for?

@NingLin-P  It's used to record how many pollers there are currently, mainly used in conjunction with `expect_pool_size`.

eg: When we need to reduce the number of pollers from 8 to 4,  we immediately set expect_pool_size to 4,  after a while, when each `Poller` executes drop, it will reduce `alive_workers` by 1 and compare it with `expect_pool_size`. When it is equal, write `channel` to notify `config manager's dispatch` that our modification has been completed.

components/batch-system/src/batch.rs
ethercflow: > It looks too hack...

Yes, I agree. My requirement is to share a `channel` between the `config manager of raftstore` and `Pollers` to achieve synchronization, avoid that the last `Pollers` adjustment have not been completed when the next configuration update request comes. I thought about `callback` before, but can I get the same `channel` when I add `Pollers` multiple times to ensure that the `config manager` can be notified when I drop `Pollers`?

@NingLin-P   Would you please give me some suggestions?

components/batch-system/src/batch.rs
```diff
@@ -432,6 +487,26 @@ where
             max_batch_size: self.max_batch_size,
             reschedule_duration: self.reschedule_duration,
             before_pause_wait_us: self.before_pause_wait_us,
+            pool_size: if priority == Priority::Normal {
```
zhangjinpeng1987, line 490: any tests?

zhangjinpeng1987:
Tests are required.

components/batch-system/src/batch.rs
```diff
@@ -325,14 +327,32 @@ pub trait PollHandler<N, C> {
     }
 }
 
+type JoinHandleCollection = Arc<Mutex<HashMap<ThreadId, (Option<JoinHandle<()>>, bool)>>>;
```
5kbpers, line 330: How about define a struct for `(Option<JoinHandle<()>>, bool)`>?

components/batch-system/src/batch.rs
```diff
@@ -377,7 +397,8 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
     }
 
     // Poll for readiness and forward to handler. Remove stale peer if necessary.
-    fn poll(&mut self) {
+    pub fn poll(&mut self) {
+        fail_point!("poll");
```
5kbpers, line 401: What is this failpoint used for?

ethercflow: > 

@5kbpers It's used for integration test:  https://github.com/tikv/tikv/pull/11213/files#diff-de27562fa1da856fe25ccd7973ce3ef87ff915dbd879db43deae2cf9ac39ea54R37.

components/batch-system/src/batch.rs
```diff
@@ -463,6 +484,15 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
                 batch.schedule(&self.router, *offset, true);
             }
             to_skip_end.clear();
+            if !run && batch.control.is_some() {
+                let len = self.handler.handle_control(batch.control.as_mut().unwrap());
+                if batch.control.as_ref().unwrap().is_stopped() {
+                    batch.remove_control(&self.router.control_box);
+                } else if let Some(len) = len {
+                    batch.release_control(&self.router.control_box, len);
```
5kbpers, line 492: Why not release control fsm outside the loop?

ethercflow: Fixed, thanks!

components/batch-system/src/batch.rs
ethercflow: struct tuple looks better, but I didn’t come up with a suitable name

5kbpers (approved):

components/batch-system/src/batch.rs
```diff
@@ -471,6 +497,24 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
                 batch.schedule(&self.router, r, false);
             }
         }
+        if let Some(fsm) = batch.control.take() {
+            self.router.control_box.release(fsm);
```
NingLin-P, line 501: Should be `self.router.control_scheduler.schedule(fsm)` as there may still have unhandled messages.  /cc @5kbpers

components/batch-system/src/batch.rs
```diff
@@ -571,9 +635,10 @@ where
         info!("shutdown batch system {}", name_prefix);
         self.router.broadcast_shutdown();
         let mut last_error = None;
-        for h in self.workers.drain(..) {
-            debug!("waiting for {}", h.thread().name().unwrap());
-            if let Err(e) = h.join() {
+        for (_, mut v) in self.workers.lock().unwrap().drain() {
+            let handle = v.0.take().unwrap();
```
NingLin-P, line 639: Is it possible that `v.0` will be `None`?

components/batch-system/src/batch.rs
```diff
@@ -509,6 +554,21 @@ where
         &self.router
     }
 
+    pub fn build_pool_state<H: HandlerBuilder<N, C>>(
+        &mut self,
+        handler_builder: H,
+    ) -> Option<PoolState<N, C, H>> {
```
NingLin-P, line 560: Perhaps `build_pool_state` should not return `None`?

NingLin-P (commented):
Rest LGTM

components/batch-system/src/batch.rs
```diff
@@ -571,9 +635,10 @@ where
         info!("shutdown batch system {}", name_prefix);
         self.router.broadcast_shutdown();
         let mut last_error = None;
-        for h in self.workers.drain(..) {
-            debug!("waiting for {}", h.thread().name().unwrap());
-            if let Err(e) = h.join() {
+        for (_, mut v) in self.workers.lock().unwrap().drain() {
+            let handle = v.0.take().unwrap();
+            debug!("waiting for {}", handle.thread().name().unwrap());
+            if let Err(e) = handle.join() {
```
NingLin-P, line 641: This may introduce deadlock, if `handle.join()` run before `Poller::drop`, `Poller::drop` will wait for the lock while this thread holding the lock and waiting for `Poller::drop`.

NingLin-P (approved):

components/batch-system/src/batch.rs
ethercflow: Fixed, thanks!

components/batch-system/src/batch.rs
ethercflow: Fixed, thanks!

Issue #11148, opened 2021-10-26
Rollback optimistic txn if meeting an unexpected primary lock when resolving locks for write


sticnarf:
## Problem

Although we have pessimistic transactions now, optimistic transactions tend to be more efficient if there are few conflicts. And it's also why TiDB prefers optimistic transactions for auto-commit transactions.

When there are write conflicts among optimistic transactions, one of the conflicted transactions must be rolled back. The rollback usually happens fast, which is expected. But sometimes multiple optimistic transactions will form deadlocks, then rollback will happen only when one lock exceeds its TTL. The TTL is usually 3 seconds, causing all the involving transactions to be stuck for 3 seconds.

Here is a concrete example:

Both txn1 (start_ts = 10) and txn2 (start_ts = 20) writes `k1` and `k2`. And the primary lock of these two transactions is `k1`. 

1. Txn1 locks `k1`.
2. Txn2 locks `k2`.
3. Txn1 wants to lock `k2` but find a lock from txn2. Txn1 queries the primary lock of txn2 but it does not exist (`k1` is locked by txn1 itself). Now it keeps CheckTxnStatus `k1` until found the lock of txn2. https://github.com/tikv/client-go/blob/a5250aa7c446e06f0a05d761c5a2c3923f8b48e2/txnkv/txnlock/lock_resolver.go#L489-L492
4. Txn2 wants to lock `k1` but find a lock from txn1. Txn2 queries the primary lock of txn1 but finds it is not expired. So it keeps retry until it expires.

Note that the transaction who has locked the primary lock of another transaction must have a smaller start_ts. Otherwise, the transaction with smaller start_ts will meet a lock with larger ts when it is writing its primary lock, it will abort itself.

## Enhancement

If the primary lock is not expected, we should rollback the lock of one of the transactions when we are resolving for write.

The simpliest idea is just to rollback the lock we want to resolve. But this will rollback the transaction with larger start ts. Still using the above example, `k2` will be rolled back and there will be a rollback record with commit_ts = 20. Then, both of the transactions will be aborted. This is apparently not optimal.

So, I think a better solution is to rollback self if the primary lock is locked by the transaction itself.

As stated above, such a deadlock happens if the lock ts of the primary lock is smaller, so this rollback will only affect one transaction. It is more likely to be safe because we are rolling back self, so we don't need to worry that we will potentially break a committed async-commit transaction.

[...]

PR #11208, opened 2021-11-02
storage: return start TS of mismatching lock in TxnNotFound 

sticnarf:

Issue Number: ref #11148

To resolve another case of optimistic deadlock. See #11148 for details.

When `CheckTxnStatus` or `TxnHeartbeat` encounters a mismatching lock, return the lock TS in the `TxnNotFound` error.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Avoid deadlock in optimistic transactions.
```

### DIFF
--- a/src/storage/txn/commands/check_txn_status.rs
+++ b/src/storage/txn/commands/check_txn_status.rs
@@ -142,9 +142,11 @@ pub mod tests {
     use crate::storage::kv::Engine;
     use crate::storage::lock_manager::DummyLockManager;
     use crate::storage::mvcc::tests::*;
+    use crate::storage::txn::commands::test_util::assert_txn_not_found;
     use crate::storage::txn::commands::{pessimistic_rollback, WriteCommand, WriteContext};
     use crate::storage::txn::scheduler::DEFAULT_EXECUTION_DURATION_LIMIT;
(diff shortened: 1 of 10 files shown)


src/storage/txn/commands/check_txn_status.rs
```diff
@@ -225,20 +227,17 @@ pub mod tests {
...
+            snapshot,
+            WriteContext {
+                lock_mgr: &DummyLockManager,
+                concurrency_manager: cm,
+                extra_op: Default::default(),
+                statistics: &mut Default::default(),
+                async_apply_prewrite: false,
+            },
         );
+        res.map(|_| ()).unwrap_err()
```
MyonKeminta, line 240: 🤔 What does the `map(|_| ())` part do?

sticnarf: `unwrap_err` will print the `Ok()` varient when it panics. But `WriteContext` is not `Debug`, so it will not compile. So I use `.map(|_| ())` to workaround.

sticnarf:
Close because we change the way of resolving the issue.

Issue #10974, opened 2021-09-23
Add API V2 check for RawKV and TxnKV requests


andylokandy:
If `storage.api_version=2`:

- If the request's context has `api_version=1`:
  - Reject the request unless it's a TxnKV request and the keys starting with `m` or `t`.

- If the request's context has `api_version=2`:
  - Only allow the key that has RawKV prefix for RawKV requests.
  - Only allow the key that has TxnKV prefix for TxnKV requests.

If `storage.api_version=1`:

- Reject all requests with `api_version=2` in the context.

andylokandy:
@pingyu You may use this when you implement the request check: https://github.com/tikv/tikv/pull/11061

PR #11216, opened 2021-11-03
storage: Add API V2 check for RawKV and TxnKV requests (txn part)

pingyu:

Issue Number: #10974 

Problem Summary:

Add API V2 check for RawKV and TxnKV requests.

Proposal: [TiKV API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md)

What's Changed:

Add checker logics for the following API:

- prewrite
- pessimistic lock

### Related changes

No.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

Breaking backward compatibility: See rfc [TiKV API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) for detail.

### Release note 

```release-note
None.
```


pingyu:
@andylokandy 
Merged changes of #11228, and modified api checking for TTL storage.
PTAL~

andylokandy:
Maybe we can check txn keys here: https://github.com/tikv/tikv/blob/8eb9cb81909df74e839504559d3325736cd28aea/src/storage/mod.rs#L1160-L1181

The advantage is that check_api_version does not leak outside of storage.

pingyu:
> Maybe we can check txn keys here:
> 
> https://github.com/tikv/tikv/blob/8eb9cb81909df74e839504559d3325736cd28aea/src/storage/mod.rs#L1160-L1181
> 
> The advantage is that check_api_version does not leak outside of storage.

Done. PTAL, thanks~ @andylokandy

src/storage/mod.rs
```diff
@@ -1171,25 +1156,37 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
 
         match &cmd {
             Command::Prewrite(Prewrite { mutations, .. }) => {
-                check_key_size!(
-                    mutations.iter().map(|m| m.key().as_encoded()),
-                    self.max_key_size,
-                    callback
-                );
+                let keys: Vec<&Vec<u8>> = mutations.iter().map(|m| m.key().as_encoded()).collect();
```
andylokandy, line 1159: Is the collect() avoidable?

andylokandy (approved):
rest LGTM

andylokandy:
@iosmanthus @zhongzc  PTAL

tests/integrations/server/kv_service.rs
```diff
@@ -1744,3 +1746,207 @@ fn test_get_lock_wait_info_api() {
...
+        ),
+        // config api_version = V2.
+        // backward compatible for TiDB request, and TiDB request only.
+        (ApiVersion::V2, ApiVersion::V1, TIDB_KEY_CASE, None),
+        (
+            ApiVersion::V2,
+            ApiVersion::V1,
+            TXN_KEY_CASE,
+            Some("InvalidKeyPrefix"),
+        ),
```
zhongzc, line 1782: How about checking RAW_KEY_CASE also?

tests/integrations/server/kv_service.rs
```diff
@@ -1744,3 +1746,207 @@ fn test_get_lock_wait_info_api() {
...
     handle.join().unwrap();
 }
+
+#[test]
+fn test_api_version() {
+    const TIDB_KEY_CASE: &[u8] = b"t_a";
+    const TXN_KEY_CASE: &[u8] = &[key_prefix::TXN_KEY_PREFIX, 0, b'a'];
+    const RAW_KEY_CASE: &[u8] = &[key_prefix::RAW_KEY_PREFIX, 0, b'a'];
+
+    let test_data = vec![
```
zhongzc, line 1756: The APIv2 spec still confuse me. Can you explain the relation between APIv2 and RAW?

And I think it's better to add some comments about the spec here.

andylokandy: Is this spec sufficient?
https://github.com/pingcap/kvproto/blob/d0692eebe1ed1827811b24a377a45b3bed58f15b/proto/kvrpcpb.proto#L715-L762

tests/integrations/server/kv_service.rs
```diff
@@ -1744,3 +1746,207 @@ fn test_get_lock_wait_info_api() {
...
+        // config api_version = V1|V1ttl, for backward compatible.
+        (ApiVersion::V1, ApiVersion::V1, TIDB_KEY_CASE, None),
+        (ApiVersion::V1, ApiVersion::V1, TXN_KEY_CASE, None),
```
zhongzc, line 1759:
```suggestion
        (ApiVersion::V1, ApiVersion::V1, TIDB_KEY_CASE, None),
        (ApiVersion::V1, ApiVersion::V1, TXN_KEY_CASE, None),
        (ApiVersion::V1, ApiVersion::V1, RAW_KEY_CASE, None),
```

tests/integrations/server/kv_service.rs
```diff
@@ -1744,3 +1746,207 @@ fn test_get_lock_wait_info_api() {
...
+        // storage api_version = V1ttl, allow RawKV request only.
+        (
+            ApiVersion::V1ttl,
```
zhongzc, line 1762: We can set the storage API version as `ApiVersion::V1ttl`? @andylokandy

pingyu: Yes. Set `enable-ttl: true` and `api-version: 1` (or leave `api-version` as default)

*(codes [here](https://github.com/tikv/tikv/blob/1e1f32ab744cafd0145cd93e48249025455f6ae7/src/storage/config.rs#L121-L146))*
```
    pub fn api_version(&self) -> ApiVersion {
        match self.api_version {
            1 if self.enable_ttl => ApiVersion::V1ttl,
            1 => ApiVersion::V1,
            2 => ApiVersion::V2,
            _ => unreachable!(),
        }
    }

    pub fn set_api_version(&mut self, api_version: ApiVersion) {
        match api_version {
            ApiVersion::V1 => {
                self.api_version = 1;
                self.enable_ttl = false;
            }
            ApiVersion::V1ttl => {
                self.api_version = 1;
                self.enable_ttl = true;
            }
            ApiVersion::V2 => {
                self.api_version = 2;
                self.enable_ttl = false;
            }
        }
    }
```

pingyu:
Fixed, and improved other similar test cases.
PTAL, thanks~ @zhongzc

tests/integrations/server/kv_service.rs
andylokandy: Oh, `ApiVersion::V2` should set  `self.enable_ttl = true;`! Would you like to correct it?

zhongzc (approved):

PR #11219, opened 2021-11-03
storage: avoid leaking FlowController channel in tests

### LINKED ISSUE #11218: FlowController channels are leaked in tests
https://github.com/tikv/tikv/blob/42c7aad5fb93d0ee6a69179e314ddc8ffc474364/src/storage/txn/flow_controller.rs#L103

Previously, we leak the channel receiver for convenience. Leaking such small amount of memory is not a real problem but this is not a good code practice.

sticnarf:

Issue Number: close #11218

This PR changes `tx` in `FlowController` to use an `Option`. The related code is not in critical paths, so no worries about performance regressions.

### Check List 

Tests 

- Existing tests.

### Release note 

```release-note
None.
```


BusyJay (approved):

Connor1996 (approved):
LGTM

PR #11139, opened 2021-10-25
storage: support raw value encode for API V2

andylokandy:
ref https://github.com/tikv/tikv/issues/10938, also close https://github.com/tikv/tikv/issues/11041

### Changes

- Removed:
  - `get_expire_ts`
  - `strip_expire_ts`
  - `truncate_expire_ts`
  - `append_expire_ts`
  - all `enable_ttl` at everywhere

- Added:
  -  `enable_ttl=true` becomes `ApiVersion::V1ttl`
  - struct `RawValue` for encoding and decoding rawkv values in different API version ways
  - Use `RawValue` in `Storage`, `SSTImporter`, `TTLCompactionFilter` and `TTLProperties`
 
- Moved `KeyPrefix` from module `storage` to crate `engine_traits`.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
none
```

PR #11228, opened 2021-11-03
storage: Add API V2 check for RawKV and TxnKV requests (part 2)

pingyu:

Issue Number: #10974 

Problem Summary:

Add API V2 check for RawKV and TxnKV requests.

Proposal: [TiKV API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md)

What's Changed:

Add checker logics for the following API:

- raw_batch_get_command
- raw_batch_put
- raw_delete
- raw_delete_range
- raw_batch_delete
- raw_scan
- raw_batch_scan
- raw_get_key_ttl
- raw_compare_and_swap_atomic
- raw_batch_put_atomic
- raw_batch_delete_atomic
- raw_checksum
- scan
- scan_lock
- delete_range

### Related changes

No.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

Breaking backward compatibility: See rfc [TiKV API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) for detail.

### Release note 

```release-note
None.
```


src/storage/key_prefix.rs
```diff
@@ -9,24 +9,65 @@ pub const TIDB_RANGES_COMPLEMENT: &[(&[u8], &[u8])] =
...
+/// KeyPrefixToCheck
 ///
-/// Returning true doesn't mean that the key is certainly written by
-/// TiDB, but instead, it matches the definition of TiDB key in API V2,
-/// therefore, the key is treated as TiDB data in order to fulfill the
-/// compatibility.
-pub fn is_tidb_key(key: &[u8]) -> bool {
-    matches!(KeyPrefix::parse(key).0, KeyPrefix::TiDB)
+/// Accept different kinds of keys, to avoid tedious casting at caller side.
+pub trait KeyPrefixToCheck {
```
iosmanthus, line 15: We may implement this trait for any type can deref as &[u8]

pingyu: Could you give more detail ?

I had tried to using coercion here but failed. `KeyPrefixToCheck` is invoked in [`check_api_version()`](https://github.com/pingyu/tikv/blob/98d7c823b97a83cbc1f823f5b3cef3966b190b27/src/storage/mod.rs#L377) and callers pass keys by trait `IntoIterator<Item = xx>`, where the "deref" is not performed.

From https://doc.rust-lang.org/nomicon/coercions.html:
> Note that we do not perform coercions when matching traits...

src/storage/mod.rs
```diff
@@ -938,15 +937,20 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         &self,
         mut ctx: Context,
         max_ts: TimeStamp,
-        start_key: Option<Key>,
-        end_key: Option<Key>,
+        raw_start_key: Option<Vec<u8>>,
+        raw_end_key: Option<Vec<u8>>,
```
andylokandy, line 941: We can accept `Vec<u8>` and call `Key::from_raw_maybe_unbounded` here.

src/storage/mod.rs
```diff
@@ -792,19 +783,24 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
     pub fn scan(
         &self,
         mut ctx: Context,
-        start_key: Key,
-        end_key: Option<Key>,
+        raw_start_key: Vec<u8>,
+        raw_end_key: Option<Vec<u8>>,
```
andylokandy, line 787: ditto

src/storage/key_prefix.rs
andylokandy: I think this is not idiomatic to rust. The caller is responsible for taking care of the ownership, and this kind of automatics increases the difficulty for the reader about which implementation on earth is used.

src/storage/mod.rs
pingyu: I check unbou and transform to `Option<Vec<u8>>` in [`service/kv.rs`](https://github.com/pingyu/tikv/blob/98d7c823b97a83cbc1f823f5b3cef3966b190b27/src/server/service/kv.rs#L1424-L1433). So just `Key::from_raw` is needed here.

src/storage/key_prefix.rs
pingyu: PTAL, thanks~ @iosmanthus @andylokandy

src/storage/mod.rs
andylokandy: I mean by checking it here, we save repeating the logic of `Key::from_raw_maybe_unbounded` in kv.rs and also no long bother to `raw_start_key.as_ref().map(|x| Key::from_raw(&x[..]))` later on.

src/storage/mod.rs
```diff
@@ -369,29 +369,32 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
     ///   * Request of V1 from TiDB, for compatiblity.
     ///   * Request of V2 with legal prefix.
     /// See rfc https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md for detail.
-    fn check_api_version<'a>(
+    fn check_api_version<T: std::convert::AsRef<[u8]>>(
         storage_api_version: ApiVersion,
         req_api_version: ApiVersion,
         cmd: CommandKind,
-        keys: impl IntoIterator<Item = &'a [u8]>,
+        keys: impl IntoIterator<Item = T>,
```
andylokandy, line 376: Does this work?
```suggestion
        keys: impl IntoIterator<Item = impl AsRef<[u8]>>,
```

src/storage/mod.rs
```diff
@@ -1983,7 +2032,12 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         keys: Vec<Vec<u8>>,
         callback: Callback<()>,
     ) -> Result<()> {
-        // TODO: check_api_version
+        Self::check_api_version(
+            self.api_version,
+            ctx.api_version,
+            CommandKind::raw_atomic_store,
+            keys.iter(),
```
andylokandy, line 2039:
```suggestion
            &keys,
```

src/storage/mod.rs
```diff
@@ -2015,14 +2069,20 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
                 if algorithm != ChecksumAlgorithm::Crc64Xor {
                     return Err(box_err!("unknown checksum algorithm {:?}", algorithm));
                 }
 
+                let mut keys = Vec::with_capacity(ranges.len() << 1);
+                for range in &ranges {
+                    keys.push(range.get_start_key());
+                    keys.push(range.get_end_key());
+                }
+                Self::check_api_version(api_version, ctx.api_version, CMD, keys)?;
```
andylokandy, line 2084: This may work
```suggestion
                Self::check_api_version(api_version, ctx.api_version, CMD, &[range.get_start_key(), range.get_end_key()])?;
```

src/storage/mod.rs
pingyu: But if so, we need to check `is_empty()` again before passing to `check_api_version()`.

Besides, the signature as `Option` telling that the argument is optional for unbounded. I think it's more readable.

src/storage/mod.rs
pingyu: En...how ? `range` is iterated from `ranges`.

pingyu:
@andylokandy 
All comments are addressed or replied. PTAL, thanks~

pingyu:
@andylokandy The following changes was committed. PTAL, thanks~
1. Resolve conflicts with #11139 
2. Refine codes involved `iter::once(...)` and `vec![...]` to `[...]`
3. Add utility function `raw_key_maybe_unbounded_into_option` to convert "maybe unbounded" keys to `Option`s.
4. Check TTL by api_version in `check_api_version()` (logics for `sched_command` is in another PR #11216).

src/storage/mod.rs
```diff
@@ -2043,6 +2094,13 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
```
andylokandy, line 2102: same comment as `raw_batch_scan`

src/storage/mod.rs
```diff
@@ -2233,6 +2291,11 @@ impl<E: Engine, L: LockManager> TestStorageBuilder<E, L> {
         self
     }
 
+    pub fn set_enable_ttl(mut self, enable_ttl: bool) -> Self {
+        self.config.enable_ttl = enable_ttl;
+        self
+    }
```
andylokandy, line 2297: I suggest removing this one and modifying the `set_api_version` above.

src/storage/mod.rs
```diff
@@ -1671,7 +1696,10 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
                     .get(priority_tag)
                     .inc();
 
-                // TODO: check_api_version
+                let keys = Some(&start_key)
+                    .into_iter()
+                    .chain(end_key.as_ref().into_iter());
```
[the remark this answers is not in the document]
pingyu: Why a "None" end_key (means "unbounded" I think) should also be checked ?

Besides, under current `check_api_version()` for Api V2 storage, an empty key will always be illegal, as it is of type "KeyPrefix::Unknown".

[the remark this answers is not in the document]
andylokandy: It's because the unbounded range is prohibited in APIV2. Imaging a raw_delete_range request which ranges from "r\0k1" to "", if this is allowed, all transaction keys with "t{keyspaceid}" prefix will also be deleted.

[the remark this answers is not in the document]
pingyu: Got it.
Then I think it would be better to explicitly reject unbounded range and return corresponding error info.

src/storage/mod.rs
```diff
@@ -4313,6 +4311,18 @@ mod tests {
...
+                if reverse {
+                    Some(b"r\0\0".to_vec())
+                } else {
+                    Some(b"r\0z".to_vec())
+                }
+            } else {
+                None
+            }
+        };
+
```
andylokandy, line 4325:
```suggestion
        let (end_key, end_key_reverse_scan) = if let ApiVersion::V2 = api_version {
            (Some(b"r\0z".to_vec()), Some(b"r\0\0".to_vec()))
        } else {
            (None, None)
        };

```

src/storage/mod.rs
```diff
@@ -4775,14 +4801,15 @@ mod tests {
...
-        let ranges: Vec<KeyRange> = vec![b"r\0a".to_vec(), b"r\0b".to_vec(), b"r\0c".to_vec()]
+        let mut ranges: Vec<KeyRange> = vec![b"r\0a".to_vec(), b"r\0b".to_vec(), b"r\0c".to_vec()]
             .into_iter()
             .map(|k| {
                 let mut range = KeyRange::default();
                 range.set_start_key(k);
                 range
             })
             .collect();
+        make_ranges_for_api_v2(&mut ranges, b"r\0z".to_vec());
```
andylokandy, line 4812:
```suggestion
        let mut ranges: Vec<KeyRange> = vec![b"r\0a".to_vec(), b"r\0b".to_vec(), b"r\0c".to_vec(), b"r\0z".to_vec()]
            .windows(2)
            .map(|keys| {
                let mut range = KeyRange::default();
                range.set_start_key(keys[0]);
                if let ApiVersion::V1 | ApiVersion::V1ttl = api_version {
                    range.set_end_key(keys[1]);
                }
                range
            })
            .collect();
```

pingyu: Done. PTAL~

andylokandy (approved):

zhongzc (approved):

andylokandy (approved):

BusyJay:
This PR makes test test_txn_store_rawkv unstable in recent builds. See https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/12889/pipeline/115.

pingyu:
> This PR makes test test_txn_store_rawkv unstable in recent builds. See https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/12889/pipeline/115.

Let me see see.

PR #11234, opened 2021-11-03
Add data_key to get_regions_in_range

### LINKED ISSUE #11217: TiKV GcKeys task doesn't work when called with multiple keys (at least in 5.1 but I think for everything)
## Bug Report

In gc worker, https://github.com/tikv/tikv/blob/master/src/server/gc_worker/gc_worker.rs#L340 the GcKeys GC task when called with multiple keys tries to get the list of regions overlapping the key range for the task and then do a sorted merge of the keys and the space encompassed by the regions. However, it fails to prepend the keys with the data prefix ('z') when passing them to get_regions_in_range, get_regions_in_range doesn't prepend the keys, and so the keys form an invalid range and no regions are ever returned. This makes garbage collection of delete markers not happen.

To verify the issue I added some logging statements in a branch: https://github.com/frew/tikv/pull/1/files#diff-fad0fef4b49a4159243096a9212032ae40e8b88cd0deb3c60df42a3dbdc639dfR385 - an example line from the first logging statement showing the issue in our cluster:
```
[2021/11/03 06:46:28.397 +00:00] [INFO] [region_info_accessor.rs:385] ["gc get_regions_in_range"] [start_key="[116, 128, 0, 0, 0, 0, 0, 0, 255, 159, 95, 114, 128, 0, 0, 12, 81, 255, 152, 74, 144, 0, 0, 0, 0, 0, 250]"] [first_region="Some(([122, 116, 128, 0, 0, 0, 0, 0, 0, 255, 19, 0, 0, 0, 0, 0, 0, 0, 248], 53))"] [regions_size=126845]
```

(note the prefix 122 on the first_region but not on the start_key)

I suggest adding data_key() to https://github.com/tikv/tikv/blob/master/src/server/gc_worker/gc_worker.rs#L338 and the following line, but defer to your expertise.

5.1.2

Linux on GCP n2d

frew:

Issue Number: close #11217 

Problem Summary: GcKeys never actually garbage collects key ranges due to encoding issue.

Add data key prefix so that GcKeys actually garbage collects keys

What's Changed:

### Related changes

- Need to cherry-pick to the release branch if appropriate

### Check List 

Tests 

- No test included - can somebody on the PingCAP side take adding appropriate testing?

### Release note 

```release-note
Fix issue with garbage collection not properly collecting delete markers.
```

### DIFF
--- a/components/raftstore/src/coprocessor/region_info_accessor.rs
+++ b/components/raftstore/src/coprocessor/region_info_accessor.rs
@@ -382,6 +382,8 @@ impl RegionCollector {
         callback: Callback<Vec<Region>>,
     ) {
         let mut regions = vec![];
+        let start_key = data_key(&start_key);
+        let end_key = data_key(&end_key);
         for (_, region_id) in self
             .region_ranges
             .range((Included(start_key), Included(end_key)))


frew:
Actually this has a problem - `get_regions_in_range` is also called from CompactionGuardGenerator, where we already have a data key. I _think_ that function is also busted because it then calls `data_end_key(&region.end_key)` where I think region end_key is already data encoded? This is starting to seem like it really needs an actual implementation in the type system for data encoded keys vs. non data encoded keys or something and I'm not convinced I have enough context to pursue.

PR #11238, opened 2021-11-04
*: support read-through lock

youjiali1995:

Issue Number: close https://github.com/tikv/tikv/issues/11402 

Problem Summary:

As https://github.com/tikv/client-go/issues/356 says, this PR implements read-through-lock function.

Proposal: [xxx](url) 

What's Changed:

Add a `committed_locks` field to RPC context as opposed to `resolved_locks`. All readers are changed. If a reader is blocked by a lock and the lock is in the `resolved_locks`, reader will read the value from it.

### Check List 

Tests 

- Unit test
- Integration test

I tested it in a workload that all reads meet a lock belonging to a committed transaction, and I injected 50ms write latency to TiKV. With this PR, write latency doesn't affect read operations.
image

### Release note 

```release-note
TiKV supports read-through-lock so that read operations needn't waiting for resolving secondary locks.
```


src/storage/mvcc/reader/point_getter.rs
```diff
@@ -239,9 +257,7 @@ impl<S: Snapshot> PointGetter<S> {
             use_near_seek = true;
 
             let cursor_key = self.write_cursor.key(&mut self.statistics.write);
-            if !Key::is_user_key_eq(cursor_key, user_key.as_encoded().as_slice()) {
-                return Ok(None);
-            }
+            // No need to compare user key because it uses prefix seek.
```
sticnarf, line 260: Seems they are leftovers after `PointGetter` changed to always use prefix seek. Now the "multi" option looks totally useless to me 😂

src/storage/mvcc/reader/point_getter.rs
```diff
@@ -343,6 +352,34 @@ impl<S: Snapshot> PointGetter<S> {
...
+                    }
+                    None => {
+                        let value = self.load_data_from_default_cf(lock.ts, user_key)?;
+                        self.statistics.processed_size += user_key.len() + value.len();
+                        Ok(Some(value))
+                    }
+                }
+            }
+            LockType::Delete => Ok(None),
+            LockType::Lock | LockType::Pessimistic => unreachable!(),
```
sticnarf, line 380: If this happens, it can be the fault from the TiKV client. So, I think returning an error should be more reasonable than panicking.

src/storage/mvcc/reader/scanner/mod.rs
```diff
@@ -471,6 +483,48 @@ where
...
+                            lock.ts,
+                            statistics,
+                        )
+                    }?;
+                    Ok(Some(value))
+                }
+            }
+        }
+        LockType::Delete => Ok(None),
+        LockType::Lock | LockType::Pessimistic => unreachable!(),
```
sticnarf, line 524: ditto

src/storage/mvcc/reader/point_getter.rs
youjiali1995: Yes. I kept the `multi` option to tell the `cursor` use `ScanMode::Mixed` but seems point getter can always use `ScanMode::Mixed`. Let me remove it in another PR.

src/storage/mvcc/reader/point_getter.rs
youjiali1995: It can't happen because only when fails to `Lock::check_ts_conflict()`, this function is called, so the `LockType` can't be lock or  pessimsitic.

sticnarf: I see

src/storage/mod.rs
```diff
@@ -6574,18 +6594,110 @@ mod tests {
...
+        req.set_key(k1.clone());
+        req.set_version(110);
+        let consumer = GetConsumer::new();
+        block_on(storage.batch_get_command(vec![req], vec![1], consumer.clone(), Instant::now()))
+            .unwrap();
+        let res = consumer.take_data();
+        assert_eq!(res.len(), 1);
+        assert_eq!(res[0].as_ref().unwrap(), &Some(v1.clone()));
+        // scan
+        for desc in &[false] {
```
you06, line 6678: May also test reverse scan?

src/storage/mvcc/reader/scanner/mod.rs
```diff
@@ -88,6 +88,16 @@ impl<S: Snapshot> ScannerBuilder<S> {
         self
     }
 
+    /// Set locks that the scanner can read through. Locks with start_ts in the specified set will be
+    /// accessed during scanning.
+    ///
+    /// Default is empty.
+    #[inline]
+    pub fn access_locks(mut self, locks: TsSet) -> Self {
```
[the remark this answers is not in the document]
youjiali1995: `access_locks` is opposed to `bypass_locks` which is a verb. Locks in `committed_locks` should be accessed to get the value, and locks in `resolved_locks` should be bypassed.

src/storage/mod.rs
youjiali1995: Yes. Seems I forgot to add it.

youjiali1995: I just noticed that the `start_key` should be greater than `end_key` in reverse scan, but checking memory locks still uses [`start_key`, `end_key`). /cc @sticnarf

sticnarf (approved):
LGTM

src/storage/mvcc/reader/point_getter.rs
```diff
@@ -207,15 +222,18 @@ impl<S: Snapshot> PointGetter<S> {
                 self.met_newer_ts_data = NewerTsCheckState::Met;
             }
             if let Err(e) =
-                Lock::check_ts_conflict(Cow::Owned(lock), user_key, self.ts, &self.bypass_locks)
+                Lock::check_ts_conflict(Cow::Borrowed(&lock), user_key, self.ts, &self.bypass_locks)
             {
                 self.statistics.lock.processed_keys += 1;
+                if self.access_locks.contains(lock.ts) {
+                    return Ok(Some(lock));
```
cfzjywxk, line 229: Should L227 be moved under this check as in the original logic `processed_keys` is increased in the error branch?

src/storage/mvcc/reader/scanner/backward.rs
```diff
@@ -163,6 +163,21 @@ impl<S: Snapshot> BackwardKvScanner<S> {
                         .map_err(Into::into);
                         if result.is_err() {
                             self.statistics.lock.processed_keys += 1;
+                            if self.cfg.access_locks.contains(lock.ts) {
```
cfzjywxk, line 166: Ditto the L165.

src/storage/mvcc/reader/point_getter.rs
youjiali1995: I intend to do it because this lock is indeed processed.

cfzjywxk: Make sense, the meaning of the `processed_keys`  changed a few times since release-4.0, I'm a bit confused about it.

cfzjywxk (approved):
LGTM

Issue #11226, opened 2021-11-03
TiKV rollback encounters panic "failed to load_latest_options"


tabokie:
### Description

Rollback from version greater than v5.2.0 to version lower than [v5.0.4, v5.1.2], TiKV panics with message:
```
failed to load_latest_options \"Invalid argument: Unable to parse the specified CF option disable_write_stall\"
```

### Root cause

https://github.com/tikv/rocksdb/pull/252 introduced a new RocksDB configuration item, lower versions that are unaware of this item can't parse options file.

### Workaround

First rollback to version v5.0.4 or v5.1.2, then rollback to the desired version.

### Related fix

https://github.com/tikv/tikv/pull/10842 makes v5.0.4 accept unrecognized items
https://github.com/tikv/tikv/pull/10843 makes v5.1.2 accept unrecognized items

tabokie:
Closing this as we have an acceptable workaround.

PR #11246, opened 2021-11-04
engine_rocks: switch rust-rocksdb to branch tikv-5.2

tabokie:

Issue Number: #11226

What's Changed:

Switch rust-rocksdb to branch tikv-5.2.

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_rocks/Cargo.toml
+++ b/components/engine_rocks/Cargo.toml
@@ -74,6 +74,7 @@ fail = "0.4"
 git = "https://github.com/tikv/rust-rocksdb.git"
 package = "rocksdb"
 features = ["encryption", "static_libcpp"]
+branch = "tikv-5.2"
 
 [dev-dependencies]
 rand = "0.8"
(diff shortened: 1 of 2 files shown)


components/engine_rocks/Cargo.toml
```diff
@@ -74,6 +74,7 @@ fail = "0.4"
 git = "https://github.com/tikv/rust-rocksdb.git"
 package = "rocksdb"
 features = ["encryption", "static_libcpp"]
+branch = "tikv-5.2"
```
BusyJay, line 77: Why 5.2 instead of 5.3?

tabokie: Reuse it, since there is no changes in rust-rocksdb during this period. (We also reused the tikv-5.0 branch for v5.0/5.1 before)

BusyJay (approved):

Connor1996 (approved):
LGTM

components/engine_rocks/Cargo.toml
yiwu-arbug: should probably call it for example `tikv-5.2+` for future branches.

Issue #11177, opened 2021-10-28
Remove or disable all unstable tests.


TennyZhuang:
There are too many unstable tests in TiKV and not fixed for too long time. A part of them were listed in #10506 , but not all.

E.g. in #11095 , a simple patch, some unrelated test failed 4 times unexpectedly. The unstable tests waste both CI resources and **the time of developers** who should not be responsible for these bugs. Please note that they can do nothing to handle the failure except retyping /run-tests , which is really meaningless.

Another problem is that, the unstable tests are really confusing for new contributors from community. They may waste much time to debug their codes for non-existent bugs.

Why not just remove them or disable them in master, and add them back with new pull requests? For tracking, we can assign every unstable test to specific developer, only these developers spend time on them. We can also create a new branch instead of master to run these tests periodically. Of course, we can still track them with tracking issue.

@zhouqiang-cl @BusyJay @tisonkun @breeswish

BusyJay:
I'm working on some of the cases and also assigning some of them to owners. Hope to make most of them stable in short time.

Some test cases involve heavy IO/CPU operations, perhaps using memory disk/high end disk can also help improve the success rate.

TennyZhuang:
> I'm working on some of the cases and also assigning some of them to owners. Hope to make most of them stable in short time.
> 
> Some test cases involve heavy IO/CPU operations, perhaps using memory disk/high end disk can also help improve the success rate.

Please note that this issue is not about these unstable tests per se, but about how to deal with them. Any test that might cause the master ci to fail should be rolled back or disabled as soon as possible, rather than hanging around causing problems for others and waiting for someone to fix it.

TennyZhuang:
Boosting the fix speed did not solve the issue

BusyJay:
I agree new test cases should be rollback or fixed as soon as possible, but I don't think it's appropriate to disable unstable test cases. Even test cases are unstable, they are still value at covering some part of the system. Disabling it or removing it will lose the opportunity to find errors early.

If there are too many unstable test cases, it means the code quality becomes low, we need to pause and find out what's going wrong instead of producing more low quality code.

PR #11249, opened 2021-11-04
tests: speed up tests with batch put and mem disk

BusyJay:

There are many heavy write in raftstore tests, I have seen some write IO
task takes several seconds to finished, which will certainly make most
test cases timeout. This PR utilizes memory disk to speed up those
operations.
    
Ref #11177

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


tests/integrations/raftstore/test_conf_change.rs
```diff
@@ -605,26 +604,8 @@ fn test_transfer_leader_safe<T: Simulator>(cluster: &mut Cluster<T>) {
...
-        );
-    }
 
-    // Test transfer leader after a safe duration.
-    thread::sleep(cfg.raft_store.raft_reject_transfer_leader_duration.into());
-    assert_eq!(
-        cluster.leader_of_region(region_id).unwrap().get_id(),
-        leader_id
-    );
+    // Any uptodate follower can become leader.
```
gengliqi, line 608:
```suggestion
    // Any up-to-date follower can become leader.
```
```suggestion
    // Any uptodate follower can become leader.
```

components/test_util/src/lib.rs
```diff
@@ -93,3 +93,50 @@ pub fn alloc_port() -> u16 {
...
+        }
+    }
+}
+
+/// Memory disk is provided for CI. By default, all raftstore test cases
+/// are run on memory disk. But in some cases, testing IO stats from system
+/// for example, you may want to run cases on physical disk instead.
+///
+/// The correctness of this API depends on running test case one by one.
+pub fn guarantee_on_physical_disk() -> GuaranteeOnPhysicalDisk {
```
NingLin-P, line 133: Could we add some config (like `try_men_disk`) to control using what kind of disk instead? And setting the default value based on `#[cfg(any(test))]`, if `try_men_disk` is true then call `temp_dir_in_mem` in `test_raftstore`.

gengliqi (approved):
LGTM

Connor1996 (approved):
LGTM

components/encryption/src/encrypted_file/mod.rs
```diff
@@ -69,8 +69,7 @@ impl<'a> EncryptedFile<'a> {
         let mut tmp_file = OpenOptions::new()
             .create(true)
             .write(true)
-            .open(&tmp_path)
-            .unwrap_or_else(|e| panic!("EncryptedFile::write {:?}: {}", &tmp_path.to_str(), e));
+            .open(&tmp_path)?;
```
gregwebs, line 72: It looks like we have lost a very useful error message.

components/encryption/src/manager/mod.rs
```diff
@@ -691,14 +691,12 @@ impl DataKeyManager {
...
-        self.rotate_terminal
-            .send(())
-            .expect("DataKeyManager drop send");
-        self.background_worker
-            .take()
-            .expect("DataKeyManager worker take")
-            .join()
-            .expect("DataKeyManager worker join");
+        if let Err(e) = self.rotate_terminal.send(()) {
+            info!("failed to terminate background rotation, are we shutting down?"; "err" => %e);
```
gregwebs, line 695: Previously this was a panic? Shouldn't it at least be logged at the error level?

BusyJay: Yes, it's panic in the past. This happens when shutting down, so may not be an error at all.

components/encryption/src/encrypted_file/mod.rs
BusyJay: Good point, let me send a new PR to log the path.

Issue #11251, opened 2021-11-04
remove useless response in apply thread


gengliqi:
## Development Task

```
message Response {
    CmdType cmd_type = 1;
    GetResponse get = 2;
    PutResponse put = 4;
    DeleteResponse delete = 5;
    SnapResponse snap = 6;
    PrewriteResponse prewrite = 7;
    DeleteRangeResponse delte_range = 8;
    IngestSSTResponse ingest_sst = 9;
    ReadIndexResponse read_index = 10;
}
```
The size of this data structure is 152 bytes.
The only useful thing in this struct is `ReadIndexResponse` and it's handled by store thread.
However, in apply thread, every request will generate a `Response` and push it to `RaftCmdResponse::responses`, which is very wasteful.

PR #11252, opened 2021-11-04
raftstore: reduce some cpu waste of apply thread

gengliqi:

Issue Number: close https://github.com/tikv/tikv/issues/11251

Problem Summary:
See https://github.com/tikv/tikv/issues/11251, this PR removes many useless `Response`.
Furthermore, this PR reduces some other waste.
1. the clone of `Region` in `prepare_for` function. This cloned `Region` is used for `post_apply_query` and `post_apply_admin` but for now no function really need this `Region`. I use a empty `Region` instead.
2. the clone of `RaftApplyState` when creating `ExecContext` in `apply_raft_cmd`. The `RaftApplyState` is cloned each time a raft cmd is applied. 
3. allocation for key in `handle_put` and `handle_delete`. Use a buffer instead of allocating every time. (Although the jemalloc is good enough to reduce most of the real allocation in this case, why not totally remove it? )

What's Changed:
As above mentioned.

### Related changes

### Check List 

Tests 

- No code

Side effects

- No

### Release note 

```release-note
None.
```

### DIFF
--- a/components/keys/src/lib.rs
+++ b/components/keys/src/lib.rs
@@ -210,6 +210,12 @@ pub fn data_key(key: &[u8]) -> Vec<u8> {
     v
 }
 
+pub fn data_key_with_buffer(key: &[u8], buffer: &mut Vec<u8>) {
+    buffer.clear();
+    buffer.extend_from_slice(DATA_PREFIX_KEY);
+    buffer.extend_from_slice(key);
+}
+
 pub fn origin_key(key: &[u8]) -> &[u8] {
     assert!(
         validate_data_key(key),
--- a/components/keys/src/lib.rs
+++ b/components/keys/src/lib.rs
@@ -434,6 +434,8 @@ mod tests {
         let mut buffer = vec![];
         data_key_with_buffer(b"abc", &mut buffer);
         assert_eq!(buffer, data_key(b"abc"));
+        data_key_with_buffer(b"cde", &mut buffer);
+        assert_eq!(buffer, data_key(b"cde"));
 
         let mut region = Region::default();
         // uninitialised region should not be passed in `enc_start_key` and `enc_end_key`.
--- a/components/keys/src/lib.rs
+++ b/components/keys/src/lib.rs
@@ -429,8 +429,11 @@ mod tests {
 
     #[test]
     fn test_data_key() {
-        assert!(validate_data_key(&data_key(b"abc")));
         assert!(!validate_data_key(b"abc"));
+        assert!(validate_data_key(&data_key(b"abc")));
+        let mut buffer = vec![];
+        data_key_with_buffer(b"abc", &mut buffer);
+        assert_eq!(buffer, data_key(b"abc"));
 
         let mut region = Region::default();
         // uninitialised region should not be passed in `enc_start_key` and `enc_end_key`.
--- a/components/raftstore/src/coprocessor/dispatcher.rs
+++ b/components/raftstore/src/coprocessor/dispatcher.rs
@@ -9,11 +9,16 @@ use engine_traits::{CfName, KvEngine};
 use kvproto::metapb::Region;
 use kvproto::pdpb::CheckPolicy;
 use kvproto::raft_cmdpb::{ComputeHashRequest, RaftCmdRequest};
+use lazy_static::lazy_static;
 use raft::eraftpb;
 use tikv_util::box_try;
 
 use super::*;
 use crate::store::CasualRouter;
+
+lazy_static! {
+    static ref EMPTY_REGION: Region = Region::default();
+}
 
 struct Entry<T> {
     priority: u32,
@@ -506,7 +511,7 @@ impl<E: KvEngine> CoprocessorHost<E> {
         }
         for batch in &cmd_batches {
             for cmd in &batch.cmds {
-                self.post_apply(&batch.region, cmd);
+                self.post_apply(&EMPTY_REGION, cmd);
             }
         }
         for observer in &self.registry.cmd_observers {
@@ -732,7 +737,7 @@ mod tests {
         assert_all!(&[&ob.called], &[55]);
 
         let observe_info = CmdObserveInfo::from_handle(ObserveHandle::new(), ObserveHandle::new());
-        let mut cb = CmdBatch::new(&observe_info, Region::default());
+        let mut cb = CmdBatch::new(&observe_info, &Region::default());
         cb.push(&observe_info, 0, Cmd::default());
         host.on_flush_applied_cmd_batch(cb.level, vec![cb], &PanicEngine);
         // `post_apply` + `on_flush_applied_cmd_batch` => 13 + 6 = 19
--- a/components/raftstore/src/coprocessor/dispatcher.rs
+++ b/components/raftstore/src/coprocessor/dispatcher.rs
@@ -507,7 +507,7 @@ impl<E: KvEngine> CoprocessorHost<E> {
         }
         for batch in &cmd_batches {
             for cmd in &batch.cmds {
-                self.post_apply(&Region::default_instance(), cmd);
+                self.post_apply(Region::default_instance(), cmd);
             }
         }
         for observer in &self.registry.cmd_observers {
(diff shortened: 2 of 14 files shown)


components/keys/src/lib.rs
```diff
@@ -210,6 +210,12 @@ pub fn data_key(key: &[u8]) -> Vec<u8> {
     v
 }
 
+pub fn data_key_with_buffer(key: &[u8], buffer: &mut Vec<u8>) {
```
BusyJay, line 213: Any tests?

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -9,12 +9,17 @@ use engine_traits::{CfName, KvEngine};
...
 use kvproto::raft_cmdpb::{ComputeHashRequest, RaftCmdRequest};
+use lazy_static::lazy_static;
 use raft::eraftpb;
 use tikv_util::box_try;
 
 use super::*;
 use crate::store::CasualRouter;
 
+lazy_static! {
+    static ref EMPTY_REGION: Region = Region::default();
```
BusyJay, line 20: Why not use `Region::default_instance()`?

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -506,7 +511,7 @@ impl<E: KvEngine> CoprocessorHost<E> {
         }
         for batch in &cmd_batches {
             for cmd in &batch.cmds {
-                self.post_apply(&batch.region, cmd);
+                self.post_apply(&EMPTY_REGION, cmd);
```
BusyJay, line 514: /cc @NingLin-P PTAL

components/raftstore/src/coprocessor/dispatcher.rs
gengliqi: Done.

components/raftstore/src/coprocessor/dispatcher.rs
NingLin-P: The region here is actually used by the `loop_ob` macro to construct `ObserverContext`, and currently the only observer of `post_apply` is `LockObserver`, which simply ignored `ObserverContext`. But using an empty region here may be unexpected by incoming observers.

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -506,7 +507,7 @@ impl<E: KvEngine> CoprocessorHost<E> {
         }
         for batch in &cmd_batches {
             for cmd in &batch.cmds {
-                self.post_apply(&batch.region, cmd);
+                self.post_apply(Region::default_instance(), cmd);
```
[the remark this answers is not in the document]
gengliqi: A default immutable `Region` with static lifetime.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3538,38 +3565,67 @@ where
         msgs: &mut Vec<Msg<EK>>,
     ) {
         let mut drainer = msgs.drain(..);
+        let mut batch_apply = None;
```
NingLin-P, line 3568: What is the benefit to batch apply messages of the same message batch? I think the writes are already batched.

components/raftstore/src/store/peer.rs
```diff
@@ -2197,6 +2190,11 @@ where
                     .proposals
                     .find_propose_time(entry.get_term(), entry.get_index());
                 if let Some(propose_time) = propose_time {
+                    // We must renew current_time because this value may be created a long time ago.
+                    // If we do not renew it, this time may be smaller than propose_time of a command,
+                    // which was proposed in another thread while this thread receives its AppendEntriesResponse
+                    // and is ready to calculate its commit-log-duration.
+                    ctx.current_time.replace(monotonic_raw_now());
```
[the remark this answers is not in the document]
gengliqi: To reduce unnecessary overhead of getting time.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3538,38 +3565,67 @@ where
...
-                    // low-priority.
-                    self.delegate.priority = Priority::Normal;
-                    self.handle_apply(apply_ctx, apply);
-                    if let Some(ref mut state) = self.delegate.yield_state {
-                        state.pending_msgs = drainer.collect();
-                        break;
+
+                    if let Some(batch) = batch_apply.as_mut() {
+                        if batch.try_batch(&mut apply) {
+                            return;
```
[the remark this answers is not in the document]
gengliqi: Change to `continue`.

components/raftstore/src/store/fsm/apply.rs
NingLin-P: Found that `WriteBatch` is append-only, and each `finish_for` call will append a `ApplyState`. To reduce write of `ApplyState` of the same region we can also cache the `ApplyState` with `HashMap` and merge them to the `WriteBatch` before it is committed to rocksdb.

gengliqi: I change this logic, it doesn't need to use a `HashMap`. The `ApplyState` will be written only when
1. no next msg or next msg is not apply msg
2. yield

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
BusyJay: So it will be calculated in every loop?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1508,18 +1489,41 @@ where
     }
 }
 
+#[cfg(any(test, feature = "testexport"))]
```
BusyJay, line 1492: I don't think it's a good idea. It means the test will not cover the production.

components/raftstore/src/coprocessor/dispatcher.rs
BusyJay: If it's not used at all, can it be removed completely?

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
gengliqi: It will only be calculated once because `lease_to_be_updated` will be set to `false`.

components/raftstore/src/store/fsm/apply.rs
gengliqi: I remove this optimization to simplify this PR.

components/raftstore/src/store/fsm/apply.rs
gengliqi: Removed

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -286,22 +286,6 @@ pub enum ApplyResult<S> {
     WaitMergeSource(Arc<AtomicU64>),
 }
 
-struct ExecContext {
-    apply_state: RaftApplyState,
```
BusyJay, line 290: This was introduced to solve failures that leaves dirty changes. Though there is no such failures anymore, but it's pitfall for future refactors. Better leave some comments.

components/keys/src/lib.rs
```diff
@@ -210,6 +210,12 @@ pub fn data_key(key: &[u8]) -> Vec<u8> {
...
+    buffer.clear();
```
BusyJay, line 214: This is not covered in test.

BusyJay (approved):
Rest LGTM.

BusyJay (approved):

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -77,6 +77,7 @@ pub trait AdminObserver: Coprocessor {
     fn pre_apply_admin(&self, _: &mut ObserverContext<'_>, _: &AdminRequest) {}
 
     /// Hook to call after applying admin request.
+    /// For now, the `region` in `ObserverContext` is a empty region.
```
NingLin-P, line 80:
```suggestion
    /// For now, the `region` in `ObserverContext` is an empty region.
```

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -92,6 +93,7 @@ pub trait QueryObserver: Coprocessor {
     fn pre_apply_query(&self, _: &mut ObserverContext<'_>, _: &[Request]) {}
 
     /// Hook to call after applying write request.
+    /// For now, the `region` in `ObserverContext` is a empty region.
```
NingLin-P, line 96: ditto

NingLin-P (commented):
Rest LGTM

youjiali1995 (approved):

PR #11261, opened 2021-11-05
raftstore: Fix a flaky test

v01dstar:

Issue Number: #11262 

Problem Summary:
We found this test some times panics, saying "region update for unsafe recover should only occur in leaderless region", this happens when it tries to update the region's meta which is only allowed when the region does not have a leader (during unsafe recover). This test, in the past, waited for 2 election timeout cycles before doing the update, which I believe was unstable for a leader to lose leadership.
 
Instead of blindly wait for 2 election cycles, check leadership periodically in a loop with a timeout to ensure the region has lost its leader.

### Related changes

N/A

### Check List 

N/A

### Release note

```release-note
NONE
```


Connor1996 (approved):
LGTM

BusyJay (approved):

Issue #11187, opened 2021-10-29
Stale non-pessimistic lock may be incompatible with CDC and possibly affect data correctness


sticnarf:
### What version of TiKV are you using?

v4.0.11
### Steps to reproduce

Pessimistic transactions may contain keys that don't need pessimistic locks. These locks are prewritten without any contraint checks.

So, there is a possible case. After a pessimistic transaction is committed, a non-pessimistic lock of the same transaction can  be also prewritten successfully. The lock TS of such a lock is smaller than the commit TS of the latest write record.

This breaks the assumption of the `DeltaScanner`. When a lock is encountered, it needs to find latest write record as the old value. It asserts that the commit TS of the record must be smaller than the lock TS. https://github.com/tikv/tikv/blob/v4.0.11/src/storage/mvcc/reader/scanner/mod.rs#L389-L392

This is probably the cause of the panic below.

### What did you expect?

CDC can still work in this case.

### What did happened?

TiKV panics.

```
stack backtrace:
   0: tikv_util::set_panic_hook::{{closure}}
             at components/tikv_util/src/lib.rs:481
   1: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:475
   2: rust_begin_unwind
             at src/libstd/panicking.rs:375
   3: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:326
   4: tikv::storage::mvcc::reader::scanner::seek_for_valid_write
             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/<::std::macros::panic macros>:9
      tikv::storage::mvcc::reader::scanner::seek_for_valid_value
             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/src/storage/mvcc/reader/scanner/mod.rs:418
   5: <tikv::storage::mvcc::reader::scanner::forward::DeltaEntryPolicy as tikv::storage::mvcc::reader::scanner::forward::ScanPolicy<S>>::handle_lock
             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/src/storage/mvcc/reader/scanner/forward.rs:609
      tikv::storage::mvcc::reader::scanner::forward::ForwardScanner<S,P>::read_next
             at /home/jenkins/agent/workspace/build_tikv_multi_branch_v4.0.11/tikv/src/storage/mvcc/reader/scanner/forward.rs:258
[...]

PR #11264, opened 2021-11-05
txn: Check constraint when retrying prewrite of non-pessimistic-locked keys in pessimistic transactions

MyonKeminta:

ref  #11187
(Imperfect fix, so do not close that issue).

Problem Summary: Prewrites of pessimistic transactions skips checking constraints on data. For a pessimistic transaction, if a prewrite request on a non-pessimistic-lock key is repeated due to something like network issues, it's possible that it successfully acquires the lock even the transaction has already been successfully committed (and maybe even some more versions were written to the key).

This affects CDC and stale read (luckily CDC has other approach to avoid this problem), and might affect data consistency in some very rare cases.

In this PR, prewrite requests of non-pessimistic-lock keys in pessimistic transactions with is_retry_request flags (......) won't skip constraint check, which mostly solves the problem but still imperfect (considering that it's  theoretically possible that the first request arrives later than a retrying request).

What's Changed: This PR changes the logic of `skip_constraint_check` function so that pessimistic prewrite requests of those kind stated above will not skip constraint check.
This change also covers the case in https://github.com/tikv/tikv/pull/10600 , but should have no affect to it.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression

### Release note 

```release-note
Fix the bug that prewrite request retrying in pessimistic transactions have risk to affect data consistency in some rare cases.
```

I actually don't know how to write release note for this thing...


cfzjywxk:
` prewrite requests of non-pessimistic-lock keys in pessimistic transactions with is_retry_request flags (......)` 
Quite a long phrase

cfzjywxk (approved):
LGTM

sticnarf (approved):
LGTM

Issue #11290, opened 2021-11-08
Panic in rocksdb compaction job


gengliqi:
### What version of TiKV are you using?
master

### What operating system and CPU are you using?
doesn't matter

### Steps to reproduce
run TPCC

### What did you expect?
no panic

### What did happened?
```
[...]

PR #11294, opened 2021-11-08
raftstore: Fix compaction guard panic on non-data key

Connor1996:

Issue Number: close https://github.com/tikv/tikv/issues/11290 

Problem Summary:

Proposal: [xxx](url) 

What's Changed: Check if it is a non-data key when calling `origin_key` in compaction guard.

### Related changes

- None

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


yiwu-arbug (approved):
Thanks for the fix!

MyonKeminta (approved):

PR #11300, opened 2021-11-09
copr: add paging API support for streaming-like processling

### LINKED ISSUE #11448: TiKV supports paging copr request
## Development Task

Now TiKV focuses on optimizing unary requests, and there will be a great performance impact in the streaming API of copr. While some use cases can benefit from the streaming-like process, this issue refers to another way towards the optimization, paging API. With paging API, TiKV only scans enough data for the current page instead of the whole range and the query can be streaming-like.

you06:

Issue Number: close #11448 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

This PR makes TiKV accepts a paging parameter from a unary request, once there is enough data for this page, TiKV stop scanning the rest and returned it to the client.

### Related changes

### Check List 

Tests 

Side effects

- Performance regression
    - Consumes more CPU

In some use cases, a unary request can be separated into several paging requests, this may consume more CPU resources. 

### Release note 

```release-note
Support paging-API for streaming-like processling.
```


youjiali1995 (commented):
Please add some tests.

you06:
> Please add some tests.

I added some tests for it, PTAL again.

components/tidb_query_executors/src/runner.rs
```diff
@@ -424,10 +429,20 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
+                    .map_or(false, |p| chunks.len() >= p as usize)
+            {
                 self.out_most_executor
                     .collect_exec_stats(&mut self.exec_stats);
 
+                let range = if self.paging_size.is_some() {
+                    Some(self.out_most_executor.take_scanned_range())
+                } else {
+                    None
+                };
```
youjiali1995, line 444:
```suggestion
                let range = self
                    .paging_size
                    .map(|_| self.out_most_executor.take_scanned_range());
```

tests/failpoints/cases/test_coprocessor.rs
```diff
@@ -172,3 +174,39 @@ fn test_region_error_in_scan() {
...
+    let product = ProductTable::new();
+    let (_, endpoint) = init_with_data(&product, &data);
+    // set batch size and grow size to 1, so that only 1 row will be scanned in each batch.
+    fail::cfg("copr_batch_initial_size", "return(1)").unwrap();
+    fail::cfg("copr_batch_grow_size", "return(1)").unwrap();
+    for paging_size in 1..=4 {
+        let req = DAGSelect::from(&product)
+            .paging_size(paging_size as u64)
+            .build();
+        let mut resp = handle_select(&endpoint, req);
```
youjiali1995, line 196: Should also check range in the response.

youjiali1995 (commented):
rest LGTM

tests/failpoints/cases/test_coprocessor.rs
you06: It seems not easy to check it accurately, I checked its boundaries.

youjiali1995 (dismissed):
Please fix DCO and CI.

youjiali1995:
@sticnarf PTAL when you are free.

components/tidb_query_executors/src/runner.rs
```diff
@@ -395,7 +400,7 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
         BATCH_INITIAL_SIZE
     }
 
-    pub async fn handle_request(&mut self) -> Result<SelectResponse> {
+    pub async fn handle_request(&mut self) -> Result<(SelectResponse, Option<IntervalRange>)> {
```
sticnarf, line 403: Could you add some comments about the returned `IntervalRange`? For example, what does it mean, what will be the behavior if there are multiple ranges to select and whether there are constraint when multiple ranges are involved?

tests/failpoints/cases/test_coprocessor.rs
```diff
@@ -172,3 +175,50 @@ fn test_region_error_in_scan() {
             .contains("region seek error")
     );
 }
+
+#[test]
+fn test_paging_scan() {
```
sticnarf, line 180: I suggest we should validate the returned range of paging scan will cover the full range if the page limit is not reached. Because otherwise, we could be in a dead loop but cannot scan over the full range.

components/tidb_query_executors/src/runner.rs
sticnarf, line 403: Could you add some comments about the returned `IntervalRange`? For example, what does it mean, what will be the behavior if there are multiple ranges to select and whether there are constraint when multiple ranges are involved?

tests/failpoints/cases/test_coprocessor.rs
you06: Added a multi ranges test, in which the case drains the ranges is handled.

components/tidb_query_executors/src/runner.rs
you06: done

you06:
The selection duration increases in the latest commit. I'm investigating it now, hold this PR.

components/tidb_query_executors/src/runner.rs
```diff
@@ -424,10 +434,18 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
                 chunks.push(chunk);
             }
 
-            if drained {
+            if drained
+                || self
+                    .paging_size
+                    .map_or(false, |p| chunks.len() >= p as usize)
```
you06, line 440: I only changed this line from
`.map_or(false, |p| record_len >= p as usize)`
to
`.map_or(false, |p| chunks.len() >= p as usize)`

Because the stop condition does not make sense before though the performance looks great(stop when a batch scanned more rows than a batch? that's weird).

Then the duration of selection request increases, I tried changing it back, everything works fine :(

youjiali1995: I see. A `chunk` is a batch of data and `record_len` is how many rows are in this `chunk`. You should aggregate all chunks' length here.

you06: After adding some logs, I found the root cause. The duration of selection increases because of too large page. I was thought there will be more rows if using `record_len >= p`, that's wrong. Even if the client asked for a page with over 10,000 rows, 10 rows may be returned in this case.
If we use `chunks.len() >= p`, then for the first page, the default value will be 64, which would be a slow scan.
After I tuning the start paging size to 2, it's fine now. Anyway, that's out of this PR.

sticnarf (approved):
LGTM

you06:
I tested indexlookup in a single node with 50,000 records in a range, the result looks good now. @youjiali1995 PTAL

image

youjiali1995 (approved):

Issue #11303, opened 2021-11-09
Reduce the number of apply state keys


gengliqi:
## Development Task

The apply state(i.e. `RaftApplyState`) is used to ensure the atomicity of applying raft cmd.
However, I find the number of this key is too much to ignore.
I added a metric to count the normal key and apply state key.
TPCC workload
![image](https://user-images.githubusercontent.com/21253569/140898300-20c50e0d-3641-415e-9d94-a5b82d4a6e29.png)
Sysbench insert workload
![image](https://user-images.githubusercontent.com/21253569/140898871-624279e0-8a71-4661-bfb4-a639e4016595.png)

Since the WAL is buffered write, the duration of writing kv db mainly comes from memtable, which is dominated by the number of the written keys. 
Therefore, we need to reduce the number of apply state keys.

PR #11304, opened 2021-11-09
raftstore: batch apply msg to reduce apply state keys

gengliqi:

Issue Number: ref https://github.com/tikv/tikv/issues/11303

Problem Summary:
As https://github.com/tikv/tikv/issues/11303 said, this PR reduces some apply state keys through batching apply msg.
Although we know the apply state key can be significantly reduced by writing it regularly, there are some cases that need to think further. This PR uses a simple method to alleviate this problem. Adjacent apply msgs can be batched together.
The test result shows it can reduce about 30% apply state keys.

What's Changed:
Mentioned above.

### Related changes

### Check List 

Tests 

I compared this PR and master by running TPCC.
QPS of them are almost the same because the bottleneck is TiDB(16core)
image
image

When the apply thread is full, this PR can reduce much of apply duration.
image
So the async write duration is also reduced. 
image

The key flow of kv rocksdb shows the number of apply state key is reduced by 0.2 million.
image

Side effects

- No

### Release note 

```release-note
None.
```


NingLin-P:
Perhaps cache the updated `ApplyState` with `HashMap` and merge them to the `WriteBatch` before it is committed to RocksDB is much simpler and more efficient since it can handle well with non-adjacent apply msg.

gengliqi:
> Perhaps cache the updated `ApplyState` with `HashMap` and merge them to the `WriteBatch` before it is committed to RocksDB is much simpler and more efficient since it can handle well with non-adjacent apply msg.

I have tried to just set `RaftApplyState` before writing to db. I found a problem that the `RaftApplyState` is not included in `ApplyMetrics`. As #11303 said, the number of `RaftApplyState` can not be ignored.
I choose to batch adjacent apply msg just for simplicity and the result is good enough because most of the msg is apply msg.

NingLin-P:
> But I found the load imbalance in this version. I suspect the root cause is that RaftApplyState is not included in ApplyMetrics. As #11303 said, the number of RaftApplyState can not be ignored.

Sorry I don't get it, `ApplyMetrics` should be counting from `WriteBatch` directly, `RaftApplyState` should not be missed. Also, the `ApplyMetrics` is just for updating store level write flow, which mean we can add the `ApplyMetrics` to any region's `ApplyRes` or just add an additional message to do that.

> I choose to batch adjacent apply msg just for simplicity and the result is good enough because most of the msg is apply msg.

Still, I think the above-mentioned approach is much simpler

gengliqi:
> > But I found the load imbalance in this version. I suspect the root cause is that RaftApplyState is not included in ApplyMetrics. As #11303 said, the number of RaftApplyState can not be ignored.
> 
> Sorry I don't get it, `ApplyMetrics` should be counting from `WriteBatch` directly, `RaftApplyState` should not be missed. Also, the `ApplyMetrics` is just for updating store level write flow, which mean we can add the `ApplyMetrics` to any region's `ApplyRes` or just add an additional message to do that.
> 
> > I choose to batch adjacent apply msg just for simplicity and the result is good enough because most of the msg is apply msg.
> 
> Still, I think the above-mentioned approach is much simpler

If `RaftApplyState` is not pushed into `WriteBatch` before sending `ApplyResult`, the `ApplyMetrics` will miss this state. And then the `written_bytes` and `written_keys` will send to PD through heartbeat which is peer level, not store level.
The simplicity I mentioned means this method does not change the original behavior as much as possible. This is because I want this PR can be merged as soon as possible.
As the description said, we can significantly reduce the apply state keys by writing them regularly. So after this method is implemented and all the corner case is considered, the code of batching apply msg will be removed.

NingLin-P:
> If RaftApplyState is not pushed into WriteBatch before sending ApplyResult, the ApplyMetrics will miss this state. And then the written_bytes and written_keys will send to PD through heartbeat which is peer level, not store level.

I have a better idea, we can use `HashMap` to store `ApplyRes` too, then we can update it for each `Apply` msg instead of creating a new one, this way we can reduce `ApplyRes` msg as much as `RaftApplyState` write. So that we can also update the `ApplyMetrics` in `ApplyRes` for the `RaftApplyState`.

> The simplicity I mentioned means this method minimizes the change of original behavior as much as possible. This is because I want this PR can be merged as soon as possible.

I  think simplicity of code is also the key to let PR be merged quickly.

gengliqi:
> > If RaftApplyState is not pushed into WriteBatch before sending ApplyResult, the ApplyMetrics will miss this state. And then the written_bytes and written_keys will send to PD through heartbeat which is peer level, not store level.
> 
> I have a better idea, we can use `HashMap` to store `ApplyRes` too, then we can update it for each `Apply` msg instead of creating a new one, this way we can reduce `ApplyRes` msg as much as `RaftApplyState` write. So that we can also update the `ApplyMetrics` in `ApplyRes` for the `RaftApplyState`.
> 
> > The simplicity I mentioned means this method minimizes the change of original behavior as much as possible. This is because I want this PR can be merged as soon as possible.
> 
> I think simplicity of code is also the key to let PR be merged quickly.

I don't understand why you think this PR is *complex*. Also, the method you said should be no better performance improvement compared with this PR because most of apply msgs can be batched together.

The simplicity doesn't just mean less code. I choose to just batch Adjacent msgs because I worry about the reorder of different types of msgs which has a risk and I have not much time to think it through.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -472,14 +475,26 @@ where
...
 
     fn commit_opt(&mut self, delegate: &mut ApplyDelegate<EK>, persistent: bool) {
+        let has_unflushed_data =
+            delegate.last_flush_applied_index != delegate.apply_state.get_applied_index();
+        if !delegate.pending_remove && has_unflushed_data {
+            delegate.write_apply_state(self.kv_wb_mut());
+        }
         delegate.update_metrics(self);
+        // If has unflushed data, push a new ApplyRes
+        if has_unflushed_data {
```
BusyJay, line 489: Why change these orders?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2868,6 +2868,17 @@ impl<S: Snapshot> Apply<S> {
             }
         }
     }
+
+    fn try_batch(&mut self, other: &mut Apply<S>) -> bool {
+        assert_eq!(self.region_id, other.region_id);
+        if self.peer_id == other.peer_id && self.term == other.term {
```
BusyJay, line 2874: Why check for term?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3200,24 +3211,35 @@ where
         fail_point!("on_handle_apply_2", self.delegate.id() == 2, |_| {});
         fail_point!("on_handle_apply", |_| {});
 
-        if apply.entries.range.is_empty() || self.delegate.pending_remove || self.delegate.stopped {
+        if self.delegate.pending_remove || self.delegate.stopped {
```
BusyJay, line 3214: Why remove this?

gengliqi: Because `apply.entries` is a `SmallVec` now and it should be always non-empty.

components/raftstore/src/store/fsm/apply.rs
gengliqi: Removed.

NingLin-P:
> I don't understand why you think this PR is complex. Also, the method you said should be no better performance improvement compared with this PR because most of apply msgs can be batched together.

I didn't say this PR is complex, what I mean is, in relatively there is a better approach that is simpler and more efficient (it also reduces the number of `ApplyRes` msg) why not adopt it? If you are worried about any risk or complexity *this approach* may bring, would you describe more about them?

components/raftstore/src/store/peer.rs
```diff
@@ -2251,7 +2251,8 @@ where
                 cbs,
             );
             apply.on_schedule(&ctx.raft_metrics);
-            self.mut_store().trace_cached_entries(apply.entries.clone());
+            self.mut_store()
+                .trace_cached_entries(apply.entries[0].clone());
```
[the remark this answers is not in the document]
gengliqi: The `apply.entries` is a `SmallVec`. After initialization, it always has one `CachedEntries`.

BusyJay:
I also thought about the write state in flush before. But I think the current way also works and a bit easier to review as besides changing processing orders,  the logic is the same as before except batching up entries. Both ways are OK and acceptable.

components/raftstore/src/store/fsm/apply.rs
gengliqi: Change back to original code.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3200,24 +3214,35 @@ where
...
+        for cached_entries in apply.entries {
+            let (mut e, sz) = cached_entries.take_entries();
+            dangle_size += sz;
+            if e.is_empty() {
+                let rid = self.delegate.region_id();
+                let StdRange { start, end } = cached_entries.range;
+                e = Vec::with_capacity((end - start) as usize);
+                self.delegate
+                    .raft_engine
+                    .fetch_entries_to(rid, start, end, None, &mut e)
```
BusyJay, line 3233: Why not just `fetch_entries_to(rid, start, end, None, &mut entries)`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2868,6 +2869,19 @@ impl<S: Snapshot> Apply<S> {
             }
         }
     }
+
+    fn try_batch(&mut self, other: &mut Apply<S>) -> bool {
+        assert_eq!(self.region_id, other.region_id);
+        if self.peer_id == other.peer_id {
+            self.entries.append(&mut other.entries);
```
BusyJay, line 2876: May need flow control to avoid batching too many entries and cause OOM.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1732,7 +1732,7 @@ pub fn write_peer_state<T: Mutable>(
 pub struct CachedEntries {
     pub range: Range<u64>,
     // Entries and dangle size for them. `dangle` means not in entry cache.
-    entries: Arc<Mutex<(Vec<Entry>, usize)>>,
+    pub entries: Arc<Mutex<(Vec<Entry>, usize)>>,
```
BusyJay, line 1735: Why pub?

BusyJay (approved):

components/raftstore/src/store/peer_storage.rs
gengliqi: Oh I forgot to change it back

NingLin-P (approved):

PR #11308, opened 2021-11-09
tikv_util: ProcessCollector should use IntGauge to provide better performance

### LINKED ISSUE #11306: tikv_util: ProcessCollector should use IntGauge to provides better performance
https://github.com/tikv/tikv/blob/0dcb7b7d457b31cb9c78a1e60fee39597f4a24e3/components/tikv_util/src/metrics/process_linux.rs#L21-L28

procfs::process::Process all return value is integer not a float, should use IntGauge

```
/// The integer version of [`Gauge`]. Provides better performance if metric values are
/// all integers.
pub type IntGauge = GenericGauge<AtomicI64>;
```

I think ProcessCollector need these enhancement:
1. ProcessCollector should use IntGauge
2. remove procfs default chrono feature because it is unused
3. start_time is a **immutable/const** value should not calculate on each collect

pymongo:

improve ProcessCollector performance

close #11306 

### Check List

Tests

- Unit test

### Release note

```release-note
None
```

### DIFF
--- a/components/tikv_util/src/metrics/process_linux.rs
+++ b/components/tikv_util/src/metrics/process_linux.rs
@@ -7,31 +7,28 @@ use std::io::{Error, ErrorKind, Result};
 
 use lazy_static::lazy_static;
 use prometheus::core::{Collector, Desc};
-use prometheus::{proto, Counter, Gauge, Opts};
-use std::sync::Mutex;
+use prometheus::{proto, IntCounter, IntGauge, Opts};
 
 /// Monitors current process.
 pub fn monitor_process() -> Result<()> {
-    let pid = unsafe { libc::getpid() };
-    let tc = ProcessCollector::new(pid);
-    prometheus::register(Box::new(tc)).map_err(|e| Error::new(ErrorKind::Other, e.to_string()))
+    let pc = ProcessCollector::new();
+    prometheus::register(Box::new(pc)).map_err(|e| Error::new(ErrorKind::Other, e.to_string()))
 }
 
 /// A collector to collect process metrics.
 pub struct ProcessCollector {
-    pid: libc::pid_t,
     descs: Vec<Desc>,
-    cpu_total: Mutex<Counter>,
-    vsize: Gauge,
-    rss: Gauge,
-    start_time: Gauge,
+    cpu_total: IntCounter,
+    vsize: IntGauge,
+    rss: IntGauge,
+    start_time: IntGauge,
 }
 
 impl ProcessCollector {
-    pub fn new(pid: libc::pid_t) -> Self {
+    pub fn new() -> Self {
         let mut descs = Vec::new();
 
-        let cpu_total = Counter::with_opts(Opts::new(
+        let cpu_total = IntCounter::with_opts(Opts::new(
             "process_cpu_seconds_total",
             "Total user and system CPU time spent in \
                  seconds.",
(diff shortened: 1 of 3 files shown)


pymongo:
https://github.com/tikv/tikv/blob/0dcb7b7d457b31cb9c78a1e60fee39597f4a24e3/components/tikv_util/src/metrics/process_linux.rs#L21-L28

Why ProcessCollector field cpu_total need Mutex?

The prometheus::Counter(or IntCounter) has a *Arc<AtomicF64>* inside, I think the Mutex warp of Counter is **redundant**

components/tikv_util/src/metrics/process_linux.rs
```diff
@@ -21,45 +21,51 @@ pub fn monitor_process() -> Result<()> {
...
+    rss: IntGauge,
+    start_time: IntGauge,
 }
 
 impl ProcessCollector {
     pub fn new(pid: libc::pid_t) -> Self {
         let mut descs = Vec::new();
 
-        let cpu_total = Counter::with_opts(Opts::new(
+        let cpu_total = IntCounter::with_opts(Opts::new(
```
skyzh, line 34: I'm afraid that this might break Grafana functionalities for existing users, if we change float counter to int counter.

components/tikv_util/src/metrics/process_linux.rs
```diff
@@ -21,45 +21,51 @@ pub fn monitor_process() -> Result<()> {
...
             "process_start_time_seconds",
             "Start time of the process since unix epoch \
                  in seconds.",
         ))
         .unwrap();
         descs.extend(start_time.desc().into_iter().cloned());
+        // proc_start_time
+        if let Some(boot_time) = *BOOT_TIME {
+            if let Ok(p) = procfs::process::Process::new(pid) {
+                start_time.set(p.stat.starttime as i64 / *CLK_TCK + boot_time as i64);
```
skyzh, line 66: Why not simply call `procfs::boot_time_secs()` here?

skyzh (commented):
Generally LGTM

components/tikv_util/src/metrics/process_linux.rs
pymongo: Thanks, the `static BOOT_TIME` only call once, doesn't need to be static

components/tikv_util/src/metrics/process_linux.rs
pymongo: https://github.com/tikv/rust-prometheus/blob/d1528e11ab9dfe11a2528744ec7c42485b82ed1e/src/value.rs#L87-L102

```rust
match self.val_type {
    ValueType::Counter => {
        let mut counter = Counter::default();
        counter.set_value(val.into_f64());
        m.set_counter(counter);
    }
    ValueType::Gauge => {
        let mut gauge = Gauge::default();
        gauge.set_value(val.into_f64());
        m.set_gauge(gauge);
    }
}
```

the prometheus only has f64 numberic type link javascript, IntCounter/IntGauge finally would convert to f64

if we use IntGauge the benefit is only need to call once `as f64`

but if we use Gauge, all procfs int type must cast to f64 multitimes

pymongo:
since prometheus numerics data type only has f64 link javascript

https://github.com/tikv/rust-prometheus/blob/d1528e11ab9dfe11a2528744ec7c42485b82ed1e/src/value.rs#L87-L102

```rust
match self.val_type {
    ValueType::Counter => {
        let mut counter = Counter::default();
        counter.set_value(val.into_f64());
        m.set_counter(counter);
    }
    ValueType::Gauge => {
        let mut gauge = Gauge::default();
        gauge.set_value(val.into_f64());
        m.set_gauge(gauge);
    }
}
```

both IntGauge and Gauge would convert to f64 at the last, doesn't need to worry about not compatible to users.

---

I think `cpu_totals: GaugeVec` in thread_linux.rs can also refactor to `IntGaugeVec`

but it relative to a lot of files and code, it would better to edit it in another PR?

components/tikv_util/src/metrics/process_linux.rs
pymongo: prometheus only has float/string two data type or literal from their document:

https://prometheus.io/docs/prometheus/latest/querying/basics/#float-literals

skyzh:
> but it relative to a lot of files and code, it would better to edit it in another PR?

@pymongo It's better to do this in another PR. But before proceeding, it's better to see if we really need to do this.

cc @breeswish @hicqu would you please take a look?

zhongzc (approved):
Nice improvement.

skyzh (approved):

zhongzc:
The code is port from [rust-prometheus](https://github.com/tikv/rust-prometheus). Can you help improve rust-prometheus in the same way?

pymongo:
> @pymongo The code is ported from [rust-prometheus](https://github.com/tikv/rust-prometheus). Can you help improve rust-prometheus in the same way?

Yeah, my company's project is using rust-prometheus also, but recent I am block on reviewing some **procfs** crate's PR and spent a lot of time to improve procfs code, I would read rust-prometheus source and make some enhancement later

PR #11310, opened 2021-11-09
raftstore: reduce the frequency of flushing raft messages

### LINKED ISSUE #11309: Reduce the frequency of flushing raft message
## Development Task

https://github.com/tikv/tikv/pull/11111 introduced a config `raft-msg-flush-interval-us` which is used to reduce the frequency of flushing raft messages. 
However, after some investigation, I found many flush events come from the `pause` function when the store threads are not full.

gengliqi:

Issue Number: close #11309

Problem Summary:
See #11309.

What's Changed:
If there is no active fsm in store thread, it will wait an additional `raft_msg_flush_interval_us`, if there is still no active fsm, the `pause` function should be called.

This PR also changes some other small parts
1. change the default `store-io-pool-size` to 1 because if it's basically enough for other default configs (e.g. 2 store threads, 2 apply threads and 5 grpc threads) in most cases. Users can increase this config and other relevant configs to achieve better performance if the CPU is sufficient.
2. remove the config `cmd-batch-concurrent-ready-max-count` from `config-template.toml` because this config should not be changed in almost all cases. (BTW, the config name may be changed in future)
3. `raft_entry_max_size` can be updated dynamically. But the `BatchRaftCmdRequestBuilder` always uses the original one.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No.

### Release note 

```release-note
None.
```


components/raftstore/src/store/config.rs
```diff
@@ -299,7 +299,7 @@ impl Default for Config {
             local_read_batch_size: 1024,
             apply_batch_system: BatchSystemConfig::default(),
             store_batch_system: BatchSystemConfig::default(),
-            store_io_pool_size: 2,
+            store_io_pool_size: 1,
```
BusyJay, line 302: Why?

components/batch-system/src/config.rs
```diff
@@ -11,6 +11,7 @@ pub struct Config {
     pub pool_size: usize,
     pub reschedule_duration: ReadableDuration,
     pub low_priority_pool_size: usize,
+    pub before_pause_wait_us: Option<u64>,
```
BusyJay, line 14: How about reusing `raft_msg_flush_interval_us`?

components/raftstore/src/store/config.rs
gengliqi: You can see the description.

components/batch-system/src/config.rs
gengliqi: This config is useless for apply batch system. And the `raft_msg_flush_interval_us` belongs to raftstore not batch system so I think just using two configs is ok.
The default value of store batch system's `before_pause_wait_us` is `raft_msg_flush_interval_us `.

BusyJay: You can mark it skip serde skip and use `raft_msg_flush_interval_us` to initialize the value.

components/batch-system/src/config.rs
```diff
@@ -11,6 +11,9 @@ pub struct Config {
     pub pool_size: usize,
     pub reschedule_duration: ReadableDuration,
     pub low_priority_pool_size: usize,
+    #[doc(hidden)]
+    #[serde(skip_serializing)]
```
BusyJay, line 15: It should be `#[serde(skip)]`.

components/batch-system/src/batch.rs
```diff
@@ -273,9 +274,25 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
...
-            if let Ok(fsm) = self.fsm_receiver.recv() {
-                return batch.push(fsm);
+            if let Some(d) = self.before_pause_wait_us {
+                channel::select! {
+                    recv(self.fsm_receiver) -> msg => {
+                        if let Ok(fsm) = msg {
+                            return batch.push(fsm);
+                        }
+                    }
+                    recv(tick(Duration::from_micros(d))) -> _ => {
```
BusyJay, line 284: Why not use `after`?

BusyJay (commented):
Can you also post the improvement in PR?

Rest LGTM

gengliqi:
From left to right, master, this PR with 250us `raft_msg_flush_interval_us `, this PR with 500us `raft_msg_flush_interval_us `
gsMmZdoUJR

components/raftstore/src/store/config.rs
```diff
@@ -489,6 +489,8 @@ impl Config {
         } else {
             self.store_batch_system.max_batch_size = Some(1024);
         }
+        self.store_batch_system.before_pause_wait_us =
+            Some(std::cmp::min(self.raft_msg_flush_interval_us, 1000));
```
BusyJay, line 493: I suggest to change the time to `ReadableDuration` for consistency.

BusyJay (approved):
Rest LGTM

Little-Wallace (approved):
LGTM

PR #11338, opened 2021-11-10
tikv_util: support microsecond unit for ReadableDuration

### LINKED ISSUE #11337: Support microsecond unit for ReadableDuration
## Development Task
The minimum unit of ReadableDuration is a millisecond for now. 
However, the microsecond unit is also useful for some configs. e.g.  1 millisecond may be too long for delaying flushing raft messages.

gengliqi:

Issue Number: close #11337

Problem Summary:
Support microsecond unit for `ReadableDuration`.

What's Changed:
This PR also renames `raft-msg-flush-interval-us` to `raft-msg-flush-interval` and change its type from `u64` to `ReadableDuration`.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No

### Release note 

```release-note
Support microsecond unit for configuration.
```


components/tikv_util/src/config.rs
```diff
@@ -463,9 +475,14 @@ impl fmt::Display for ReadableDuration {
...
         }
+        if dur >= MS {
+            written = true;
+            write!(f, "{}ms", dur / MS)?;
+            dur %= MS;
+        }
         if dur > 0 {
             written = true;
-            write!(f, "{}ms", dur)?;
+            write!(f, "{}us", dur)?;
```
BusyJay, line 485: This seems a incompatible change. If a common configuration is set to `us` in the latest version and then rollback to older version, it will fail to start. I think it's acceptable, but better clarify in some places.

components/raftstore/src/store/config.rs
```diff
@@ -717,8 +719,8 @@ impl Config {
             .with_label_values(&["io_reschedule_hotpot_duration"])
             .set(self.io_reschedule_hotpot_duration.as_secs() as f64);
         CONFIG_RAFTSTORE_GAUGE
-            .with_label_values(&["raft_msg_flush_interval_us"])
-            .set(self.raft_msg_flush_interval_us as f64);
+            .with_label_values(&["raft_msg_flush_interval"])
+            .set(self.raft_msg_flush_interval.as_secs() as f64);
```
BusyJay, line 723: This can always be 0.

components/tikv_util/src/config.rs
gengliqi: Got it. I will update the doc. https://docs.pingcap.com/zh/tidb/stable/command-line-flags-for-tikv-configuration#tikv-%E9%85%8D%E7%BD%AE%E5%8F%82%E6%95%B0

components/raftstore/src/store/config.rs
```diff
@@ -717,8 +719,8 @@ impl Config {
             .with_label_values(&["io_reschedule_hotpot_duration"])
             .set(self.io_reschedule_hotpot_duration.as_secs() as f64);
         CONFIG_RAFTSTORE_GAUGE
-            .with_label_values(&["raft_msg_flush_interval_us"])
-            .set(self.raft_msg_flush_interval_us as f64);
+            .with_label_values(&["raft_msg_flush_interval"])
+            .set(self.raft_msg_flush_interval.as_micros() as f64);
```
BusyJay, line 723: I think you need a new method that returns seconds as f64.

BusyJay: It seems to be supported: https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f64

BusyJay (approved):

components/tikv_util/src/time.rs
```diff
@@ -29,9 +29,17 @@ pub fn duration_to_sec(d: Duration) -> f64 {
     d.as_secs() as f64 + (nanos / 1_000_000_000.0)
 }
 
+/// Converts Duration to microseconds.
+#[inline]
+pub fn duration_to_us(d: Duration) -> u64 {
+    let nanos = u64::from(d.subsec_nanos());
+    // Most of case, we can't have so large Duration, so here just panic if overflow now.
+    d.as_secs() * 1_000_000 + (nanos / 1_000)
```
NingLin-P, line 37: Actually, when compile in release mode, integer overflow will cause wrap around silently instead of panic.

gengliqi: Update the comments

NingLin-P (approved):

PR #11347, opened 2021-11-12
Remove duplicate enum and macro about numberic_enum_mod

### LINKED ISSUE #11297: Remove duplicate `perf_level_serde`
## Development Task
Currently, there have two `perf_level_serde` in `engine_rocks` and `engine_traits`,we can freely delete one of them

```
pub use crate::perf_context::PerfLevel;

// FIXME This is duplicated from engine_rocks
...

numeric_enum_mod! {perf_level_serde PerfLevel {
    Uninitialized = 0,
    Disable = 1,
    EnableCount = 2,
    EnableTimeExceptForMutex = 3,
    EnableTimeAndCPUTimeExceptForMutex = 4,
    EnableTime = 5,
    OutOfBounds = 6,
}}
```

yuqi1129:

Issue Number: close #11297 

Problem Summary:

What's Changed:

Remove duplicated enum and related macros about `PerfLevel`

### Related changes

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
 None.
```


tabokie:
Oh, I forgot this one. You shouldn't delete the enum in `engine_trait`.

The `engine_trait` is an abstraction of RocksDB engine, meaning we should be able to plug in different kv engines under the same interface set. That's why we should accept generic configuration (`engine_trait::PerfLevel`) and convert it to specific configuration (`engine_rocks::PerfLevel`) when entering specific engine implementation.

Instead, you can delete the `engine_rocks::perf_level_serde`. That doesn't seem useful anymore.

yuqi1129:
@tabokie Sorry to miss this, Got it

yuqi1129:
@tabokie Maybe we should remove duplicated macro `numeric_enum_mod` and `rocksdb_enum_mod` either in this issue. For example put the macors in the crate `tikv-util`

tabokie:
@yuqi1129 That's a good idea. You can try it.

components/tikv_util/src/config.rs
```diff
@@ -689,6 +689,88 @@ pub fn check_max_open_fds(expect: u64) -> Result<(), ConfigError> {
...
 
+#[macro_export]
+macro_rules! rocksdb_enum_mod {
+    ($name:ident $enum:ident { $($variant:ident = $value:expr, )* }) => {
+        pub mod $name {
+            use std::fmt;
+
+            use serde::{Serializer, Deserializer};
+            use serde::de::{self, Unexpected, Visitor};
+            use crate::$enum;
```
tabokie, line 700: How about this, more flexible.
```suggestion
            use super::$enum;
```

components/engine_rocks/src/config.rs
```diff
@@ -5,6 +5,10 @@ use rocksdb::{DBCompressionType, DBInfoLogLevel, DBTitanDBBlobRunMode};
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
+pub use rocksdb::{CompactionPriority, DBCompactionStyle, DBRateLimiterMode, DBRecoveryMode};
```
tabokie, line 8: And remove the pub use.

components/engine_traits/src/lib.rs
```diff
@@ -253,12 +253,11 @@
...
 #![feature(min_specialization)]
 
-#[allow(unused_extern_crates)]
-extern crate tikv_alloc;
 #[cfg(test)]
-#[macro_use]
 extern crate serde_derive;
 extern crate slog_global;
+#[allow(unused_extern_crates)]
+extern crate tikv_alloc;
```
tabokie, line 260: Better leave these unchanged.

components/tikv_util/src/config.rs
```diff
@@ -689,6 +689,88 @@ pub fn check_max_open_fds(expect: u64) -> Result<(), ConfigError> {
     }
 }
 
+#[macro_export]
+macro_rules! rocksdb_enum_mod {
```
tabokie, line 693: rename to numeric_enum_serializing_mod

yuqi1129: OK，The name that contains 'rocksdb' is indeed improper as this is a common lib

components/engine_traits/src/lib.rs
yuqi1129: This is recommended by `cargo format`

components/engine_rocks/src/config.rs
yuqi1129: https://github.com/tikv/tikv/blob/8017333565f3525fcb49563590cd13375e596e58/components/engine_traits/src/config.rs#L3

Sounds we should also change here

components/tikv_util/src/config.rs
yuqi1129: Got it

components/engine_traits/src/lib.rs
yuqi1129: 🆗

components/tikv_util/src/config.rs
```diff
@@ -689,6 +689,88 @@ pub fn check_max_open_fds(expect: u64) -> Result<(), ConfigError> {
     }
 }
 
+#[macro_export]
```
tabokie, line 692: Right now this macro is standing between `#[cfg(unix)] pub fn check_max_open_fds` and `#[cfg(not(unix))] pub fn check_max_open_fds`. You should move it to the end.

tabokie (approved):
Rest looks good

components/engine_traits/src/config.rs
```diff
@@ -1,79 +1,9 @@
...
-                        assert!(h == holder);
-                    }
-                }
-            }
-        }
-    }
-}
-
-numeric_enum_mod! {perf_level_serde PerfLevel {
+numeric_enum_serializing_mod! {perf_level_serde PerfLevel {
```
Connor1996, line 6: move it `crate::perf_context`?

Connor1996: and can we step further to define the enum in the macro, so we don't need to duplicate the perf level's definition in two places

components/engine_rocks/src/config.rs
```diff
@@ -262,124 +266,33 @@ impl From<BlobRunMode> for DBTitanDBBlobRunMode {
...
-
-            use serde::{Serializer, Deserializer};
-            use serde::de::{self, Unexpected, Visitor};
-            use rocksdb::$enum;
-            use case_macros::*;
-
-            pub fn serialize<S>(mode: &$enum, serializer: S) -> Result<S::Ok, S::Error>
-                where S: Serializer
-            {
-                serializer.serialize_i64(*mode as i64)
```
tabokie, line 278: One more thing, use `$value:expr` here, considering `DBRateLimiterMode` starts with 1.

tabokie: And add a test asserting serialized string equals original string.

components/engine_traits/src/config.rs
yuqi1129: Good idea~

components/engine_rocks/src/config.rs
yuqi1129: > One more thing, use $value:expr here, considering DBRateLimiterMode starts with 1.

Seems DBRateLimiterMode starts with 1 not 0 does not affect the result, i have test both use 
```
serializer.serialize_i64(*mode as i64)
```
and 
```
 match mode {
         $( $enum::$variant => serializer.serialize_i64($value as i64), )*
  }
```
Test code is :

```
        let x = r#"
            rate-limiter-mode = "read-only"
        "#;

        let config: DbConfig = toml::from_str(x).unwrap();
        assert_eq!(config.rate_limiter_mode, DBRateLimiterMode::ReadOnly);

        println!("{:?}", toml::to_string(&config).unwrap());
        assert!(
            toml::to_string(&config)
                .unwrap()
                .contains("rate-limiter-mode = 1")
        );
```

Both can pass, Still, shall we use to latter form anyhow?


> And add a test asserting serialized string equals original string.

I can't catch your point. Currently any config value is serialized as numeric value not matter what the input format is, what's the meaning of  `serialized string`?

tabokie: Oh, it's because rust-rocksdb codes it from 1 too: https://github.com/tikv/rust-rocksdb/blob/06aa625e21b0b2c632cc02b7d001e6d5671bbc38/librocksdb_sys/src/lib.rs#L379. Regardless of that, using `$value` is better.

And I mean testing `toml::to_string(toml::from_str(str)) == str`, where str is "rate-limiter-mode = 1, ..."

Connor1996 (approved):
LGTM

components/engine_rocks/src/config.rs
yuqi1129: >  And I mean testing toml::to_string(toml::from_str(str)) == str, where str is "rate-limiter-mode = 1, ..." 

The test `test_serde_custom_tikv_config` in integration test has cover the case , but seems that `make test` will not trigger integration test. anyway, i wiil add some test about `rate-limiter-mode`

PR #11352, opened 2021-11-12
resolved_ts: return early after regions already had quorum

### LINKED ISSUE #11351: resolved ts lag increased after stoping a tikv
## Bug Report

image

Here it will wait until all check leader requests returned, which may cause resolved ts lag increased if one tikv was disconnected.

https://github.com/tikv/tikv/blob/999ca556005ed5c3b7ccbceb511806a5da365c45/components/cdc/src/endpoint.rs#L1072

5kbpers:

Issue Number: close #11351 

Problem Summary:

What's Changed:
* export `region_resolved_ts_store` from `resolved_ts` and import it in `cdc`
* use `select_all` instead of `join_all` for returning early

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Fix resolved ts lag increased after stoping a tikv
```


NingLin-P (approved):
LGTM

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +137,140 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
-        pd_client: Arc<dyn PdClient>,
-        security_mgr: Arc<SecurityManager>,
-        env: Arc<Environment>,
-        cdc_clients: Arc<Mutex<HashMap<u64, TikvClient>>>,
-        min_ts: TimeStamp,
-    ) -> Vec<u64> {
-        #[cfg(feature = "failpoint")]
-        (|| fail_point!("before_sync_replica_read_state", |_| regions))();
+// Confirms leadership of region peer before trying to advance resolved ts.
+// This function broadcasts a special message to all stores, get the leader id of them to confirm whether
```
liuzix, line 143:
```suggestion
// This function broadcasts a special message to all stores, gets the leader id of them to confirm whether
```

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +137,140 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
+// current peer has a quorum which accept its leadership.
```
liuzix, line 144:
```suggestion
// current peer has a quorum which accepts its leadership.
```

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +137,140 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
-                    .into_iter()
-                    .map(move |region_id| (store_id, region_id))
-            })
-            .flatten()
-            .for_each(|(store_id, region_id)| {
+            .boxed()
+        })
+        .collect();
+    for _ in 0..store_count {
+        let (res, _, remains) = select_all(stores).await;
```
liuzix, line 226: Can you add comments on why using `select_all` as opposed to the original solution using `join_all`?

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +137,140 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
-                    );
-                    None
+                if region_has_quorum(&region_map[&region_id], &resp_map[&region_id]) {
+                    valid_regions.insert(region_id);
                 }
-            })
-            .collect()
+            }
+        }
+        if valid_regions.len() == regions.len() {
```
liuzix, line 236:
```suggestion
        if valid_regions.len() == regions.len() {
            // break here because all regions have quorum,
            // so there is no need waiting for other stores to respond.
```

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +137,140 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
-            .map_or(0, |regions| regions[0].compute_size());
-        let stores = store_map.into_iter().map(|(store_id, regions)| {
-            let cdc_clients = cdc_clients.clone();
+        region_map.insert(region_id, peer_list);
+    }
+    // Approximate `LeaderInfo` size
+    let leader_info_size = store_map
+        .values()
+        .next()
+        .map_or(0, |regions| regions[0].compute_size());
```
liuzix, line 198: Can there be `0` regions?

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +137,140 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
-        // region_id -> `(Vec<Peer>, LeaderInfo)`
-        let info_map = region_read_progress.dump_leader_infos(&regions);
+    // store_id -> leaders info, record the request to each stores
+    let mut store_map: HashMap<u64, Vec<LeaderInfo>> = HashMap::default();
+    // region_id -> region, cache the information of regions
+    let mut region_map: HashMap<u64, Vec<Peer>> = HashMap::default();
+    // region_id -> peers id, record the responses
+    let mut resp_map: HashMap<u64, Vec<u64>> = HashMap::default();
+    // region_id -> `(Vec<Peer>, LeaderInfo)`
+    let info_map = region_read_progress.dump_leader_infos(&regions);
```
liuzix, line 170: How expensive is this function?

5kbpers: O(n), n is the number of regions

components/resolved_ts/src/advance.rs
```diff
@@ -135,129 +140,144 @@ impl<E: KvEngine> AdvanceTsWorker<E> {
...
+    min_ts: TimeStamp,
+) -> Result<(u64, CheckLeaderResponse)> {
+    if tikv_clients.lock().unwrap().get(&store_id).is_none() {
+        let store = box_try!(pd_client.get_store_async(store_id).await);
+        let cb = ChannelBuilder::new(env.clone());
+        let channel = security_mgr.connect(cb, &store.address);
+        tikv_clients
+            .lock()
+            .unwrap()
+            .insert(store_id, TikvClient::new(channel));
```
overvenus, line 266: When do we release these connections?

NingLin-P: There is an issue https://github.com/tikv/tikv/issues/11400 about it, I think we can fix it in another PR.

5kbpers: Fixed

overvenus (approved):
LGTM

hicqu (approved):

Issue #11348, opened 2021-11-12
tidb-cluster can not run pod of basic-tikv-0


fhj939057500:
### What version of TiKV are you using?

5.2.1
### What operating system and CPU are you using?

Linux TENCENT64.site  x86_64 x86_64 x86_64 GNU/Linux
### Steps to reproduce

1.Refer to the official website document，quickly run tidb-cluster.
2.Execute instructions：
    kubectl -n tidb-cluster apply -f https://raw.githubusercontent.com/pingcap/tidb-operator/master/examples/basic/tidb-cluster.yaml
3. kubectl logs basic-tikv-0 -n tidb-cluster,find log :
    ParseIntError { kind: PosOverflow}' , components/tikv_util/src/cgroupp.rs.251:18
3. find basic

### What did you expect?
fix bug

sticnarf:
It's because we cannot handle a limit that is larger than 2^63-1. I think we can treat `PosOverflow` error same as max.

sticnarf:
@fhj939057500 Could you tell more about your kubernetes environment?

Are you using a managed k8s service provided by cloud, or you deployed your own k8s?

fhj939057500:
> @fhj939057500 Could you tell more about your kubernetes environment?
> 
> Are you using a managed k8s service provided by cloud, or you deployed your own k8s?

yes,it.s run by: https://docs.pingcap.com/zh/tidb-in-kubernetes/stable/get-started#%E8%BF%9E%E6%8E%A5-tidb-%E9%9B%86%E7%BE%A4

sticnarf:
Closed by #11354

PR #11354, opened 2021-11-12
fix a panic about parsing memory.limit_in_bytes

### LINKED ISSUE #11353: cgroup code parses memory.limit_in_bytes fail
### What did happened?
图片

hicqu:

Issue Number: close #11353.

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```


components/tikv_util/src/sys/cgroup.rs
```diff
@@ -247,7 +248,12 @@ fn parse_memory_max(line: &str) -> i64 {
     if line == "max" {
         return -1;
     }
-    line.parse().unwrap()
+    match line.parse::<i64>() {
+        Ok(x) => x,
+        Err(e) if matches!(e.kind(), IntErrorKind::PosOverflow) => i64::MAX,
+        Err(e) if matches!(e.kind(), IntErrorKind::NegOverflow) => i64::MIN,
```
sticnarf, line 254: What about treating any wrong value as no limit? `i64::MIN` is not so meaningful here, as any negative value is treated as no limit.

BusyJay, line 254: This is not covered.

sticnarf:
Also closes https://github.com/tikv/tikv/issues/11348

components/tikv_util/src/sys/cgroup.rs
hicqu: I prefer to keep the raw value, because they will be log out at components/tikv_util/src/sys/mod.rs:72.

sticnarf: If the value is negative, the logged value is also strange after casted to u64. Maybe it's better to fix the logging?

hicqu: The log line is also updated in the PR. It prints i64 instead of casted u64, so we can know the original value from that.

sticnarf (approved):
LGTM

BusyJay (approved):

PR #11357, opened 2021-11-12
codec: add compile time assert to guarantee safety

### LINKED ISSUE #11356: `try_decode_first_internal` is unsafe as incorrect implemented trait can lead to undefined behavior
## Development Task

If `T::get_raw_padding_ptr()` returns a pointer, which points to a buffer less than `padding_size`, out of bound access may occur in `memcmp`.

https://github.com/tikv/tikv/blob/37e3395231bfb477895884126734526e898f9d26/components/codec/src/byte.rs#L374-L380

Fortunately, the trait is correctly implemented for all structs currently.

BusyJay:

Issue Number: close #11356.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/codec/src/byte.rs
+++ b/components/codec/src/byte.rs
@@ -367,7 +367,9 @@ impl MemComparableByteCodec {
                     // is faster than checking pad bytes one by one, since it will compare multiple
                     // bytes at once.
                     let base_padding_ptr = dest_ptr.sub(padding_size);
-                    // Force a compile time check to ensure safety. The check will be optimized away.
+                    // Force a compile time check to ensure safety. The check will be optimized away
+                    // if PADDING's size is larger than MEMCMP_GROUP_SIZE, because it's checked
+                    // aboved that padding_size <= MEMCMP_GROUP_SIZE.
                     let expected_padding_ptr = T::PADDING[..padding_size].as_ptr();
                     let cmp_result = libc::memcmp(
                         base_padding_ptr as *const libc::c_void,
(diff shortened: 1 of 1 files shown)


skyzh (commented):
Why is this a compile-time check? It seems that `padding_size` is calculated by `let padding_size = T::parse_padding_size(*src_ptr.add(MEMCMP_GROUP_SIZE));`, which can only be known at runtime.

components/codec/src/byte.rs
```diff
@@ -372,7 +367,8 @@ impl MemComparableByteCodec {
                     // is faster than checking pad bytes one by one, since it will compare multiple
                     // bytes at once.
                     let base_padding_ptr = dest_ptr.sub(padding_size);
-                    let expected_padding_ptr = T::get_raw_padding_ptr();
+                    // Force a compile time check to ensure safety. The check will be optimized away.
+                    let expected_padding_ptr = T::PADDING[..padding_size].as_ptr();
```
breeswish, line 371: padding_size is not a constant but a variable, how would compiler discover that there is out of bound possibility?

BusyJay: Because L362 has checked that `padding_size <= MEMCMP_GROUP_SIZE`, and the len of `T::PADDING` is exact `MEMCMP_GROUP_SIZE`. /cc @skyzh

skyzh: This sounds reasonable to me.

skyzh (approved):

breeswish (approved):

BusyJay:
/test

The failed case is traced at #11365.

PR #11385, opened 2021-11-16
cdc: introduce TsFilter into incremental scan

### LINKED ISSUE #11384: TsFilter may be helpful to speed up CDC incremental scan
## Development Task

Currently CDC incremental scan iterators step over all SST files covered by the relative Regions. However generally incremental data size after a given CDC checkpoint is much less than the total data size, so it wastes I/O resource.

So maybe it's possible to use `TsFilter` to filter SST files which doesn't contain any changes after given CDC checkpoints.

But it needs an extra operation to fetch old values because they can be filtered out by `TsFilter`, so performance could drop in some extra cases. We need to test it carefully.

hicqu:

Issue Number: close #11384.

* Add a new configuration `cdc.incremental-scan-use-ts-filter`, which indicates `TsFilter` can be used for incremental scan or not
* Only use TsFilter if it can filter more than 2/3 recoreds
* Use `OldValueReader` to re-fetch old values when `TsFilter` is used

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/cdc/src/endpoint.rs
+++ b/components/cdc/src/endpoint.rs
@@ -2,17 +2,21 @@
 
 use std::f64::INFINITY;
 use std::fmt;
-use std::marker::PhantomData;
 use std::sync::{Arc, Mutex as StdMutex};
 use std::time::Duration;
 
 use collections::{HashMap, HashSet};
 use concurrency_manager::ConcurrencyManager;
 use crossbeam::atomic::AtomicCell;
-use engine_traits::{KvEngine, Snapshot as EngineSnapshot};
+use engine_rocks::PROP_MAX_TS;
+use engine_traits::{
+    KvEngine, Range, Snapshot as EngineSnapshot, TablePropertiesCollection, TablePropertiesExt,
+    UserCollectedProperties, CF_WRITE,
+};
 use fail::fail_point;
 use futures::compat::Future01CompatExt;
 use grpcio::Environment;
+use keys::{data_end_key, data_key};
 use kvproto::cdcpb::{
     ChangeDataRequest, ClusterIdMismatch as ErrorClusterIdMismatch,
     DuplicateRequest as ErrorDuplicateRequest, Error as EventError, Event, Event_oneof_event,
@@ -34,9 +38,10 @@ use security::SecurityManager;
 use tikv::config::CdcConfig;
 use tikv::storage::kv::{PerfStatisticsInstant, Snapshot};
 use tikv::storage::mvcc::{DeltaScanner, ScannerBuilder};
-use tikv::storage::txn::TxnEntry;
-use tikv::storage::txn::TxnEntryScanner;
+use tikv::storage::txn::{TxnEntry, TxnEntryScanner};
+use tikv::storage::{Cursor, Statistics};
 use tikv_kv::PerfStatisticsDelta;
+use tikv_util::codec::number;
 use tikv_util::sys::inspector::{self_thread_inspector, ThreadInspector};
 use tikv_util::time::{Instant, Limiter};
 use tikv_util::timer::SteadyTimer;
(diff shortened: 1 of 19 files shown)


components/cdc/src/endpoint.rs
```diff
@@ -1437,6 +1493,53 @@ impl Initializer {
...
+        let collection = match self.engine.table_properties_collection(CF_WRITE, &[range]) {
+            Ok(collection) => collection,
+            Err(_) => return false,
+        };
+
+        let hint_min_ts = self.checkpoint_ts.into_inner();
+        let (mut total_count, mut filtered_count, mut tables) = (0, 0, 0);
+        collection.iter_user_collected_properties(|prop| {
+            tables += 1;
+            if let Some((_, keys)) = prop.approximate_size_and_keys(&start_key, &end_key) {
```
hicqu, line 1514: Here @tabokie . I think comparing with the scan itself, this price is ok.

tabokie: Got it

src/storage/txn/store.rs
```diff
@@ -135,16 +135,68 @@ pub enum TxnEntry {
...
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub enum OldValue {
+    Some(Value),
+    None,
+    /// Not sure whether the old value exists or not because it can be filtered by `hint_min_ts`.
+    /// So users can seek CF_WRITE to the given position to take a look.
+    SeekWrite(Key),
+}
```
overvenus, line 168: It's similar to `txn_types::OldValue`, could you unify them?

components/cdc/src/metrics.rs
```diff
@@ -5,6 +5,10 @@ use prometheus::*;
 use prometheus_static_metric::*;
 use std::cell::RefCell;
 use tikv::storage::kv::PerfStatisticsDelta;
+use tikv::storage::Statistics;
+
+pub const TAG_DELTA_CHANGE: &str = "delta_change";
+pub const TAG_INCREMENTAL_SCAN: &str = "incremental_scan";
```
overvenus, line 11: Could you add some comments about their meaning?

components/cdc/src/endpoint.rs
```diff
@@ -1066,26 +1077,36 @@ impl Initializer {
...
+            hint_min_ts = Some(self.checkpoint_ts);
+            old_value_reader = Some(OldValueReader::new(snap.clone()));
+        }
 
-        let start = Instant::now_coarse();
-        // Time range: (checkpoint_ts, current]
-        let current = TimeStamp::max();
-        let mut scanner = ScannerBuilder::new(snap, current)
+        // Time range: (checkpoint_ts, max]
+        let mut scanner = ScannerBuilder::new(snap, TimeStamp::max())
```
overvenus, line 1099: @amyangfei PTAL is it ok to scan `(checkpoint_ts, max]`?

components/cdc/src/old_value.rs
```diff
@@ -194,7 +196,7 @@ pub fn get_old_value<S: EngineSnapshot>(
                         (value, Some(statistics))
                     }
                     // Unspecified should not be added into cache.
```
overvenus, line 198:
```suggestion
                    // Unspecified and SeekWrite should not be added into cache.
```

src/storage/mvcc/reader/scanner/forward.rs
```diff
@@ -504,15 +504,15 @@ impl<S: Snapshot> ScanPolicy<S> for LatestEntryPolicy {
                     break Some(TxnEntry::Commit {
                         default: entry_default,
                         write: entry_write,
-                        old_value: OldValue::None,
+                        old_value: OldValue::Unspecified,
```
overvenus, line 507: Could you elaborate why change to `Unspecified`?

hicqu: Fixed by updating `tests::EntryBuilder`.

overvenus (approved):
LGTM

liuzix (approved):

PR #11388, opened 2021-11-16
 Support to iterate table properties for KvEngine 

### LINKED ISSUE #11387: Support to iterate table properties for KvEngine
## Development Task

Currently how to get a `MvccProperties` for a given range:
```rust
let mvcc_properties = Default::default();
let collections = engine.get_table_properties_collection(range);
for (_, table_properties) in collections {
    let user_properties = table_properties.user_collected_properties();
    let one_mvcc = user_properties.decode();
    mvcc_properties.add(one_mvcc);
}
```

Now I have someting like `MvccProperties` based on `UserCollectedProperties` from all SST files in a given range. So I think it's necessary to introduce `TablePropertiesCollection`, `UserCollectedProperties` to `KvEngine` trait. Then I can calculate my things based on it.

hicqu:

Issue Number: close #11387

### Check List 

Tests 

- No code

Side effects

### Release note 

```release-note
None
```


hicqu:
@brson would you like to take a look?

components/engine_rocks/src/table_properties.rs
```diff
@@ -1,9 +1,44 @@
...
+    fn get(&self, index: &[u8]) -> Option<&[u8]> {
+        self.0.get(index)
+    }
+}
+
+#[repr(transparent)]
+pub struct TablePropertiesCollection(rocksdb::TablePropertiesCollection);
+impl engine_traits::TablePropertiesCollection for TablePropertiesCollection {
+    type UserCollectedProperties = UserCollectedProperties;
+    fn iter_user_collected_properties<F>(&self, mut f: F)
```
Connor1996, line 18: can we implement `Iterator` for it?

hicqu: For the rocksdb implementation it's hard because `Iterator` is implemented for `rocksdb::TablePropertiesCollectionIter` instead of `rocksdb::TablePropertiesCollection`. So exposing an `Iterator` will require another `struct CollectionIter`. So here I use a closure to avoid too many structs.

Connor1996 (approved):
LGTM

yiwu-arbug (approved):

Issue #11404, opened 2021-11-17
raftstore: TiKV can not deleted the truncated raft log as soon as possible


Little-Wallace:
### What version of TiKV are you using?

5.1.2
### Steps to reproduce

I have add some metrics in TiKV in PR: https://github.com/tikv/tikv/pull/11377/files
And we find that TiKV was busy with write the deleted keys to RaftDB. Because most of GC Batch is small but the number of regions which need to clean is large.
![image](https://user-images.githubusercontent.com/16376959/142194352-18790a91-8b33-492b-8901-b93a46070186.png)

### What did you expect?
I hope TiKV can consume the GC task as soon as possible,

### What did happened?
There are more and more data in RaftDB.  I got thousands  of SST files in RaftDB and it usually was small (maybe tens).

VelocityLight:
update

PR #11405, opened 2021-11-17
raftstore: increase batch for raftlog-gc-worker

Little-Wallace:

Issue Number: close https://github.com/tikv/tikv/issues/11404

Batch more deleted keys together. As we all know ,the deleted key must be much small than the entry which has been put. So 256 is too small for every writebatch in raftlog-gc-worker.
And the gc tasks of most of regions are small. They are often smaller than 100. So I batch the keys of multiple regions and write them together.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/engine_rocks/src/raft_engine.rs
```diff
@@ -97,6 +97,38 @@ impl RaftEngineReadOnly for RocksEngine {
...
+        raft_wb: &mut RocksWriteBatch,
+    ) -> Result<usize> {
+        if from >= to {
+            return Ok(0);
+        }
+        if from == 0 {
+            let start_key = keys::raft_log_key(raft_group_id, 0);
+            let prefix = keys::raft_log_prefix(raft_group_id);
+            match self.seek(&start_key)? {
+                Some((k, _)) if k.starts_with(&prefix) => from = box_try!(keys::raft_log_index(&k)),
```
5kbpers, line 115: Make sure `from < to` after `seek`

5kbpers, line 115: Make sure from < to after seek

components/engine_rocks/src/raft_engine.rs
```diff
@@ -172,35 +204,27 @@ impl RaftEngine for RocksEngine {
...
-            let start_key = keys::raft_log_key(raft_group_id, 0);
-            let prefix = keys::raft_log_prefix(raft_group_id);
-            match self.seek(&start_key)? {
-                Some((k, _)) if k.starts_with(&prefix) => from = box_try!(keys::raft_log_index(&k)),
-                // No need to gc.
-                _ => return Ok(0),
-            }
-        }
-
+    fn batch_gc(&self, groups: Vec<(u64, u64, u64)>) -> Result<usize> {
```
5kbpers, line 207: Better to use a struct instead of a tuple.

5kbpers (approved):

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -108,8 +105,8 @@ impl<EK: KvEngine, ER: RaftEngine, R: CasualRouter<EK>> Runner<EK, ER, R> {
         });
         RAFT_LOG_GC_KV_SYNC_DURATION_HISTOGRAM.observe(start.saturating_elapsed_secs());
         let tasks = std::mem::take(&mut self.tasks);
+        let mut groups = Vec::with_capacity(tasks.len());
```
BusyJay, line 108: Better use a HashMap, so releated deletes are clustered together.

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -149,6 +135,20 @@ impl<EK: KvEngine, ER: RaftEngine, R: CasualRouter<EK>> Runner<EK, ER, R> {
                 }
             }
         }
+        let start = Instant::now();
+        match self.gc_raft_log(groups) {
```
BusyJay, line 139: This reorder the `Task::Purge` and `Task::Gc`, I'm not sure if it's OK. /cc @tabokie

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -120,21 +117,10 @@ impl<EK: KvEngine, ER: RaftEngine, R: CasualRouter<EK>> Runner<EK, ER, R> {
...
-                            RAFT_LOG_GC_FAILED.inc();
-                        }
-                        Ok(n) => {
-                            debug!("gc log entries"; "region_id" => region_id, "entry_count" => n);
-                            self.report_collected(n);
-                            RAFT_LOG_GC_DELETED_KEYS_HISTOGRAM.observe(n as f64);
-                        }
-                    }
-                    RAFT_LOG_GC_WRITE_DURATION_HISTOGRAM.observe(start.saturating_elapsed_secs());
+                    groups.push((region_id, start_idx, end_idx));
```
BusyJay, line 120: Should log abnormal gap and its region id.

components/raftstore/src/store/worker/raftlog_gc.rs
tabokie: It's fine, they are unrelated.

BusyJay (approved):

sticnarf:
Does this PR need to be picked to other release branches?

BusyJay:
It was picked to 5.1, but it should also be picked to 5.2 and 5.3.

Issue #11320, opened 2021-11-10
Raft should fetch log entries in async way


Connor1996:
## Development Task

We encounter a latency spike issue when some TiKV machines are down and restart after a while. 
<img width="920" alt="截屏2021-11-10 下午4 15 56" src="https://user-images.githubusercontent.com/13497871/141075539-559ecb86-3610-407b-8638-58c80e128905.png">

Process ready duration is increased greatly at the time of the down TiKVs back online.
<img width="921" alt="截屏2021-11-10 下午1 52 30" src="https://user-images.githubusercontent.com/13497871/141075705-ae107ca0-34e3-4bf5-b12e-63507c89716e.png">

After investigation, it turns out that fetching raft log entries cost a lot of time which blocks the raftstore for 100ms
<img width="926" alt="截屏2021-11-10 下午1 52 21" src="https://user-images.githubusercontent.com/13497871/141076789-e01f637e-04c2-4b69-9c79-8b6d7e6ced02.png">
<img width="1849" alt="截屏2021-11-10 下午1 52 42" src="https://user-images.githubusercontent.com/13497871/141076804-eb9acc9b-b616-41fb-9a10-873bd79108d1.png">
 
To alleviate this issue, we should fetch log entries in async way

Lily2025:
when open raft engine，while network fault recover，QPS dropped a lot
![pCr3ivwqbp](https://user-images.githubusercontent.com/84712107/152974633-86892040-0db2-488b-9551-987560890959.png)

Connor1996:
@Lily2025 Async fetch raft log has been merged, you can test it again to see if it's relieved.

PR #11409, opened 2021-11-18
raftstore: Fetch raft log in async manner

### LINKED ISSUE #10408: QPS drop a lot when a tikv is killed
v5.0.2

tidbcloud T3.standard

### Steps to reproduce
1. deploy a 2db 6kv(3 zone) cluster on tidbcloud staging environment
2. go-tpc prepare 1000 warehouses 
3. run `go-tpc tpcc --warehouses 1000 run -T 100`
4. kill and start one tikv
5. wait till the tps and duration is stable
6. kill two tikv in one zone

### What did you expect?
1. ~QPS drops not much(about 1/6 in ideal condition) when a tikv is killed.~
2. ~QPS drops not much(about 1/3 in ideal condition) when two tikvs are killed.~
3. The QPS drops less when a tikv is killed than when two tikvs are killed. (FAILED)
4. The QPS should recover in one or two minutes. (PASS)
5. After the killed tikv restart, there should be no jitter in QPS. (FAILED)

### What did happened?
1. QPS drops significantly when a tikv is killed 
image
image

2. When a tikv is killed, the QPS drops more than when two tikvs are killed.
image

Connor1996:

Issue Number: close https://github.com/tikv/tikv/issues/11320, close #10408

Problem Summary:

What's Changed:
This PR is divided into several parts:
- Fetch raft entries asynchronously in a standalone worker if entry cache is miss
    - make raft support fetch entries in async way https://github.com/tikv/raft-rs/pull/459
    - update kvproto https://github.com/pingcap/kvproto/pull/859
    - Introduce significant router https://github.com/tikv/tikv/pull/11750
    - Introduce raft log fetch worker https://github.com/tikv/tikv/pull/11900
- Reject transfer leader on start 
    - ~~if any region has log lag within 10min after the start https://github.com/tikv/tikv/pull/11859~~
    - report apply busy to PD https://github.com/tikv/tikv/pull/11921
- Handle leaders' apply tasks with higher priority than followers' apply tasks https://github.com/tikv/tikv/pull/11788
- Update grafana template
截屏2022-02-08 下午3 00 52

### Check List 

Tests 

- Unit test
- Integration test

### Test Result

master: a latency spike exceeding 1s because raftstore is blocked by fetching raft log
image

async fetch log: still have a latency spike because for the newly started tikv a lot of logs to be applied at that time
image

async fetch log + reject transfer leader when having log lag: no obvious latency spike even though the apply duration is high for newly started tikv 
image

### Release note 

```release-note
Fetch raft log in async manner
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -633,6 +633,29 @@ where
                 PeerMsg::UpdateRegionForUnsafeRecover(region) => {
                     self.on_update_region_for_unsafe_recover(region)
                 }
+                PeerMsg::RaftLogFetched {
```
Little-Wallace, line 636: Is it a typo ? `RaftEntryFetched`?

Connor1996: cause there is a term`raftlog_gc`, so I change `entry` to `log`😂

Connor1996: Oh, I forget to push the latest commit. What you have reviewed is stale, please check again

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```
zhangjinpeng1987, line 1: This filename is better to be 'raftlog_fetcher.rs'?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -755,7 +795,137 @@ where
...
+                high,
+                max_size: (max_size as usize),
+            })
+            .unwrap();
+        Err(raft::Error::Store(
+            raft::StorageError::LogTemporarilyUnavailable,
+        ))
+    }
+
+    pub fn entries(
```
tonyxuqqi, line 922: The current behavior of "entries()" if it was called twice with exactly same low, high, max_size, is that it will fetch the entry anyway.
But with this change, if it's called twice with same async_to parameter, it can call async_fetch which could return an LogTemporarilyUnavailable error. It changes the semantic of entries API.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -650,8 +681,13 @@ where
         low: u64,
         high: u64,
         max_size: impl Into<Option<u64>>,
+        async_to: Option<u64>,
```
tonyxuqqi, line 684: which code calls entries with non-empty async_to. Cannot find the caller using async_to.

Connor1996: in `raft-rs`, you'd better review tikv/raft-rs#459 first

components/raftstore/src/store/peer_storage.rs
```diff
@@ -755,7 +795,137 @@ where
...
+                                                ents.len(),
+                                                idx + (high - low) as usize,
+                                            ),
+                                        )
+                                        .collect());
+                                } else {
+                                    // the fetched entries doesn't cover [low, high]
+                                    self.raftlog_fetch_stats.fetch_invalid.update(|m| m + 1);
+                                }
+                            } else if *first > low {
```
tonyxuqqi, line 872: "if *first > low" is not necessary.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -755,7 +795,137 @@ where
...
+                                return Ok(ents);
+                            } else if *first < low {
+                                if low <= last {
+                                    // the fetched entries covers [low, ..], high may be
+                                    let idx = (low - first) as usize;
+                                    return Ok(ents
+                                        .drain(
+                                            idx..std::cmp::min(
+                                                ents.len(),
+                                                idx + (high - low) as usize,
```
tonyxuqqi, line 864: "idx + high - low" ==> high - first as it's more clear.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -757,7 +796,116 @@ where
...
+        context: GetEntriesContext,
+        buf: &mut Vec<Entry>,
+    ) -> raft::Result<usize> {
+        if let Some(res) = self.async_fetch_res.replace(None) {
+            if res.low == low {
+                let mut ents = res.ents?;
+                return Ok(match ents.first() {
+                    None => 0,
+                    Some(Entry { index: first, .. }) => {
+                        // no need to compare the index of last entry with high, it's okay to not return the full entries
```
BusyJay, line 827: The result can be used only when `first index == low && result max size == max_size`.

Connor1996: when the size of result < max_size, seems it doesn't make any problem.

BusyJay: For sending logs, it may be OK. But it will not be OK in other cases. When caller calls `entries` with size big enough, then it expects to get all the entries between `low` and `high`. Returning partial results breaks the assumption of the API.

Connor1996: if that it should also consider `high`. But it may cause async fetch results always invalid.

Connor1996: seems we can give the assumption only when `max_size == None`?

BusyJay: > if that it should also consider `high`. But it may cause async fetch results always be invalid.

Why? The key point is whether the old result is stopped at high limit or size limit. If it hits size limit, then as long as `low` and `size_limit` are the same, the result can be reused directly. If it stops at high limit, then it should either fetches more from cache or trigger a second fetch.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -757,7 +797,105 @@ where
         Ok(())
     }
 
-    pub fn entries(&self, low: u64, high: u64, max_size: u64) -> raft::Result<Vec<Entry>> {
+    pub fn set_async_fetch_res(&mut self, res: RaftlogFetchResult) {
+        let prev_res = self.async_fetch_res.replace(Some(res));
```
BusyJay, line 801: Technically, it should validate whether the logs still match the actual logs. For example, a leader fetches logs between [10, 50) asynchronously. And then it becomes follower, and logs are overwritten by new leader from 40. And then it becomes a leader again. If the returned value is reused, then it may produce incorrect logs sequence.

An easy way to solve the problem is store the term when trigger the task, and check whether term changes when the task is returned. Though this may be a waste to drop all logs if they are not overwritten at all.

Connor1996:
I notice one peer is regarded as pending only when the matched index is smaller than the truncated index. But the rejecting transfer leader mechanism needs to know the count of peers whose last index - matched index < max_log_lag. 
Here are two choices: 
1. introduce lagging peer term distinguishing from pending peer 
2. broaden the scope of pending peer to include peers whose last index - matched index < max_log_lag. Though it may cause stores reject to transfer leader on start when all available logs are caught up but some peers are still applying snapshot.
What do you think @BusyJay

components/raftstore/src/store/peer_storage.rs
Connor1996: seems we should add a term field in GetEntriesGetContext? peer storage has no access to the current term

Connor1996:
Now this PR is clean and ready to review, most of the code is test. PTAL @BusyJay @NingLin-P

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +799,191 @@ where
         Ok(())
     }
 
+    pub fn set_async_fetch_res(&mut self, res: Option<RaftlogFetchResult>) {
+        let in_fetching = match self.async_fetch_res.borrow().as_ref() {
+            r @ Some(RaftlogFetchResult::Fetched { .. }) => {
+                warn!(
```
BusyJay, line 805: I think it's a normal behavior. For example, it fallbacks to catch up logs by snapshot.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -784,24 +995,35 @@ where
...
-            self.engines.raft.fetch_entries_to(
-                region_id,
-                low,
-                high,
-                Some(max_size as usize),
-                &mut ents,
-            )?;
-            return Ok(ents);
+            return if context.can_async() {
+                self.async_fetch(region_id, low, high, max_size, context, &mut ents)?;
```
BusyJay, line 999: I'm afraid it may cause serious regression even under normal workload. It may make the slow node always lagging behind as it always takes more time to fetch logs.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +799,191 @@ where
...
+        let tried_cnt = if let Some(RaftlogFetchResult::Fetched {
+            ents: res_ents,
+            low: res_low,
+            max_size: res_max_size,
+            hit_size_limit: res_hit_size_limit,
+            tried_cnt: res_tried_cnt,
+            term: res_term,
+        }) = self.async_fetch_res.replace(None)
+        {
+            if res_low == low && res_term == self.hard_state().get_term() {
```
BusyJay, line 850: Though it may be OK to use term of storage, using term from raft directly seems a safer choice.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +799,191 @@ where
...
+        match *self.async_fetch_res.borrow() {
+            Some(RaftlogFetchResult::Fetching) => {
+                // already an async fetch in flight
+                return Err(raft::Error::Store(
+                    raft::StorageError::LogTemporarilyUnavailable,
+                ));
+            }
+            _ => {}
+        }
+        let tried_cnt = if let Some(RaftlogFetchResult::Fetched {
```
BusyJay, line 841: How about defining it as a struct? So matching every field can be avoided.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1183,8 +1183,11 @@ where
             SignificantMsg::RaftLogGcFlushed => {
                 self.on_raft_log_gc_flushed();
             }
-            SignificantMsg::RaftlogFetched { context, .. } => {
+            SignificantMsg::RaftlogFetched { context, res } => {
+                self.fsm.peer.mut_store().set_async_fetch_res(Some(res));
                 self.fsm.peer.raft_group.on_entries_fetched(context);
+                // clean the async fetch result immediately if not used to free memory
+                self.fsm.peer.mut_store().set_async_fetch_res(None);
```
NingLin-P, line 1190: `async_fetch_res` is set and clean immediately, is it possible that other places can still see `async_fetch_res = RaftlogFetchResult::Fetched`?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+            return Err(raft::Error::Store(
+                raft::StorageError::LogTemporarilyUnavailable,
+            ));
+        }
+
+        let tried_cnt = if let Some(RaftlogFetchState::Fetched(res)) =
+            self.async_fetch_results.borrow_mut().remove(&low)
+        {
+            assert_eq!(res.low, low);
+            let mut ents = res.ents?;
```
NingLin-P, line 882: If `ent` is an error then `raft-rs` will panic, is it expected?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+        max_size: u64,
+        context: GetEntriesContext,
+        buf: &mut Vec<Entry>,
+    ) -> raft::Result<usize> {
+        if let Some(RaftlogFetchState::Fetching) = self.async_fetch_results.borrow().get(&low) {
+            // already an async fetch in flight
+            return Err(raft::Error::Store(
+                raft::StorageError::LogTemporarilyUnavailable,
+            ));
+        }
```
NingLin-P, line 876: If there is more than one peer requesting log entries start from `low` then only one peer will get the logs by async fetching and other peers may have to wait until raft heartbeat to trigger getting logs for another peer.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+                "region_id" => self.region.get_id(),
+                "peer_id" => self.peer_id,
+                "first" => first,
+                "last" => last,
+                "low" => low,
+                "high" => high,
+                "max_size" => max_size,
+                "res_max_size" => res.max_size,
+            );
+            // low index or max size is changed, the result is not fit for the current range, so refetch again.
```
NingLin-P, line 952: How to clean up the `async_fetch_results` if the low index is changed?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+                // async fetch res covers [low, high)
+                ents.truncate((high - first) as usize);
+                assert_eq!(ents.last().map(|e| e.index).unwrap(), high - 1);
+                if max_size < res.max_size {
+                    limit_size(&mut ents, Some(max_size));
+                }
+                let count = ents.len();
+                buf.append(&mut ents);
+                return Ok(count);
+            } else if res.hit_size_limit && max_size <= res.max_size {
```
NingLin-P, line 900: Perhaps we don't need to drop the `fetch_res` even the `max_size` is set to a larger value.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+        };
+
+        // the first try: get [low, high) asynchronously
+        // the second try: get [low, persisted) asynchronously
+        // the third try:
+        //  - if term and low are matched: use result of [low, persisted) and get [persisted, high) synchronously
+        //  - else: get [low, high) synchronously
+        if tried_cnt == MAX_ASYNC_FETCH_TRY_CNT - 1 {
+            // the async fetch is retried several times, so use a larger range to avoid being invalid again
+            high = std::cmp::max(high, self.cache.persisted);
```
NingLin-P, line 966: Why set to a larger range can help the fetch result being valid?

components/raftstore/src/store/peer_storage.rs
Connor1996: yes, the error would be `ErrCompacted` or `ErrUnavailable`. They are possible returned errors that sync API may return.

components/raftstore/src/store/peer_storage.rs
Connor1996: That's what `set_async_fetch_results(low, None)` does

components/raftstore/src/store/peer_storage.rs
Connor1996: Yes, that's a defect though I think it's rare. Using `low+high` as the map index could be a solution, but when high is changed the result can't be reused which I think it's a more common case.

At the very previous implementation, peer id is used as the index. But to make the context transparent to the application, we just drop it.

components/raftstore/src/store/peer_storage.rs
Connor1996: but the `fetch_res` contains partial result under the larger `max_size`

components/raftstore/src/store/peer_storage.rs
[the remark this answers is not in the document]
Connor1996: why could `persisted` be wrong?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+                raft::StorageError::LogTemporarilyUnavailable,
+            ));
+        }
+
+        let tried_cnt = if let Some(RaftlogFetchState::Fetched(res)) =
+            self.async_fetch_results.borrow_mut().remove(&low)
+        {
+            assert_eq!(res.low, low);
+            let mut ents = res.ents?;
+            if ents.is_empty() {
```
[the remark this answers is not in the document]
Connor1996: I was thought `max_size` may make it empty. But seems it make sure there is at least one entry.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+                return Ok(0);
+            }
+            let first = ents.first().map(|e| e.index).unwrap();
+            assert_eq!(first, res.low);
+            let last = ents.last().map(|e| e.index).unwrap();
+
+            if last >= high - 1 {
```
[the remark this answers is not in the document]
Connor1996: I think it's more intuitive, as the index of the last entry is >= the largest index that high limit wants (high - 1)

components/raftstore/src/store/peer_storage.rs
Connor1996: being invalid is due to the cache is dropped. So fetching the entries as much as we can narrows down the possibility of being invalid again

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,188 @@ where
...
+        };
+
+        match res {
+            Some(res) => {
+                if let Some(RaftlogFetchState::Fetched(prev)) = self
+                    .async_fetch_results
+                    .borrow_mut()
+                    .insert(low, RaftlogFetchState::Fetched(res))
+                {
+                    warn!(
```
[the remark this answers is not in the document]
Connor1996: it's actually unexpected. Every time, we'll clean the fetched result if not used. So encountering a fetched result when setting a new fetched result is impossible.

components/raftstore/src/store/peer_storage.rs
NingLin-P: > At the very previous implementation, peer id is used as the index. But to make the context transparent to the application, we just drop it.

Sorry, I don't get it. The peer id index is used for `async_fetch_results` and `async_fetch_results` is the application's internal state.

components/raftstore/src/store/peer_storage.rs
NingLin-P: > but the fetch_res contains partial result under the larger max_size

Don't see any harm it brings? IOM, drop and re-fetching log entries is more harmful.

components/raftstore/src/store/peer_storage.rs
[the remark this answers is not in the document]
BusyJay: `persisted` is updated lazily. For example, if [3, 10] becomes [3, 8], and `persisted` won't be updated to 8 from 10 until 8 is persisted.

components/raftstore/src/store/peer_storage.rs
Connor1996: Jay said it's the promise for the API. https://github.com/tikv/tikv/pull/11409#discussion_r789414101

components/raftstore/src/store/peer_storage.rs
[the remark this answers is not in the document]
BusyJay: It's the same. `[low, high)` means `low <= last < high`, so `last + 1 <= high`. I suggest to use plus is because underflow is more likely to happen generally.

components/raftstore/src/store/peer_storage.rs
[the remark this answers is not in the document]
BusyJay: Yes, but does it matter?

components/raftstore/src/store/peer_storage.rs
Connor1996: the `to_peer_id` is not the application's internal state. We need to get the peer id to which the entries are sent from raft.

components/raftstore/src/store/peer_storage.rs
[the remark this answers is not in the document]
Connor1996: so shouldn't use `debug` to conceal the problem.

components/raftstore/src/store/peer_storage.rs
Connor1996: test in under 8c with sysbench-insert(400, 800 concurrency), sysbench-write-only(400, 800 concurrency) which is the case for testing async-io

Under normal workload, it can hardly trigger async fetch, See
截屏2022-02-07 下午12 33 10
截屏2022-02-07 下午12 33 26

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,182 @@ where
         Ok(())
     }
 
+    pub fn clean_async_fetch_res(&mut self, low: u64) {
+        self.async_fetch_results.borrow_mut().remove(&low);
+    }
+
+    pub fn set_async_fetch_res(&mut self, low: u64, res: Option<Box<RaftlogFetchResult>>) {
+        // if it's in fetching, don't clean the async fetch result.
```
BusyJay, line 829:
```suggestion
        // If it's in fetching, don't clean the async fetch result.
```

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,182 @@ where
...
+                raft::StorageError::LogTemporarilyUnavailable,
+            ));
+        }
+
+        let tried_cnt = if let Some(RaftlogFetchState::Fetched(res)) =
+            self.async_fetch_results.borrow_mut().remove(&low)
+        {
+            assert_eq!(res.low, low);
+            let mut ents = res.ents?;
+            assert!(!ents.is_empty());
```
BusyJay, line 879: This seems unnecessary given it's checked already at L880.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +821,182 @@ where
...
+                                ents.push(ent);
+                            }
+                        }
+                    }
+                }
+                let count = ents.len();
+                buf.append(&mut ents);
+                return Ok(count);
+            }
+            info!(
```
BusyJay, line 935: What if `last < high - 1` and `res.hit_size_limit` is false but `fetched_size >= max_size`?

Connor1996: I have considered it, but it needs to compute fetched size every time and in this case `max_size` must be changed  which is really rare

components/raftstore/src/store/peer_storage.rs
NingLin-P: How about implementing the `Hash` trait for `GetEntriesContext` (namely hash the `to_peer_id`) and using the hash value (an `u64`) for indexing.

Connor1996: Can't. In other context, there is no peer id, e.g. `GenReady`

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +825,180 @@ where
         Ok(())
     }
 
+    pub fn clean_async_fetch_res(&mut self, low: u64) {
+        self.async_fetch_results.borrow_mut().remove(&low);
+    }
+
+    pub fn set_async_fetch_res(&mut self, low: u64, res: Option<Box<RaftlogFetchResult>>) {
```
BusyJay, line 832: How about adding comment to explain what `None` means?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -767,12 +825,180 @@ where
...
+                    limit_size(&mut ents, Some(max_size));
+                };
+                let count = ents.len();
+                buf.append(&mut ents);
+                return Ok(count);
+            } else if last + RAFT_LOG_MULTI_GET_CNT > high - 1
+                && res.tried_cnt + 1 == MAX_ASYNC_FETCH_TRY_CNT
+            {
+                let mut fetched_size = ents.iter().fold(0, |acc, e| acc + e.compute_size() as u64);
+                if max_size < fetched_size {
```
BusyJay, line 912: It should be OK to also check `==`.

components/raftstore/src/store/peer_storage.rs
BusyJay: It seems async fetches are always triggered when the workload is finished.

tests/integrations/raftstore/test_compact_log.rs
```diff
@@ -84,6 +84,76 @@ fn check_compacted(
     true
 }
 
+#[test]
+fn test_node_cache_compact_with_one_node_down() {
```
BusyJay, line 88: Can you explain the purpose of the case?

tests/integrations/raftstore/test_compact_log.rs
```diff
@@ -84,6 +84,76 @@ fn check_compacted(
...
+        let mut state: RaftApplyState = get_raft_msg_or_default(engines, &keys::apply_state_key(1));
+        let after_state = state.take_truncated_state();
+
+        let before_state = &before_states[&id];
+        let idx = after_state.get_index();
+        assert_eq!(idx, before_state.get_index());
+    }
+
+    // logs should be replicated to node 1 successfully.
+    for i in 1..60 {
```
BusyJay, line 138: What if it's not replicated to node 1 yet?

components/raftstore/src/store/peer_storage.rs
Connor1996: because there is a restart

metrics/grafana/tikv_details.json
```diff
@@ -14,7 +14,7 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "7.5.7"
+      "version": "7.5.11"
```
BusyJay, line 17: How about only including necessary changes?

tests/integrations/raftstore/test_compact_log.rs
```diff
@@ -84,6 +84,80 @@ fn check_compacted(
...
+        let v = k.clone();
+        cluster.must_put(&k, &v);
+    }
+
+    // wait log gc.
+    sleep_ms(500);
+
+    cluster.run_node(1).unwrap();
+
+    // limit has not reached, should not gc.
```
BusyJay, line 129: How to know async fetch is triggered?

BusyJay (commented):
Rest LGTM

metrics/grafana/tikv_details.json
Connor1996: it should be changed anyway, as the grafana has been updated

BusyJay: Yes, but other changes like `gridPos` are not necessary.

tests/integrations/raftstore/test_compact_log.rs
Connor1996: now, check the metrics to make sure async-fetch is performed.

metrics/grafana/tikv_details.json
BusyJay: Does v5.0 use the same version?

metrics/grafana/tikv_details.json
```diff
@@ -39133,6 +39236,6 @@
   },
   "timezone": "browser",
   "title": "Test-Cluster-TiKV-Details",
-  "uid": "RDVQiEzZz",
+  "uid": "RDVQiEzZz1",
```
BusyJay, line 39239: Why change this?

metrics/grafana/tikv_details.json
Connor1996: clean the unnecessary changes now

tests/failpoints/cases/test_async_fetch.rs
```diff
@@ -0,0 +1,149 @@
...
+        assert_eq!(v2, Some(v));
+
+        if i > 100 && check_compacted(&cluster.engines, &before_states, 1) {
+            return;
+        }
+    }
+    panic!("cluster is not compacted after inserting 200 entries.");
+}
+
+fn check_compacted(
```
BusyJay, line 100: Can it be extracted to test_raftstore?

BusyJay (approved):

NingLin-P (approved):

PR #11418, opened 2021-11-19
raftstore: support synchronous write

gengliqi:

Problem Summary:
After https://github.com/tikv/tikv/pull/10289, raftstore supports asynchronous write.
However, the performance may be degraded because the effect of batching command may be worse than before especially when the CPU is very limited. After merging this PR, synchronous write can be used when `store-io-pool-size` is 0.

What's Changed:
Mentioned above

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No.

### Release note 

```release-note
None.
```


etc/config-template.toml
```diff
@@ -419,6 +419,7 @@
 # store-pool-size = 2
 
 ## Use how many threads to handle raft io tasks
+## If it is 0, it means io tasks are handled in store threads.
 # store-io-pool-size = 1
```
zhangjinpeng1987, line 423: Please keep the default value 0.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -584,13 +592,17 @@ where
         let send_time = duration_to_sec(now2.saturating_duration_since(now));
         STORE_WRITE_SEND_DURATION_HISTOGRAM.observe(send_time);
 
-        for (region_id, (peer_id, ready_number)) in &self.batch.readies {
-            self.notifier
-                .notify_persisted(*region_id, *peer_id, *ready_number);
+        let mut callback_time = 0f64;
```
hicqu, line 595: Seems can be moved into `if notify` block

components/raftstore/src/store/fsm/store.rs
```diff
@@ -787,25 +788,49 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
     }
 
     fn end(&mut self, peers: &mut [Box<PeerFsm<EK, ER>>]) {
-        for peer in peers {
+        for peer in peers.iter_mut() {
```
hicqu, line 791: Why `iter_mut`?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -787,25 +788,49 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
+                    >= self.poll_ctx.cfg.raft_msg_flush_interval.0
+            {
+                self.last_flush_msg_time = now;
+                self.poll_ctx.trans.flush();
+            }
+
+            if now.saturating_duration_since(self.last_flush_time) >= Duration::from_millis(1) {
+                self.last_flush_time = now;
+                self.need_flush_events = false;
+                self.flush_events();
```
hicqu, line 814: Seems `flush_events` is a new logic? Can you comment it?

components/raftstore/src/store/async_io/write.rs
gengliqi: The `slow_log` below needs this `callback_time`.

components/raftstore/src/store/fsm/store.rs
gengliqi: Because `peers` will be moved if using `into_iter` implicitly.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -827,48 +852,57 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
-            self.flush_events();
-        } else {
-            self.need_flush_events = true;
-        }
-
         if !self.poll_ctx.pending_latency_inspect.is_empty() {
+            let latency_inspect = std::mem::take(&mut self.poll_ctx.pending_latency_inspect);
             for inspector in &mut self.poll_ctx.pending_latency_inspect {
                 inspector.record_store_process(dur);
             }
```
NingLin-P, line 870: `self.poll_ctx.pending_latency_inspect` will be empty after `mem::take`

gengliqi: Fixed

components/raftstore/src/store/fsm/store.rs
```diff
@@ -771,41 +776,79 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
     }
 
-    fn end(&mut self, peers: &mut [Box<PeerFsm<EK, ER>>]) {
-        for peer in peers {
+    fn light_end(
+        &mut self,
+        peers: &mut [Option<impl DerefMut<Target = PeerFsm<EK, ER>>>],
+        _to_skip_end: &mut Vec<usize>,
+    ) {
+        for peer in peers.iter_mut().flatten() {
```
BusyJay, line 807: Should flush raft client if it's written synchronously.

components/test_raftstore/src/common-test.toml
```diff
@@ -67,6 +67,7 @@ merge-check-tick-interval = "100ms"
 pd-heartbeat-tick-interval = "20ms"
 dev-assert = true
 hibernate-regions = true
+store-io-pool-size = 0
```
BusyJay, line 70: Why 0?

zhangjinpeng1987: turn off by default

BusyJay (approved):
This is not the original synchronous write implementation. Instead, it's new feature that implementing synchronous write based on asynchronous write. Compared to asynchronous write, its biggest difference is reducing write threads.

Time is too limited to review through, I can see asynchronous write is the same as before. For synchronous write, LGTM if CI says OK.

NingLin-P (approved):

Issue #11295, opened 2021-11-09
panic when preparing tpcc/ch benchmark dataset


dragonly:
## Bug Report

when doing
```
go-tpc tpcc cleanup -H 127.0.0.1 -P 4001 --warehouses 1 -D ch
```
and
```
go-tpc ch cleanup -H 127.0.0.1 -P 4001 --warehouses 1 -D ch
```

sometimes tikv will panic

```
tikv quit: exit status 1
[2021/11/09 11:43:14.772 +08:00] [INFO] [raft.rs:1200] ["became leader at term 6"] [term=6] [raft_id=80] [region_id=79]
[2021/11/09 11:43:14.772 +08:00] [INFO] [peer.rs:3778] ["require updating max ts"] [initial_status=25769803854] [region_id=79]
[2021/11/09 11:43:14.773 +08:00] [INFO] [pd.rs:1246] ["succeed to update max timestamp"] [region_id=79]
[2021/11/09 11:43:14.773 +08:00] [INFO] [endpoint.rs:382] ["deregister observe region"] [observe_id=ObserveID(73)] [region_id=2] [store_id=Some(1)]
[2021/11/09 11:43:14.773 +08:00] [INFO] [endpoint.rs:309] ["register observe region"] [region="id: 2 start_key: 7480000000000000FF5100000000000000F8 region_epoch { conf_ver: 1 version: 39 } peers { id: 3 store_id: 1 }"]
[2021/11/09 11:43:14.773 +08:00] [INFO] [endpoint.rs:233] ["Resolver initialized"] [pending_data_index=0] [snapshot_index=662] [observe_id=ObserveID(75)] [region=2]
[2021/11/09 11:43:14.773 +08:00] [INFO] [endpoint.rs:309] ["register observe region"] [region="id: 79 start_key: 7480000000000000FF4F00000000000000F8 end_key: 7480000000000000FF5100000000000000F8 region_epoch { conf_ver: 1 version: 39 } peers { id: 80 store_id: 1 }"]
[2021/11/09 11:43:14.773 +08:00] [INFO] [endpoint.rs:233] ["Resolver initialized"] [pending_data_index=0] [snapshot_index=6] [observe_id=ObserveID(76)] [region=79]
[...]

Issue #11192, opened 2021-10-30
Panic in analyze sampling


gengliqi:
### What version of TiKV are you using?
https://github.com/tikv/tikv/tree/d0c129dde8d3f41bbac26ab786419bb4b5e24878 

### What operating system and CPU are you using?
doesn't matter

### Steps to reproduce
I run tipocket `pipeline` test. I find several TiKV panic and stuck in an infinite crash loop.

### What did you expect?
Not panic.

### What did happened?

```
[...]

PR #11425, opened 2021-11-19
copr:fix panic when tidb change the analyze behaviors using sampling

Jayice-zjw:

ref https://github.com/tikv/tikv/issues/11192 https://github.com/tikv/tikv/issues/11295

Optimized code logic https://github.com/tikv/tikv/blob/e81fe9a26f88fcbe2196b2bfa40e804e1a7d1686/src/coprocessor/statistics/analyze.rs#L718

### Release note 

```release-note
```


src/coprocessor/statistics/analyze.rs
```diff
@@ -715,10 +715,18 @@ impl RowSampleCollector for ReservoirRowSampleCollector {
         let cur_rng = self.base.rng.gen_range(0, i64::MAX);
         if self.samples.len() < self.max_sample_size {
             need_push = true;
-        } else if self.samples.peek().unwrap().0.0 < cur_rng {
-            need_push = true;
-            let (_, evicted) = self.samples.pop().unwrap().0;
-            self.base.memory_usage -= evicted.iter().map(|x| x.capacity()).sum::<usize>();
+        } else {
```
bufferflies, line 718: I think this can not solve the upgrade consistency problem. it will run sample branch if sample size=1 in master

Jayice-zjw: but according to https://github.com/tikv/tikv/issues/11295, samples will be None in some cases

winoros (commented):
Oh, can we just throw an error?

winoros (commented):
Or we change the function signature `new_collector` to throw an error when the values passed in is wrong.

Jayice-zjw:
> Or we change the function signature `new_collector` to throw an error when the values passed in is wrong.

alright~  but how do we define "a wrong value"?，😂

BusyJay:
@winoros Can you make further explanation?

PR #11447, opened 2021-11-23
storage: fix reverse scan check memory locks

### LINKED ISSUE #11440: Reverse scan can't detect memory locks
## Bug Report

`Storage::scan()` always uses range [start_key, end_key) to check memory locks

https://github.com/tikv/tikv/blob/0b3655275712b0dd04de5fd83cddc408034f4b8b/src/storage/mod.rs#L843

but reverse scan's start_key is greater than end_key
https://github.com/pingcap/kvproto/blob/03abd340988f539630450c78ead7b13dcf5e056d/proto/kvrpcpb.proto#L56-L58. It results in that reverse scan can't detect memory locks and may read old data.

youjiali1995:

Issue Number:close #11440 

Problem Summary:

Reverse scan's start_key is greater than end_key, but storage always uses [start_key, end_key) to check memory locks.

What's Changed:

Swap start_key and end_key if is a reverse scan.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
Fix the issue that reverse scan can't detect memory locks and may read stale data.
```

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -785,8 +785,8 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
     }
 
     /// Scan keys in [`start_key`, `end_key`) up to `limit` keys from the snapshot.
-    ///
-    /// If `end_key` is `None`, it means the upper bound is unbounded.
+    /// If `reverse_scan` is true, it scans [`end_key`, `start_key`) in descending order.
+    /// If `end_key` is `None`, it means the upper bound or the lower bound if reverse scan is unbounded.
     ///
     /// Only writes committed before `start_ts` are visible.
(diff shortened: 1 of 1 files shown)


sticnarf (approved):
LGTM

cfzjywxk (approved):
Do we need to add some test coverage enhancement task [here](https://github.com/PingCAP-QE/transaction-qa)?

youjiali1995:
I can't open it https://github.com/PingCAP-QE/transaction-qa. The reason why we didn't find this bug in tests is TiDB doesn't use scan except in backfilling. I think it's not valuable to add integration tests for it and unit tests in TiKV are enough. @cfzjywxk

PR #11457, opened 2021-11-23
raftstore: destroy uninitialized peer can make it possible to recreate old peer

### LINKED ISSUE #10533: destroy uninitialized peer can make it possible to recreate old peer
v5.0.3

Linux

### Steps to reproduce
Supposing there are 4 stores 1, 2, 3, 4. And we schedule a region to add peer a, b, c on store 1, 2, 3 respectively. Since then a is isolated, b is leader. b decides to remove c and add d on store 4. It can succeed as quorum remains connected during the whole time. After that, 2 adds a new learner e on store 3, but remove it immediately before sending any snapshot to it. So e is uninitialized. After 2 hours, e will ask PD and find itself deleted, so will mark itself as tombstone with an **empty region**. If a is recovered and send vote requests to c, the messages can pass the peer id check and epoch check as region is uninitialized. So c will be recreated as an uninitialized peer and vote for a. Then a will become leader. There will be two leaders a and b for the region.

### What did you expect?
Only one leader at a term and the cluster work correctly.

### What did happened?
Correctness is broken.

PatrickNicholas:
Signed-off-by: p4tr1ck <patricknicholas@foxmail.com>

Issue Number: close #10533  

Problem Summary:

For an uninitialized peer, both epoch and peer list are empty, so a message target to a staled peer which override by an uninitialized peer will be create again. 

What's Changed:

This PR will save the current peer into peer list if this region is uninitialized.

### Related changes

### Check List 

Tests 

- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Fix the problem that destroying an uninitialized replica may cause a stalled replica be created again.
```

### DIFF
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -6,13 +6,15 @@ use std::sync::{mpsc, Arc, Mutex, RwLock};
 use std::time::Duration;
 use std::{result, thread};
 
+use crossbeam::channel::TrySendError;
 use futures::executor::block_on;
 use kvproto::errorpb::Error as PbError;
 use kvproto::metapb::{self, PeerRole, RegionEpoch, StoreLabel};
 use kvproto::pdpb;
 use kvproto::raft_cmdpb::*;
 use kvproto::raft_serverpb::{
-    self, RaftApplyState, RaftLocalState, RaftMessage, RaftTruncatedState, RegionLocalState,
+    self, PeerState, RaftApplyState, RaftLocalState, RaftMessage, RaftTruncatedState,
+    RegionLocalState,
 };
 use raft::eraftpb::ConfChangeType;
 use tempfile::TempDir;
@@ -1569,6 +1571,38 @@ impl<T: Simulator> Cluster<T> {
             try_cnt += 1;
             sleep_ms(20);
         }
+    }
+
+    pub fn destroy_peer(
+        &mut self,
+        region_id: u64,
+        node_id: u64,
+        peer: metapb::Peer,
+    ) -> std::result::Result<(), TrySendError<RaftMessage>> {
+        let router = self.sim.rl().get_router(node_id).unwrap();
+
+        let mut message = RaftMessage::default();
+        message.set_region_id(region_id);
+        message.set_from_peer(peer.clone());
+        message.set_to_peer(peer);
+        message.set_region_epoch(self.get_region_epoch(region_id));
+        message.set_is_tombstone(true);
+        router.send_raft_message(message)
+    }
+
+    pub fn must_destroy_peer(&mut self, region_id: u64, node_id: u64, peer: metapb::Peer) {
+        for _ in 0..250 {
+            self.destroy_peer(region_id, node_id, peer.clone()).unwrap();
+            if self.region_local_state(region_id, node_id).get_state() == PeerState::Tombstone {
+                return;
+            }
+            sleep_ms(20);
+        }
+
+        panic!(
+            "destroy peer timeout: region id {}, node id {}, peer {:?}",
+            region_id, node_id, peer
+        );
     }
 }
 
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -1573,7 +1573,7 @@ impl<T: Simulator> Cluster<T> {
         }
     }
 
-    pub fn destroy_peer(
+    pub fn gc_peer(
         &mut self,
         region_id: u64,
         node_id: u64,
@@ -1590,9 +1590,9 @@ impl<T: Simulator> Cluster<T> {
         router.send_raft_message(message)
     }
 
-    pub fn must_destroy_peer(&mut self, region_id: u64, node_id: u64, peer: metapb::Peer) {
+    pub fn must_gc_peer(&mut self, region_id: u64, node_id: u64, peer: metapb::Peer) {
         for _ in 0..250 {
-            self.destroy_peer(region_id, node_id, peer.clone()).unwrap();
+            self.gc_peer(region_id, node_id, peer.clone()).unwrap();
             if self.region_local_state(region_id, node_id).get_state() == PeerState::Tombstone {
                 return;
             }
@@ -1600,7 +1600,7 @@ impl<T: Simulator> Cluster<T> {
         }
 
         panic!(
-            "destroy peer timeout: region id {}, node id {}, peer {:?}",
+            "gc peer timeout: region id {}, node id {}, peer {:?}",
             region_id, node_id, peer
         );
     }
--- a/tests/failpoints/cases/test_stale_peer.rs
+++ b/tests/failpoints/cases/test_stale_peer.rs
@@ -213,6 +213,7 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
     must_get_none(&cluster.get_engine(3), b"k1");
 }
 
+/// Test if destroy a uninitialized peer through tombstone msg would allow a staled peer be created again.
 #[test]
 fn test_destroy_uninitialized_peer_when_there_exists_old_peer() {
     // 4 stores cluster.
@@ -273,7 +274,7 @@ fn test_destroy_uninitialized_peer_when_there_exists_old_peer() {
     // Remove and destroy the uninitialized 5
     let peer_5 = new_learner_peer(3, 5);
     pd_client.must_remove_peer(r1, peer_5.clone());
-    cluster.must_destroy_peer(r1, 3, peer_5);
+    cluster.must_gc_peer(r1, 3, peer_5);
 
     let region = block_on(pd_client.get_region_by_id(r1)).unwrap();
     must_region_cleared(&cluster.get_all_engines(3), &region.unwrap());
--- a/tests/failpoints/cases/test_stale_peer.rs
+++ b/tests/failpoints/cases/test_stale_peer.rs
@@ -255,17 +255,12 @@ fn test_destroy_uninitialized_peer_when_there_exists_old_peer() {
 
     // Remove 3 and add 4
     pd_client.must_add_peer(r1, new_learner_peer(4, 4));
-    // must_get_equal(&cluster.get_engine(4), b"k10", b"v1");
-
     pd_client.must_add_peer(r1, new_peer(4, 4));
     pd_client.must_remove_peer(r1, new_peer(3, 3));
 
     for i in 11..20 {
         cluster.must_put(format!("k{}", i).as_bytes(), b"v1");
     }
-
-    // Add learner on store 5
-    pd_client.must_add_peer(r1, new_learner_peer(3, 5));
 
     // Ensure 5 drops all snapshot
     let (notify_tx, _notify_rx) = mpsc::channel();
@@ -274,6 +269,9 @@ fn test_destroy_uninitialized_peer_when_there_exists_old_peer() {
         .wl()
         .add_recv_filter(3, Box::new(DropSnapshotFilter::new(notify_tx)));
 
+    // Add learner 5 on store 3
+    pd_client.must_add_peer(r1, new_learner_peer(3, 5));
+
     for i in 21..30 {
         cluster.must_put(format!("k{}", i).as_bytes(), b"v1");
     }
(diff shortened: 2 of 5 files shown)


PatrickNicholas:
@BusyJay @NingLin-P 
PTAL, feel free to let me know if you have any other questions.

BusyJay:
Can you explain how this is fixed in the PR description?

components/keys/src/lib.rs
```diff
@@ -56,6 +60,9 @@ pub const SNAPSHOT_RAFT_STATE_SUFFIX: u8 = 0x04;
 // For region meta
 pub const REGION_STATE_SUFFIX: u8 = 0x01;
 
+// For peer meta
+pub const PEER_STATE_SUFFIX: u8 = 0x01;
```
BusyJay, line 64: I don't think it's a good idea to add a new state just for one corner case, especially when it can be fixed by no extra states.

PatrickNicholas: I think this is not a corner case. On the contrary, it is the correct way to give each removed peer a separate key to record its status.

PatrickNicholas: >  especially when it can be fixed by no extra states.

@BusyJay Could you roughly describe the solution?

BusyJay: It's quoted in the original issue that "adding the current peer into peer list when writing tombstone".

PatrickNicholas: > It's quoted in the original issue that "adding the current peer into peer list when writing tombstone".

This is to introduce a new concept just to solve a corner case, and only part of the problem can be solved. You also need to introduce some other mechanisms to resolve other case. For example, merge a region need the region info in the RegionLocalState to continue ticking for overlapped, so I need save both tombstone state and region info.

The most intuitive and easy to understand way is to record state and region info on the state of the peer, not on the state of the region. In this way, any related problems can be solved once and for all.

BusyJay: > ...introduce a new concept just to solve a corner case...

What's the new concept? Using epoch and peer list to check for existence is existing flow.

> For example, merge a region need the region info in the RegionLocalState to continue ticking for overlapped, so I need save both tombstone state and region info.

Can you explain more? I don't see how adding the current peer into peer list can break this.

PatrickNicholas: > Using epoch and peer list to check for existence is existing flow.
Maybe I misunderstood what you said "add current peer into peer list", do you mean add the current peer into the region in RegionLocalState if it is uninitialized?

BusyJay: Yes. More precisely, if a peer is about to be destroyed, and it's uninitialized, then add the peer to the region peer list without changing epoch. So when a stale heartbeat arrives, it will find the local state in tombstone with epoch set to 0. So it will check the peer list to find if there is a peer with larger or equal ID, then hit the added peer and return.

PatrickNicholas: Make sense, I will try to modify this PR.

PatrickNicholas:
@BusyJay I have modified the code according to the comments, PTAL.

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+    // must_get_equal(&cluster.get_engine(4), b"k10", b"v1");
+
+    pd_client.must_add_peer(r1, new_peer(4, 4));
+    pd_client.must_remove_peer(r1, new_peer(3, 3));
+
+    for i in 11..20 {
+        cluster.must_put(format!("k{}", i).as_bytes(), b"v1");
+    }
+
+    // Add learner on store 5
```
BusyJay, line 267: You mean store 3?

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+    pd_client.must_add_peer(r1, new_learner_peer(3, 5));
+
+    // Ensure 5 drops all snapshot
+    let (notify_tx, _notify_rx) = mpsc::channel();
+    cluster
```
BusyJay, line 272: Should probably add filter before doing the conf change.

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+    }
+
+    // Remove 3 and add 4
+    pd_client.must_add_peer(r1, new_learner_peer(4, 4));
+    // must_get_equal(&cluster.get_engine(4), b"k10", b"v1");
+
+    pd_client.must_add_peer(r1, new_peer(4, 4));
+    pd_client.must_remove_peer(r1, new_peer(3, 3));
+
+    for i in 11..20 {
```
BusyJay, line 263: What's the purpose?

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+        .add_recv_filter(3, Box::new(DropSnapshotFilter::new(notify_tx)));
+
+    for i in 21..30 {
+        cluster.must_put(format!("k{}", i).as_bytes(), b"v1");
+    }
+
+    // Remove and destroy the uninitialized 5
+    let peer_5 = new_learner_peer(3, 5);
+    pd_client.must_remove_peer(r1, peer_5.clone());
+    cluster.must_destroy_peer(r1, 3, peer_5.clone());
```
BusyJay, line 284: You can use `configure_for_merge` to make stale peer check quicker without destroying the peer explicitly.

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+    cluster
+        .sim
+        .wl()
+        .add_recv_filter(2, Box::new(PartitionFilter::new(vec![1, 3])));
+    cluster
+        .sim
+        .wl()
+        .add_recv_filter(4, Box::new(PartitionFilter::new(vec![1, 3])));
+
+    cluster.sim.wl().clear_send_filters(1);
```
BusyJay, line 299: Why not just 
```
cluster.clear_send_filters();
cluster.partition(vec![1, 3], vec![2, 4]);
```
?

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+
+    // Add learner on store 5
+    pd_client.must_add_peer(r1, new_learner_peer(3, 5));
+
+    // Ensure 5 drops all snapshot
+    let (notify_tx, _notify_rx) = mpsc::channel();
+    cluster
+        .sim
+        .wl()
+        .add_recv_filter(3, Box::new(DropSnapshotFilter::new(notify_tx)));
```
BusyJay, line 275: Or you can use send filter like `RegionPacketFilter`, so in the end explicit clean can be avoided.

tests/failpoints/cases/test_stale_peer.rs
```diff
@@ -211,3 +212,99 @@ fn test_stale_peer_destroy_when_apply_snapshot() {
...
+        .sim
+        .wl()
+        .add_recv_filter(4, Box::new(PartitionFilter::new(vec![1, 3])));
+
+    cluster.sim.wl().clear_send_filters(1);
+    cluster.sim.wl().clear_send_filters(3);
+    cluster.sim.wl().clear_recv_filters(3);
+
+    sleep_ms(
+        6u64 * cluster.cfg.raft_store.raft_election_timeout_ticks as u64
```
BusyJay, line 304: What is `6`?

tests/failpoints/cases/test_stale_peer.rs
PatrickNicholas: This is not an atomic operation, maybe some messages are send from 1 to 2 before  `cluster.partition`, after `cluster.clear_send_filters`.

BusyJay: Yes, but it's simple and too fast to meet race.

tests/failpoints/cases/test_stale_peer.rs
PatrickNicholas: Randomly selected, just ensure enough intervals so that peer 1 would trigger a election.

tests/failpoints/cases/test_stale_peer.rs
BusyJay: Then a single write should be sufficient.

tests/failpoints/cases/test_stale_peer.rs
BusyJay: Shouldn't `3` be enough? The default election timeout is 250ms in test, 6 times means 1.5 seconds.

tests/integrations/raftstore/test_conf_change.rs
```diff
@@ -79,8 +79,8 @@ fn test_simple_conf_change<T: Simulator>(cluster: &mut Cluster<T>) {
     // peer 3 must exist
     must_get_equal(&engine_3, b"k3", b"v3");
 
-    // add peer 2 then remove it again.
-    pd_client.must_add_peer(r1, new_peer(2, 2));
+    // add peer 22 then remove it again.
```
BusyJay, line 82: Why change this?

tests/failpoints/cases/test_stale_peer.rs
PatrickNicholas: I think it would be easier to understand if you explicitly destroying the peer?

tests/integrations/raftstore/test_conf_change.rs
PatrickNicholas: Please ignore, this is cherry-picked from the previous changes, I have deleted it.

tests/failpoints/cases/test_stale_peer.rs
PatrickNicholas: Ok, I will change to 3. Would it be better to add a helper to unify this logic?

BusyJay: Yes. Perhaps you can add a send filter to capture `VoteResponse` from store 3, which should collect nothing and always timeout.

tests/failpoints/cases/test_stale_peer.rs
BusyJay: Yes, I don't have strong opinion on this. Stale peer check is the real code path that can trigger problem.

BusyJay (commented):
The case can pass even without the fix.

components/test_raftstore/src/cluster.rs
```diff
@@ -1570,6 +1572,38 @@ impl<T: Simulator> Cluster<T> {
             sleep_ms(20);
         }
     }
+
+    pub fn destroy_peer(
```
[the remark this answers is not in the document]
PatrickNicholas: I have renamed this function to gc_peer, PTAL.

BusyJay:
> The case can pass even without the fix.

Any updates? You can verify it by running
```
env EXTRA_CARGO_ARGS=test_destroy_uninitialized_peer_when_there_exists_old_peer make test
```

PatrickNicholas:
> > The case can pass even without the fix.
> 
> Any updates? You can verify it by running
> 
> ```
> env EXTRA_CARGO_ARGS=test_destroy_uninitialized_peer_when_there_exists_old_peer make test
> ```

Do you mean `test_destroy_uninitialized_peer_when_there_exists_old_peer` would pass even without saving current peer into RegionLocalState? I have verified by:

```
git checkout c73c5512d1bcadec784fb6f82f886db8901e6b7a
cargo test -p tests --test failpoints -- test_destroy_uninitialized_peer_when_there_exists_old_peer
```

The result is :

```
failures:
    cases::test_stale_peer::test_destroy_uninitialized_peer_when_there_exists_old_peer

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 167 filtered out; finished in 2.78s

error: test failed, to rerun pass '-p tests --test failpoints'
```

Now it is the latest change, please give a some sugession if you have any question.

BusyJay:
I test it with:
```
git checkout master -- components/raftstore
cargo test -p tests --test failpoints -- test_destroy_uninitialized_peer_when_there_exists_old_peer
```
.

And no failure is reported.

I guess the latest change break the test case.

PatrickNicholas:
> I guess the latest change break the test case.

Yes, I have fixed.

BusyJay (approved):

BusyJay:
Remember to add release note for the fix.

NingLin-P (approved):

PR #11465, opened 2021-11-24
support to get approximate keys and size for a given range from UserCollectedProperties

### LINKED ISSUE #11464: Support to get approximate keys and size from user collected properties
## Development Task

Currently approximate keys and size are only available for `KvEngine`, not for every tables. It's necessary to add interfaces to access them with `UserCollectedProperties`.

hicqu:

Issue Number: close #11464.
Extracted from #11385 .

### Check List 

Tests 

- No code

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_rocks/src/table_properties.rs
+++ b/components/engine_rocks/src/table_properties.rs
@@ -1,6 +1,6 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-use crate::{util, RocksEngine};
+use crate::{util, RangeProperties, RocksEngine};
 use engine_traits::{Error, Range, Result};
 
 #[repr(transparent)]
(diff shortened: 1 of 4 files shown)


Connor1996 (approved):
LGTM

components/engine_rocks/src/table_properties.rs
```diff
@@ -9,6 +9,12 @@ impl engine_traits::UserCollectedProperties for UserCollectedProperties {
     fn get(&self, index: &[u8]) -> Option<&[u8]> {
         self.0.get(index)
     }
+
+    fn approximate_size_and_keys(&self, start: &[u8], end: &[u8]) -> Option<(usize, usize)> {
+        let rp = RangeProperties::decode(&self.0).ok()?;
```
tabokie, line 14: Where will this new method be called? This seems expensive.

tabokie (approved):

components/engine_rocks/src/table_properties.rs
hicqu: Here: https://github.com/tikv/tikv/pull/11385/files#r755931248.

Issue #11452, opened 2021-11-23
Tracking Issue: In-memory Pessimistic Locks


sticnarf:
RFC: https://github.com/tikv/rfcs/pull/77

## Development Task

- [x] https://github.com/tikv/tikv/pull/11459
- [x] https://github.com/tikv/tikv/pull/11471
- [x] https://github.com/tikv/tikv/pull/11481
- [x] https://github.com/tikv/tikv/pull/11529
- [x] https://github.com/tikv/tikv/pull/11591
- [x] https://github.com/tikv/tikv/pull/11577
- [x] https://github.com/tikv/tikv/pull/11841
- [x] https://github.com/tikv/tikv/pull/11883
- [x] https://github.com/tikv/tikv/pull/11655
- [x] https://github.com/tikv/tikv/pull/11985
- [x] https://github.com/tikv/tikv/pull/11758
- [x] https://github.com/tikv/tikv/pull/12089
- [x] https://github.com/tikv/tikv/pull/12078
- [ ] https://github.com/tikv/tikv/pull/12153

PR #11471, opened 2021-11-25
*: replace snapshot `CbContext` with `SnapshotExt`

sticnarf:
Based on https://github.com/tikv/tikv/pull/11459

Issue Number: #11452

This PR is a refactoring about the snapshot interfaces.

It was messy as the raftstore related information is returned in two ways. Some are returned through `CbContext`, such as the term and the `txn_extra_op`. And some are through the `Snapshot` interface directly, like the `data_version` and the sync status.

This PR uniforms the interface returning information specialized for `RaftKv`, through `SnapshotExt`.

### Check List 

Tests 

Refactoring only.

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/region_snapshot.rs
+++ b/components/raftstore/src/store/region_snapshot.rs
@@ -4,8 +4,10 @@
 use engine_traits::{
     IterOptions, KvEngine, Peekable, ReadOptions, Result as EngineResult, Snapshot,
 };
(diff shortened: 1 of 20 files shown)


youjiali1995 (approved):
LGTM. Much cleaner and unified.

longfangsong (approved):

PR #11488, opened 2021-11-26
Update max_ts for Prewrite and AcquirePessimisticLock with should_not_exist flag.

### LINKED ISSUE #11315: Prewrite and AcquirePessimisticLock's should_not_exist check doesn't update max_ts for async commit

Master

### Steps to reproduce

It might not be able to easily reproduce in a real cluster, but the problem exists theoretically.

Suppose a key `k` exists at `commit_ts = 5`, and currently `concurrency_manager.max_ts == 5`

1. Txn1 (`start_ts = 10`) prewrites or acquires pessimistic lock on `k`, with `should_not_exist` flag set. Since the key exists, it returns `AlreadyExist` error. `max_ts` is not updated.
2. Txn1 (`start_ts = 8`) prewrites a DELETE operation and succeeds, then committed at `commit_ts = 9`.
3. And now the key `k` doesn't exist at `ts=10`, which means, txn1 shouldn't report `AlreadyExist`.

This somewhat breaks linearizability. But I think it's not severe, since it seems unlikely that any user might depend on  this kind of linearizability.

dwangxxx:
Signed-off-by: dwangxxx <dwang_slam@163.com>

Issue Number: close #11315 

Problem Summary: Update max_ts when Prewrite and AcquirePessimisticLock with should_not_exist flag set for async commit.

### Check List 

Tests 

- Unit test

Side effects

- No

### Release note 

```release-note
None.
```


src/storage/txn/actions/acquire_pessimistic_lock.rs
```diff
@@ -25,6 +25,11 @@ pub fn acquire_pessimistic_lock<S: Snapshot>(
         crate::storage::mvcc::txn::make_txn_error(err, &key, reader.start_ts).into()
     ));
 
+    // Update max_ts for Insert operation to guarante linearizability and snapshot isolation
```
sticnarf, line 28: Could you also add a unit test for this?

src/storage/txn/commands/prewrite.rs
```diff
@@ -1511,6 +1511,64 @@ mod tests {
         }
     }
 
+    // this test for prewrite with should_not_exist flag
+    #[test]
+    fn test_prewrite_should_not_exist() {
```
MyonKeminta, line 1516: You should be able to get the current `max_ts` from `cm.max_ts()`. How about also checking the exact value in the tests?

src/storage/txn/actions/acquire_pessimistic_lock.rs
dwangxxx: Done. 
PTAL

src/storage/txn/commands/prewrite.rs
dwangxxx: Done.
PTAL

src/storage/txn/actions/acquire_pessimistic_lock.rs
```diff
@@ -1009,4 +1014,69 @@ pub mod tests {
...
+            for_update_ts,
+            need_value,
+            min_commit_ts,
+            need_old_value,
+        )
+        .unwrap_err();
+
+        assert_eq!(cm.max_ts().into_inner(), 15);
+
+        // T3: start_ts = 8, commit_ts = max_ts + 1 = 16, prewrite a DELETE operation on k
```
sticnarf, line 1055: Emm, you're simulating manually committing with async commit 😂 not a big problem, though
Actually, personally I think only checking the current `max_ts` is enough...

sticnarf (approved):
LGTM

src/storage/txn/actions/acquire_pessimistic_lock.rs
dwangxxx: Got it.

MyonKeminta (approved):

PR #11493, opened 2021-11-26
raftclient: count term and index in estimated size

### LINKED ISSUE #9714: batch messages can be too large in new raft client implementation

BusyJay:

Issue Number: close #9714.

Problem Summary:

If there are many entries in a message, the estimated size of message can be way smaller than
the actual size. This PR fixes the error by also counting index and term in estimation.

It also remove the hard limit as the estimation is closed enough.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Fix connection abort when too many raft entries are batched into one messages
```


src/server/raft_client.rs
```diff
@@ -174,7 +174,9 @@ impl Buffer for BatchMessageBuffer {
         let mut msg_size = msg.start_key.len()
             + msg.end_key.len()
             + msg.get_message().context.len()
-            + msg.extra_ctx.len();
+            + msg.extra_ctx.len()
+            // index: 3, term: 2, data tag and size: 3, entry tag and size: 3
+            + 11 * msg.get_message().get_entries().len();
```
NingLin-P, line 179: There are also other fields in `eraftpb::Message` like `to`, `from`, `term`, etc.

BusyJay: Yes, but there are at most 128 messages, so the size can be omitted.

src/server/raft_client.rs
```diff
@@ -568,7 +570,6 @@ where
 
         let cb = ChannelBuilder::new(self.builder.env.clone())
             .stream_initial_window_size(self.builder.cfg.grpc_stream_initial_window_size.0 as i32)
-            .max_send_message_len(self.builder.cfg.max_grpc_send_msg_len)
```
tonyxuqqi, line 571: So the max message limit is removed in grpc layer,  do we have metrics ( histogram) of the grpc size for this particular Raft batch API?  This is to validate if the size check in RaftClient code works in all the cases and help to diagnostics perf related issues

src/server/raft_client.rs
tonyxuqqi, line 179: minor:  per entry length can be added in the loop below.
msg_size += entry.data.len() + 11; // index: 3, term: 2, data tag and size: 3, entry tag and size: 3

src/server/raft_client.rs
BusyJay: Not yet. I plan to add one once middleware is supported in grpc-rs.

src/server/raft_client.rs
BusyJay: Yes, but it's efficient to use multiply than loop.

src/server/raft_client.rs
hicqu: Can we remove the entry `max_grpc_send_msg_len` from `Config`? Or mark it as deprecated.

BusyJay: No, it's used for controlling the batch size.

NingLin-P (approved):

hicqu (approved):

Issue #11119, opened 2021-10-21
Integrate with raft engine


tabokie:
## Development Task

TiKV subtask of tikv/raft-engine#95

- [x] Support write batch interfaces: `persist_size()`, `merge(rhs)`: https://github.com/tikv/raft-engine/pull/82, https://github.com/tikv/raft-engine/pull/122
- [x] Issue IO requests through IO rate limiter: #10937
- [ ] Tag IO in threads owned by raft engine
- [x] Support encryption at rest: #10937
- [x] Add Grafana monitoring: #11415

ztelur:
Hi! I am interested in raft. Is there some work I can contribue to ?

tabokie:
@ztelur Hi and welcome. The TiKV part of tikv/raft-engine#95 is almost done, you can still contribute to raft-engine.

Right now we are looking to releasing raft-engine in upcoming release (5.4.0), so code quality is the priority here. It basically boils down to two tasks:
1) Refactor critical code paths, add unit tests and assertions.
2) Implement "emergency exit" to handle possible data inconsistency caused by bug or disk corruption.

Other than these two, you can also contribute to a performance-oriented task: make raft engine supports multiple WALs. It's not yet on the roadmap, but we can discuss the details if you are interested.

ztelur:
@tabokie ok

PR #11497, opened 2021-11-28
engine: update raft engine for important bugfixes

tabokie:

Issue Number: ref #11119 

Problem Summary:

Found several critical bugs during functional test.

What's Changed:

See https://github.com/tikv/raft-engine/pull/160. Also implement `batch_gc` for raft engine.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raft_log_engine/src/engine.rs
+++ b/components/raft_log_engine/src/engine.rs
@@ -7,7 +7,8 @@ use std::sync::Arc;
 
 use encryption::{DataKeyManager, DecrypterReader, EncrypterWriter};
 use engine_traits::{
-    CacheStats, RaftEngine, RaftEngineReadOnly, RaftLogBatch as RaftLogBatchTrait, Result,
+    CacheStats, RaftEngine, RaftEngineReadOnly, RaftLogBatch as RaftLogBatchTrait, RaftLogGCTask,
+    Result,
 };
 use file_system::{IOOp, IORateLimiter, IOType};
 use kvproto::raft_serverpb::RaftLocalState;
@@ -246,8 +247,8 @@ impl RaftEngineReadOnly for RaftLogEngine {
 impl RaftEngine for RaftLogEngine {
     type LogBatch = RaftLogBatch;
 
-    fn log_batch(&self, _capacity: usize) -> Self::LogBatch {
-        RaftLogBatch::default()
+    fn log_batch(&self, capacity: usize) -> Self::LogBatch {
+        RaftLogBatch(LogBatch::with_capacity(capacity))
     }
 
     fn sync(&self) -> Result<()> {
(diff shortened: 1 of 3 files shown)


components/raft_log_engine/src/engine.rs
```diff
@@ -297,8 +298,34 @@ impl RaftEngine for RaftLogEngine {
...
+        }])
+    }
+
+    fn batch_gc(&self, tasks: Vec<RaftLogGCTask>) -> Result<usize> {
+        let mut batch = self.log_batch(tasks.len());
+        let mut old_first_index = Vec::with_capacity(tasks.len());
+        for task in &tasks {
+            batch
+                .0
+                .add_command(task.raft_group_id, Command::Compact { index: task.to });
```
hunterlxt, line 315: According to my understanding,`Compact` can only remove the previous range [-∞, to], but not the range selected by gc, am i right?

hunterlxt (approved):
Most LGTM

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1285,6 +1287,26 @@ impl<EK: KvEngine, ER: RaftEngine> RaftBatchSystem<EK, ER> {
         let raftlog_gc_scheduler = workers
             .background_worker
             .start_with_timer("raft-gc-worker", raftlog_gc_runner);
+        let router_clone = self.router();
+        let engines_clone = engines.clone();
+        workers.purge_worker.spawn_interval_task(
```
Connor1996, line 1292: move `raft-gc-worker` out of background worker instead of adding a new worker.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1285,6 +1287,26 @@ impl<EK: KvEngine, ER: RaftEngine> RaftBatchSystem<EK, ER> {
...
+            cfg.value().raft_engine_purge_interval.0,
+            move || {
+                match engines_clone.raft.purge_expired_files() {
+                    Ok(regions) => {
+                        for region_id in regions {
+                            let _ = router_clone.send(
+                                region_id,
+                                PeerMsg::CasualMessage(CasualMessage::ForceCompactRaftLogs),
+                            );
```
hunterlxt, line 1301: can you explain why leader needs gc log when spawning

components/raftstore/src/store/fsm/store.rs
tabokie: They should be in different threads, 1min+ purge job will block gc, causing OOM.

components/raftstore/src/store/fsm/store.rs
tabokie: That's a good point, but we have no way to tell if the region is leader here, so it will send out message regardlessly.

components/raft_log_engine/src/engine.rs
tabokie: Yes.

Connor1996 (approved):
LGTM

PR #11501, opened 2021-11-29
storage: make it cheaper to clone Scheduler

### LINKED ISSUE #11500: Cloning the scheduler wastes much CPU
Running a TiDB sysbench write only workload on a recent nightly TiKV, I it uses about 5% total CPU to clone and drop the inner `Engine` (the inner `RaftKv`):
image

`RaftKv` is a rather complex struct. And we actually don't need to clone them because it can be fetched from the TLS of scheduler threads.

sticnarf:

Issue Number: closes #11500

This PR removes the engine from `Scheulder`. This should reduce some CPU usage of the scheduler threads.

And it also moves `control_mutex` into `SchedulerInner`, so we can avoid some clones when it is not used.

### Check List 

Tests 

- Existing tests.

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/scheduler.rs
+++ b/src/storage/txn/scheduler.rs
@@ -23,6 +23,7 @@
 
 use crossbeam::utils::CachePadded;
 use parking_lot::{Mutex, MutexGuard};
+use std::marker::PhantomData;
 use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
@@ -184,6 +185,8 @@ struct SchedulerInner<L: LockManager> {
 
     flow_controller: Arc<FlowController>,
 
+    control_mutex: Arc<tokio::sync::Mutex<bool>>,
+
     lock_mgr: L,
 
     concurrency_manager: ConcurrencyManager,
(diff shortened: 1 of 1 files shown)


youjiali1995 (approved):
Good point. I missed it after always getting snapshot in the scheduler pool.

src/storage/txn/scheduler.rs
```diff
@@ -283,10 +286,10 @@ impl<L: LockManager> SchedulerInner<L> {
 /// Scheduler which schedules the execution of `storage::Command`s.
 #[derive(Clone)]
 pub struct Scheduler<E: Engine, L: LockManager> {
-    // `engine` is `None` means currently the program is in scheduler worker threads.
-    engine: Option<E>,
```
BusyJay, line 287: It doesn't seem to be used at all.

ekexium (approved):

src/storage/txn/scheduler.rs
sticnarf: Right. But we need to keep the type, so `with_tls_engine` can know what the actual type is.

BusyJay:
How come the CI of my PRs almost green while this one almost red?

PR #11502, opened 2021-11-29
raftstore: prevent removing voters directly in 2 replica raft group

### LINKED ISSUE #10595: Remove voter from a 2 replica raft group may lead to region unavailable

v5.1.0

Doesn't matter

### Steps to reproduce

In a 2 replica raft group, PD decides to remove a voter with simple `confchange` command, but the leader changed to the removing peer during the process and cause the leader to destroy itself. If the leader didn't sync the commitment of the `confchange` command to another peer (perhaps due to network problem or another store is temporarily down) before it is destroyed, another peer can't become leader since it needs the old leader's vote, and hence lead to region unavailable

There are few conditions that lead to this case:
- Leader change to the removing peer and cause the leader to destroy itself
- Removing the peer with simple `confchange` which allow removing voter directly, usually this is okay since there won't be more than one inflight simple `confchange` and removing one voter will still have the majority of voter alive. But with 2 replica raft group, there is not majority after one voter destroy
- `raft-rs` will broadcast a msg when a `confchange` log is committed, but it is failed due to network problem or other store is temporarily down

### What did you expect?
Region work normal

### What did happened?
Region unavailable

w41ter:
Signed-off-by: p4tr1ck <patricknicholas@foxmail.com>

Issue Number: close #10595 

Problem Summary:

It may lead to region unavailable when removing a voter from a 2 replica raft group directly.

What's Changed:

prevent removing voters directly in 2 replica raft group.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -2963,6 +2963,15 @@ where
                         cp
                     ));
                 }
+                (ConfChangeType::RemoveNode, PeerRole::Voter)
+                    if kind == ConfChangeKind::Simple && current_voter.len() == 2 =>
+                {
+                    return Err(box_err!(
+                        "{} unsafe request {:?}, to remove voter directly from a 2 voters raft group, please demoting voter firstly. see issue 10595 for detail",
+                        self.tag,
+                        cp
+                    ));
+                }
                 (ConfChangeType::RemoveNode, _)
                 | (ConfChangeType::AddNode, PeerRole::Voter)
                 | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {}
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -2967,7 +2967,7 @@ where
                     if kind == ConfChangeKind::Simple && current_voter.len() == 2 =>
                 {
                     return Err(box_err!(
-                        "{} unsafe request {:?}, to remove voter directly from a 2 voters raft group, please demoting voter firstly. see issue 10595 for detail",
+                        "{} removing voter directly from a 2 voters raft group may make region unavailable, please demoting voter firstly, request: {:?}",
                         self.tag,
                         cp
                     ));
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -1210,6 +1210,17 @@ impl<T: Simulator> Cluster<T> {
         }
     }
 
+    pub fn remove_peer(&mut self, region_id: u64, peer: metapb::Peer) -> RaftCmdResponse {
+        let epoch = self.get_region_epoch(region_id);
+        let remove_peer = new_admin_request(
+            region_id,
+            &epoch,
+            new_change_peer_request(ConfChangeType::RemoveNode, peer),
+        );
+        self.call_command_on_leader(remove_peer, Duration::from_secs(5))
+            .unwrap()
+    }
+
     pub fn transfer_leader(&mut self, region_id: u64, leader: metapb::Peer) {
         let epoch = self.get_region_epoch(region_id);
         let transfer_leader = new_admin_request(region_id, &epoch, new_transfer_leader_cmd(leader));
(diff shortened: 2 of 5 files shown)


PatrickNicholas:
@NingLin-P @BusyJay 
PTAL

PatrickNicholas:
@BusyJay @NingLin-P 
PTAL

components/raftstore/src/store/peer.rs
```diff
@@ -2963,6 +2963,15 @@ where
                         cp
                     ));
                 }
+                (ConfChangeType::RemoveNode, PeerRole::Voter)
+                    if kind == ConfChangeKind::Simple && current_voter.len() == 2 =>
+                {
+                    return Err(box_err!(
+                        "{} unsafe request {:?}, to remove voter directly from a 2 voters raft group, please demoting voter firstly. see issue 10595 for detail",
```
BusyJay, line 2970: How about leaving `issue 10595` in comment? It seems weird to referencing issue number in API resource.

components/test_raftstore/src/cluster.rs
```diff
@@ -1210,6 +1210,17 @@ impl<T: Simulator> Cluster<T> {
         }
     }
 
+    pub fn remove_peer(&mut self, region_id: u64, peer: metapb::Peer) -> RaftCmdResponse {
```
BusyJay, line 1213: You mean `self.async_remove_peer(region_id, peer).unwrap().recv().unwrap()`?

tests/integrations/raftstore/test_conf_change.rs
```diff
@@ -929,3 +929,40 @@ fn test_conf_change_fast() {
...
 }
+
+#[test]
+fn test_crash_when_remove_node_from_two_member_cluster() {
+    let mut cluster = new_node_cluster(0, 2);
+    let pd_client = Arc::clone(&cluster.pd_client);
+    pd_client.disable_default_operator();
+
+    let r1 = cluster.run_conf_change();
+    pd_client.must_add_peer(r1, new_peer(1, 1));
```
BusyJay, line 940: run_conf_change will add the first peer automatically.

components/raftstore/src/store/peer.rs
BusyJay: I don't think it's "unsafe" if it can only lead to unavailable. Unavailability is not safety.

tests/integrations/raftstore/test_conf_change.rs
PatrickNicholas: Thanks for your reply, I have address the comments in the above.

NingLin-P:
To fix #10595, I think it is also important to prevent PD from producing such operators. Otherwise, with this PR PD will enter a timeout-retry loop.

PatrickNicholas:
> To fix #10595, I think it is also important to prevent PD from producing such operators. Otherwise, with this PR PD will enter a timeout-retry loop.

@NingLin-P They are two repos, I will submit a new PR to PD once this PR is merged.

NingLin-P:
> @NingLin-P They are two repos, I will submit a new PR to PD once this PR is merged.

Perhaps it would be better to fix the PD side first. As currently, some nodes need to crash to trigger the issue, but with this PR the timeout-retry loop may be inevitable.

PatrickNicholas:
> Perhaps it would be better to fix the PD side first. As currently, some nodes need to crash to trigger the issue, but with this PR the timeout-retry loop may be inevitable.

@NingLin-P I have submit a PR to PD, PTAL.  https://github.com/tikv/pd/pull/4412

components/raftstore/src/store/peer.rs
```diff
@@ -2963,6 +2963,15 @@ where
                         cp
                     ));
                 }
+                (ConfChangeType::RemoveNode, PeerRole::Voter)
+                    if kind == ConfChangeKind::Simple && current_voter.len() == 2 =>
```
tonyxuqqi, line 2967: Should it be current_voter.len() <= 2

LintianShi:
@w41ter Are you willing to continue this PR? The work in PD side has been done in this PR https://github.com/tikv/pd/pull/4564. Or do I submit a new PR?

PR #11506, opened 2021-11-29
coprocesser: update encoding lib

### LINKED ISSUE #11505: make gbk encoding lib we used exactly the same with golang
## Development Task
we should make gbk encoding lib the same with golang.

xiongjiwei:

Issue Number: close #11505 

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.
```


components/tidb_query_datatype/Cargo.toml
```diff
@@ -37,4 +37,4 @@ tidb_query_common = { path = "../tidb_query_common", default-features = false }
 tikv_alloc = { path = "../tikv_alloc" }
 bstr = "0.2.8"
 log_wrappers = { path = "../log_wrappers" }
-encoding_rs = "0.8.28"
+encoding_rs = { git = "https://github.com/xiongjiwei/encoding_rs.git", branch = "golang"}
```
breeswish, line 40: It's surprising that Golang and Rust have different implementations for an encoding. Is it a fault in Golang side or a fault in Rust side?

andylokandy: If we have to use private repo to make it merge ASAP,  please open an issue for tracking further merging with `encoding_rs` or TiKV official fork. I'm also curious about which side to blame on. If golang is to blame, should it be fixed in golang side?

xiongjiwei: they all claimed they are following the https://encoding.spec.whatwg.org/. but they really have the difference, see https://gist.github.com/xiongjiwei/d78d276445a511dcef4497aa0a578f1f. All the different characters in golang can not encode and decode. I think they are all in private use part.

breeswish: How about opening an issue in the upstream and make a pull request for them?

xiongjiwei: they do not want to change how to handle the private use code point. see https://github.com/hsivonen/encoding_rs/issues/76. Also, we change `0x80` to make it compatible with MySQL both in golang and rust.

breeswish (approved):

wshwsh12 (approved):

sticnarf:
Still feeling this is not a perfect solution... Looks like the golang implementation is incorrent. Yet this can be accepted as a temporary workaround.

/merge

PR #11508, opened 2021-11-29
*: return wait time and process time for kv get

### LINKED ISSUE #8942: Return wait time and process time for KvGet and KvBatchGet
## Description

Now we have wait time and process time for Coprocessor request. However for other KV requests, like Get and BatchGet, this information is not available. Adding such information may help us figure out where latency comes from, for example, when TiDB's latency is 100ms while TiKV's latency is 70ms, the delta 30ms may come from network latency or Go routine scheduling.

This is a subtask extracted from https://github.com/tikv/tikv/issues/8938, only focus on KvGet and KvBatchGet, so that the development can be easier.

You need to do the following things:

1. Add wait time and process time fields in the Get and BatchGet response in [kvproto](https://github.com/pingcap/kvproto).
2. Count the wait time and the process time in TiKV KvGet and KvBatchGet, and return these time in the response. 

## Score

- 600

## Mentor(s)

- @breeswish
- @zhongzc

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Rust
* Protobuf

mornyx:

Close #8942 

What's Changed:

- Return wait time and process time for KvGet and KvBatchGet.

### Check List 

Tests 

- Manual test

Check the returned results of KvGet/KvBatchGet in `tests/integrations/server/kv_service.rs#test_mvcc_basic`.
I didn't write assertion into tests, because we might get 0ms in the test due to the simplicity of the logic. It is meaningless to assert whether the result is 0ms, and asserting whether the result will be greater than 0ms will lead unstable test.

### Release note 

```release-note
None
```


mornyx:
/cc @crazycs520 @zhongzc @breeswish 
PTAL, Thanks~

src/storage/mod.rs
```diff
@@ -696,7 +714,15 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         mut ctx: Context,
         raw_keys: Vec<Vec<u8>>,
         start_ts: TimeStamp,
-    ) -> impl Future<Output = Result<(Vec<Result<KvPair>>, Statistics, PerfStatisticsDelta)>> {
+    ) -> impl Future<
+        Output = Result<(
+            Vec<Result<KvPair>>,
+            Statistics,
+            PerfStatisticsDelta,
+            KvGetLatencyStats,
```
zhongzc, line 722: extract as a strcut?

src/storage/mod.rs
```diff
@@ -425,7 +425,15 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
-    ) -> impl Future<Output = Result<(Option<Value>, Statistics, PerfStatisticsDelta)>> {
+    ) -> impl Future<
+        Output = Result<(
+            Option<Value>,
+            Statistics,
+            PerfStatisticsDelta,
+            KvGetLatencyStats,
+        )>,
+    > {
+        let stage_begin_ts = Instant::now();
```
zhongzc, line 436: `now_coarse` instead?

src/storage/mod.rs
mornyx: This struct is currently only used for KvGet/KvBatchGet, so it is named KvGetXxx. If we will add similar statistics to more interfaces in the future, then we may need to change KvGetXxx to a more general name.

src/storage/mod.rs
zhongzc: at L436 start schedule waiting

at L447 finish schedule waiting
```rust
        let res = self.read_pool.spawn_handle(
            async move {
                // finished schedule wait here.

                tls_collect_query(
                    ctx.get_region_id(),
                    ctx.get_peer(),
                    key.as_encoded(),
                    key.as_encoded(),
                    false,
                    QueryKind::Get,
                );
            }

            // ...
        );
```

mornyx: > 

I learned that `wait_wall_time = schedule_wait_time + snapshot_wait_time`, and we seem to have no need to return `schedule_wait_time` , so I did not get unnecessary timestamp at the stage of `Scheduled`.

src/storage/mod.rs
zhongzc: I mean the return type seems complicated. How about extracting `(Statistics, PerfStatisticsDelta, KvGetLatencyStats)` as a struct?

mornyx: This is reasonable~

src/storage/mod.rs
mornyx: `now_coarse()` seems to have less precision, is it acceptable here?

zhongzc: > 

We can split them first and merge when responding.

mornyx: We get `wait_wall_time` by sub `begin_ts` from `snapshot_received_ts`, it seems unnecessary to calculate by adding `schedule_wait_time` and `snapshot_wait_time`. But I can put a temporarily useless `_stage_scheduled_ts` here for future needs.

zhongzc: > `now_coarse()` seems to have less precision, is it acceptable here?

Get request is more latency-sensitive than coprocessor. As you can see, most timing methods for coprocessor are `now_coarse`s, so it makes sense to get requests to act the same.

mornyx: > Get request is more latency-sensitive than coprocessor. As you can see, most timing methods for coprocessor are now_coarses, so it makes sense to get requests to act the same.

Got it

zhongzc: > We get `wait_wall_time` by sub `begin_ts` from `snapshot_received_ts`, it seems unnecessary to calculate by adding `schedule_wait_time` and `snapshot_wait_time`. But I can put a temporarily useless `_stage_scheduled_ts` here for future needs.

I prefer to place it to `KvGetLatencyStats`.

src/storage/mod.rs
mornyx: done [src/storage/mod.rs#L2503](https://github.com/tikv/tikv/pull/11508/files#diff-e3216d051c82e0ef881cdaeee1b11ef716ff9d72655f381da72a7828e9ca4b3cR2503)

src/storage/mod.rs
mornyx: done
[src/storage/mod.rs#L440](https://github.com/tikv/tikv/pull/11508/files#diff-e3216d051c82e0ef881cdaeee1b11ef716ff9d72655f381da72a7828e9ca4b3cR440)
[src/storage/mod.rs#L530](https://github.com/tikv/tikv/pull/11508/files#diff-e3216d051c82e0ef881cdaeee1b11ef716ff9d72655f381da72a7828e9ca4b3cR530)

zhongzc (approved):
LGTM, and at next PR we can do the same thing for `batch_get` and `batch_get_command`.

src/storage/mod.rs
```diff
@@ -2448,6 +2498,34 @@ pub mod test_util {
...
+    pub perf_stats: PerfStatisticsDelta,
+    pub latency_stats: KvGetLatencyStats,
+}
+
+/// Latency indicators related to KvGet/KvBatchGet.
+///
+/// The detailed meaning of the indicators is as follows:
+///
+/// ```text
+/// ------> Begin ------> Scheduled ------> SnapshotReceived ------> Finished ------>
```
breeswish, line 2514: I think this model is actually useful for all kv requests and even coprocessor requests. Is it possible to use this for all?

src/storage/mod.rs
```diff
@@ -2448,6 +2498,34 @@ pub mod test_util {
...
+/// |----- schedule_wait_time -----|
+///                                |-- snapshot_wait_time --|
+/// |------------------- wait_wall_time --------------------|
+///                                                         |-- process_wall_time --|
+/// |------------------------------ kv_read_wall_time ------------------------------|
```
breeswish, line 2519: Nice comments!

src/storage/mod.rs
mornyx: If so, I think we can change it to a more general name.
cc @zhongzc

zhongzc: Agreed

mornyx: In order to be used by multiple modules in the future, I moved it to [`components/tikv_kv/src/stats.rs`](https://github.com/tikv/tikv/pull/11508/files#diff-c206d7590a3c260a9ce287f7ee097981b45728491d48b4072e36fbe21633a679R268) and renamed it to `StageLatencyStats` which means `latency indicators for multi-execution-stages.`

zhongzc (approved):

breeswish (commented):
I have concerns below:

1. Currently there are two different ways of runtime measurements, one is the "StageLatencyStats" you introduced for KV Get (in Storage module) in this PR, and one is the original one for other requests like Coprocessor Request (in gRPC interface implementations). It is not a good idea to have different mechanisms for the same purpose. You'd better either keep the old mechanism, or change other places of the old mechanism to use the new mechanism. If you would like to do things step by step (for example, this PR is the first step of transiting to the new measurements), please open an issue including your roadmaps (e.g. what all tasks you want to do) to track it.

2. There are more detailed time measurements now, which is a good improvement but not enough. For example, the `schedule_wait_time` is not returned to TiDB, and TiDB does not utilize it as well (e.g. presenting it in slow query log and slow query tables), which means this PR is a semi-finished implementation and is not complete. For something that is far from complete, a roadmap tracking issue would be good.

3. If no new mechanism will be introduced, I'm worried that a TiDB side PR is also needed in order to utilize new information provided by this PR. Please take a look @crazycs520

components/tikv_kv/src/stats.rs
```diff
@@ -251,3 +251,23 @@ impl StatisticsSummary {
...
+/// ------> Begin ------> Scheduled ------> SnapshotReceived ------> Finished ------>
+/// |----- schedule_wait_time -----|
+///                                |-- snapshot_wait_time --|
+/// |------------------- wait_wall_time --------------------|
+///                                                         |-- process_wall_time --|
+/// |------------------------------ kv_read_wall_time ------------------------------|
+/// ```
+#[derive(Debug, Default, Copy, Clone)]
+pub struct StageLatencyStats {
+    pub schedule_wait_time: u64,
```
breeswish, line 269: It would be better to indicate the unit in the field name.

components/tikv_kv/src/stats.rs
```diff
@@ -251,3 +251,23 @@ impl StatisticsSummary {
...
+///
+/// The detailed meaning of the indicators is as follows:
+///
+/// ```text
+/// ------> Begin ------> Scheduled ------> SnapshotReceived ------> Finished ------>
+/// |----- schedule_wait_time -----|
+///                                |-- snapshot_wait_time --|
+/// |------------------- wait_wall_time --------------------|
+///                                                         |-- process_wall_time --|
+/// |------------------------------ kv_read_wall_time ------------------------------|
```
breeswish, line 265: This model might not fully work for all requests. I guess transactions have different stages. This could be a blocking stone to generalize it. It might also not work for batch get commands (which merges multiple get requests to be one).

src/storage/mod.rs
```diff
@@ -792,7 +820,29 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
                         .get(CMD)
                         .observe(command_duration.saturating_elapsed_secs());
```
breeswish, line 821: We have so many duplicated `elapsed` counts and `instant::now` calls, and they actually serve similar purposes. I'm looking forward to seeing refinements later. For example, some underlying time measurements provides both metrics and per-request statistics.

mornyx:
> I have concerns below:
> 
> 1. Currently there are two different ways of runtime measurements, one is the "StageLatencyStats" you introduced for KV Get (in Storage module) in this PR, and one is the original one for other requests like Coprocessor Request (in gRPC interface implementations). It is not a good idea to have different mechanisms for the same purpose. You'd better either keep the old mechanism, or change other places of the old mechanism to use the new mechanism. If you would like to do things step by step (for example, this PR is the first step of transiting to the new measurements), please open an issue including your roadmaps (e.g. what all tasks you want to do) to track it.
> 2. There are more detailed time measurements now, which is a good improvement but not enough. For example, the `schedule_wait_time` is not returned to TiDB, and TiDB does not utilize it as well (e.g. presenting it in slow query log and slow query tables), which means this PR is a semi-finished implementation and is not complete. For something that is far from complete, a roadmap tracking issue would be good.
> 3. If no new mechanism will be introduced, I'm worried that a TiDB side PR is also needed in order to utilize new information provided by this PR. Please take a look @crazycs520

I created an issue to track this roadmap: https://github.com/tikv/tikv/issues/11525

components/tikv_kv/src/stats.rs
mornyx: done

components/tikv_kv/src/stats.rs
mornyx: I recorded this note in roadmap issue: https://github.com/tikv/tikv/issues/11525

src/storage/mod.rs
mornyx: I listed it as a separate task in the roadmap issue.

crazycs520 (commented):
LGTM

mornyx:
/assign @breeswish 
Please see if this PR can be merged, thanks~

breeswish (approved):

breeswish (approved):

PR #11516, opened 2021-11-30
txn: Support check existence without getting value during pessimistic lock requests

MyonKeminta:

Ref https://github.com/pingcap/tidb/issues/26833

What's Changed: This is part of https://github.com/pingcap/tidb/issues/26833 and was split out from #10977 .

This PR makes PessimisticLock requests support checking if the key already exists. The logic of reading is already implemented for returning value from pessimistic lock requests, which is reused in this PR. This PR just makes it possible to avoid transmitting the values through network when it's not necessary.

The PR is marked as draft because the related kvproto is not merged yet.

### Related changes

- Requires kvproto: https://github.com/pingcap/kvproto/pull/824

### Check List 

Tests 

- Unit test

Side effects

\-

### Release note 

```release-note
None
```


MyonKeminta:
How can I pass the commit message check...

ekexium:
> How can I pass the commit message check...

I guess there must be a related issue in tikv

src/storage/types.rs
```diff
@@ -142,6 +144,10 @@ impl PessimisticLockRes {
                     (v.unwrap_or_default(), is_not_found)
                 })
                 .unzip(),
+            PessimisticLockRes::Existence(mut vals) => {
+                vals.iter_mut().for_each(|x| *x = !*x);
+                (vec![], vals)
```
cfzjywxk, line 149: The returned result formats are diffrent for `value return` and `existence check`, would this confuse the `PessimisticLockResponse` users?

ekexium: Who would be confused?

MyonKeminta: 🤔 Sorry I didn't understand... This behavior only happens when `return_value` flag in the request is not set but `check_existence` is set, and when setting the flag, the user should know what (s)he is doing.

src/storage/txn/commands/acquire_pessimistic_lock.rs
```diff
@@ -101,12 +106,12 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for AcquirePessimisticLock
                 should_not_exist,
                 self.lock_ttl,
                 self.for_update_ts,
-                self.return_values,
+                self.return_values || self.check_existence,
```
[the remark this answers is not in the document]
MyonKeminta: Ah, yes.

src/storage/types.rs
cfzjywxk: The two fields `values` and `not_founds` in `PessimisticLockResponse` would be used by both `return_value` and `check_exisstence` flags after this change. The result formats are a bit different, the user needs to know exactly the difference. 
It dose not matter much and for simplicy seems this is the only way.

MyonKeminta:
@cfzjywxk @ekexium PTAL again, thanks!

ekexium (approved):

cfzjywxk (approved):

youjiali1995 (approved):

PR #11085, opened 2021-10-18
resrc_mtr: introduce scanned keys statistics

mornyx:
### What problem does this PR solve?

close https://github.com/tikv/tikv/issues/11690

Added support for the collection and reporting of the scanned keys.

### What is changed and how it works?

What's Changed:

- Added `SummaryRecorder` which implements `SubRecorder`.
- Three fields have been added to `LocalStorage` for statistics of scanned keys.
- Added the statistical logic of scanned keys in the context of ResourceMeteringTag.
- Added some fields related to scanned keys to the structure in `model.rs`.
- Handle extra fields in `client.rs`.

### Related changes

- [pingcap/kvproto#814](https://github.com/pingcap/kvproto/pull/814)

### Check List

Tests

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note

```release-note
Support the statistics of scanned keys in resource metering.
```

PR #11522, opened 2021-12-01
rsmeter: tag factory instead of global variable

### LINKED ISSUE #11229: test_resource_metering is too slow
test_resource_metering takes 4 minutes to finish, which is too long.

https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/10151/pipeline/112#step-333-log-108

zhongzc:

Close #11229

What's Changed:

Remove global variable `STORAGE_CHANS`and `COLLECTOR_REG_CHAN`
- Add `ResourceTagFactory`
- Add `CollectorRegistry`

image

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


zhongzc:
cc @mornyx @andylokandy

components/resource_metering/src/localstorage.rs
```diff
@@ -49,35 +25,5 @@ pub struct LocalStorage {
...
-        .unwrap();
-        let mut count = 0;
-        while let Ok(r) = rx.try_recv() {
-            assert_ne!(r.id, 0);
-            count += 1;
-        }
-        // This value may be greater than 2 if other test threads access `STORAGE` in parallel.
-        assert!(count >= 2);
-    }
+    pub storage: LocalStorage, // TODO: change to shared_ptr
```
mornyx, line 28: Does this TODO mean to use `SharedTagPtr` directly instead? We will add more fields to `LocalStorage` in the scan-row PR, so here we still need to refer to the entire LocalStorage instead of a single tag.

mornyx:
Generally LGTM, some implementation details need more time to review.

components/resource_metering/src/localstorage.rs
zhongzc: `SharedTagPtr` contains an Arc. Can we use the same `Arc` to wrap other fields together? Of course, will rename `SharedTagPtr`.

mornyx: Yes, we can put them in LocalStorage, of course they can also be placed under `SharedTagPtr` (need rename). See [#11085 L76-L79](https://github.com/tikv/tikv/pull/11085/files#diff-8a4ce4dc3080cfb7d9bf6ae0f53d8e877a4fedc8321298ed79b695914373ed17R76-R79)

zhongzc: Then we can simply `cleanup` recorder.

andylokandy (approved):

components/resource_metering/src/recorder/mod.rs
```diff
@@ -238,28 +239,36 @@ impl RecorderBuilder {
     /// Spawn a new thread that executes the main loop of the [Recorder].
     ///
     /// This function does not return the `Recorder` instance directly, instead
-    /// it returns a [RecorderHandle] that is used to control the execution.
-    pub fn spawn(self) -> io::Result<RecorderHandle> {
+    /// it returns a [RecorderHandle] that is used to control the execution, and
+    /// a [ResourceTagFactory] that is used for construction of resource tag.
+    pub fn spawn(self) -> io::Result<(RecorderHandle, CollectorRegistry, ResourceTagFactory)> {
```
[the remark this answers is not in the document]
zhongzc: Why not? The collector is tight to the recorder.

src/storage/mod.rs
```diff
@@ -1421,7 +1426,7 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
         const CMD: CommandKind = CommandKind::raw_batch_get;
         let priority = ctx.get_priority();
         let priority_tag = get_priority_tag(priority);
-        let resource_tag = ResourceMeteringTag::from_rpc_context(&ctx);
+        let resource_tag = self.tag_factory.new_tag(&ctx);
```
breeswish, line 1429: maybe call it `resource_tag_factory` is better, as we have so many "tags".  A single "tag_factory" will cause ambiguity.

breeswish (commented):
The rest LGTM

components/resource_metering/src/lib.rs
```diff
@@ -90,20 +96,55 @@ impl ResourceMeteringTag {
...
+
+    fn register_local_storage(&self, storage: &LocalStorage) -> bool {
+        let lsr = LocalStorageRef {
+            id: utils::thread_id(),
+            storage: storage.clone(),
+        };
+        match self.tx.send(lsr) {
+            Ok(_) => true,
+            Err(err) => {
+                warn!("failed to register thread"; "err" => ?err);
```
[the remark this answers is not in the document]
zhongzc: The channel is full or the receiver is dropped (i.e. the recorder thread is down).

components/resource_metering/src/lib.rs
```diff
@@ -70,14 +66,24 @@ impl ResourceMeteringTag {
     ///
     /// [STORAGE]: crate::localstorage::STORAGE
     pub fn attach(&self) -> Guard {
-        STORAGE.with(|s| {
-            if s.is_set.get() {
-                panic!("nested attachment is not allowed")
+        STORAGE.with(move |s| {
+            let mut ls = s.borrow_mut();
+
+            if unlikely(!ls.registered && ls.register_failed_times < MAX_THREAD_REGISTER_RETRY) {
```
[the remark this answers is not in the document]
zhongzc: > When will register fail?
The channel is full or the receiver is dropped (i.e. the recorder thread is down).

If the channel is full, it means a lot of threads have requests attached in a short period of time, and these threads have no requests attached before.

if the receiver is dropped, it means some panic happens on the record thread.

> if it failed what will happen?
Won't register if failed too many times before or reregister.

breeswish (approved):

PR #11528, opened 2021-12-01
backup: pipeline scan and save step

### LINKED ISSUE #11350: backup: run rocksDB operations (which are blocking) in separate worker pool
## Development Task
After #11158, the 'save file' part has been refactored to asynchronous functions. However, the `scan` part (which involves rocksDB operations, which includes blocking local disk I/O.) is still blocking and CPU bound. 
Some test cases shows that, with `noop://` backend, one backup worker thread can utilize 100% of CPU.

image

For now, we have a `semaphore` to limit the concurrency of backup tasks, but backup threads keep the `permit` when they are doing async i/o too:

Synchronous asynchronous I_O

We can reduce the problem by separating the control thread pool into two parts:

1. A tiny (maybe 1 or 2 threads) for running real async tasks (write, scan regions, etc.).
2. A worker pool (maybe `yatp::ThreadPool`) for scan tasks.

Task list:

- [ ] decouple scanning and writing.
- [ ] make local backend use `tokio::fs::File` instead of `std::fs::File`. (which is essential, or when we backing up to NFS, the blocking fs operations may block everything)
- [ ] add a new config for changing the tokio runtime size (not allowed to be modified in runtime).

YuJuncen:

Close #11350.

What's Changed:

Added a new worker pool, and run I/O in it.

### TODO

- [x] ~~Make local file I/O (`local://` storages) async.~~ (This has already be done when refactoring backup to async...)
- [x] Add config terms.

### Related changes

### Check List 

Tests 

- Integration test (The old test cases can help)

### Release note 

```release-note
None
```


components/backup/src/endpoint.rs
```diff
@@ -151,6 +152,118 @@ pub struct BackupRange {
...
+
+impl KvWriter {
+    async fn save(self, storage: &dyn ExternalStorage) -> Result<Vec<File>> {
+        match self {
+            Self::TiDB(writer) => writer.save(storage).await,
+            Self::Raw(writer) => writer.save(storage).await,
+        }
+    }
+
+    fn need_flush_keys(&self) -> bool {
```
3pointer, line 170: why not use this before send to `InMemBackupFiles` channel?

YuJuncen:
```
[2021-12-15T05:44:27.172Z] thread 'main' panicked at 'assertion failed: `(left != right)`
[2021-12-15T05:44:27.172Z]   left: `0`,
[2021-12-15T05:44:27.172Z]  right: `0`', components/backup/src/endpoint.rs:1507:13
```
🤔

components/backup/src/endpoint.rs
```diff
@@ -151,6 +152,118 @@ pub struct BackupRange {
...
+    start_key: Vec<u8>,
+    end_key: Vec<u8>,
+    start_version: TimeStamp,
+    end_version: TimeStamp,
+    region: Region,
+}
+
+impl std::fmt::Debug for InMemBackupFiles {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("InMemBackupFiles")
```
kennytm, line 189: should just implement a dummy `Debug` for `KvWriter` and `#[derive(Debug)]` for `InMemBackupFiles`.

components/backup/src/endpoint.rs
```diff
@@ -151,6 +152,118 @@ pub struct BackupRange {
     cf: CfName,
 }
 
+/// The generic saveable writer. for generic `InMemBackupFiles`.
+/// Maybe what we really need is make Writer a trait...
+enum KvWriter {
+    TiDB(BackupWriter),
```
kennytm, line 158: this should be `Txn` not `TiDB`? the backup API isn't TiDB-specific.

components/backup/src/endpoint.rs
```diff
@@ -293,33 +389,35 @@ impl BackupRange {
                 return Err(e);
             }
         }
+        drop(snap_store);
+        let stat = scanner.take_statistics();
+        if start_scan.saturating_elapsed_secs() > 30.0 {
```
kennytm, line 394: place `start_scan.saturating_elapsed_secs()` into a variable. the elapsed duration is not a pure function.

YuJuncen:
[2021-12-20T06:39:59.159Z] thread 'main' panicked at 'assertion failed: `(left != right)`
[2021-12-20T06:39:59.159Z]   left: `0`,
[2021-12-20T06:39:59.159Z]  right: `0`', components/backup/src/endpoint.rs:1513:13

Which means RocksDB may not trigger any disk I/O, can we just commit this?

kennytm:
what do you mean by "commit this"

YuJuncen:
> what do you mean by "commit this"

Just comment the assertion of `stats.fetch(IOType::Export, IOOp::Read) == 0`.

BTW, I guess this would be a bug introduced by the PR(Which made the statistic invalid.): during backup, the task may be scheduled to other threads (which doesn't set the thread local I/O type by `file_system::set_io_type`.).

kennytm:
ok you meant "comm<b>ent</b> (out)" not "comm<b>it</b>".

components/backup/src/metrics.rs
```diff
@@ -41,4 +41,21 @@ lazy_static! {
...
+    )
+    .unwrap();
+    pub static ref BACKUP_SCAN_KV_COUNT : IntCounterVec = register_int_counter_vec!(
+        "tikv_backup_scan_kv_count",
+        "Total number of kvs backed up",
+        &["cf"],
+    )
+    .unwrap();
+    pub static ref BACKUP_SCAN_KV_SIZE : IntCounterVec = register_int_counter_vec!(
+        "tikv_backup_scan_kv_size",
```
kennytm, line 56:
```suggestion
        "tikv_backup_scan_kv_size_bytes",
```

following the naming convention https://prometheus.io/docs/practices/naming/#metric-names

components/backup/src/endpoint.rs
```diff
@@ -1401,7 +1501,7 @@ pub mod tests {
             assert!(!resp.has_error(), "{:?}", resp);
             let file_len = if *len <= SHORT_VALUE_MAX_LEN { 1 } else { 2 };
             let files = resp.get_files();
-            info!("{:?}", files);
+            println!("{:?}", files);
```
kennytm, line 1504: revert?

kennytm (approved):
rest lgtm

3pointer (approved):
LGTM

PR #11529, opened 2021-12-01
storage: read and write in-memory pessimistic locks in the scheduler 

sticnarf:
- [x] Based on #11481

Ref #11452 

What's Changed:

This PR supports reading and writing pessimistic locks in the transaction layer. 

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -93,7 +93,7 @@ use kvproto::kvrpcpb::{
     RawGetRequest,
 };
 use kvproto::pdpb::QueryKind;
-use raftstore::store::util::build_key_range;
+use raftstore::store::{util::build_key_range, TxnExt};
 use raftstore::store::{ReadStats, WriteStats};
 use rand::prelude::*;
 use resource_metering::{FutureExt, ResourceMeteringTag};
@@ -105,6 +105,7 @@ use std::{
         Arc,
     },
 };
+use tikv_kv::SnapshotExt;
 use tikv_util::time::{Instant, ThreadReadId};
 use txn_types::{Key, KvPair, Lock, OldValues, RawMutation, TimeStamp, TsSet, Value};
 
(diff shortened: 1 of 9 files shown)


src/storage/txn/scheduler.rs
```diff
@@ -739,221 +741,276 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
             };
 
+            // FIXME: avoid this clone
             task.cmd
-                .process_write(snapshot, context)
+                .process_write(snapshot.clone(), context)
                 .map_err(StorageError::from)
         };
-        match deadline
+        let WriteResult {
```
sticnarf, line 749: Here most changes are caused by reduce indentation with early return.

src/storage/txn/scheduler.rs
```diff
@@ -740,220 +742,274 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+        let sched_pool = scheduler.get_sched_pool(priority).pool.clone();
+        let txn_ext = snapshot.ext().get_txn_ext().cloned();
+        // The callback to receive async results of write prepare from the storage engine.
+        let engine_cb = Box::new(move |result: EngineResult<()>| {
+            sched_pool
+                .spawn(async move {
+                    fail_point!("scheduler_async_write_finish");
+
+                    let ok = result.is_ok();
+                    if ok {
```
youjiali1995, line 907:
```suggestion
                    if ok && !locks_to_be_removed.is_empty() {
```

src/storage/txn/scheduler.rs
```diff
@@ -740,220 +742,274 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+        let write_size = to_be_write.size();
+
+        // Mutations on the lock CF should overwrite the memory locks.
+        // These mutations happen on a working leader, so memory pessimistic locks
+        // are only written through the scheduler. Protected by the latches, there will
+        // be no concurrent write on the same keys in the lock table. So, it is safe
+        // for us to only delete keys that are already in the lock table.
+        // This will help us reduce the cost of cloning the keys for the commit command
+        // or when the feature is disabled.
+        let locks_to_be_removed = {
```
youjiali1995, line 814: `pessimistic_rollback` still writes to disk even if all pessimistic locks are in memory?

src/storage/mod.rs
```diff
@@ -2182,6 +2183,112 @@ impl TestStorageBuilder<RocksEngine, DummyLockManager> {
     }
 }
 
+#[derive(Clone)]
+pub struct TxnEngine<E: Engine> {
```
youjiali1995, line 2187: `TxnEngine` and `TxnSnapshot` are only used for tests? I thought you added a new formal wrapper at first glance...

sticnarf: Ah, yes. They are only used in tests. I will add "Test" to the names.

src/storage/txn/scheduler.rs
sticnarf: Right. It can be optimized but the current approach is simpler.
To implement what you said, we need to check whether the lock exists in the memory table in the pessimistic rollback command and add a new modify type. We can do it in the future.

src/storage/txn/scheduler.rs
```diff
@@ -740,220 +742,274 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
-        match deadline
+        let WriteResult {
+            ctx,
+            mut to_be_write,
+            rows,
+            pr,
+            lock_info,
+            lock_guards,
+            response_policy,
+        } = match deadline
```
cfzjywxk, line 756: Besides deadline check do we have other situations the `high_priority_pool` will be used? There're some ddl transation options which would set the command priority but they're rarely used.

youjiali1995: resolve-lock requests use `high_priority_pool`.

cfzjywxk (approved):
LGTM

src/storage/txn/scheduler.rs
```diff
@@ -740,220 +742,274 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+        let mut is_async_apply_prewrite = false;
+        let write_size = to_be_write.size();
+
+        // Mutations on the lock CF should overwrite the memory locks.
+        // These mutations happen on a working leader, so memory pessimistic locks
+        // are only written through the scheduler. Protected by the latches, there will
+        // be no concurrent write on the same keys in the lock table. So, it is safe
+        // for us to only delete keys that are already in the lock table.
+        // This will help us reduce the cost of cloning the keys for the commit command
+        // or when the feature is disabled.
```
youjiali1995, line 813: > // or when the feature is disabled.

Why locks_to_be_removed may not be empty even if this feature is disabled? To prevent memory locks leak if the feature is turned off online?

youjiali1995 (approved):

src/storage/txn/scheduler.rs
sticnarf: Yes, it can be turned off online. Actually, I didn't think about it, I just naturally use a uniform way to try reducing the cost = =

PR #11536, opened 2021-12-01
storage: Increase default reserve space to 16 GB

LINKIWI:

What's Changed:

This change increases the default value of `storage.reserve-space` from 5 GB to 16 GB.

By default, RocksDB uses 8 background threads and requires at least 2 GB of disk space each for compaction. In order to play nicely with RocksDB defaults, also increase the reserve space to 16 GB to be consistent with this requirement.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`: I will file a PR to update the docs if there is interest in moving forward with this change.
- PR to update `pingcap/tidb-ansible`: N/A

### Check List 

Tests 

Manual test:

TODO: will post details for manual verification soon.

Side effects

This may break for installations deployed on disks smaller than 16 GB, though in practice this is an unlikely deployment environment.

### Release note 

```release-note
storage: Increase default reserve space to 16 GB
```


hicqu:
@tier-cap would you like to explain space reserving policy based on the latest implementation?

PR #11550, opened 2021-12-02
tikv_util: reimplement poll_future_notify

### LINKED ISSUE #11549: TiKV PD worker thread deadlocks

v5.3.0

\-

### Steps to reproduce

The exact way to trigger the issue is currently uncertain.

### What did you expect?

TiKV runs normally.

### What did happened?

1.  Its found that some queries in TiDB reports this error:

```
MySQL [test]> insert into t1(id,name) value(1,'dfdfd'),(2,'dfdfdffff');
ERROR 9011 (HY000): TiKV max timestamp is not synced
```

This means async commit / 1pc transaction fails due to not updating the max_ts associated to this region.

2. One of the TiKV process is alive, but PD shows it's down. The TiKV's log is nearly empty, excepts some PD reconnection related logs that's printed every 10 minutes.

3. The metrics shows this TiKV node's PD heartbeat is zero, however, TiKV's GC worker is updating GC safepoint normally.

I think it looks like a deadlock related to the PD worker thread. It's confirmed by @BusyJay .

Partial stack provided by @BusyJay :

```
Thread 36 (Thread 0x7fb7b5df3700 (LWP 13739)):
#0  0x00007fb82e2dd54d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007fb82e2d8e9b in _L_lock_883 () from /lib64/libpthread.so.0
#2  0x00007fb82e2d8d68 in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00005597d049cd69 in std::sys::unix::mutex::Mutex::lock (self=<optimized out>) at /rustc/2faabf579323f5252329264cc53ba9ff803429a3/library/std/src/sys/unix/mutex.rs:63
#4  std::sys_common::mutex::MovableMutex::raw_lock (self=<optimized out>) at /rustc/2faabf579323f5252329264cc53ba9ff803429a3/library/std/src/sys_common/mutex.rs:76
#5  std::sync::mutex::Mutex<T>::lock (self=<optimized out>) at /rustc/2faabf579323f5252329264cc53ba9ff803429a3/library/std/src/sync/mutex.rs:261
#6  <tikv_util::future::BatchCommandsWaker as futures_task::arc_wake::ArcWake>::wake_by_ref (arc_self=<optimized out>) at components/tikv_util/src/future.rs:84
```

```
36   Thread 0x7fb7b5df3700 (LWP 13739) "pd-worker-0"     0x00007fb82e2dd54d in __lll_lock_wait () from /lib64/libpthread.so.0
```

BusyJay:

The current implementation is very easy to get deadlock if the future is
woken immediately during polling. This PR resolves the issue by
maintaining states instead.

Note, if the future is implemented in the wrong way that always
notifying the future without checking if it has readiness, the new
implementation can run into dead loop.

Close #11549.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Fix deadlock in some rare cases that futures get resolved too fast
```

### DIFF
--- a/components/tikv_util/src/future.rs
+++ b/components/tikv_util/src/future.rs
@@ -86,7 +86,7 @@ const IDLE: u8 = 0;
 /// The future is being polled by some thread.
 const POLLING: u8 = 1;
 /// The future is woken when being polled.
-const PENDING: u8 = 2;
+const NOTIFIED: u8 = 2;
 
 /// A waker that will poll the future immediately when waking up.
 struct PollAtWake {
@@ -113,7 +113,7 @@ impl PollAtWake {
                 POLLING => {
                     match arc_self.state.compare_exchange_weak(
                         POLLING,
-                        PENDING,
+                        NOTIFIED,
                         Ordering::SeqCst,
                         Ordering::SeqCst,
                     ) {
@@ -122,7 +122,7 @@ impl PollAtWake {
                         Err(s) => state = s,
                     }
                 }
-                PENDING => {
+                NOTIFIED => {
                     // It will be polled again, so we don't need to do anything here.
                     return;
                 }
@@ -158,10 +158,10 @@ impl PollAtWake {
             {
                 Ok(_) => return,
                 Err(s) => {
-                    if s == PENDING {
-                        // Only this thread can change the state from PENDING, so it has to succeed.
+                    if s == NOTIFIED {
+                        // Only this thread can change the state from NOTIFIED, so it has to succeed.
                         match arc_self.state.compare_exchange(
-                            PENDING,
+                            NOTIFIED,
                             POLLING,
                             Ordering::SeqCst,
                             Ordering::SeqCst,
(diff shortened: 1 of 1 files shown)


components/tikv_util/src/future.rs
```diff
@@ -72,25 +74,152 @@ where
 /// thread calling `wake()`.
```
sticnarf, line 74:
```suggestion
/// thread calling `poll_future_notify()`.
```

components/tikv_util/src/future.rs
```diff
@@ -72,25 +74,152 @@ where
...
 }
 
-// BatchCommandsWaker is used to make business pool notifiy completion queues directly.
-struct BatchCommandsWaker(Mutex<Option<BoxFuture<'static, ()>>>);
+/// The future is not processed by any one.
+const IDLE: u8 = 0;
+/// The future is being polled by some thread.
+const POLLING: u8 = 1;
+/// The future is woken when being polled.
+const PENDING: u8 = 2;
```
sticnarf, line 89: I was confused by the name `PENDING` at first. Personally, I think `NOTIFIED` is easier to understand...

components/tikv_util/src/future.rs
BusyJay: `wake` is correct. It's polled in `wake`.

sticnarf: I understand the comment now.

sticnarf (approved):
LGTM

youjiali1995 (approved):

hicqu (approved):

PR #11551, opened 2021-12-02
raftstore: reduce decode raft command time on leader

### LINKED ISSUE #10783: Cache `RaftCmdRequest` along with entry
I can see some CPU (about 10% of total apply CPU) of apply worker is consumed by decoding `RaftCmdRequest`.

We already cache the entry data, which is useful when broadcasting entries to followers. We can also cache the `RaftCmdRequest` along with the data. This can reduce some CPU usage on the leader side, who has already known the RaftCmdRequest. To reduce memory consumption, we may also want to enable bytes to `RaftCmdRequest`. Of course, we need to prototype it and see whether it can bring enough improvement before actually optimizing it.

gengliqi:

Close #10783

What's Changed:

#10783 proposed a method to reduce decode command time on leader, which is caching `RaftCmdRequest` along with `Entry`.
But this PR uses a simpler method that has the same optimization effect.

The `RaftCmdRequest` can be carried with `Proposal` after proposing. Therefore, the `RaftCmdRequest` can be easily taken from `pending_cmds` when applying this cmd.

### Related changes

### Check List 

Tests 

Side effects

- No

### Release note 

```release-note
None.
```


BusyJay:
How does it perform?

gengliqi:
> How does it perform?

TiKV uses default config. The workload is sysbench insert.
The sequence is master, this PR, master, this PR.
image
image
image
The improvement is not very obvious but it's expected.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1126,7 +1130,33 @@ where
         }
     }
 
-    fn find_pending(
+    /// Take the pending normal command.
+    /// Note that the `PendingCmd` of this index and term is not popped.
+    fn take_pending_normal_cmd(&mut self, index: u64, term: u64) -> Option<RaftCmdRequest> {
+        while let Some(mut head) = self.pending_cmds.pop_normal(index, term) {
```
BusyJay, line 1136: How about introducing a new method to get a mutable reference instead? So the command doesn't need to be moved twice.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1126,7 +1130,33 @@ where
         }
     }
 
-    fn find_pending(
+    /// Take the pending normal command.
+    /// Note that the `PendingCmd` of this index and term is not popped.
+    fn take_pending_normal_cmd(&mut self, index: u64, term: u64) -> Option<RaftCmdRequest> {
```
BusyJay, line 1135: Should comment why conf change doesn't need to be handled.

BusyJay (commented):
The PR will make the memory control inefficient on pending apply tasks.

Issue #11572, opened 2021-12-04
Extra next call in raw_scan


BusyJay:
There will be always an extra next call if any key is about to be returned.

https://github.com/tikv/tikv/blob/fd24cc9e3b43f77f0dceba5bf20f7b4f9e9dae44/src/storage/raw/store.rs#L175-L192

PR #11576, opened 2021-12-06
storage: reduce extra next/prev in raw_scan

JayiceZ:
Signed-off-by: JayiceZ <1185430411@qq.com>

 optimize logic in raw_scan

close https://github.com/tikv/tikv/issues/11572

What's Changed:
optimize logic in raw_scan

### Check List 

Tests 

- Unit test

### Release note 

```release-note
optimize logic in raw_scan
```


src/storage/raw/store.rs
```diff
@@ -189,6 +189,9 @@ impl<'a, S: Snapshot> RawStoreInner<S> {
                     cursor.value(statistics).to_owned()
                 },
             )));
+            if pairs.len() >= limit {
```
BusyJay, line 192: How about `if pairs.len() < limit { cursor.next() } else { break }`? The difference is the first branch is likely to hit.

Connor1996 (approved):
LGTM

src/storage/raw/store.rs
JayiceZ: Sure! It is better, I will modify it.

src/storage/raw/store.rs
```diff
@@ -189,7 +189,11 @@ impl<'a, S: Snapshot> RawStoreInner<S> {
                     cursor.value(statistics).to_owned()
                 },
             )));
-            cursor.next(statistics);
+            if pairs.len() < limit {
+                cursor.next(statistics);
+            }else{
```
Connor1996, line 194: please make format

JayiceZ: done.PTAL

JayiceZ: Fixed. PTAL

Connor1996:
Please fix the DCO check

BusyJay (approved):

Connor1996:
Please provide the associated issue number in the commit messages, for example: close #12345, or ref #12345.

src/storage/raw/store.rs
```diff
@@ -172,7 +172,7 @@ impl<'a, S: Snapshot> RawStoreInner<S> {
         let mut pairs = vec![];
         let mut row_count = 0;
         let mut time_slice_start = Instant::now();
-        while cursor.valid()? && pairs.len() < limit {
+        while cursor.valid()? {
```
pingyu, line 175: It seems that this modification would cause wrong result when `limit == 0`.

And this also causes the failure of unittest `test_txn_store_rawkv` (see the [log](https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/12889/pipeline/115/#step-326-log-202)).
To be specific, this [test](https://github.com/tikv/tikv/blob/1f3b5cfa804077add976144296bdf770cff0246f/tests/integrations/storage/test_storage.rs#L773):
```
store.raw_scan_ok("".to_string(), b"".to_vec(), None, 0, vec![]);
```

JayiceZ: Fixed.
Sorry, I forgot to run tests after my modification, I will pay attention next time.

BusyJay:
Can you fix the "FORMAT CHECKER NOTIFICATION" warnings?

JayiceZ:
Commit was messed up by me accidentally. I think it is more convenient to open a new pr.😭 Instead of reverting them

PR #11577, opened 2021-12-06
raftstore: propose in-memory pessimistic locks before leader transfer

sticnarf:
- [x] Based on https://github.com/tikv/tikv/pull/11591

Refer to #11452 

To avoid losing the in-memory pessimistic locks after transferring leader, this PR proposes writing commands of the pessimistic locks before leader transfer.

After proposing pessimistic locks, we propose another `TransferLeader` command. This will avoid proposing pessimistic locks increasing unavailable time before `MsgTimeout`, and give chance for the write command that have been sent to the raftstore to propose.

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```


components/raftstore/src/store/peer.rs
```diff
@@ -3696,16 +3704,72 @@ where
...
+            mem::take(&mut pessimistic_locks.map)
+        } else {
+            // If `valid` is false, the locks may have been proposed but not committed.
+            // So, it is not safe to use a Raft message to transfer leader in case the lock
+            // proposal fails.
+            return Ok(true);
+        };
+        drop(pessimistic_locks);
+
+        // 2. Propose pessimistic locks
```
youjiali1995, line 3750: What if there are some raft logs that will remove in-memory pessimistic locks queued before these requests? Can these requests bring locks back?

sticnarf: Yes, it will be incorrect...

I'm thinking if removing the in-memory pessimistic locks in the `proposed_cb` is enough.

sticnarf: Well, the current `proposed_cb` may be not suitable because it's not guaranteed to be called...

components/raftstore/src/store/peer.rs
```diff
@@ -3696,16 +3704,72 @@ where
...
+            // If there is no lock when setting `valid` to false, we can guarantee that no lock
+            // needs to be migrated to the new leader. In this case, we can just send a message
+            // to trigger leader transfer instead of making a proposal.
+            if pessimistic_locks.map.is_empty() {
+                return Ok(false);
+            }
+            mem::take(&mut pessimistic_locks.map)
+        } else {
+            // If `valid` is false, the locks may have been proposed but not committed.
+            // So, it is not safe to use a Raft message to transfer leader in case the lock
```
youjiali1995, line 3744: I think it's safe because the leader will wait for the transferee's log progress and all memory lock requests have been proposed to the leader.

components/raftstore/src/store/peer.rs
youjiali1995, line 3750: Is it possible that the leader fails to transfer leadership due to `leader_transfer_max_log_lag` even if you propose the transfer-leader request other than using raft message? I'm thinking can we propose memory lock requests after the check and add some actions to reduce the unavailable duration due to it.

youjiali1995: It's not easy to solve. Let us think about it.

sticnarf: I think there should be no unavailable duration. Do you mean the duration of not being able to write in-memory pessimistic locks?

components/raftstore/src/store/peer.rs
sticnarf: I'm thinking that Raft messages don't keep the order with proposals (potentially because of internal retry mechanisms). To my knowledge such a raft message can be prior to a previously proposed command...

components/raftstore/src/store/peer.rs
youjiali1995: No. The unavailable duration is caused by leadership transfer in which the leader refuses to accept proposals until the transferee catches up with the leader or timeout. If there are many memory-lock requests, it may take a while for the transferee to catch up.

components/raftstore/src/store/peer.rs
youjiali1995: > To my knowledge such a raft message can be prior to a previously proposed command

Yes, but the leader has to replicate all raft logs to the transferee before it transfers leadership to the transferee.

components/raftstore/src/store/peer.rs
sticnarf: I change the way of writing in-memory pessimistic locks in https://github.com/tikv/tikv/pull/11577/commits/72184155ecf838c6b435d4ddafe3486788622803. I think removing the locks before proposing should work.

sticnarf: The `TransferLeader` command is proposed after proposing memory locks. And the raft transfer leader is triggered after the `TransferLeader` command is applied, which is after the memory-lock logs.
So, I don't think the unavailable duration should be related to how many lock requests there are... It seems to me that the unavailable duration due to rejecting proposals is about the same as now...

youjiali1995: I have the same idea at the beginning, but I'm thinking it may lose some memory locks if fails to propose these requests that remove `removed_pessimistic_locks` due to leadership transfer. After digging into it, I think the possibility is very low because there is pre-transfer-leader.

youjiali1995: > The TransferLeader command is proposed after proposing memory locks. And the raft transfer leader is triggered after the TransferLeader command is applied, which is after the memory-lock logs.
So, I don't think the unavailable duration should be related to how many lock requests there are... It seems to me that the unavailable duration due to rejecting proposals is about the same as now...

Agree.

> Is it possible that the leader fails to transfer leadership due to leader_transfer_max_log_lag even if you propose the transfer-leader request other than using raft message?

I think the possibility that rejects to transfer leader due to `leader_transfer_max_log_lag`  is the same as now with your PR, and it's wasteful to replicate memory locks in such a case.

> I'm thinking can we propose memory lock requests after the check and add some actions to reduce the unavailable duration due to it.

My point is how about proposing memory locks after passing the pre-transfer-leader check, if so, no waste but with longer unavailable duration, so we need some actions to reduce it...

sticnarf: Yes... Locks may be lost if the scheduler thread runs much slower than the raftstore, then the write request may be proposed after transfer leader. The possibility is low, but it's good to avoid it.

sticnarf: > I think the possibility that rejects to transfer leader due to leader_transfer_max_log_lag is the same as now with your PR, and it's wasteful to replicate memory locks in such a case.

Ah... I got what you mean... I was assuming the possibility of rejecting transfer leader is low, so it will unlikely to cause waste. But it seems not always the case after another thought...

sticnarf:
@youjiali1995 I've updated the PR to solve the raised problems.

Proposing locks are moved after the log gap check.

The scheduler will keep a read lock guard until the command is sent to the raftstore router. And the extra `TransferLeader` command makes the real effective transfer leader to go another round of raftstore, so there will be no chance for the write command to fail due to leader transfer.

src/storage/txn/scheduler.rs
```diff
@@ -898,24 +879,103 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+                _ => vec![],
+            }
+        };
+        // Keep the read lock guard of the pessimistic lock table until the request is sent to the raftstore.
+        //
+        // If some in-memory pessimistic locks need to be proposed, we will propose another TransferLeader
+        // command. Then, we can guarentee even if the proposed locks don't include the locks deleted here,
+        // the response message of the transfer leader command must be later than this write command because
+        // this write command has been sent to the raftstore. Then, we don't need to worry this request will
+        // fail due to the voluntary leader transfer.
```
youjiali1995, line 955: Are these write commands sure to be proposed successfully?

src/storage/txn/scheduler.rs
```diff
@@ -898,24 +879,103 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+        // to other peers may include these deleted locks unexpectedly. Here is a example:
+        // 1. Here we propose deleting a lock (but not removing the lock from the memory first).
+        // 2. The region is going to transfer leader, so it proposes all its in-memory pessimistic
+        //    locks.
+        // 3. The follower deletes the lock first, and adds the migrated locks later.
+        // This leads to an incorrect result. Deleted locks appear again.
+        //
+        // Protected by the latches, be no concurrent write on the same keys. So, removing these
+        // locks now is safe. If the write fails, we will later recover these deleted locks
+        // before releasing the latches. We will not mistakenly lose the pessimistic locks.
```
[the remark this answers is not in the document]
sticnarf: I think it is safe if the in-memory locks are removed.

If some in-memory locks are removed before this point for any reason, such as transfer leader or splitting, we can guarantee (1) the write request will not succeed because of changed term or epoch (2) the removed locks will not be added back to the new lock table because we will check term and epoch before we add them back.

By the way, I changed in https://github.com/tikv/tikv/pull/11591 to let the pessimistic locks save the term and epoch version of the region, helping us comparing them more precisely.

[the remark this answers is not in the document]
youjiali1995: Agree, no safety issues, but I think it's possible some memory locks are lost in some cases.

sticnarf:
I change this PR to base on https://github.com/tikv/tikv/pull/11591, which should have less controversy. And some other changes done in that PR (for example, changing a single `epoch` to term and version) is synchronized to this one.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2239,17 +2239,88 @@ where
...
         }
     }
 
+    // Returns whether we should propose another TransferLeader command. This is for:
+    // 1. Considering the amount of pessimistic locks can be big, it can reduce
+    // unavailable time caused by waiting for the transferree catching up logs.
+    // 2. Make transferring leader strictly after write commands that executes
+    // before proposing the locks, preventing unexpected lock loss.
+    fn propose_locks_before_transfer_leader(&mut self) -> bool {
+        // 1. Disable in-memory pessimistic locks.
```
sticnarf, line 2278: I just find it incorrect to take away the locks here. After the locks are taken away, write requests will fail to read the locks here, while the locks are also missing from the underlying rocksdb.

I'll fix it and add tests to cover the case.

src/storage/txn/scheduler.rs
```diff
@@ -903,23 +879,101 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+        let mut pessimistic_locks_guard = txn_ext2
+            .as_ref()
+            .map(|txn_ext| txn_ext.pessimistic_locks.write());
+        let removed_pessimistic_locks = {
+            match pessimistic_locks_guard.as_mut() {
+                Some(locks) if !locks.map.is_empty() => to_be_write
+                    .modifies
+                    .iter()
+                    .filter_map(|write| match write {
+                        Modify::Put(cf, key, ..) | Modify::Delete(cf, key) if *cf == CF_LOCK => {
```
youjiali1995, line 933: I'm thinking does it more safely check term and version here.

components/raftstore/src/store/txn_ext.rs
```diff
@@ -47,7 +47,47 @@ impl fmt::Debug for TxnExt {
...
+    ///   2. Propose locks -> propose write -> transfer leader
+    ///      No lock will be lost in normal cases because the write request has been sent to the
+    ///      raftstore, it is likely to be proposed successfully, while the leader will need at
+    ///      least another round to receive the transfer leader message from the transferree.
+    ///  
+    /// - Split region
+    ///   The lock with the deleted mark SHOULD be moved to new regions on region split.
+    ///   Considering the following cases with different orders:
+    ///   1. Propose write -> propose split -> apply write -> execute split
+    ///      The write will be applied earlier than split. So, the lock will be deleted earlier
```
youjiali1995, line 72: Deleting the lock is executed in the scheduler worker pool. It may be after executing split.

components/raftstore/src/store/txn_ext.rs
```diff
@@ -47,7 +47,47 @@ impl fmt::Debug for TxnExt {
...
+    ///      than moving locks to new regions.
+    ///   2. Propose split -> propose write -> ready split -> apply write
+    ///      The write will be skipped because its version is lower than the new region. So, no
```
youjiali1995, line 75: Write proposals can't be proposed because of `CmdEpochChecker`.

components/raftstore/src/store/txn_ext.rs
sticnarf: Ah, I didn't notice it. But this should be easy to solve. We can move deleting the lock to be executed in the apply pool.

components/raftstore/src/store/txn_ext.rs
sticnarf: I was assuming `CmdEpochChecker` could be skipped if not applied to the current term....

youjiali1995: ok

src/storage/mvcc/reader/reader.rs
```diff
@@ -192,7 +192,11 @@ impl<S: EngineSnapshot> MvccReader<S> {
                 .read()
                 .map
                 .get(key)
-                .map(|lock| lock.to_lock())
+                .map(|(lock, deleted)| {
+                    // Protected by latch, deleted must be false.
```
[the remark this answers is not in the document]
sticnarf: I'm not very sure about whether it should read the deleted pessimistic lock. I'm actually inclined to yes.

For example, when a prewrite has marked the lock deleted but hasn't applied yet. If we use `MvccByKey` to read the key but skip the deleted lock, we will see there isn't any lock on it. It may be a strange state.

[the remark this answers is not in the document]
sticnarf: Of course, there is also a state when the prewrite is applied but the lock isn't deleted. But this time gap is much smaller.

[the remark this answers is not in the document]
youjiali1995: > MvccByKey and MvccByStartTs can read a deleted pessimistic lock.

I don't mean these commands should read deleted locks... My point is it can trigger `debug_assert!(!deleted)`.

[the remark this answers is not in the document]
sticnarf: I'm also thinking about checking term and version here, to solve the problem when the map is already cleared after the region becomes follower.

src/storage/mvcc/reader/reader.rs
```diff
@@ -185,15 +218,36 @@ impl<S: EngineSnapshot> MvccReader<S> {
...
-                .map
-                .get(key)
-                .map(|lock| lock.to_lock())
-        })
+    fn load_in_memory_pessimistic_lock(&self, key: &Key) -> Result<Option<Lock>> {
+        self.snapshot
+            .ext()
+            .get_txn_ext()
+            .and_then(|txn_ext| {
+                // If the term or region version has changed, do not read the lock table.
```
youjiali1995, line 226: I'm not sure if it's necessary to check it, but it's safer.

youjiali1995 (approved):

src/storage/mvcc/reader/reader.rs
sticnarf: It seems not necessary to check the version. I add both checks to become safer and consistent.

sticnarf: I need to comment out the version check part. Because we haven't maintained it well. Then, it cause unavailabilities after split and merge, even when this feature is not enabled.
The term is always updated when becoming leader, so I think the current code is safe.

sticnarf:
@youjiali1995 @longfangsong PTAL again when you're free.

A flag and special Raft message context are added to recognize the correct transfer leader response.

components/raftstore/src/store/peer.rs
```diff
@@ -4449,7 +4454,10 @@ pub trait RequestInspector {
             if apply::is_conf_change_cmd(req) {
                 return Ok(RequestPolicy::ProposeConfChange);
             }
-            if get_transfer_leader_cmd(req).is_some() {
+            if get_transfer_leader_cmd(req).is_some()
+                && !WriteBatchFlags::from_bits_truncate(req.get_header().get_flags())
+                    .contains(WriteBatchFlags::TRANSFER_LEADER_PROPOSAL)
```
youjiali1995, line 4459: Without this change, the `TransferLeader` command after memory locks proposals triggers another round of leader transfer and results in a dead loop?

youjiali1995 (approved):

components/raftstore/src/store/peer.rs
sticnarf: Previously no. Because `is_valid` is set to false, so we skipped sending another lock proposal.

But now, with https://github.com/tikv/tikv/pull/11577/commits/ab86391583e6a8b362fc30ccf783ab171c4d0081#diff-7a4718b9cddd27a4ef7e563024d5283cf14b4ba3abf26edc33b974b87680a5fbL2282 change, without `msg.get_context() != TRANSFER_LEADER_COMMAND_REPLY_CTX`, there will be a dead loop.

components/raftstore/src/store/txn_ext.rs
```diff
@@ -47,7 +47,47 @@ impl fmt::Debug for TxnExt {
 
 /// Pessimistic locks of a region peer.
 pub struct PeerPessimisticLocks {
-    pub map: HashMap<Key, PessimisticLock>,
+    /// The table that stores pessimistic locks.
```
cfzjywxk, line 50: Should this part be added in the rfc document or the test plan, it will be helpful to the test participants and to verify if all the situations are covered.

cfzjywxk (approved):

components/raftstore/src/store/txn_ext.rs
sticnarf: Ah, yes. There are quite some changes to the original design.

longfangsong (approved):

components/raftstore/src/store/peer.rs
```diff
@@ -4443,7 +4454,10 @@ pub trait RequestInspector {
             if apply::is_conf_change_cmd(req) {
                 return Ok(RequestPolicy::ProposeConfChange);
             }
-            if get_transfer_leader_cmd(req).is_some() {
+            if get_transfer_leader_cmd(req).is_some()
+                && !WriteBatchFlags::from_bits_truncate(req.get_header().get_flags())
```
BusyJay, line 4458: I don't get it. Isn't propose transfer leader after propose locks removal enough? Why need to persist the transfer leader command and apply?

sticnarf: Sorry, perhaps the explanation in the PR description is not clear enough.

I was afraid that the proposal containing all pessimistic locks is too large. That proposal may take a bit long time to apply. So, if the real transfer leader happens after the transferree applies the pessimistic locks, the gap will be much smaller. Then, the unavailable duration won't increase compared to when there're no in-memory pessimistic locks.

BusyJay: Perhaps you can use a callback instead of issuing an actual write.

PR #11615, opened 2021-12-08
cdc: load uninlined value more effectively

### LINKED ISSUE #11557: CDC fetching old values utilizes too much CPU
## Development Task

It's introduced in https://github.com/tikv/tikv/pull/11385. For a CDC incremental scan with `hint_min_ts` activated, long old values (not inlined in write CF) are fetched from default CF with `rocksdb::get` call, which utilizes too much CPU. However without `hint_min_ts` those old values are fetched with `Cursor::near_seek`, which should be a better policy.

hicqu:

Close #11557.

This PR contains these parts:
* Refactor `cdc::old_value` module.
  * `OldValueReader` is removed and methods are extracted into functions to make them easier to use.
  * Test cases in the module can cover the refactor.
* `old_value::get_old_value` returns `Result<Option>` instead of `Option` now.
  So that if RocksdB APIs return errors, CDC tasks can detect it instead of returning wrong things to downstreams.
* `flush_oldvalue_stats` won’t be called too frequently on `on_multi_batch`, which will utilizes too much CPU.
  For the current CPU usage without the PR, we can take a look at this [profiling](https://github.com/tikv/tikv/files/7646997/with-ts-filter.xml.zip).
* `ScanPolicy::fetch_old_value` is removed and the logic to handle `TsFilter` is moved into `seek_for_valid_value`.
  * so that we can avoid to load long value from CF default with some conditions.
  * This change is covered by test case `test_old_value_with_hint_min_ts`.

### Related changes

### Check List 

Tests 

- Unit test
- Manual test
  Here is the test result for the patch:
  图片
  图片

  We can know
  * `hint_min_ts` introduced in #11385 utilizes much more CPU (from 140% to 300%)
  * this patch can fix it, and reduces CDC task initialization time from 7.2min to 6.7min.

### Release note 

```release-note
None
```


components/cdc/src/delegate.rs
```diff
@@ -522,29 +524,24 @@ impl Delegate {
...
         is_one_pc: bool,
     ) -> Result<()> {
         let txn_extra_op = self.txn_extra_op;
-        let mut read_old_value = |row: &mut EventRow, read_old_ts| {
+        let mut read_old_value = |row: &mut EventRow, read_old_ts| -> Result<()> {
             if txn_extra_op == TxnExtraOp::ReadOldValue {
                 let key = Key::from_raw(&row.key).append_ts(row.start_ts.into());
-                let start = Instant::now();
-                let (old_value, statistics) = old_value_cb(key, read_old_ts, old_value_cache);
+                let old_value = old_value_cb(key, read_old_ts, old_value_cache, statistics)?;
```
overvenus, line 534: What will happen if it returns an error?

hicqu: The delegate will be deregistered.

> When will `old_value_cb` return an error?
Currently errors only come from RocksDB APIs.

overvenus (approved):
LGTM

hicqu:
@liuzix friendly ping~

liuzix (approved):

Issue #10483, opened 2021-06-29
Support online recovery for Raft majority failure


Connor1996:
## Development Task

Currently, we have a tool to recover Raft groups from majority failure:
```
tikv-ctl unsafe-recover remove-fail-stores -s <store_id, store_id> …
tikv-ctl recreate-region -r <region_id> ...
```

The first command is used to recover Raft groups of which the majority but not all replicas failed. Its mechanism is removing failed stores from membership configurations of those Raft groups.

The second command is used to recover Raft groups of which all replicas failed. Before running the command we need to find those Raft groups, which is not a documented routine. So there are no users who can do this recovery without our DBAs’ help.
Both of the 2 commands should be run on stopped TiKV instances, which makes the recovery process so long and complex that not all users can perform all steps correctly.

And there are some corner cases that could lead to inconsistent membership configuration among all replicas. If the Region splits later, this mistake will lead to inconsistent Region boundaries, which is very hard to locate and fix.
So we need to re-design and re-implement a new feature to recover a TiKV cluster from majority failure. The feature should be:
- able to be executed online
- able to avoid all inconsistencies
- a feature instead of a tool, focus on Raft majority failure, shouldn’t be used for other situations
- easy to use, at most 1 or 2 steps or command

### Specs
We must know there are no silver bullets for Raft majority failure. There must be something lost after a TiKV cluster recovers from that. Considering such integrations or consistencies:
- Committed writes  can be lost, so linear consistency of Raft can be broken;
- Some prewrites can be lost for a given transaction, so Transaction Atomicity and/or Consistency can be broken;
- Commits can be lost for a given transaction, so Transaction Consistency and/or Durability can be broken;
- For a given table, both records or indices can be lost, so Data Consistency can be broken;
- For a given transaction, it can be partially committed, so Data Integrity can be broken.

#### Stage 1: Implement basic online recovery 
- [x] Add `recover-store-failure` command of pd-ctl https://github.com/v01dstar/pd/pull/1
- [x] Add framework of unsafe recovery worker in PD  https://github.com/v01dstar/pd/pull/2
- [x] Support `DumpUnavailableRegions` command
  - [x] Change `StoreHeartbeatRequest` and `StoreHeartbeatResponse` in kvproto  https://github.com/v01dstar/kvproto/pull/1
  - [x] Dispatch it by `StoreHeartbeatResponse` in PD https://github.com/v01dstar/pd/pull/2
[...]

PR #11626, opened 2021-12-10
Stop sending store report as the result of plan execution

v01dstar:

TiKV issue #10483

What's Changed:

This PR makes TiKV stop sending report at the end of unsafe recovery plan execution as the result, since PD can use region info reported from region heartbeats to determine whether a plan has been executed or not.

### Related changes

- PR to update `tikv/pd`: https://github.com/tikv/pd/pull/4447

### Check List 

Tests 

- Unit test

Side effects

N/A

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/worker/pd.rs
+++ b/components/raftstore/src/store/worker/pd.rs
@@ -1090,18 +1090,18 @@ where
                         }
                     } else if resp.has_plan() {
                         info!("asked to execute recovery plan");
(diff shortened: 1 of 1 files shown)


Connor1996 (commented):
What we need to do is not send peer report until all these msg are handled

Connor1996:
Reimpl in https://github.com/tikv/tikv/pull/11716

Issue #11514, opened 2021-11-29
tikv-ctl doesn't handle missing parameters


kolbe:
### What version of TiKV are you using?

```
TiKV Control (tikv-ctl)
Release Version:   5.4.0-alpha
Edition:           Community
Git Commit Hash:   171ffa9179794f35432a3943785110c30a44665b
Git Commit Branch: master
UTC Build Time:    2021-11-29 22:08:48
Rust Version:      rustc 1.56.0-nightly (2faabf579 2021-07-27)
Enable Features:   jemalloc test-engines-rocksdb cloud-aws cloud-gcp
Profile:           release
```

### What operating system and CPU are you using?

```
Darwin monebase 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:24 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T8101 arm64

```

### Steps to reproduce

```
./tikv-ctl consistency-check -r 1
```

### What did you expect?

The `--host` parameter is missing, and `tikv-ctl` should inform the user of this rather than panicking.

### What did happened?

```
$ RUST_BACKTRACE=1 ./tikv-ctl consistency-check -r 1
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cmd/tikv-ctl/src/main.rs:106:29
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: tikv_ctl::new_debug_executor
   4: tikv_ctl::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

kolbe:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cmd/tikv-ctl/src/main.rs:106:29
stack backtrace:
   0:        0x100a32750 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h868959edd43e2d09
   1:        0x100648b98 - core::fmt::write::ha504d799b94a55bf
   2:        0x100a2d600 - std::io::Write::write_fmt::hdd42eaf7cc314d21
   3:        0x100a36774 - std::panicking::default_hook::{{closure}}::ha8b5ad372d004d6e
   4:        0x100a36ff0 - std::panicking::rust_panic_with_hook::h85b111d5ada04830
[...]

PR #11639, opened 2021-12-10
ctl:add prompt for ctl when required param is not specified

JayiceZ:

Close https://github.com/tikv/tikv/issues/11514

What's Changed:

### Check List 

Tests 

ref https://tikv.org/docs/dev/reference/tools/tikv-ctl/
and run some commands like
```
tikv-ctl consistency-check -r 2
tikv-ctl --host 127.0.0.1:20160 consistency-check -r 2

tikv-ctl --to-escaped 0xaaff

tikv-ctl compact-cluster 
tikv-ctl --pd 127.0.0.1:2379 compact-cluster

....
```

### Release note 

```release-note
None
```

### DIFF
--- a/cmd/tikv-ctl/src/main.rs
+++ b/cmd/tikv-ctl/src/main.rs
@@ -2016,7 +2016,7 @@ fn main() {
             }
             String::new()
         });
-        let data_dir = if data_dir.is_empty(){
+        let data_dir = if data_dir.is_empty() {
             None
         } else {
             Some(data_dir.as_str())
--- a/cmd/tikv-ctl/src/main.rs
+++ b/cmd/tikv-ctl/src/main.rs
@@ -1983,33 +1983,18 @@ fn main() {
         let key = unescape(&key);
         split_region(&pd_client, mgr, region_id, key);
     } else {
-        // Deal with all subcommands about db or host.
-        let mut contains_db = true;
-        let data_dir = opt.data_dir.unwrap_or_else(|| {
-            contains_db = false;
-            String::new()
-        });
-        let host = opt.host.unwrap_or_else(|| {
-            if !contains_db {
-                clap::Error {
-                    message: String::from("[host|data-dir] is not specified"),
-                    kind: ErrorKind::MissingRequiredArgument,
-                    info: None,
-                }
-                .exit();
+        let data_dir = opt.data_dir.as_deref();
+        let host = opt.host.as_deref();
+
+        if data_dir.is_none() && host.is_none() {
+            clap::Error {
+                message: String::from("[host|data-dir] is not specified"),
+                kind: ErrorKind::MissingRequiredArgument,
+                info: None,
             }
-            String::new()
-        });
-        let data_dir = if data_dir.is_empty() {
-            None
-        } else {
-            Some(data_dir.as_str())
-        };
-        let host = if host.is_empty() {
-            None
-        } else {
-            Some(host.as_str())
-        };
+            .exit();
+        }
+
         let skip_paranoid_checks = opt.skip_paranoid_checks;
 
         let debug_executor =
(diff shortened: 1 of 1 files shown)


cmd/tikv-ctl/src/main.rs
```diff
@@ -2090,156 +1965,315 @@ fn main() {
...
     } = cmd
     {
+        let pd = opt.pd.unwrap_or_else(|| {
+            clap::Error {
+                message: String::from("--pd is required for this command"),
+                kind: ErrorKind::MissingRequiredArgument,
+                info: None,
+            }
+            .exit();
+        });
```
sticnarf, line 1975: This seems to duplicate. Personally, I recommend extracting this out. For example:

```rust
if matches!(cmd, Cmd::CompactCluster { .. } | Cmd::SplitRegion { .. }) {
    // check pd
}
```

JayiceZ: It makes sense. I will fix it, thanks

sticnarf (approved):

cmd/tikv-ctl/src/main.rs
```diff
@@ -1955,10 +1956,160 @@ fn main() {
         return;
     }
 
-    // Deal with all subcommands needs PD.
-    if let Some(pd) = opt.pd.as_deref() {
-        let pd_client = get_pd_rpc_client(pd, Arc::clone(&mgr));
-        if let Cmd::CompactCluster {
+    if matches!(cmd, Cmd::CompactCluster { .. } | Cmd::SplitRegion { .. }) {
```
andylokandy, line 1959: Why bother to check command type and pd here rather than in the command-handling block? We can embed the `--pd` checking into the `get_pd_rpc_client` like this: `let pd_client = get_pd_rpc_client(&*opt.pd, Arc::clone(&mgr))?;`

andylokandy: If so, we can avoid repeating the command type twice.

JayiceZ: > If so, we can avoid repeating the command type twice.

Fixed.

andylokandy (approved):

cmd/tikv-ctl/src/main.rs
```diff
@@ -2090,156 +1965,299 @@ fn main() {
...
-            cfg.endpoints = pd_urls;
-            if let Err(e) = cfg.validate() {
-                panic!("invalid pd configuration: {:?}", e);
+            Some(data_dir.as_str())
+        };
+        let host = if host.is_empty() {
+            None
+        } else {
+            Some(host.as_str())
+        };
```
peng1999, line 2012: There's no need to re-bind `data_dir` and `host` here. Just use somthing like
```rust
if data_dir.is_none() && host.is_none() {
    clap::Error {
        message: String::from("[host|data-dir] is not specified"),
        kind: ErrorKind::MissingRequiredArgument,
        info: None,
    }
    .exit();
}
```

cmd/tikv-ctl/src/main.rs
```diff
@@ -1983,33 +1983,18 @@ fn main() {
...
-        let host = opt.host.unwrap_or_else(|| {
-            if !contains_db {
-                clap::Error {
-                    message: String::from("[host|data-dir] is not specified"),
-                    kind: ErrorKind::MissingRequiredArgument,
-                    info: None,
-                }
-                .exit();
+        let data_dir = opt.data_dir.as_deref();
+        let host = opt.host.as_deref();
```
andylokandy, line 1987:
```suggestion
        let data_dir = opt.data_dir.filter(|s| !s.is_empty());
        let host = opt.host.filter(|s| !s.is_empty());
```

JayiceZ: But the following method's parameters are Option<&str>, I think it is better to use `deref` here rather than modify the following methods all.

andylokandy: ok

JayiceZ:
@andylokandy @peng1999 @sticnarf  Appreciate your work on reviewing!

PR #11655, opened 2021-12-14
raftstore: move in-memory pessimistic locks to split regions

sticnarf:
- [x] Based on https://github.com/tikv/tikv/pull/11577

Refer to #11452

This PR moves the in-memory pessimistic lock in the original region to the new regions when region split finishes. 

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2715,6 +2715,19 @@ where
...
         let region_id = derived.get_id();
+
+        // Group in-memory pessimistic locks in the original region into new regions. The locks of
+        // new regions will be put into the corresponding new regions later. And the locks belonging
+        // to the old region will stay in the original map.
+        let region_locks = {
+            let mut pessimistic_locks = self.fsm.peer.txn_ext.pessimistic_locks.write();
+            // Update the version so the concurrent reader will fail due to EpochNotMatch
+            // instead of PessimisticLockNotFound.
+            pessimistic_locks.version = derived.get_region_epoch().get_version();
```
cfzjywxk, line 2726: L2736 sets the `pessimistic_locks` fields for the current peer, is this duplicated?

sticnarf: No, it is intended. Setting this version must be protected with the `RwLock` locked. Otherwise, a concurrent reader may find some lock missing while not realizing the epoch version has changed. 

This is covered by the `test_split_pessimistic_locks_with_concurrent_prewrite` test.

cfzjywxk (approved):

youjiali1995 (approved):

Issue #11651, opened 2021-12-14
Log Archiving and Rotation


ethercflow:
### Is your feature request related to a problem? Please describe:
The log behavior is different from TiDB and PD, lack of archiving function.

### Describe the feature you'd like:
Log parameters and behavior are consistent with TiDB and PD.

youjiali1995:
I have a dream that one day TiDB, TiKV and TiFlash can use the same programing language.

PR #11657, opened 2021-12-14
logger: support archiving and rotation (#11651)

ethercflow:
close #11651

Close #11651  

What's Changed:

Add a runner to handle archive request

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- No

### Release note 

```release-note
logger: support archiving and rotation
```


ethercflow:

components/tikv_util/src/logger/file_log.rs
```diff
@@ -83,6 +94,9 @@ impl RotatingFileLoggerBuilder {
 
     pub fn build(mut self) -> io::Result<RotatingFileLogger> {
         let file = open_log_file(&self.path)?;
+        let mut worker = LazyWorker::new("archive-worker");
+        assert!(worker.start(Runner::new(&self.path, self.max_backups, self.max_days)));
+        worker.scheduler().schedule(Task::Archive).unwrap();
```
5kbpers, line 99: Why schedule a task here?

ethercflow: I think there may be such a scene:

User previously set not to archive. After a period of time, the user found that the number of rotated files was too many, so they changed the archiving rules and restarted TiKV. If we don't schedule a task here, we can only perform the archiving operation when the rotate is triggered next time, which may be a bit long.

components/server/src/setup.rs
```diff
@@ -35,11 +36,22 @@ macro_rules! fatal {
...
-    new_path.push(format!(
-        ".{}",
-        Local::now().format(logger::DATETIME_ROTATE_SUFFIX)
-    ));
+    let parent = path.parent().unwrap().as_os_str();
+    let prefix = path.file_stem().unwrap();
+    let mut new_path = OsString::from(parent);
+    new_path.push("/");
+    new_path.push(prefix);
+    let dt = Local::now().to_rfc3339_opts(SecondsFormat::Millis, false);
```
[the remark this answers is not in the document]
ethercflow: Added, thanks!

components/server/src/setup.rs
```diff
@@ -34,12 +35,25 @@ macro_rules! fatal {
...
+    let mut dt = dt.chars();
+    // remove zone
+    for _ in 0..6 {
+        dt.next_back();
```
5kbpers, line 50: How about using `split_at` here?

components/server/src/setup.rs
```diff
@@ -211,11 +241,13 @@ pub fn initial_metric(cfg: &MetricConfig) {
 #[allow(dead_code)]
 pub fn overwrite_config_with_cmd_args(config: &mut TiKvConfig, matches: &ArgMatches<'_>) {
     if let Some(level) = matches.value_of("log-level") {
-        config.log_level = logger::get_level_by_string(level).unwrap();
+        config.log.level = logger::get_level_by_string(level).unwrap();
+        config.log_level = slog::Level::Info;
```
5kbpers, line 245: How about keeping these parameters changed for backward compatibility?

src/config.rs
```diff
@@ -2797,6 +2856,56 @@ impl TiKvConfig {
...
+
+        let default_tikv_cfg = TiKvConfig::default();
+        let default_log_cfg = LogConfig::default();
+        if self.log_level != default_tikv_cfg.log_level {
+            warn!("deprecated configuration, log-level has been moved to log.level");
+            if self.log.level == default_log_cfg.level {
+                warn!("override log.level with log-level, {:?}", self.log_level);
+                self.log.level = self.log_level;
+            }
+            self.log_level = default_tikv_cfg.log_level;
```
5kbpers, line 2868: (ditto)

components/server/src/setup.rs
ethercflow: Much better, fixed thanks!

components/tikv_util/src/logger/file_log.rs
```diff
@@ -230,12 +193,117 @@ impl Rotator for RotateBySize {
...
+        assert!(path.pop());
+        Runner {
+            log_dir: path,
+            file_name,
+            max_backups,
+            max_days,
+        }
+    }
+
+    fn dt_from_file_name(&self, path: &Path) -> Option<DateTime<Local>> {
```
5kbpers, line 231: Better to add a unit test for this function.

5kbpers: And seems this function could be defined as an ordinary function, not a member function

ethercflow: Agree, changed to this, and added a unit test. Thanks!

5kbpers (approved):

components/tikv_util/src/logger/file_log.rs
```diff
@@ -230,12 +193,117 @@ impl Rotator for RotateBySize {
     }
 }
 
+pub enum Task {
+    Archive,
+}
+
+impl Display for Task {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "")
```
youjiali1995, line 202: You can adds `#[derive(Debug)]` attribute to `Task` and use it.
```rust
write!(f, "{:?}", self)
```

components/server/src/setup.rs
```diff
@@ -34,12 +35,22 @@ macro_rules! fatal {
...
-    ));
+    let parent = path.parent().unwrap().as_os_str();
+    let prefix = path.file_stem().unwrap();
+    let mut new_path = OsString::from(parent);
+    new_path.push("/");
+    new_path.push(prefix);
+    let dt = Local::now().to_rfc3339_opts(SecondsFormat::Millis, false);
+    // remove zone
+    let (dt, _) = dt.split_at(dt.len() - 6);
+    new_path.push(format!("-{}", dt.replace(":", "-")));
```
sticnarf, line 49: What about `Local::now().format("%Y-%m-%dT%H-%M-%S%.3f")`?

components/server/src/setup.rs
```diff
@@ -34,12 +35,22 @@ macro_rules! fatal {
...
-    let mut new_path = path.to_path_buf().into_os_string();
-    new_path.push(format!(
-        ".{}",
-        Local::now().format(logger::DATETIME_ROTATE_SUFFIX)
-    ));
+    let parent = path.parent().unwrap().as_os_str();
+    let prefix = path.file_stem().unwrap();
+    let mut new_path = OsString::from(parent);
+    new_path.push("/");
+    new_path.push(prefix);
```
sticnarf, line 45: I prefer to create a `PathBuf` first and then `push` the new filename. If the original log is `/x.log`, then you will create a path beginning with `//` (although this is still valid...)

components/server/src/setup.rs
```diff
@@ -34,12 +35,22 @@ macro_rules! fatal {
 // TODO: There is a very small chance that duplicate files will be generated if there are
 // a lot of logs written in a very short time. Consider rename the rotated file with a version
 // number while rotate by size.
+//
+// The file name format after rotated is as follows: tikv-2021-12-21T00-19-38.378.log
```
youjiali1995, line 39: The format is not unified with comments above `RotatingFileLogger`
> /// After rotating, the original log file would be renamed to "{original name}.{%Y-%m-%d-%H:%M:%S}".

components/server/src/setup.rs
sticnarf: If you mean to set it to the default value, I think it's better to write ``TiKvConfig::default().log_level` here?

etc/config-template.toml
```diff
@@ -63,6 +42,25 @@
...
+## Enable automatic timestamps in log output, if not set, it will be defaulted to true.
+# enable-timestamp = true
+
+[log.file]
+## Usually it is set through command line.
+# filename = ""
+## max log file size in MB (upper limit to 4096MB)
+# max-size = "300MB"
+## max log file keep days
+# max-days = "0d"
```
youjiali1995, line 60: "0d" means no clean up according to log's timestamp?

components/server/src/setup.rs
ethercflow: Changed to this, thanks!

etc/config-template.toml
```diff
@@ -63,6 +42,25 @@
...
+## maximum number of old log files to retain
+# max-backups = 0
```
youjiali1995, line 62: "0" means no clean up accroding to log count?

ethercflow: Yes

etc/config-template.toml
ethercflow: Yes

components/server/src/setup.rs
ethercflow: Fixed, thanks!

components/tikv_util/src/logger/file_log.rs
ethercflow: Fixed, thanks!

components/server/src/setup.rs
ethercflow: Fixed thanks!

youjiali1995 (approved):

components/tikv_util/src/logger/file_log.rs
```diff
@@ -230,12 +193,118 @@ impl Rotator for RotateBySize {
...
+    if let Some(offset) = dt.find('T') {
+        unsafe {
+            // -2021-12-13T16-08-27.621 => -2021-12-13T16:08:27.621
+            let dt = dt.as_bytes_mut();
+            if dt.len() < offset + 6 {
+                return None;
+            }
+            dt[offset + 3] = b':';
+            dt[offset + 6] = b':';
+        };
```
sticnarf, line 228: The code extracting time looks unsound. It can make the program panic in L230 if you modify the some bytes to `:` causing the string invalid UTF-8.

A safe way is changing it to raw bytes, and then convert it to string again...

components/tikv_util/src/logger/file_log.rs
```diff
@@ -230,12 +193,118 @@ impl Rotator for RotateBySize {
...
+        dt.push_str("+00:00");
```
sticnarf, line 229: The timezone seems not correct. The file is created using the local timezone, but always use UTC+0 here.

components/tikv_util/src/logger/file_log.rs
```diff
@@ -230,12 +193,118 @@ impl Rotator for RotateBySize {
...
+            while let Some(log) = logs.last() {
+                if log.dt < threshold {
+                    remove.push(logs.pop().unwrap());
+                } else {
+                    break;
+                }
+            }
+        }
+        for log in remove {
+            fs::remove_file(log.f.path()).unwrap();
```
sticnarf, line 296: I think we don't need to panic here. Instead, we can log the error.

sticnarf (commented):
The rest lgtm.

components/tikv_util/src/logger/file_log.rs
sticnarf: Actually, I think we can use the `parse_from_str` method provided by `chrono` directly.

components/tikv_util/src/logger/file_log.rs
ethercflow: This is intentional, because the time zone has been removed when the file name is generated, and it is added here only for time conversion. When we compare time, we use the +00:00 time zone uniformly

sticnarf: OK, this is not a big problem. I was thinking that using the local timezone might be better.

components/tikv_util/src/logger/file_log.rs
ethercflow: Sure, fixed, thanks!

components/tikv_util/src/logger/file_log.rs
ethercflow: Changed to use the correct time zone, I also think it’s better

components/tikv_util/src/logger/file_log.rs
ethercflow: You are right, it has been modified, much better than my own parser, thanks!

sticnarf (approved):

PR #11658, opened 2021-12-14
ctl: split into multiple modules

andylokandy:
Signed-off-by: andylokandy <andylokandy@hotmail.com>

- Split `tikv-ctl/src/main.rs` into multiple files. (pure copy-and-paste)
- Turn `if let` style command pattern match into `match`. (tikv-ctl/src/main.rs)
- ~Filter out tombstone region on tikv-ctl getting all regions' id. (src/server/debug.rs)~
- Skip region on `tikv-ctl unsafe-recovery drop-unapplied-raftlog` if no raftlog to drop. (src/server/debug.rs)

What's Changed:

### Release note 

```release-note
none
```


andylokandy:
cc @peng1999

src/server/debug.rs
```diff
@@ -142,12 +142,17 @@ impl<ER: RaftEngine> Debugger<ER> {
...
         let mut regions = Vec::with_capacity(128);
-        box_try!(db.scan_cf(cf, start_key, end_key, false, |key, _| {
+        box_try!(db.scan_cf(cf, start_key, end_key, false, |key, value| {
             let (id, suffix) = box_try!(keys::decode_region_meta_key(key));
-            if suffix != keys::REGION_STATE_SUFFIX {
-                return Ok(true);
+            if suffix == keys::REGION_STATE_SUFFIX {
+                let mut region_state = RegionLocalState::default();
+                region_state.merge_from_bytes(value)?;
+                if region_state.get_state() == PeerState::Tombstone {
```
BusyJay, line 150: Then how to know how many tombstone states in the DB?

src/server/debug.rs
```diff
@@ -682,8 +687,13 @@ impl<ER: RaftEngine> Debugger<ER> {
             })?;
 
             let applied_index = old_raft_apply_state.applied_index;
+            let commit_index = old_raft_apply_state.commit_index;
             let last_index = old_raft_local_state.last_index;
 
+            if last_index == applied_index && commit_index == applied_index {
```
BusyJay, line 693: Why the change?

src/server/debug.rs
andylokandy: Currently, it's impossible to know whether a region returned from this API is tombstoned. So it brings confusion. When do you need to know how many tombstone states in the DB?

src/server/debug.rs
andylokandy: Because if so the API will not change any state but will print too many unnecessary logs.

src/server/debug.rs
BusyJay: I believe you can use `tikv-ctl raft region -r id` to know if a region is tombstoned. I don't have a practical case for now, but it should be useful when it needs to analyze historical information.

src/server/debug.rs
BusyJay: Oh, sorry, I misread. Why check for `commit_index == applied_index`?

src/server/debug.rs
andylokandy: Reasonable. But just to mention, `tikv-ctl raft region -r id` doesn't show the peer state so far.

src/server/debug.rs
andylokandy: Because the code below will modify `commit_index` to be equal `applied_index`

zhongzc (approved):

cmd/tikv-ctl/src/main.rs
```diff
@@ -1959,327 +176,310 @@ fn main() {
...
-                    let resp = client.list_fail_points_opt(&list_req, option).unwrap();
-                    println!("{:?}", resp.get_entries());
+                Cmd::Store {} => {
+                    debug_executor.dump_store_info();
+                }
+                Cmd::Cluster {} => {
+                    debug_executor.dump_cluster_info();
+                }
+                _ => {
+                    Opt::clap().print_help().ok();
```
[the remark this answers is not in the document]
andylokandy: What if the user don't give any command?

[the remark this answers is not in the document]
andylokandy: Ok I get it.

cmd/tikv-ctl/src/util.rs
```diff
@@ -0,0 +1,78 @@
...
+use std::error::Error;
+use std::str::FromStr;
+use std::{process, str, u64};
+
+use tikv::config::TiKvConfig;
+
+pub fn init_ctl_logger(level: &str) {
+    let mut cfg = TiKvConfig::default();
+    cfg.log_level = slog::Level::from_str(level).unwrap();
+    cfg.rocksdb.info_log_dir = "./ctl-engine-info-log".to_owned();
```
iosmanthus, line 14: Could we change this path to constant or refactor it as a parameter?

iosmanthus (approved):
LGTM

PR #11659, opened 2021-12-15
raftstore:  support multi-files' snapshot

### LINKED ISSUE #11595: Multiple file snapshot support
If the region size is very large, put all data inside a file may introduce a large compaction. It's better to split a large snapshot into several small files.

tonyxuqqi:

Issue Number: close #11595

What's Changed:

- Before this PR, TiKV's only generates single SST file per cf for snapshot. This does not work well when region size is huge (such as 10GB) as large SST could lead to larger write amplification in compactions. This change would split the large snapshot into multiple files according to the sst's raw size (before compression). The size is controlled by config: server.max_snapshot_file_size. For backward compatibility consideration, the feature is only enabled in V6.1.0 or newer.  That means it will still send one snapshot file when the whole cluster version is older than v6.1.0.

- Make snapshot GC execution in the background thread

### Related changes
None

### Check List 

Tests 

- Unit test
- Integration test
   Deploy older version and update one tikv with this version, the snapshot feature still works as expected in both old and new version's tikv.  And then upgrade all tikv nodes to new version and set the max snapshot file size to 50MB, the snapshot feature still works and multiple snapshot files can be observed. 

Side effects
- Breaking backward compatibility (only when incorrect config and rollback to older version, the new versions' snapshot files cannot be deleted by older version's code)

### Release note 

```release-note
TiKv supports multi-files' snapshot for large snapshot. It's only enabled in v6.1.0 or after for now. 
```


components/server/src/server.rs
```diff
@@ -662,10 +664,20 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         let bps = i64::try_from(self.config.server.snap_max_write_bytes_per_sec.0)
             .unwrap_or_else(|_| fatal!("snap_max_write_bytes_per_sec > i64::max_value"));
 
+        let max_snapshot_file_raw_size = if self
```
BusyJay, line 667: This will requires restart twice to make every node generate snapshot in multiple files.

tonyxuqqi: good point. fixed.

tonyxuqqi: > This will requires restart twice to make every node generate snapshot in multiple files.

good point. fixed.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2213,86 +2222,24 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
...
-
     fn on_snap_mgr_gc(&mut self) {
-        if let Err(e) = self.handle_snap_mgr_gc() {
-            error!(?e;
-                "handle gc snap failed";
+        // refresh multi_snapshot_files enable flag
+        self.ctx.snap_mgr.set_enable_multi_snapshot_files(
+            self.ctx
+                .feature_gate
+                .can_enable(MULTI_FILES_SNAPSHOT_FEATURE),
```
5kbpers, line 2230: Maybe we need a test for this feature gate.

components/raftstore/src/store/worker/cleanup_snapshot.rs
```diff
@@ -0,0 +1,176 @@
+// Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: 2018 -> 2021

tonyxuqqi: > 2018 -> 2021

good catch..

components/raftstore/src/store/fsm/store.rs
tonyxuqqi: > Maybe we need a test for this feature gate.

Good point. Added an integration test.

5kbpers (approved):
LGTM

components/raftstore/src/store/snap.rs
```diff
@@ -465,35 +584,43 @@ impl Snapshot {
...
-                    file_for_recving.encrypter = Some(
-                        create_aes_ctr_crypter(
-                            mthd,
-                            &enc_info.key,
-                            Mode::Encrypt,
-                            Iv::from_slice(&enc_info.iv)?,
-                        )
-                        .map_err(|e| RaftStoreError::Snapshot(box_err!(e)))?,
-                    );
+            cf_file.file_for_recving = vec![];
```
hicqu, line 590: Is the line necessary?

components/raftstore/src/store/snap.rs
```diff
@@ -742,24 +920,78 @@ impl Snapshot {
...
+        macro_rules! try_delete_snapshot_files {
+            ($cf_file: ident, $file_name_func: ident) => {
+                let mut file_id = 0;
+                loop {
+                    let file_path = $cf_file.path.join($cf_file.$file_name_func(file_id));
+                    if file_exists(&file_path) {
+                        delete_file_if_exist(&file_path).unwrap();
+                        file_id += 1;
+                    } else {
+                        break;
```
hicqu, line 932: What if the tikv-server instance is killed after some SSTs are deleted? How about keeping try to delete more files?

components/raftstore/src/store/snap.rs
```diff
@@ -1072,19 +1337,21 @@ struct SnapManagerCore {
...
     encryption_key_manager: Option<Arc<DataKeyManager>>,
+    max_per_file_size: Arc<AtomicU64>,
+    enable_multi_snapshot_files: Arc<AtomicBool>,
 }
 
 /// `SnapManagerCore` trace all current processing snapshots.
 pub struct SnapManager {
     core: SnapManagerCore,
-    max_total_size: AtomicU64,
+    max_total_size: Arc<AtomicU64>,
```
hicqu, line 1347: Why `Arc`? Is it a bug?

components/raftstore/src/store/snap.rs
tonyxuqqi: Currently the delete() is on critical path of fsm. Trying to delete more files (such as hard coded 100 files) will significantly add IOs and slow the fsm processing. And even with that, this is still the best effort style. 
Once the snapshot gc is moved to background thread, then we can be more aggressive here. 

Practically the code path should work fine because in most cases, the snapshot files of one CF is only 1.

components/raftstore/src/store/snap.rs
tonyxuqqi: Because the SnapshotManager can have several instances and any instance changes the max_total_size should impact all other instances.

hicqu (approved):

PR #11661, opened 2021-12-15
coprocesser: refactor encoding and add gbk_chinese_ci map

### LINKED ISSUE #11660: coprocesser: it will report error if use gbk_chinese_ci

xiongjiwei:

Close https://github.com/tikv/tikv/issues/11660

- refactor encoding and add gbk_chinese_ci map
- add ASCII and binary encoding

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/tidb_query_datatype/src/codec/collation/encoding/ascii.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+#[derive(Debug)]
+pub struct EncodingAscii;
+
+impl Encoding for EncodingAscii {
+    #[inline]
+    fn decode(data: BytesRef) -> Result<Bytes> {
+        for x in data {
+            if *x >= 0x80_u8 {
```
zhongzc, line 22:
```suggestion
            if x.is_ascii() {
```

xiongjiwei: updated

components/tidb_query_datatype/src/def/field_type.rs
```diff
@@ -145,20 +146,25 @@ impl fmt::Display for Collation {
...
 
-#[derive(Debug, Clone, Copy)]
+#[derive(PartialEq, Debug, Clone, Copy)]
 pub enum Charset {
     UTF8,
     UTF8Mb4,
     Latin1,
     GBK,
+    Binary,
+    Ascii,
```
zhongzc, line 156: Should  `Collation` at L107-116 be updated?

xiongjiwei: nope, this two is the missing charset, just add them here is fine

components/tidb_query_datatype/src/codec/collation/encoding/utf8.rs
```diff
@@ -2,20 +2,37 @@
...
+impl<T: UTF8CompatibleEncoding> Encoding for T {
     #[inline]
     fn decode(data: BytesRef) -> Result<Bytes> {
         match str::from_utf8(data) {
             Ok(v) => Ok(Bytes::from(v)),
-            Err(_) => Err(Error::cannot_convert_string("utf8mb4")),
+            Err(_) => Err(Error::cannot_convert_string(T::NAME)),
         }
     }
+}
```
andylokandy, line 17: Why add this trait rather than impl `Encoding` on all of them?

andylokandy: I think it can't save too much codes.

zhongzc:
I don't know if test cases cover here's changes. Please take a look. @andylokandy

components/tidb_query_datatype/src/codec/collation/encoding/utf8.rs
xiongjiwei: `EncodingUTF8Mb4`, `EncodingUTF8` and `EncodingLatin1` share the same `decode` implemention, only different at the name. 

Sure it saves little codes, but it is not hard to understand and only have the file scope

andylokandy (approved):

zhongzc (approved):

Issue #10938, opened 2021-09-13
Implement API V2


andylokandy:
API V2 brings a bunch of new features for TiKV, including keyspace, mix usage of RawKV and TxnKV, enhancement on TTL and RawCAS.

See the [RFC](https://github.com/tikv/rfcs/pull/69) for details.

## Tasklist :spiral_calendar:

- [ ] Add configuration item `server.api-version` and store the config in kvdb
- [ ] Add `api_version` in `import_sstpb.SSTMeta` and `kvrpckv.Context` in kvproto
- [ ] Add new key encode structure for API V2
- [ ] Add new RawKV value encode for API V2 in `RawStore` and `TTLStore`
- [ ] Add API V2 key-value encode support in `sst_importer`
- [ ] Add data check during `server.api-version` change during startup
- [ ] Add API V2 check for RawKV requests and deprecate `for_cas` in `RawPut` and `RawDelete` request
- [ ] Add API V2 check for TxnKV requests

Support API V2 on clients:
- [ ] Java client
- [ ] Rust client
- [ ] Go client
- [ ] Python client
- [ ] Node client
- [ ] C++ client

PR #11664, opened 2021-12-15
sst_importer: check api version when importing

peng1999:

Part of #10938 

What's Changed:

When importing SST file using ingest API, the version in `SstMeta` is checked if is compatible to current tikv's api version.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


andylokandy (approved):

components/sst_importer/src/import_file.rs
```diff
@@ -281,14 +285,77 @@ impl ImportDir {
...
+                _ => {
+                    let path = self.join(meta)?;
+                    let path_str = path.save.to_str().unwrap();
+                    let env = get_env(key_manager.clone(), get_io_rate_limiter())?;
+                    let sst_reader = RocksSstReader::open_with_env(path_str, Some(env))?;
+
+                    for &(start, end) in TIDB_RANGES_COMPLEMENT {
+                        let mut unexpected_data_key = None;
+                        sst_reader.scan(start, end, false, |key, _| {
+                            unexpected_data_key = Some(key[DATA_KEY_PREFIX_LEN..].to_vec());
```
zhongzc, line 313: Why needs `DATA_KEY_PREFIX_LEN` here?

andylokandy: It's because we don't want the `z` prefix to show up in the user key when printing the error to user.

zhongzc: I looked into `TIDB_RANGES_COMPLEMENT`, and it doesn't have `z` prefix.
So it means that, as arguments, keys won't contain the prefix, but the iterator will meet the prefix?

peng1999: `TIDB_RANGES_COMPLEMENT` do contains keys having `z` prefix. It have range `(&[b't' + 1], &[])`, which means [`'u'`, +∞).

peng1999: OK. The `[DATA_KEY_PREFIX_LEN..]` have been removed.

zhongzc (approved):
Please fix the lint, Rest LGTMA

PR #11665, opened 2021-12-15
update rust toolchain version and crates version

LemonHX:
update rust toolchain to latest,
fix compiling errors.

Signed-off-by: lemonhx <lemonhx@lemonhx.tech>

Close #xxx 

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
none
```


components/encryption/src/io.rs
```diff
@@ -387,8 +387,9 @@ impl CrypterCore {
 
     fn reset_buffer(&mut self, size: usize) {
         // OCrypter require the output buffer to have block_size extra bytes, or it will panic.
-        self.buffer.reserve(size + self.block_size);
+        #[allow(clippy::uninit_vec)]
```
BusyJay, line 390: I don't think it's a safe implementation. /cc @hunterlxt

components/external_storage/export/src/export.rs
```diff
@@ -182,6 +182,7 @@ fn create_backend_inner(
         },
         #[cfg(not(any(feature = "cloud-gcp", feature = "cloud-aws")))]
         _ => return Err(bad_backend(backend.clone())),
+        Backend::AzureBlobStorage(_) => todo!(),
```
BusyJay, line 185: We use `unimplemented!` in TiKV.

BusyJay (commented):
You should keep the crates version untouched. We don't upgrade dependencies unless necessary.

components/encryption/src/io.rs
sticnarf: Agree... Strictly speaking, every slice must be initialized. The problem is that the interface provided by the openssl library accepts a `& mut [u8]`. Passing an uninitialized buffer is UB but should work correctly in real world...

components/external_storage/export/src/export.rs
LemonHX: ok get it

components/encryption/src/io.rs
tabokie: This is an output buffer, and uninit-ed vec is safe as long as it's not read before written to. Still, I'm not against using `resize` here, the perf difference should be negligible.

peng1999: Using `Vec::set_len` like this is not panic-safe and should be avoided. The data should be written *before* the `Vec::set_len` call. See https://github.com/rust-lang/rust-clippy/issues/4483

tabokie: I understand, but panic safety is only a blocker if we are storing value with non-trivial `Drop` implementation, which isn't the case here.

sticnarf:
Hi, @LemonHX. Do you have time to continue this pull request?

sticnarf:
Close due to inactivity.

PR #10979, opened 2021-09-24
file_system: support collect the io flow by proc

MuZhou233:
### What problem does this PR solve?

Issue Number: task of #10867 

### What is changed and how it works?

What's Changed:

- Add a new file which collect the io flow from `/proc/self/task/<tid>/io`

### Check List 

Tests 

- Unit test
- Manual test (add detailed scripts or steps below)
- Bench
```
running 1 test
test thread_io::tests::bench_fetch_io_bytes      ... bench:       5,253 ns/iter (+/- 519)
```

Issue #10867, opened 2021-08-31
Support limiting read flow for TiKV IO rate limiter


Connor1996:
## Description

Before, we added an [io snooper supported by BPF](https://github.com/tikv/tikv/pull/9153) to get the actual read flow categorized by threads. With the limitation of BPF that needs sudo permission, it isn't used in production actually.

Actually, [proc](https://man7.org/linux/man-pages/man5/proc.5.html) provides a simpler way to get the actual read flow categorized by threads by `/proc/<pid>/task/<tid>/io/read_bytes`.

## Document Collection

- Weekly report: ${weekly report link}

## Talent Challenge Program information

- Mentor of this issue: @Connor1996 
- Recommended skills: Rust, Linux
- Estimated Workloads: 1XL(XL, man-month)

## Milestones and action items

Milestone 1: Expected finish date: ${0.5XL}

- [ ] Collect the actual read flow by `proc` https://github.com/tikv/tikv/pull/11667
- [ ] Remove the entrance of enabling io snooper  https://github.com/tikv/tikv/pull/11394
- [ ] Adjust grafana template
- [ ] Fix https://github.com/tikv/tikv/issues/10659

Milestone 2: Expected finish date: ${0.5XL}

- [ ] Test

PR #11667, opened 2021-12-15
file_system: support collect the io flow by proc

MuZhou233:

Issue Number: task of #10867 
old Pull Request: #10979

What's Changed:

- Add a new file which collect the io flow from `/proc/self/task/<tid>/io`

### Check List 

Tests 

- Unit test
- Bench
```
running 1 test
test thread_io::tests::bench_fetch_io_bytes      ... bench:       5,253 ns/iter (+/- 519)
```

### Release note 

```release-note
Support collect the io flow by proc
```


Connor1996 (approved):
LGTM

Connor1996 (commented):
add a reference to the old PR

tabokie (approved):

components/file_system/src/thread_io.rs
```diff
@@ -0,0 +1,258 @@
...
+
+pub fn get_io_type() -> IOType {
+    THREAD_IO_PRIVATE.with(|private| private.get().io_type)
+}
+
+fn fetch_thread_io_bytes(_io_type: IOType) -> IOBytes {
+    fetch_exact_thread_io_bytes(&ThreadID::current())
+}
+
+fn fetch_exact_thread_io_bytes(id: &ThreadID) -> IOBytes {
```
hicqu, line 142: How about reuse `ThreadInspector` in `components/tikv_util/src/sys/inspector.rs`?

tabokie: I agree. The merged version needs to stay at "file_system" though, to avoid circular dependency.

components/file_system/src/thread_io.rs
```diff
@@ -0,0 +1,258 @@
...
+use std::io::{BufRead, BufReader};
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc, Mutex, MutexGuard,
+};
+use std::time::Duration;
+
+use nix::unistd::{getpid, gettid, Pid};
```
sticnarf, line 15: `gettid` is not available on `macOS`. So, this PR breaks our builds on macOS. Could you please switch it to `thread-id` to finish it?

Connor1996: @MuZhou233 PTAL

MuZhou233: As I know, `thread-id` is actually `pthread id` on linux, Which is different from `tid` here. And `macOS` didn't have `/proc` for us to read.
I think it's better to mark this as linux only. But how do I deal with other target os? @Connor1996

Connor1996: let's target it only for linux

components/file_system/src/thread_io.rs
MuZhou233: It seems `ThreadInspector` can only get `io_stat` of **current** thread now. But we need function to get `io_stat` of other thread here.

components/file_system/src/thread_io.rs
sticnarf: @MuZhou233 Will you try to fix it by limiting the platform? Or I can do it for you.

MuZhou233: @sticnarf I will open a new PR to fix it tonight.

PR #11693, opened 2021-12-21
status_server: Support online config update for configuration hosting platform

### LINKED ISSUE #11692: Support online config update for configuration hosting platform
## Development Task

Cloud platforms such as K8s will perform configuration management, and TiKV does not have write permissions on these platforms. Therefore, an interface needs to be added to support this situation.

ethercflow:
… platform (#11692)

close #11692

Close #11692  

What's Changed:  
- add a new API in status_server

### Related changes

- None

### Check List 

Tests 

- Manual test 
   1. Setup tikv-server
   2. run curl -X PUT

Side effects

No

### Release note 

```release-note
None.
```


ethercflow:

src/config.rs
```diff
@@ -3231,6 +3231,21 @@ impl ConfigController {
...
+    }
+
+    pub fn update_from_toml_file(&self) -> CfgResult<()> {
+        let current = self.get_current();
+        let incoming = TiKvConfig::from_file(Path::new(&current.cfg_path), None);
+        let diff = current.diff(&incoming);
+        self._update(diff, None)
+    }
+
+    fn _update(
```
5kbpers, line 3244: Prefer to name it `update_impl`. I think underscore prefixed function names is not commonly used in Rust

ethercflow: Fixed, thanks!

5kbpers (approved):
LGTM

src/config.rs
```diff
@@ -3231,6 +3231,21 @@ impl ConfigController {
 
     pub fn update(&self, change: HashMap<String, String>) -> CfgResult<()> {
         let diff = to_config_change(change.clone())?;
+        self.update_impl(diff, Some(change))
+    }
+
+    pub fn update_from_toml_file(&self) -> CfgResult<()> {
```
NingLin-P, line 3237: Plz add some test cases for this method

src/config.rs
```diff
@@ -3231,6 +3231,21 @@ impl ConfigController {
...
+        let current = self.get_current();
+        let incoming = TiKvConfig::from_file(Path::new(&current.cfg_path), None);
```
NingLin-P, line 3239: `TiKvConfig::from_file` will panic if there is no such file, better to change it to return an `Err` and we print a log here instead.

ethercflow: Fixed, thanks!

src/config.rs
ethercflow: Added, thanks!

NingLin-P (approved):
LGTM

Issue #11000, opened 2021-09-28
Find better default thread numbers for backup and import


YuJuncen:
## Feature Request
Currently, we use 75% of logical CPU cores for backup and 8 threads for importing SST.

1. For backup, with so many CPU resource used, backup would probably **impact the online cluster**.
2. For importing, we can utilize nearly all the resources in the cluster, but **the constant 8 is too small** for some conditions.

### Is your feature request related to a problem? Please describe:
An internal test shows, for a write-only workload, the impact to the cluster is relative to the CPU used by BR.
![image](https://user-images.githubusercontent.com/36239017/135052672-018dc2c9-6549-439f-95f3-c217f1c0ec68.png)
> The green line is QPS.
> Chinese text in the graph, from left to right: 
> 1. Only read the disk.
> 2. Run BR, the red line is CPU-cost, and the blue line is the CPU used by backup worker
> 3. Run empty for loops. The orange line is the CPU costed by the empty for loops.

... Even the CPU isn't full. The cluster would still be impacted by the CPU usage.

The test also shows it seems the bottleneck isn't the disk:
<img width="1869" alt="image" src="https://user-images.githubusercontent.com/36239017/135054206-f3b9e249-f5d4-4bee-a381-966f9cda58f8.png">
> The gold line is disk usage, blue line is BR CPU usage, orange line is total CPU usage, red line is empty loop CPU usage, green line is QPS.

### Describe the feature you'd like:
Choose a better percentage of default thread number to the total CPU.
Modify the default import thread number by the resource of the node.

### Describe alternatives you've considered:
Fix the problem throughput down by CPU usage even the CPU isn't full. Or implement some cheaper backup / restore method, say, copy SST files directly.

### Teachability, Documentation, Adoption, Migration Strategy:
We can avoid teaching the user how to modify the config file before importing / backing up.

YuJuncen:
The available resource for backing up would be vary with workload, e.g. Consider a heavy workload which occupies 60% of total CPU resource and a light workload (say, in midnight) which only uses 10% of total CPU, the best thread number for backup for each of them is indeed different.

As a solution, instead of let ops to decide the best value via experience, we can detect and auto tune the backup threads via a  soft limit. Briefly, it was designed like this:

1. We make a "token bucket" for backup tasks, with some traits:
    a. This token bucket wouldn't generate token automatically.
    b. When each backup worker wants to backup a `BackupRange`, it must acquire a token from the bucket, and put it back to the bucket when finished.
[...]

PR #11556, opened 2021-12-03
backup: fix background panic in `endpoint::tests::test_adjust_thread_pool_size`

YuJuncen:
Signed-off-by: yujuncen <yujuncen@pingcap.com>

### What is changed and how it works?

Close #11548

What's Changed:

See comments in #11548

### Check List 

Tests 

- Manual test (add detailed scripts or steps below)
image

### Release note 

```release-note
None
```

PR #11699, opened 2021-12-22
backup: reduce the default thread pool size of backup, enable auto tune by default

YuJuncen:

Close(again!) #11000 

What's Changed:

1. The default thread pool size of backup is now `clamp(CPU * 0.5, 1, 8)`.
2. Auto tune would clamp its permit to `[1, backup.num-threads]`.
3. Enable auto tune by default.
4. Moved `ControlThreadPool` to the `utils` package.
5. Auto tune remain threads is now `round(0.2 * CPU)`.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes less CPU.
    - Backup speed would be slower.

### Release note 

```release-note
Reduce the default thread pool size of backup, and enable backup restrict itself when the workload is heavy. The backup speed may be slower. You can set `backup.num-threads` by tikv-ctl or config file to rollback that change.
```


kennytm:
🤔 

```
[2021-12-22T10:08:06.926Z] test endpoint::tests::test_adjust_thread_pool_size ... thread 'main' panicked at 'assertion failed: `(left == right)`
[2021-12-22T10:08:06.926Z]   left: `2`,
[2021-12-22T10:08:06.926Z]  right: `3`', components/backup/src/endpoint.rs:1617:9
```

YuJuncen:
> 🤔
> 
> ```
> [2021-12-22T10:08:06.926Z] test endpoint::tests::test_adjust_thread_pool_size ... thread 'main' panicked at 'assertion failed: `(left == right)`
> [2021-12-22T10:08:06.926Z]   left: `2`,
> [2021-12-22T10:08:06.926Z]  right: `3`', components/backup/src/endpoint.rs:1617:9
> ```

This cannot be reproduced in local... Why...?🤔

/run-tests

YuJuncen:
@kennytm I guess I have got what happened.
1. When dropping the runtime, the task will be 'killed'(not being polled anymore, I guess).
2. When softlimit enabled, there may not be enough permits to make the initial 2 tasks finish.
3. So, when the runtime dropped, one of the two tasks would be canceled.

I think we shouldn't adapt the origin trick (let all tasks reference the runtime) for allowing all tasks success to finish because that causes a background panic (even it is harmless, cc https://github.com/tikv/tikv/pull/11556).

Maybe we can make the trick better? 🤔
```rust
#[derive(Clone)]
struct DaemonRuntime(Option<Runtime>);

impl DaemonRuntime {
  fn spawn(self: Arc<Self>, f: impl Future) {
    self.0.unwrap().spawn(async move {
      f.await;
      drop(self)
    })
  }
}

impl Drop for DaemonRuntime {
  fn drop(&mut self){
    // it is safe because all tasks should be finished.
    self.take().shutdown_background()
  }
}
```

YuJuncen:
(BTW, I'm not sure how can the test thread use up to 5000% CPU...)
```
[components/backup/src/softlimit.rs:126] (name, usage) = (
    "endpoint::tests",
    5000,
)
```

kennytm (approved):

3pointer (approved):

YuJuncen:
```
[2021-12-28T06:29:17.552Z]     raftstore::test_snap::test_inspected_snapshot
[2021-12-28T06:29:17.552Z]     raftstore::test_split_region::test_incompatible_node_auto_split_region
```
/run-tests

PR #11704, opened 2021-12-22
backup: return api-version to BR when backup

peng1999:

Part of #10938 

What's Changed:

Return api-version for BR when invoking backup, so that BR can write the corrent api-version info into backupmeta.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/external_storage/export/src/export.rs
```diff
@@ -180,7 +180,7 @@ fn create_backend_inner(
                 return Err(bad_backend(Backend::CloudDynamic(dyn_backend.clone())));
             }
         },
-        #[cfg(not(any(feature = "cloud-gcp", feature = "cloud-aws")))]
+        #[cfg(not(all(feature = "cloud-gcp", feature = "cloud-aws", feature = "cloud-azure")))]
```
andylokandy, line 183: Why use all?

peng1999: We can omit the catch-all match arm only if all the features are provided. Otherwise, a non exhausted error will occur.

andylokandy: Why not leave the non-exhausted branch?

peng1999: I don't know why this is in the first place. 😂 
According to the blame this `cfg` is introduced by @gregwebs in 70c7467c, and I believe the original line does not work as expected. (It will fail to compile when `cloud-gcp` is present but `cloud-aws` is not) This change in this PR is just a bugfix to make the code compile after https://github.com/pingcap/kvproto/commit/78a55fb8e69c1052a55945b69a3a571fd248bb7d breaks the code.

andylokandy: Ok, so I suggest the following:
```suggestion
        #[allow(unreachable_patterns)]
```

gregwebs: yes, this should have been `all`. Unreachable would make sense as well.

andylokandy (approved):

Little-Wallace (approved):
LGTM

PR #11705, opened 2021-12-22
raftclient: delay flush

BusyJay:

Ref #11309.

What's Changed:

#11310 introduces delay in raftstore thread to reduce the flush rate.
It can only reduce flush rate from one thread. If there are many
raftstore threads or IO threads, messages can still be flush frequently.
This PR reduces the flush rate by introducing delay in raft client. So
delay will work at connection level and achieve the maximun batch
affect.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Delay in raft client to reduce syscalls and improve CPU efficiency
```


BusyJay:
Benchmark shows that after enabling asyncio, the throughput increases about 10% ~ 35% in sysbench oltp_write_only and oltp_insert with this patch.

components/batch-system/src/config.rs
```diff
@@ -16,10 +15,6 @@ pub struct Config {
     pub reschedule_duration: ReadableDuration,
     #[online_config(skip)]
     pub low_priority_pool_size: usize,
-    #[doc(hidden)]
-    #[serde(skip)]
-    #[online_config(skip)]
-    pub before_pause_wait: Option<Duration>,
```
Connor1996, line 22: should we keep it for compatibility?

Connor1996 (commented):
rest LGTM

components/batch-system/src/config.rs
BusyJay: No, it's marked as `serde(skip)` on purpose in the past to be able to be removed without breaking compatibility.

gengliqi (commented):
Rest LGTM

components/raftstore/src/store/fsm/store.rs
```diff
@@ -908,17 +903,14 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport> PollHandler<PeerFsm<EK, ER>, St
...
-            }
+        }
+        if self.poll_ctx.trans.need_flush() {
+            self.poll_ctx.trans.flush();
+        }
+        if self.need_flush_events {
+            self.last_flush_time = TiInstant::now();
+            self.need_flush_events = false;
+            self.flush_events();
         }
```
gengliqi, line 914:
```suggestion
        } else {
            if self.poll_ctx.trans.need_flush() {
                self.poll_ctx.trans.flush();
            }
            if self.need_flush_events {
                self.last_flush_time = TiInstant::now();
                self.need_flush_events = false;
                self.flush_events();
            }
        }
```

gengliqi (approved):

NingLin-P (approved):

src/storage/mod.rs
```diff
@@ -5340,15 +5341,19 @@ mod tests {
...
             let res = block_on(storage.raw_get_key_ttl(ctx.clone(), "".to_string(), key.clone()))
+                .unwrap()
                 .unwrap();
             if ttl != 0 {
-                if ttl > u64::MAX - ttl_current_ts() {
-                    assert_eq!(res, Some(u64::MAX - ttl_current_ts()));
-                } else {
-                    assert_eq!(res, Some(ttl));
-                }
+                let lower_bound = before_written.saturating_add(ttl) - ttl_current_ts();
```
BusyJay, line 5347: /cc @Connor1996 @andylokandy

PR #11708, opened 2021-12-23
coprocessor: not return rows when there is no input for simple aggregation

### LINKED ISSUE #11735: wrong  `firstrow` aggregation result when there are regions returning empty result
## Bug Report

This is an alias for https://github.com/pingcap/tidb/issues/30923

windtalker:

Close #11735, Close pingcap/tidb#30923 

What's Changed:
For `SimpleAggregation` does not return row if there is no input rows
### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Fix wrong `any_value` result when there are regions returning empty result 
```

### DIFF
--- a/components/tidb_query_executors/src/simple_aggr_executor.rs
+++ b/components/tidb_query_executors/src/simple_aggr_executor.rs
@@ -105,7 +105,7 @@ impl<Src: BatchExecutor> BatchSimpleAggregationExecutor<Src> {
     ) -> Result<Self> {
         // Empty states is fine because it will be re-initialized later according to the content
         // in entities.
-        let aggr_impl = SimpleAggregationImpl { states: Vec::new() };
+        let aggr_impl = SimpleAggregationImpl { states: Vec::new(), has_input_rows: false };
 
         Ok(Self(AggregationExecutor::new(
             aggr_impl,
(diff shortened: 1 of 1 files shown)


sticnarf (approved):
Aggregation implementations on TiKV are always a bit different from TiDB, for example, the output of `avg()`. I think it's OK to assume at this time that simple aggr will not be the final result and return zero row.

components/tidb_query_executors/src/simple_aggr_executor.rs
```diff
@@ -203,7 +215,9 @@ impl<Src: BatchExecutor> AggregationExecutorImpl<Src> for SimpleAggregationImpl
         mut iteratee: impl FnMut(&mut Entities<Src>, &[Box<dyn AggrFunctionState>]) -> Result<()>,
     ) -> Result<Vec<LazyBatchColumn>> {
         assert!(src_is_drained);
-        iteratee(entities, &self.states)?;
+        if self.has_input_rows {
+            iteratee(entities, &self.states)?;
+        }
```
breeswish, line 220: Maybe you also need to change L207

breeswish (commented):
BTW seems that TiKV is keeping this behavior since v2.0.0, so this might be a bug exists for a lot of versions. Better to check v3.0 and v2.1 as well.

components/tidb_query_executors/src/simple_aggr_executor.rs
windtalker: This [comment](https://github.com/tikv/tikv/blob/a7e89a85bb5661f6ea211b319ea6253341a53197/components/tidb_query_executors/src/util/aggr_executor.rs#L65) indicates that `group_len` can be inaccurate since it is just a hint. But I've updated L207 as you suggested.

windtalker:
> pingcap/tidb#30923

TiDB v3.1.2/v3.0.20/v2.1.19, all of them have the same issue.

windtalker:
> BTW seems that TiKV is keeping this behavior since v2.0.0, so this might be a bug exists for a lot of versions. Better to check v3.0 and v2.1 as well.

Tested in TiDB v3.1.2/v3.0.20/v2.1.19, all of them have the same issue.

breeswish (approved):

windtalker:
@breeswish can you help merge it?

PR #11709, opened 2021-12-23
*: fix invalid failpoint caused by typo

### LINKED ISSUE #11734: Some failpoints don't work as expected
There are two fail points that don't work due to typo, it should be changed from `failpoint` to `failpoints`.

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

Close #11734  

What's Changed: change `failpoint` to `failpoints`.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


5kbpers (approved):

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3445,7 +3445,7 @@ where
             .iter()
             .any(|res| res.region_id == self.delegate.region_id())
             && self.delegate.last_flush_applied_index != applied_index;
-        #[cfg(feature = "failpoint")]
+        #[cfg(feature = "failpoints")]
```
BusyJay, line 3448: How about removing it completely? It should be optimized away in release mode.

rleungx: Do you mean remove `apply_on_handle_snapshot_sync` or just the feature?

NingLin-P (approved):

components/raftstore/src/store/fsm/apply.rs
BusyJay: Just the feature.

components/resolved_ts/src/advance.rs
```diff
@@ -6,6 +6,8 @@ use std::time::Duration;
 use collections::{HashMap, HashSet};
 use concurrency_manager::ConcurrencyManager;
 use engine_traits::KvEngine;
+#[cfg(feature = "failpoints")]
```
BusyJay, line 9: It doesn't need to be guarded.

rleungx: `make clippy` will report a warning.

BusyJay: If two feature cfgs are removed, then there should be no warnings.

components/resolved_ts/src/advance.rs
```diff
@@ -156,8 +157,7 @@ pub async fn region_resolved_ts_store(
     tikv_clients: Arc<Mutex<HashMap<u64, TikvClient>>>,
     min_ts: TimeStamp,
 ) -> Vec<u64> {
-    #[cfg(feature = "failpoint")]
-    (|| fail_point!("before_sync_replica_read_state", |_| regions))();
+    fail_point!("before_sync_replica_read_state", |_| regions.clone());
```
BusyJay, line 160: Either the old implementation or the new implementation is wrong. /cc @NingLin-P

BusyJay (approved):
Rest LGTM

PR #11716, opened 2021-12-24
raftstore: Wait for the apply index equals to the commit index

v01dstar:

TiKV issue #10483

What's Changed:

This PR ensures that when unsafe recovery is triggered, and TiKV is requested to send a report to PD, it sends an up-to-date report. This is achieved by waiting for all peers' apply index equals to the commit index when reporting request is received.

### Related changes

N/A

### Check List 

Tests 

- Integration Test

Side effects

N/A

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -674,10 +674,9 @@ where
                 PeerMsg::UpdateRegionForUnsafeRecover(region) => {
                     self.on_update_region_for_unsafe_recover(region)
                 }
-                PeerMsg::UnsafeRecoveryWaitApply {
-                    commit_index,
-                    counter,
-                } => self.on_unsafe_recovery_wait_apply(commit_index, counter),
+                PeerMsg::UnsafeRecoveryWaitApply(counter) => {
+                    self.on_unsafe_recovery_wait_apply(counter)
+                }
             }
         }
         // Propose batch request which may be still waiting for more raft-command
@@ -874,11 +873,13 @@ where
         self.register_raft_base_tick();
     }
 
-    fn on_unsafe_recovery_wait_apply(&mut self, commit_index: u64, counter: Arc<AtomicUsize>) {
-	// Checks the applied index here first, in case the log at the target index has been
-	// applied right before this peer msg is handled. If the log has not been applied here,
-	// further checks will be performed in on_apply_res().
-        if self.fsm.peer.raft_group.store().applied_index() >= commit_index {
+    fn on_unsafe_recovery_wait_apply(&mut self, counter: Arc<AtomicUsize>) {
+        // Checks the applied index here first, in case the log at the target index has been
+        // applied right before this peer msg is handled. If the log has not been applied here,
+        // further checks will be performed in on_apply_res().
+        if self.fsm.peer.raft_group.store().applied_index()
+            == self.fsm.peer.raft_group.store().commit_index()
+        {
             if counter.fetch_sub(1, Ordering::Relaxed) == 1 {
                 let stats = StoreStats::default();
                 let store_info = StoreInfo {
@@ -896,9 +897,10 @@ where
                 }
             }
         } else {
-	    self.fsm.unsafe_recovery_target_commit_index = commit_index;
-	    self.fsm.unsafe_recovery_wait_apply_counter = counter.clone();
-	}
+            self.fsm.unsafe_recovery_target_commit_index =
+                self.fsm.peer.raft_group.store().commit_index();
+            self.fsm.unsafe_recovery_wait_apply_counter = counter.clone();
+        }
     }
 
     fn on_casual_msg(&mut self, msg: CasualMessage<EK>) {
@@ -1527,10 +1529,17 @@ where
 
     fn on_apply_res(&mut self, res: ApplyTaskRes<EK::Snapshot>) {
         fail_point!("on_apply_res", |_| {});
-	// After a log has been applied, check if we need to trigger the unsafe recovery reporting procedure.
-	if self.fsm.unsafe_recovery_target_commit_index != 0 {
-            if self.fsm.peer.raft_group.store().applied_index() >= self.fsm.unsafe_recovery_target_commit_index {
-                if self.fsm.unsafe_recovery_wait_apply_counter.fetch_sub(1, Ordering::Relaxed) == 1 {
+        // After a log has been applied, check if we need to trigger the unsafe recovery reporting procedure.
+        if self.fsm.unsafe_recovery_target_commit_index != 0 {
+            if self.fsm.peer.raft_group.store().applied_index()
+                >= self.fsm.unsafe_recovery_target_commit_index
+            {
+                if self
+                    .fsm
+                    .unsafe_recovery_wait_apply_counter
+                    .fetch_sub(1, Ordering::Relaxed)
+                    == 1
+                {
                     let stats = StoreStats::default();
                     let store_info = StoreInfo {
                         kv_engine: self.ctx.engines.kv.clone(),
@@ -1546,9 +1555,9 @@ where
                         panic!("fail to send detailed report to pd {:?}", e);
                     }
                 }
-		self.fsm.unsafe_recovery_target_commit_index = 0;
-	    }
-	}
+                self.fsm.unsafe_recovery_target_commit_index = 0;
+            }
+        }
         match res {
             ApplyTaskRes::Apply(mut res) => {
                 debug!(
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -899,7 +899,7 @@ where
         } else {
             self.fsm.unsafe_recovery_target_commit_index =
                 self.fsm.peer.raft_group.store().commit_index();
-            self.fsm.unsafe_recovery_wait_apply_counter = counter.clone();
+            self.fsm.unsafe_recovery_wait_apply_counter = counter;
         }
     }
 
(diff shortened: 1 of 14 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -817,6 +823,36 @@ where
         self.register_raft_base_tick();
     }
 
+    fn on_unsafe_recovery_wait_apply(&mut self, commit_index: u64, counter: Arc<AtomicUsize>) {
+        if self.fsm.peer.raft_group.store().applied_index() != commit_index {
+            sleep(Duration::from_millis(1000));
```
Connor1996, line 828: it will block the raftstore, please record the commit index, and check it on the update of the apply index, which is in `on_apply_res`

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1080,13 +1080,72 @@ where
...
+                                        )
+                                        .unwrap()
+                                    {
+                                        None => return Ok(true),
+                                        Some(value) => value,
+                                    };
+                                    if apply_state.get_applied_index()
+                                        != apply_state.get_last_commit_index()
+                                    {
+                                        commit_indices.insert(
```
Connor1996, line 1118: no need to check here. broadcast the message to every peer fsm. And in each peer fsm, you can get the apply index and commit index from memory.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1080,13 +1080,22 @@ where
...
-                        if let Err(e) = scheduler.schedule(task) {
-                            error!("notify pd failed"; "err" => ?e);
+                        // Init the counter with 1 in case the msg processing is faster than the distributing thus cause FSMs race to send a report.
+                        let counter = Arc::new(AtomicUsize::new(1));
+                        let counter_clone = counter.clone();
+                        router.broadcast_normal(|| {
+                            let _ = counter_clone.fetch_add(1, Ordering::Relaxed);
+                            PeerMsg::UnsafeRecoveryWaitApply(counter_clone.clone())
+                        });
+                        if counter.fetch_sub(1, Ordering::Relaxed) == 1 {
```
Connor1996, line 1090: add a comment here

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1080,13 +1080,22 @@ where
                     }
                     if resp.get_require_detailed_report() {
                         info!("required to send detailed report in the next heartbeat");
```
Connor1996, line 1082: Add a detailed comment about how the message flow between tikv's several workers. Such like the merge process flow
截屏2022-01-19 上午10 32 27

components/raftstore/src/store/worker/pd.rs
v01dstar: Done

components/raftstore/src/store/worker/pd.rs
v01dstar: Done.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -143,6 +144,12 @@ where
     /// Before actually destroying a peer, ensure all log gc tasks are finished, so we
     /// can start destroying without seeking.
     logs_gc_flushed: bool,
+
+    /// To make sure the reported store/peer meta is up to date, each peer has to wait for the log
+    /// at its target commit index to be applied. The last peer does so triggers the next procedure
+    /// which is reporting the store/peer meta to PD.
+    unsafe_recovery_target_commit_index: u64,
```
Connor1996, line 151: use Option

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1524,6 +1568,34 @@ where
...
         }
+        // After a log has been applied, check if we need to trigger the unsafe recovery reporting procedure.
+        if self.fsm.unsafe_recovery_target_commit_index != 0
+            && self.fsm.peer.raft_group.store().applied_index()
+                >= self.fsm.unsafe_recovery_target_commit_index
+        {
+            if self
+                .fsm
+                .unsafe_recovery_wait_apply_counter
+                .fetch_sub(1, Ordering::Relaxed)
```
Connor1996, line 1579: sub it when the peer is destroyed.

tests/failpoints/cases/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,81 @@
...
+    let apply_triggered_pair = Arc::new((Mutex::new(false), Condvar::new()));
+    let apply_triggered_pair2 = Arc::clone(&apply_triggered_pair);
+    let apply_released_pair = Arc::new((Mutex::new(false), Condvar::new()));
+    let apply_released_pair2 = Arc::clone(&apply_released_pair);
+    fail::cfg_callback("on_handle_apply_store_1", move || {
+        {
+            let (lock, cvar) = &*apply_triggered_pair2;
+            let mut triggered = lock.lock().unwrap();
+            *triggered = true;
+            cvar.notify_one();
```
Connor1996, line 37: seems a sync channel is enough?

tests/failpoints/cases/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,81 @@
...
+use std::iter::FromIterator;
+use std::sync::{Arc, Condvar, Mutex};
+
+use futures::executor::block_on;
+use pd_client::PdClient;
+use raftstore::store::util::find_peer;
+use test_raftstore::*;
+
+#[test]
+fn test_unsafe_recover_send_report() {
```
Connor1996, line 13: Please add some comments in the test, to describe the purpose of every paragraph of the test code.

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: I did a little bit investigation. Seems to me that, `sync channel` does not work here, because the failpoint callback may be called multiple times, and `apply-triggered sender` cannot send messages multiple times while `apply-triggered receiver` only receive once, `apply-released receiver` will be blocked forever since `apply-released sender` only sends once. Correct me if I am wrong

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -873,33 +873,45 @@ where
         self.register_raft_base_tick();
     }
 
+    fn finish_unsafe_recovery_wait_apply(&mut self) {
+        if self
+            .fsm
+            .unsafe_recovery_wait_apply_counter
+            .as_ref()
+            .unwrap()
```
Connor1996, line 881: it may panic here

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2631,6 +2621,9 @@ where
         // We can't destroy a peer which is handling snapshot.
         assert!(!self.fsm.peer.is_handling_snapshot());
 
+        // No need to wait for the apply anymore.
+        self.finish_unsafe_recovery_wait_apply();
```
Connor1996, line 2625: it may not be waiting for unsafe recovery

v01dstar: Thank you for catching this.

components/raftstore/src/store/fsm/peer.rs
v01dstar: Are you worrying about the "destroy peer" case may panic here, or panic due to race condition?

Connor1996: about the "destroy_peer", now it's sovled

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -859,6 +873,48 @@ where
...
+    fn finish_unsafe_recovery_wait_apply(&mut self) {
+        if self
+            .fsm
+            .unsafe_recovery_wait_apply_counter
+            .as_ref()
+            .unwrap()
+            .fetch_sub(1, Ordering::Relaxed)
+            == 1
+        {
+            let stats = StoreStats::default();
```
NingLin-P, line 885: The default `StoreStats` does not set the `store_id` field, may cause PD confuse about where the store heartbeat comes from.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -859,6 +873,48 @@ where
...
+            };
+            if let Err(e) = self.ctx.pd_scheduler.schedule(task) {
+                panic!("fail to send detailed report to pd {:?}", e);
+            }
+        }
+        self.fsm.unsafe_recovery_target_commit_index = None;
+        self.fsm.unsafe_recovery_wait_apply_counter = None;
+    }
+
+    fn on_unsafe_recovery_wait_apply(&mut self, counter: Arc<AtomicUsize>) {
```
NingLin-P, line 904: Should also check `self.fsm.stopped` here, in case `destroy_peer` is already called.

NingLin-P (commented):
Rest LGTM

tests/failpoints/cases/test_unsafe_recovery.rs
Connor1996: seems you just want to block the apply process, why not use `fail::cfg("on_handle_apply_store_1", "pause");`?

tests/failpoints/cases/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,94 @@
...
+        let mut released = lock2.lock().unwrap();
+        *released = true;
+        cvar2.notify_all();
+    }
+
+    // Store reports are sent once the entries are applied.
+    let mut reported = false;
+    for _ in 0..20 {
+        if pd_client.must_get_store_reported() > 0 {
+            reported = true;
```
Connor1996, line 87: should check the report to make sure it's the state that you expect

Connor1996 (commented):
rest LGTM

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: We can definitely use `pause` to replace the functionality of `apply-released` switch. But, we still need `apply-triggered` switch to ensure that the "reporting" logic happens exactly after an entry is **committed** but **not applied** on this peer.  I combined these 2 requirements in this callback, instead of using a callback for `apply-released` plus a `fail point pause`.

components/raftstore/src/store/fsm/peer.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Done

NingLin-P: I still worry about a default `StoreStats` that many fields like `bytes_written` are set to 0 will cause PD to make some wrong schedule decision.

NingLin-P: How about introducing a new `StoreMsg` to let the store fsm (reuse `store_heartbeat_pd`) send a `StoreHeartbeat` (with `send_detailed_report: true`) to the PD worker immediately.

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Currently, it has been taken care of, see the PD [code here](https://github.com/tikv/pd/blob/master/server/grpc_service.go#L598). The best way to address that concern, while not bringing in other side effects, is to use a separate RPC instead of the heartbeat. This can be done as a optimization in the future? What do you think? @Connor1996

NingLin-P (approved):

Connor1996 (approved):
LGTM

components/raftstore/src/store/fsm/peer.rs
Connor1996: Yes, let's do it then if it's really an issue.

PR #11717, opened 2021-12-24
Standalone streaming for reporting region flow

5kbpers:

Ref https://github.com/tikv/tikv/issues/11718

What's Changed:
* Add a grpc streaming for reporting region flow
* Add a new tick of raftstore to schedule region flow reporting

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Standalone streaming for reporting region flow
```


components/raftstore/src/store/worker/pd.rs
```diff
@@ -1468,6 +1487,67 @@ where
     fn handle_region_cpu_records(&mut self, records: Arc<RawRecords>) {
         calculate_region_cpu_records(self.store_id, records, &mut self.region_cpu_records);
     }
+
+    fn region_flow(&mut self, region_id: u64, written_bytes: u64, written_keys: u64) -> RegionFlow {
```
lhy1024, line 1491: If we report data at the region granularity, will it also cause too much heartbeat pressure?

lhy1024: I think another problem with reporting at the region granularity may be that, as in previous versions, the reporting intervals and times for store and regions are inconsistent.

Issue #11718, opened 2021-12-24
Standalone streaming for reporting buckets flow


5kbpers:
Currently, region flow is reported through region heartbeat every 60s. After introducing dynamic region, we need more timely information for region split, so buckets flow should be reported more frequently.
For more details see [0082-dynamic-size-region.md#flow-report](https://github.com/tikv/rfcs/blob/master/text/0082-dynamic-size-region.md#flow-report)

lhy1024:
At present, TiKV report hot read statistics to PD by store heartbeat and report hot write statistic to PD by region heartbeat. I think we can unify them to store heartbeat.

lhy1024:
So it's really two questions：
1. Do we need Standalone streaming?
2. Is the granularity of report store or region or bucket?

nolouch:
We can use store stream to report TopN same as read flow. but change region heartbeat to store heartbeat will influence keyviz for now.

5kbpers:
> So it's really two questions：
> 
> 1. Do we need Standalone streaming?
> 2. Is the granularity of report store or region or bucket?

1. The benefit of standalone streaming is that we can change the duration of reporting flow individually. Reporting flow through store heartbeats seems to be more simple, so we can implement it in this way and introduce standalone streaming later if we do have needs. 
2. Will be buckets.

cc @lhy1024 @nolouch

5kbpers:
> We can use store stream to report TopN same as read flow. but change region heartbeat to store heartbeat will influence keyviz for now.

We can keep the flow reporting in region heartbeat.

bufferflies:
There will be consistency problem if keeps two reporting channel.

BusyJay:
Can this be closed?

PR #11730, opened 2021-12-27
pd: report write flow through store heartbeat

5kbpers:

Ref #11718 

What's Changed:
* New tick to report write flow to pd worker
* Report write flow through store heartbeat 

kvproto: https://github.com/pingcap/kvproto/pull/845

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Report write flow through store heartbeat
```


5kbpers:
cc @bufferflies

components/raftstore/src/store/worker/pd.rs
```diff
@@ -943,26 +964,36 @@ where
...
             region_peer.last_store_report_read_bytes = region_peer.read_bytes;
             region_peer.last_store_report_read_keys = region_peer.read_keys;
+            region_peer.last_store_report_written_bytes = region_peer.written_bytes;
+            region_peer.last_store_report_written_keys = region_peer.written_keys;
             region_peer
                 .last_store_report_query_stats
                 .fill_query_stats(&region_peer.query_stats);
             if read_bytes < hotspot_byte_report_threshold()
                 && read_keys < hotspot_key_report_threshold()
+                && written_bytes < hotspot_byte_report_threshold()
```
rleungx, line 980: `write` should have a different threshold. See https://github.com/tikv/pd/blob/ea57c6d6a0c9ad081aded76545f3d97bc465635f/server/statistics/hot_peer_cache.go#L46-L53

components/raftstore/src/store/worker/pd.rs
```diff
@@ -523,31 +523,49 @@ where
...
 }
 
-const HOTSPOT_KEY_RATE_THRESHOLD: u64 = 128;
-const HOTSPOT_QUERY_RATE_THRESHOLD: u64 = 128;
-const HOTSPOT_BYTE_RATE_THRESHOLD: u64 = 8 * 1024;
+const HOTSPOT_READ_KEY_RATE_THRESHOLD: u64 = 128;
+const HOTSPOT_READ_QUERY_RATE_THRESHOLD: u64 = 128;
+const HOTSPOT_READ_BYTE_RATE_THRESHOLD: u64 = 8 * 1024;
+
+const HOTSPOT_WRITE_KEY_RATE_THRESHOLD: u64 = 32;
```
lhy1024, line 530: we also need a threshold for the write query rate, how about 32?

5kbpers: But we don't use it now. Perhaps could add it later

components/raftstore/src/store/config.rs
```diff
@@ -300,7 +301,7 @@ impl Default for Config {
             // Disable consistency check by default as it will hurt performance.
             // We should turn on this only in our tests.
             consistency_check_interval: ReadableDuration::secs(0),
-            report_region_flow_interval: ReadableDuration::minutes(1),
+            report_region_flow_interval: ReadableDuration::secs(1),
```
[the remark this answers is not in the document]
5kbpers: This is an interval that raftstore reports write flow to pd worker, change it to 1s for making the flow updated promptly.

components/raftstore/src/store/worker/pd.rs
5kbpers: Done

PR #11750, opened 2021-12-29
raftstore: Extract significant router

Connor1996:

Part of https://github.com/tikv/tikv/pull/11409

What's Changed:
Extract significant router, so raft log fetcher worker can use it to dispatch related msg.

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```


tabokie (approved):

components/raftstore/src/router.rs
```diff
@@ -33,7 +37,9 @@ where
         &self,
         region_id: u64,
         msg: SignificantMsg<EK::Snapshot>,
-    ) -> RaftStoreResult<()>;
+    ) -> RaftStoreResult<()> {
+        <Self as SignificantRouter<EK>>::send(self, region_id, msg)
```
BusyJay, line 41: How about removing `significant_send` and renaming `send` to `significant_send`?

Connor1996: `send_casual_msg` and `send_store_msg` does it in the same manner, better to keep it consistent.

BusyJay: Now I think keeping them different name may be better. As if a trait replace with another accidentally, it will report error instead of succeed silently.

BusyJay (approved):

PR #11753, opened 2021-12-30
res_meter: optimize top k

### LINKED ISSUE #11754: res_meter: optimize top k
## Development Task

Now we have the logic to retain top k before each report, but we still need to retain the top k items after collecting CPU data per-second.

UPDATED: then remove the logic which retain top k before each report.

### Expected modification

- Implement `keep_top_k()` for `RawRecords` in `components/resource_metering/src/model.rs`.
- Call `keep_top_k()` in `Reporter::handle_records()`.

mornyx:
Signed-off-by: mornyx <mornyx.z@gmail.com>

### What's changed

Close #11754 

- Optimize top k logic.
- Remove `recorder::RecorderHandle` and introduce `recorder::ConfigChangeNotifier`.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/resource_metering/src/config.rs
+++ b/components/resource_metering/src/config.rs
@@ -40,7 +40,7 @@ impl Default for Config {
         Config {
             receiver_address: "".to_string(),
             report_receiver_interval: ReadableDuration::minutes(1),
-            max_resource_groups: 200,
+            max_resource_groups: 100,
             precision: ReadableDuration::secs(1),
         }
     }
--- a/components/resource_metering/src/config.rs
+++ b/components/resource_metering/src/config.rs
@@ -1,7 +1,7 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
-use crate::recorder::RecorderHandle;
-use crate::reporter::ConfigChangeNotifier;
+use crate::recorder::ConfigChangeNotifier as RecorderConfigChangeNotifier;
+use crate::reporter::ConfigChangeNotifier as ReporterConfigChangeNotifier;
 use crate::AddressChangeNotifier;
 
 use std::error::Error;
@@ -40,7 +40,7 @@ impl Default for Config {
         Config {
             receiver_address: "".to_string(),
             report_receiver_interval: ReadableDuration::minutes(1),
-            max_resource_groups: 2000,
+            max_resource_groups: 200,
             precision: ReadableDuration::secs(1),
         }
     }
@@ -84,22 +84,23 @@ impl Config {
 /// to control the dynamic update of the configuration.
 pub struct ConfigManager {
     current_config: Config,
-    recorder: RecorderHandle,
-    config_notifier: ConfigChangeNotifier,
+
+    recorder_notifier: RecorderConfigChangeNotifier,
+    reporter_notifier: ReporterConfigChangeNotifier,
     address_notifier: AddressChangeNotifier,
 }
 
 impl ConfigManager {
     pub fn new(
         current_config: Config,
-        recorder: RecorderHandle,
-        config_notifier: ConfigChangeNotifier,
+        recorder_notifier: RecorderConfigChangeNotifier,
+        reporter_notifier: ReporterConfigChangeNotifier,
         address_notifier: AddressChangeNotifier,
     ) -> Self {
         ConfigManager {
             current_config,
-            recorder,
-            config_notifier,
+            recorder_notifier,
+            reporter_notifier,
             address_notifier,
         }
     }
@@ -110,15 +111,13 @@ impl online_config::ConfigManager for ConfigManager {
         let mut new_config = self.current_config.clone();
         new_config.update(change);
         new_config.validate()?;
-        if self.current_config.precision != new_config.precision {
-            self.recorder.precision(new_config.precision.0);
-        }
         if self.current_config.receiver_address != new_config.receiver_address {
             self.address_notifier
                 .notify(new_config.receiver_address.clone());
         }
         // Notify reporter that the configuration has changed.
-        self.config_notifier.notify(new_config.clone());
+        self.recorder_notifier.notify(new_config.clone());
+        self.reporter_notifier.notify(new_config.clone());
         self.current_config = new_config;
         Ok(())
     }
--- a/components/resource_metering/src/model.rs
+++ b/components/resource_metering/src/model.rs
@@ -92,6 +92,30 @@ impl RawRecords {
         STATIC_BUF.with(move |b| b.set(buf));
         others
     }
+
+    /// Returns (TopK, Evicted).
+    /// It is caller's responsibility to ensure that k < records.len().
+    pub fn top_k(
+        &self,
+        k: usize,
+    ) -> (
+        impl Iterator<Item = (&Arc<TagInfos>, &RawRecord)>,
+        impl Iterator<Item = (&Arc<TagInfos>, &RawRecord)>,
+    ) {
+        assert!(self.records.len() > k);
+        let mut buf = STATIC_BUF.with(|b| b.take());
+        buf.clear();
+        for record in self.records.values() {
+            buf.push(record.cpu_time);
+        }
+        pdqselect::select_by(&mut buf, k, |a, b| b.cmp(a));
+        let kth = buf[k];
+        STATIC_BUF.with(move |b| b.set(buf));
+        (
+            self.records.iter().filter(move |(_, v)| v.cpu_time > kth),
+            self.records.iter().filter(move |(_, v)| v.cpu_time <= kth),
+        )
+    }
 }
 
 /// Resource statistics.
@@ -190,7 +214,11 @@ impl From<Records> for Vec<ResourceUsageRecord> {
 
 impl Records {
     /// Aggregates [RawRecords] into [Records].
-    pub fn append(&mut self, raw_records: Arc<RawRecords>) {
+    pub fn append<'a>(
+        &mut self,
+        ts: u64,
+        iter: impl Iterator<Item = (&'a Arc<TagInfos>, &'a RawRecord)>,
+    ) {
         // # Before
         //
         // ts: 1630464417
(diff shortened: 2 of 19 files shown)


mornyx:
/cc @zhongzc @breeswish 
PTAL, Thanks

components/resource_metering/src/model.rs
```diff
@@ -64,10 +65,36 @@ impl Default for RawRecords {
...
             duration: Duration::default(),
             records: HashMap::default(),
+            others: RawRecord::default(),
         }
     }
 }
 
+impl RawRecords {
+    /// Keep a maximum of `k` self.records and aggregate the others into self.others.
+    pub fn keep_top_k(&mut self, k: usize) {
```
breeswish, line 75: Seems that we have two `top_k` functions. Is it possible to be one?

components/resource_metering/src/model.rs
```diff
@@ -64,10 +65,36 @@ impl Default for RawRecords {
...
+        if self.records.len() <= k {
+            return;
+        }
+        let mut buf = STATIC_BUF.with(|b| b.take());
+        buf.clear();
+        // Find kth top cpu time.
+        for record in self.records.values() {
+            buf.push(record.cpu_time);
+        }
+        pdqselect::select_by(&mut buf, k, |a, b| b.cmp(a));
```
breeswish, line 85: How about select_by self.records directly? Then you don't need to use the buf any more.

components/resource_metering/src/recorder/sub_recorder/cpu.rs
```diff
@@ -47,6 +56,10 @@ impl SubRecorder for CpuRecorder {
         });
     }
 
+    fn collect(&mut self, records: &mut RawRecords, _: &mut HashMap<usize, LocalStorage>) {
+        records.keep_top_k(self.max_resource_groups);
```
breeswish, line 60: For per-second top K, we can leave only top 100 (MAX_MAX_RESOURCE_GROUPS = 100). BTW I think there is no need to do a 1-minute top K after that. What do you think? @zhongzc

components/resource_metering/src/model.rs
mornyx: `self.records` is an unordered `HashMap`, but `pdqselect::select_by()` requires to be done on `&mut [T]`, it will sort in place.

components/resource_metering/src/model.rs
mornyx: It depends on whether we can unify the behavior of `keep_top_k(0)`. By the way, why does `Records::keep_top_k(0)` clean up `others` instead of merging all records into `others`? cc @zhongzc 
```rust
         if k == 0 {
             self.records.clear();
             self.others.clear();
             return;
         }
```

components/resource_metering/src/recorder/sub_recorder/cpu.rs
mornyx: This may require two configuration items, one for per-second and one for per-minute(current `max_resource_groups`).

If we don’t need it to be configurable, we can simply use constant.

components/resource_metering/src/model.rs
```diff
@@ -64,10 +65,36 @@ impl Default for RawRecords {
...
+        for (_, record) in evicted_records {
+            others.merge(&record);
+        }
+        STATIC_BUF.with(move |b| b.set(buf));
+    }
+}
+
 /// Resource statistics.
 ///
 /// TODO(mornyx): Optimize to Vec<Item{timestamp,cpu_time,read_keys,write_keys}>
```
zhongzc, line 100:
```suggestion
```

components/resource_metering/src/recorder/mod.rs
```diff
@@ -290,6 +290,7 @@ impl RecorderHandle {
 /// This function is intended to simplify external use.
 pub fn init_recorder(
     precision_ms: u64,
+    max_resource_groups: usize,
```
zhongzc, line 293:
```suggestion
    config: Config,
```

components/resource_metering/src/recorder/mod.rs
```diff
@@ -298,7 +299,7 @@ pub fn init_recorder(
 ) {
     let (recorder, recorder_handle) = RecorderBuilder::default()
         .precision_ms(Arc::new(AtomicU64::new(precision_ms)))
-        .add_sub_recorder(Box::new(CpuRecorder::default()))
+        .add_sub_recorder(Box::new(CpuRecorder::new(max_resource_groups)))
         .add_sub_recorder(Box::new(SummaryRecorder::default()))
```
zhongzc, line 303: add a comment about ordering (why cpu first and summary second)?

components/resource_metering/src/model.rs
zhongzc: The per-minute one is being removed. This issue is resolved.

zhongzc:
The default value of the config should be modified.

components/resource_metering/src/model.rs
mornyx: We still have:
```rust
pub struct Record {
    pub timestamps: Vec<u64>,
    pub cpu_time_list: Vec<u32>,
    pub read_keys_list: Vec<u32>,
    pub write_keys_list: Vec<u32>,
    pub total_cpu_time: u32,
}
```

This TODO refers to here.

components/resource_metering/src/model.rs
breeswish: This PR should remove the per-minute one.

mornyx: updating

components/resource_metering/src/recorder/sub_recorder/cpu.rs
breeswish: We can drop the per-minute, and modifying the default value to be 100. In this way, no new config will be introduced.

components/resource_metering/src/reporter/mod.rs
```diff
@@ -87,8 +88,14 @@ impl Reporter {
...
-        self.records.append(records);
-        self.records.keep_top_k(self.config.max_resource_groups);
+        let mut records = records.deref().clone();
+        let others = records.keep_top_k(self.config.max_resource_groups);
+        self.records
+            .others
+            .entry(records.begin_unix_time_secs)
+            .or_default()
+            .merge(&others);
+        self.records.append(Arc::new(records));
```
zhongzc, line 98:
```suggestion
        let (top_k, others) = raw_records.top_k(self.config.max_resource_groups);
        self.records
            .others
            .entry(records.begin_unix_time_secs)
            .or_default()
            .merge(&others);
        self.records.append(top_k);
```

components/resource_metering/src/config.rs
```diff
@@ -40,7 +40,7 @@ impl Default for Config {
         Config {
             receiver_address: "".to_string(),
             report_receiver_interval: ReadableDuration::minutes(1),
-            max_resource_groups: 2000,
+            max_resource_groups: 200,
```
breeswish: Please change to 100~

mornyx: Has been modified before, try to refresh :)

components/resource_metering/src/reporter/mod.rs
```diff
@@ -87,8 +88,14 @@ impl Reporter {
     }
 
     fn handle_records(&mut self, records: Arc<RawRecords>) {
-        self.records.append(records);
-        self.records.keep_top_k(self.config.max_resource_groups);
+        let mut records = records.deref().clone();
+        let others = records.keep_top_k(self.config.max_resource_groups);
```
breeswish, line 92: Why is records a `Arc<...>` so that we need a clone here? Is it possible to avoid that clone?

breeswish (approved):
the rest LGTM

zhongzc:
We can save unnecessary clones by implementing topk like following for Records.

```rust
    pub fn top_k(&self, k: usize) -> (impl Iterator<Item = &RawRecord> + '_, RawRecord) {
        assert!(self.records.len() > k);

        // Evict records with cpu time less or equal than `kth`
        let mut buf = STATIC_BUF.with(|b| b.take());
        buf.clear();
        // Find kth top cpu time.
        for record in self.records.values() {
            buf.push(record.cpu_time);
        }
        pdqselect::select_by(&mut buf, k, |a, b| b.cmp(a));
        let kth = buf[k];
        STATIC_BUF.with(move |b| b.set(buf));

        // Record evicted into others
        let others = self
            .records
            .values()
            .filter(move |r: &&RawRecord| (*r).cpu_time <= kth)
            .fold(RawRecord::default(), |mut others, r| {
                others.merge(r);
                others
            });

        (
            self.records
                .values()
                .filter(move |r: &&RawRecord| (*r).cpu_time > kth),
            others,
        )
    }
```

components/resource_metering/src/reporter/mod.rs
mornyx: 1. The reason why records is `Arc` is that we may have multiple `Collector`s to share this read-only records.
2. Yes, I'm trying to let `top_k()` to return tow Iterators.

components/resource_metering/src/reporter/mod.rs
```diff
@@ -87,8 +87,16 @@ impl Reporter {
...
+        let ts = records.begin_unix_time_secs;
+        if self.config.max_resource_groups >= records.records.len() {
+            self.records.append(ts, records.records.iter());
+            return;
+        }
+        let (top, evicted) = records.top_k(self.config.max_resource_groups);
+        self.records.append(ts, top);
+        evicted.for_each(|(_, v)| {
+            self.records.others.entry(ts).or_default().merge(v);
+        });
```
zhongzc, line 99:
```suggestion
        let others = self.records.others.entry(ts).or_default();
        evicted.for_each(|(_, v)| { others.merge(v); });
```

zhongzc:
rest LGTM

zhongzc (approved):

PR #11756, opened 2021-12-30
coprocesser: implement upper and lower function on GBK charset

xiongjiwei:

Close https://github.com/tikv/tikv/issues/11757

some characters only have lower or upper case under gbk encoding. if so, we should make its lower or upper eq to itself

see https://github.com/pingcap/tidb/pull/28869 for more details

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


xiongjiwei:
/cc @andylokandy @zhongzc 
PTAL

zhongzc (approved):

breeswish (approved):

PR #11758, opened 2021-12-30
raftstore: propose in-memory pessimistic locks before prepare merge

sticnarf:
- [x] Based on https://github.com/tikv/tikv/pull/11655

Refer to #11452 

Before prepare merge, we cannot propose pessimistic locks immediately because there can be write requests removing pessimistic locks proposed but not applied.

So, if there are pessimistic locks, we firstly ban all write requests and record the current propose index, which is called `prepare_merge_fence`. After the applied index reaches `prepare_merge_fence`, we can safely propose all pessimistic locks and then propose prepare merge.

### Check List 

Tests 

- Unit tests

### Release note 

```release-note
None
```


youjiali1995 (commented):
Can memory locks are piggybacked by `CommitMerge`?

components/raftstore/src/store/peer.rs
```diff
@@ -3556,19 +3602,68 @@ where
                 cmd_type
             ));
         }
-        if entry_size as f64 > ctx.cfg.raft_entry_max_size.0 as f64 * 0.9 {
+        let entry_size_limit = ctx.cfg.raft_entry_max_size.0 as usize * 10 / 9;
```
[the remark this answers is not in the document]
sticnarf: Sorry, it's a mistake...

components/raftstore/src/store/peer.rs
```diff
@@ -3496,12 +3509,45 @@ where
...
+                        ));
+                    }
+                    self.prepare_merge_fence = last_index;
+                    info!("start rejecting new proposals before prepare merge"; "region_id" => self.region_id);
+                }
+                if self.prepare_merge_fence > 0
+                    && self.get_store().applied_index() < self.prepare_merge_fence
+                {
+                    self.pending_prepare_merge = Some(mem::take(req));
+                    return Err(box_err!(
```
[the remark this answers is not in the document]
sticnarf: I change to ignore printing the log specially for a pending merge.

sticnarf:
> Can memory locks are piggybacked by `CommitMerge`?

It's not easy.

`CommitMerge` is proposed from the source region peer which is in the same TiKV with the target region leader. The source region peer can be a follower, which does not include the in-memory pessimistic locks. So, we cannot simply let the source region collect in-memory pessimistic locks and generate a `CommitMerge` proposal with locks.

components/raftstore/src/store/peer.rs
```diff
@@ -3556,19 +3602,68 @@ where
...
+        ctx: &mut PollContext<EK, ER, T>,
+        size_limit: usize,
+    ) -> Result<()> {
+        let pessimistic_locks = self.txn_ext.pessimistic_locks.upgradable_read();
+        if pessimistic_locks.is_empty() {
+            let mut pessimistic_locks = RwLockUpgradableReadGuard::upgrade(pessimistic_locks);
+            pessimistic_locks.is_valid = false;
+            return Ok(());
+        }
+        if pessimistic_locks.memory_size > size_limit {
```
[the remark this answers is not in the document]
youjiali1995: I get it. Locks are piggybacked in `CommitMerge` indirectly by proposing these before the `PrepareMerge`.

components/raftstore/src/store/peer.rs
```diff
@@ -3600,13 +3694,14 @@ where
...
+    fn propose_normal<T: Transport>(
         &mut self,
         poll_ctx: &mut PollContext<EK, ER, T>,
         mut req: RaftCmdRequest,
     ) -> Result<Either<u64, u64>> {
-        if self.pending_merge_state.is_some()
-            && req.get_admin_request().get_cmd_type() != AdminCmdType::RollbackMerge
+        if (self.pending_merge_state.is_some()
+            && req.get_admin_request().get_cmd_type() != AdminCmdType::RollbackMerge)
+            || (self.prepare_merge_fence > 0 && !req.has_admin_request())
```
youjiali1995, line 3704: What admin requests can be useful here except `PrepareMerge`? These admin requests may forbid merge.

components/raftstore/src/store/peer.rs
```diff
@@ -3496,12 +3509,42 @@ where
...
         ctx: &mut PollContext<EK, ER, T>,
         req: &mut RaftCmdRequest,
     ) -> Result<()> {
         let last_index = self.raft_group.raft.raft_log.last_index();
+
+        // Record current proposed index. If there are some in-memory pessimistic locks, we should
+        // wait until applying to the proposed index before proposing pessimistic locks and
+        // PrepareMerge. Otherwise, if an already proposed command will remove a pessimistic lock,
+        // we will make some deleted locks appear again.
+        {
```
youjiali1995, line 3523: Is it a good idea to move this check after checking `merge_max_log_gap`, admin requests and `raft_entry_max_size`? And when there is a `prepare_merge_fence`, only check apply index and do not check these again.

components/raftstore/src/store/peer.rs
sticnarf: I was thinking other admin requests are just acceptable here. `PrepareMerge` can just fail in this case.

components/raftstore/src/store/peer.rs
sticnarf: > Is it a good idea to move this check after checking `merge_max_log_gap`, admin requests and `raft_entry_max_size`?

Ah, yes. We shouldn't block write if the log gap is too large.

Now I move it after these checks.

> And when there is a `prepare_merge_fence`, only check apply index and do not check these again.

I'm not sure... It may make the code more complex. And when the applied index reaches the fence, some pessimistic locks may be deleted and the log gap may shrink too. It will be more likely to pass the size check. Considering the frequency of region merge, perhaps a second calculation is acceptable.

components/raftstore/src/store/peer.rs
```diff
@@ -3556,19 +3569,96 @@ where
...
+                }
+                if self.prepare_merge_fence > 0
+                    && self.get_store().applied_index() < self.prepare_merge_fence
+                {
+                    self.pending_prepare_merge = Some(mem::take(req));
+                    return Err(Error::PendingPrepareMerge);
+                }
+            }
         }
+
```
youjiali1995, line 3603:
```suggestion
        // Check passed, clear fence and start proposing PrepareMerge.
        self.prepare_merge_fence = 0;
        self.pending_prepare_merge = None;
```

components/raftstore/src/store/peer.rs
youjiali1995: > And when the applied index reaches the fence, some pessimistic locks may be deleted and the log gap may shrink too. It will be more likely to pass the size check.

The size check can't block merge in this case, but is it possible there is a new admin request within the log gap that forbid to merge because admin requests can pass `prepare_merge_fence` check? We should clear fence if it happens.

youjiali1995: Clear the fence when:
1. pass all checks and start proposing `PrepareMerge`.
2. fail to propose `PrepareMerge` and the fence exsits.

sticnarf: > is it possible there is a new admin request within the log gap that forbid to merge because admin requests can pass `prepare_merge_fence` check

Definitely. Now, I use the similar way as before, clear the fence before prepare merge checks. Hope to reduce subtleties burning brains...

youjiali1995 (approved):
LGTM, but it's better to add tests to cover recent changes.

components/raftstore/src/store/peer.rs
```diff
@@ -3556,19 +3583,92 @@ where
...
+            .set_region_epoch(self.region().get_region_epoch().clone());
+        cmd.mut_header().set_peer(self.peer.clone());
+
+        {
+            let mut pessimistic_locks = RwLockUpgradableReadGuard::upgrade(pessimistic_locks);
+            pessimistic_locks.is_valid = false;
+        }
+        debug!("propose {} pessimistic locks before prepare merge", cmd.get_requests().len();
+            "region_id" => self.region_id);
+        self.propose_normal(ctx, cmd)?;
```
5kbpers, line 3666: How about using `propose_raft_command_internal` here?

components/raftstore/src/store/peer.rs
```diff
@@ -3556,19 +3583,92 @@ where
...
+        };
+
+        // Record current proposed index. If there are some in-memory pessimistic locks, we should
+        // wait until applying to the proposed index before proposing pessimistic locks and
+        // PrepareMerge. Otherwise, if an already proposed command will remove a pessimistic lock,
+        // we will make some deleted locks appear again.
+        if !passed_merge_fence {
+            let pessimistic_locks = self.txn_ext.pessimistic_locks.read();
+            if !pessimistic_locks.is_empty() {
+                if !pessimistic_locks.is_valid {
```
5kbpers, line 3600: Could we assert it should always be true because all cases have already been checked?

components/raftstore/src/store/peer.rs
sticnarf: Seems not possible to call `propose_raft_command_internal` which is a function of `PeerFsmDelegate` from `Peer`?

components/raftstore/src/store/peer.rs
sticnarf: I think it's possible that it's going to have a transfer leader, so `is_valid` becomes false. And related commands also have been committed, so the checks are pased.

5kbpers: Can you add some comments for clarifying this?

components/raftstore/src/store/peer.rs
```diff
@@ -3501,11 +3514,32 @@ where
         Ok((min_m, min_c))
     }
 
-    fn pre_propose_prepare_merge<T>(
-        &self,
+    fn pre_propose_prepare_merge<T: Transport>(
```
NingLin-P, line 3517: Seems here is the only place (except leader change) that may update `prepare_merge_fence`, does it mean we very rely on PD to retry the merge command to push forward the merge process? If the merge request is keep failing (perhaps due to the `pessimistic_locks` is too large) and PD end the merge operator with a timeout error hence won't retry the merge command anymore, the region may remain in merging mode forever and won't accept incoming proposals.

sticnarf: It is also retried in `on_apply_res`. Which means, when the `applied_index` reaches `prepare_merge_fence`, we will continue.

components/raftstore/src/store/peer.rs
```diff
@@ -3560,19 +3594,99 @@ where
...
+    ) -> Result<()> {
+        let pessimistic_locks = self.txn_ext.pessimistic_locks.upgradable_read();
+        if pessimistic_locks.is_empty() {
+            let mut pessimistic_locks = RwLockUpgradableReadGuard::upgrade(pessimistic_locks);
+            pessimistic_locks.is_valid = false;
+            return Ok(());
+        }
+        // The proposed pessimistic locks here will also be carried in CommitMerge. Check the size
+        // to avoid CommitMerge exceeding the size limit of a raft entry.
+        if pessimistic_locks.memory_size > size_limit {
```
NingLin-P, line 3654: Seems `pessimistic_locks.memory_size` is the sum of all `key.len() + locks.memory_size()`, although we have reserve `raft_entry_max_size * 0.1` for extra fields in `CommitMerge`, depend on the number of pessimistic locks, more extra fields will bring by `PutRequest`, will their size exceed the reserved size?

components/raftstore/src/store/peer.rs
```diff
@@ -1658,6 +1668,9 @@ where
                     // Init the in-memory pessimistic lock table when the peer becomes leader.
                     self.activate_in_memory_pessimistic_locks();
 
+                    self.prepare_merge_fence = 0;
+                    self.pending_prepare_merge = None;
```
NingLin-P, line 1672: Perhaps we should clear them when the leader step down instead.

components/raftstore/src/store/peer.rs
NingLin-P: In `on_apply_res`, we retried with the previous saved `PrepareMerge` request, it is also required the request can still pass all the checks before `pre_propose_prepare_merge`

sticnarf: Sounds like something worth considering. But what check may it fail to pass before `pre_propose_prepare_merge` while there is not any region change and the region can still serve request?

NingLin-P: I have been checked the code, checks that rely on states that outside of the source peer, like the disk full check and the target peer check:
https://github.com/tikv/tikv/blob/4e2af0426c13657c6eeb69bffc7b9847f2f61b47/components/raftstore/src/store/fsm/peer.rs#L3886-L3902

which may change after waiting for the source peer's `appied_index` catch up. 

The source peer's state can also change as apply more log, but with the `epoch_checker` there won't have `EpochNotMatch` error though.

components/raftstore/src/store/peer.rs
sticnarf: Protobuf integers and bytes use variant encoding, it is hard to compare (usually the proposal will be smaller than the original map).
I will add a second check of the size.

components/raftstore/src/store/peer.rs
sticnarf: Now I always clear the fence after proposing `PrepareMerge` in `on_apply_res`. I think the region will return available even if `PrepareMerge` fails to propose.

components/raftstore/src/store/peer.rs
sticnarf: Moved to `clear_in_memory_pessimistic_locks`

sticnarf:
@NingLin-P Thanks for the comments! I've addressed them and could you take another look?

components/raftstore/src/store/peer.rs
```diff
@@ -3560,19 +3591,108 @@ where
...
-        &self,
+    fn propose_locks_before_prepare_merge<T: Transport>(
+        &mut self,
+        ctx: &mut PollContext<EK, ER, T>,
+        size_limit: usize,
+    ) -> Result<()> {
+        let pessimistic_locks = self.txn_ext.pessimistic_locks.upgradable_read();
+        if pessimistic_locks.is_empty() {
+            let mut pessimistic_locks = RwLockUpgradableReadGuard::upgrade(pessimistic_locks);
+            pessimistic_locks.is_valid = false;
```
NingLin-P, line 3646: If the `PrepareMerge` command fails to propose later on (`raft_group.propose` may still return error), this flag may reject incoming retrying and cause the region can't be merged until leadership change or `pessimistic_locks ` become empty.

NingLin-P (commented):
Rest LGTM

NingLin-P (approved):

sticnarf:
I will merge it first.

I've added FIXME in apply to the last two comments. And I will address them in https://github.com/tikv/tikv/pull/11883.

sticnarf:
> I will merge it first.
> 
> I've added FIXME in reply to the last two comments. And I will address them in #11883.

After some thoughts, the problem is not suitable to be fixed by tick. Instead, I open https://github.com/tikv/tikv/pull/11985 to fix it (unit tests to be added)

PR #11760, opened 2021-12-30
rocksdb: consider env sharing when configuring background threads

tabokie:

What's Changed:

#9649 increased the default setting of kvdb background flushes (3 for 8c and 4 for 16c).

This PR takes another approach: use smaller concurrency for individual dbs, but configure the global thread pool size with the sum of kvdb and raftdb concurrency. This way we won't over-configure flush workers when raft-engine is enabled.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


tabokie:
Threads allocation compared:
(kv flushes, kv compactions) + (raft flushes, raft compactions) = (worker flushes, worker compactions)
|     | old                   | new                   | new with raft engine |
| --- | --------------------- | --------------------- | -------------------- |
| 4c  | (2,2) + (1,3) = (2,3) | (1,2) + (1,2) = (2,2) | (1,2) + () = (1,2)   |
| 8c  | (3,5) + (1,3) = (3,5) | (2,5) + (1,3) = (3,5) | (2,5) + () = (2,5)   |
| 16c | (4,6) + (1,3) = (4,6) | (3,6) + (1,3) = (4,6) | (3,6) + () = (3,6)   |

src/config.rs
```diff
@@ -2955,6 +2953,14 @@ impl TiKvConfig {
         // Reserve some space for page cache. The
         ReadableSize((total as f64 * MEMORY_USAGE_LIMIT_RATE) as u64)
     }
+
+    pub fn configure_shared_rocks_env(&self, env: &Env) {
+        if !self.raft_engine.enable {
+            env.set_high_priority_background_threads(
```
hunterlxt, line 2959: I suggest to add some explanation where the changes to the TiKV based on internal logic of rocksdb, which is convenient for maintenance.

hunterlxt (commented):
btw can you post negative effects caused by over-config? which will let me know the significance of this change

hunterlxt:
another question, why not just modify `opts.set_max_background_compactions` with sum of raftdb and kvdb but set bg workers manually.

hunterlxt:
another question, seems that `opts.set_max_background_flushes` in `config.rs` is not used?

tabokie:
> btw can you post negative effects caused by over-config? which will let me know the significance of this change

Not much difference under most cases. It just feels right to configure them this way.

> another question, seems that `opts.set_max_background_flushes` in `config.rs` is not used?

What do you mean? It's used in `build_opt`.

hunterlxt:
> > another question, seems that `opts.set_max_background_flushes` in `config.rs` is not used?
> 
> What do you mean? It's used in `build_opt`.

but you set it in `configure_shared_rocks_env`

src/config.rs
Connor1996, line 2959: no need to set `background_jobs`?

tabokie:
@hunterlxt @Connor1996 Are you suggesting to scale up kvdb flush concurrency to total flush worker count (make the behavior completely consistent with before)? I was thinking if the actual workers are enough, we don't have to give kvdb that much concurrency.

hunterlxt:
> @hunterlxt @Connor1996 Are you suggesting to scale up kvdb flush concurrency to total flush worker count (make the behavior completely consistent with before)? I was thinking if the actual workers are enough, we don't have to give kvdb that much concurrency.

No ideas but just don't know the difference of doing this compared to before.

tabokie:
@hunterlxt I thought the https://github.com/tikv/tikv/pull/11760#issuecomment-1006290734 is clear enough, it's to reduce flush workers when raft-engine is enabled (while not changing the disabled case).

hunterlxt:
@tabokie  I meant didn't see the improvement.

tabokie:
@hunterlxt I see.
I just think this is the way it's supposed to be implemented in the first place, because sharing `Env` is not something the RocksDB side would expect, and there should be some extra work on our side to make sure the sharing is done right. Before this PR, the complexity is moved to tuning the kvdb concurrency (to make room for both kvdb and raftdb).

Connor1996 (approved):
LGTM

hunterlxt (approved):

you06:
@tabokie Seems the update of [config-template.toml](https://github.com/tikv/tikv/blob/ac32d5c8099a85994a9790ae1840755033dd0fac/etc/config-template.toml#L471-L483) is missing in both #9649 and this PR.

Issue #11759, opened 2021-12-30
TiKV supports buckets


tonyxuqqi:
In large region size,  we need a smaller "sub-region" called bucket to parallelize the scan, collect the traffic statistics and do proper split/merge. 
TiKV needs to automatically identify the buckets and collect their traffic statistics such as read, scan, write throughput and report them to PD.

BusyJay:
Can this be closed?

PR #11763, opened 2021-12-30
TiKV supports buckets

tonyxuqqi:

Ref #11759 

What's Changed:

### Related changes

- PR to update `pingcap/kvproto`: https://github.com/pingcap/kvproto/pull/849

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
TiKv supports buckets
```


components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -85,8 +85,47 @@ impl<'a, E> Host<'a, E> {
...
+            self.cfg.region_max_size.0
+        };
+
+        const MIN_BUCKET_COUNT_PER_REGION: u64 = 2;
+        const SAMPLE_COUNT_PER_BUCKET: u64 = 10;
+        if region_size >= self.cfg.region_bucket_size.0 * MIN_BUCKET_COUNT_PER_REGION {
+            let mut bucket_checker = size::Checker::new(
+                self.cfg.region_bucket_size.0 * MIN_BUCKET_COUNT_PER_REGION
+                    / SAMPLE_COUNT_PER_BUCKET,
+                self.cfg.region_bucket_size.0 / SAMPLE_COUNT_PER_BUCKET,
```
youjiali1995, line 105: useless

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -85,8 +85,47 @@ impl<'a, E> Host<'a, E> {
...
 
+    pub fn approximate_bucket_keys<Kv: engine_traits::KvEngine>(
+        &mut self,
+        region: &Region,
+        engine: &Kv,
+    ) -> Result<Vec<Vec<u8>>> {
+        let region_size = if let Ok(size) = get_region_approximate_size(engine, region, 0) {
+            size
+        } else {
+            self.cfg.region_max_size.0
```
youjiali1995, line 96: I don't think it's a good idea to use max size as fallback. It may split too many buckets.

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -198,10 +197,35 @@ where
...
+                    vec![]
+                }
+            };
+            info!("starting approximate_bucket_keys {}", bucket_keys.len());
+            if !bucket_keys.is_empty() {
+                let mut region_buckets = Buckets::default();
+                region_buckets.set_keys(::protobuf::RepeatedField::from_vec(bucket_keys));
+                let _ = self.router.send(
+                    region.get_id(),
+                    CasualMessage::RefreshRegionBuckets { region_buckets },
```
youjiali1995, line 224: It's unnecessary if the region has to split by scan.

youjiali1995 (commented):
Is it necessary to refresh buckets immediately when the region range is changed?

tests/integrations/config/mod.rs
```diff
@@ -677,6 +677,8 @@ fn test_serde_custom_tikv_config() {
         region_split_keys: 100000,
         consistency_check_method: ConsistencyCheckMethod::Raw,
         perf_level: PerfLevel::EnableTime,
+        enable_region_bucket: false,
+        region_bucket_size: ReadableSize::gb(1),
```
bufferflies, line 681: Why the config default value(1G) is deferrent from DEFAULT_BUCKET_SIZE(96M).

youjiali1995: It tests whether the config is serialized and deserialized correctly, so its value should be different.

components/raftstore/src/coprocessor/split_check/mod.rs
tonyxuqqi: which part is useless here?
Do you mean SAMPLE_COUNT_PER_BUCKET? This is to mitigate the issue that that split keys are not appropriate when the SST files have overlapped ranges.

components/raftstore/src/coprocessor/config.rs
```diff
@@ -75,6 +81,8 @@ impl Default for Config {
             region_max_keys: SPLIT_KEYS / 2 * 3,
             consistency_check_method: ConsistencyCheckMethod::Mvcc,
             perf_level: PerfLevel::EnableCount,
+            enable_region_bucket: false,
+            region_bucket_size: DEFAULT_BUCKET_SIZE,
```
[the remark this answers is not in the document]
tonyxuqqi: What do you mean validate its range?  It's just the default value.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4390,6 +4393,10 @@ where
         self.register_pd_heartbeat_tick();
     }
 
+    fn on_refresh_region_buckets(&mut self, region_buckets: metapb::Buckets) {
```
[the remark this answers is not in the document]
tonyxuqqi: good catch

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -198,10 +197,35 @@ where
             "end_key" => log_wrappers::Value::key(&end_key),
         );
         CHECK_SPILT_COUNTER.all.inc();
-
+        let bucket_check_policy = policy;
         let mut host =
             self.coprocessor
                 .new_split_checker_host(region, &self.engine, auto_split, policy);
+        if bucket_check_policy == CheckPolicy::Approximate && host.enable_region_bucket() {
```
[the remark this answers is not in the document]
tonyxuqqi: You're right. Maybe host.policy() is better, though it does not work well either in big region with current code because only region_size >= host.cfg.region_max_size.0 will trigger CheckPolicy::approximate. Probably we need to change the code there too to enforce approximate when region size is large enough .

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -198,10 +197,35 @@ where
...
+                        "failed to get approximate bucket key";
+                        "region_id" => region_id,
+                    );
+                    vec![]
+                }
+            };
+            info!("starting approximate_bucket_keys {}", bucket_keys.len());
+            if !bucket_keys.is_empty() {
+                let mut region_buckets = Buckets::default();
+                region_buckets.set_keys(::protobuf::RepeatedField::from_vec(bucket_keys));
```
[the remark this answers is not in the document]
tonyxuqqi: ok.

components/raftstore/src/coprocessor/config.rs
Connor1996, line 85: should it consider the count of keys too?

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -186,8 +185,8 @@ where
         }
     }
 
-    /// Checks a Region with split checkers to produce split keys and generates split admin command.
-    fn check_split(&mut self, region: &Region, auto_split: bool, policy: CheckPolicy) {
+    /// Checks a Region with split and bucket checkers to produce split keys and buckets keys and generates split admin command.
+    fn check_split_and_bucket(&mut self, region: &Region, auto_split: bool, policy: CheckPolicy) {
```
Connor1996, line 189: should change `region_split_check_diff` when the region is enlarged, otherwise it would seldom trigger this check.

components/raftstore/src/store/worker/split_check.rs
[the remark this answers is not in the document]
Connor1996: should use the host.policy(), and change the condition `region_size >= host.cfg.region_max_size.0` when region size is enlarged.

components/raftstore/src/coprocessor/config.rs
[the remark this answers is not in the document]
youjiali1995: `region_bucket_size` should be less than `region_split_size`. You can validate it in `Config::validate()`.

components/raftstore/src/store/worker/split_check.rs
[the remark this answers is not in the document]
youjiali1995: Using `into()` or `Default::default()` has the benefit that once we switch from `rust-protobuf` to `prost`, we needn't change many codes.

components/raftstore/src/coprocessor/split_check/mod.rs
youjiali1995: `Checker::approximate_split_keys()` doesn't use `max_size` and `split_size`. It calls `get_approximate_split_keys()` with `batch_split_limit` directly.

components/raftstore/src/store/worker/split_check.rs
youjiali1995: In your implementation, you first collect bucket keys by approximate way and if the region has to be splitted by scan, bucket keys are also collected by scan, so you calculate bucket keys twice if a region has to be splitted by scan and the approximate calculation is unnecessary.

components/raftstore/src/coprocessor/config.rs
tonyxuqqi: Today region has max keys count and split keys count, so I agree that bucket should have it as well.

components/raftstore/src/store/worker/split_check.rs
Connor1996: better to calculate a reasonable default value for `region_split_check_diff`, no one want to change it.

components/raftstore/src/coprocessor/config.rs
```diff
@@ -95,6 +107,20 @@ impl Config {
                 self.region_split_keys
             ));
         }
+        if self.region_split_size.0 < self.region_bucket_size.0 {
```
BusyJay, line 110: And it should not be zero.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4545,6 +4551,23 @@ where
...
     }
 
+    fn on_refresh_region_buckets(
+        &mut self,
+        region_epoch: metapb::RegionEpoch,
+        region_buckets: metapb::Buckets,
+    ) {
+        let region = self.fsm.peer.region();
+        if util::is_epoch_stale(&region_epoch, region.get_region_epoch()) {
+            warn!(
```
BusyJay, line 4561: `info` should be enough as split/merge is expected behaviors.

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -196,12 +195,46 @@ where
...
+                    );
+                    vec![]
+                }
+            };
+            info!(
+                "starting approximate_bucket_keys {}, bucket {:?}",
+                bucket_keys.len(),
+                bucket_keys
+            );
+            if !bucket_keys.is_empty() {
```
5kbpers, line 223: Should we initialize the buckets with `start_key` and `end_key` if `bucket_keys` is empty?

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -196,12 +195,46 @@ where
...
+                bucket_keys.insert(0, region.get_start_key().to_vec());
+                bucket_keys.push(region.get_end_key().to_vec());
+                let mut region_buckets = Buckets::default();
+                region_buckets.set_keys(bucket_keys.into());
+                let _ = self.router.send(
+                    region.get_id(),
+                    CasualMessage::RefreshRegionBuckets {
```
5kbpers, line 230: Could this message be skipped if no bucket is changed?

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -196,12 +195,46 @@ where
...
+            info!(
+                "starting approximate_bucket_keys {}, bucket {:?}",
+                bucket_keys.len(),
+                bucket_keys
+            );
+            if !bucket_keys.is_empty() {
+                bucket_keys.insert(0, region.get_start_key().to_vec());
+                bucket_keys.push(region.get_end_key().to_vec());
+                let mut region_buckets = Buckets::default();
+                region_buckets.set_keys(bucket_keys.into());
```
5kbpers, line 227: The version of buckets should be updated.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4545,6 +4551,23 @@ where
         self.register_pd_heartbeat_tick();
     }
 
+    fn on_refresh_region_buckets(
+        &mut self,
+        region_epoch: metapb::RegionEpoch,
+        region_buckets: metapb::Buckets,
+    ) {
+        let region = self.fsm.peer.region();
+        if util::is_epoch_stale(&region_epoch, region.get_region_epoch()) {
```
5kbpers, line 4560: We need to check the version of buckets here.

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -85,8 +85,40 @@ impl<'a, E> Host<'a, E> {
...
+            let mut bucket_checker = size::Checker::new(
+                self.cfg.region_bucket_size.0, /* not used */
+                self.cfg.region_bucket_size.0, /* not used */
+                region_size / self.cfg.region_bucket_size.0,
+                CheckPolicy::Approximate,
+            );
+            let keys = box_try!(bucket_checker.approximate_split_keys(region, engine));
+            if !keys.is_empty() {
+                return Ok(keys);
+            }
```
youjiali1995, line 105:
```suggestion
            return bucket_checker.approximate_split_keys(region, engine);
```

components/raftstore/src/store/msg.rs
```diff
@@ -377,6 +377,10 @@ pub enum CasualMessage<EK: KvEngine> {
     RejectRaftAppend {
         peer_id: u64,
     },
+    RefreshRegionBuckets {
+        region_epoch: RegionEpoch,
+        region_buckets: metapb::Buckets,
```
youjiali1995, line 382: How about using `regions_buckets: Vec<Vec<u8>>,` instead and constructing the `metapb::Buckets` in `on_refresh_region_buckets()`? You can add the start/end key(now there are two places doing it) and set the bucket version here.

components/raftstore/src/store/fsm/peer.rs
youjiali1995: I prefer to `CasualMessage::RefreshRegionBuckets` only containing bucket keys(not including region's start/end key) and constructing the bucket here, and buckets versions is a `CLOCK_MONOTONIC_*` timestamp.

components/raftstore/src/coprocessor/config.rs
tonyxuqqi: done.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: done

components/raftstore/src/store/worker/split_check.rs
tonyxuqqi: For now this buckets generation logic will regenerate all buckets for each run.
But in next PR it will only update buckets having significant write qps. And if no such buckets, the message will be skipped.

components/raftstore/src/store/worker/split_check.rs
tonyxuqqi: good catch.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: > 
good point.

tests/integrations/config/mod.rs
```diff
@@ -678,6 +678,9 @@ fn test_serde_custom_tikv_config() {
         region_split_keys: 100000,
         consistency_check_method: ConsistencyCheckMethod::Raw,
         perf_level: PerfLevel::EnableTime,
+        enable_region_bucket: false,
```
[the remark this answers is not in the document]
tonyxuqqi: done

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4545,6 +4551,23 @@ where
...
+            warn!(
+                "receive a stale refresh region bucket message";
+                "region_id" => self.fsm.region_id(),
+                "peer_id" => self.fsm.peer_id(),
+            );
```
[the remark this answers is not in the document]
tonyxuqqi: good point.

components/raftstore/src/coprocessor/split_check/mod.rs
tonyxuqqi: good catch. I followed the pattern somewhere else, but apparently that was in a loop.

components/raftstore/src/store/msg.rs
tonyxuqqi: agree.

components/raftstore/src/store/worker/split_check.rs
tonyxuqqi: i think we should. Because the region size could be shrunk and we should update the buckets upon the shrink.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4545,6 +4551,31 @@ where
...
+                "peer_id" => self.fsm.peer_id(),
+                "epoch" => ?region_epoch,
+                "current_epoch" => ?region.get_region_epoch(),
+            );
+            return;
+        }
+        let mut region_buckets = Buckets::default();
+        let now = monotonic_raw_now();
+        const NANOSECONDS_PER_SECOND: u64 = 1_000_000_000;
+        region_buckets.version = (now.sec as u64) * NANOSECONDS_PER_SECOND + now.nsec as u64;
```
youjiali1995, line 4569: It's better to encapsulate a `timespec_to_ns()` function in `components/tikv_util/src/time.rs`.

youjiali1995 (approved):

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4545,6 +4551,31 @@ where
...
+        region_buckets.set_keys(bucket_keys.into());
+        // add region's start/end key
+        region_buckets
```
5kbpers, line 4572: Maybe using `VecDeque` for `buckets_keys` could reduce the cost of constructing `keys`.

5kbpers (commented):
rest LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -603,6 +603,8 @@ where
     persisted_number: u64,
     /// The context of applying snapshot.
     apply_snap_ctx: Option<ApplySnapshotContext>,
+    /// region buckets.
+    pub region_buckets: metapb::Buckets,
```
5kbpers, line 607:
```suggestion
    pub region_buckets: Buckets,
```

tonyxuqqi: done

5kbpers (approved):

youjiali1995 (approved):

PR #11764, opened 2021-12-31
rsmeter: deprecate single target

### LINKED ISSUE #11755: rsmeter: deprecate single target mode
TiKV depending on a diag node is anti-patten. We have introduced pub/sub mode to provide resource metering data, so the single target can be deprecated.

zhongzc:

Close #11755

What's Changed:

* Remove single target
* Alter tests

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
none
```


components/resource_metering/src/config.rs
```diff
@@ -20,9 +19,6 @@ const MIN_REPORT_RECEIVER_INTERVAL: ReadableDuration = ReadableDuration::millis(
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct Config {
-    /// Data reporting destination address.
```
breeswish, line 23: Prefer to keep it but not respect it. Otherwise for old config TiKV will throw errors.

breeswish:
Let's hold and merge it for release v5.5. In this way, there is one version for users to switch out.

PR #11766, opened 2021-12-31
*: support using raft engine in tests

tabokie:

Issue Number: ref #11119

What's Changed:

Support using raft engine in tests. Enable this feature by default.

### Check List 

Tests
- Unit test

### Release note 

```release-note
None
```


tests/integrations/raftstore/test_early_apply.rs
```diff
@@ -39,23 +38,23 @@ where
     let last_index = cluster.raft_local_state(1, 1).get_last_index();
     action(cluster);
     cluster.wait_last_index(1, 1, last_index + 1, Duration::from_secs(3));
-    let mut snaps = vec![(1, RocksSnapshot::new(cluster.get_raft_engine(1)))];
+    let mut rafts = vec![(1, cluster.get_raft_engine(1))];
```
5kbpers, line 41: Does it perform as a snapshot?

tabokie: No, I'll fix it by loading the data into memory.

BusyJay:
Once we have switched to raft engine, how can we also test against rocks engine?

tabokie:
@BusyJay Agree there are some perceivable behavioral differences between the two. I'll make it generic and controlled by feature gate.

tabokie:
@BusyJay It's done.

components/engine_test/src/lib.rs
```diff
@@ -24,15 +25,16 @@
 //! The engine for each module is chosen at compile time with feature flags:
 //!
 //! - `--features test-engine-kv-rocksdb`
-//! - `--features test-engine-raft-rocksdb`
```
5kbpers, line 27: Shouldn't we keep it?

tabokie: It's moved down a bit.

BusyJay:
> I'll make it generic and controlled by feature gate.

Actually what I'm concerned is how to ensure rocks engine and raft engine both work. In my opinion, we should guarantee raft engine work well enough that we can't drop the support of rocks engine. Otherwise, all existing test cases may need to be run twice.

tabokie:
> Actually what I'm concerned is how to ensure rocks engine and raft engine both work. In my opinion, we should guarantee raft engine work well enough that we can't drop the support of rocks engine. Otherwise, all existing test cases may need to be run twice.

You mean "we can drop the support of rocks engine" right? Personally I would run tests with all engine variants regardless of whether they are officially supported or not. That way any unsound dependency on engine's private behavior can be detected early.

But if we can only run tests with one variant (given the resource budget), then there is no other option than to run it against the default one. Maybe run other variants in master branch might help?

And for your question, there's no guarantee that Raft Engine can win over RocksDB in every aspect, even though our limited testing indicates so.

BusyJay:
> And for your question, there's no guarantee that Raft Engine can win over RocksDB in every aspect, even though our limited testing indicates so.

But there is no regression in every known aspect right? Maintaining two engines is hard no matter code complexity or tests resources.

tabokie:
> But there is no regression in every known aspect right? Maintaining two engines is hard no matter code complexity or tests resources.

No known regression. But there are a few "by-design" differences:
- Space usage: RocksDB can take less space than Raft Engine under default configuration.
- Read performance: Raft Engine has compression and no internal cache. That makes reads much slower than RocksDB (100ms vs 100us). Issues like #11320 are more prominent with Raft Engine.

tabokie:
Ping @BusyJay @5kbpers, could we proceed with this one? Hit a bug (#12162) due to lacking integration tests.

tests/integrations/server/kv_service.rs
```diff
@@ -706,15 +709,14 @@ fn test_debug_raft_log() {
     // Put some data.
     let engine = cluster.get_raft_engine(store_id);
     let (region_id, log_index) = (200, 200);
-    let key = keys::raft_log_key(region_id, log_index);
     let mut entry = eraftpb::Entry::default();
     entry.set_term(1);
-    entry.set_index(1);
```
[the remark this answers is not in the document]
tabokie: You mean before? I guess we didn't verify the entry in this test. But Raft Engine will extract log index from the entry internally.

components/test_raftstore/Cargo.toml
```diff
@@ -32,6 +39,7 @@ pd_client = { path = "../pd_client", default-features = false }
 protobuf = { version = "2.8", features = ["bytes"] }
 raft = { version = "0.7.0", default-features = false, features = ["protobuf-codec"] }
 raftstore = { path = "../raftstore", default-features = false }
+raft_log_engine = { path = "../raft_log_engine", default-features = false }
```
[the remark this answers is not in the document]
tabokie: Stale code. Removed.

tests/failpoints/cases/test_async_fetch.rs
```diff
@@ -84,20 +84,20 @@ fn test_node_cache_compact_with_one_node_down() {
...
     // logs should be replicated to node 1 successfully.
-    for i in 1..60 {
+    for i in 1..60u32 {
         let k = i.to_string().into_bytes();
         let v = k.clone();
         must_get_equal(cluster.engines[&1].kv.as_inner(), &k, &v);
     }
 
-    for i in 60..200 {
+    for i in 60..200u32 {
```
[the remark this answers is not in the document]
tabokie: The type can't be inferred. I don't remember the rust version I got the error from. Can't reproduce now.

tests/integrations/raftstore/test_early_apply.rs
```diff
@@ -150,20 +165,23 @@ fn test_update_internal_apply_index() {
...
     cluster.clear_send_filters();
     must_get_equal(&cluster.get_engine(1), b"k2", b"v2");
     must_get_equal(&cluster.get_engine(2), b"k2", b"v2");
 
     // Simulate data lost in raft cf.
-    for (id, snap) in &snaps {
-        cluster.stop_node(*id);
-        cluster.restore_raft(1, *id, snap);
-        cluster.run_node(*id).unwrap();
+    for (id, mut batch) in snaps.into_iter() {
```
BusyJay, line 178:
```suggestion
    for (id, mut batch) in snaps {
```

components/test_raftstore/src/util.rs
```diff
@@ -1162,29 +1195,51 @@ pub fn get_tso(pd_client: &TestPdClient) -> u64 {
...
+    key: &[u8],
+) -> M {
+    engines
+        .kv
+        .get_msg_cf(CF_RAFT, key)
+        .unwrap()
+        .unwrap_or_default()
+}
+
+pub fn is_compacted(
```
BusyJay, line 1209: `check_compacted` is better as it can panic inside.

tests/integrations/raftstore/test_compact_log.rs
```diff
@@ -89,11 +88,23 @@ fn test_compact_count_limit<T: Simulator>(cluster: &mut Cluster<T>) {
         let v2 = cluster.get(&k);
         assert_eq!(v2, Some(v));
 
-        if i > 100 && check_compacted(&cluster.engines, &before_states, 1) {
+        if i > 100
+            && is_compacted(
+                &cluster.engines,
+                &before_states,
+                1,
+                true, /*must_compacted*/
```
BusyJay, line 96: You mean `false`?

scripts/check-build-opts.py
```diff
@@ -38,19 +38,39 @@ def cargo_check_default():
...
+        if has_test_engine_features:
+            args += ["--features", f"test-engines-{engines}"]
+
+        run_and_collect_errors(args)
+
+def cargo_check_test_engines_ext(kv_engine, raft_engine):
+    cargo_run_test_engines("check", [], kv_engine, raft_engine)
+
+def cargo_test_test_engines_ext(kv_engine, raft_engine):
+    cargo_run_test_engines("test", ["--no-run"], kv_engine, raft_engine)
```
BusyJay, line 71: You mean `cargo_run_test_engines_ext`?

tests/integrations/raftstore/test_early_apply.rs
tabokie: The batch are drained, I think into is better.

BusyJay: Actually, `for a in vec` works like `let iter = vec.into_iter(); while let Some(i) = iter.next()`. For loop only accepts types implements `IntoIterator` trait.

tabokie: Oh, right.

scripts/check-build-opts.py
```diff
@@ -38,19 +38,39 @@ def cargo_check_default():
...
+
+        args = ["cargo", cmd, "-p", crate, "--no-default-features"]
+        args += extra_args
+        if has_test_engine_features:
+            args += ["--features", f"test-engines-{engines}"]
+
+        run_and_collect_errors(args)
+
+def cargo_check_test_engines_ext(kv_engine, raft_engine):
+    cargo_run_test_engines("check", [], kv_engine, raft_engine)
```
BusyJay, line 68: You mean `cargo_run_test_engines_ext`?

components/test_raftstore/src/util.rs
```diff
@@ -130,6 +129,25 @@ pub fn must_region_cleared(engine: &Engines<RocksEngine, RocksEngine>, region: &
     );
 }
 
+pub fn dump_raft_entries(
```
BusyJay, line 132: The name is not correct any more.

components/test_raftstore/src/util.rs
```diff
@@ -633,22 +651,33 @@ pub fn create_test_engine(
...
+    #[allow(clippy::redundant_clone)]
+    let env = get_env(key_manager.clone(), limiter.clone()).unwrap();
     let cache = cfg.storage.block_cache.build_shared_cache();
 
-    let raft_path = dir.path().join("raft");
-    let raft_path_str = raft_path.to_str().unwrap();
-
-    let mut raft_db_opt = cfg.raftdb.build_opt();
-    raft_db_opt.set_env(env.clone());
+    #[cfg(feature = "test-engine-raft-rocksdb")]
```
BusyJay, line 658: Is it possible to reuse `ConfiguredRaftEngine`?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2118,7 +2118,7 @@ mod tests {
     ) -> PeerStorage<KvTestEngine, RaftTestEngine> {
         let mut store = new_storage(region_scheduler, raftlog_fetch_scheduler, path);
         let mut write_task = WriteTask::new(store.get_region_id(), store.peer_id, 1);
-        store.append(ents[1..].to_vec(), &mut write_task);
+        store.append(ents[0..].to_vec(), &mut write_task);
```
BusyJay, line 2121: Why? The first entry is just a placeholder.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2197,6 +2199,7 @@ mod tests {
         }
     }
 
+    // Does not include raftdb state key.
```
BusyJay, line 2202: What if new states are added to raft engine? For example, we may move all states from KV engine to raft engine in the future.

components/raftstore/src/store/peer_storage.rs
tabokie: I don't understand

BusyJay: The first entry don't need to be stored. Only its index and term are used to initialized truncated state. The test logic is from tikv/raft-rs.

tabokie: Seems weird to me.. But I'll restore it.

tabokie: Actually, it won't work with raft engine. Raft Engine's `first_index()` is directly deduced from the stored entries.

BusyJay: So there will be at least one entry left even if all of them are deleted by raft? What if it catches logs by snapshot? There will be no stored entries.

tabokie: Turns out `first_index` is not a trait interface. I'll try restoring them again.

tabokie: It doesn't work. In some test cases we try to append [3,5] entries to a region of `new_storage_from_ents([3,5])`, but Raft Engine will panic if we append [3,5] to [4,5].

components/raftstore/src/store/snap.rs
```diff
@@ -577,7 +577,7 @@ impl Snapshot {
     }
 
     fn get_display_path(dir_path: impl AsRef<Path>, prefix: &str) -> String {
-        let cf_names = "(".to_owned() + &SNAPSHOT_CFS.join("|") + ")";
+        let cf_names = "(".to_owned() + SNAPSHOT_CFS.join("|").as_str() + ")";
```
BusyJay, line 580: Why the change?

components/raft_log_engine/Cargo.toml
```diff
@@ -19,5 +19,5 @@ serde = "1.0"
 serde_derive = "1.0"
 kvproto = { git = "https://github.com/pingcap/kvproto.git" }
 raft = { version = "0.7.0", default-features = false, features = ["protobuf-codec"] }
-raft-engine = "0.1.0"
+raft-engine = { git = "https://github.com/tikv/raft-engine.git" }
```
BusyJay, line 22: How about bumping a patch version?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2118,7 +2118,7 @@ mod tests {
     ) -> PeerStorage<KvTestEngine, RaftTestEngine> {
         let mut store = new_storage(region_scheduler, raftlog_fetch_scheduler, path);
         let mut write_task = WriteTask::new(store.get_region_id(), store.peer_id, 1);
-        store.append(ents[1..].to_vec(), &mut write_task);
+        store.append(ents.to_vec(), &mut write_task);
```
BusyJay, line 2121: Better change the case instead of changing it here as it makes no sense to test impossible operations.

tabokie: What do you mean? I didn’t know the entries stored in raftdb are inconsistent with raft state by nature.

components/raft_log_engine/Cargo.toml
tabokie: I think we should use latest raft engine to catch bugs early. The patch version will be release in sync with 6.1.

components/raftstore/src/store/peer_storage.rs
BusyJay: What I mean is remove all entries prior to (3, 3) in test cases. entries should be consistent with raft state and no entries prior to truncated state can be appended.

components/raft_log_engine/Cargo.toml
BusyJay: This will be slow down compilation. One goal of tirocks is to skip git pulling for every fresh build.

components/raftstore/src/store/snap.rs
tabokie: I don't even know; some compile errors just pop up even though master works fine.

tests/integrations/raftstore/test_early_apply.rs
```diff
@@ -150,20 +165,23 @@ fn test_update_internal_apply_index() {
...
     // Simulate data lost in raft cf.
-    for (id, snap) in &snaps {
-        cluster.stop_node(*id);
-        cluster.restore_raft(1, *id, snap);
-        cluster.run_node(*id).unwrap();
+    for (id, mut batch) in snaps {
+        cluster.stop_node(id);
+        cluster
+            .get_raft_engine(id)
+            .consume(&mut batch, true /*sync*/)
```
BusyJay, line 182: Extra entries are not deleted.

components/test_raftstore/src/util.rs
```diff
@@ -112,16 +115,13 @@ pub fn must_region_cleared(engine: &Engines<RocksEngine, RocksEngine>, region: &
...
     }
-    let log_min_key = keys::raft_log_key(id, 0);
-    let log_max_key = keys::raft_log_key(id, u64::MAX);
+
     engine
         .raft
-        .scan(&log_min_key, &log_max_key, false, |k, v| {
-            panic!("[region {}] unexpected log ({:?}, {:?})", id, k, v);
-        })
+        .scan_entries(id, |_| panic!("[region {}] unexpected entry", id))
```
BusyJay, line 121: Why not just `dump_data().is_empty()`?

components/raftstore/src/store/snap.rs
```diff
@@ -1791,24 +1802,22 @@ pub mod tests {
 
     #[test]
     fn test_empty_snap_file() {
-        test_snap_file(open_test_empty_db, None);
-        test_snap_file(open_test_empty_db, Some(gen_db_options_with_encryption()));
```
BusyJay, line 1795: Why remove this?

components/raftstore/src/store/async_io/write_tests.rs
```diff
@@ -234,10 +237,9 @@ fn test_worker() {
     let engines = new_temp_engine(&path);
     let mut t = TestWorker::new(&Config::default(), &engines);
 
-    let mut task_1 = WriteTask::<KvTestEngine, KvTestEngine>::new(1, 1, 10);
+    let mut task_1 = WriteTask::<KvTestEngine, RaftTestEngine>::new(1, 1, 10);
     init_write_batch(&engines, &mut task_1);
     put_kv(&mut task_1.kv_wb, b"kv_k1", b"kv_v1");
-    put_kv(&mut task_1.raft_wb, b"raft_k1", b"raft_v1");
```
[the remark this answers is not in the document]
tabokie: I think that will be contradicting with task.raft_state,

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2229,19 +2231,29 @@ mod tests {
         store
             .engines
             .raft
-            .scan(&raft_start, &raft_end, false, |_, _| {
+            .scan_entries(region_id, |_| {
```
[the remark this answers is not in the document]
tabokie: Different engine has different concept of "key". I think it's not good to have it as a trait function.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2229,19 +2231,29 @@ mod tests {
...
+        }
+
         count
     }
 
     #[test]
     fn test_storage_clear_meta() {
         let worker = Worker::new("snap-manager").lazy_build("snap-manager");
-        let cases = vec![(0, 0), (5, 1)];
+        let cases = vec![(0, 0), (3, 0)];
```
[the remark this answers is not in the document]
tabokie: RocksDB clears entries one by one, while raft engine clears them all at once. So it's impossible (and unnecessary) to test a partial `clear_meta`.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -3306,35 +3298,32 @@ mod tests {
...
-            .raft
-            .put_msg(&log_key, &new_entry(20, RAFT_INIT_LOG_TERM))
-            .unwrap();
         raft_state.set_last_index(20);
-        engines.raft.put_msg(&raft_state_key, &raft_state).unwrap();
+        let entries = (12..=20)
+            .map(|index| new_entry(index, RAFT_INIT_LOG_TERM))
+            .collect();
+        engines.raft.append(1, entries).unwrap();
+        raft_state.mut_hard_state().set_commit(12);
```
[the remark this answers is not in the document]
tabokie: Doesn't hurt to be more specific.

components/test_raftstore/src/util.rs
tabokie: Cannot identify if it's entry or state.

components/raftstore/src/store/snap.rs
tabokie: I adjust the testing code to match how we actually use encryption in tikv. After that, this test logic doesn't work anymore. I believe we shouldn't call `Snapshot::new_xx` directly to make sure encryption works consistently.

components/raftstore/src/store/peer_storage.rs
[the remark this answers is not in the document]
BusyJay: It's confusing.

components/raftstore/src/store/async_io/write_tests.rs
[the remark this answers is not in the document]
BusyJay: Then how to know "if raft_wb will be merged correctly"?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2846,28 +2862,6 @@ mod tests {
     fn test_storage_append() {
         let ents = vec![new_entry(3, 3), new_entry(4, 4), new_entry(5, 5)];
         let mut tests = vec![
-            (
-                vec![new_entry(3, 3), new_entry(4, 4), new_entry(5, 5)],
```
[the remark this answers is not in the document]
BusyJay: It applies to the other cases as well.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -2847,26 +2863,16 @@ mod tests {
...
         let mut tests = vec![
             (
-                vec![new_entry(3, 3), new_entry(4, 4), new_entry(5, 5)],
-                vec![new_entry(4, 4), new_entry(5, 5)],
+                vec![new_entry(4, 6), new_entry(5, 6)],
+                vec![new_entry(4, 6), new_entry(5, 6)],
             ),
             (
-                vec![new_entry(3, 3), new_entry(4, 6), new_entry(5, 6)],
+                vec![new_entry(4, 6), new_entry(5, 6)],
```
BusyJay, line 2870: This is the same as last case.

BusyJay (approved):

tabokie:
Verified RocksDB tests still work (at least for now) in https://github.com/tikv/tikv/pull/11766/commits/661da35b9e424b4abca05d5475dab4f4ed058fbc.

tabokie:
Ping @5kbpers @tier-cap

5kbpers (approved):

PR #11772, opened 2022-01-04
raftclient: enable delay only in high load

### LINKED ISSUE #11769: 21% tps degradation on ycsb workloada from commit 74cd8a

v5.4.0

x86-64
### Steps to reproduce

1

```
    export AWS_ACCESS_KEY_ID=xyz AWS_SECRET_ACCESS_KEY=xyz
    br restore db --db=test --pd "pd-peer:2379" \
        --storage "s3://benchmark/ycsb-100m-v5.0.0" \
        --s3.endpoint "http://xyz" \
        --send-credentials-to-tikv=true \
                --checksum-concurrency 64 \
                --check-requirements=false

go-ycsb run mysql -P /ycsb/workloads/workloada -p mysql.host=tiup-peer -p mysql.port=3390 -p recordcount=100000000 -p operationcount=50000000 -p workload=core -p fieldcount=10 -p fieldlength=100 -p threadcount=500 -p requestdistribution=zipfian
```

compared the ycsb workoada between 74cd8ae2a9ff930460cba15acdf1847726eb6db8 and the previous commit
### What did you expect?
no performance degradation

### What did happened?
21% tps degradation

BusyJay:

Close #11769 

What's Changed:

#11705 adds delay to raft client, which can reduce syscalls and improve CPU efficiency under high load. But it turns out this can lead to significant regression in normal load. So this PR adds CPU stats and only enabling delay in high load.

This is also the same strategy used in v4.x.

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

### Release note 

```release-note
None
```


BusyJay:
I have been tested the PR to see whether the improvement still works. But I can't reproduce the regression reported in https://github.com/tikv/tikv/issues/11309. But the regression reported in #11769 was resolved. So I tend to merge this PR first. The strategy is configurable, if we reproduce the regression at some point, we can still adjust the configuration.

src/server/load_statistics/linux.rs
```diff
@@ -13,27 +16,27 @@ pub struct ThreadLoadStatistics {
...
     cur_pos: usize,
-    cpu_usages: Vec<f64>,
+    cpu_usages: Vec<HashMap<Pid, f64>>,
     instants: Vec<Instant>,
-    thread_load: Arc<ThreadLoad>,
+    thread_loads: Arc<ThreadLoadPool>,
 }
 
 impl ThreadLoadStatistics {
     /// Create a thread load statistics for all threads with `prefix`. `ThreadLoad` is stored into
```
gengliqi, line 25: The comment should be updated.

src/server/load_statistics/linux.rs
```diff
@@ -56,29 +59,45 @@ impl ThreadLoadStatistics {
...
         }
         let current_instant = self.instants[self.cur_pos];
-        let current_cpu_usage = self.cpu_usages[self.cur_pos];
+        let current_cpu_usages = &self.cpu_usages[self.cur_pos];
 
         let next_pos = (self.cur_pos + 1) % self.slots;
         let earlist_instant = self.instants[next_pos];
-        let earlist_cpu_usage = self.cpu_usages[next_pos];
+        let earlist_cpu_usages = &self.cpu_usages[next_pos];
         self.cur_pos = next_pos;
```
gengliqi, line 76: I don't read this code before. Seems the calculation of CPU usage is only related to the previous record. So 2 slot num should be enough?

src/server/config.rs
```diff
@@ -358,6 +353,12 @@ impl Config {
             ));
         }
 
+        if self.heavy_load_threshold > 100 {
+            // The configuration has been changed to describe CPU usage of a single thread instead
+            // of all threads. So migrate from the old style.
+            self.heavy_load_threshold /= 4;
```
gengliqi, line 359: Seems `heavy_load_threshold` is still possible greater than 100.

src/server/raft_client.rs
```diff
@@ -450,12 +457,19 @@ where
...
+                        if Pin::new(timeout).poll(ctx).is_pending() {
+                            return Poll::Pending;
+                        } else {
+                            RAFT_MESSAGE_FLUSH_COUNTER.timeout.inc_by(1);
+                        }
+                    } else {
+                        RAFT_MESSAGE_FLUSH_COUNTER.eof.inc_by(1);
                     }
+                } else if s.flush_timeout.is_some() {
+                    RAFT_MESSAGE_FLUSH_COUNTER.full_after_timeout.inc_by(1);
```
NingLin-P, line 470: Seems `full_after_delay` are more suitable.

src/server/load_statistics/linux.rs
BusyJay: It's a sliding window to calculate a smooth usage. /cc @hicqu

src/server/load_statistics/linux.rs
```diff
@@ -56,29 +60,45 @@ impl ThreadLoadStatistics {
...
         let millis = (current_instant - earlist_instant).as_millis() as usize;
+        let mut total_usage = 0;
         if millis > 0 {
-            let mut cpu_usage = calc_cpu_load(millis, earlist_cpu_usage, current_cpu_usage);
-            if cpu_usage > self.tids.len() * 100 {
-                cpu_usage = self.tids.len() * 100;
+            let mut loads = self.thread_loads.stats.lock();
+            for tid in current_cpu_usages.keys() {
+                loads.entry(*tid).or_default();
+            }
```
NingLin-P, line 85: Since `ThreadLoadStatistics::tids` won't be changed, we can determine what should store at `ThreadLoadPool::stats` in `ThreadLoadStatistics::new`, how about moving these codes to there?

NingLin-P (approved):

PR #11782, opened 2022-01-05
raftstore: skip deleting snapshot files in peer pending_remove is true

### LINKED ISSUE #11746: raftstore: snapshot files may deleted while the peer state is still `PeerState::Applying`

Matser

Not related

### Steps to reproduce

A peer with the peer state set to `PeerState::Applying` is started the apply snapshot process. If before it actually starts to apply snapshot data, it received a message (gc message or raft msg with larger peer id) telling it to destroy itself, the peer will try to cancel applying snapshot and submit the change of peer state (to `PeerState::Tombstone`) to the `WriteBatch`.

At the meantime, a canceled snapshot will be gc by the `StoreTick::SnapGc` tick, which will delete the snapshot files immediately and the `WriteBatch` may not persist yet. If TiKV then crashes and recover, the peer state is still `PeerState::Applying` and TiKV will panic because the snapshot does not exist.

This case can happen in scenarios like a TiKV is down while there are regions applying snapshot, and recover after a long time, by then these regions will try to continue to apply snapshot and the leader will tell them to gc themself, at this time the TiKV is crashed.

### What did you expect?

Not panic

### What did happened?

Panic because snapshot files not exist

tonyxuqqi:
skip deleting snapshot files in peer pending_remove is true and the snapshot is being applied and canceled -- close #11746

This is to avoid the potential panic when the snapshot files are deleted, but the peer's status (Tombstone) is not persisted in disk due to tikv crash.

Close #11746 

What's Changed:
Skip deleting snapshot files in peer.rs when the snapshot is being applied and just canceled and the peer is pending removal. 
### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
Fix a potential panic (#11746) when snapshot files have been deleted but the peer's status is still Applying.
```


tests/failpoints/cases/test_snap.rs
```diff
@@ -240,9 +249,27 @@ fn test_destroy_peer_on_pending_snapshot() {
 
     fail::remove(before_handle_normal_3_fp);
 
+    assert_eq!(pending_remove_is_triggered.load(Ordering::Acquire), true);
```
NingLin-P, line 252: It may possible that the peer is destroyed before the gc snap tick is triggered, how about `sleep_ms` between `fail::remove(apply_snapshot_fp)` and `fail::remove(before_handle_normal_3_fp)`?

NingLin-P (commented):
Rest LGTM

tests/failpoints/cases/test_snap.rs
tonyxuqqi: So before snap_mgr_gc_tick_interval, we have sleep_ms(100),  is it not enough to run a gc snap (Right now the snap_mgr_gc_tick_interval is 50ms) ?

Added 50ms anyway in case.

NingLin-P: Snapshot gc is start only after `fail::remove(apply_snapshot_fp)`

NingLin-P (approved):

tests/failpoints/cases/test_snap.rs
```diff
@@ -231,18 +231,46 @@ fn test_destroy_peer_on_pending_snapshot() {
     let before_handle_normal_3_fp = "before_handle_normal_3";
     fail::cfg(before_handle_normal_3_fp, "pause").unwrap();
 
+    let pending_remove_is_triggered = Arc::new(AtomicBool::new(false));
```
BusyJay, line 234: If it's a channel, then L250 doesn't need to sleep an arbitrary time.

tonyxuqqi: good point.will fix it.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1033,6 +1036,8 @@ where
                     }
                 };
                 self.ctx.snap_mgr.delete_snapshot(&key, a.as_ref(), false);
+            } else if self.fsm.peer.pending_remove {
+                fail_point!("pending_remove_is_true");
```
BusyJay, line 1040: The case is tricky. It checks whether `pending_remove` is checked rather than checking whether it works well if `pending_remove` is skipped. In my opinion, whether `pending_remove` is checked is not important, it's just an implementation details. What the case should check is if destroying a peer is delayed for quite a long time, raftstore can still work as usual.

tonyxuqqi: Well, this is unit-test/integration test and check some implementation details is normal--otherwise how to differentiate if the snapshot file is deleted by peer's gc or store's gc. Using other fail_points to disable/enable store gc are also implementation details. 

From user's perspective,  it checked if snapshot files are finally deleted and peer will be destroyed and re-created successfully. This is what user cares.   And the check of is "pending_remove_is_true" is kind of white-box test that snapshot file is not deleted by peer's snap gc. 
The whole change's goal is just to prevent snapshot files being deleted when peer is being destroyed and after it's destroyed, they should be deleted. 
Regarding destroying a peer is delayed or not,  and if raftstore still work, it's still within existing code's 100ms range and covered by existing code.

BusyJay: > otherwise how to differentiate if the snapshot file is deleted by peer's gc or store's gc

There is no difference. So when gc is tested, it should check whether files will be deleted at the end instead of checking whether it's deleted in peer gc or store gc.

> Using other fail_points to disable/enable store gc are also implementation details.

I'm not sure what's you refer to. But disable/enable gc is more like a configuration change to me.

> The whole change's goal is just to prevent snapshot files being deleted when peer is being destroyed and after it's destroyed, they should be deleted.

Yes, so it should check the consequence of deleting snapshots while destroying a peer. For example, if destroy peer is skipped for a given time, the snapshot files should not be deleted; or if a peer is being destroy, gc has happened before destroy state is persisted, the node should be able to be restarted and resumed.

Utilizing implementation details for timing is OK, but correctness depends on the details is bad. The goal of writing a unit test is not just to ensure current implementation is correct, it also checks refactoring is correct. So correctness should depend on the behavior instead of whether a field is used or not.

tests/failpoints/cases/test_snap.rs
```diff
@@ -237,14 +237,96 @@ fn test_destroy_peer_on_pending_snapshot() {
...
+    cluster.clear_send_filters();
+    // Wait for leader send snapshot.
+    sleep_ms(100);
+
+    // Don't send check stale msg to PD
+    let peer_check_stale_state_fp = "peer_check_stale_state";
+    fail::cfg(peer_check_stale_state_fp, "return()").unwrap();
+
+    pd_client.must_remove_peer(r1, new_peer(3, 3));
+    pd_client.must_add_peer(r1, new_peer(3, 4)); // Without it, pd_client.must_remove_peer does not trigger destroy_peer!
```
BusyJay, line 290: That's because when a peer is applying snapshot, leader will not replicate new logs to it to avoid OOM.

tests/failpoints/cases/test_snap.rs
```diff
@@ -237,14 +237,96 @@ fn test_destroy_peer_on_pending_snapshot() {
...
+
+    let before_handle_normal_3_fp = "before_handle_normal_3";
+    fail::cfg(before_handle_normal_3_fp, "pause").unwrap(); // to pause ApplyTaskRes::Destroy so that peer gc could finish
```
BusyJay, line 293: In TiKV, we use the style that comment is put in the lines above statements.

BusyJay (approved):

tests/failpoints/cases/test_snap.rs
tonyxuqqi: Will fix it. Thanks.

sticnarf:
Does this PR need to be backported to release branches?

BusyJay:
The change is quite small, I suggest to also pick it to 5.x.

PR #11785, opened 2022-01-05
*: unify thread related native APIs

### LINKED ISSUE #11784: duplicated codes handling threads
## Development Task

tikv_utils provides methods to access threads stats,

https://github.com/tikv/tikv/blob/master/components/tikv_util/src/metrics/threads_linux.rs

while resource_metering also provides a bunch of methods for the same purpose:

https://github.com/tikv/tikv/blob/master/components/resource_metering/src/utils.rs

They should be merged and use only one set.

BusyJay:

Close #11784.

What's Changed:

Now there are duplicated codes calling syscalls to fetch thread
informations. This PR provides unified APIs and remove duplicated codes.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/file_system/src/iosnoop/biosnoop.rs
```diff
@@ -167,7 +168,7 @@ pub fn init_io_snooper() -> Result<(), String> {
         }
     }
 
-    let code = include_str!("biosnoop.c").replace("##TGID##", &nix::unistd::getpid().to_string());
+    let code = include_str!("biosnoop.c").replace("##TGID##", &thread::process_id().to_string());
```
zhongzc, line 171: L65

```rust
            let tid = nix::unistd::gettid().as_raw() as u32;
```

disksing:
I'm not very familar with these code. Better to assign another reviewer. @BusyJay

breeswish (approved):
Cool change! @mornyx PTAL

mornyx:
Great changes~ In fact, the implementation of non-linux platform is incomplete and cannot work properly. If we extract it as a public component, I think it's better to explain it in comments to avoid misuse of non-linux platforms.

mornyx:
LGTM

BusyJay:
Module doc is added to declare platform compatibility.

/merge

PR #11788, opened 2022-01-05
raftstore: Handle apply tasks from leaders with higher priority

Connor1996:

Part of  https://github.com/tikv/tikv/pull/11409 
What's Changed:
Use a separate channel to send apply tasks of leaders, and apply worker polls the channel first.

### Related changes

- None

### Check List 

Tests 

### Release note 

```release-note
None
```

### DIFF
--- a/components/batch-system/src/batch.rs
+++ b/components/batch-system/src/batch.rs
@@ -81,6 +81,7 @@ macro_rules! impl_sched {
     };
 }
 
+impl_sched!(HighPriNormalScheduler, FsmTypes::Normal, Fsm = N);
 impl_sched!(NormalScheduler, FsmTypes::Normal, Fsm = N);
 impl_sched!(ControlScheduler, FsmTypes::Control, Fsm = C);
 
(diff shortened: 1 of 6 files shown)


Connor1996:
Though didn't find much difference under sysbench workload

NingLin-P:
> Though didn't find much difference under sysbench workload

IMO, as long as leader and follower share the same io thread, it won't have too much improvement even let the leader's task has higher priority to be handled.

With well scheduling, there should always have more follower tasks than the leader's (namely `Leader:Follower = 1:2` with 3 replicas). Suppose tasks handle in the order of `LLFFFF`, the first two leader tasks `LL` may respond to the client very quickly, but later coming leader task may have to wait for the four follower task `FFFF` to finish before being handled, in other words the latency transfer to apply wait duration.

Connor1996:
> > Though didn't find much difference under sysbench workload
> 
> IMO, as long as leader and follower share the same io thread, it won't have too much improvement even let the leader's task has higher priority to be handled.
> 
> With well scheduling, there should always have more follower tasks than the leader's (namely `Leader:Follower = 1:2` with 3 replicas). Suppose tasks handle in the order of `LLFFFF`, the first two leader tasks `LL` may respond to the client very quickly, but later coming leader task may have to wait for the four follower task `FFFF` to finish before being handled, in other words the latency transfer to apply wait duration.

Yes, but high follower apply wait duration doesn't affect grpc message duration.

NingLin-P:
> Yes, but high follower apply wait duration doesn't affect grpc message duration.

Leader & follower share the same io thread, apply wait duration should affect both of them.

PR #11792, opened 2022-01-05
logger: invoke logger_compatible_adjust before run_tikv (#11651)

### LINKED ISSUE #11789: tikv have not logs saved in k8s
/ # ./tikv-server -V
TiKV
Release Version: 5.4.0-alpha
Edition: Community
Git Commit Hash: 99b3436
Git Commit Branch: heads/refs/tags/v5.4.0-nightly
UTC Build Time: 2022-01-04 01:15:55
Rust Version: rustc 1.56.0-nightly (2faabf579 2021-07-27)
Enable Features: jemalloc mem-profiling portable sse test-engines-rocksdb cloud-aws cloud-gcp cloud-azure
Profile: dist_release

8core 16G

### Steps to reproduce
no matter

### What did you expect?
tikv logs can be saved 

### What did happened?
tikv have not logs saved in k8s 
image

ethercflow:
close #11789 

Close #11789  

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
None
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2557,7 +2557,7 @@ impl Default for TiKvConfig {
             log_level: slog::Level::Info,
             log_file: "".to_owned(),
             log_format: LogFormat::Text,
-            log_rotation_timespan: ReadableDuration::hours(24),
+            log_rotation_timespan: ReadableDuration::hours(0),
             log_rotation_size: ReadableSize::mb(300),
             slow_log_file: "".to_owned(),
             slow_log_threshold: ReadableDuration::secs(1),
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -2787,6 +2787,59 @@ impl TiKvConfig {
         Ok(())
     }
 
+    // As the init of `logger` is very early, this adjust needs to be separated and called
```
5kbpers, line 2790: Can we add a test for this case?

5kbpers (approved):

ethercflow:
This PR is to fix a bug: when `log-file` is not set in the command line, `log-file` is set in the configuration file, but `log.file.filename` is not configured, and the log information is printed to the terminal.

The reason is that the log module uses log parameters for initialization earlier than the compatibility adjustment of the parameters.

gengliqi (approved):

src/config.rs
```diff
@@ -2780,6 +2780,59 @@ impl TiKvConfig {
...
+    // immediately after parsing the command line.
+    pub fn logger_compatible_adjust(&mut self) {
+        let default_tikv_cfg = TiKvConfig::default();
+        let default_log_cfg = LogConfig::default();
+        if self.log_level != default_tikv_cfg.log_level {
+            println!("deprecated configuration, log-level has been moved to log.level");
```
BusyJay, line 2789: You should not use `println!`.

If the logger is initialized, then you should use `info!` instead, if it's not, `eprintln!` should be used instead.

ethercflow: Sorry, I missed email notification. I'll submit another PR to fix this. Thanks!

Issue #11542, opened 2021-12-02
Add functionalities that helps detecting potential data inconsistency


MyonKeminta:
## Development Task

This is part of https://github.com/pingcap/tidb/issues/26833 , and there needs an issue in TiKV repo that the PRs' can refer to in the commit messages. 😕

The TiKV part is to support an assertion mechanism that, during  a prewrite request, asserts a key exists or not exists before, and when an assertion is violated, it must indicates that there's some bug.

Related PRs:

* https://github.com/tikv/tikv/pull/11516
* https://github.com/tikv/tikv/pull/11518
* https://github.com/tikv/tikv/pull/10977

ekexium:
Can we close it?

MyonKeminta:
> Can we close it?

I think yes.

PR #11799, opened 2022-01-06
txn: Do not check assertion if the mutation is insert or check_not_exist

MyonKeminta:

Ref #11542

This PR avoids doing any assertion when the mutation is a `Mutation::Insert` or a `Mutation::CheckNotExist`. These two kinds of mutations implies existence check and should not require a key must exist or not exist.

What's Changed:

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -294,7 +294,7 @@ codegen-units = 4
 
 [profile.test]
 opt-level = 0
-debug = 2 # enable line numbers by default for easy test debugging
+debug = 1 # enable line numbers by default for easy test debugging
 codegen-units = 16
 lto = false
 incremental = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -294,7 +294,7 @@ codegen-units = 4
 
 [profile.test]
 opt-level = 0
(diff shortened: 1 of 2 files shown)


ekexium (approved):

cfzjywxk (approved):
Will the assertion field still be filled for this type of write requests in tidb or client-go side?

MyonKeminta:
> Will the assertion field still be filled for this type of write requests in tidb or client-go side?

It's possible. For example, if a optimistic transaction inserts a key and then deletes it, there will be a `CheckNotExist` with `Assertion::NotExist`. It won't cause any problem if we check assertion after finishing checking the constraints.

Another case is, if amending lock happened on a pessimistic transaction, the current implementation will cause an assertion check during amending, before the constraint check. I'm not sure if this may happen. I think it's still not possible to cause any false-positive even in this case (as long as there's no unknown bugs), however, for safety, I submitted this PR to skip the check.

PR #11820, opened 2022-01-07
fix make build on Mac M1

gregwebs:
make dev still has issues

I have not tested this on other platforms, but the PR build should test Linux.

### Release note

```release-note
None
```


gregwebs:
This issue is fixed on master now.

PR #11841, opened 2022-01-11
*: check memory limit when inserting in-memory pessimistic locks

sticnarf:

ref #11452

What's Changed:

This PR adds memory tracing for in-memory pessimistic locks and check memory limit before insertion.

To make the memory tracing more precise, the `map` is made private, so the map can be only changed through public functions which maintains the memory usage.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore/src/store/txn_ext.rs
```diff
@@ -45,6 +47,19 @@ impl fmt::Debug for TxnExt {
...
+        "tikv_pessimistic_lock_memory_size",
+        "Total memory size of pessimistic locks in bytes."
+    )
+    .unwrap();
+}
+
+const GLOBAL_MEM_SIZE_LIMIT: usize = 100 << 20; // 100 MiB
+
+// 512 KiB, so pessimistic locks in one region can be proposed in a single command.
+const PEER_MEM_SIZE_LIMIT: usize = 512 << 10;
```
youjiali1995, line 61: Are these too small? Is it necessary to make them adjustable?

youjiali1995 (approved):
BTW, don't forget to adjust `group_locks_by_regions()` after https://github.com/tikv/tikv/pull/11655 is merged.

components/raftstore/src/store/txn_ext.rs
sticnarf: Let's first use a small const, because we don't yet split proposing command before leader transfer.

sticnarf:
> BTW, don't forget to adjust `group_locks_by_regions()` after #11655 is merged.

I move this function to `PeerPessimisticLocks` and add test for it. You can take another look.

youjiali1995 (approved):

PR #11843, opened 2022-01-11
[compile] fix part of compile warnings #8336

### LINKED ISSUE #8336: Fix Clippy warnings

haojinming:
What problem does this PR solve?
Issue Number: close #8336

### Problem Description:
There are several compile warning when build with "cargo clippy".
This PR fix part of them.

### Check List for Tests
This PR has been tested by at least one of the following methods:
Manual test (add detailed scripts or steps below)

### DIFF
--- a/cmd/tikv-ctl/src/util.rs
+++ b/cmd/tikv-ctl/src/util.rs
@@ -11,8 +11,10 @@ use tikv::config::TiKvConfig;
 const LOG_DIR: &str = "./ctl-engine-info-log";
 
 pub fn init_ctl_logger(level: &str) {
-    let mut cfg = TiKvConfig::default();
-    cfg.log_level = slog::Level::from_str(level).unwrap();
+    let mut cfg = TiKvConfig {
+    	log_level: slog::Level::from_str(level).unwrap(),
+	..Default::default()
+    };
     cfg.rocksdb.info_log_dir = LOG_DIR.to_owned();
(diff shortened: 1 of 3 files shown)


pingyu:
@haojinming 
Are all similar cases fixed ? If so, we can enable `field_reassign_with_default` of clippy.

Please fix DCO by signing off. See [here](https://github.com/tikv/tikv/pull/11843/checks?check_run_id=4773424484).

Please update release note in PR description. You can set it to "None".

Rest LGTM.

iosmanthus:
What about the results of `make clippy` @haojinming

haojinming:
@iosmanthus  Actually we disable this kind of warning when executing "make clippy". This issue is complaining the warning whe exexuting "cargo clippy".

haojinming:
@pingyu sorry for missing this for much time. I will check the clippy issue and continue again.

PR #11849, opened 2022-01-12
raftstore: change PeerTicks to enum

### LINKED ISSUE #11848: Change PeerTicks to a simple enum
`PeerTicks` is defined as bitflags:

https://github.com/tikv/tikv/blob/0743109814002b23d539379777a33c56797362fb/components/raftstore/src/store/msg.rs#L183

But `PeerTicks` doesn't need to be bitflags. Only when scheduling ticks need we check if a `PeerTick` exists in a set, but it can be achieved in other ways efficiently too.

It also makes extending the types of `PeerTick` more difficult. It is now defined as `u8` and if we need more tick types, we have to change it to u16. It also wastes space because `tick_batch` needs 2^n space where n is the type of ticks.

Therefore, I think it more appropriate to define `PeerTicks` as an enum. This will make it easier to extend more tick types.

sticnarf:

close #11848

This PR defines `PeerTicks` as an enum instead of bitflags. And the length of `tick_batch` is shorten to the number of tick types.

### Check List 

Tests 

- Existing tests

### Release note 

```release-note
None
```


BusyJay:
schedule tick is a frequent operation when there are a lot of regions.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -97,7 +97,7 @@ where
 {
     pub peer: Peer<EK, ER>,
     /// A registry for all scheduled ticks. This can avoid scheduling ticks twice accidentally.
-    tick_registry: PeerTicks,
+    tick_registry: [bool; PeerTick::get_all_ticks().len()],
```
BusyJay, line 100: How about define it as an associated const value?

components/raftstore/src/store/msg.rs
```diff
@@ -179,45 +178,46 @@ where
...
+    PdHeartbeat = 3,
+    CheckMerge = 4,
+    CheckPeerStaleState = 5,
+    EntryCacheEvict = 6,
+    CheckLeaderLease = 7,
 }
 
-impl PeerTicks {
+impl PeerTick {
+    pub const NUM_TYPES: usize = Self::get_all_ticks().len();
```
BusyJay, line 195: How about naming it `VARIANT_COUNT`?

BusyJay (approved):

components/raftstore/src/store/msg.rs
sticnarf: OK, no opinion.

5kbpers (approved):
LGTM

PR #11853, opened 2022-01-12
raftstore: move scan delete to raft log gc worker

### LINKED ISSUE #10210: Cleaning raft log costs too much time during destroying a peer

BusyJay:

Close #10210.

What's Changed:

When clearing raft metas, raftstore will scan raft logs and delete them
one by one. Seeking can be slow if there are a lot of tombstone keys.
This PR moves the operation to raft log gc worker to reduce the impact.

It offloads the operation by two strategies:
- For destroy, always flush raft gc worker first, so current `last_compacted_idx` is always reliable for cleaning up logs;
- For applying snapshot, use current first index directly, and schedule a raft log gc task after the snapshot is applied. This is a best effort, if TiKV is shutdown before the task is handled, there may still be stale logs. They will be deleted when next compact log task is executed or when the peer is destroyed. This is the same as potential leaking of regular raft gc task.

The final solution should be also moving remaining IO operations to
async write IO threads.

### Check List 

Tests 

- Unit test
- Integration test

### Release note

```release-note
fix potential high latency caused by destroying a peer
```


components/raft_log_engine/src/engine.rs
```diff
@@ -242,6 +242,10 @@ impl RaftEngineReadOnly for RaftLogEngine {
             .fetch_entries_to::<MessageExtTyped>(raft_group_id, begin, end, max_size, to)
             .map_err(transfer_error)
     }
+
+    fn get_all_entries_to(&self, _region_id: u64, _buf: &mut Vec<Entry>) -> Result<()> {
```
BusyJay, line 246: /cc @tabokie PTAL, I'm not sure how to implement it correctly with raft engine.

components/raft_log_engine/src/engine.rs
```diff
@@ -272,6 +276,7 @@ impl RaftEngine for RaftLogEngine {
     fn clean(
         &self,
         raft_group_id: u64,
+        _: u64,
```
BusyJay, line 279: Raft engine should also avoid seeking, but I don't know how to do it either.

tabokie: The underlying implementation is just deleting a memtable and appending a tombstone, no seek involved. So no special care needed.

components/raft_log_engine/src/engine.rs
tabokie: There is no shortcut here. Something like this:
```
if let (Some(first), Some(last)) = (self.0.first_index(id), self.0.last_index(id)) {
    buf.reserve(last - first + 1);
    self.fetch_entries_to(id, first, last + 1, None /*max_size*/, &mut buf)?;
}
```

BusyJay: The thing is `get_all_entries_to` should not check if entries are contiguous. All it does is fetching all available entries in the region.

tabokie: Raft engine doesn't allow non-contiguous entries, so there is not extra check either.

tabokie: Why `first_index` is not reliable?

BusyJay: For example, there are [4, 16] logs in a peer. The peer schedules raft log gc task delete [4, 8), and set first index to 8. While the gc task is pending in queue, the peer is deleted. So logs in range [8, 16] are deleted first, [4, 8) are deleted later.

If TiKV is shutdown when the task is in queue and the peer is not deleted, then after restart, its first index becomes 8. If it applies a snapshot, then it will delete logs in [8, 16] first, and then apply the snapshot. It may append logs further, for example, [20, 24]. And then delete [4, 8). So there is a time both [4, 8) and [20, 24] exist in the engine.

BusyJay: `get_all_entries_to` is used to make sure no matter the order of executions, there will be no garbage left in the end.

tabokie: Oh, this won't happen with raft engine. Raft engine's `compact_to` ignores the `from` index and deletes all entries in `[0, to]`.

tabokie: Feel free to modify. It actually returns wrong result if there are multiple tasks of the same region.

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -17,37 +17,39 @@ use crate::store::worker::metrics::*;
 const MAX_GC_REGION_BATCH: usize = 512;
 const MAX_REGION_NORMAL_GC_LOG_NUBER: u64 = 10240;
```
Connor1996, line 18:
```suggestion
const MAX_REGION_NORMAL_GC_LOG_NUMBER: u64 = 10240;
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2393,31 +2416,113 @@ where
...
+            assert!(self.ctx.sync_write_worker.is_none());
+            // The destroy must be delayed if there are some unpersisted readies.
+            // Otherwise there is a race of writting kv db and raft db between here
+            // and write worker.
+            return Some(DelayReason::UnPersistedReady);
+        }
+
+        if !self.fsm.logs_gc_flushed {
+            let start_index = self.fsm.peer.last_compacted_idx;
+            let mut end_index = start_index;
```
Connor1996, line 2431: if `start_index` == `end_index`, the task is only to make sure all the previous gc tasks of the region are flushed. Can we set the `end_index` with `last_index`, so there is nearly nothing to write in the following destroy process of raftstore.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1210,7 +1212,18 @@ where
...
+            // It's possible that logs between `last_compacted_idx` and `first_index` are
+            // being deleted in raftlog_gc worker. But it's OK as:
+            // 1. If the peer accepts a new snapshot, it must start with an index larger than
+            //    this `first_index`;
+            // 2. If the peer accepts new entries after this snapshot or new snapshot, it must
+            //    start with the new applied index, which is larger than `first_index`.
+            // So new logs won't be deleted by on going raftlog_gc task accidentally.
+            // It's possible that there will be some logs between `last_compacted_idx` and
+            // `first_index` are not deleted. So a cleanup task for the range should be triggered
+            // after applying the snapshot.
```
NingLin-P, line 1225: Wasn't `last_compacted_idx` and `first_index` are the same? namely `truncated_state().get_index() + 1`

components/raftstore/src/store/peer.rs
```diff
@@ -1865,6 +1899,15 @@ where
                     let msgs = self.build_raft_messages(ctx, snap_ctx.msgs);
                     self.send_raft_messages(ctx, msgs);
 
+                    if self.last_compacted_idx == 0
+                        && snap_ctx.last_first_index >= RAFT_INIT_LOG_INDEX
+                    {
+                        // There may be stale logs in raft engine, so schedule a task to clean it
+                        // up. This is a best effort, if TiKV is shutdown before the task is
+                        // handled, there can still be stale logs not being deleted until next
+                        // log gc command is executed. This will delete range [0, last_first_index).
```
NingLin-P, line 1908: How next log gc command can delete stale logs? For example, the apply thread executes the log gc command, update and persisted the `truncated_state` but crashes before the log is actually gc. After restart it will use `truncated_state.get_index() + 1` as the `last_compacted_idx` and `first_index`

components/raftstore/src/store/peer_storage.rs
BusyJay: `last_compacted_idx` is 0 when initialized.

components/raftstore/src/store/peer.rs
BusyJay: No, after being restarted, `last_compacted_idx` is always initialized to 0.

NingLin-P: It doesn't help because the `compact_index` of the log gc command is not calculated from `last_compacted_idx` and the apply thread also rejects log gc commands that `compact_index <= first_index`

components/raftstore/src/store/fsm/peer.rs
NingLin-P: It may be okay if we also let the log gc worker write other metadata (like `PeerState`, `RegionLocalState`) in the same `WriteBacth` that deletes raft log

BusyJay: I don't think it's a good idea to make these essential operations happen in multiple places.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2393,31 +2416,114 @@ where
...
-            // and write worker.
-            if let Some(mbt) = self.fsm.delayed_destroy {
+
+        if let Some(reason) = self.maybe_delay_destroy() {
+            if self
+                .fsm
+                .delayed_destroy
+                .map_or(false, |delay| delay.reason == reason)
+            {
                 panic!(
```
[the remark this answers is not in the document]
BusyJay: It's marked as `panic` because I don't think it's possible.

components/raftstore/src/store/peer_storage.rs
BusyJay: No, it should be scheduled not earlier than the persistent of the deletions.

NingLin-P: The logs between `last_compacted_idx` and `first_index` should be stale logs that won't be accessed anymore and safe to delete whenever we can.

NingLin-P (approved):

5kbpers (approved):

PR #11859, opened 2022-01-12
raftstore: Reject transferring leader to store that having log lag after a restart

Connor1996:

A part of #11409  

What's Changed:
- Collect replicate log lag in `raft_log_gc_tick` and apply log lag in `on_apply_res`
- Reject transfer leader to the store having region with replicate log lag within 10min after the start
- Reject transfer leader to the store having region with apply log lag within 10min after the start

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Reject transferring leader to store that having log lag after a restart
```

### DIFF
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -1500,6 +1500,7 @@ where
         ctx: &mut ApplyContext<EK, W>,
         req: &Request,
     ) -> Result<()> {
+        PEER_WRITE_CMD_COUNTER.put.inc();
         let (key, value) = (req.get_put().get_key(), req.get_put().get_value());
         // region key range has no data prefix, so we must use origin key to check.
         util::check_key_in_region(key, &self.region)?;
@@ -1546,6 +1547,7 @@ where
         ctx: &mut ApplyContext<EK, W>,
         req: &Request,
     ) -> Result<()> {
+        PEER_WRITE_CMD_COUNTER.delete.inc();
         let key = req.get_delete().get_key();
         // region key range has no data prefix, so we must use origin key to check.
         util::check_key_in_region(key, &self.region)?;
@@ -1595,6 +1597,7 @@ where
         ranges: &mut Vec<Range>,
         use_delete_range: bool,
     ) -> Result<()> {
+        PEER_WRITE_CMD_COUNTER.delete_range.inc();
         let s_key = req.get_delete_range().get_start_key();
         let e_key = req.get_delete_range().get_end_key();
         let notify_only = req.get_delete_range().get_notify_only();
@@ -1666,6 +1669,7 @@ where
         req: &Request,
         ssts: &mut Vec<SSTMetaInfo>,
     ) -> Result<()> {
+        PEER_WRITE_CMD_COUNTER.ingest_sst.inc();
         let sst = req.get_ingest_sst().get_sst();
 
         if let Err(e) = check_sst_for_ingestion(sst, &self.region) {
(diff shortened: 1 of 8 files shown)


BusyJay:
> ...within 10min after the start

Why?

Connor1996:
> > ...within 10min after the start
> 
> Why?

In heavy write workload, it may have apply log lag due to high apply duration. In that case, it's unexpected to reject transferring log. So just check it within a time span after restart. 10min is just a time span with which log lag should be possibly caught up.

BusyJay:
> In heavy write workload, it may have apply log lag due to high apply duration.

Yes, but isn't it checked by the current pre-transfer implementation already? Why need a time span?

> In that case, it's unexpected to reject transferring log.

I don't get it, if it has large apply log lag, then reject transferring should be a safe move.

Connor1996:
> Yes, but isn't it checked by the current pre-transfer implementation already? Why need a time span?

What the current implementation does is reject transferring leader if the region has log lag. This PR does if the store has any regions having log lag, reject transferring leader to the store no matter which region.

Connor1996:
> I don't get it, if it has large apply log lag, then reject transferring should be a safe move.

If all the stores are under high apply duration, then the cluster wouldn't do any transfer leader operation. It's unexpected.

Connor1996:
Why it's unexpected? The reason it doesn't transfer leader is to avoid triggering down time for any regions.

That's why only limit this PR's mechanism within 10min after a restart. The current pre-transfer implementation is still kept.

> I don't think it's a good idea. A single peer has log lag may just because it's a hotspot. Other peers can still apply logs in a reasonable speed. We have developed yield mechanism which allows idle peers apply logs even being scheduled with a busy peer.

Even with yield mechanism, the high apply duration still affects the tail latency. See 
截屏2022-01-17 下午5 22 50
截屏2022-01-17 下午5 23 06

BusyJay:
Why not just let PD don't schedule leader to a recently started busy node if possible?

Connor1996:
> Why not just let PD don't schedule leader to a recently started busy node if possible?

But how to define `busy`. Still need TiKV to check the replicate log lag and apply log lag. The replicate log lag can be only checked on the leader side, it's a little weird to report other stores' statuses by leader's store heartbeat report.

BusyJay:
> But how to define `busy`. Still need TiKV to check the replicate log lag and apply log lag. The replicate log lag can be only checked on the leader side, it's a little weird to report other stores' statuses by leader's store heartbeat report.

It can just check the apply duration. Actually log gap is an indirect observation of busy, the CPU usage or the apply duration is a direct observation. On the other hand, scheduling an operator and let it timeout is just a waste. But make PD not scheduling such operator in the first place is a win for all.

Connor1996:
> It can just check the apply duration. Actually log gap is an indirect observation of busy, the CPU usage or the apply duration is a direct observation. On the other hand, scheduling an operator and let it timeout is just a waste. But make PD not scheduling such operator in the first place is a win for all.

I agree not letting PD schedule transfer leader operator is a better choice for apply log lag. But how does it handle replicate log lag. Before logs are caught up, the apply duration is still low. At that time, PD would permit transfer leader, whereas after the log are caught up, the apply duration may high but the leaders are already on that store.

NingLin-P:
> I agree not letting PD schedule transfer leader operator is a better choice for apply log lag. But how does it handle replicate log lag. Before logs are caught up, the apply duration is still low. At that time, PD would permit transfer leader, whereas after the log are caught up, the apply duration may high but the leaders are already on that store.

The pre-transfer mechanism check `last_index >= index + ctx.cfg.leader_transfer_max_log_lag` where `last_index` is the index of leader's last raft log and `index` is the follower's applied index, so both apply/replicate log lag should have been checked.

But with the default config `leader_transfer_max_log_lag: 128` and `raft_entry_max_size: ReadableSize::mb(8)`, the size of the lagging logs may still be quite large due to write hotspot and request batching (which batch multiple requests into one raft log).

Connor1996:
Implemented in https://github.com/tikv/tikv/pull/16239

PR #11862, opened 2022-01-12
server/config: keep compatible using option

### LINKED ISSUE #11861: downgrade from nightly to v5.2.x will fail
nightly

doesn't matter

### Steps to reproduce
1. Deploy a cluster v5.2.x
2. Upgrade to nightly
3. Downgrade to v5.2.x

### What did you expect?
Succeed.

### What did happened?
It fails to start as nightly change the value of `heavy_load_wait_duration` to `50us`, which is only supported after v5.3.0. It should be changed to an `Option` value instead.

BusyJay:

Close #11861.

What's Changed:

Using `Option` for `heavy_load_wait_duration` to keep compatible with
versions prior to v5.3.

### Check List 

Tests 

- Manual test

    Deploy a cluster using this PR, stop TiKV and replace it with v5.2.0, start TiKV.

### Release note 

```release-note
None
```

### DIFF
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -2,6 +2,7 @@
 
 use std::sync::Arc;
 use std::{cmp, i32, isize};
+use std::time::Duration;
 
 use super::Result;
 use grpcio::CompressionAlgorithms;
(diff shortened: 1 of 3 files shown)


NingLin-P (approved):

Connor1996 (approved):
LGTM

PR #11866, opened 2022-01-13
raftstore: Support adjusting max_inflight_msgs dynamically

### LINKED ISSUE #11865: Support adjusting max_inflight_msgs dynamically

A dynamic config modification is performed from TiKV-Control command line (documentation). So expect to adjust the max inflight msgs size in this way:

```
set config tikv `raftstore.raft-max-inflight-msgs`=128;
```

### Technical proposal
When `on_tick` if the peer is leader  then loop call `adjust_max_inflight_msgs` with peer list to implement the modification.

ethercflow:
close #11865

Close #11865  

What's Changed:

### Related changes

### Check List 

Tests 

- Manual test
  use mysql client to test
- No code

Side effects

- None

### Release note 

```release-note
raftstore: Support adjusting max_inflight_msgs dynamically 
```


components/raftstore/src/store/config.rs
```diff
@@ -425,6 +424,14 @@ impl Config {
             ));
         }
 
+        // The adjustment of this value is related to the number of regions, usually 4096 is
+        // already a large enough value
+        if self.raft_max_inflight_msgs == 0 || self.raft_max_inflight_msgs > 4096 {
```
5kbpers, line 429: As I remember, some users are using a value bigger than 4096, maybe we need to enlarge the upper bound.

components/raftstore/src/store/peer.rs
```diff
@@ -4536,6 +4539,18 @@ where
         });
         !has_overlapped_reads && !has_overlapped_writes
     }
+
+    pub fn adjust_raft_max_inflight_msgs_if_cfg_changed<T>(
```
5kbpers, line 4543: Better to name it `adjust_cfg_if_changed`, then we can add more config changes here in the future.

components/raftstore/src/store/peer.rs
```diff
@@ -4536,6 +4539,18 @@ where
...
+        &mut self,
+        ctx: &PollContext<EK, ER, T>,
+    ) {
+        if ctx.cfg.raft_max_inflight_msgs != self.raft_max_inflight_msgs {
+            self.raft_group
+                .raft
+                .adjust_max_inflight_msgs(self.peer_id(), ctx.cfg.raft_max_inflight_msgs);
```
NingLin-P, line 4550: The `raft_max_inflight_msgs` is actually used by the leader for all other peers, so we should call `adjust_max_inflight_msgs` for all peers in the peer list and also should make sure incoming peers will set the same `max_inflight_msgs ` value

ethercflow: Fixed, thanks!

components/raftstore/src/store/peer.rs
ethercflow: Fixed, thanks!

components/raftstore/src/store/config.rs
ethercflow: Fixed, thanks!

NingLin-P (approved):

ethercflow:
Another way is to traverse all regions and update by sending `CasualMessage`, but it has the risk of failure due to full message, so we choose this way.

5kbpers (approved):

PR #11869, opened 2022-01-13
logger: use eprintln! if the logger is not initialized

ethercflow:
ref #11651

Close #11651 

What's Changed:

### Related changes

### Check List 

Tests 

- No code

Side effects

- None

### Release note 

```release-note
None
```


NingLin-P (approved):

5kbpers (approved):

BusyJay:
Is it possible to register a term logger first, and then change it to a configured one? So we can always use `info/debug/error`.

src/config.rs
```diff
@@ -2827,17 +2827,17 @@ impl TiKvConfig {
...
         let default_log_cfg = LogConfig::default();
         if self.log_level != default_tikv_cfg.log_level {
-            println!("deprecated configuration, log-level has been moved to log.level");
+            eprintln!("deprecated configuration, log-level has been moved to log.level");
             if self.log.level == default_log_cfg.level {
-                println!("override log.level with log-level, {:?}", self.log_level);
+                eprintln!("override log.level with log-level, {:?}", self.log_level);
                 self.log.level = self.log_level;
             }
             self.log_level = default_tikv_cfg.log_level;
```
BusyJay, line 2835: I don't get it. If `log_level` is set, then user must have customized the configuration, so it should be respected. The old configuration is more likely to be a stale one. Why overwrite a customized one with a stale one?

ethercflow: > I don't get it. If `log_level` is set, then user must have customized the configuration, so it should be respected. The old configuration is more likely to be a stale one. Why overwrite a customized one with a stale one?

I agree with you, but I think the current code matches your idea.  I don't understand which is the stale one?

BusyJay: I mis-read the line. What's the point of overwriting it with the default value?

BusyJay (approved):

PR #11873, opened 2022-01-17
file_system: overhaul proc-based io tracing utilities

### LINKED ISSUE #11775: Log level for IO snooper failure is too high
/ # ./tikv-server -V
TiKV
Release Version: 5.4.0-alpha
Edition: Community
Git Commit Hash: 99b3436
Git Commit Branch: heads/refs/tags/v5.4.0-nightly
UTC Build Time: 2022-01-04 01:15:55
Rust Version: rustc 1.56.0-nightly (2faabf579 2021-07-27)
Enable Features: jemalloc mem-profiling portable sse test-engines-rocksdb cloud-aws cloud-gcp cloud-azure
Profile: dist_release

8core 16G

### Steps to reproduce
oltp_stability_rel_002
1、
/pd-ctl schedule add shuffle-leader-scheduler:Success!
/pd-ctl schedule add shuffle-region-scheduler:Success!
/pd-ctl schedule add random-merge-scheduler:Success!
2、
go-tpc tpcc run -D tpcc --host tc-tidb.endless-oltp--tps-543619-1-288 -P4000 --warehouses 1000 -T 512 --time 36000m --ignore-error '2013,1213,1105,1205,8022,8027,8028,9004,9007,1062‘
3、
Run chaos {"name": "network-loss", "selectors": ["endless-oltp--tps-543619-1-288/tc-tikv-2"], "experiment": {"Duration":"","Scheduler":null,"Loss":"76","Correlation":"25"}}

### What did you expect?
logs level are right for below log
[2022/01/04 07:05:58.465 +00:00] [ERROR] [server.rs:1071] ["failed to init io snooper"] [err_code=KV:Unknown] [err="\"IO snooper is not started due to not compiling with BCC\""]

### What did happened?
this log should not set to ERRPOR level
[2022/01/04 07:05:58.465 +00:00] [ERROR] [server.rs:1071] ["failed to init io snooper"] [err_code=KV:Unknown] [err="\"IO snooper is not started due to not compiling with BCC\""]

tabokie:

Ref #10867, Fix #11775

What's Changed:

The old implementation is pure wrong. I fixed it and improved the performance by caching the procfs reader. Micro-bench diff:

```
Before:
test tracing::proc::tests::bench_fetch_thread_io_bytes ... bench:       8,463 ns/iter (+/- 884)
test tracing::proc::tests::bench_fetch_thread_io_bytes ... bench:       8,762 ns/iter (+/- 708)
test tracing::proc::tests::bench_fetch_io_bytes        ... bench:      71,775 ns/iter (+/- 11,017)
test tracing::proc::tests::bench_fetch_io_bytes        ... bench:      74,738 ns/iter (+/- 3,355)
test tracing::proc::tests::bench_set_io_type           ... bench:       9,495 ns/iter (+/- 799)
test tracing::proc::tests::bench_set_io_type           ... bench:       9,512 ns/iter (+/- 1,349)

After:
test io_stats::proc::tests::bench_fetch_thread_io_bytes ... bench:       4,715 ns/iter (+/- 494)
test io_stats::proc::tests::bench_fetch_thread_io_bytes ... bench:       4,695 ns/iter (+/- 205)
test io_stats::tests::bench_fetch_io_bytes              ... bench:      39,349 ns/iter (+/- 3,405)
test io_stats::tests::bench_fetch_io_bytes              ... bench:      39,758 ns/iter (+/- 5,069)
test io_stats::tests::bench_set_io_type                 ... bench:       4,896 ns/iter (+/- 290)
test io_stats::tests::bench_set_io_type                 ... bench:       5,049 ns/iter (+/- 309)
```

### Related changes

None.

### Check List 

Tests

- Unit test

### Release note 

```release-note
None
```


Connor1996 (commented):
How is the old implementation wrong?

tabokie:
@Connor1996 The call stack before:
https://github.com/tikv/tikv/blob/da1c069a2f9b5222ddbd32820a5a2ae1de50c78c/components/file_system/src/thread_io.rs#L38-L40
https://github.com/tikv/tikv/blob/da1c069a2f9b5222ddbd32820a5a2ae1de50c78c/components/file_system/src/thread_io.rs#L124-L141
It accesses `THREAD_IO_PRIVATE` whose value is local to the caller thread.

components/file_system/src/tracing/proc.rs
```diff
@@ -0,0 +1,251 @@
...
+    let io_bytes = sentinel.id.fetch_io_bytes();
+    GLOBAL_IO_STATS[sentinel.io_type as usize]
+        .fetch_add(io_bytes - sentinel.last_flushed, Ordering::Relaxed);
+    sentinel.last_flushed = io_bytes;
+    if let Some(new_io_type) = new_io_type {
+        sentinel.io_type = new_io_type;
+    }
+}
+
+pub fn init() -> Result<(), String> {
```
Connor1996, line 141: give a better name `init_io_stats_collector`?

components/server/src/server.rs
```diff
@@ -1066,18 +1066,18 @@ impl<ER: RaftEngine> TiKVServer<ER> {
     }
 
     fn init_io_utility(&mut self) -> BytesFetcher {
-        // Always set to false because BPF needs root permission which is not feasible in a real deployment.
-        // Maybe enable it when the newer stable kernel has better permission control for BPF.
-        let io_snooper_on = false;
+        let io_tracing = file_system::init_tracing()
```
Connor1996, line 1069: `tracing` is a little confusing

tabokie:
Demo screenshot:

image
The surge of reads happened when I called `sudo echo 3 > /proc/sys/vm/drop_caches`.

Will run benchbot to check performance later.

tabokie:
Performance looks okay (up is patched):
image

But I noticed there is a big difference in foreground write I/O statistics:
image

tabokie:
Did some digging, the proc write counter looks good enough. It tracks page dirtying bytes, but does not count writes on an already dirtied page.

Document: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/proc.rst,
Source (4.4.299):
```
__set_page_dirty_buffers
 + __set_page_dirty if (newly_dirty)
  + account_page_dirtied
   + task_io_account_write
    + current->ioac.write_bytes += bytes;
```
Ran a test on IDC machine, "IO Breakdown" matches "Disk Performance" better than before:

image
image

Connor1996 (commented):
Seems we still need to fetch io byte diff on every rate limiter request?

Connor1996:
> Performance looks okay (up is patched): image
> 
> But I noticed there is a big difference in foreground write I/O statistics: image

why there is a big difference?

tabokie:
> why there is a big difference?

Possibly caused by amplification between user write size and OS paged write size. I checked the average Bytes/Written is around 500B, whereas page size is normaly 4KB.

metrics/grafana/tikv_details.json
```diff
@@ -8578,15 +8578,15 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "max_over_time((sum(rate(tikv_io_bytes{tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", op=\"write\"}[30s])) by (type))[30s:10s])",
+              "expr": "sum(rate(tikv_io_bytes{tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", op=\"write\"}[45s])) by (type)",
```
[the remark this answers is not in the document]
tabokie: Before using proc, stats from rate limiter is always smaller than "Disk Performance", so I used `max_over_time` to compensate. Now, the counter itself is good enough, and 45s is just to make the graph looks smoother.

components/file_system/src/io_stats/proc.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+#[derive(Default)]
+struct AtomicIOBytes {
+    read: AtomicU64,
+    write: AtomicU64,
+}
+
+impl AtomicIOBytes {
+    fn load(&self, order: Ordering) -> IOBytes {
```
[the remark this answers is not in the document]
tabokie: This is copied from old code, It's a wrapper type so I think accepting order is fine.

metrics/grafana/tikv_details.json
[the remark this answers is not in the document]
Connor1996: but larger time span conceal the flow spike

Connor1996 (commented):
rest LGTM

components/file_system/src/io_stats/proc.rs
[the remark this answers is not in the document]
Connor1996: but seems no reason to use other ordering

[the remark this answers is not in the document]
Connor1996: @tabokie

metrics/grafana/tikv_details.json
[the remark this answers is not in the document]
tabokie: I was doing it to avoid "zigzagging" pattern. But I checked again and it's not an issue anymore. So I'll tune it down now.

components/file_system/src/io_stats/proc.rs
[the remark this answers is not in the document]
tabokie: I'm still not very convinced. Hardcoding the order can save us some typing, but exposing a std-like `load` interface can also make the code familiar and self-explanatory.

Connor1996 (approved):
LGTM

components/file_system/src/file.rs
```diff
@@ -231,6 +234,19 @@ impl Default for OpenOptions {
     }
 }
 
+#[cfg(target_os = "linux")]
+impl OpenOptionsExt for OpenOptions {
```
BusyJay, line 238: Why reinvent the wheel?

tabokie: `file_system` is supposed to replace `std::fs`, therefore traits that work for `std::fs::OpenOptions` should also work for `file_system::OpenOptions`.

BusyJay: What I mean is why replace `std::fs`? Why not partially replaced? Or provide patch functions?

tabokie: Partial replacement won't help this case, `OpenOptions` is tightly coupled with `File` type. Patch function like `fn wrap(std::File) -> file_system::File` is inconvenient compared to replacing. Such [trivial inconvenience](https://www.lesswrong.com/posts/reitXJgJXFzKpdKyd/beware-trivial-inconveniences) is enough to discourage other contributors to use our customized file.

tabokie:
@Connor1996 Changing the metrics interval back to 45s. 30s still has that zigzagging effect:

image

components/file_system/src/io_stats/biosnoop.rs
```diff
@@ -146,22 +147,26 @@ pub fn get_io_type() -> IOType {
...
+    let mut bytes = Default::default();
     unsafe {
         if let Some(ctx) = BPF_CONTEXT.as_mut() {
-            let io_type_buf_ptr = &mut io_type as *mut IOType as *mut u8;
-            let mut io_type_buf =
-                std::slice::from_raw_parts_mut(io_type_buf_ptr, std::mem::size_of::<IOType>());
-            if let Ok(e) = ctx.stats_table.get(&mut io_type_buf) {
-                assert!(e.len() == std::mem::size_of::<IOBytes>());
-                return std::ptr::read_unaligned(e.as_ptr() as *const IOBytes);
+            for io_type in IOType::iter() {
```
hunterlxt, line 154: Why return an array containing all Type? Not all types are required where I saw below. like `let mut compaction_bytes_before = fetch_io_bytes()[IOType::Compaction as usize];`

components/file_system/src/io_stats/biosnoop.rs
```diff
@@ -146,22 +147,26 @@ pub fn get_io_type() -> IOType {
...
+                }
             }
         }
     }
-    IOBytes::default()
+    bytes
 }
 
-pub fn init_io_snooper() -> Result<(), String> {
+pub fn init() -> Result<(), String> {
```
hunterlxt, line 169: consider a more clear name?

components/file_system/src/io_stats/mod.rs
```diff
@@ -0,0 +1,88 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+#[cfg(not(any(target_os = "linux", feature = "bcc-iosnoop")))]
```
hunterlxt, line 3: In my understanding, this is to replace `null.rs`?

components/file_system/src/io_stats/biosnoop.rs
```diff
@@ -244,7 +249,8 @@ macro_rules! flush_io_latency {
     };
 }
 
-pub(crate) fn flush_io_latency_metrics() {
+#[allow(dead_code)]
+pub fn flush_io_latency_metrics() {
```
hunterlxt, line 253: I saw it's not used in `components/file_system/src/metrics_manager.rs` why not removed here

components/file_system/src/io_stats/proc.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    fn fetch_add(&self, other: IOBytes, order: Ordering) {
+        self.read.fetch_add(other.read, order);
+        self.write.fetch_add(other.write, order);
+    }
+}
+
+/// Flushes the local I/O stats to global I/O stats.
+#[inline]
+fn flush_thread_io(sentinel: &mut LocalIOStats) {
```
hunterlxt, line 140: How often is this happend. Seem that this function involves IO read

components/file_system/src/io_stats/biosnoop.rs
tabokie: `init` belongs to `io_stats` crate. So I think it wouldn't be confusing.

components/file_system/src/io_stats/mod.rs
tabokie: Yes. It's a stub of interfaces that each io_stats implementation must adhere to.

components/file_system/src/io_stats/biosnoop.rs
tabokie: All codes related to biosnoop is deprecated and useless now. But a lot of efforts went into making it work, so I prefer keeping it for now.

components/file_system/src/io_stats/proc.rs
tabokie: It involves syscall, but no IO. As procfs is a virtual file that is redirected to kernel memory. Plus the metrics flusher only calls it once every 10s. For the overhead, you can check out the benchmark results in https://github.com/tikv/tikv/pull/11873#issuecomment-1017080413.

components/file_system/src/io_stats/biosnoop.rs
hunterlxt: just a suggestion, usually it will be used `init()` not `biosnoop::init()`.

components/file_system/src/io_stats/proc.rs
hunterlxt: got

components/file_system/src/io_stats/biosnoop.rs
hunterlxt: Do you mean only `proc::fetch_io_bytes` work now and `biosnoop.rs` deprecated?

tabokie: Yes.

components/file_system/src/io_stats/biosnoop.rs
tabokie: biosnoop or proc or stub is all private implementation details. We only expose `io_stats` crate.

hunterlxt (approved):
LGTM

PR #11878, opened 2022-01-19
raftstore: move scan delete to raft log gc worker (#11853)

### LINKED ISSUE #10210: Cleaning raft log costs too much time during destroying a peer

ti-srebot:
cherry-pick #11853 to release-5.1
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/11878
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/11878:release-5.1-6170255604af
```

---

Close #10210.

What's Changed:

When clearing raft metas, raftstore will scan raft logs and delete them
one by one. Seeking can be slow if there are a lot of tombstone keys.
This PR moves the operation to raft log gc worker to reduce the impact.

It offloads the operation by two strategies:
- For destroy, always flush raft gc worker first, so current `last_compacted_idx` is always reliable for cleaning up logs;
- For applying snapshot, use current first index directly, and schedule a raft log gc task after the snapshot is applied. This is a best effort, if TiKV is shutdown before the task is handled, there may still be stale logs. They will be deleted when next compact log task is executed or when the peer is destroyed. This is the same as potential leaking of regular raft gc task.

The final solution should be also moving remaining IO operations to
async write IO threads.

### Check List 

Tests 

- Unit test
- Integration test

### Release note

```release-note
fix potential high latency caused by destroying a peer
```


sticnarf:
PTAL @5kbpers @NingLin-P

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -131,6 +143,12 @@ where
     max_inflight_msgs: usize,
 
     trace: PeerMemoryTrace,
+    /// Destroy is delayed because of some unpersisted readies in Peer.
+    /// Should call `destroy_peer` again after persisting all readies.
```
NingLin-P, line 147: Plz update the comment

NingLin-P (approved):

5kbpers (approved):

PR #11883, opened 2022-01-19
raftstore: reactivate in-memory pessimistic locking when leader transfer fails

sticnarf:

Refer to #11452

What's Changed:

This PR adds a new tick `ReactivateMemoryLock`. If leader transfer does not succeed after some ticks, we set `is_valid` back to true.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4652,6 +4658,27 @@ where
...
+        // be enabled when this region becomes leader again.
+        if !self.fsm.peer.is_leader() {
+            return;
+        }
+
+        self.fsm.reactivate_memory_lock_ticks += 1;
+        if self.fsm.reactivate_memory_lock_ticks >= self.ctx.cfg.reactive_memory_lock_timeout_tick {
+            self.fsm.peer.txn_ext.pessimistic_locks.write().is_valid = true;
+        } else if !self.fsm.peer.txn_ext.pessimistic_locks.read().is_valid {
+            // Continue ticking only if `is_valid` is false. If it becomes true, probably this
```
BusyJay, line 4677: How do you know it fails? It can still succeed.

sticnarf: The comment describes right the case when transfer leader succeeds.

If transfer leader succeeds, `is_valid` will be `true`. Then, we don't need to register tick again.

NingLin-P: Consider the following case:
1. PD send command to transfer leader from A to B
2. A register the `PeerTick::ReactivateMemoryLock` and start transfer leader 
3. For some reason the transfer is aborted
4. PD retry the transfer command and A start transfer leader again (set `is_valid` to `false`)
5. The `PeerTick::ReactivateMemoryLock` triggered and set `is_valid` to `true` hence A accept more memory locks
6. Transfer succeeded and B become the new leader but lost some memory locks

sticnarf: In step 4, when `is_valid` is set to `false`, `reactivate_memory_lock_ticks` will be reset to 0. Then, in step 5 when `PeerTick::ReactivateMemoryLock` is triggered, the `self.fsm.reactivate_memory_lock_ticks >= self.ctx.cfg.reactive_memory_lock_timeout_tick` condition will not be satistied, so `is_valid` will not be set back to `true`.

NingLin-P: I see, but it may still be possible that the transfer takes more time than the `PeerTick::ReactivateMemoryLock` reset `is_valid`, just like the abovementioned case but without step 3 and step 4.

sticnarf: I got what you mean. Using tick is a best-effort solution. We don't 100% guarantee that no pessimistic locks will be lost.

But to be better, perhaps we can check `self.raft_group.raft.lead_transferee` in the tick. If it's `Some`, we don't set it to `true` and continue ticking.

NingLin-P: > I got what you mean. Using tick is a best-effort solution. We don't 100% guarantee that no pessimistic locks will be lost.

Yes, the current implementation is okay to me

> But to be better, perhaps we can check self.raft_group.raft.lead_transferee in the tick. If it's Some, we don't set it to true and continue ticking.

Seems a good idea

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4804,6 +4822,36 @@ where
...
+
+        self.fsm.reactivate_memory_lock_ticks += 1;
+        let transferring_leader = self.fsm.peer.raft_group.raft.lead_transferee.is_some();
+        // `lead_transferee` is not set immediately after the lock status changes. So, we need
+        // the tick count condition to avoid reactivating too early.
+        if !transferring_leader
+            && self.fsm.reactivate_memory_lock_ticks
+                >= self.ctx.cfg.reactive_memory_lock_timeout_tick
+        {
+            pessimistic_locks.status = LocksStatus::Normal;
```
NingLin-P, line 4849: How about also reset `reactivate_memory_lock_ticks` here for robustness.

NingLin-P (commented):
Rest LGTM

youjiali1995 (approved):

youjiali1995:
@NingLin-P PTAL

NingLin-P (approved):

PR #11894, opened 2022-01-21
copr: support Bit column push down

xiejiandong:
## TiKV coprocessor: support Bit column push down
Issue: https://github.com/tikv/tikv/issues/11893
TiDB old version not support Bit column push down;  
One reason is TiKV not support,  another reason is TiDB set it unsupport. 

### This PR have Some files and functions update: 
1. components/tidb_query_aggr/src/util.rs
    rewrite_exp_for_sum_avg() 
2. components/tidb_query_datatype/src/def/eval_type.rs  
    TryFrom()
3. components/tidb_query_datatype/src/codec/chunk/column.rs
    from_raw_datums()
    from_vector_value()    
    +append_bit_datum()
4、components/tidb_query_expr/src/impl_cast.rs
     cast_bit_as_string()

TiDB's corresponding PR: pingcap/tidb#31884

### Tests
- Unit test
- [Integration test](https://github.com/tikv/copr-test)
- Manual test：[test reposrt](https://docs.google.com/document/d/1jPOzbBZlxkdusJPQfh38IW7KfDv08YERKqpMAfG7oEM/edit?usp=sharing)

### Side effects
- No side effects.
- Be backward compatibility
- Need TiDB upgrade together to use the feature.

### Release note
Can realease.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,12 +115,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dabe5a181f83789739c194cbe5a897dde195078fac08568d09221fd6137a7ba8"
 
 [[package]]
-name = "arc-swap"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5d78ce20460b82d3fa150275ed9d55e21064fc7951177baacf86a145c4a4b1f"
-
-[[package]]
 name = "arrayvec"
 version = "0.4.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4083,7 +4077,6 @@ dependencies = [
 name = "resource_metering"
 version = "0.0.1"
 dependencies = [
- "arc-swap 1.5.0",
  "collections",
  "crossbeam",
  "futures 0.3.15",
@@ -4696,7 +4689,7 @@ name = "slog-global"
 version = "0.1.0"
 source = "git+https://github.com/breeswish/slog-global.git?rev=d592f88e4dbba5eb439998463054f1a44fbf17b9#d592f88e4dbba5eb439998463054f1a44fbf17b9"
 dependencies = [
- "arc-swap 0.4.8",
+ "arc-swap",
  "lazy_static",
  "log",
  "slog",
--- a/components/resource_metering/Cargo.toml
+++ b/components/resource_metering/Cargo.toml
@@ -21,7 +21,6 @@ slog = { version = "2.3", features = ["max_level_trace", "release_max_level_debu
 slog-global = { version = "0.1", git = "https://github.com/breeswish/slog-global.git", rev = "d592f88e4dbba5eb439998463054f1a44fbf17b9" }
 futures = "0.3"
 pdqselect = "0.1"
-arc-swap = "1.5.0"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 procinfo = { git = "https://github.com/tikv/procinfo-rs", rev = "5125fc1a69496b73b26b3c08b6e8afc3c665a56e" }
--- a/components/resource_metering/src/lib.rs
+++ b/components/resource_metering/src/lib.rs
@@ -79,14 +79,18 @@ impl ResourceMeteringTag {
                 }
             }
 
-            assert!(!ls.is_set, "nested attachment is not allowed");
-            ls.attached_tag.store(Some(self.infos.clone()));
+            // unexpected nested attachment
+            if ls.is_set {
+                debug_assert!(false, "nested attachment is not allowed");
+                return Guard;
+            }
+
+            let prev_tag = ls.attached_tag.swap(Some(self.infos.clone()));
+            debug_assert!(prev_tag.is_none());
             ls.is_set = true;
             ls.summary_cur_record.reset();
 
-            Guard {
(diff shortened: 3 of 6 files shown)


components/tidb_query_aggr/src/util.rs
```diff
@@ -39,10 +39,20 @@ pub fn rewrite_exp_for_sum_avg(schema: &[FieldType], exp: &mut RpnExpression) ->
...
             return Ok(());
         }
-        EvalType::Int => FieldTypeBuilder::new()
-            .tp(FieldTypeTp::NewDecimal)
-            .flen(tidb_query_datatype::MAX_DECIMAL_WIDTH)
-            .build(),
+        EvalType::Int => {
+            if ret_field_type.tp() == FieldTypeTp::Bit {
+                FieldTypeBuilder::new()
+                    .tp(FieldTypeTp::Double)
```
skyzh, line 45: Why is `Bit`'s field type `Double` in this part? Is this specified in TiDB?

components/tidb_query_datatype/src/codec/chunk/column.rs
```diff
@@ -73,7 +73,15 @@ impl Column {
         let eval_type = box_try!(EvalType::try_from(field_type.as_accessor().tp()));
         match eval_type {
             EvalType::Int => {
-                if field_type.is_unsigned() {
+                if field_type.as_accessor().tp() == FieldTypeTp::Bit {
+                    if field_type.as_accessor().flen() > 64 {
+                        unimplemented!()
+                    }
+                    let start_idx: usize = ((64 - field_type.as_accessor().flen()) / 8) as usize;
```
skyzh, line 80: Is there any document on `Bit` vector (and scalar)'s encoding? These operations that doesn't look straightforward to me, and I guess it's a kind of hack to make this work.

skyzh (commented):
Rest LGTM.

What's TiDB's corresponding PR to this TiKV PR?

And the same time, would you please make the test report accessible by URL, so that community members can look into that?

components/tidb_query_datatype/src/codec/chunk/column.rs
xiejiandong: Bit is defined bits range from 1 to 64:  https://dev.mysql.com/doc/refman/8.0/en/bit-type.html
For higher version of mysql, Bit will be more bytes, but TiDB has't support yet，and TiDB defined Bit storage len be 8 bytes（tidb/parser/types/field_type.go +352）now，hard code； So，for future TiDB that Bit support more bytes，we should fix TiKV codes here；
Maybe I should has some comments here.

components/tidb_query_aggr/src/util.rs
xiejiandong: Codes tidb\expression\aggregation\base_func.go  define typeInfer4Sum()、typeInfer4Avg(), For Bit RetType is Double.
TiDB's corresponding PR:  https://github.com/pingcap/tidb/pull/31884

skyzh (approved):
LGTM from my side -- you'll need another reviewer for this PR.

At the same time, you should add this case to integration-copr-test. This check is expected to run for every PR, but it seems that the EE team has removed it. Anyway, you can merge the PRs in TiDB and TiKV first, and then add push-down integration test (instead of doing manual test every time).

See https://github.com/tikv/copr-test for copr-test examples and information.

Good work!

xiejiandong:
status/can-merge

components/tidb_query_datatype/src/def/eval_type.rs
```diff
@@ -75,7 +76,7 @@ impl std::convert::TryFrom<crate::FieldTypeTp> for EvalType {
             | crate::FieldTypeTp::Null => EvalType::Bytes,
             crate::FieldTypeTp::Enum => EvalType::Enum,
             _ => {
-                // In TiDB, Bit's eval type is Int, but it is not yet supported in TiKV.
+                // In TiDB, Bit's eval type is Int
```
zhongzc, line 79:
```suggestion
```

components/tidb_query_datatype/src/codec/chunk/column.rs
zhongzc: Comments are needed for this line of code.

components/tidb_query_datatype/src/codec/chunk/column.rs
```diff
@@ -523,6 +552,27 @@ impl Column {
...
+    pub fn append_bit_datum(&mut self, src_datum: &[u8], start_idx: &usize) -> Result<()> {
+        if src_datum.is_empty() {
+            return Err(Error::InvalidDataType(
+                "Failed to decode datum flag".to_owned(),
+            ));
+        }
+        let flag = src_datum[0];
+        match flag {
+            datum::NIL_FLAG => self.append_null(),
+            datum::UINT_FLAG => self.append_bytes(&src_datum[(1 + start_idx)..])?,
```
zhongzc, line 565: How `1 +` works? Need comments I think.

xiejiandong: It skip [0] flag; ok;

components/tidb_query_datatype/src/codec/chunk/column.rs
xiejiandong: ok, has added it.

PR #11897, opened 2022-01-21
server: tikv should panic when read request met engine corruption error

### LINKED ISSUE #11617: TiKV should panic when read request met engine corruption error

When data files are corrupted, TiKV won't panic until rocksdb's background job (like compaction) met such corrupted files.  But before that, TiKV will keep response errors like:
```
Engine("Corruption: truncated block read from xxx.sst ...")
```
for read request that hit the corrupted file, and since the leadership is healthy TiDB will keep backoff retry on this TiKV until timeout.

TiKV should panic when reading such corrupted files, so that the peer on other TiKV can be elected as the new leader (if there are still majority peers alive) and serve the read request 

TiKV doesn't panic on this case so that the TiKV can serve read request of other uncorrupted files until the background job trigger panic. But since there are corrupted files, panic is inevitable.

NingLin-P:

Close #11617 

What's Changed:

Panic when read request met engine corruption error

### Check List 

Tests 

    - Run a read-only workload
    - Corrupt sst files (e.g replace the content with some random bytes)
    - TiKV store, those sst files were corrupted, should panic as read request hit these corrupted sst files
    - Peers on other stores should be elected as the new leader
    - TiDB see a latency spike then become normal as the new leader elected

Side effects

### Release note 

```release-note
server: tikv should panic when read request met engine corruption error
```

### DIFF
--- a/src/coprocessor/endpoint.rs
+++ b/src/coprocessor/endpoint.rs
@@ -602,11 +602,22 @@ impl<E: Engine> Endpoint<E> {
     }
 }
 
+fn is_engine_corruption_error(e: &Error) -> bool {
+    // Hack! it depend on the error message to identify the engine corruption error
+    let err_msg = format!("{:?}", e).to_lowercase();
+    err_msg.contains("engine") && err_msg.contains("corruption") && err_msg.contains("sst")
+}
+
 fn make_error_response(e: Error) -> coppb::Response {
     warn!(
         "error-response";
         "err" => %e
     );
+
+    if is_engine_corruption_error(&e) {
+        panic!("met engine corruption error: {:?}", e);
+    }
+
     let mut resp = coppb::Response::default();
     let tag;
     match e {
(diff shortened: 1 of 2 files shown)


sticnarf:
Is it possible to create an error variant like `Corruption` to `engine_trait::Error` and generate it in the engine impl?

src/coprocessor/endpoint.rs
```diff
@@ -602,11 +602,23 @@ impl<E: Engine> Endpoint<E> {
     }
 }
 
+fn is_engine_corruption_error(e: &Error) -> bool {
+    // Hack! it depend on the error message to identify the engine corruption error
+    let err_msg = e.to_string().to_lowercase();
```
hunterlxt, line 607: what's the purpose

hunterlxt (approved):
LGTM

NingLin-P:
> Is it possible to create an error variant like `Corruption` to `engine_trait::Error`?

Rocksdb return error in `String` so string matching is inevitable. Though we can do that in the engine interfaces, I'm not sure if there are common non-corrupt errors that can be handled inside tikv, and the string matching introduces cost for them.

> Or, what about panicking directly in the engine impl?

Ideally, rocksdb should panic directly. Doing that in the engine interfaces may face the abovementioned problem.

src/coprocessor/endpoint.rs
NingLin-P: The `engine_traits::Error` is set to the `Error::message` field.

PR #11900, opened 2022-01-24
raftstore: Introduce raft log fetcher

Connor1996:

Part of https://github.com/tikv/tikv/pull/11409 

What's Changed:

- Introduce raft log fetcher(but not used, will be used in next PR)
- Update raft and some related API changes

### Related changes

- None

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,7 +112,7 @@ pin-project = "1.0"
 pnet_datalink = "0.23"
 pprof = { version = "^0.6", default-features = false, features = ["flamegraph", "protobuf"] }
 protobuf = { version = "2.8", features = ["bytes"] }
-raft = { version = "0.6.0-alpha", default-features = false, features = ["protobuf-codec"] }
+raft = { version = "0.7.0", default-features = false, features = ["protobuf-codec"] }
 raftstore = { path = "components/raftstore", default-features = false }
 raft_log_engine = { path = "components/raft_log_engine", default-features = false }
 rand = "0.7.3"
@@ -195,8 +195,8 @@ procinfo = { git = "https://github.com/tikv/procinfo-rs", rev = "5125fc1a69496b7
 # When you modify TiKV cooperatively with kvproto, this will be useful to submit the PR to TiKV and the PR to
 # kvproto at the same time.
 # After the PR to kvproto is merged, remember to comment this out and run `cargo update -p kvproto`.
-# [patch.'https://github.com/pingcap/kvproto']
-# kvproto = {git = "https://github.com/your_github_id/kvproto", branch="your_branch"}
+[patch.'https://github.com/pingcap/kvproto']
+kvproto = {git = "https://github.com/Connor1996/kvproto", branch="async-fetch-entry"}
 
 [workspace]
 # See https://github.com/rust-lang/rfcs/blob/master/text/2957-cargo-features2.md
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,8 +195,8 @@ procinfo = { git = "https://github.com/tikv/procinfo-rs", rev = "5125fc1a69496b7
 # When you modify TiKV cooperatively with kvproto, this will be useful to submit the PR to TiKV and the PR to
 # kvproto at the same time.
 # After the PR to kvproto is merged, remember to comment this out and run `cargo update -p kvproto`.
-[patch.'https://github.com/pingcap/kvproto']
-kvproto = {git = "https://github.com/Connor1996/kvproto", branch="async-fetch-entry"}
+# [patch.'https://github.com/pingcap/kvproto']
+# kvproto = {git = "https://github.com/your_github_id/kvproto", branch="your_branch"}
 
 [workspace]
 # See https://github.com/rust-lang/rfcs/blob/master/text/2957-cargo-features2.md
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -1504,6 +1504,7 @@ where
         ctx: &mut ApplyContext<EK, W>,
         req: &Request,
     ) -> Result<()> {
+        PEER_WRITE_CMD_COUNTER.put.inc();
         let (key, value) = (req.get_put().get_key(), req.get_put().get_value());
         // region key range has no data prefix, so we must use origin key to check.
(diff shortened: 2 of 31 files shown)


Cargo.toml
```diff
@@ -195,8 +195,8 @@ procinfo = { git = "https://github.com/tikv/procinfo-rs", rev = "5125fc1a69496b7
 # When you modify TiKV cooperatively with kvproto, this will be useful to submit the PR to TiKV and the PR to
 # kvproto at the same time.
 # After the PR to kvproto is merged, remember to comment this out and run `cargo update -p kvproto`.
-# [patch.'https://github.com/pingcap/kvproto']
-# kvproto = {git = "https://github.com/your_github_id/kvproto", branch="your_branch"}
+[patch.'https://github.com/pingcap/kvproto']
```
BusyJay, line 198: Missing patch for raft.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1504,6 +1504,7 @@ where
         ctx: &mut ApplyContext<EK, W>,
         req: &Request,
     ) -> Result<()> {
+        PEER_WRITE_CMD_COUNTER.put.inc();
```
BusyJay, line 1507: How about using local metrics?

components/raftstore/src/store/msg.rs
```diff
@@ -303,6 +303,29 @@ where
...
+    // Reports the result of asynchronous Raft logs fetching.
+    RaftlogFetched {
+        context: GetEntriesContext,
+        res: RaftlogFetchResult,
+    },
+}
+
+#[derive(Debug)]
+pub enum RaftlogFetchResult {
+    Fetching,
```
BusyJay, line 315: When will it be used?

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -0,0 +1,115 @@
...
+            Task::PeerStorage {
+                region_id,
+                low,
+                high,
+                max_size,
+                context,
+                tried_cnt,
+                term,
+            } => {
+                let mut ents = vec![];
```
BusyJay, line 83: How about specifying capacity?

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -0,0 +1,115 @@
...
+                let res = self.raft_engine.fetch_entries_to(
+                    region_id,
+                    low,
+                    high,
+                    Some(max_size),
+                    &mut ents,
+                );
+
+                let hit_size_limit = res
```
BusyJay, line 92: If it's an error, then no need to return the entries as next try has to fail.

components/raftstore/src/store/fsm/apply.rs
Connor1996: `auto_flush_from!` already make it be local metrics.

components/raftstore/src/store/msg.rs
Connor1996: in next PR, when `async_fetch` is called the fisrt time. To make sure only one async fetch in flight per region.

components/raftstore/src/store/worker/raftlog_fetch.rs
Connor1996: the error would be `ErrCompacted`, so need to return and trigger raft callback.

Cargo.toml
Connor1996: on line 176, `raft = { git = "https://github.com/tikv/raft-rs", branch = "master" }`

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -0,0 +1,115 @@
...
+                    .map(|c| (*c as u64) != high - low)
+                    .unwrap_or(false);
+
+                // it may return a region not found error as the region could be merged.
+                let _ = self.router.significant_send(
+                    region_id,
+                    SignificantMsg::RaftlogFetched {
+                        context,
+                        res: RaftlogFetchResult::Fetched {
+                            ents: res.map(|_| ents).map_err(|e| e.into()),
```
BusyJay, line 103: Perhaps `ents` should be returned as well. In async raft, raft may request logs between [low, high) and only [low, cache_low) are persisted and able to be fetched. So unavailable may always be returned.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1361,6 +1370,12 @@ impl<EK: KvEngine, ER: RaftEngine> RaftBatchSystem<EK, ER> {
                 };
             },
         );
+
+        let raftlog_fetch_scheduler = workers.raftlog_fetch_worker.start(
```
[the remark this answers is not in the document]
Connor1996: in `RaftBatchSystem::shutdown`, workers are drained and be stopped one by one.

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    fn run(&mut self, task: Task) {
+        match task {
+            Task::PeerStorage {
+                region_id,
+                low,
+                high,
+                max_size,
+                context,
+                tried_cnt,
```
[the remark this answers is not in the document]
Connor1996: it will be used in next PR, tried_cnt will be added in `Storage::entries()`

components/raftstore/src/store/worker/raftlog_fetch.rs
Connor1996: We can't rely on the underlying engine's implementation to ensure `ents` are partial fetched when error is returned

components/raftstore/src/store/fsm/store.rs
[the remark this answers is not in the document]
NingLin-P: Sorry, I don't get it, workers should stop in:
https://github.com/tikv/tikv/blob/d1cdf642f6034fd91a5cb310f7602d98b67c3dff/components/raftstore/src/store/fsm/store.rs#L1554-L1559

BusyJay (approved):

NingLin-P (approved):

PR #11904, opened 2022-01-24
gc_worker: fix incorrect scheduled_tasks counting

### LINKED ISSUE #11903: False GcWorkerTooBusy caused by incorrect scheduled_tasks
## Bug Report

If TiKV is upgraded from a version with https://github.com/tikv/tikv/issues/11217 bug, there can be a lot of accumulated MVCC versions. Then, the GC worker can be very busy for quite some time.

During the period, the gc worker may increase its `scheduled_tasks` count but fail to schedule the task:

 https://github.com/tikv/tikv/blob/e8718be241/src/server/gc_worker/gc_worker.rs#L937

When it fails to schedule the task, it forgets to decrease the count. So, `scheduled_tasks` is incorrectly maintained. It can increase beyond `GC_MAX_EXECUTING_TASKS` and never decrease. After that, all later `unsafe_destroy_range` requests will be rejected due to false `GcWorkerTooBusy` errors.

This bug is caused by https://github.com/tikv/tikv/pull/11429.

sticnarf:

Closes #11903

What's Changed:

1. Callback is always called in `handle_gc_task_schedule_error`, so the `scheduled_tasks` counter is maintained correctly.
2. Allow `unsafe_destroy_range` to schedule even if the worker is full. Then, `unsafe_destroy_range` will still have the chance to execute when the GC worker is busy.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Fixes the bug that unsafe_destroy_range does not get executed when GC worker is busy
```

### DIFF
--- a/src/server/gc_worker/gc_worker.rs
+++ b/src/server/gc_worker/gc_worker.rs
@@ -668,7 +668,17 @@ where
 /// When we failed to schedule a `GcTask` to `GcRunner`, use this to handle the `ScheduleError`.
 fn handle_gc_task_schedule_error(e: ScheduleError<GcTask<impl KvEngine>>) -> Result<()> {
     error!("failed to schedule gc task"; "err" => %e);
-    Err(box_err!("failed to schedule gc task: {:?}", e))
+    let res = Err(box_err!("failed to schedule gc task: {:?}", e));
+    match e.into_inner() {
+        GcTask::Gc { callback, .. } | GcTask::UnsafeDestroyRange { callback, .. } => {
+            callback(Err(Error::from(ErrorInner::GcWorkerTooBusy)))
+        }
+        GcTask::PhysicalScanLock { callback, .. } => {
+            callback(Err(Error::from(ErrorInner::GcWorkerTooBusy)))
+        }
+        _ => {}
+    }
+    res
 }
 
 /// Schedules a `GcTask` to the `GcRunner`.
@@ -935,8 +945,13 @@ where
     ) -> Result<()> {
         GC_COMMAND_COUNTER_VEC_STATIC.unsafe_destroy_range.inc();
         self.check_is_busy(callback).map_or(Ok(()), |callback| {
+            // Use schedule_force to allow unsafe_destroy_range to schedule even if
+            // the GC worker is full. This will help free up space in the case when
+            // the GC worker is busy with other tasks.
+            // Unsafe destroy range is in store level, so the number of them is
+            // quite small, so we don't need to worry about its memory usage.
             self.worker_scheduler
-                .schedule(GcTask::UnsafeDestroyRange {
+                .schedule_force(GcTask::UnsafeDestroyRange {
                     ctx,
                     start_key,
                     end_key,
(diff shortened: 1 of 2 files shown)


cfzjywxk (approved):

youjiali1995 (approved):

src/server/gc_worker/gc_worker.rs
```diff
@@ -935,8 +945,13 @@ where
     ) -> Result<()> {
         GC_COMMAND_COUNTER_VEC_STATIC.unsafe_destroy_range.inc();
         self.check_is_busy(callback).map_or(Ok(()), |callback| {
```
Connor1996, line 947: can we remove `check is busy` and use scheduler capacity instead? seems `GcTask::GcKeys` can bypass it. Mix-use may still cause potential problems in the future.

src/server/gc_worker/gc_worker.rs
```diff
@@ -668,7 +668,17 @@ where
...
     error!("failed to schedule gc task"; "err" => %e);
-    Err(box_err!("failed to schedule gc task: {:?}", e))
+    let res = Err(box_err!("failed to schedule gc task: {:?}", e));
+    match e.into_inner() {
+        GcTask::Gc { callback, .. } | GcTask::UnsafeDestroyRange { callback, .. } => {
+            callback(Err(Error::from(ErrorInner::GcWorkerTooBusy)))
+        }
+        GcTask::PhysicalScanLock { callback, .. } => {
+            callback(Err(Error::from(ErrorInner::GcWorkerTooBusy)))
+        }
```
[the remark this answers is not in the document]
sticnarf: This does not compile, the `callback` type of `PhysicalScanLock` is different.

src/server/gc_worker/gc_worker.rs
sticnarf: Makes sense. I totally remove `check_is_busy`.

src/server/gc_worker/gc_worker.rs
```diff
@@ -959,17 +950,16 @@ where
...
-                    max_ts,
-                    start_key,
-                    limit,
-                    callback,
-                })
-                .or_else(handle_gc_task_schedule_error)
-        })
+
+        self.worker_scheduler
+            .schedule(GcTask::PhysicalScanLock {
```
Connor1996, line 955: Don't know when `phsical_scan_lock` is triggered. Would it block the outer process if it's not been executed for a long time.

sticnarf: It's not expected to use any longer. It's used by Green GC, but Green GC does seem to be enabled in a short time.

src/server/gc_worker/gc_worker.rs
sticnarf: @cfzjywxk @youjiali1995 Wanna take another look for any other concerns?

src/server/gc_worker/gc_worker.rs
```diff
@@ -668,7 +668,17 @@ where
...
-    Err(box_err!("failed to schedule gc task: {:?}", e))
+    let res = Err(box_err!("failed to schedule gc task: {:?}", e));
+    match e.into_inner() {
+        GcTask::Gc { callback, .. } | GcTask::UnsafeDestroyRange { callback, .. } => {
+            callback(Err(Error::from(ErrorInner::GcWorkerTooBusy)))
+        }
+        GcTask::PhysicalScanLock { callback, .. } => {
+            callback(Err(Error::from(ErrorInner::GcWorkerTooBusy)))
+        }
+        _ => {}
```
Connor1996, line 679: make sure other tasks added later won't forget to handle it.
```suggestion
        _ => unreachable!();
```

Connor1996 (commented):
rest LGTM

src/server/gc_worker/gc_worker.rs
youjiali1995: None

src/server/gc_worker/gc_worker.rs
sticnarf: OK. I explicitly match all variant and add a notice.

Connor1996 (approved):
LGTM

youjiali1995 (approved):

src/server/gc_worker/gc_worker.rs
```diff
@@ -668,7 +668,20 @@ where
 /// When we failed to schedule a `GcTask` to `GcRunner`, use this to handle the `ScheduleError`.
 fn handle_gc_task_schedule_error(e: ScheduleError<GcTask<impl KvEngine>>) -> Result<()> {
     error!("failed to schedule gc task"; "err" => %e);
-    Err(box_err!("failed to schedule gc task: {:?}", e))
+    let res = Err(box_err!("failed to schedule gc task: {:?}", e));
+    match e.into_inner() {
```
Connor1996, line 672: after checking `unsafe_destroy_range` in `service/kv.rs`, seems we don't need to call the callback. When the error is returned, callback won't called.
see
```rust
        let (cb, f) = paired_future_callback();
        let res = self.gc_worker.unsafe_destroy_range(
            req.take_context(),
            Key::from_raw(&req.take_start_key()),
            Key::from_raw(&req.take_end_key()),
            cb,
        );

        let task = async move {
            let res = match res {
                Err(e) => Err(e),       <-----------
                Ok(_) => f.await?,      <-----------
            };
            let mut resp = UnsafeDestroyRangeResponse::default();
            // Region error is impossible here.
            if let Err(e) = res {
                resp.set_error(format!("{}", e));
            }
            sink.success(resp).await?;
            GRPC_MSG_HISTOGRAM_STATIC
                .unsafe_destroy_range
                .observe(duration_to_sec(begin_instant.saturating_elapsed()));
            ServerResult::Ok(())
        }
```

Connor1996 (changes requested):
do not merge now

src/server/gc_worker/gc_worker.rs
sticnarf: I think this does not matter. It's better to keep all callbacks called to avoid potential problems.

Connor1996 (approved):

PR #11907, opened 2022-01-25
raftstore: skip deleting snapshot files in peer pending_remove is true (#11782)

### LINKED ISSUE #11746: raftstore: snapshot files may deleted while the peer state is still `PeerState::Applying`

ti-srebot:
cherry-pick #11782 to release-5.2
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/11907
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/11907:release-5.2-8c31f1a392ba
```

---

skip deleting snapshot files in peer pending_remove is true and the snapshot is being applied and canceled -- close #11746

This is to avoid the potential panic when the snapshot files are deleted, but the peer's status (Tombstone) is not persisted in disk due to tikv crash.

Close #11746 

What's Changed:
Skip deleting snapshot files in peer.rs when the snapshot is being applied and just canceled and the peer is pending removal. 
### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
Fix a potential panic (#11746) when snapshot files have been deleted but the peer's status is still Applying.
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2072,6 +2075,35 @@ where
 
     fn destroy_peer(&mut self, merged_by_target: bool) {
         fail_point!("destroy_peer");
+<<<<<<< HEAD
+=======
```
MyonKeminta, line 2079: Please resolve conflicts

tonyxuqqi: > Please resolve conflicts

done.

Connor1996 (approved):
LGTM

BusyJay (approved):

PR #11918, opened 2022-01-25
gc_worker: fix incorrect scheduled_tasks counting (#11904)

### LINKED ISSUE #11903: False GcWorkerTooBusy caused by incorrect scheduled_tasks

sticnarf:
cherry-pick #11904 to release-5.1-20220125

Closes #11903

What's Changed:

1. Callback is always called in `handle_gc_task_schedule_error`, so the `scheduled_tasks` counter is maintained correctly.
2. Allow `unsafe_destroy_range` to schedule even if the worker is full. Then, `unsafe_destroy_range` will still have the chance to execute when the GC worker is busy.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Fixes the bug that unsafe_destroy_range does not get executed when GC worker is busy
```

### DIFF
--- a/components/tikv_util/src/worker/pool.rs
+++ b/components/tikv_util/src/worker/pool.rs
@@ -114,6 +114,14 @@ impl<T: Display + Send> Scheduler<T> {
         if self.counter.load(Ordering::Acquire) >= self.pending_capacity {
             return Err(ScheduleError::Full(task));
         }
+        self.schedule_force(task)
+    }
+
+    /// Schedules a task to run.
+    ///
+    /// Different from the `schedule` function, the task will still be scheduled
+    /// if pending task number exceeds capacity.
+    pub fn schedule_force(&self, task: T) -> Result<(), ScheduleError<T>> {
         self.counter.fetch_add(1, Ordering::SeqCst);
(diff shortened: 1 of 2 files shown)


sticnarf:
PTAL @youjiali1995 @cfzjywxk

cfzjywxk (approved):

youjiali1995 (approved):

PR #11931, opened 2022-02-07
copr: Extra physical table id column

### LINKED ISSUE #11888: Add support for returning Table ID as a Column (EXTRA_PHYSICAL_TABLE_ID_COL_ID)

In TiDB, table partitioning is implemented by having several physical tables presented as one logical table, where each partition is mapped to a Physical Table ID.

When dynamic prune mode for table partitioning was implemented, a single TableReader was used for all needed partitions, and sent a coprocessor requests which included multiple Physical Table ID ranges resulting in row data could come from several partitions in the result chunk. This works well for handling the data, but if there is an ongoing transaction or pessimistic locks are requested (SELECT FOR UPDATE/SelectLock) then each row needs to have which partition/physical table id it came from, to be able to lookup in TiDBs transaction buffer or to send a lock to TiKV (which both needs the Physical Table ID).

Since the Physical Table ID is encoded in the Key for each row in the KV Store, we could solve this problem if the Physical Table ID could be added as an extra column in the result chunk.

Add a static column id '-3' / EXTRA_PHYSICAL_TABLE_ID_COL_ID, similar to int handle (-1) or EXTRA_PARTITION_ID_COL_ID (-2).
If a column with id -3 is requested, use the Table ID from the Key to fill in its 64 bit integer value. This should be done for both index and table reads.

- Add all partitioning columns to the read request and then recalculate which partition the row came from (a lot of overhead, both data/network wise and processing wise). Would not work for indexes not including all partitioning columns
- Get full key for each row and decode the table id on TiDB side. More data and processing overhead, similar implementation and no other use case yet.

### Teachability, Documentation, Adoption, Migration Strategy:

[Proof of Concept for TiDB UniStore](https://github.com/pingcap/tidb/pull/31634/files#diff-89dead167fcc4733d04c187d2bcd4ce4b987ecf204fed9ef9c15ead6b6604f07), see the unistore path code changes for details, `model.ExtraPhysTblID` is what is proposed to be `EXTRA_PHYSICAL_TABLE_ID_COL_ID` in TiKV.

mjonss:

Issue Number: Close #11888

What's Changed:
Added support for returning the Physical Table ID as a column for table scan and index scan

```commit-message
Added EXTRA_PHYSICAL_TABLE_ID_COL_ID support, to support TiDB's table partition dynamic prune mode, where a single request includes multiple partitions, but when Pessimistic lock (like `SELECT FOR UPDATE`) or ongoing transaction (having something in the tidb session local transaction buffer) each row needs its partition id / physical table ID to either lock that row or to check against the transaction buffer.
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Added support for EXTRA_PHYSICAL_TABLE_ID_COL_ID needed for TiDB's table partition dynamic prune mode.
```


components/tidb_query_executors/src/index_scan_executor.rs
```diff
@@ -738,7 +786,7 @@ impl IndexScanExecutorImpl {
             DecodePartitionIdOp::PID(pid) => {
                 // If need partition id, append partition id to the last column.
```
wjhuang2016, line 787: Should we update this comment?

wjhuang2016 (commented):
rest LGTM

components/tidb_query_executors/src/index_scan_executor.rs
mjonss: Updated.

wjhuang2016 (approved):

components/tidb_query_executors/src/index_scan_executor.rs
```diff
@@ -727,6 +763,18 @@ impl IndexScanExecutorImpl {
...
 
+    #[inline]
+    fn process_physical_table_id_column(
+        &mut self,
+        key: &[u8],
+        columns: &mut LazyBatchColumnVec,
+    ) -> Result<()> {
+        let table_id = table::decode_table_id(key)?;
+        let col_index = columns.columns_len() - 1;
+        columns[col_index].mut_decoded().push_int(Some(table_id));
```
andylokandy, line 774: Why `physical_table_id_column` must be the last column?

components/tidb_query_executors/src/index_scan_executor.rs
```diff
@@ -736,9 +784,10 @@ impl IndexScanExecutorImpl {
         match decode_pid {
             DecodePartitionIdOp::Nop => {}
             DecodePartitionIdOp::PID(pid) => {
-                // If need partition id, append partition id to the last column.
+                // If need partition id, append partition id to the last column
+                // before physical table id column if exists.
                 let pid = NumberCodec::decode_i64(pid);
-                let idx = columns.columns_len() - 1;
+                let idx = columns.columns_len() - 1 - self.physical_table_id_column_cnt;
```
andylokandy, line 790:
```suggestion
                let idx = columns.columns_len() - self.physical_table_id_column_cnt - 1;
```

components/tidb_query_executors/src/index_scan_executor.rs
mjonss: When added to the list of columns in TiDB, it will be last (idea was taken from the Global Index handling of / EXTRA_PARTITION_ID_COL_ID). Also I don't really understand why the order seems to be important for index scan, but not for table scan.

@andylokandy do you want me to try to change the code so it is not dependent of column order?

Also I have not really understood why TiKV should be aware of EXTRA_PARTITION_ID_COL_ID at all, since I wonder if it is not just a simple i64 column for TiKV, that TiDB should handle fully?

breeswish:
Looks like TiDB already knows the table id and index id from the start key. Is it necessary for TiKV to return it again?

mjonss:
> Looks like TiDB already knows the table id and index id from the start key, when the request is going to be sent to TiKV. Is it necessary for TiKV to return it again?

When sending the ids/ranges under dynamic partition prune mode, all partitions for a table are mixed together so there is no-longer a 1:1 relationship between partition and result, which it had in the static prune mode. Which is the reason for adding the Physical Table ID in the result, already in the storage engine, so all coprocessors and executors can use it directly and send it to its parent. It is only used for pessimistic locking (SELECT FOR UPDATE) and TiDB transaction buffer (through the UnionScan executor).


Another idea I tried to explore was the startKey which is returned in the distsql/select_result, but at least for UniStore engine it did not work:

If the region is not split, like the configuration file contains `split-table = false` then the start key will not be per physical table, but for several tables/partitions in the same region.


SQL to test with:
```
create table pt (id bigint primary key, col varchar(255)) partition by hash (id) partitions 3;
insert into pt values (1, "First AAA"), (2, "Second BBB"), (3, "Third CCC"), (5, "Fifth EEE");
set tidb_partition_prune_mode = 'dynamic';
select * from pt for update;
```

Hack to log the result from engine:
```
diff --git a/distsql/select_result.go b/distsql/select_result.go
index b2eef6e6f5..192fe9a1ce 100644
--- a/distsql/select_result.go
+++ b/distsql/select_result.go
@@ -239,6 +239,12 @@ func (r *selectResult) fetchResp(ctx context.Context) error {
                        r.feedback.Update(resultSubset.GetStartKey(), r.selectResp.OutputCounts, r.selectResp.Ndvs)
                }
                r.partialCount++
+               logutil.Logger(ctx).Info("fetchResp",
+                       zap.String("resultSubset.startKey", resultSubset.GetStartKey().String()))
+               for i := range r.selectResp.Chunks {
+                       logutil.Logger(ctx).Info("fetchResp",
+                               zap.String("r.selectResp.chunk[i]", r.selectResp.Chunks[i].String()))
+               }
 
                hasStats, ok := resultSubset.(CopRuntimeStats)
                if ok {
```

Log output when using the UniStore:
```
[2022/02/17 00:22:55.088 +01:00] [INFO] [select_result.go:242] ["fetchResp"] [conn=3] [resultSubset.startKey=7480000000000000405f720000000000000000]
[2022/02/17 00:22:55.088 +01:00] [INFO] [select_result.go:245] ["fetchResp"] [conn=3] ["r.selectResp.chunk[i]"="rows_data:\"\\010\\006\\002\\022Third CCC\\010\\200\\001\\010\\002\\002\\022First AAA\\010\\202\\001\\010\\004\\002\\024Second BBB\\010\\204\\001\\010\\n\\002\\022Fifth EEE\\010\\204\\001\" "]
```
Here you can see all the rows from the different partitions in the same chunk for the same startKey.
Which means rows from several partitions/ranges is merged into the same chunk and startKey, which is why I assumed we needed to include the Physical Table ID for every row in the chunk already from the store.

For TiKV I got three concurrent sets, each with its own startKey (also for 8192 partitions, so it seems to be consistent).

@breeswish is that guaranteed that each range/physical table id will be sent with its own start key from TiKV?
Is that a bug in UniStore mock-engine that it merges rows from different ranges/physical table ids into the same chunk?

Would it still be OK to add the Physical Table ID in the engine, in case some coprocessors needs the correct count of columns or something else?
Since this solution makes it consistent data flow wise (otherwise will the column for Physical Table ID first be created when the result is received by TiDB).

breeswish:
@mjonss Sorry for the late response. What TiKV receives depends on how the TiDB's Coprocessor module splits the request. The TiKV indeed works correctly now if the request's range contains multiple partitions, as long as they have the same schema: TiKV only scans the range that TiDB tells, and use the schema from the TiDB to decode the rows.

breeswish:
@mjonss You also mentioned about the `startKey` in the response. If I remember correctly it should be a field that you should never rely on. Previously it is introduced by the Streaming mode. However the implementation of the Streaming mode is buggy so that Streaming mode is completely deprecated now. For the same reason, we can never trust what TiKV returns in the `startKey`.

components/tidb_query_executors/src/index_scan_executor.rs
andylokandy: > do you want me to try to change the code so it is not dependent of column order?

I'd like to make it an explicit const number denoting the nth column counting back from the last one.

> Also I have not really understood why TiKV should be aware of EXTRA_PARTITION_ID_COL_ID at all, since I wonder if it is not just a simple i64 column for TiKV, that TiDB should handle fully?

@wjhuang2016 Do you have any idea?

wjhuang2016: > > do you want me to try to change the code so it is not dependent of column order?
> 
> I'd like to make it an explicit const number denoting the nth column counting back from the last one.
> 
> > Also I have not really understood why TiKV should be aware of EXTRA_PARTITION_ID_COL_ID at all, since I wonder if it is not just a simple i64 column for TiKV, that TiDB should handle fully?
> 
> @wjhuang2016 Do you have any idea?

No， @tiancaiamao PTAL

breeswish: An explicit order looks good to me, as this module is a tidb_query module and its logic can be highly coherent with the TiDB.

components/tidb_query_executors/src/index_scan_executor.rs
```diff
@@ -854,13 +903,20 @@ mod tests {
                 ci.set_pk_handle(true);
                 ci
             },
+            {
+                let mut ci = ColumnInfo::default();
+                ci.as_mut_accessor().set_tp(FieldTypeTp::LongLong);
+                ci.set_column_id(table::EXTRA_PHYSICAL_TABLE_ID_COL_ID);
+                ci
+            },
```
breeswish, line 911: Could you also have a test to ensure that, when PhysicalTableId column is NOT pushed down, the code works?

breeswish (commented):
The rest LGTM. Thanks

breeswish (approved):
Cool. Thanks!

Issue #6107, opened 2019-11-29
tikv-ctl unsafe recover can lead to inconsistent replica list


BusyJay:
Suppose region 1 has three replicas A, B and C, and A is leader. A adds a new replica D and then both A and D are corrupted, and B has applied the conf change while D has not. Using tikv-ctl will remove both A and D from B's local peer list and just remove A from C's. After B and C are started, C will apply the conf change add D to its list. So B's peer list becomes (B, C), and C's list is (B, C, D).

To get a consistent list, we need to skip mutating the list after an unsafe recovery. A simple way is maintain a tombstone store list on every node. When applying conf change, ignore the targets that locate on the store. Note that conf version should still be increased to reach the correct version.

/cc @disksing @hicqu

hicqu:
Stores could be only `tombstone` for some special regions, but still useful for other regions.
So, there are 2 question:
1. The list is for all regions, or for some special regions?
2. When will the list be clear?

BusyJay:
Another way is to scan all the logs and mutate the id to a special value, which we should treat as no-op. It requires to update region peer list and logs atomically.

disksing:
any update?

BusyJay:
It's still under discussion.

I think the safest way for the problem is generating a snapshot from a node that has applied most up to date data, and restoring it on all other remaining node.

This approach is generic and work on many situation.

NingLin-P:
I have an idea for this problem: for each peer removed by tikv-ctl append one `RemovePeer` raft log on every peer of the region because every peer has these logs so they must commit and send to the apply worker. Also tikv-ctl remove peer from store's local peer list as before, and because there is at most one unapplied `confchange`, so at most one `RemovePeer` logs appended by tikv-ctl will success, others will result in a `remove missing peer` error log, for the epoch of these `RemovePeer` logs we can simply set to the same `epoch { ver: exist_ver, conf_ver: exist_confver + 1 }`.

BusyJay:
> because every peer has these logs so they must commit and send to the apply worker

So you mean propose a conf change after restarting all the remaining nodes?

>  because there is at most one unapplied confchange

No, it's only guaranteed that at most one uncommitted conf change.

NingLin-P:
> So you mean propose a conf change after restarting all the remaining nodes
No, just append log (write to rocksdb) without propose when tikv-ctl remove peer.
 
> No, it's only guaranteed that at most one uncommitted conf change.
[...]

PR #11932, opened 2022-02-07
raftstore: Introduce force leader state

Connor1996:

Issue Number: Close https://github.com/tikv/tikv/issues/6107

What's Changed:

Introduce force leader state to advance commit index even if the quorum is not formed. This is used for online unsafe recover.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None 
```

### DIFF
--- a/components/error_code/src/raftstore.rs
+++ b/components/error_code/src/raftstore.rs
@@ -27,6 +27,7 @@ define_error_codes!(
     SERVER_IS_BUSY => ("ServerIsBusy", "", ""),
     DATA_IS_NOT_READY => ("DataIsNotReady", "", ""),
     DEADLINE_EXCEEDED => ("DeadlineExceeded", "", ""),
+    RECOVERY_IN_PROGRESS => ("RecoveryInProgress", "", ""),
 
     SNAP_ABORT => ("SnapAbort", "", ""),
     SNAP_TOO_MANY => ("SnapTooMany", "", ""),
@@ -59,6 +60,8 @@ impl ErrorCodeExt for errorpb::Error {
             PROPOSAL_IN_MERGING_MODE
         } else if self.has_data_is_not_ready() {
             DATA_IS_NOT_READY
+        } else if self.has_data_is_not_ready() {
+            RECOVERY_IN_PROGRESS
         } else {
             UNKNOWN
         }
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -590,6 +590,12 @@ where
                         continue;
                     }
 
+                    if self.fsm.peer.force_leader {
+                        // in force leader state, forbid requests to make the recovery progress less error-prone
+                        cmd.callback.invoke_with_response(new_error(Error::RecoveryInProgress(self.region_id())));
+                        continue;
+                    }
+
                     let req_size = cmd.request.compute_size();
                     if self.ctx.cfg.cmd_batch
                         && self.fsm.batch_req_builder.can_batch(&cmd.request, req_size)
@@ -978,8 +984,29 @@ where
             } => self.on_capture_change(cmd, region_epoch, callback),
             SignificantMsg::LeaderCallback(cb) => {
                 self.on_leader_callback(cb);
-            }
+            },
+            SignificantMsg::EnterForceLeaderState => self.on_enter_force_leader(),
+            SignificantMsg::ExitForceLeaderState => self.on_exit_force_leader(),
         }
+    }
+
+    fn on_enter_force_leader(&mut self) {
+        self.fsm.peer.force_leader = true;
+
+        // become candidate first to increase term
+        self.fsm.peer.raft_group.raft.become_candidate();
+        self.fsm.peer.raft_group.raft.become_leader();
+
+        // append an empty entry to truncate logs that are not committed
+        let mut entry = raft::eraftpb::Entry::default();
+        entry.term = self.fsm.peer.term();
+        entry.index = self.fsm.peer.raft_group.raft.raft_log.committed + 1;
+        self.fsm.peer.raft_group.raft.raft_log.append(&[entry]);
+    }
+
+    fn on_exit_force_leader(&mut self) {
+        self.fsm.peer.force_leader = false;
+        self.fsm.peer.raft_group.raft.become_follower(self.fsm.peer.term(), raft::INVALID_ID);
     }
 
     fn on_persisted_msg(&mut self, peer_id: u64, ready_number: u64, _send_time: TiInstant) {
@@ -1037,6 +1064,13 @@ where
                 self.register_split_region_check_tick();
                 self.fsm.peer.heartbeat_pd(self.ctx);
                 self.register_pd_heartbeat_tick();
+            }
+
+            if self.fsm.peer.force_leader {
+                if ss.raft_state == StateRole::Follower {
+                    // for some reason, it's not leader anymore
+                    self.on_enter_force_leader();
+                }
             }
         }
     }
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -590,12 +590,6 @@ where
                         continue;
                     }
 
-                    if self.fsm.peer.force_leader {
-                        // in force leader state, forbid requests to make the recovery progress less error-prone
-                        cmd.callback.invoke_with_response(new_error(Error::RecoveryInProgress(self.region_id())));
-                        continue;
-                    }
-
                     let req_size = cmd.request.compute_size();
                     if self.ctx.cfg.cmd_batch
                         && self.fsm.batch_req_builder.can_batch(&cmd.request, req_size)
@@ -994,8 +988,10 @@ where
         self.fsm.peer.force_leader = true;
 
         // become candidate first to increase term
-        self.fsm.peer.raft_group.raft.become_candidate();
-        self.fsm.peer.raft_group.raft.become_leader();
+        if !self.fsm.peer.is_leader() {
+            self.fsm.peer.raft_group.raft.become_candidate();
+            self.fsm.peer.raft_group.raft.become_leader();
+        }
 
         // append an empty entry to truncate logs that are not committed
         let mut entry = raft::eraftpb::Entry::default();
@@ -3446,6 +3442,13 @@ where
         let region_id = self.region_id();
         let leader_id = self.fsm.peer.leader_id();
         let request = msg.get_requests();
+
+        if self.fsm.peer.force_leader {
+            // in force leader state, forbid requests to make the recovery progress less error-prone
+            if !(msg.has_admin_request() && msg.get_admin_request().get_cmd_type() == AdminCmdType::ChangePeer) {
+                return Err(Error::RecoveryInProgress(self.region_id()));
+            }
+        }
 
         // ReadIndex can be processed on the replicas.
         let is_read_index_request =
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -985,6 +985,11 @@ where
     }
 
     fn on_enter_force_leader(&mut self) {
+        info!(
+            "enter force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+        );
         self.fsm.peer.force_leader = true;
 
         // become candidate first to increase term
@@ -992,17 +997,17 @@ where
             self.fsm.peer.raft_group.raft.become_candidate();
             self.fsm.peer.raft_group.raft.become_leader();
         }
+        assert!(self.fsm.peer.is_leader());
 
-        // append an empty entry to truncate logs that are not committed
-        let mut entry = raft::eraftpb::Entry::default();
-        entry.term = self.fsm.peer.term();
-        entry.index = self.fsm.peer.raft_group.raft.raft_log.committed + 1;
-        self.fsm.peer.raft_group.raft.raft_log.append(&[entry]);
+        // forward commit index
+        self.fsm.peer.raft_group.raft.raft_log.committed = self.fsm.peer.raft_group.raft.raft_log.last_index();
+        self.fsm.has_ready = true;
     }
 
     fn on_exit_force_leader(&mut self) {
         self.fsm.peer.force_leader = false;
         self.fsm.peer.raft_group.raft.become_follower(self.fsm.peer.term(), raft::INVALID_ID);
+        self.fsm.has_ready = true;
     }
 
     fn on_persisted_msg(&mut self, peer_id: u64, ready_number: u64, _send_time: TiInstant) {
@@ -1065,7 +1070,7 @@ where
             if self.fsm.peer.force_leader {
                 if ss.raft_state == StateRole::Follower {
                     // for some reason, it's not leader anymore
-                    self.on_enter_force_leader();
+                    warn!("step to follower in force leader state");
                 }
             }
         }
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -978,7 +978,7 @@ where
             } => self.on_capture_change(cmd, region_epoch, callback),
             SignificantMsg::LeaderCallback(cb) => {
                 self.on_leader_callback(cb);
-            },
+            }
             SignificantMsg::EnterForceLeaderState => self.on_enter_force_leader(),
             SignificantMsg::ExitForceLeaderState => self.on_exit_force_leader(),
         }
@@ -1000,13 +1000,18 @@ where
         assert!(self.fsm.peer.is_leader());
 
         // forward commit index
-        self.fsm.peer.raft_group.raft.raft_log.committed = self.fsm.peer.raft_group.raft.raft_log.last_index();
+        self.fsm.peer.raft_group.raft.raft_log.committed =
+            self.fsm.peer.raft_group.raft.raft_log.last_index();
         self.fsm.has_ready = true;
     }
 
     fn on_exit_force_leader(&mut self) {
         self.fsm.peer.force_leader = false;
-        self.fsm.peer.raft_group.raft.become_follower(self.fsm.peer.term(), raft::INVALID_ID);
+        self.fsm
+            .peer
+            .raft_group
+            .raft
+            .become_follower(self.fsm.peer.term(), raft::INVALID_ID);
         self.fsm.has_ready = true;
     }
 
@@ -3450,7 +3455,9 @@ where
 
         if self.fsm.peer.force_leader {
             // in force leader state, forbid requests to make the recovery progress less error-prone
-            if !(msg.has_admin_request() && msg.get_admin_request().get_cmd_type() == AdminCmdType::ChangePeer) {
+            if !(msg.has_admin_request()
+                && msg.get_admin_request().get_cmd_type() == AdminCmdType::ChangePeer)
+            {
                 return Err(Error::RecoveryInProgress(self.region_id()));
             }
         }
--- a/components/raftstore/src/store/msg.rs
+++ b/components/raftstore/src/store/msg.rs
@@ -291,6 +291,8 @@ where
         callback: Callback<SK>,
     },
     LeaderCallback(Callback<SK>),
+    EnterForceLeaderState,
+    ExitForceLeaderState,
 }
 
 /// Message that will be sent to a peer.
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -427,6 +427,8 @@ where
     /// If a snapshot is being applied asynchronously, messages should not be sent.
     pending_messages: Vec<eraftpb::Message>,
 
+    pub force_leader: bool,
+
     /// Record the instants of peers being added into the configuration.
     /// Remove them after they are not pending any more.
     pub peers_start_pending_time: Vec<(u64, Instant)>,
@@ -586,6 +588,7 @@ where
             leader_unreachable: false,
             pending_remove: false,
             should_wake_up: false,
+            force_leader: false,
             pending_merge_state: None,
             want_rollback_merge_peers: HashSet::default(),
             pending_request_snapshot_count: Arc::new(AtomicUsize::new(0)),
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -2500,7 +2500,7 @@ where
 
         let promoted_commit_index = after_progress.maximal_committed_index().0;
         if current_progress.is_singleton() // It's always safe if there is only one node in the cluster.
-            || promoted_commit_index >= self.get_store().truncated_index()
+            || promoted_commit_index >= self.get_store().truncated_index() || self.force_leader
         {
             return Ok(());
         }
@@ -3268,6 +3268,9 @@ where
             // The message is dropped silently, this usually due to leader absence
             // or transferring leader. Both cases can be considered as NotLeader error.
             return Err(Error::NotLeader(self.region_id, None));
+        } else if self.force_leader {
+            // forward the commit index
+            self.raft_group.raft.raft_log.committed = self.raft_group.raft.raft_log.last_index();
         }
 
         Ok(propose_index)
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -3071,6 +3071,9 @@ where
             // The message is dropped silently, this usually due to leader absence
             // or transferring leader. Both cases can be considered as NotLeader error.
             return Err(Error::NotLeader(self.region_id, None));
+        } else if self.force_leader {
+            // forward the commit index
+            self.raft_group.raft.raft_log.committed = self.raft_group.raft.raft_log.last_index();
         }
 
         Ok(Either::Left(propose_index))
--- a/components/raftstore/src/store/util.rs
+++ b/components/raftstore/src/store/util.rs
@@ -746,10 +746,6 @@ impl<
     }
 }
 
-pub fn integration_on_half_fail_quorum_fn(voters: usize) -> usize {
-    (voters + 1) / 2 + 1
-}
-
 #[derive(PartialEq, Eq, Debug)]
 pub enum ConfChangeKind {
     // Only contains one configuration change
@@ -1892,16 +1888,6 @@ mod tests {
                 check_region_epoch(&req, &region, false).unwrap_err();
                 check_region_epoch(&req, &region, true).unwrap_err();
             }
-        }
-    }
-
-    #[test]
-    fn test_integration_on_half_fail_quorum_fn() {
-        let voters = vec![1, 2, 3, 4, 5, 6, 7];
-        let quorum = vec![2, 2, 3, 3, 4, 4, 5];
-        for (voter_count, expected_quorum) in voters.into_iter().zip(quorum) {
-            let quorum = super::integration_on_half_fail_quorum_fn(voter_count);
-            assert_eq!(quorum, expected_quorum);
         }
     }
 
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -27,11 +27,11 @@ use engine_traits::{
 };
 use file_system::IORateLimiter;
 use pd_client::PdClient;
+use raftstore::router::RaftStoreRouter;
 use raftstore::store::fsm::store::{StoreMeta, PENDING_MSG_CAP};
 use raftstore::store::fsm::{create_raft_batch_system, RaftBatchSystem, RaftRouter};
 use raftstore::store::transport::CasualRouter;
 use raftstore::store::*;
-use raftstore::router::RaftStoreRouter;
 use raftstore::{Error, Result};
 use tikv::config::TiKvConfig;
 use tikv::server::Result as ServerResult;
@@ -1289,20 +1289,16 @@ impl<T: Simulator> Cluster<T> {
 
     pub fn enter_force_leader(&mut self, region_id: u64, store_id: u64) {
         let router = self.sim.rl().get_router(store_id).unwrap();
-        router.significant_send(
-            region_id,
-            SignificantMsg::EnterForceLeaderState,
-        )
-        .unwrap();
+        router
+            .significant_send(region_id, SignificantMsg::EnterForceLeaderState)
+            .unwrap();
     }
 
     pub fn exit_force_leader(&mut self, region_id: u64, store_id: u64) {
         let router = self.sim.rl().get_router(store_id).unwrap();
-        router.significant_send(
-            region_id,
-            SignificantMsg::ExitForceLeaderState,
-        )
-        .unwrap();
+        router
+            .significant_send(region_id, SignificantMsg::ExitForceLeaderState)
+            .unwrap();
     }
 
     pub fn must_split(&mut self, region: &metapb::Region, split_key: &[u8]) {
--- a/components/test_raftstore/src/cluster.rs
+++ b/components/test_raftstore/src/cluster.rs
@@ -31,6 +31,7 @@ use raftstore::store::fsm::store::{StoreMeta, PENDING_MSG_CAP};
 use raftstore::store::fsm::{create_raft_batch_system, RaftBatchSystem, RaftRouter};
 use raftstore::store::transport::CasualRouter;
 use raftstore::store::*;
+use raftstore::router::RaftStoreRouter;
 use raftstore::{Error, Result};
 use tikv::config::TiKvConfig;
 use tikv::server::Result as ServerResult;
@@ -1282,6 +1283,24 @@ impl<T: Simulator> Cluster<T> {
                 callback: cb,
                 source: "test".into(),
             },
+        )
+        .unwrap();
+    }
+
+    pub fn enter_force_leader(&mut self, region_id: u64, store_id: u64) {
+        let router = self.sim.rl().get_router(store_id).unwrap();
+        router.significant_send(
+            region_id,
+            SignificantMsg::EnterForceLeaderState,
+        )
+        .unwrap();
+    }
+
+    pub fn exit_force_leader(&mut self, region_id: u64, store_id: u64) {
+        let router = self.sim.rl().get_router(store_id).unwrap();
+        router.significant_send(
+            region_id,
+            SignificantMsg::ExitForceLeaderState,
         )
         .unwrap();
     }
--- a/components/raftstore/src/errors.rs
+++ b/components/raftstore/src/errors.rs
@@ -58,6 +58,9 @@ pub enum Error {
     #[error("store ids {0:?}, errmsg {1}")]
     DiskFull(Vec<u64>, String),
 
+    #[error("region {0} is the recovery progress")]
+    RecoveryInProgress(u64),
+
     #[error(
         "key {} is not in region key range [{}, {}) for region {}",
         log_wrappers::Value::key(.0),
@@ -273,6 +276,7 @@ impl ErrorCodeExt for Error {
             Error::RegionNotFound(_) => error_code::raftstore::REGION_NOT_FOUND,
             Error::NotLeader(..) => error_code::raftstore::NOT_LEADER,
             Error::DiskFull(..) => error_code::raftstore::DISK_FULL,
+            Error::RecoveryInProgress(..) => error_code::raftstore::RECOVERY_IN_PROGRESS,
             Error::StaleCommand => error_code::raftstore::STALE_COMMAND,
             Error::RegionNotInitialized(_) => error_code::raftstore::REGION_NOT_INITIALIZED,
             Error::KeyNotInRegion(..) => error_code::raftstore::KEY_NOT_IN_REGION,
--- a/tests/integrations/raftstore/test_hibernate.rs
+++ b/tests/integrations/raftstore/test_hibernate.rs
@@ -27,13 +27,12 @@ fn test_force_leader() {
 
     let put = new_put_cmd(b"k2", b"v2");
     let mut region = cluster.get_region(b"k2");
-    let req = new_request(
-        region.get_id(),
-        region.take_region_epoch(),
-        vec![put],
-        true,
+    let req = new_request(region.get_id(), region.take_region_epoch(), vec![put], true);
+    assert!(
+        cluster
+            .call_command_on_leader(req, Duration::from_millis(10))
+            .is_err()
     );
-    assert!(cluster.call_command_on_leader(req, Duration::from_millis(10)).is_err());
 
     cluster.enter_force_leader(1, 1);
     cluster.pd_client.must_remove_peer(1, new_peer(3, 3));
@@ -41,13 +40,10 @@ fn test_force_leader() {
     cluster.pd_client.must_remove_peer(1, new_peer(5, 5));
     let put = new_put_cmd(b"k3", b"v3");
     let mut region = cluster.get_region(b"k2");
-    let req = new_request(
-        region.get_id(),
-        region.take_region_epoch(),
-        vec![put],
-        true,
-    );
-    let resp = cluster.call_command_on_leader(req, Duration::from_millis(10)).unwrap();
+    let req = new_request(region.get_id(), region.take_region_epoch(), vec![put], true);
+    let resp = cluster
+        .call_command_on_leader(req, Duration::from_millis(10))
+        .unwrap();
     assert!(resp.get_header().has_error());
     cluster.exit_force_leader(1, 1);
 
--- a/tests/integrations/raftstore/test_hibernate.rs
+++ b/tests/integrations/raftstore/test_hibernate.rs
@@ -12,6 +12,7 @@ use test_raftstore::*;
 use tikv_util::time::Instant;
 use tikv_util::HandyRwLock;
 
+#[test]
 fn test_force_leader() {
     let mut cluster = new_node_cluster(0, 5);
 
@@ -23,11 +24,35 @@ fn test_force_leader() {
     cluster.stop_node(4);
     cluster.stop_node(5);
 
-    assert!(cluster.put(b"k2", b"v2").is_err());
-    cluster.force_leader();
+    let put = new_put_cmd(b"k2", b"v2");
+    let mut region = cluster.get_region(b"k2");
+    let req = new_request(
+        region.get_id(),
+        region.take_region_epoch(),
+        vec![put],
+        true,
+    );
+    assert!(cluster.call_command_on_leader(req, Duration::from_millis(10)).is_err());
 
-    cluster.must_put(b"k3", b"v3");
-    cluster.stop_nod
+    cluster.enter_force_leader(1, 1);
+    cluster.pd_client.must_remove_peer(1, new_peer(3, 3));
+    cluster.pd_client.must_remove_peer(1, new_peer(4, 4));
+    cluster.pd_client.must_remove_peer(1, new_peer(5, 5));
+    let put = new_put_cmd(b"k3", b"v3");
+    let mut region = cluster.get_region(b"k2");
+    let req = new_request(
+        region.get_id(),
+        region.take_region_epoch(),
+        vec![put],
+        true,
+    );
+    assert!(cluster.call_command_on_leader(req, Duration::from_millis(10)).is_err());
+    cluster.exit_force_leader(1, 1);
+
+    cluster.must_put(b"k4", b"v4");
+    assert_eq!(cluster.must_get(b"k2"), None);
+    assert_eq!(cluster.must_get(b"k3"), None);
+    assert_eq!(cluster.must_get(b"k4"), Some(b"v4".to_vec()));
 }
 
 #[test]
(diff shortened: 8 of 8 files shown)


Connor1996:
PTAL too @v01dstar

v01dstar (commented):
Losing leadership in election timeout while `check_quorum = true` should be taken into consideration?

Connor1996:
> 

Yes, do you want to add an API to change that? If not, I'll add it.

Connor1996:
> Losing leadership in election timeout while `check_quorum = true` should be taken into consideration?

Yes, do you want to add an API to change that? If not, I'll do it.

components/error_code/src/raftstore.rs
```diff
@@ -59,6 +60,8 @@ impl ErrorCodeExt for errorpb::Error {
             PROPOSAL_IN_MERGING_MODE
         } else if self.has_data_is_not_ready() {
             DATA_IS_NOT_READY
+        } else if self.has_data_is_not_ready() {
```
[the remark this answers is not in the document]
Connor1996: oh, should update the proto first

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1279,6 +1312,11 @@ where
                 self.register_raft_gc_log_tick();
                 self.register_check_leader_lease_tick();
             }
+
+            if self.fsm.peer.force_leader && r == StateRole::Follower {
+                // for some reason, it's not leader anymore
+                warn!("step to follower in force leader state");
```
5kbpers, line 1318: How could this happen? (And if we need this log, `peer_id` and `region_id` should be included.)

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3946,6 +3984,15 @@ where
         let leader_id = self.fsm.peer.leader_id();
         let request = msg.get_requests();
 
+        if self.fsm.peer.force_leader {
```
5kbpers, line 3987: Do we need a test for this?

Connor1996: it's already covered by the test, will add a comment in the test

Connor1996:
> I think there is a remaining problem about the learner role with the force leader mechanism, see [#6107 (comment)](https://github.com/tikv/tikv/issues/6107#issuecomment-869450923)

Good point. Adding a test to see if the learner can call `became_leader` successfully.

Connor1996:
> I think there is a remaining problem about the learner role with the force leader mechanism, see [#6107 (comment)](https://github.com/tikv/tikv/issues/6107#issuecomment-869450923)

Add a case to force the learner to be the leader. Seems no problem, as we can promote it to be a voter once in force leader state.

NingLin-P:
> Add a case to force the learner to be the leader. Seems no problem, as we can promote it to be a voter once in force leader state.

It may more than just whether a learner can become leader, I worry about what may happen after a learner became leader as we always assume that only voter can become leader. /cc @BusyJay

components/raftstore/src/store/peer.rs
```diff
@@ -3676,6 +3679,9 @@ where
             // The message is dropped silently, this usually due to leader absence
             // or transferring leader. Both cases can be considered as NotLeader error.
             return Err(Error::NotLeader(self.region_id, None));
+        } else if self.force_leader {
```
tonyxuqqi, line 3682: minor suggestion: else if can be if.  Same for other similar coded.
also maybe it's easier to read and maintain to create a tiny method forward_commit_index to wrap the logic as it appears in multiple places.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1200,9 +1200,42 @@ where
...
+        }
+        assert!(self.fsm.peer.is_leader());
+
+        // forward commit index
+        self.fsm.peer.raft_group.raft.raft_log.committed =
+            self.fsm.peer.raft_group.raft.raft_log.last_index();
+        self.fsm.has_ready = true;
+    }
+
+    fn on_exit_force_leader(&mut self) {
```
tonyxuqqi, line 1229: minor: should add similar log info like on_enter_force_leader

components/raftstore/src/store/msg.rs
```diff
@@ -308,6 +308,25 @@ where
         context: GetEntriesContext,
         res: Box<RaftlogFetchResult>,
```
tonyxuqqi, line 309: question: Will the compiler be confused which RaftlogFetchResult to use? As it's imported from store::RaftlogFetchResult and defined below too.

components/raftstore/src/store/msg.rs
```diff
@@ -308,6 +308,25 @@ where
...
     },
+    EnterForceLeaderState,
+    ExitForceLeaderState,
+}
+
+#[derive(Debug)]
+pub enum RaftlogFetchResult {
```
tonyxuqqi, line 316: question: where is it used? besides fetching and fetched, is it possible to fail?

components/raftstore/src/store/msg.rs
Connor1996: my bad, define it twice when resolving the conflict.

components/raftstore/src/store/msg.rs
Connor1996: unrelated with this PR, see https://github.com/tikv/tikv/pull/11409
And it's possible to fail returning with ents of `raft::Result`

components/raftstore/src/store/peer.rs
Connor1996: forwarding commit index is simple enough, no need to extract it.

components/error_code/src/raftstore.rs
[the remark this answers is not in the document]
Connor1996: done

Connor1996:
> > Add a case to force the learner to be the leader. Seems no problem, as we can promote it to be a voter once in force leader state.
> 
> It may more than just whether a learner can become leader, I worry about what may happen after a learner became leader as we always assume that only voter can become leader. /cc @BusyJay

@NingLin-P We'll make sure on PD side that making learner to be leader only if it has the latest log

BusyJay:
> It may more than just whether a learner can become leader, I worry about what may happen after a learner became leader as we always assume that only voter can become leader. /cc @BusyJay

Seems it's the only choice if quorum are gone forever.

components/raftstore/src/store/peer.rs
```diff
@@ -3830,6 +3838,9 @@ where
...
+            // forward the commit index
+            self.raft_group.raft.raft_log.committed = self.raft_group.raft.raft_log.last_index();
```
BusyJay, line 3843: I suggest to return error instead of committing any logs.

components/raftstore/src/store/peer.rs
```diff
@@ -4058,6 +4069,9 @@ where
...
```
BusyJay, line 4074: This breaks the assumption that only persisted entries get committed.

components/test_raftstore/src/cluster.rs
```diff
@@ -1288,6 +1288,20 @@ impl<T: Simulator> Cluster<T> {
         .unwrap();
     }
 
+    pub fn enter_force_leader(&mut self, region_id: u64, store_id: u64) {
```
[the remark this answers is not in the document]
Connor1996: what do you mean by PD protocol?

components/raftstore/src/store/peer.rs
Connor1996: we need to commit the new proposal, e.g. remove the failed peer, in force leader state

components/raftstore/src/store/peer.rs
Connor1996: change it

Connor1996:
PTAL again @BusyJay @NingLin-P

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1260,9 +1260,58 @@ where
...
+            "enter force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+        );
+        self.fsm.peer.force_leader = true;
+
+        if !self.fsm.peer.is_leader() {
+            // become candidate first to increase term
+            self.fsm.peer.raft_group.raft.become_candidate();
+            // trigger candidate twice to increase term by 2, to avoid there is
```
BusyJay, line 1279: What if there is a leader whose term is more than 3?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1260,9 +1260,58 @@ where
...
+            // a existing leader by accident.
+            self.fsm.peer.raft_group.raft.become_candidate();
+            self.fsm.peer.raft_group.raft.become_leader();
+        }
+        assert!(self.fsm.peer.is_leader());
+        self.fsm.peer.raft_group.raft.set_check_quorum(false);
+
+        // forward commit index
+        self.fsm.peer.raft_group.raft.raft_log.committed =
```
BusyJay, line 1288: What if `committed` is larger than `persisted`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1260,9 +1260,58 @@ where
...
+        );
+        self.fsm.peer.force_leader = false;
+        self.fsm
+            .peer
+            .raft_group
+            .raft
+            .become_follower(self.fsm.peer.term(), raft::INVALID_ID);
+        // expire leader lease immediately, because the campaign may choose to be leader at once.
+        // term changes while the old leader lease still holds may cause panic.
+        self.fsm.peer.leader_lease.expire();
```
BusyJay, line 1308: How about not maintaining leader lease in the first place?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1373,6 +1422,14 @@ where
                 self.register_raft_gc_log_tick();
                 self.register_check_leader_lease_tick();
             }
+
+            if self.fsm.peer.force_leader && r == StateRole::Follower {
+                // for some reason, it's not leader anymore
```
BusyJay, line 1427: Like what?

components/raftstore/src/store/peer.rs
```diff
@@ -1684,13 +1692,12 @@ where
                         );
                     }
                 }
-                StateRole::Follower => {
+                StateRole::Follower | StateRole::Candidate | StateRole::PreCandidate => {
```
BusyJay, line 1695: Why?

components/raftstore/src/store/fsm/peer.rs
Connor1996: it's never enough. And considering we use prevote, increasing term by 2 avoids most of the accident.

components/raftstore/src/store/peer.rs
Connor1996: For the exit force leader case, it becomes a follower, and then `campaign()` makes it becomes a pre-candidate, so it needs to clean some states in candidate/pre-candidate state as well.

components/raftstore/src/store/fsm/peer.rs
Connor1996: Don't know, it's unexpected.

components/raftstore/src/store/fsm/peer.rs
Connor1996: If there is a restart, how are some of the committed logs recovered as they aren't persisted?

components/raftstore/src/store/fsm/peer.rs
Connor1996: you mean do not renew leader lease in force leader state?

BusyJay: Yes.

components/raftstore/src/store/fsm/peer.rs
BusyJay: How about making it campaign instead of becoming leader directly? If PD chooses a node with latest logs, it should be able to collect all remaining votes. If all remaining votes are collected, we force it as leader.

components/raftstore/src/store/fsm/peer.rs
NingLin-P: It is possible in the race condition that after the majority of the alive peer have removed all down peers and before the force leader received `SignificantMsg::ExitForceLeaderState` (i.e the force leader apply log slower than other peers or the force leader is isolated from PD), a new leader is elected and cause the force leader to step down.


 `ExitForceLeaderState` message is triggered by PD. After all peers on down stores are removed and before the message is received, there may have a new leader and cause this panic.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4089,6 +4146,15 @@ where
         let leader_id = self.fsm.peer.leader_id();
         let request = msg.get_requests();
 
+        if self.fsm.peer.force_leader {
+            // in force leader state, forbid requests to make the recovery progress less error-prone
+            if !(msg.has_admin_request()
+                && msg.get_admin_request().get_cmd_type() == AdminCmdType::ChangePeer)
```
NingLin-P, line 4152: Should also check `AdminCmdType::ChangePeerV2` if using joint consensus

components/raftstore/src/store/fsm/peer.rs
NingLin-P: Like network problem or raftstore stuck

Connor1996: seems better to exit force leader here, and let's PD retry later

components/raftstore/src/store/peer.rs
```diff
@@ -2516,6 +2523,10 @@ where
         self.report_commit_log_duration(pre_commit_index, &ctx.raft_metrics);
 
         let persist_index = self.raft_group.raft.raft_log.persisted;
+        if self.force_leader {
+            // forward commit index
+            self.raft_group.raft.raft_log.committed = persist_index;
```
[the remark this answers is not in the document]
Connor1996: seems raft base tick is enough, just to make sure it's not in hibernate state.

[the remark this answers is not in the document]
NingLin-P: > make sure it's not in hibernate state.

Or send a `PeerMsg::Noop` for simplicity

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1260,9 +1260,58 @@ where
...
+        self.fsm.peer.raft_group.raft.set_check_quorum(false);
+
+        // forward commit index
+        self.fsm.peer.raft_group.raft.raft_log.committed =
+            self.fsm.peer.raft_group.raft.raft_log.persisted;
+
+        self.fsm.has_ready = true;
+    }
+
+    fn on_exit_force_leader(&mut self) {
```
NingLin-P, line 1294: How is PD detect when to exit the force leader state? if PD relies on the region heatbeat it may possible that the force leader append, commit and apply these `confchange` logs, and send the region heatbeat while other peers may not even have received these `confchange` logs.

Connor1996: PD dispatches the recovery plan, and once TiKV finished applying the conf-changes, TiKV will let these regions exit force leader state.

NingLin-P: What I mean is when the leader applies these `confchange` logs and exit force leader state, other peers may not even have received these `confchange` logs hence still can't elect a new leader.

Connor1996: As the conf-change log is applied on the force leader, so after exit force leader state, this peer can elect as a new leader successfully.

NingLin-P: Oh, I got it

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
Connor1996: `PeerMsg::Noop` doesn't trigger a ready, and it still can't be called.

components/raftstore/src/store/fsm/peer.rs
Connor1996: done

components/raftstore/src/store/peer.rs
```diff
@@ -2516,6 +2523,10 @@ where
         self.report_commit_log_duration(pre_commit_index, &ctx.raft_metrics);
 
         let persist_index = self.raft_group.raft.raft_log.persisted;
+        if self.force_leader {
```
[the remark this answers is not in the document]
Connor1996: done

Connor1996:
Comments are addressed, PTAL @BusyJay @NingLin-P @v01dstar

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,7 +1270,131 @@ where
...
+            .fsm
+            .peer
+            .raft_group
+            .raft
+            .prs()
+            .conf()
+            .voters()
+            .ids()
+            .iter()
+            .filter(|v| !failed_stores.contains(v))
```
BusyJay, line 1295: Probably wrong to comparing store id to peer id.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,7 +1270,131 @@ where
...
+            self.fsm.peer.leader_lease.expire();
+            self.fsm
+                .peer
+                .raft_group
+                .raft
+                .become_follower(self.fsm.peer.term(), raft::INVALID_ID);
+        }
+
+        // become candidate first to increase term
+        self.fsm.peer.raft_group.raft.become_candidate();
```
BusyJay, line 1315: How about sending itself `MsgHup`?

components/raftstore/src/store/fsm/peer.rs
Connor1996: oh, my fault. the store id is the same as peer id, so the tests can pass.

components/raftstore/src/store/fsm/peer.rs
Connor1996: actually, `campaign()` does this thing. Here calling `become_candidate` is to increase the term which would be easy for us to notice the change in logs if we are investigating a problem.

components/raftstore/src/store/fsm/peer.rs
Connor1996: fixed

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1953,6 +2098,73 @@ where
         self.fsm.hibernate_state.count_vote(from.get_id());
     }
 
+    #[inline]
+    fn check_force_leader(&mut self) {
+        if let Some(ForceLeaderState::PreForceLeader {
```
v01dstar, line 2103: nit;
From code readability perspective, I think this part of code can be improved if there is less nested parentheses (especially when the code inside the block is large), and make the whole procedure more clearly by moving the “core” logic (`on_enter_force_leader()`)  to the end of this function. Something like:

```
if !self.fsm.peer.force_leader.is_some() {
    return;
}
let Some(term, expected..) = ..;
if ... {
    warn!("");
    return;
} else if .. {
    warn!("");
    return;
} else {
    ...
}
self.on_enter_force_leader();
```

I know that you were trying to make the code efficient and reuse the code as much as possible. But, it might not be very friendly to a future maintainer who has no context about this feature.

I will leave this for you to decide though. Rest LGTM

Connor1996: change it, PTAL again

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,7 +1270,139 @@ where
...
             }
+            SignificantMsg::EnterForceLeaderState { failed_stores } => {
+                self.on_enter_pre_force_leader(failed_stores)
+            }
+            SignificantMsg::ExitForceLeaderState => self.on_exit_force_leader(),
+        }
+    }
+
+    fn on_enter_pre_force_leader(&mut self, failed_stores: HashSet<u64>) {
+        if self.fsm.peer.force_leader.is_some() {
```
BusyJay, line 1281: Should check whether `expected_alive_voter` changes before ignoring.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,7 +1270,139 @@ where
...
+            return;
+        }
+
+        let region = self.region();
+        let expected_alive_voter = self
```
BusyJay, line 1286: So, why not use `self.region().get_peers().iter().filter()`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,7 +1270,139 @@ where
...
+        // become candidate first to increase term
+        self.fsm.peer.raft_group.raft.become_candidate();
+
+        // trigger vote request to all voters, will check the vote result in `check_force_leader`
+        self.fsm.peer.raft_group.campaign().unwrap();
+        assert!(
+            self.fsm.peer.raft_group.raft.state == StateRole::PreCandidate
+                || self.fsm.peer.raft_group.raft.state == StateRole::Candidate
+        );
+        if !self
```
BusyJay, line 1331: How can it be possible?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,7 +1270,139 @@ where
...
+            return;
+        }
+
+        info!(
+            "exit force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+        );
+        self.fsm.peer.force_leader = None;
+        self.fsm
```
BusyJay, line 1394: May need to register tick again.

components/raftstore/src/store/fsm/peer.rs
Connor1996: when there is a pending configuration change

components/raftstore/src/store/fsm/peer.rs
Connor1996: need to decide which are voters and which aren't, better leave raft to do this

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1382,6 +1515,18 @@ where
                 self.register_raft_gc_log_tick();
                 self.register_check_leader_lease_tick();
             }
+
+            if let Some(ForceLeaderState::ForceLeader) = self.fsm.peer.force_leader {
+                if r != StateRole::Leader {
+                    // for some reason, it's not leader anymore
+                    info!(
```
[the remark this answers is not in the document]
Connor1996: Don't understand, can you elaborate on this?

components/raftstore/src/store/fsm/peer.rs
Connor1996: register which tick?

components/raftstore/src/store/fsm/peer.rs
BusyJay: It's possible. For example, the connection is lost when PD asks it to exit force_leader mode.

components/raftstore/src/store/fsm/peer.rs
BusyJay: They should match each other exactly. Using `region` seems more readable.

components/raftstore/src/store/fsm/peer.rs
BusyJay: raft base tick.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
BusyJay: Oh, what I'm asking is how to retry if it fails to collect enough votes in an election timeout.

components/raftstore/src/store/fsm/peer.rs
BusyJay: How can it be possible giving that wait apply is called before force leader?

Connor1996: just defensive programming, even though it's impossible in normal process

BusyJay:
`wait_apply` may be wrong if the peer is handling snapshot. Apply index will be updated immediately but it has actually not apply to the current index yet. And it should not become leader.

Connor1996:
> `wait_apply` may be wrong if the peer is handling snapshot. Apply index will be updated immediately but it has actually not apply to the current index yet. And it should not become leader.

Where the apply index is updated? And we check the apply index in `on_apply_res` which shouldn't be called on snapshot process.

BusyJay:
It's updated at

https://github.com/tikv/tikv/blob/b5559844ce26a453900ca64c86902647dc807e28/components/raftstore/src/store/peer_storage.rs#L1498.

Connor1996:
> It's updated at
> 
> https://github.com/tikv/tikv/blob/b5559844ce26a453900ca64c86902647dc807e28/components/raftstore/src/store/peer_storage.rs#L1498
> 
> .

Emmm... It does have a problem. /cc @v01dstar Please file another PR to cover the snapshot case.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
BusyJay: In practice, it's quite normal to complete campaign in 2 ~ 3 rounds. The failure rate may be high.

components/raftstore/src/store/peer.rs
```diff
@@ -3772,6 +3812,9 @@ where
         poll_ctx: &mut PollContext<EK, ER, T>,
         mut req: RaftCmdRequest,
     ) -> Result<Either<u64, u64>> {
+        // Should not propose normal in force leader state.
+        assert!(self.force_leader.is_none());
```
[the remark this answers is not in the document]
BusyJay: What I mean is when it's panic, instead of seeing an `assert`, it's better to also log which region and peer goes wrong with what state.

components/raftstore/src/store/fsm/peer.rs
Connor1996: yes, they are the same. But I want to reuse the logic of raft to decide the voters considering the joint consensus.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
Connor1996: Okay. It will issue a new round of request-vote after the election timeout. And `check_force_leader` will pass when receiving enough votes. Will add a test for it.

Connor1996:
Comments are addressed and add more test cases, PTAL @BusyJay

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,224 @@ where
...
+                "region_id" => self.fsm.region_id(),
+                "peer_id" => self.fsm.peer_id(),
+            );
+            return;
+        }
+
+        // Do not use prevote as prevote won't set `vote` to itself.
+        // When PD issues force leader on two different peer, it may cause
+        // two force leader in same term.
+        self.fsm.peer.raft_group.raft.pre_vote = false;
```
BusyJay, line 1332: How about using `MsgTimeoutNow`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,224 @@ where
...
+        // trigger vote request to all voters, will check the vote result in `check_force_leader`
+        self.fsm.peer.raft_group.campaign().unwrap();
+        assert!(
+            self.fsm.peer.raft_group.raft.state == StateRole::PreCandidate
```
BusyJay, line 1336: Stale assert.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,224 @@ where
...
+    fn on_enter_force_leader(&mut self) {
+        info!(
+            "enter force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+        );
+        assert!(!self.fsm.peer.is_leader());
+
+        // become candidate first to increase term
+        self.fsm.peer.raft_group.raft.become_candidate();
```
BusyJay, line 1383: Now that `prevote` is not used, it should be candidate already.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,224 @@ where
...
+        self.fsm.peer.raft_group.raft.become_leader();
```
BusyJay, line 1384: Instead of calling `become_leader` explicitly, you can cast fake vote to it, and let it become leader internally. So all mechanism like commit by vote, bcastappend right after winning election can be reused.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,224 @@ where
...
+            "exit force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+        );
+        self.fsm.peer.force_leader = None;
+        // make sure it's not hibernated
+        self.reset_raft_tick(GroupState::Ordered);
+        // leader lease shouldn't be renewed in force leader state.
+        assert!(self.fsm.peer.leader_lease.is_expired());
+        self.fsm
```
BusyJay, line 1419: How about `MsgHup`?

components/raftstore/src/store/peer.rs
BusyJay: But there is no lease in force leader state.

components/raftstore/src/store/util.rs
```diff
@@ -502,6 +502,10 @@ impl Lease {
         matches!(self.bound, Some(Either::Left(_)))
     }
 
+    pub fn is_expired(&self) -> bool {
```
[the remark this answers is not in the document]
Connor1996: To avoid calling `monotonic_raw_now` in `inspect`, but the frequency is low, will change

components/raftstore/src/store/fsm/peer.rs
Connor1996: can't, learner ignores `MsgTimeoutNow`

components/raftstore/src/store/util.rs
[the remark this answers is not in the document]
BusyJay: Using `inspect(None)` will not request time.

components/raftstore/src/store/fsm/peer.rs
Connor1996: why? `campaign` after this line is to trigger `MsgHup`, here we need to step it down to follower

components/raftstore/src/store/util.rs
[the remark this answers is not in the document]
Connor1996: `None` causes using `monotonic_raw_now` to get a timestamp

components/raftstore/src/store/peer.rs
```diff
@@ -1599,6 +1620,34 @@ where
...
+                .unwrap()
+                .get_store_id();
+            if !expected_alive_voter.contains(&store_id) {
+                continue;
+            }
+            if replicated_idx > p.matched {
+                replicated_idx = p.matched;
+            }
+        }
+        self.raft_group.raft.raft_log.committed =
```
BusyJay, line 1647: I suggest to check if the `replicated_idx` is at current term before forcing advance the commit index.

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,535 @@
...
+    cluster.run();
+    cluster.must_put(b"k1", b"v1");
+
+    let region = cluster.get_region(b"k1");
+    cluster.must_split(&region, b"k9");
+    let region = cluster.get_region(b"k2");
+    let peer_on_store5 = find_peer(&region, 5).unwrap();
+    cluster.must_transfer_leader(region.get_id(), peer_on_store5.clone());
+
+    // try to enter force leader, it can't succeed due to majority isn't lost
```
BusyJay, line 310: Where is it checked?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,255 @@ where
...
+            .peer
+            .raft_group
+            .raft
+            .become_follower(self.fsm.peer.term() + 1, raft::INVALID_ID);
+
+        self.fsm.peer.raft_group.raft.set_check_quorum(true);
+        self.fsm.peer.raft_group.raft.pre_vote = true;
+        // let it trigger election immediately.
+        let _ = self.fsm.peer.raft_group.campaign();
```
BusyJay, line 1459: What if it's still learner after being recovered?

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,535 @@
...
+        .must_remove_peer(region.get_id(), find_peer(&region, 5).unwrap().clone());
+    cluster.exit_force_leader(region.get_id(), 1);
+
+    // majority is formed, can propose command successfully now
+    cluster.must_put(b"k4", b"v4");
+    assert_eq!(cluster.must_get(b"k4"), Some(b"v4".to_vec()));
+}
+
+#[test]
+fn test_force_leader_multiple_election_rounds() {
```
BusyJay, line 491: Better add comment to tests to explain its goal and expected behavior.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1269,6 +1270,255 @@ where
...
+        );
+        assert_eq!(self.fsm.peer.get_role(), StateRole::Candidate);
+
+        if let Some(ForceLeaderState::PreForceLeader {
+            expected_alive_voter,
+        }) = self.fsm.peer.force_leader.take()
+        {
+            let peer_ids: Vec<_> = self
+                .fsm
+                .peer
```
BusyJay, line 1385: This is unreadable.

components/raftstore/src/store/fsm/peer.rs
NingLin-P: Even promote the learner to voter or allow learner become leader, if the learner happens in a tiflash node, it may be required to generate snapshot in SST format, not sure if it is supported currently

tests/integrations/raftstore/test_unsafe_recovery.rs
Connor1996: `check_force_leader` will fail and exit pre force leader

components/raftstore/src/store/fsm/peer.rs
Connor1996: @BusyJay PD ensures it will be promoted as a voter, though better check `promotable` here. If not, do not campaign. will change

Connor1996: @NingLin-P It's not supported yet. For the tiflash peer which has the latest log, PD will trigger remove the learner locally and not let it be the leader. I'm supplementing the issues about merge and tiflash in design doc, and let's discuss it then.

tests/integrations/raftstore/test_unsafe_recovery.rs
BusyJay: But the peer will start campaign any way. It can break correctness as we assume it should not campaign if there is still leader.

Connor1996: note that `promotable` is still false after being force leader, so it will not start campaign.

BusyJay: How can it be related to `promotable`?

components/raftstore/src/store/peer.rs
```diff
@@ -2489,6 +2538,11 @@ where
 
             let persist_index = self.raft_group.raft.raft_log.persisted;
             self.mut_store().update_cache_persisted(persist_index);
+
+            if let Some(ForceLeaderState::ForceLeader { .. }) = self.force_leader {
+                // forward commit index, the committed entries will be applied in the next raft base tick round
+                self.force_forward_commit_index();
```
NingLin-P, line 2544: `force_forward_commit_index` only forward commit index if all alive peers have also persisted the logs, if the leader failed to forward, then it may not get a second chance to retry `force_forward_commit_index` here

tests/integrations/raftstore/test_unsafe_recovery.rs
Connor1996: sorry, misread with the below comment. forget about it.

`It can break correctness as we assume it should not campaign if there is still leader.`
Yes, we don't use `prevote` now, so it can't avoid interrupting other peers if there is a leader.

components/raftstore/src/store/peer.rs
Connor1996: you're right, should check it in base tick too.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1276,6 +1277,247 @@ where
...
+            })
+            .collect();
+        info!(
+            "enter pre force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+            "alive_voter" => ?expected_alive_voter,
+        );
+
+        if self.fsm.peer.is_leader() {
```
BusyJay, line 1322: How about moving it to L1299?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1276,6 +1277,247 @@ where
...
+
+    fn on_enter_force_leader(&mut self) {
+        info!(
+            "enter force leader state";
+            "region_id" => self.fsm.region_id(),
+            "peer_id" => self.fsm.peer_id(),
+        );
+        assert_eq!(self.fsm.peer.get_role(), StateRole::Candidate);
+
+        if let Some(ForceLeaderState::PreForceLeader {
```
BusyJay, line 1381:
```
let expected_alive_voter = match xxx {
    .. => expected_alive_voter,
   ... unreachable!()
};
...
```
is more readable.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1276,6 +1277,247 @@ where
...
+        // leader lease shouldn't be renewed in force leader state.
+        assert_eq!(
+            self.fsm.peer.leader_lease.inspect(None),
+            LeaseState::Expired
+        );
+        self.fsm
+            .peer
+            .raft_group
+            .raft
+            .become_follower(self.fsm.peer.term() + 1, raft::INVALID_ID);
```
BusyJay, line 1446: Unnecessary to increase term.

components/raftstore/src/store/peer.rs
```diff
@@ -478,7 +484,7 @@ where
 
     proposals: ProposalQueue<EK::Snapshot>,
     leader_missing_time: Option<Instant>,
-    leader_lease: Lease,
+    pub leader_lease: Lease,
```
BusyJay, line 487: How about using an access method?

components/raftstore/src/store/peer.rs
```diff
@@ -1608,6 +1629,40 @@ where
...
+                .get_store_id();
+            if !expected_alive_voter.contains(&store_id) {
+                continue;
+            }
+            if replicated_idx > p.matched {
+                replicated_idx = p.matched;
+            }
+        }
+
+        if self.raft_group.store().term(replicated_idx).unwrap() < self.term() {
```
BusyJay, line 1657: Should use `map_or` in case replicated_idx is 0 or out of bound. A case is missing.

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    // quorum is formed, can propose command successfully now
+    cluster.must_put(b"k4", b"v4");
+    assert_eq!(cluster.must_get(b"k2"), None);
+    assert_eq!(cluster.must_get(b"k3"), None);
+    assert_eq!(cluster.must_get(b"k4"), Some(b"v4".to_vec()));
+}
+
+// Test the case that three of five nodes fail and force leader on the rest node
+// which is a learner.
```
BusyJay, line 228: How about joint state?

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    // quorum is formed, can propose command successfully now
+    cluster.must_put(b"k4", b"v4");
+    assert_eq!(cluster.must_get(b"k2"), None);
+    assert_eq!(cluster.must_get(b"k3"), None);
+    assert_eq!(cluster.must_get(b"k4"), Some(b"v4".to_vec()));
+    cluster.must_transfer_leader(region.get_id(), find_peer(&region, 1).unwrap().clone());
+}
+
+// Test the case that none of five nodes fails and force leader on one of the nodes.
```
BusyJay, line 292: The case may provide the wrong impression that force_leader will not break correctness if triggered accidentally. Should comment its limitation.

BusyJay (commented):
Rest LGTM

components/raftstore/src/store/peer.rs
Connor1996: Other fields are public, I think no bother to add an access method for it only.

tests/integrations/raftstore/test_unsafe_recovery.rs
Connor1996: It doesn't break the correctness. Pre-force-leader check probits it to become force leader.

tests/integrations/raftstore/test_unsafe_recovery.rs
Connor1996: The case for joint consesus would be elaborated in this PR https://github.com/tikv/tikv/pull/12022

components/raftstore/src/store/peer.rs
BusyJay: If it's just for read access, it's still necessary. For example, I have found it very convenient multiple times that knowing `proposals` and `pending_reads` are private so I only need to check the single file for potential pitfalls.

tests/integrations/raftstore/test_unsafe_recovery.rs
BusyJay: How? For example, a group of a, b and c. c is isolated from a, a is the leader. If c has increased its term by 2 somehow (for example false prevote success two time) and force leader is sent to b, then b will reject a's heartbeat while can vote for c. So c becomes leader and there are two leaders in the group.

Connor1996: but b won't become leader because it receives vote from an unexpected peer.

NingLin-P (approved):

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,651 @@
...
```
BusyJay, line 390: Still missing the comment that it may not cover all misuse cases.

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -0,0 +1,651 @@
...
+    cluster.wait_log_truncated(region.get_id(), 1, state.get_index() + 40);
+    let state = cluster.truncated_state(region.get_id(), 1);
+
+    cluster.stop_node(3);
+    cluster.stop_node(4);
+    cluster.stop_node(5);
+
+    let state = cluster.truncated_state(region.get_id(), 2);
+
+    cluster.stop_node(1);
```
BusyJay, line 342: Why stop it?

Connor1996: used for the debugging, verbose now.

components/raftstore/src/store/peer.rs
```diff
@@ -1638,23 +1641,19 @@ where
...
-            if !expected_alive_voter.contains(&store_id) {
+            if !expected_alive_voter.contains(peer_id) {
                 continue;
             }
+            count += 1;
             if replicated_idx > p.matched {
                 replicated_idx = p.matched;
             }
         }
+        assert_eq!(count, expected_alive_voter.len());
```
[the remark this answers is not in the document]
Connor1996: Though if PD fails to wait for apply, it indeed has a potential assert fail. Will change it to exit and print a log instead of assertion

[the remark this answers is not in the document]
BusyJay: What I mean is it executes a plan sent by PD and changes its configuration; or have a pending configuration change in uncommitted logs before becoming leader.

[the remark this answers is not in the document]
Connor1996: 1. The plan only removes the failed peers, the assertion keeps true.
2. It's a problem. Seems we should record the failed store id instead of alive peers to handle the case conf change is applied during that.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1276,6 +1277,285 @@ where
...
+                !failed_stores.contains(&store_id)
+            })
+            .collect();
+        if self
+            .fsm
+            .peer
+            .raft_group
+            .raft
+            .prs()
+            .has_quorum(&expected_alive_voter)
```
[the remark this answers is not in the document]
Connor1996: Emmm.... In this case, force leader can't be performed as we add pre-force-leader check expecting all votes from expected alive voters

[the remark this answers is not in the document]
BusyJay: Indeed. Perhaps what we should do is check if there is rejection instead of expecting all approvals.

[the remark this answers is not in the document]
Connor1996: add an TODO item in https://github.com/tikv/tikv/issues/10483

BusyJay (approved):

PR #11942, opened 2022-02-09
metrics: load base split event with same period

### LINKED ISSUE #11941: metrics: load base split event with different period
## Bug Report

load_fit run with per second

no_fit_key or prepare_to_split run with per ten seconds

no_balance_key or no_uncross_key run with per sample key
image

lhy1024:

Issue Number: Close #11941

### Release note 

```release-note
metrics: load base split event with same period
```

### DIFF
--- a/components/raftstore/src/store/worker/split_controller.rs
+++ b/components/raftstore/src/store/worker/split_controller.rs
@@ -256,6 +256,9 @@ impl Recorder {
             if sample.key.is_empty() {
                 continue;
             }
+            LOAD_BASE_SPLIT_EVENT
+                .with_label_values(&["sample_key"])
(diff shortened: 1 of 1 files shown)


nolouch (commented):
need test？could you show the different between before and after?

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -436,8 +442,6 @@ impl AutoSplitController {
                 continue;
             }
 
-            LOAD_BASE_SPLIT_EVENT.with_label_values(&["load_fit"]).inc();
```
rleungx, line 439: Is it expected that `load_fit` is removed?

PR #11954, opened 2022-02-10
rust: update toolchain to fix missing rls

### LINKED ISSUE #11953: Rls is missing

master

No related

### Steps to reproduce

```
$ rustup component add rls                                                                
error: component 'rls' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel 'nightly-2022-01-07'
Sometimes not all components are available in any given nightly. If you don't need the component, you can remove it with:
```

### What did you expect?
Have a rls component

Connor1996:

Issue Number: Close #11953 

What's Changed:

Update to rust nightly 2022-02-14 which is presented with rls, see https://rust-lang.github.io/rustup-components-history/

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```


tabokie (approved):

Cargo.lock
```diff
@@ -6330,9 +6320,9 @@ dependencies = [
 
 [[package]]
 name = "x86"
-version = "0.37.0"
+version = "0.45.0"
```
BusyJay, line 6323: Why change this?

Connor1996: `llvm_asm!` is replaced with `asm!` in recent nightly, so update the dependency to fix compile issue.

5kbpers:
Please fix the lint errors.

cmd/tikv-ctl/src/util.rs
```diff
@@ -10,6 +10,7 @@ use tikv::config::TiKvConfig;
 
 const LOG_DIR: &str = "./ctl-engine-info-log";
 
+#[allow(clippy::field_reassign_with_default)]
```
tabokie, line 13: This should be fixed.

Connor1996: the original way is cleaner, and we also need to change some subfield of the field, so it needs be written in reassign anyway

Connor1996:
PTAL

components/txn_types/src/timestamp.rs
```diff
@@ -144,12 +145,14 @@ impl TsSet {
 
     pub fn from_u64s(ts: Vec<u64>) -> Self {
         // This conversion is safe because TimeStamp is a transparent wrapper over u64.
+        #[allow(clippy::unsound_collection_transmute)]
```
5kbpers, line 148: This is undefined behavior. Better to fix this through iteration like `ts.into_iter().map(TimeStamp::from).collect()`

components/txn_types/src/timestamp.rs
```diff
@@ -144,12 +145,14 @@ impl TsSet {
...
         let ts = unsafe { ::std::mem::transmute::<Vec<u64>, Vec<TimeStamp>>(ts) };
         Self::new(ts)
     }
 
     pub fn vec_from_u64s(ts: Vec<u64>) -> Self {
         // This conversion is safe because TimeStamp is a transparent wrapper over u64.
+        #[allow(clippy::unsound_collection_transmute)]
```
5kbpers, line 155: (ditto)

components/concurrency_manager/src/lib.rs
```diff
@@ -79,6 +79,7 @@ impl ConcurrencyManager {
             result[index] = MaybeUninit::new(self.lock_table.lock_key(key).await);
         }
         #[allow(clippy::unsound_collection_transmute)]
+        #[allow(clippy::transmute_undefined_repr)]
```
5kbpers, line 82: It seems also to be a UB

sticnarf: I don't understand. Why is it a UB to transmute from `Vec<MaybeUninit<T>>` to `Vec<T>`?

components/txn_types/src/timestamp.rs
sticnarf: I think it will be 100% safe if we add `#[repr(transparent)]` to `TimeStamp`.

Connor1996: But `TimeStamp` is just a wrapper of u64. Seems these warnings are only due to clippy's check is more strict

components/concurrency_manager/src/lib.rs
Connor1996: don't understand. Not sure if it's due to the clippy's check is more strict or the compile really changes the memory layout.

5kbpers: Ok I read the type of `result` wrong

components/txn_types/src/timestamp.rs
Connor1996: @sticnarf Tried it before, `repr(C)` eliminates `transmute_undefined_repr`, whereas `repr(transparent)` eliminates `unsound_collection_transmute`. There is always a clippy error anyway. 🤣

Connor1996: Though `repr(transparent)` is more appropriate, will change it from `repr(C)`

components/concurrency_manager/src/lib.rs
sticnarf: Ah, I know why it is UB. It is indeed possible that `Wrapper<MaybeUninit<T>>` may be different from `Wrapper<T>`. https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#layout has an example of `Option`.

Actually this function `lock_keys` is only used in tests and I don't believe `Vec` can have similar behavior as `Option` XD

Connor1996: got it

sticnarf (approved):
LGTM

Connor1996:
PTAL @5kbpers

sticnarf:
I see `test_read_index_retry_lock_checking` fails because of failing to transfer leader. But I can always pass the test at my dev env...

Looks strange...

PR #11974, opened 2022-02-16
raftstore: support dynamic adjust max_batch_size 

### LINKED ISSUE #11982: online config: support adjusting apply/store-max-batch-size dynamically

A dynamic config modification is performed from TiKV-Control command line (documentation). So expect to adjust the `max-batch-size` for both apply and store batch system.
```
set config tikv `raftstore.store-max-batch-size`=512;
set config tikv `raftstore.apply-max-batch-size`=512;
```

glorv:

Issue Number: Close #11982

What's Changed:

- In the current implementation, Poller will call `hander.start` to do some preprocess. In both apply/store handler, it will check and apply raftstore's online config change, so we can let Poller register a callback here, when config version changes, handle calls this callback to let batch-system adjust its relation configs.
- Adjust the `max-batch-size` in `PoolState` at `RaftstoreConfigManager` so it can use the newest value to create new poller thread when adjust pool-size.

### Related changes

### Check List 

Tests 

- Unit test
  - Running sysbench(oltp_read_write) on a 3 tikv cluster with 50 threads, and use a bash script to update one of these two config every 5s with a random value from [64, 2048] for 10h. There is no significant qps volatility and no error in each tikv log.
- No code

Side effects

### Release note 

```release-note
raftstore: Support adjusting apply_max_batch_size and store_max_batch_size dynamically 
```


components/raftstore/src/store/config.rs
```diff
@@ -532,13 +533,16 @@ impl Config {
...
-        if let Some(size) = self.apply_batch_system.max_batch_size {
-            if size == 0 {
-                return Err(box_err!("apply-max-batch-size should be greater than 0"));
-            }
-        } else {
-            self.apply_batch_system.max_batch_size = Some(256);
+
+        if self.apply_batch_system.max_batch_size == 0 {
+            self.apply_batch_system.max_batch_size = 256;
+        }
```
NingLin-P, line 539: Actually, when the online config is first implemented, the `TiKvConfig::validate` function is used for validation only not for adjusting, so the configs that adjusted in `TiKvConfig::validate` won't be actually used. 

Changing
https://github.com/tikv/tikv/blob/8b86c063c21f111a2fb0d4fcc4a328c3df7bd6f3/src/config.rs#L3480-L3484
to
```rust
let diff = {
        let mut incoming = self.get_current();
        incoming.update(diff);
        // Config may be adjusted in `validate`
        incoming.validate()?;
        self.get_current().diff(&incoming)
};
```
may work.

NingLin-P:
Although the effect of changing `max_batch_size` is hard to measure, I think we still need uint tests to ensure `max_batch_size` indeed can be changed dynamically (see `test_update_raftstore_config` for example).

components/raftstore/src/store/config.rs
glorv: done.

glorv:
> Although the effect of changing `max_batch_size` is hard to measure, I think we still need uint tests to ensure `max_batch_size` indeed can be changed dynamically (see `test_update_raftstore_config` for example).

Seems it's not easy to verify that the config takes effect in `Poller` and `PoolState`, so I just add the new config fields in `test_update_raftstore_config`

src/config.rs
```diff
@@ -3474,14 +3465,17 @@ impl ConfigController {
...
         change: Option<HashMap<String, String>>,
     ) -> CfgResult<()> {
-        {
+        diff = {
             let mut incoming = self.get_current();
-            incoming.update(diff.clone());
-            incoming.validate()?;
-        }
+            let mut updated = incoming.clone();
+            updated.update(diff.clone());
```
NingLin-P, line 3474: Unnecessary cloning

tests/integrations/config/dynamic/raftstore.rs
```diff
@@ -154,6 +154,14 @@ fn test_update_raftstore_config() {
             "raftstore.raft-log-gc-threshold".to_owned(),
             "54321".to_owned(),
         );
+        m.insert(
+            "raftstore.apply-max-batch-size".to_owned(),
+            "1234".to_owned(),
```
NingLin-P, line 159: Plz add more cases like `0`, `10241`

glorv:
@NingLin-P @5kbpers  PTAL again, thanks~

components/batch-system/src/config.rs
```diff
@@ -8,8 +8,7 @@ use tikv_util::config::ReadableDuration;
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct Config {
-    #[online_config(skip)]
-    pub max_batch_size: Option<usize>,
+    pub max_batch_size: usize,
```
NingLin-P, line 11: This change may not be needed after https://github.com/tikv/tikv/pull/12133 is merged

NingLin-P (commented):
Rest LGTM

NingLin-P (approved):

components/raftstore/src/store/worker/refresh_config.rs
```diff
@@ -103,26 +103,39 @@ where
...
+pub enum BatchComponent {
     Store,
     Apply,
 }
 
+impl Display for BatchComponent {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match *self {
+            BatchComponent::Store => {
+                write!(f, "apply")
```
5kbpers, line 116: apply -> raft

5kbpers (approved):

PR #11978, opened 2022-02-16
diagnosis: support get cpu time of threads for macOS

### LINKED ISSUE #11977: thread cpu time is unavailable for macOS
master

macOS, x86-64

### What did you expect?
CPU information to be available for resource metering

### What did happened?
not supported yet

zhongzc:

Issue Number: Close #11977

What's Changed:

```commit-message
Add the variant for macOS into `components/tikv_util/src/sys/thread.rs`
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Support thread CPU time for macOS
```


components/tikv_util/build.rs
```diff
@@ -0,0 +1,170 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+// Reference: https://github.com/larksuite/perf-monitor-rs/blob/master/build.rs
```
sticnarf, line 3: Is it possible to use `perf-monitor-rs` directly?

components/tikv_util/src/sys/cpu_time.rs
```diff
@@ -126,16 +86,17 @@ mod imp {
             let mut num_cpus_u = 0;
             let mut cpu_info = ptr::null_mut();
             let mut msg_type = 0;
-            let err = host_processor_info(
+            let ret = host_processor_info(
```
BusyJay, line 89: I think you can get these APIs from libc without any build scripts. https://docs.rs/libc/latest/x86_64-apple-darwin/libc/index.html

zhongzc: libc doesn't contain all my needs, e.g. `task_threads`.

components/tikv_util/build.rs
zhongzc: I'd like to manipulate the raw API but `perf-monitor-rs` provides a high abstraction. If using `perf-monitor-rs`, I have to do the reverse of many of the operations it does.

components/tikv_util/src/sys/cpu_time.rs
BusyJay: Any other symbols it misses? Just defining `task_threads` seems a lot simpler and lightweight than using a build script.

zhongzc:
comments addressed. PTAL @BusyJay @sticnarf

components/tikv_util/src/sys/thread.rs
```diff
@@ -181,7 +181,120 @@ mod imp {
     }
 }
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(target_os = "macos")]
+#[allow(bad_style)]
```
BusyJay, line 185: Why allow bad style?

zhongzc: Want to keep the same type names as original APIs. But Clippy suggests that `type task_inspect_t` should be `type TaskInspectT`.

components/tikv_util/src/sys/thread.rs
```diff
@@ -181,7 +181,120 @@ mod imp {
...
+    #[derive(Default)]
+    pub struct FullStat {
+        pub stime: i64,
+        pub utime: i64,
+        pub command: String,
+    }
+
+    #[inline]
+    pub fn clock_tick() -> i64 {
+        1_000_000
```
BusyJay, line 218: Why use a fixed value?

components/tikv_util/src/sys/thread.rs
```diff
@@ -264,10 +381,43 @@ mod tests {
...
+                    while !stop_threads.load(Ordering::Relaxed) {}
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let ids: HashSet<_> = thread_ids(process_id()).unwrap();
+        {
+            let actual_tids = actual_tids.lock().unwrap();
+            for tid in &*actual_tids {
+                assert!(ids.contains(tid));
```
BusyJay, line 406: Race with L395, should use channel instead.

components/tikv_util/src/sys/thread.rs
```diff
@@ -181,7 +181,120 @@ mod imp {
...
+    }
+
+    pub fn full_thread_stat(_pid: Pid, tid: Pid) -> io::Result<FullStat> {
+        // https://www.gnu.org/software/hurd/gnumach-doc/Thread-Information.html
+
+        unsafe {
+            let flavor = THREAD_BASIC_INFO;
+            let mut info = std::mem::zeroed::<thread_basic_info>();
+            let mut thread_info_cnt =
+                (size_of::<thread_basic_info>() / size_of::<natural_t>()) as mach_msg_type_number_t;
```
BusyJay, line 266: You mean `THREAD_BASIC_INFO_COUNT`?

zhongzc: yes, and not found from libc

components/tikv_util/src/sys/thread.rs
zhongzc: updated

components/tikv_util/src/sys/thread.rs
zhongzc: updated

components/tikv_util/src/sys/thread.rs
BusyJay: Interesting, but it does compile when using the constant.

components/tikv_util/src/sys/thread.rs
```diff
@@ -264,10 +385,38 @@ mod tests {
...
+        let (tx, rx) = crossbeam::channel::bounded(THREAD_COUNT);
+        let stop_threads = Arc::new(AtomicBool::new(false));
+
+        let threads = (0..THREAD_COUNT)
+            .map(|_| {
+                let tx = tx.clone();
+                let stop_threads = stop_threads.clone();
+                std::thread::spawn(move || {
+                    tx.send(thread_id()).unwrap();
+                    while !stop_threads.load(Ordering::Relaxed) {}
```
sticnarf, line 400: It doesn't seem a good idea to let 10 threads watch a bool flag in a busy loop. I suggest adding use condvar or at least add some sleep.

zhongzc: updated

components/tikv_util/src/sys/thread.rs
zhongzc: updated

components/tikv_util/src/sys/thread.rs
```diff
@@ -181,7 +181,123 @@ mod imp {
...
+
+    #[derive(Default)]
+    pub struct FullStat {
+        pub stime: i64,
+        pub utime: i64,
+        pub command: String,
+    }
+
+    /// The time unit of `stime` and `utime` is in microseconds.
+    /// See [`full_thread_stat()`]
```
BusyJay, line 219:
```suggestion
    /// Unlike Linux, the unit of `stime` and `utime` is microseconds instead of ticks.
    /// See [`full_thread_stat()`]
```

BusyJay (approved):
Rest LGTM

sticnarf (approved):
LGTM

PR #11980, opened 2022-02-17
copr: support extract scalar value from bit

Yisaer:

Issue Number: close https://github.com/tikv/tikv/issues/11893

What's Changed:

In https://github.com/tikv/tikv/pull/11968 ,  we supported push bit cloumn down but we forgot to support mysqlBit during extract scalar value.

This pr fixes it.

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/tidb_query_expr/src/types/expr_builder.rs
+++ b/components/tidb_query_expr/src/types/expr_builder.rs
@@ -906,4 +906,14 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_extract_scalar_value_uint64_from_bits() {
+        let mut res = extract_scalar_value_uint64_from_bits(vec![0x01, 0x56, 0x12, 0x34]).unwrap();
+        assert_eq!(ScalarValue::Int(Some(0x1561234)), res);
+        res = extract_scalar_value_uint64_from_bits(vec![0x56, 0x34, 0x12, 0x78]).unwrap();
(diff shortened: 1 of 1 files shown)


Yisaer:
/cc @skyzh @zhongzc 

And I don't know how to design unit testcase for this,  could you help me? @skyzh

zhongzc (approved):

components/tidb_query_expr/src/types/expr_builder.rs
```diff
@@ -395,6 +398,16 @@ fn extract_scalar_value_int64(val: Vec<u8>) -> Result<ScalarValue> {
     Ok(ScalarValue::Int(Some(value)))
 }
 
+#[inline]
+fn extract_scalar_value_uint64_from_bits(val: Vec<u8>) -> Result<ScalarValue> {
+    debug_assert!(val.len() <= 8);
+    let mut res = 0;
+    for (i, v) in val.iter().enumerate() {
+        res |= (*v as u64) << ((val.len() - i - 1) * 8);
+    }
```
zhongzc, line 407: If I don't think wrong, could the logic to be:
```rust
let mut res = 0;
for v in val {
    res <<= 8;
    res |= v as u64;
}
```

Yisaer: right, yours is more clear, updated.

longfangsong (approved):

skyzh (approved):

PR #11985, opened 2022-02-18
raftstore: revert pessimistic locks status when failing to propose PrepareMerge

sticnarf:

Issue Number: Ref #11452 

What's Changed:

To fix the problem mentioned in https://github.com/tikv/tikv/pull/11758#discussion_r807606597, this PR adds `post_propose_fail` callback. If it fails to propose `PrepareMerge`, we should change the status of `PeerPessimisticLocks` back to normal.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


tests/integrations/raftstore/test_merge.rs
```diff
@@ -1385,3 +1387,140 @@ fn test_merge_pessimistic_locks_when_gap_is_too_large() {
...
+            .insert(vec![(Key::from_raw(b"k1"), lock.clone())])
+            .is_ok()
+    );
+
+    // Filter MsgAppend, so the proposed PrepareMerge will not succeed
+    cluster.add_send_filter(CloneFilterFactory(
+        RegionPacketFilter::new(left.id, 2)
+            .msg_type(MessageType::MsgAppend)
+            .direction(Direction::Recv),
+    ));
```
NingLin-P, line 1506: Propose is the procedure to generate `MsgAppend` it can still success even can't send the msg to other peers, how about using a fail point to return an error early when proposing to simulate the case.

tests/integrations/raftstore/test_merge.rs
```diff
@@ -1385,3 +1387,140 @@ fn test_merge_pessimistic_locks_when_gap_is_too_large() {
...
+        min_commit_ts: 30.into(),
+    };
+    assert!(
+        txn_ext
+            .pessimistic_locks
+            .write()
+            .insert(vec![(Key::from_raw(b"k1"), lock)])
+            .is_ok()
+    );
+    // Filter MsgTimeoutNow, so transfer leader will not actually succeed.
```
NingLin-P, line 1430: Transfer leader may abort after some tick and accept proposals again which make the case unstable, how about using a fail point instead.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2408,13 +2408,16 @@ where
...
+        // have been proposed. But we still need to return true to propose another TransferLeader
+        // command. Otherwise, some write requests that have marked some locks as deleted will fail
+        // because raft rejects more proposals.
+        // It is OK to return true here if it's in other states like MergingRegion or NotLeader.
+        // In those cases, the locks will fail to propose and nothing will happen.
+        if !pessimistic_locks.is_writable() {
             return true;
         }
-        pessimistic_locks.is_valid = false;
+        pessimistic_locks.status = LocksStatus::TransferingLeader;
```
NingLin-P, line 2420: Should we also revert the `TransferingLeader` status if the transfer is aborted? Abortion can happen in a few places (see `abort_leader_transfer` in `raft-rs`)  but we don't have a unified way to get notified...

tests/integrations/raftstore/test_merge.rs
sticnarf: It doesn't need to really fail. Instead, I want it to succeed, but not until I remove the filters. 

This is to generate the case that a second PrepareMerge is proposed before the previous one succeeds. This is to test against this part:

```rust
            // If we just failed to propose PrepareMerge, the pessimistic locks status
            // may become MergingRegion incorrectly. So, we have to revert it here.
            // But we have to rule out the case when the region has successfully
            // proposed PrepareMerge or has been in merging, which is decided by
            // the boolean expression below.
            let is_merging = self.is_merging()
                || self
                    .cmd_epoch_checker
                    .last_cmd_index(AdminCmdType::PrepareMerge)
                    .is_some();
```

components/raftstore/src/store/fsm/peer.rs
sticnarf: Yes, absolutely. I don't know how to get notified, so I decide o use a tick to do so in https://github.com/tikv/tikv/pull/11883...

tests/integrations/raftstore/test_merge.rs
NingLin-P: I see, but filtering just one peer may not be enough as there are still majority of peers.

sticnarf: I configure rhe simulated cluster to only have 2 nodes.

tests/integrations/raftstore/test_merge.rs
sticnarf: I switch to use a fail point instead. I make a more flexible failpoint, so I can also use it in system tests. PTAL again, thanks very much!

tests/failpoints/cases/test_merge.rs
```diff
@@ -1597,3 +1599,68 @@ fn test_retry_pending_prepare_merge_fail() {
...
+
+    cluster.merge_region(left.id, right.id, Callback::None);
+    thread::sleep(Duration::from_millis(200));
+    assert_eq!(
+        txn_ext.pessimistic_locks.read().status,
+        LocksStatus::MergingRegion
+    );
+
+    // With the fail point set, we will fail to propose the locks or the PrepareMerge request.
+    fail::cfg("raft_propose", "return()").unwrap();
```
NingLin-P, line 1653: Do you mean `return(:)`

NingLin-P (commented):
Rest LGTM

tests/failpoints/cases/test_merge.rs
sticnarf: The ":" should also be omitted if `store_id` is not specified.

sticnarf:
@youjiali1995 PTAL, thanks!

youjiali1995 (approved):

youjiali1995 (approved):

PR #11986, opened 2022-02-18
split_controller: move sample_threshold check out of the split_key iteration

JmPotato:

What's Changed:

```commit-message
split_controller: move sample_threshold check out of the split_key iteration
```

Since the `evaluated_key_num` will always equal to the length of `self.key_ranges` inside a `Recorder`, so we do not need to do this check in every iteration of the `samples.split_key`. This PR moves `sample_threshold` check out of the `split_key` iteration to reduce the branch checking inside a loop.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```


lhy1024:
What happens if a region has less than sample_threshold keys but qps exceeds the threshold？

lhy1024 (commented):
What happens if a region has less than sample_threshold keys but qps exceeds the threshold？

JmPotato:
> What happens if a region has less than sample_threshold keys but qps exceeds the threshold？

I will return an empty split key result as before.

lhy1024 (approved):

rleungx (approved):

disksing (approved):

PR #11991, opened 2022-02-20
cdc: separate resolved region outliers

### LINKED ISSUE #11993: Separate broadcasing outlier regions and normal regions
See more: https://github.com/tikv/tikv/pull/11991

overvenus:

Issue Number: close #11993 close pingcap/tiflow#4516 close pingcap/tiflow#4311 ref pingcap/tiflow#4146

What's Changed:

```commit-message
Separate broadcasing outlier regions and normal regions,
so 1) downstreams know where they should send resolve lock requests,
and 2) resolved ts of normal regions does not fallback.
```

### Related changes

### Check List 

Tests 

- Unit test
  Restart tikv nodes one by one.

### Release note 

```release-note
Reduce CDC recovery time by reduce the number regions that need resolved lock.
```


components/cdc/src/endpoint.rs
```diff
@@ -710,29 +776,75 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
...
-        self.broadcast_resolved_ts(resolved_regions);
+
+        // Separate broadcasing outlier regions and normal regions,
+        // so 1) downstreams know where they should send resolve lock requests,
+        // and 2) resolved ts of normal regions does not fallback.
+        //
+        // Max number of outliers, in most cases, only a few regions are outliers.
+        let max_outlier_count = 32;
+        let (outlier_min_resolved_ts, outlier_regions) =
+            resolved_regions.pop_outliers(max_outlier_count);
```
hicqu, line 838: How about pass in a timestamp into `pop_outliers` instead of a count?

hicqu:
rest LGTM.

components/cdc/src/endpoint.rs
```diff
@@ -710,29 +776,75 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
...
+                    advance_failed_none += 1;
                 }
             }
         }
-        self.resolved_region_count = resolved_regions.len();
+        let lag_millis = min_ts
+            .physical()
+            .saturating_sub(self.min_resolved_ts.physical());
+        if Duration::from_millis(lag_millis) > WARN_RESOLVED_TS_LAG_THRESHOLD {
+            warn!("cdc resolved ts lag too large";
```
5kbpers, line 819: Is it too frequent to output every 1s?

components/cdc/src/endpoint.rs
5kbpers: Or maybe we could name this function `pop` rather than `pop_outliers`.

components/cdc/src/endpoint.rs
```diff
@@ -900,6 +1015,12 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
                     Err(err) => panic!("failed to schedule min ts event, error: {:?}", err),
                 }
             }
+            let lag_millis = min_ts_pd.physical().saturating_sub(min_ts.physical());
```
5kbpers, line 1018: (ditto)

hicqu (approved):

5kbpers (approved):

Tianion:
Hello, we have recently encountered this issue as well. I can understand why TiKV might cause TiCDC to hang, but I am not clear on why TiKV would exhibit this behavior. Could you or someone else help clarify this issue ?

PR #12003, opened 2022-02-22
copr: support rest greatest/least functions

Yisaer:
Signed-off-by: yisaer <disxiaofei@163.com>

Issue Number: Ref https://github.com/tikv/tikv/issues/5751

What's Changed:

Support rest greatest/least functions in order to let tidb push `greatest` and `least` functions down into TiKV

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


skyzh (approved):
LGTM for the implementation. How is this function look like in SQL? Can we apply some techniques in `IN` expression for such functions? e.g., if there are constant cells, we can first calculate their greatest value and store them in query context; and for other columns, we can directly compare them with the greatest among the constants.

longfangsong (approved):

Yisaer:
> LGTM for the implementation. How is this function look like in SQL? Can we apply some techniques in `IN` expression for such functions? e.g., if there are constant cells, we can first calculate their greatest value and store them in query context; and for other columns, we can directly compare them with the greatest among the constants.

I think your suggestion can be applied later to revise such functions.

PR #12013, opened 2022-02-23
split_controller: fix the incorrect sampled key ranges number check

### LINKED ISSUE #12012: Incorrect sampled key ranges number check in the Load Base Split

After merging https://github.com/tikv/tikv/pull/11986. 

### What did you expect?

The number of sampled key ranges that will be evaluated is actually the number of `self.key_ranges` flatten.

https://github.com/tikv/tikv/blob/cbcaab6ce56cdc0a503a5f9171470bff3c0b53da/components/raftstore/src/store/worker/split_controller.rs#L258-L260

### What did happened?

However, the check uses `self.key_ranges.len()` to do the check incorrectly.

https://github.com/tikv/tikv/blob/cbcaab6ce56cdc0a503a5f9171470bff3c0b53da/components/raftstore/src/store/worker/split_controller.rs#L252

JmPotato:

Issue Number: Close #12012.

What's Changed:

```commit-message
split_controller: fix the incorrect sampled key ranges number check
```

### Related changes

- Problem cause: https://github.com/tikv/tikv/pull/11986

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/raftstore/src/store/worker/split_controller.rs
+++ b/components/raftstore/src/store/worker/split_controller.rs
@@ -632,6 +632,25 @@ mod tests {
     }
 
     #[test]
+    fn test_recorder() {
+        let mut config = SplitConfig::default();
+        config.detect_times = 10;
+        config.sample_threshold = 20;
+
+        let mut recorder = Recorder::new(config.detect_times);
+        for _ in 0..config.detect_times {
+            assert!(!recorder.is_ready());
+            recorder.record(vec![
+                build_key_range(b"a", b"b", false),
+                build_key_range(b"b", b"c", false),
+            ]);
+        }
+        assert!(recorder.is_ready());
+        let key = recorder.collect(&config);
+        assert_eq!(key, b"b");
+    }
+
+    #[test]
     fn test_hub() {
         // raw key mode
         let raw_key_ranges = vec![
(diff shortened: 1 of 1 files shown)


lhy1024 (approved):
Do we need to add a test to avoid the similar problem?

disksing (approved):

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -247,15 +247,16 @@ impl Recorder {
             |x| x,
         );
         let mut samples = Samples::from(sampled_key_ranges);
+        let recorded_key_ranges: Vec<&KeyRange> = self.key_ranges.iter().flatten().collect();
         // Because we need to observe the number of `no_enough_key` of all the actual keys,
         // so we do this check after the samples are calculated.
-        if (self.key_ranges.len() as u64) < config.sample_threshold {
+        if (recorded_key_ranges.len() as u64) < config.sample_threshold {
```
rleungx, line 253: How about using `size_hint`?

JmPotato: `size_hint`'s output is a tuple and has an `Option` inside it, I think using `collect` is more simple here.

rleungx: I think maybe you can use `.0` so that we can avoid `collect` and `into_iter`.

JmPotato: Also, after `flatten()`, the result of `size_hint()` will be `(0, None)`.

origin_img_v2_85e192ec-377a-449b-bb6e-51ce6b1730ag

PR #12018, opened 2022-02-24
raftstore: adjust raft max msg size dynamically

### LINKED ISSUE #12017: online config: support adjust raftstore.raft-max-size-per-msg dynamically

A dynamic config modification is performed from TiKV-Control command line (documentation). So expect to adjust the `raft-max-size-per-msg` in some scenario to optimize performance.
```
set config tikv `raftstore.raft-max-size-per-msg`="8MiB";
```

glorv:

Issue Number: Close #12017

What's Changed:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

### Release note 

```release-note
raftstore: Support adjusting the max size of raft message dynamically 
```


NingLin-P:
I think `raftstore.raft-entry-max-size` is not supported to change dynamically. 

When region merge is triggered, before proposing the `PrepareMerge` command, the source region will check whether all the lagging raft logs can be fitted into a single raft entry. And then after `PrepareMerge` is applied, all the lagging raft logs will be fitted into the `CommitMerge` command and proposed by the target region. 

If the source region check passed then `raftstore.raft-entry-max-size` change to a smaller value, it may cause the `CommitMerge` command to fail to propose due to its size is larger than `raftstore.raft-entry-max-size` now, then the source region is stuck in merging state and can't handle any read/write request.

NingLin-P:
Oh, sorry, I had made a mistake, this PR tries to support changing `raft_max_size_per_msg` dynamically not `raft_entry_max_size`, and `raft_entry_max_size` already support changing dynamically...

glorv:
> Plz also add `#[online_config(skip)]` to `raft_entry_max_size` to disable dynamically change of `raft_entry_max_size`

done

NingLin-P (approved):

components/raftstore/src/store/config.rs
```diff
@@ -432,6 +432,14 @@ impl Config {
             ));
         }
 
+        if self.raft_max_size_per_msg.0 == 0
+            || self.raft_max_size_per_msg.0 > ReadableSize::gb(32).0
```
5kbpers, line 436: Maybe it should be limited to 1GiB.

5kbpers (commented):
rest LGTM

components/raftstore/src/store/config.rs
glorv: @BusyJay said there are users that set this to `3gb`.

BusyJay: What I mean is not larger than 3GiB. The value comes from rocksdb's limitation on value size.

glorv: done. Sorry, I misunderstood your meaning.

glorv:
@5kbpers PTAL again, thanks.

5kbpers (approved):

PR #12016, opened 2022-02-24
copr: fix missing flag solvement for bit column in decoding

Yisaer:
### What is changed and how it works?

Issue Number: Close https://github.com/pingcap/tidb/issues/32506

What's Changed:

For `MysqlBit` type column datum, it will carry `uint_flag` or `var_uint_flag` according to 
https://github.com/pingcap/tidb/blob/e42f26c3f8f9bfb3a0cbc903b4204fac13ac4eb6/util/codec/codec.go#L120-L121
in tidb

Also, the we should directly `append_bytes` for datum into column according to https://github.com/pingcap/tidb/blob/e42f26c3f8f9bfb3a0cbc903b4204fac13ac4eb6/util/chunk/chunk.go#L567-L568
in tidb

The previous implementation does not match the tidb, this request fixed it.

```commit-message
```

### Check List 

Tests 

- Unit test
- Manual test (add detailed scripts or steps below)

I manually tested this requests with https://github.com/pingcap/tidb/pull/32497 which enabled push bit column down,  here is the test result:

```sql
mysql> create table t (a int);
er table `t` add column c bit(16) NULL default b'1100110111001';
update t set c = b'11100000000111';
update t set c = 'AB';
select * from t;
select * from t where concat(c,'A') = 'ABA';
explain select * from t where concat(c,'A') = 'ABA';Query OK, 0 rows affected (0.09 sec)

mysql> insert into t set a=1;
Query OK, 1 row affected (0.00 sec)

mysql> alter table `t` add column c bit(16) NULL default b'1100110111001';
Query OK, 0 rows affected (0.26 sec)

mysql> update t set c = b'11100000000111';
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> update t set c = 'AB';
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select * from t;
+------+----------+
| a    | c        |
+------+----------+
|    1 |       AB |
+------+----------+
1 row in set (0.00 sec)

mysql> select * from t where concat(c,'A') = 'ABA';
+------+----------+
| a    | c        |
+------+----------+
|    1 |       AB |
+------+----------+
1 row in set (0.00 sec)

mysql> explain select * from t where concat(c,'A') = 'ABA';
+-------------------------+----------+-----------+---------------+-------------------------------------------------------+
| id                      | estRows  | task      | access object | operator info                                         |
[...]

PR #12021, opened 2022-02-24
copr: fix panic during push bit column down

Yisaer:
Signed-off-by: yisaer <disxiaofei@163.com>

Issue Number: ref https://github.com/pingcap/tidb/issues/32506 https://github.com/tikv/tikv/pull/12016

What's Changed:
Bit Column should be solved separately according to its flag, otherwise it will cause panic.

```commit-message
```

### Check List 

Tests 

same in https://github.com/tikv/tikv/pull/12016

### Release note 

```release-note
None
```

### DIFF
--- a/components/tidb_query_datatype/src/codec/chunk/column.rs
+++ b/components/tidb_query_datatype/src/codec/chunk/column.rs
@@ -560,14 +560,16 @@ impl Column {
             ));
         }
         let flag = src_datum[0];
-        let raw_datum = &src_datum[1..];
+        let mut raw_datum = &src_datum[1..];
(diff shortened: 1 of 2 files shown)


longfangsong (approved):

shiyuhang0:
How long will it take to fix this issue. TiSpark meet some problems because of this bug :)

Yisaer:
> How long will it take to fix this issue. TiSpark meet some problems because of this bug :)

I will revert pr which caused bug firstly in https://github.com/tikv/tikv/pull/12030 as the fixing and testing costs much time.

Issue #11965, opened 2022-02-14
Implement RawKV Cross Cluster Replication


pingyu:
## Development Task

Issue Number: #11745 
RFC: https://github.com/tikv/rfcs/pull/86

##### Development:

- [x] API v2: @pingyu #12036
  - [x] New RawKV key codec
- [x] Raw store: @haojinming #12087
- [x] Causal timestamp: @pingyu #12099
  - [x] Timestamp allocation
  - [x] Timestamp management
    - [x] Advance timestamp on leader transfer
  - [x] Server
    - [x] Feature switch & configuration
- [x] Raw interfaces: @pingyu #12343 
  - [x] Reads
  - [x] Writes
  - [x] Deletes @haojinming #12090
  - [x] Atomic
- [x] Garbage collection: @ystaticy 
  - [x] TiKV gc_worker #12424 
  - [x] Standalone GC worker @haojinming https://github.com/tikv/migration/pull/85
- [x] BR for cross API version data migration
  - [x] TiKV br component #12303 
  - [x] TiKV-BR https://github.com/tikv/migration/issues/67
- [x] Others
  - [x] Bulk load #12389 
  - [x] Metrics #12494 

##### Testing: 

- [x] Test
  - [x] Integration Test for causality validation
    - [x] Leader Transfer
    - [x] Region split + leader transfer
    - [x] Region merge + leader transfer
- [x] Benchmark: baseline (v5.5-alpha) v.s. API v2, observing QPS and P99 latency
  - [x] raw_put
  - [x] raw_delete
  - [x] raw_get

#### Phase 2 (CDC):

- [x] CDC component @zeminzhou
  - [x] Rawkv procedure #12387
    - [x] Raw changefeed (receive rawkv changefeed from TiKV-CDC)
    - [x] Incremental scan
    - [x] Raw sink (send rawkv data to TiKV-CDC)
  - [x] Resolved_ts #12866
- [x] Metrics
    - [x] TiKV-CDC Grafana board https://github.com/tikv/migration/pull/130
    - [x] Resolved-ts metrics
- [x] Test
  - [x] Integration Test for causality validation
    - [x] TiKV - CDC - TiKV
    - [x] Leader Transfer
    - [x] Region split + leader transfer
    - [x] Region merge + leader transfer
- [x] Benchmark: baseline (v5.5-alpha) v.s. CDC feature disabled v.s CDC feature enabled, observing QPS and P99 latency
  - [x] raw_put
  - [x] raw_delete
[...]

PR #12036, opened 2022-02-28
api_version: Codec for RawKV key

pingyu:

Issue Number: ref #11965 

What's Changed:

Key of RawKV is encoded as `user-key + memcomparable-padding + timestamp` in API v2.
A structure `RawKey` is added.

Refer to RFC:
1. https://github.com/pingyu/tikv-rfcs/blob/rfc_api_v2_pd/text/0069-api-v2.md (https://github.com/tikv/rfcs/pull/79)
2. https://github.com/pingyu/tikv-rfcs/blob/rawkv-cdc/text/0083-rawkv-cross-cluster-replication.md (https://github.com/tikv/rfcs/pull/86)

```commit-message
Key of RawKV is encoded as `user-key + memcomparable-padding + timestamp` in API v2.
```

### Related changes

No.
No.
No.

### Check List 

Tests 

- Unit test
- Integration test

Side effects
No.

### Release note 

```release-note
RawKV key is encoded as user-key + memcomparable-padding + timestamp in API v2.
```


andylokandy (approved):

haojinming (approved):

pingyu:
> /cc @innerr @BusyJay

PTAL, thanks~

components/api_version/src/api_v2.rs
```diff
@@ -107,4 +107,58 @@ impl APIVersion for APIV2 {
...
+fn verify_encoded_key(encoded_key: &Key, with_ts: bool) -> bool {
+    let encoded_bytes = &encoded_key.as_encoded()[..];
+    if !matches!(APIV2::parse_key_mode(encoded_bytes), KeyMode::Raw) {
+        return false;
+    }
+    let mut len = encoded_bytes.len();
+    if with_ts {
+        len = len.checked_sub(number::U64_SIZE).unwrap();
+    }
+    len > 0 && (len % (tikv_util::codec::bytes::ENC_GROUP_SIZE + 1) == 0)
```
BusyJay, line 153: How about `bytes::decode_bytes(&mut encoded_bytes, false).is_ok() && encoded_bytes.len() == 8 * (with_ts as usize)`? So it doesn't need to depend on the details of codec.

components/api_version/src/api_v2.rs
```diff
@@ -107,4 +107,58 @@ impl APIVersion for APIV2 {
...
+        let encoded_key = Key::from_raw(user_key);
+        if let Some(ts) = ts {
+            encoded_key.append_ts(ts)
+        } else {
+            encoded_key
+        }
+    }
+
+    // `encode_raw_key_owned` has improvement only for APIV1|APIV1TTL, but not for APIV2.
+    // As there is no "in place" implementation available for encoding.
```
BusyJay, line 137: Actually there is. See `encode_all_in_place`.

pingyu:
@BusyJay Thanks for your review !
Comments addressed, PTAL~

components/api_version/src/api_v2.rs
```diff
@@ -107,4 +108,65 @@ impl APIVersion for APIV2 {
...
+        } else {
+            encoded_key
+        }
+    }
+
+    fn encode_raw_key_owned(mut user_key: Vec<u8>, ts: Option<TimeStamp>) -> Key {
+        debug_assert_eq!(Self::parse_key_mode(&user_key), KeyMode::Raw);
+        let src_len = user_key.len();
+        let encoded_len = MemComparableByteCodec::encoded_len(src_len);
+        user_key.reserve(encoded_len - src_len + number::U64_SIZE * (ts.is_some() as usize));
```
BusyJay, line 141: This is redundant as it will be called inside `resize`.

components/tikv_util/src/codec/bytes.rs
```diff
@@ -7,7 +7,7 @@ use super::{BytesSlice, Error, Result};
 use crate::codec::number::{self, NumberEncoder};
 use std::ptr;
 
-const ENC_GROUP_SIZE: usize = 8;
+pub const ENC_GROUP_SIZE: usize = 8;
```
BusyJay, line 10: Doesn't have to be public.

BusyJay (commented):
Rest LGTM

components/api_version/src/api_v2.rs
pingyu: We maybe need 8 more bytes for timestamp. Is it possible to extend the capacity again when we invoke `append_ts` later, if we do not reserve 8 more bytes here ?

BusyJay: Got it.

BusyJay (approved):

PR #12047, opened 2022-02-28
server: disallow coexist of raft engine and raftdb

### LINKED ISSUE #12045: RocksDB and Raft Engine data exist simultaneously due to operational errors

nightly

N/A

### Steps to reproduce

1. Start TiKV with nightly version
2. Downgrade TiKV to 5.3.0
3. **Raft Engine reports "Parse file name error", TiKV won't start**
4. Start TiKV with nightly version
5. **TiKV panicks with "log at recorded commit index [19] 102 doesn't exist, may lose data"**

### What did you expect?

Step 5 TiKV can start normally.

### What did happened?

TiKV by default disable Raft Engine in v5.3.0. It therefore creates a raftdb directory at step 3.

In step 5, TiKV tries to dump the data from raftdb directory into Raft Engine. In that process, Raft Engine data is cleared via this function:
https://github.com/tikv/tikv/blob/6170255604aff4a52b365c91d09859e4315045c5/components/server/src/raft_engine_switch.rs#L46-L53

tabokie:

Issue Number: Close #12045 

What's Changed:

```commit-message
```

### Related changes

### Check List 

Tests
- No code

### Release note 

```release-note
None
```


src/config.rs
```diff
@@ -2671,6 +2671,11 @@ impl TiKvConfig {
         {
             return Err("default rocksdb doesn't exist, but raftdb or raft engine exists".into());
         }
+        if RocksEngine::exists(&self.raft_store.raftdb_path)
```
BusyJay, line 2674: What if TiKV crashes during upgrading data from rocksdb to raft engine?

tabokie: I'll add a deferred cleanup. But in case the directory is still not cleaned up due to filesystem error, user has to manually do it.

BusyJay: How about using a temporary directory and rename to do atomic upgrade?

tabokie: I think your solution is safer, but it involves some extra work to reopen the db. Plus it won't prevent coexist if filesystem does err (rename of target directory succeeds, rename of source directory fails).

BusyJay: I think the process should be:
1. Upgraded to tempdir (you can even used a fixed name), let's refer as raft0;
2. Rename rocksdb directory to a temp directory, let's refer it as rocks1 -> rocks0;
3. Rename tempdir to raft engine directory, let's refer it as raft0 -> raft1;
4. Delete rocksdb directory.

So no matter which steps it crashes, it can still resume automatically. Note following situations are impossible:
1. raft0 and raft1 coexist;
2. rocks0 and rocks1 coexist;
3. raft1 and rocks1 coexist.

So if raft0 and rocks0 coexist, just resume from step 3. If rocks1 exists, delete raft0 if exists and resume from step 1. If raft1 exists, delete rocks0 if exists, and finish upgrading.

tabokie: Nice, but if we allow a weird intermediate state to exist, I can do better without reopening DB:
[A] -> [A.0] -> [A.0, B] -> [B] (A.0 takes priority over any directory)

Maybe I'll try it later.

BusyJay: I have to say I don't understand what you mean.

tabokie: A/B refers to your raft1/rocks1, A.0 refers to your raft0/rocks0.

BusyJay: I don't think there is an atomic way to delete `A.0`. It's totally possible that only some files in `A.0` are deleted before TiKV crashes.

tabokie: Oh, just add another rename. The old code already does it this way.

tabokie: On a second thought, I think this introduces a lot of complexities, especially to the config check and engine init part. I'll stick to the current solution.

BusyJay: > introduces a lot of complexities, especially to the config check

Why? What it needs to do is just guarantee the three asserts listed above.

> and engine init part

It just needs to follow the migration order mentioned above.

tabokie: We might have different understanding for "complex". I consider gluing config check and engine dumping with a shared state machine a serious complexity. And I prefer not to do it given the current solution already prevents data corruption.

BusyJay: So we know it's a possible state, and we know the state is fine, and the correct way to fix the problem is just delete the directory. However, somehow we choose to have someone to be called up to handle it instead of adding about tens of lines? 🤣

tabokie: I understand this sentiment. But the possibility of failing to rm a directory (while succeed in opening a DB) is pretty low. And we give clear instruction to the user (rm a specific directory manually). Meanwhile a decent implementation (without looking like a patchwork) of the novel solution we discussed would probably takes hundreds of lines.

tabokie: sigh.. Just noticed we use `process::abort` in panic hook, so that defer block won't be called. Maybe I still have to do it the hard way.

components/engine_traits/src/util.rs
```diff
@@ -21,3 +24,151 @@ pub fn check_key_in_range(
         })
     }
 }
+
+pub struct RaftDataStateMachine {
```
BusyJay, line 28: Can you add some comments to explain how it's expected to migrate?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +24,151 @@ pub fn check_key_in_range(
...
+    fn data_exists(path: &Path) -> bool {
+        if !path.exists() || !path.is_dir() {
+            return false;
+        }
+        fs::read_dir(&path).unwrap().next().is_some()
+    }
+
+    fn remove_dir_safe(path: &Path) {
+        if path.exists() {
+            let trash = path.with_extension("REMOVE");
```
BusyJay, line 112: Actually, this can leave dirty data forever if it's renamed but failed to remove.

components/server/src/raft_engine_switch.rs
```diff
@@ -137,22 +50,81 @@ pub fn check_and_dump_raft_db(
...
+        "Cannot transfer data from RaftDb to non-empty Raft Engine."
+    );
+}
+
+fn check_raft_db_is_empty(engine: &RocksEngine) {
+    let start_key = keys::raft_log_key(0, 0);
+    let end_key = keys::raft_log_key(u64::MAX, u64::MAX);
+    let mut count = 0;
+    engine
+        .scan(&start_key, &end_key, false, |_, _| {
```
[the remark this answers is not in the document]
tabokie: I copied the old code. And using `raft_log_key` is probably more readable.

components/engine_traits/src/util.rs
```diff
@@ -21,3 +24,182 @@ pub fn check_key_in_range(
...
+        if Self::data_exists(&self.target_staging) {
+            // InvMigrating -> Completed
+            assert!(!Self::data_exists(&self.source_staging));
+            Self::remove_dir_safe(&self.source);
+            Self::rename_dir_safe(&self.target_staging, &self.target);
+            return None;
+        }
+        if Self::data_exists(&self.source_staging) {
+            // Double-check.
+            Self::remove_dir_safe(&self.source);
```
BusyJay, line 115: How can this happen?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +24,182 @@ pub fn check_key_in_range(
...
+        }
+        Some(self.source_staging.clone())
+    }
+
+    /// Exits the `Migrating` state and enters the `Completed` state.
+    pub fn after_dump_data(&mut self) {
+        assert!(Self::data_exists(&self.target));
+        assert!(!Self::data_exists(&self.source));
+        Self::remove_dir_safe(&self.source_staging);
+        Self::remove_dir_safe(&self.target_staging);
```
BusyJay, line 131: Removing `target_staging` seems useless.

components/engine_traits/src/util.rs
```diff
@@ -21,3 +24,182 @@ pub fn check_key_in_range(
...
+    }
+
+    fn data_exists(path: &Path) -> bool {
+        if !path.exists() || !path.is_dir() {
+            return false;
+        }
+        fs::read_dir(&path).unwrap().next().is_some()
+    }
+
+    fn remove_dir_safe(path: &Path) {
```
BusyJay, line 141: I suggest to always log the operation for auditing.

components/server/src/raft_engine_switch.rs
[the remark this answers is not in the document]
BusyJay: Using `raft_log_key` seems confused to me. First, `raft_state_key` is technically encoded as following the `raft_log_key`, using `raft_log_key` may raise the question that is it possible to leave some dirty data? Second, raft log key across regions is meaningless as log key space is not continuous. Using `b""` is clear that it checks if there is any key inside the db.

components/engine_traits/src/util.rs
```diff
@@ -21,3 +24,182 @@ pub fn check_key_in_range(
...
+        let mut new_source_file = new_source;
+        new_source_file.push("file");
+        let mut target_file = target.clone();
+        target_file.push("file");
+        fs::copy(&new_source_file, &target_file).unwrap();
+        validate_state(&source, &target, true);
+        fs::remove_file(&new_source_file).unwrap();
+        validate_state(&source, &target, true);
+        state.after_dump_data();
+        validate_state(&source, &target, true);
```
BusyJay, line 203: Seems invmigrate not cover.

BusyJay (commented):
Rest LGTM.

components/engine_traits/src/util.rs
tabokie: It can't happen. But the protocol itself doesn't require `source` to be absent in this condition too. I'll make the requirement stronger and replace these loose ends with assertions.

tabokie:
Moved the thing to `tikv_util` to avoid introducing slog dependency to `engine_traits`.

components/tikv_util/src/config.rs
```diff
@@ -1416,6 +1416,148 @@ macro_rules! numeric_enum_serializing_mod {
...
+    fn data_exists(path: &Path) -> bool {
+        if !path.exists() || !path.is_dir() {
+            return false;
+        }
+        fs::read_dir(&path).unwrap().next().is_some()
+    }
+
+    fn remove_dir_safe(path: &Path) {
+        if path.exists() {
+            info!("Removing directory: {}", path.display());
```
BusyJay, line 1546: Should use the format `info!("Removing directory", "path" => %path.display())`.

BusyJay (approved):

components/server/src/server.rs
```diff
@@ -1259,23 +1259,41 @@ impl TiKVServer<RocksEngine> {
...
 
-        // Create raft engine.
-        let raft_db_path = Path::new(&self.config.raft_store.raftdb_path);
+        let mut raft_data_state_machine = RaftDataStateMachine::new(
+            &self.config.raft_engine.config().dir,
+            &self.config.raft_store.raftdb_path,
+        );
+        let dump_source = raft_data_state_machine.before_open_target();
+
+        let raftdb_path = &self.config.raft_store.raftdb_path;
```
Connor1996, line 1274: keep `raft_db`, which is consistent with `kv_db`

components/server/src/raft_engine_switch.rs
```diff
@@ -137,22 +50,79 @@ pub fn check_and_dump_raft_db(
...
+        "Cannot transfer data from RaftDb to non-empty Raft Engine."
+    );
+}
+
+fn check_raft_db_is_empty(engine: &RocksEngine) {
+    let mut count = 0;
+    engine
+        .scan(b"", &[0xFF, 0xFF], false, |_, _| {
+            count += 1;
+            Ok(true)
```
Connor1996, line 117:
```suggestion
            Ok(false)
```
return directly if there is any.

components/server/src/raft_engine_switch.rs
```diff
@@ -1,126 +1,39 @@
...
-    clear_raft_engine(engine).expect("clear_raft_engine");
-
-    let config_raftdb = &config.raftdb;
-    let mut raft_db_opts = config_raftdb.build_opt();
-    raft_db_opts.set_env(env.clone());
-    let raft_db_cf_opts = config_raftdb.build_cf_opts(&None);
-    let db = engine_rocks::raw_util::new_engine_opt(raftdb_path, raft_db_opts, raft_db_cf_opts)
-        .unwrap_or_else(|s| fatal!("failed to create origin raft db: {}", s));
-    let src_engine = RocksEngine::from_db(Arc::new(db));
+pub fn dump_raftdb_to_raft_engine(source: &RocksEngine, target: &RaftLogEngine, threads: usize) {
```
Connor1996, line 15: can we move these functions to `RaftDataStateMachine`? so the logic of the struct is more coherent.

components/server/src/server.rs
```diff
@@ -1259,23 +1259,41 @@ impl TiKVServer<RocksEngine> {
...
             .build_shared_rocks_env(self.encryption_key_manager.clone(), get_io_rate_limiter())
             .unwrap();
 
-        // Create raft engine.
-        let raft_db_path = Path::new(&self.config.raft_store.raftdb_path);
+        let mut raft_data_state_machine = RaftDataStateMachine::new(
+            &self.config.raft_engine.config().dir,
+            &self.config.raft_store.raftdb_path,
+        );
+        let dump_source = raft_data_state_machine.before_open_target();
```
Connor1996, line 1272: Seems it's okay to do the check after opening

tabokie: Opening first will create a non-empty directory. `target` and `source` coexist.

components/server/src/server.rs
tabokie: I'll use `kvdb` instead. It's in line with how we name the configurations.

components/server/src/raft_engine_switch.rs
tabokie: It's tempting. But I also think state machine should avoid depending on specific engine implementation.

Connor1996 (approved):
LGTM

PR #12068, opened 2022-03-02
storage: support scale store scheduler pool size

### LINKED ISSUE #12067: online config: support adjusting scheduler-worker-pool-size dynamically

A dynamic config modification is performed from TiKV-Control command line (documentation). So expect to adjust the `scheduler-worker-pool-size` for storage.
```
set config tikv `storage.scheduler-worker-pool-size`=16;
```

glorv:

Issue Number: Close #12067

What's Changed:
Update yatp to the newest version to support dynamic scale yatp thread pool. Change `FuturePool` by set min/max thread count to 1/num_cpu to support dynamic adjust the active thread count of the future pool.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

```release-note
storage: Support adjusting scheduler-worker-pool-size dynamically 
```


glorv:
/cc @ethercflow

Some of the logic change in this PR should also affect online config reader pool size, so please also take a look.

src/storage/txn/scheduler.rs
```diff
@@ -663,6 +673,7 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
         let resource_tag = self.inner.resource_tag_factory.new_tag(task.cmd.ctx());
         async {
             let tag = task.cmd.tag();
+            println!("start process cmd: {}", task.cmd);
```
ethercflow, line 676: Legacy debug statements?

glorv:
@NingLin-P  @5kbpers PTAL, thanks~

src/storage/config_manager.rs
```diff
@@ -0,0 +1,126 @@
...
+                            .set_options_cf(cf, &[("disable_write_stall", "false")])
+                            .unwrap();
+                    }
+                    self.flow_controller.enable(false);
+                }
+            }
+        } else if let Some(v) = change.get("scheduler_worker_pool_size") {
+            let pool_size: usize = v.into();
+            let max_pool_size = std::cmp::max(1, SysQuota::cpu_cores_quota() as usize);
+            if pool_size == 0 || pool_size > max_pool_size {
```
ethercflow, line 97: When a configuration file is loaded or updated, configuration verification is performed and the validate function of each configuration block is called, so I think we'd better move this to validate too.

src/config.rs
```diff
@@ -4176,6 +4193,36 @@ mod tests {
...
+            scheduler
+                .get_sched_pool(CommandPri::Normal)
+                .pool
+                .get_pool_size(),
+            4
+        );
+
+        cfg_controller
+            .update_config("storage.scheduler-worker-pool-size", "1")
+            .unwrap();
```
NingLin-P, line 4214: Plz add more cases like 0 and also check the `high_priority_pool`

src/storage/config.rs
```diff
@@ -113,6 +109,15 @@ impl Config {
                     .into(),
             );
         };
+        let max_pool_size = std::cmp::max(1, SysQuota::cpu_cores_quota() as usize);
+        if self.scheduler_worker_pool_size == 0 || self.scheduler_worker_pool_size > max_pool_size {
```
NingLin-P, line 113: Not sure if it is a good idea, since TiKV uses block io currently, the user may want to use more threads to get full utilization of cpu cores. And since now we support changing the pool size online, the user can always decrease the pool size if it hurt performance, so may not need such restricted restraint.

glorv: Because yatp thread pool need to set the `max-thread-count` before initializing, and after the pool is built, and dynamic pool size change can not exceed this threshold. And, bacause yatp pool need to pre-create `max-thread-count` threads, so I think the cpu_num should be a proper upper limit.  /cc @ethercflow

ethercflow: > Because yatp thread pool need to set the max-thread-count before initializing, and after the pool is built, and dynamic pool size change can not exceed this threshold. And, bacause yatp pool need to pre-create max-thread-count threads, so I think the cpu_num should be a proper upper limit. /cc @ethercflow

For IDC environment, I think `max_thread_count` is `SysQuota::cpu_cores_quota()` maybe enough.
For cloud environment, eg: 4 core or 8 core,  I've no idea.  How much do we need at most in this type of scenario? @NingLin-P

NingLin-P: /cc @BusyJay

BusyJay: I don't think we have ever configure the scheduler worker pool larger than the cpu cores. But the situation may not be valid for 4 cores or even 1 cores. And we may need to bench how it performs when too many threads are spawn even without being used.

glorv: So, how about set a hard threshold of max-threshold-count when the total CPU Quota is relative small, e.g. 4, or a dynamic config like (CPUQuota  * 2)?

glorv: @NingLin-P @BusyJay @ethercflow what do you think? Or what other benchmark do we need to do?

glorv: according @BusyJay 's recommendation, set the min value of `max-thread-count` to 4 when the cpu quota is smaller.

NingLin-P (approved):

src/storage/config.rs
```diff
@@ -113,6 +109,15 @@ impl Config {
                     .into(),
             );
         };
+        let max_pool_size = std::cmp::max(1, SysQuota::cpu_cores_quota() as usize);
```
NingLin-P, line 112: Should be `std::cmp::max(1, std::cmp::min(4, cpu_num as usize))`

glorv: updated. can be simplified to `std::cmp::max(4, cpu_num)`

PR #12076, opened 2022-03-03
raftstore: drop raft message by replication state

### LINKED ISSUE #12074: replication mode: drop raft message by replication state
## Development Task

To avoid breaking ACID integraty of dr center, we need to cut off network connections between normal stores and unhealthy stores (stores that cannot connect to PD).

disksing:

Issue Number: Close #12074

What's Changed: Drop raft messages when the peer store is not available judged by replication mode state.

```commit-message
```

### Check List 

Tests 
- Unit test
- Manual test

Side effects
- Performance regression
    - Consumes more CPU

### Release note 

```release-note
None
```


BusyJay:
Can you describe more on what's is available stores? Or any public link to show the definition?

disksing:
available stores are stores that are allowed to communicate judged by the replication mode from PD's perspective. 
Specifically.
1) When the PD determines that DR  has failed and needs to switch to the async state, all nodes in the DR will be marked as unavailable store
2) When a small number of nodes in PRIMARY fail and cannot update the replication mode, these failed nodes will also be marked as unavailable store

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1752,6 +1752,23 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
             MEMTRACE_RAFT_MESSAGES.trace(TraceEvent::Sub(heap_size));
         });
 
+        if !self
+            .ctx
+            .global_replication_state
+            .lock()
```
nolouch, line 1758: is it ok？ cc @BusyJay

components/raftstore/src/store/replication_mode.rs
```diff
@@ -143,13 +143,23 @@ impl StoreGroup {
...
 /// A global state that stores current replication mode and related metadata.
-#[derive(Default, Debug)]
+#[derive(Debug)]
 pub struct GlobalReplicationState {
     status: ReplicationStatus,
     pub group: StoreGroup,
     group_buffer: Vec<(u64, u64)>,
 }
 
+impl Default for GlobalReplicationState {
```
nolouch, line 153: can drive `Default`.

components/raftstore/src/store/replication_mode.rs
```diff
@@ -185,6 +195,16 @@ impl GlobalReplicationState {
             ReplicationMode::Majority => None,
         }
     }
+
+    pub fn check_store_available(&self, store_id: u64) -> bool {
+        match self.status.get_mode() {
+            ReplicationMode::Majority => true,
+            ReplicationMode::DrAutoSync => {
+                let available_stores = self.status.get_dr_auto_sync().get_available_stores();
+                available_stores.is_empty() || available_stores.contains(&store_id)
```
nolouch, line 204: Why empty is true?

nolouch, line 204: Why empty is true?

components/raftstore/src/store/fsm/store.rs
BusyJay, line 1758: I'm afraid this can be the bottleneck of the system since raft messages are very frequent. I would be better to implement it on the network side. If it's in blocklist, the connection should be aborted and rejected on established.

disksing: That way could work too. But I have not found a way to notify to network layer (Transport, RaftClient, etc.) as soon as the replication state is changed. To be more rigorous, we would actually also need the network layer to return success when the setup is complete before the lower layers can continue to synchronize new state to the PD. Do you have any suggestions for implementing such a notification mechanism?

components/raftstore/src/store/replication_mode.rs
disksing: PD only set it in `wait_async` and `async` states. Other states, we do not care the store list.

src/server/raft_client.rs
```diff
@@ -834,6 +834,7 @@ pub struct RaftClient<S, R, E> {
     builder: ConnectionBuilder<S, R>,
     engine: PhantomData<E>,
     last_hash: (u64, u64),
+    store_whitelist: Vec<u64>,
```
BusyJay, line 837: This is not shared by threads. It's cloned to every threads. So setting it once may not update the copy of other threads.

Technically, it should  kill the connection in `ConnectionPool` and replace it with a blackhole. So other threads will refresh its connection list and find it's blocked.

It will be easier if `CachedQueue` support `pause`.

disksing: It looks that simply pause `CacheQueue` may not be enough. We should make sure that connections to blocked stores not be created in the future.

components/raftstore/src/store/async_io/write_tests.rs
```diff
@@ -75,6 +75,7 @@ impl Transport for TestTransport {
         self.tx.send(msg).unwrap();
         Ok(())
     }
+    fn set_store_whitelist(&mut self, _: Vec<u64>) {}
```
BusyJay, line 78:
```suggestion
    fn set_store_allowlist(&mut self, _: Vec<u64>) {}
```

src/server/raft_client.rs
```diff
@@ -46,6 +46,7 @@ struct Queue {
     buf: ArrayQueue<RaftMessage>,
     /// A flag indicates whether the queue can still accept messages.
     connected: AtomicBool,
+    paused: AtomicBool,
```
BusyJay, line 49: It can be merged with `connected`.

components/test_raftstore/src/node.rs
```diff
@@ -124,6 +124,8 @@ impl Transport for ChannelTransport {
         }
     }
 
+    fn set_store_allowlist(&mut self, _whitelist: Vec<u64>) {}
```
BusyJay, line 127:
```suggestion
    fn set_store_allowlist(&mut self, _allowlist: Vec<u64>) {}
```

src/server/raft_client.rs
BusyJay: Yes, what I mean is using an `AtomicU8` to merge `connected` and `paused`. Apparently, a connection can either be connected, paused or disconnected (should never connect again).

disksing: ok

src/server/raft_client.rs
```diff
@@ -66,25 +74,19 @@ impl Queue {
...
-                Err(_) => return Err(DiscardReason::Full),
-            }
-        } else {
-            return Err(DiscardReason::Disconnected);
-        }
-        if self.connected.load(Ordering::SeqCst) {
-            Ok(())
-        } else {
-            Err(DiscardReason::Disconnected)
+        match self.conn_state.load(Ordering::SeqCst) {
```
nolouch, line 77: How about using `into` with:
```
impl From<u8> for ConnState
```

nolouch (commented):
rest LGTM

src/server/raft_client.rs
disksing: I'll give a try.

BusyJay (commented):
How about also adding a case to verify raftstore will accept messages from PD and update allowlist correctly?

Rest LGTM

BusyJay (approved):

nolouch (approved):
lgtm

PR #12077, opened 2022-03-03
*: bump master branch version to v6.0.0-alpha

sticnarf:
### What is changed and how it works?

Issue Number: Close #12075 

Features that should be only enabled in the next release may depend on this change.

### Release note 

```release-note
None
```

PR #12078, opened 2022-03-03
storage: set in-memory pessimistic lock feature gate to 6.0.0

sticnarf:
Depends on:

- [x] https://github.com/tikv/tikv/pull/12077

Issue Number: Ref #11452

What's Changed:

```commit-message
The in-memory pessimistic lock feature should be only enabled after
all TiKV instances have been upgraded to the new version. Otherwise,
transferring leader may be blocked because of the new protocol and
procedure when in-memory pessimistic locks exist.

So, we need to use a feature gate to make sure this feature is only
enabled after all TiKVs have been upgraded to the new version.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


youjiali1995 (approved):
Is it a good idea to make `FeatureGate` a lazy_static one?

sticnarf:
> Is it a good idea to make `FeatureGate` a lazy_static one?

I don't know. Global static one may be easier to use but less easier for tests.

cfzjywxk (approved):

PR #12087, opened 2022-03-03
APIV2: Implement RawMvccSnapshot & RawMvccIterator

haojinming:

Issue Number: Ref #11965 

What's Changed:

Implement RawEncodeSnapshot::get related function with seek mode in API v2, because we will encode timestamp in APIV2 key. 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
we had a demo to test the performance regression, please ref: https://github.com/tikv/rfcs/pull/86

```release-note
There are multi version of one same user key in api v2. We return the newest one when client call raw_get/raw_scan.
```


src/storage/raw/store.rs
```diff
@@ -223,6 +223,53 @@ impl<'a, S: Snapshot> RawStoreInner<S> {
         Ok(pairs)
     }
 
+    pub async fn forward_raw_scan_v2(
```
pingyu, line 226: I think we need to skip entries: 1. without the largest timestamp of the same user key, and 2. logical deleted.

pingyu:
> /cc @andylokandy @BusyJay @Little-Wallace

PTAL, thanks~

src/storage/raw/store.rs
haojinming: now i implement the raw_get_key_value/ forward_raw_scan/reverse_raw_scan operation in RawStoreV2 to have unifued view. Same key is skipped, and deletion can only be judged after PR#12090 merged.

src/storage/raw/store.rs
```diff
@@ -317,3 +317,200 @@ impl<'a, S: Snapshot> RawStoreInner<S> {
...
+            let cur_key = cursor.key(statistics).to_owned();
+            pairs.push(Ok((cur_key.clone(), if key_only {
+                vec![]
+            } else {
+                cursor.value(statistics).to_owned()
+            })));
+            if pairs.len() < limit {
+                cursor.next(statistics);
+                while cursor.valid()? && cur_key == cursor.key(statistics) {
+                    cursor.next(statistics);
```
BusyJay, line 404: If there are too many same versions, this can be very slow.

haojinming: yes, the scan perfromance will be impacted by multi ts of same user keys. Do you have any idea to optimize this?

BusyJay: Transaction layer has done a lot of improvement on these cases, why not reuse the code?

haojinming: currently, mvcc scanner cannot be used here, as timestamp is not encode in key for now. i prefer to keep current implentation and do an optimization after all apiv2 code is ready. Another thing is,  mvcc scanner has  some process handling IsolationLevel and Lock, we need ensure that IsolationLevel can be empty in RawKv mode and donnot bring in much burden.

src/storage/raw/encoded.rs
```diff
@@ -44,6 +44,28 @@ impl<S: Snapshot, API: APIVersion> RawEncodeSnapshot<S, API> {
         }
     }
 
+    pub fn seek_first_key_value_cf(&self, cf: CfName, key: &Key) -> Result<Option<Value>> {
+        let mut iter_opt = IterOptions::default();
+        iter_opt.set_fill_cache(false);
```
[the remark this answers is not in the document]
haojinming: should we fill cache here?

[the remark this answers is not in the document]
haojinming: addressd, default to be true if no option input.

src/storage/raw/encoded.rs
```diff
@@ -44,6 +44,28 @@ impl<S: Snapshot, API: APIVersion> RawEncodeSnapshot<S, API> {
...
+        iter_opt.use_prefix_seek();
+        iter_opt.set_prefix_same_as_start(true);
+        let end_key = key.clone().append_ts(TimeStamp::zero());
+        iter_opt.set_upper_bound(end_key.as_encoded(), DATA_KEY_PREFIX_LEN);
+        let mut iter = self.iter_cf(cf, iter_opt)?;
+        if !iter.seek(key)? {
+            Ok(None)
+        } else if iter.valid()? {
```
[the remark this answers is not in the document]
haojinming: addressd.

src/storage/raw/store.rs
haojinming: Per discussion with @pingyu @andylokandy , rawkv  need process the ttl and logical deletion after mvcc. Define rawstore currently as RawStoreInner<RawEncodeSnapshot<RawMvccSnapshot<S>, APIV2>>， RawMvccSnapshot process mvcc of keys and RawEncodeSnapshot process meta in value.

src/storage/raw/store.rs
haojinming: Define apiv2 RawStore as RawStoreInner<RawEncodeSnapshot<RawMvccSnapshot, APIV2>>, RawMvccSnapshot process mvcc of keys and RawEncodeSnapshot process meta in value

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,197 @@
...
+        iter_opt.use_prefix_seek();
+        iter_opt.set_prefix_same_as_start(true);
+        let end_key = key.clone().append_ts(TimeStamp::zero());
+        iter_opt.set_upper_bound(end_key.as_encoded(), DATA_KEY_PREFIX_LEN);
+        let mut iter = match cf {
+            Some(cf_name) => self.iter_cf(cf_name, iter_opt)?,
+            None => self.iter(iter_opt)?,
+        };
+        if iter.seek(key)? {
+            // Following line can be removed after mce padding in api v2.
```
BusyJay, line 37: Actually, without padding, this function will not work as expected, returning `None` is also wrong.

BusyJay (commented):
/cc @Connor1996 PTAL

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,197 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+// #[PerformanceCriticalPath]
```
Connor1996, line 3: remove it

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,197 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```
Connor1996, line 1:
```suggestion
// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```

src/storage/raw/raw_mvcc.rs
Connor1996: +1

haojinming: emm... yes. So currently, case of apiv2 are using keys with same length. I think we can add more cases with different length keys after do mce padding in raw_put/raw_batch_put functions. Mce padding PR is one the way by @pingyu.

haojinming: For now, i will remove this equal judgement.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,197 @@
...
+impl<S: Snapshot> RawMvccSnapshot<S> {
+    pub fn from_snapshot(snap: S) -> Self {
+        RawMvccSnapshot { snap }
+    }
+
+    pub fn seek_first_key_value_cf(
+        &self,
+        cf: Option<CfName>,
+        opts: Option<ReadOptions>,
+        key: &Key,
```
Connor1996, line 24: In `raw_get()` of `storage.rs`, `Key` is constructed by `Key::from_encoded(raw_key)` which is just to reuse the  `Key` struct for raw operations, but the key isn't actually an encoded key. It doesn't matter for rawkv V1, but it is wrong for the mvcc rawkvV2.

Connor1996: need to adjust the code in `storage.rs` as well

Connor1996 (commented):
Please add test case to cover this

src/storage/raw/raw_mvcc.rs
haojinming: we will use from_raw to encode the apiv2 key as txn kv does. The pr is on the way by@pingyu.

src/storage/raw/raw_mvcc.rs
haojinming: addressed

src/storage/raw/raw_mvcc.rs
haojinming: addressed

src/storage/raw/raw_mvcc.rs
haojinming: Actually we define a struct RawKey to encode apiv2 key and encode timestamp in key. Please ref: https://github.com/tikv/rfcs/pull/86

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,191 @@
...
+    fn seek_to_last(&mut self) -> Result<bool> {
+        let res = self.inner.seek_to_last();
+        self.move_to_prev_max_ts(res)
+    }
+
+    fn valid(&self) -> Result<bool> {
+        match self.is_valid {
+            Some(v) => Ok(v),
+            None => self.inner.valid(),
+        }
```
andylokandy, line 171: You can use [`unwrap_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or) for this pattern.

haojinming: adressed

haojinming:
Review comments are addressed. please have another look at this pr. Thanks. @andylokandy @pingyu @BusyJay .

src/storage/mod.rs
```diff
@@ -8344,4 +8348,98 @@ mod tests {
         // Prewrite still succeeds
         assert!(rx.recv().unwrap().is_ok());
     }
+
+    #[test]
+    fn test_raw_mvcc_snapshot() {
+        let storage = TestStorageBuilder::new(DummyLockManager {}, ApiVersion::V2)
```
BusyJay, line 8354:
```suggestion
        let storage = TestStorageBuilder::new(DummyLockManager, ApiVersion::V2)
```

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    fn clear_cur_kv(&mut self) {
+        self.cur_key = None;
+        self.cur_value = None;
+        self.is_valid = None;
+    }
+
+    fn move_to_prev_max_ts(&mut self, res: Result<bool>) -> Result<bool> {
+        if *res.as_ref().unwrap() && self.inner.valid()? {
```
BusyJay, line 124: Seems res is always unwrap, why not just take a bool instead?

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+        res
+    }
+}
+
+impl<I: Iterator> Iterator for RawMvccIterator<I> {
+    fn next(&mut self) -> Result<bool> {
+        let cur_key = self.inner.key().to_owned();
+        let mut res = self.inner.next();
+        while *res.as_ref().unwrap()
```
BusyJay, line 147: Why unwrap instead of returning error?

src/storage/raw/raw_mvcc.rs
haojinming: good idea. This let me rethink of the usage of unwrap and res here. I think res is not suitable to be an paramenter of this function. I remove it from this function.

src/storage/mod.rs
haojinming: addressed

src/storage/raw/raw_mvcc.rs
```diff
@@ -120,60 +120,64 @@ impl<I: Iterator> RawMvccIterator<I> {
...
 impl<I: Iterator> Iterator for RawMvccIterator<I> {
     fn next(&mut self) -> Result<bool> {
         let cur_key = self.inner.key().to_owned();
-        let mut res = self.inner.next();
-        while *res.as_ref().unwrap()
-            && self.inner.valid()?
-            && is_user_key_eq(&cur_key, self.inner.key())
-        {
-            res = self.inner.next();
+        let mut res = self.inner.next()?;
```
BusyJay, line 147: I think it's better to make sure why it's chosen to panicked in the old code.

src/storage/raw/raw_mvcc.rs
```diff
@@ -120,60 +120,64 @@ impl<I: Iterator> RawMvccIterator<I> {
...
     }
 
     fn seek(&mut self, key: &Key) -> Result<bool> {
         self.inner.seek(key)
     }
 
     fn seek_for_prev(&mut self, key: &Key) -> Result<bool> {
-        let res = self.inner.seek_for_prev(key);
-        self.move_to_prev_max_ts(res)
+        if self.inner.seek_for_prev(key)? {
```
BusyJay, line 164: It seems to left `cur_key` untouched.

haojinming: only if "seek_for_prev(key)?"  return err, cur_key/cur_value are untouched, in this case, caller should not call key() or value(), cur_key is useless. If call key()/value() abort will occur.

src/storage/raw/raw_mvcc.rs
haojinming: actually this file is newly added in this pr. Panic is a bug, i think, as it's a common case that seek_for_prev return error when disk error or something else. In this case, we should return the error to outside other than panic.

haojinming: thanks for pointing this issue.

src/storage/raw/raw_mvcc.rs
haojinming: test_raw_mvcc_snapshot is added to cover this file.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,215 @@
...
+            inner,
+            cur_key: None,
+            cur_value: None,
+            is_valid: None,
+        }
+    }
+
+    fn update_cur_kv(&mut self) {
+        if let Some(ref mut key) = self.cur_key {
+            key.clear();
```
BusyJay, line 113: It may need to be shrink if key becomes too long.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,215 @@
...
+        let mut res = self.inner.next()?;
+        while res && self.inner.valid()? && is_user_key_eq(&cur_key, self.inner.key()) {
+            res = self.inner.next()?;
+        }
+        self.clear_cur_kv();
+        Ok(res)
+    }
+
+    fn prev(&mut self) -> Result<bool> {
+        self.move_to_prev_max_ts()
```
BusyJay, line 166: This is wrong. For example, if there are only one key in the db and iterator is pointing at the only one key, calling prev should return false.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    fn seek(&mut self, key: &Key) -> Result<bool> {
+        self.clear_cur_kv();
+        self.inner.seek(key)
+    }
+
+    fn seek_for_prev(&mut self, key: &Key) -> Result<bool> {
+        if self.inner.seek_for_prev(key)? {
+            self.move_to_prev_max_ts()
```
BusyJay, line 176: This is confusing as `seek` doesn't require moving to max_ts.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn seek_to_last(&mut self) -> Result<bool> {
+        if self.inner.seek_to_last()? {
+            self.move_to_prev_max_ts()
+        } else {
+            Ok(false)
+        }
+    }
+
+    fn valid(&self) -> Result<bool> {
+        self.is_valid.map_or_else(|| self.inner.valid(), |v| Ok(v))
```
BusyJay, line 196: How about `self.is_valid.map_or_else(|| self.inner.valid(), Ok)`

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    fn validate_key(&self, key: &Key) -> Result<()> {
+        self.inner.validate_key(key)
+    }
+
+    fn key(&self) -> &[u8] {
+        // need map_or_else to lazy evaluate the default func, as it will abort when invalid.
+        self.cur_key
+            .as_ref()
+            .map_or_else(|| self.inner.key(), |k| k)
```
BusyJay, line 207: You mean `unwrap_or_else`?

src/storage/raw/raw_mvcc.rs
haojinming: seek_for_prev also need move to max_ts. As seek_for_prev reach the end the user key, which is the min ts written.

src/storage/raw/raw_mvcc.rs
haojinming: ok

src/storage/raw/raw_mvcc.rs
haojinming: In this case, move_to_prev_max_ts actually return Ok(false). when caller call seek_for_prev, move_to_prev_max_ts  is also called once, so engine iterator has been invalid. Then call move_to_prev_max_ts  again will return Ok(false)

src/storage/raw/raw_mvcc.rs
haojinming: good proposal. how about shrinking the vector when capacity is 2 times than the key size?

src/storage/raw/raw_mvcc.rs
haojinming: yes

src/storage/raw/raw_mvcc.rs
BusyJay: There should be a hard limit. For example, it's meaningless to shrink a 512 buffer.

src/storage/raw/raw_mvcc.rs
BusyJay: What if it calls `seek` and then calls `prev`?

haojinming: also works. Caller should check valid() after call prev(), and valid() return false in this case.

src/storage/raw/raw_mvcc.rs
haojinming: rawkv get interface now do not accept timestamp parameter and always seek the lastest one for user.

BusyJay: Since it's a public interface, its behavior should better be well documented.

src/storage/raw/raw_mvcc.rs
haojinming: ok

src/storage/raw/raw_mvcc.rs
haojinming: Is this the proper usage of a iterator? This iterator is exposed to RawStoreInner and it does not have the usage like this.

src/storage/raw/raw_mvcc.rs
haojinming: ok, comment is added to clear this.

src/storage/raw/raw_mvcc.rs
BusyJay: It's not the optimal usage, but It's OK to do so.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,244 @@
...
+// ts is desc encoded after user key, so it's placed the first one for the same user key.
+// Only one-way direction scan is supported. Like `seek` then `next` or `seek_for_prev` then `prev`
+impl<I: Iterator> Iterator for RawMvccIterator<I> {
+    fn next(&mut self) -> Result<bool> {
+        if !self.is_forward {
+            return Err(Error::from(ErrorInner::Other(Box::from(
+                "invalid raw mvcc operation",
+            ))));
+        }
+        let cur_key = self.inner.key().to_owned();
```
BusyJay, line 174: Probably need to ensure it's valid before calling key.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,244 @@
...
+            is_forward: true,
+        }
+    }
+
+    fn update_cur_kv(&mut self) {
+        if let Some(ref mut key) = self.cur_key {
+            key.clear();
+            key.extend_from_slice(self.inner.key());
+            if key.capacity() > key.len() * 2 && key.capacity() > VEC_SHRINK_THRESHOLD {
+                key.shrink_to(key.len() * 2);
```
BusyJay, line 120: How about `shrint_to_fit`?

BusyJay (approved):
Rest LGTM

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,244 @@
...
+            self.update_cur_kv();
+            self.inner.prev()?;
+        } else {
+            self.clear_cur_kv();
+            return Ok(false);
+        }
+        while self.inner.valid()? {
+            // cur_key should not be None here.
+            if is_user_key_eq(self.cur_key.as_ref().unwrap(), self.inner.key()) {
+                self.update_cur_kv();
```
Connor1996, line 154: why need to update cur kv again and again

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,244 @@
...
+        self.is_valid.map_or_else(|| self.inner.valid(), Ok)
+    }
+
+    fn validate_key(&self, key: &Key) -> Result<()> {
+        self.inner.validate_key(key)
+    }
+
+    fn key(&self) -> &[u8] {
+        // need map_or_else to lazy evaluate the default func, as it will abort when invalid.
+        self.cur_key
```
Connor1996, line 232: cur_kv seems unnecessary, why not just use inner.key() and inner.value() directly?

src/storage/raw/raw_mvcc.rs
Connor1996: oh, I was wrong, seems it's necessary.

src/storage/raw/raw_mvcc.rs
```diff
@@ -0,0 +1,244 @@
...
+    is_valid: Option<bool>,
+    is_forward: bool,
+}
+
+fn is_user_key_eq(left: &[u8], right: &[u8]) -> bool {
+    let len = left.len();
+    if len != right.len() {
+        return false;
+    }
+    Key::is_user_key_eq(left, &right[..len - U64_SIZE])
```
Connor1996, line 101: use `truncate_ts_for` to truncate ts

src/storage/raw/store.rs
```diff
@@ -2,6 +2,7 @@
 
 // #[PerformanceCriticalPath]
```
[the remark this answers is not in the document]
haojinming: done

src/storage/raw/raw_mvcc.rs
haojinming: i think it is caller's responsibility to check validition before calling next().

src/storage/raw/raw_mvcc.rs
haojinming: done

haojinming: done

src/storage/raw/raw_mvcc.rs
haojinming: currently return error if caller call `seek` then `prev` or `seek_for_prev` then `next`

Connor1996 (approved):
LGTM

src/storage/raw/raw_mvcc.rs
BusyJay: It's hard. That's why we have features "nortcheck" in a lot of components.

src/storage/raw/raw_mvcc.rs
haojinming: done

PR #12089, opened 2022-03-04
*: add more metrics and logs for in-memory pessimistic locking

sticnarf:

Issue Number: Ref #11452

What's Changed:

```commit-message
It adds logs when proposing locks during transferring leader and moving locks to new split regions.

The metrics about the memory used by pessimistic locks and how many locks fail to be inserted due to
reaching the memory limit are displayed in the grafana.
```

### Related Changes

- Remeber to update https://github.com/pingcap/docs-cn/blob/master/grafana-tikv-dashboard.md

### Check List 

Tests 

- Manual test

The "Lock manager" row is renamed to "Pessimistic Locking":

image

Two more panels are added:

image

image

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3033,6 +3039,7 @@ where
         // to the old region will stay in the original map.
         let region_locks = {
             let mut pessimistic_locks = self.fsm.peer.txn_ext.pessimistic_locks.write();
+            info!("moving {} locks to new regions", pessimistic_locks.len());
```
youjiali1995, line 3042: I think it's not sound without region id...

sticnarf: Sorry, I missed it...

youjiali1995 (approved):

PR #12090, opened 2022-03-04
apiv2 rawkv: logical deletion flag

haojinming:

Issue Number: Ref #11965 

What's Changed:
encode logical deletion flag in apiv2 RawValue, just one bit flag in value meta.

- Unit test

### DIFF
--- a/components/api_version/src/lib.rs
+++ b/components/api_version/src/lib.rs
@@ -24,15 +24,16 @@ pub trait APIVersion: Clone + Copy + 'static + Send + Sync {
     fn decode_raw_value(bytes: &[u8]) -> Result<RawValue<&[u8]>>;
     /// This is equivalent to `decode_raw_value()` but returns the owned user value.
     fn decode_raw_value_owned(mut bytes: Vec<u8>) -> Result<RawValue<Vec<u8>>> {
-        let (len, expire_ts) = {
+        let (len, expire_ts, is_delete) = {
             let raw_value = Self::decode_raw_value(&bytes)?;
-            (raw_value.user_value.len(), raw_value.expire_ts)
+            (raw_value.user_value.len(), raw_value.expire_ts, raw_value.is_delete)
         };
         // The user value are always the first part in encoded bytes.
         bytes.truncate(len);
         Ok(RawValue {
             user_value: bytes,
             expire_ts,
+            is_delete,
         })
     }
     /// Encode the raw value and it's metadata into bytes.
@@ -149,6 +150,8 @@ pub struct RawValue<T: AsRef<[u8]>> {
     pub user_value: T,
     /// The unix timestamp in seconds indicating the point of time that this key will be deleted.
     pub expire_ts: Option<u64>,
+    /// Logical deletion flag in apiv2, should be None in APIV1 and APIV1ttl
+    pub is_delete: Option<bool>,
 }
 
 #[cfg(test)]
@@ -331,8 +334,8 @@ mod tests {
             (vec![1], ApiVersion::V2),
             (vec![1, 2, 3, 4, 5, 6, 7, 1], ApiVersion::V2),
             // Undefined flag.
-            (vec![2], ApiVersion::V2),
-            (vec![1, 2, 3, 4, 5, 6, 7, 8, 2], ApiVersion::V2),
+            (vec![4], ApiVersion::V2),
+            (vec![1, 2, 3, 4, 5, 6, 7, 8, 4], ApiVersion::V2),
         ];
 
         for (bytes, api_version) in cases {
@@ -354,6 +357,10 @@ mod tests {
         encoded_bytes: &[u8],
         api_version: ApiVersion,
     ) {
+        let is_delete = match api_version {
+            ApiVersion::V2 => Some(false),
+            _ => None,
+        };
         match_template_api_version!(
             API,
             match api_version {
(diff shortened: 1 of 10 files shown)


components/api_version/src/lib.rs
```diff
@@ -149,6 +150,8 @@ pub struct RawValue<T: AsRef<[u8]>> {
     pub user_value: T,
     /// The unix timestamp in seconds indicating the point of time that this key will be deleted.
     pub expire_ts: Option<u64>,
+    /// Logical deletion flag in apiv2, should be None in APIV1 and APIV1ttl
+    pub is_delete: Option<bool>,
```
pingyu, line 154: I think "bool" is enough. "None" seems to be unnecessary. Always be "false" in V1 and V1TTL.

components/api_version/src/lib.rs
```diff
@@ -516,4 +525,41 @@ mod tests {
             }
         )
     }
+
+    #[test]
+    fn test_raw_value_encode_delete() {
```
pingyu, line 530: Suggest add unit tests to `test_no_ttl` and `test_ttl`, as test of TTL is quite similar to delete flag.
Or we can change names to `test_no_meta` and `test_meta`.

components/api_version/src/lib.rs
```diff
@@ -149,6 +150,8 @@ pub struct RawValue<T: AsRef<[u8]>> {
     pub user_value: T,
     /// The unix timestamp in seconds indicating the point of time that this key will be deleted.
     pub expire_ts: Option<u64>,
+    /// Logical deletion flag in apiv2, should be None in APIV1 and APIV1ttl
```
pingyu, line 153:
```suggestion
    /// Logical deletion flag in APIV2, should be None in APIV1 and APIV1TTL
```

pingyu (commented):
Three suggestions, PTAL~. Rest LGTM.

components/api_version/src/lib.rs
haojinming: fixed

components/api_version/src/lib.rs
haojinming: fixed

components/api_version/src/lib.rs
haojinming: fixed

components/api_version/src/lib.rs
```diff
@@ -149,6 +150,8 @@ pub struct RawValue<T: AsRef<[u8]>> {
     pub user_value: T,
     /// The unix timestamp in seconds indicating the point of time that this key will be deleted.
     pub expire_ts: Option<u64>,
+    /// Logical deletion flag in APIV2, should be None in APIV1 and APIV1TTL
```
pingyu, line 153:
```suggestion
    /// Logical deletion flag in APIV2, should be `false` in APIV1 and APIV1TTL
```

components/api_version/src/lib.rs
```diff
@@ -311,10 +317,37 @@ mod tests {
...
+                case.4
             );
         }
         for case in &cases {
-            assert_raw_value_encode_decode_identity(case.0, Some(case.1), case.3, ApiVersion::V2);
+            assert_raw_value_encode_decode_identity(case.0, Some(case.1), case.3, ApiVersion::V2, case.4);
+        }
+
+        // (user_value, expire_ts, ecoded_bytes_v2, api_version, is_delete)
+        let positive_cases = vec![
```
pingyu, line 328: En... api_version & is_delete fields are almost the same.

How about like this:
1. fn test_no_meta: test V1, V1TTL and V2 with cases without meta
2. fn test_ttl: test V1TTL and V2 with cases with ttl
3. fn test_meta: test V2 only, with cases has meta as combination of ttl & delete flag.

pingyu (commented):
PTAL. Rest LGTM.

components/api_version/src/lib.rs
haojinming: resolved

pingyu (approved):
Great! LGTM.

BusyJay (approved):

pingyu (approved):
LGTM.

components/api_version/src/lib.rs
```diff
@@ -311,10 +318,46 @@ mod tests {
...
+            );
+        }
+    }
+
+    #[test]
+    fn test_meta() {
+        // (user_value, expire_ts, ecoded_bytes_v2, is_delete)
+        let cases = vec![
+            // only deletion flag.
+            (&b""[..], None, &[2][..], true),
```
andylokandy, line 340: Consider giving some of them a non-empty value.

components/api_version/src/lib.rs
```diff
@@ -353,6 +396,7 @@ mod tests {
         expire_ts: Option<u64>,
         encoded_bytes: &[u8],
         api_version: ApiVersion,
+        is_delete: bool,
```
andylokandy, line 399: Consider moving it two lines above.

components/api_version/src/lib.rs
haojinming: this case is to test logical delete scenario, so it just has one byte(ValueMeta). There are two cases with non-empty value in the case suit.

components/api_version/src/lib.rs
haojinming: addressed

components/api_version/src/lib.rs
```diff
@@ -311,10 +318,46 @@ mod tests {
...
+                Some(case.1),
+                case.3,
+                ApiVersion::V2,
+                false,
+            );
+        }
+    }
+
+    #[test]
+    fn test_meta() {
```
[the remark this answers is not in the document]
haojinming: addressed

andylokandy (approved):

PR #12051, opened 2022-03-01
engine: cut raft-engine to pre-alpha branch

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

### What is changed and how it works?

Issue Number: ref #11119

What's Changed:

Cut to pre-alpha branch, which includes all the major bug fixes but without interface changes in master branch.

Will switch back to master branch by @ztelur as per https://github.com/tikv/raft-engine/issues/165#issuecomment-1042860549.

### Related changes

Mirrors #12029 

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

PR #12095, opened 2022-03-04
engine: upgrade raft engine

ztelur:
Signed-off-by: Randy <ztelur@gmail.com>

What is changed and how it works?
Issue Number: ref https://github.com/tikv/tikv/issues/11119

What's Changed:

- upgrade raft-engine dependency and other library version related to raft-engine
- replace old `ManagedFileBuilder ` with `ManagedFileSystem`, more detail in raft-engine [#165](https://github.com/tikv/raft-engine/issues/165)

 Also refer to  https://github.com/tikv/tikv/pull/12051

Check List
### Tests

Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raft_log_engine/src/engine.rs
+++ b/components/raft_log_engine/src/engine.rs
@@ -2,7 +2,7 @@
 
 use std::fs;
 use std::io::{Read, Result as IoResult, Seek, SeekFrom, Write};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use encryption::{DataKeyManager, DecrypterReader, EncrypterWriter};
@@ -14,10 +14,12 @@ use file_system::{IOOp, IORateLimiter, IOType};
 use kvproto::raft_serverpb::RaftLocalState;
 use raft::eraftpb::Entry;
 use raft_engine::{
-    Command, Engine as RawRaftEngine, Error as RaftEngineError, FileBuilder, LogBatch, MessageExt,
+    env::FileSystem, Command, Engine as RawRaftEngine, Error as RaftEngineError, LogBatch,
+    MessageExt,
 };
 use tikv_util::Either;
 
+use raft_engine::env::{DefaultFileSystem, Handle, WriteExt};
 pub use raft_engine::{Config as RaftEngineConfig, RecoveryMode};
 
 #[derive(Clone)]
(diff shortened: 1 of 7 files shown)


ztelur:
cc @tabokie

components/raft_log_engine/src/engine.rs
```diff
@@ -14,10 +14,12 @@ use file_system::{IOOp, IORateLimiter, IOType};
 use kvproto::raft_serverpb::RaftLocalState;
 use raft::eraftpb::Entry;
 use raft_engine::{
-    Command, Engine as RawRaftEngine, Error as RaftEngineError, FileBuilder, LogBatch, MessageExt,
+    env::FileSystem, Command, Engine as RawRaftEngine, Error as RaftEngineError, LogBatch,
+    MessageExt,
 };
 use tikv_util::Either;
 
+use raft_engine::env::{DefaultFileSystem, Handle, WriteExt};
```
tabokie, line 22:
```suggestion
use tikv_util::Either;
use raft_engine::env::{DefaultFileSystem, Handle, WriteExt};

```

components/raft_log_engine/src/engine.rs
```diff
@@ -89,74 +97,136 @@ impl<W: Seek + Write> Write for ManagedWriter<W> {
...
         }
     }
 }
 
-impl FileBuilder for ManagedFileBuilder {
-    type Reader<R: Seek + Read + Send> = ManagedReader<R>;
-    type Writer<W: Seek + Write + Send> = ManagedWriter<W>;
+struct ManagedHandle {
+    path: PathBuf,
+    base: Arc<<DefaultFileSystem as FileSystem>::Handle>,
```
tabokie, line 144: Is the `Arc` necessary?

components/raft_log_engine/src/engine.rs
```diff
@@ -89,74 +97,136 @@ impl<W: Seek + Write> Write for ManagedWriter<W> {
...
+
+    fn new_reader(&self, handle: Arc<Self::Handle>) -> IoResult<Self::Reader> {
+        let base_reader = self
+            .base_level_file_system
+            .new_reader(handle.base.clone())?;
         if let Some(ref key_manager) = self.key_manager {
             Ok(ManagedReader {
-                inner: Either::Right(key_manager.open_file_with_reader(path, reader)?),
+                inner: Either::Right(
+                    key_manager.open_file_with_reader(handle.path.clone(), base_reader)?,
```
tabokie, line 183: Maybe no need to clone?
```suggestion
                    key_manager.open_file_with_reader(&handle.path, base_reader)?,
```

BusyJay:
Will raft engine be published to crates.io? So we don't need to query github instead, which is quite slow.

tabokie:
> Will raft engine be published to crates.io? So we don't need to query github instead, which is quite slow.

@BusyJay Yes, I'll publish 0.1.0 and update tikv dependency just before 6.0.0 release.

components/raft_log_engine/src/engine.rs
ztelur: Yes. it will also be held by the base reader or writer
``` rust
fn new_reader(&self, handle: Arc<Self::Handle>) -> IoResult<Self::Reader> {
    Ok(LogFile::new(handle))
}
```

tabokie (approved):
LGTM

Connor1996 (approved):
LGTM

PR #12096, opened 2022-03-04
*: Fix possible undefined behavior for transmuting vec

### LINKED ISSUE #12070: UB in TsSet conversions
https://github.com/tikv/tikv/blob/8b86c063c21f111a2fb0d4fcc4a328c3df7bd6f3/components/txn_types/src/timestamp.rs#L149
https://github.com/tikv/tikv/blob/8b86c063c21f111a2fb0d4fcc4a328c3df7bd6f3/components/txn_types/src/timestamp.rs#L156

This unsafe cast is UB because Vec\<u64\> and Vec\<TimeStamp\> don't have the same layout in general. For example a Rust implementation using profile-guided optimization might put the vec's fields in a different order for the two types, due to different access patterns across all the places those 2 different instantiations are used (e.g. one might store the len first while the other might store the capacity first, because len() is called more often on one than the other).

Connor1996:

Issue Number: Close #12070

What's Changed:

Recreate a vec from raw parts instead of transmuting memory directly.

### Related changes

### Check List 

Tests 

- No code

### Release note 

```release-note
Fix possible undefined behavior for transmuting vec
```

### DIFF
--- a/components/concurrency_manager/src/lib.rs
+++ b/components/concurrency_manager/src/lib.rs
@@ -10,8 +10,6 @@
 //! In order to mutate the lock of a key stored in the lock table, it needs
 //! to be locked first using `lock_key` or `lock_keys`.
 
-#![feature(vec_into_raw_parts)]
-
 use fail::fail_point;
 
 mod key_handle;
@@ -80,8 +78,7 @@ impl ConcurrencyManager {
         for (index, key) in keys_with_index {
             result[index] = MaybeUninit::new(self.lock_table.lock_key(key).await);
         }
-        let (ptr, len, cap) = result.into_raw_parts();
-        unsafe { Vec::from_raw_parts(ptr as _, len, cap) }
+        unsafe { tikv_util::memory::vec_transmute(result) }
     }
 
     /// Checks if there is a memory lock of the key which blocks the read.
--- a/components/concurrency_manager/src/lib.rs
+++ b/components/concurrency_manager/src/lib.rs
@@ -10,6 +10,8 @@
 //! In order to mutate the lock of a key stored in the lock table, it needs
 //! to be locked first using `lock_key` or `lock_keys`.
 
+#![feature(vec_into_raw_parts)]
+
 use fail::fail_point;
 
 mod key_handle;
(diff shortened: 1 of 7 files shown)


sticnarf (approved):
LGTM

components/concurrency_manager/src/lib.rs
```diff
@@ -78,11 +80,8 @@ impl ConcurrencyManager {
         for (index, key) in keys_with_index {
             result[index] = MaybeUninit::new(self.lock_table.lock_key(key).await);
         }
-        #[allow(clippy::unsound_collection_transmute)]
-        #[allow(clippy::transmute_undefined_repr)]
-        unsafe {
-            mem::transmute(result)
-        }
+        let (ptr, len, cap) = result.into_raw_parts();
```
BusyJay, line 83: How about making it a util function?

components/tikv_util/src/memory.rs
```diff
@@ -8,6 +8,13 @@ use kvproto::{
 };
 use std::mem;
 
+// Transmute vec with types of same memory layout.
+pub unsafe fn vec_transmute<F, T>(from: Vec<F>) -> Vec<T> {
+    debug_assert!(mem::size_of::<F>() == mem::size_of::<T>());
```
sticnarf, line 13: Better to also assert about the alignment.

And `static_assertions` may help check the error in compile time.

BusyJay, line 13: May also check for align.

components/tikv_util/src/memory.rs
```diff
@@ -8,6 +8,13 @@ use kvproto::{
 };
 use std::mem;
 
+// Transmute vec with types of same memory layout.
+pub unsafe fn vec_transmute<F, T>(from: Vec<F>) -> Vec<T> {
```
BusyJay, line 12: How about marking it `inline`?

BusyJay (approved):

components/tikv_util/src/memory.rs
Connor1996: can't, rust complains about it. Same goes for `static_assertions:: assert_eq_size`.
```
error[E0401]: can't use generic parameters from outer function
  --> components/tikv_util/src/memory.rs:14:56
   |
13 | pub unsafe fn vec_transmute<F, T>(from: Vec<F>) -> Vec<T> {
   |                             - type parameter from outer function
14 |     static_assertions::const_assert_eq!(mem::size_of::<F>(), mem::size_of::<T>());
   |                                                        ^ use of generic parameter from outer function
```
Seems no perfect solution for this, see https://stackoverflow.com/questions/30330519/compile-time-generic-type-size-check

sticnarf: OK, I see.

PR #12097, opened 2022-03-04
txn: support read-consistency read with tso checking

cfzjywxk:
### What is changed and how does it work?

Try to support read-consistency read processing with tso checking to optimize the latency.

Ref document: https://github.com/pingcap/tidb/pull/32806
Ref kvproto change: https://github.com/pingcap/kvproto/pull/876

What's Changed:
- If the isolation field in a read request is `RcCheckTs`,  return write conflict error directly to the client. The client may retry the whole SQL statement with a new `for_update_ts` fetched from global tso allocator if no data package has been responsed to the MySQL client.

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
Support read-consistency read with tso checking.
```

### DIFF
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -395,7 +395,7 @@ impl Lock {
         key: &Key,
         ts: TimeStamp,
         bypass_locks: &TsSet,
-        iso_level: kvproto::kvrpcpb::IsolationLevel,
+        iso_level: IsolationLevel,
     ) -> Result<()> {
         match iso_level {
             IsolationLevel::Si => Lock::check_ts_conflict_si(lock, key, ts, bypass_locks),
@@ -857,6 +857,50 @@ mod tests {
             IsolationLevel::Si,
         )
         .unwrap_err();
+    }
+
+    #[test]
+    fn test_check_ts_conflict_rc_check_ts() {
+        let k1 = Key::from_raw(b"k1");
+        let mut lock = Lock::new(
+            LockType::Put,
+            vec![],
+            100.into(),
+            3,
+            None,
+            100.into(),
+            1,
+            TimeStamp::zero(),
+        );
+
+        let empty = Default::default();
+
+        // Ignore locks that occurs in the `bypass_locks` set.
+        Lock::check_ts_conflict_rc_check_ts(
+            Cow::Borrowed(&lock),
+            &k1,
+            50.into(),
+            &TsSet::from_u64s(vec![100]),
+        )
+        .unwrap();
+
+        // Ignore locks if the lock type are Pessimistic or Lock.
+        lock.lock_type = LockType::Pessimistic;
+        Lock::check_ts_conflict_rc_check_ts(Cow::Borrowed(&lock), &k1, 50.into(), &empty).unwrap();
+        lock.lock_type = LockType::Lock;
+        Lock::check_ts_conflict_rc_check_ts(Cow::Borrowed(&lock), &k1, 50.into(), &empty).unwrap();
+
+        // Report error even if read ts is less than the lock version.
+        lock.lock_type = LockType::Put;
+        Lock::check_ts_conflict_rc_check_ts(Cow::Borrowed(&lock), &k1, 50.into(), &empty)
+            .unwrap_err();
+        Lock::check_ts_conflict_rc_check_ts(Cow::Borrowed(&lock), &k1, 110.into(), &empty)
+            .unwrap_err();
+
+        // Report error if for other lock types.
+        lock.lock_type = LockType::Delete;
+        Lock::check_ts_conflict_rc_check_ts(Cow::Borrowed(&lock), &k1, 50.into(), &empty)
+            .unwrap_err();
     }
 
     #[test]
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -4,7 +4,7 @@ use crate::timestamp::{TimeStamp, TsSet};
 use crate::types::{Key, Mutation, Value, SHORT_VALUE_PREFIX};
 use crate::{Error, ErrorInner, Result};
 use byteorder::ReadBytesExt;
-use kvproto::kvrpcpb::{LockInfo, Op};
+use kvproto::kvrpcpb::{IsolationLevel, LockInfo, Op};
 use std::{borrow::Cow, mem::size_of};
 use tikv_util::codec::bytes::{self, BytesEncoder};
 use tikv_util::codec::number::{self, NumberEncoder, MAX_VAR_I64_LEN, MAX_VAR_U64_LEN};
@@ -326,7 +326,7 @@ impl Lock {
     }
 
     /// Checks whether the lock conflicts with the given `ts`. If `ts == TimeStamp::max()`, the primary lock will be ignored.
-    pub fn check_ts_conflict(
+    fn check_ts_conflict_si(
         lock: Cow<'_, Self>,
         key: &Key,
         ts: TimeStamp,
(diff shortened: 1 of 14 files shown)


youjiali1995:
Please resolve conflicts.

src/storage/mvcc/reader/scanner/backward.rs
```diff
@@ -266,6 +264,16 @@ impl<S: Snapshot> BackwardKvScanner<S> {
                     if self.met_newer_ts_data == NewerTsCheckState::NotMetYet {
                         self.met_newer_ts_data = NewerTsCheckState::Met;
                     }
+                    if self.cfg.isolation_level == IsolationLevel::RcCheckTs {
```
you06, line 267: Is it ok to skip this record if the `WriteType` is `Lock` or `Rollback`?

src/storage/mvcc/reader/scanner/forward.rs
```diff
@@ -265,7 +268,9 @@ impl<S: Snapshot, P: ScanPolicy<S>> ForwardScanner<S, P> {
             };
 
             if has_lock {
-                if self.met_newer_ts_data == NewerTsCheckState::NotMetYet {
+                if self.met_newer_ts_data == NewerTsCheckState::NotMetYet
+                    && self.cfg.isolation_level == IsolationLevel::Si
+                {
```
you06, line 273: Why not set `met_newer_ts_data` for `IsolationLevel::RcCheckTs`?

src/storage/mvcc/reader/scanner/forward.rs
```diff
@@ -328,13 +333,26 @@ impl<S: Snapshot, P: ScanPolicy<S>> ForwardScanner<S, P> {
...
+                if key_commit_ts <= self.cfg.ts {
                     // Founded, don't need to seek again.
                     needs_seek = false;
                     break;
                 } else if self.met_newer_ts_data == NewerTsCheckState::NotMetYet {
                     self.met_newer_ts_data = NewerTsCheckState::Met;
                 }
+
+                // Report error if there's a more recent version if the isolation level is RcCheckTs.
+                if self.cfg.isolation_level == IsolationLevel::RcCheckTs {
```
you06, line 346: Ditto, may skip `WriteType::Lock` and `WriteType::Rollback`.

src/storage/mvcc/reader/scanner/forward.rs
cfzjywxk: The scanner does not rely on this field to check the latest data so it's not set.

src/storage/mvcc/reader/scanner/backward.rs
cfzjywxk: I think so, I'll add more tests to cover this path.

src/storage/mvcc/reader/scanner/forward.rs
you06: It doesn't matter in `RcCheckTs` case, and pointGetter also does not set it, but it's set in the backward scan 🤔.

src/storage/mvcc/reader/scanner/backward.rs
cfzjywxk: It's a bit complext to check if there's a newer version which is not `Lock` and `Rollback`. As the feature is recommanded to be used where there's less read-write conflict, maybe it's ok to return error and let client retry directly.

you06: May left a TODO here?

src/storage/mvcc/reader/scanner/forward.rs
cfzjywxk: I'll change the backward scanner not to set it to keep.

sticnarf: I think it more sematically correct to set it to `NewerTsCheckState::Met` here no matter the isolation level is...

src/storage/mvcc/reader/scanner/backward.rs
cfzjywxk: Ok.

src/storage/mvcc/reader/scanner/forward.rs
cfzjywxk: I've changed the logic not to depend on the `NewerTsCheckState` related behaviours and the default behaviours of it would not be affected.

src/storage/mod.rs
```diff
@@ -2401,7 +2402,13 @@ fn prepare_snap_ctx<'a>(
                 .read_key_check(key, |lock| {
                     // No need to check access_locks because they are committed which means they
                     // can't be in memory lock table.
-                    Lock::check_ts_conflict(Cow::Borrowed(lock), key, start_ts, bypass_locks)
+                    Lock::check_ts_conflict(
```
TonsnakeLin, line 2405: Here has a "if" block which is "if isolation_level == IsolationLevel::Si". Should it be modified as "if isolation_level != IsolationLevel::RC"

sticnarf: Yes, should be changed to `matches!(isolation_level, IsolationLevel::Si | IsolationLevel::RcCheckTs)`

src/storage/mvcc/reader/scanner/backward.rs
```diff
@@ -266,6 +265,18 @@ impl<S: Snapshot> BackwardKvScanner<S> {
...
```
TonsnakeLin, line 268: Does the self.cfg.isolation_level come from tidb or tikv's config file?

TonsnakeLin, line 268: Does the self.cfg.isolation_level come from tidb or tikv's config file?

cfzjywxk: No it comes from the rpc requst field `isolation_level` which is set by read requests in https://github.com/pingcap/tidb/pull/32922.

cfzjywxk:
@youjiali1995 @sticnarf @you06 @MyonKeminta 
PTAL

src/storage/mod.rs
```diff
@@ -1033,6 +1033,7 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
                                 key,
                                 start_ts,
                                 &bypass_locks,
+                                ctx.get_isolation_level(),
```
MyonKeminta, line 1036: I'm afraid there might be many code like `if ctx.get_isolation_level() == IsolationLevel::xxx` that needs to be updated bot missed. Does line 1027 needs to be updated?

cfzjywxk: Yes, thankds for reminding. The entry of `check_ts_conflict` is unified as two more isolation variables `RC` and `RcCheckTS` are introduced for different purposes.

you06: `need_check_locks_in_replica_read` needs to return true when the isolation level is `IsolationLevel::RcCheckTs`.

components/txn_types/src/lock.rs
```diff
@@ -363,6 +363,49 @@ impl Lock {
...
+        ts: TimeStamp,
+        bypass_locks: &TsSet,
+        iso_level: kvproto::kvrpcpb::IsolationLevel,
+    ) -> Result<()> {
+        match iso_level {
+            IsolationLevel::Si => Lock::check_ts_conflict_si(lock, key, ts, bypass_locks),
+            IsolationLevel::RcCheckTs => {
+                Lock::check_ts_conflict_rc_read(lock, key, ts, bypass_locks)
+            }
+            _ => Ok(()),
```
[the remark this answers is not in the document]
you06: The performance of `IsolationLevel::RcCheckTs` is certainly poorer than `IsolationLevel::RC`, however, TiDB's RC, a.k.a. read consistency, read data by `IsolationLevel::Si`. So there will be no regression on this check.

src/storage/mod.rs
cfzjywxk: The isolation level filed is not passed by the read index request yet and the `RcCheckTS` read is not compatibile with replica read. Maybe we just document about it and solve the compatible issue next?

components/txn_types/src/lock.rs
```diff
@@ -363,6 +363,49 @@ impl Lock {
...
+        key: &Key,
+        ts: TimeStamp,
+        bypass_locks: &TsSet,
+        iso_level: kvproto::kvrpcpb::IsolationLevel,
+    ) -> Result<()> {
+        match iso_level {
+            IsolationLevel::Si => Lock::check_ts_conflict_si(lock, key, ts, bypass_locks),
+            IsolationLevel::RcCheckTs => {
+                Lock::check_ts_conflict_rc_read(lock, key, ts, bypass_locks)
+            }
```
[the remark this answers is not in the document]
cfzjywxk: The format changes it to a new line, I guess the original line is a bit long.

src/storage/mod.rs
cfzjywxk: I'll add the check in the tidb side so these two would not be used together.Maybe it's not necessary to support it.

src/coprocessor/endpoint.rs
```diff
@@ -112,7 +112,10 @@ impl<E: Engine> Endpoint<E> {
         if !req_ctx.context.get_stale_read() {
             self.concurrency_manager.update_max_ts(start_ts);
         }
-        if req_ctx.context.get_isolation_level() == IsolationLevel::Si {
+        if matches!(
+            req_ctx.context.get_isolation_level(),
+            IsolationLevel::Si | IsolationLevel::RcCheckTs
```
MyonKeminta, line 117: Can these kinds of conditions be wrapped in a utility function like `is_at_least_si` or `requires_si` so that we will be less likely to forgot to miss updating the condition if we add one more `IsolationLevel` choice in the future?

sticnarf (commented):
The rest lgtm

src/storage/mvcc/reader/point_getter.rs
```diff
@@ -255,7 +255,9 @@ impl<S: Snapshot> PointGetter<S> {
         let mut use_near_seek = false;
         let mut seek_key = user_key.clone();
 
-        if self.met_newer_ts_data == NewerTsCheckState::NotMetYet {
+        if self.met_newer_ts_data == NewerTsCheckState::NotMetYet
```
youjiali1995, line 258: `NewerTsCheckState` helps a lot for  this feature.😄

components/txn_types/src/lock.rs
```diff
@@ -363,6 +363,49 @@ impl Lock {
         )))
     }
 
+    // Check if lock could be bypassed for isolation level `RcCheckTs`.
+    fn check_ts_conflict_rc_check_ts(
```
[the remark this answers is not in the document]
cfzjywxk: A test case is added for it.

youjiali1995 (approved):

sticnarf (approved):
LGTM

cfzjywxk:
@youjiali1995 @sticnarf 
Need help merge it, thx

Issue #12020, opened 2022-02-24
supporting using more efficient format for put/delete requests


BusyJay:
### Is your feature request related to a problem? Please describe:
When using single node, loading a lot of key values shows that about 30% of raftstore CPU is used in propose_normal. They are freeing memory and encoding RaftCmdRequest. This makes the commit duration not match the append duration even for single node.
![image](https://user-images.githubusercontent.com/1701473/155519583-cf9d53f6-314c-431c-afc9-479edbae775e.png)

When using protobuf message, every key and value is stored in a separate byte string and every operation is stored in a request. So it will waste a lot of CPU in loops and freeing small byte slice.

### Describe the feature you'd like:
Instead of just using protobuf message, we can store all key/value pairs in one big slice. So when proposing the message, only one byte slice needs to be freed and copied.

A simple example can be adding a new sub messages in `RaftCmdRequest`.
```
message BlobRequest {
    bytes  mutations = 1;
}

message RaftCmdRequest {
    ...
    BlobRequest blob_request = 5;
}
```

@coocood has done a prototype with his own codec that translate into prewrite/commit format. I also done a prototype that using a simple format as following.

Only put and delete need to be encoded in `BlobRequest`, all other operations still use the old format. So it doesn't require change a lot of codes. And the format of `mutations` can be:
```
('p'|'d') key_len key_content (value_len value_content)
```

Merging two requests is as simple as copying byte slice.

My prototype shows that for single node, using new codecs can make the cost of propose_normal ellageiable. And the QPS is increased from 40k to 48k.
![image](https://user-images.githubusercontent.com/1701473/155519657-42d5dec3-26b3-403d-bd6d-94847badd143.png)

A more aggressive version is store the rocksdb writebatch directly in mutations, this can remove further one copy when translating mutations to write batch in apply threads.

### Describe alternatives you've considered:
No.

youjiali1995:
> A more aggressive version is store the rocksdb writebatch directly in mutations, this can remove further one copy when translating mutations to write batch in apply threads.

I remembered @Little-Wallace did it in 2020 and it brings much improvement.

BusyJay:
Yes, see https://github.com/tikv/tikv/issues/7994. This issue is less aggressive and need less changes.

hnes:
@tonyxuqqi  OK, I will try to implement this feature.

BusyJay:
You may also discuss it for potential compatibility. /cc @pingyu

hnes:
[...]

PR #12099, opened 2022-03-05
rawkv: causal timestamp

pingyu:

Issue Number: ref #11965

What's Changed:

* Add `causal_ts` component for causal timestamp generation & management.
* Implement causal timestamp provider from "batch" TSO from PD (see `causal_ts::BatchTsoProvider`).
* Append timestamp to keys of RawKV in `QueryObserver.pre_propose_query`.
* Setup causal timestamp observer when API V2 enabled (comment out)

Refer to RFC:
1. https://github.com/pingyu/tikv-rfcs/blob/rfc_api_v2_pd/text/0069-api-v2.md (https://github.com/tikv/rfcs/pull/79)
2. https://github.com/pingyu/tikv-rfcs/blob/rawkv-cdc/text/0083-rawkv-cross-cluster-replication.md (https://github.com/tikv/rfcs/pull/86)

```commit-message
* Add `causal_ts` component for causal timestamp generation & management.
* Implement causal timestamp provider from "batch" TSO from PD (see `causal_ts::BatchTsoProvider`).
* Append timestamp to keys of RawKV in `QueryObserver.pre_propose_query`.
* Setup causal timestamp observer when API V2 enabled (comment out)
```

### Related changes
No.

### Check List 

Tests 

- Unit test

Side effects
No.

### Release note 

```release-note
New encoding for RawKV API V2.
```


components/api_version/src/api_v2.rs
```diff
@@ -151,16 +153,27 @@ impl APIVersion for APIV2 {
     }
 }
 
+impl APIV2 {
+    pub fn append_ts_on_encoded_bytes(encoded_bytes: &mut Vec<u8>, ts: TimeStamp) {
```
andylokandy, line 157: why not to use `Key::append_ts`?

components/causal_ts/src/hlc.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    fn advance(&self, to: TimeStamp) -> Result<()> {
+        self.hlc.advance(to);
+        Ok(())
+    }
+}
+
+/// A causal timestamp provider by HLC for TEST.
+#[derive(Default)]
+pub struct TestHlcProvider {
```
andylokandy, line 138: Why not use `Hlc` directly?

components/causal_ts/src/hlc.rs
```diff
@@ -0,0 +1,159 @@
...
+    hlc: Arc<Hlc>,
+    pd_client: Arc<dyn PdClient>,
+    tso_worker: Worker,
+
+    /// Interval of Physical Clock refresh from TSO.
+    /// Smaller interval will get more accurate metrics, but more pressure on PD.
+    /// Can be Duration::ZERO, which means never refresh from TSO. For test purpose ONLY.
+    tso_refresh_interval: Duration,
+
+    initialized: AtomicBool,
```
andylokandy, line 67: I suggest initializing HLC in `HlcProvider::new_opt` so that `HlcProvider` must be initialized.

components/api_version/src/api_v2.rs
pingyu: `Key::append_ts` need moving a `Vec<u8>` in and out, and is not convenient to use. E.g., if using `Key::append_ts`, codes with `append_ts_on_encoded_bytes ` would be as following:
```
req.mut_put().set_key(APIV2::append_ts_on_encoded_bytes(req.mut_put().take_key(), ts))
```

components/api_version/src/api_v2.rs
```diff
@@ -151,16 +153,27 @@ impl APIVersion for APIV2 {
...
+        debug_assert!(verify_encoded_bytes(encoded_bytes, false));
+        encoded_bytes.encode_u64_desc(ts.into_inner()).unwrap();
+    }
+}
+
 #[inline]
-fn verify_encoded_key(encoded_key: &Key, with_ts: bool) -> bool {
-    let mut encoded_bytes = &encoded_key.as_encoded()[..];
+fn verify_encoded_bytes(mut encoded_bytes: &[u8], with_ts: bool) -> bool {
```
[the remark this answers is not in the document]
pingyu: Required by `append_ts_on_encoded_bytes`, which accepts `Vec<u8>` argument, other than `Key`.

src/storage/mod.rs
```diff
@@ -1634,7 +1643,40 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
+                for req in requests {
+                    debug_assert!(matches!(req.get_cmd_type(), CmdType::Put));
+                    let ts = causal_ts_provider.get_ts().unwrap(); // `get_ts()` must not fail.
+                    api_version::APIV2::append_ts_on_encoded_bytes(req.mut_put().mut_key(), ts);
+
+                    debug!(
+                        "raw_write::pre_propose_cb";
+                        "ts" => ts,
+                        "key" => &log_wrappers::Value::key(req.get_put().get_key()),
+                        // "value" => &log_wrappers::Value::value(req.get_put().get_value()),
```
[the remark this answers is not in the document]
pingyu: En... value is too big, and make log hard to read. Let me remove the line comment out.

components/causal_ts/src/hlc.rs
pingyu: En... then `Hlc` should implement `CausalTsProvider` trait, which seems to be a little strange as it is also a member of `HlcProvider`.
Becides, `TestHlcProvider` is for test purpose, and I think there will be more methods in the future.

components/causal_ts/src/hlc.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    // self.next() == `to`
+    #[inline]
+    fn advance(&self, to: TimeStamp) {
+        let before = self.0.fetch_max(to.into_inner(), Ordering::Acquire).into();
+        debug!("Hlc::advance"; "before" => ?before, "after" => ?std::cmp::max(before, to));
+    }
+}
+
+const TSO_REFRESH_INTERVAL: Duration = Duration::from_millis(500);
```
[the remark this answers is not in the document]
pingyu: Yes, `new_opt()` can use to configure the interval. `new()` is just for convenience, as 500ms will be enough for most circumstance.

components/causal_ts/src/hlc.rs
pingyu: OK. The `init()` is an async method, as it invoking `pd_client.get_tso()`. Then I will use `block_on(pd_client.get_tso())` in `new_opt()`.

components/causal_ts/src/hlc.rs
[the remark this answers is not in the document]
pingyu: I forget to add a config item...

pingyu:
@andylokandy Comments are addressed or replied. PTAL, thanks~

pingyu:
> /cc @BusyJay

PTAL, thanks~

andylokandy (approved):

src/config.rs
```diff
@@ -2299,6 +2299,10 @@ pub struct CdcConfig {
     #[doc(hidden)]
     #[serde(skip_serializing)]
     pub old_value_cache_size: usize,
+    /// `causal_ts_refresh_interval` is used for causal timestamp of RawKV CDC, to balance metrics precision & PD load.
+    /// TODO: make doc visible when RawKV CDC feature enabled.
+    #[doc(hidden)]
+    pub causal_ts_refresh_interval: ReadableDuration,
```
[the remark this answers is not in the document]
pingyu: I have no idea...
@overvenus @zz-jason PTAL, How do you think ?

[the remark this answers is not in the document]
overvenus: Better to put it in its own submodule.

components/causal_ts/src/observer.rs
```diff
@@ -0,0 +1,286 @@
...
+                    if matches!(req.get_cmd_type(), CmdType::Put)
+                        && matches!(APIV2::parse_key_mode(key_slice), KeyMode::Raw)
+                    {
+                        debug_assert!({
+                            // verify key encoding
+                            let key = Key::from_encoded_slice(key_slice);
+                            APIV2::decode_raw_key_owned(key, true).is_ok()
+                        });
+                        let ts = Key::decode_ts_from(key_slice).unwrap();
+                        self.causal_manager.update_max_ts(batch.region_id, ts);
```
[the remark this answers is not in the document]
pingyu: Every apply in a flush will `update_max_ts` only once, other that each key. Please see the `continue 'outer` in inner loops.
So there should be minor impact on performance.

pingyu:
@andylokandy All comments are addressed or replied. PTAL, thanks~

pingyu:
> /cc @BusyJay @Connor1996 @andylokandy

Refactor according to new proposal as following:
1. Causal timestamp acquiring by "batch" TSO from PD (see `causal_ts::BatchTsoProvider`).
2. Append timestamp to keys of RawKV in `QueryObserver.pre_propose_query`.

PTAL, thanks~

components/causal_ts/src/tso.rs
```diff
@@ -0,0 +1,292 @@
...
+    fn get_ts(&self) -> Result<TimeStamp> {
+        let batch = self.batch.read();
+        match batch.pop() {
+            Some(ts) => {
+                trace!("BatchTsoProvider::get_ts: {:?}", ts);
+                Ok(ts)
+            }
+            None => {
+                error!("BatchTsoProvider::get_ts, batch used up"; "batch.size" => batch.size);
+                Err(Error::TsoBatchUsedUp(batch.size))
```
Connor1996, line 211: when it's used up, renew immediately?

components/api_version/src/api_v2.rs
```diff
@@ -147,7 +147,9 @@ impl APIVersion for APIV2 {
         debug_assert_eq!(Self::parse_key_mode(&user_key), KeyMode::Raw);
```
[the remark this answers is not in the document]
pingyu: It will be used in next PR #12128.
Please see [here](https://github.com/pingyu/tikv/blob/e922d6e2fea54969c336169669b104edf4af5fd9/src/storage/mod.rs#L1483).

components/causal_ts/src/tso.rs
pingyu: The batch size is expected to be always big enough, and used up should not happen.
The default initial batch size is set according to benchmark.
Moreover, the initial batch size is configurable, so in case there is a rare scenario, user can still set a bigger initial batch size to avoid the used up.

Renew immediately may need carefully handle burst requests in an instant, to avoid impacting the TSO service, and will make the QPS slow down. Setting to a bigger initial batch size would be a better solution.

Connor1996: why does renewing immediately impact the TSO service? tso service is light-weight which can serve high enough QPS.
And if you don't renew, the request may retry which slows down QPS as well.

components/causal_ts/src/observer.rs
```diff
@@ -0,0 +1,174 @@
...
+
+/// CausalObserver append timestamp for RawKV V2 data,
+/// and invoke causal_ts_provider.flush() on specified event, e.g. leader transfer, snapshot apply.
+/// Should be used ONLY when API v2 is enabled.
+#[derive(Clone)]
+pub struct CausalObserver {
+    causal_ts_provider: Arc<dyn CausalTsProvider>,
+}
+
+// Causality is most important and should be done first.
```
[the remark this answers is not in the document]
pingyu: I change comment to "Causal observer's priority should be higher than all other observers, to avoid being bypassed.". Is it ok ?

components/causal_ts/src/tso.rs
pingyu: En...my main idea here is that, the batch size should be always enough. If it is not (it should be very rare), user should adjust initial batch size in configuration. And also maybe we need to optimize the batch size algorithm.
So I think we don't need the renew here.

pingyu: Batch renew on used up is added. Also add some metrics, for further analysis & optimization.
PTAL, thanks~

components/causal_ts/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub use observer::*;
+
+use crate::errors::Result;
+use txn_types::TimeStamp;
+
+/// Trait of causal timestamp provider.
+pub trait CausalTsProvider: Send + Sync {
+    /// Get a new timestamp.
+    fn get_ts(&self) -> Result<TimeStamp>;
```
BusyJay, line 22: Perhaps it can accept an additional API to indicate that next returned TS should not be less than given value. So the existing batch doesn't have to be abandoned.

components/causal_ts/src/tso.rs
```diff
@@ -0,0 +1,349 @@
...
+                    None => {
+                        warn!("BatchTsoProvider::get_ts, batch used up"; "batch.size" => batch.size, "retries" => retries);
+                    }
+                }
+            }
+
+            if retries >= GET_TS_MAX_RETRY {
+                break;
+            }
+            if let Err(err) = block_on(self.renew_tso_batch(false, TSO_BATCH_RENEW_FOR_USED_UP)) {
```
BusyJay, line 258: I suggest to use a channel, so that multiple request to renew tso can be batched up.

components/tikv_kv/src/lib.rs
```diff
@@ -93,6 +94,100 @@ impl Modify {
     }
 }
 
+impl From<Modify> for raft_cmdpb::Request {
```
BusyJay, line 97: Why move it here?

pingyu: I'm sorry that I am lazy here. I will revert them.

_`Modify` to `Request` is used by the codes in this PR before. But I do not revert them because they will be used in the coming PR. `components/tikv_kv/src/rocksdb_engine.rs` will use it to call "pre_propose_query" observers for uni tests [here](https://github.com/tikv/tikv/pull/12128/files#diff-10c2f3682d5777d8dca294378acff71faa16589fabba789532041ee3db10cad7R211)._

components/causal_ts/src/lib.rs
pingyu: You mean the `flush()` method ?
The only scenario invokes `flush()` now is leader transfer. But we do not know the TS of the original leader.

BusyJay: The approach I think of is after transferring leader, the region will get a latest tso without clearing current batch, and then will use that tso as parameter to require tso for requests. So that regions without leader change can still using the existing batches, and with leader changes will use the fresher tso.

It's an improvement and not required in this PR.

pingyu: Got it ~
Then can we implement [batching up](https://github.com/tikv/tikv/pull/12099#discussion_r832135245) with this improvement in another PR ?

Connor1996 (commented):
rest LGTM

pingyu:
Fix CI error by following changes:
1. Move `test_batch_tso_provider` from `tests/integrations` to `tso.rs`. The individual integration module will cause `check-bins.py` error.
2. Fix unit test failure of `test_engine_builder` by handle `TEST_ENGINE_CFS` in `From<Request> for Modify`, and leave comments for this conversion function that it is for test purpose.

components/causal_ts/src/observer.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::sync::Arc;
+
+use crate::CausalTsProvider;
+
+/// CausalObserver appends timestamp for RawKV V2 data,
+/// and invoke causal_ts_provider.flush() on specified event, e.g. leader transfer, snapshot apply.
+/// Should be used ONLY when API v2 is enabled.
+#[derive(Clone)]
+pub struct CausalObserver {
+    causal_ts_provider: Arc<dyn CausalTsProvider>,
```
BusyJay, line 21: How about using generic?

components/causal_ts/src/observer.rs
```diff
@@ -0,0 +1,144 @@
...
+            r.get_cmd_type() == CmdType::Put
+                && APIV2::parse_key_mode(r.get_put().get_key()) == KeyMode::Raw
+        }) {
+            if ts.is_none() {
+                ts = Some(self.causal_ts_provider.get_ts().map_err(|err| {
+                    coprocessor::Error::Other(box_err!("Get causal timestamp error: {:?}", err))
+                })?);
+            }
+
+            APIV2::append_ts_on_encoded_bytes(req.mut_put().mut_key(), ts.unwrap());
```
BusyJay, line 63: These actually breaks the optimization mentioned in #12020. We may need to find a way to add ts for a batch instead of every key in the future.

components/server/Cargo.toml
```diff
@@ -30,6 +30,7 @@ backup = { path = "../backup", default-features = false }
 cdc = { path = "../cdc", default-features = false }
 chrono = "0.4"
 tempfile = "3.0"
+causal_ts = { path = "../causal_ts", default-features = false }
```
BusyJay, line 33: Why set default-features explicitly?

components/test_raftstore/src/pd.rs
```diff
@@ -1665,8 +1665,8 @@ impl PdClient for TestPdClient {
                 )),
             )));
         }
-        let tso = self.tso.fetch_add(1, Ordering::SeqCst);
-        Box::pin(ok(TimeStamp::new(tso)))
+        let tso = self.tso.fetch_add(count as u64, Ordering::SeqCst);
+        Box::pin(ok(TimeStamp::new(tso + count as u64)))
```
BusyJay, line 1669: Why? For example, if count = 1 and self.tso is 1, it returns 1 in the past, but 2 now.

src/config.rs
```diff
@@ -2467,6 +2467,35 @@ impl LogConfig {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+#[serde(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct CausalTsConfig {
```
BusyJay, line 2473: It should be put into resolve-ts module.

components/test_raftstore/src/pd.rs
pingyu: In order to meet the conversion of PD that ["returns the timestamp with the biggest logical value"](https://github.com/tikv/tikv/blob/b5559844ce26a453900ca64c86902647dc807e28/components/pd_client/src/tso.rs#L210). And otherwise the [expected result](https://github.com/tikv/tikv/pull/12099/files#diff-e84ee9750d8a3e966a336e814bb4d3c2a9b88283afe12992c8045521551dd792R368) will be different with real PD.

I search codes may concern with the actual value of TSO, and change [one](https://github.com/tikv/tikv/pull/12099/files#diff-7875915f82a683a4d13b452ed136bcec8d572338c1e197b8a27142a40f6a618dR1066) of them.

Maybe I should add some comments for this new behavior of `set_tso`.

src/config.rs
pingyu: Could you give more comments about why ?
`causal_ts` has quite different functions with `resolved-ts`. Please also refer to [this discussion](https://github.com/tikv/tikv/pull/12099#discussion_r823352929) before.

components/causal_ts/src/observer.rs
pingyu: How about reserve 8 bytes when encoding mutations into `BlobRequest` for RawKV writes, and ingest timestamp in place in `CausalObserver` ?

components/server/Cargo.toml
pingyu: En...I see that almost all our components is set as `default-features = false`, so I think it would be conventional to do this.
`causal_ts` do not have default features, so the setting here has no effect.

src/config.rs
BusyJay: I don't see there is any discussion but @overvenus is making a similar suggestion. `CausalTsConfig` works for causal timestamp, and causal timestamp is defined in its own module, so it's intuitive to put them together.

components/server/Cargo.toml
BusyJay: They are written with `default-features = false` for a reason. Just don't make code verbose.

components/causal_ts/src/observer.rs
BusyJay: Besides it's impossible to do so as the length may change, it's too hacky and strange.

components/causal_ts/src/tso.rs
```diff
@@ -0,0 +1,466 @@
...
+struct TsoBatch {
+    size: u32,
+    physical: u64,
+    logical_end: u64, // exclusive
+    logical_start: AtomicU64,
+}
+
+impl TsoBatch {
+    pub fn pop(&self) -> Option<TimeStamp> {
+        let logical = self.logical_start.fetch_add(1, Ordering::Relaxed);
```
BusyJay, line 46: Better rewrite it with `load` and `compare_exchange_weak`.

src/config.rs
pingyu: En... You mean put into `causal-ts` module ? OK, I see.

BusyJay: Sorry, I just found I wrote the wrong ts 🤣

pingyu: Never mind ~

components/causal_ts/src/observer.rs
pingyu: En... then I have another idea:

1. Append timestamp in `Storage` raw interfaces. (It will break the assumption that timestamps are non-decreasing in a region.)

2. To make "resoved_ts" work, do lockings as following:
  2.1 Acquire a memory lock in `Storage`.
  2.2 [`track_lock`](https://github.com/tikv/tikv/blob/b5559844ce26a453900ca64c86902647dc807e28/components/cdc/src/delegate.rs#L679) in CDC observer, if the region is subscribed.
  2.3 Release the memory lock in CDC observer.

3. TSO is needed per key, other than per batch.

4. Other modules require to be changed:
  4.1 Full sorting in TiCDC (other than merge sort)
  4.2 TBD...

Cons: Overhead of memory locks & track_locks, longer write duration, much more TSO, sorting in TiCDC, etc.

Pros:
* Can benefit by #12020, to overcome above cons. The 20% improvement is very attractive.
* More simple and readable codes. E.g. the "resolved_ts" procedure above is very similar to transaction writes, and I think will be easier to modify and to be understood.

I can make another prototype to verify and benchmark the new proposal.

How do you think ?

BusyJay: I don't get it why it needs to acquire memory lock in storage and release it in CDC observer. If the lock is to guarantee TS increasing on the same key, then it can just acquire memory lock before allocating TS, and then release the lock after the request is sent to raftstore.

pingyu: The memory locks & track locks is used for CDC, to indicate that there are some requests with timestamp `ts` on the way from propose to apply, and "resolved_ts" should not be larger than these `ts`.

_Memory locks are used by [`cdc::Endpoint::register_min_ts_event()`](https://github.com/tikv/tikv/blob/7c100939dc66a683996806ff0f10ce41f6709732/components/cdc/src/endpoint.rs#L987)._
_Track locks are used by [`resolved_ts::Resolver::resolve()`](https://github.com/tikv/tikv/blob/7c100939dc66a683996806ff0f10ce41f6709732/components/resolved_ts/src/resolver.rs#L138), invoked from [`cdc::Delegate::on_min_ts()`](https://github.com/tikv/tikv/blob/b5559844ce26a453900ca64c86902647dc807e28/components/cdc/src/delegate.rs#L402)_

pingyu:
@BusyJay Thanks for your review. All comments are addressed or replied, PTAL~

components/causal_ts/src/observer.rs
```diff
@@ -0,0 +1,151 @@
...
+    QueryObserver, RoleChange, RoleObserver,
+};
+use std::sync::Arc;
+
+use crate::CausalTsProvider;
+
+/// CausalObserver appends timestamp for RawKV V2 data,
+/// and invoke causal_ts_provider.flush() on specified event, e.g. leader transfer, snapshot apply.
+/// Should be used ONLY when API v2 is enabled.
+pub struct CausalObserver<TS: CausalTsProvider> {
```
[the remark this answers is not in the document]
pingyu: Fixed. PTAL, thanks~

BusyJay (approved):

PR #12115, opened 2022-03-07
*: check memory locks for replica read only on the leader

### LINKED ISSUE #12109: Memory locks are not checked if read index is redirected twice

Any released version >= 5.0.0

### What did happened?

If we need to read on a follower or a learner, we have to send the read range and timestamp to the leader through read index, to make it compatible with async commit and 1PC.

But in some rare cases, the mechanism fails to work:

The reader sends read index to the leader it supposed to be, but when the leader receives the read index message, it has stepped down to a follower. But now, a peer will check the locks and re-fill the read index context with the result no matter what role it is.

So, when the read index request is redirected to the new leader again, it no longer carries the context of lock checking. Which means, the mechanism of checking locks and updating max TS fails.

This bug will cause the reader to read stale data when it is using replica read.

This is supposed to be the cause of https://github.com/pingcap/tidb/issues/32800

sticnarf:

Issue Number: Close #12109

What's Changed:

```commit-message
Consider a reader sends read index to the leader it supposed to be, but when
the leader receives the read index message, it has stepped down to a
follower. Without this commit, a peer will check the locks and re-fill the
read index context with the result no matter what role it is. So, when the
read index request is redirected to the new leader again, it no longer carries
the context of lock checking.

This commmit changes to only do the check when the peer is a leader. Then, the
read index request will remain unchanged before being redirected to the leader.

If the lease of the leader expires, it is still safe to check on the
uncertained leader. If the heartbeat check passes and no new leader is elected,
then the check works fine. If there is a new leader, when the old leader
becomes a follower, it will clear its pending read index. Then, the reader has
to resend the read index again to the correct leader. So, generally it is safe
as long as we guarantee the check only happens on the leader.
```

### Related changes

### Check List 

Tests 

- Integration test

### Release note 

```release-note
Fix potential linearizability violation in replica reads.
```

### DIFF
--- a/components/raftstore/src/coprocessor/dispatcher.rs
+++ b/components/raftstore/src/coprocessor/dispatcher.rs
@@ -526,9 +526,9 @@ impl<E: KvEngine> CoprocessorHost<E> {
         }
     }
 
-    pub fn on_step_read_index(&self, msg: &mut eraftpb::Message) {
+    pub fn on_step_read_index(&self, msg: &mut eraftpb::Message, role: StateRole) {
         for step_ob in &self.registry.read_index_observers {
-            step_ob.observer.inner().on_step(msg);
+            step_ob.observer.inner().on_step(msg, role);
         }
     }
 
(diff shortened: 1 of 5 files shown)


youjiali1995 (approved):

5kbpers (approved):

MyonKeminta (approved):

sticnarf:
Just find a test is unstable... Let me see...

https://ci.pingcap.net/blue/rest/organizations/jenkins/pipelines/tikv_ghpr_test/runs/17277/nodes/103/steps/179/log/?start=0

PR #12123, opened 2022-03-07
txn: Fix assertion fail on non-pessimistic key with conflict

### LINKED ISSUE #12113: Assertion false-positive on non-unique index key in pessimistic transaction that's rolled back by another transaction

master with assertion supported

### Steps to reproduce

It would be hard to reproduce manually.
Suppose two transaction t1 and t2, and a row key (pessimistic-locked key) `r1` and an index key (non-pessimistic-locked key) `i1`:
* t1 (start_ts = 10) lock key `r1` and then executes slowly
* t2 found lock on `r1` and rolled it back
* t2 continue and write a new version on both `r1` and `i1` (start_ts = 20, commit_ts = 30)
* t1 continues and two prewrite requests on `r1` and `i1` are sent.
  * Here, we expect to receive a `PessimisticLockNotFound` error raised by the request to `r1`, and nothing will happen on `i1` because it doesn't need pessimistic lock or checking conflicts. However, if `AssertionLevel::Strict` is set here, the prewrite request on `i1` will see the new version written by t2 and possibly raise a false-positive assertion failure.

### Solution

It's not easy to fix it while keeping the check behavior. We are not considering to make assertion failure error less prior to other errors such as `KeyIsLocked`. Both TiKV and client-go need to be updated. The check will be weaken by this change: it's possible that in some cases that real assertion failure occurs but not reported due to meeting other normal errors.

MyonKeminta:

Issue Number: ref #12113
(needs another PR in client-go)

What's Changed:

Fixes the false-positive assertion fail on non-pessimistic key with conflict. This PR adopts the solution mentioned in issue #12113 , and needs a corresponding PR in client-go too.

```commit-message
txn: Fix assertion fail on non-pessimistic key with conflict (ref #12113)
```

### Check List 

Tests 

- Unit test

Side effects
### Release note 

```release-note
None
```


sticnarf (approved):
LGTM

ekexium (approved):

PR #12124, opened 2022-03-07
raftstore,cdc: pass leader transferee to cdc observer

### LINKED ISSUE #12111: Pass leader transferee to cdc observer to reduce TiCDC latency spike
As title.

hicqu:

Issue Number: Close #12111, close pingcap/tiflow#4761

What's Changed:

```commit-message
Pass leader transferee to cdc observer
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Pass leader transferee to cdc observer to reduce TiCDC latency spike.
```


components/cdc/src/observer.rs
```diff
@@ -252,7 +264,7 @@ mod tests {
         };
```
overvenus, line 265: Please also add the unit in https://github.com/tikv/tikv/pull/12110/files#diff-05bdfd7edb585436e8424bb42e74a6f92d53b4a1bcae9212d9d717746b2a5493R285

overvenus (approved):
LGTM!

hi-rustin (approved):

components/raftstore/src/store/peer.rs
```diff
@@ -1694,16 +1698,22 @@ where
...
-            // TODO: it may possible that only the `leader_id` change and the role
-            // didn't change
-            ctx.coprocessor_host
-                .on_role_change(self.region(), ss.raft_state);
+            ctx.coprocessor_host.on_role_change(
+                self.region(),
+                RoleChange {
+                    state: ss.raft_state,
+                    leader_id: ss.leader_id,
+                    prev_lead_transferee: self.lead_transferee,
```
BusyJay, line 1706: I don't get it. In what case when `lead_transferee` has value while it's not equal to `vote`?

hicqu: `lead_transferee` could be stale. For example, If the leadership has transfered from A to B, and then starts to transfer to C quickly.

BusyJay: In this case, `vote` should still be the correct value.

hicqu: Yes, however `vote` isn't the tranferee. We hope to only handle the leadership transferring case.

BusyJay: What's the point of transferee in this case? If what you want is the next peer to visit, then `vote` should be used instead of `lead_transferee` as no matter who the leader tries to transfer leader to, the next leader is likely the one it votes for.

Issue #11424, opened 2021-11-19
QPS dropped severely many times during scale out in DBaaS


Lily2025:
### What version of TiKV are you using?
5.3.0

### What operating system and CPU are you using?
Cluster Tier
T3.standard

### Steps to reproduce
run sysbench prepare
sysbench --config-file=config oltp_insert --tables=16 --table-size=100000000 prepare
threads=8
scale out tikv from 3 to 6

### What did you expect?
performance is stable

### What did happened?
![NYTJMTk7xU](https://user-images.githubusercontent.com/84712107/142590227-d2d79f89-f0b5-49a6-88dd-11d45f4d3cf7.png)
![zYRSkgmqqs](https://user-images.githubusercontent.com/84712107/142590256-24e8a7e6-c407-4552-8023-1a15b801fbee.png)

zhangjinpeng1987:
@Connor1996 any update for this issue?

MimeLyc:
update

PR #12127, opened 2022-03-08
engine: only override write stall configurations if unspecified

tabokie:

Issue Number: Ref #11424

What's Changed:

Follow up on #11840.

This is handy if we want to make RocksDB compaction a bit faster without triggering flow control.

### Related changes

- Doc change

### Check List 

Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note
None
```


BusyJay (approved):

src/config.rs
```diff
@@ -2721,58 +2745,42 @@ impl TiKvConfig {
...
-                .soft_pending_compaction_bytes_limit;
-            self.rocksdb.raftcf.hard_pending_compaction_bytes_limit = self
-                .storage
-                .flow_control
-                .hard_pending_compaction_bytes_limit;
         }
+        // Fill in values for unspecified write stall configurations.
+        macro_rules! fill_cf_opts {
+            ($cf_opts:expr, $flow_control_cfg:expr) => {
+                if $cf_opts.level0_slowdown_writes_trigger.is_none() {
```
Connor1996, line 2756: so if the config is specified, it won't follow the config of flow control? it's unexpected and it would cause QPS drop once L0 reaches the flow controller's threshold whereas ingest_maybe_stall() is still false.

tabokie: How about returning error when cf_opts is larger than flow_control.

Connor1996: Discuss with Jay, let's add a config check to make sure the config of rocksdb is smaller or equal to one of flow control

src/config.rs
```diff
@@ -2772,58 +2796,78 @@ impl TiKvConfig {
...
+                        Some($flow_control_cfg.hard_pending_compaction_bytes_limit);
+                }
+            };
+        }
+        let flow_control_cfg = if self.storage.flow_control.enable {
+            self.storage.flow_control.clone()
+        } else {
+            crate::storage::config::FlowControlConfig::default()
+        };
+        fill_cf_opts!(self.raftdb.defaultcf, flow_control_cfg);
```
Connor1996, line 2866: Do these check only when flow control is enabled

src/config.rs
```diff
@@ -2803,9 +2803,9 @@ impl TiKvConfig {
         }
         // Fill in values for unspecified write stall configurations.
         macro_rules! fill_cf_opts {
-            ($cf_opts:expr, $flow_control_cfg:expr) => {
+            ($cf_opts:expr, $flow_control_cfg:expr, $check:expr) => {
                 if let Some(v) = $cf_opts.level0_slowdown_writes_trigger {
-                    if v > $flow_control_cfg.l0_files_threshold as i32 {
+                    if $check && v > $flow_control_cfg.l0_files_threshold as i32 {
```
Connor1996, line 2808: $flow_control_cfg.enable

src/config.rs
```diff
@@ -2772,58 +2796,98 @@ impl TiKvConfig {
...
-                .flow_control
-                .hard_pending_compaction_bytes_limit;
         }
+        // Fill in values for unspecified write stall configurations.
+        macro_rules! fill_cf_opts {
+            ($cf_opts:expr, $flow_control_cfg:expr, $check:expr) => {
+                if let Some(v) = $cf_opts.level0_slowdown_writes_trigger {
+                    if $check && v > $flow_control_cfg.l0_files_threshold as i32 {
+                        return Err(format!(
+                            "{}.level0-slowdown-writes-trigger ({}) should not be greater than \
```
Connor1996, line 2810: It occurs to me that upgrading from the old cluster may trigger this error, how about just printing a warn and override it.

src/config.rs
```diff
@@ -2772,58 +2796,81 @@ impl TiKvConfig {
...
+                            stringify!($cf_opts), $cfg.l0_files_threshold
+                        );
+                        *v = $cfg.l0_files_threshold as i32;
+                    }
+                } else {
+                    $cf_opts.level0_slowdown_writes_trigger =
+                        Some($cfg.l0_files_threshold as i32);
+                }
+                if let Some(v) = &mut $cf_opts.level0_stop_writes_trigger {
+                    if $cfg.enable && *v > $cfg.l0_files_threshold as i32 {
```
Connor1996, line 2821: no need to check `stop_writes_trigger`

src/config.rs
```diff
@@ -280,8 +280,8 @@ macro_rules! cf_config {
             pub compaction_style: DBCompactionStyle,
             pub disable_auto_compactions: bool,
             pub disable_write_stall: bool,
-            pub soft_pending_compaction_bytes_limit: ReadableSize,
-            pub hard_pending_compaction_bytes_limit: ReadableSize,
+            pub soft_pending_compaction_bytes_limit: Option<ReadableSize>,
```
Connor1996, line 283: what's the output of `Option` after serialization

src/config.rs
tabokie: Why, I remember stop is allowed to be smaller than slow, in which case RocksDB will modify slow to match stop.

src/config.rs
tabokie: `None` will not be serialized, `Some` is serialized in the same way as normal fields.

PR #12132, opened 2022-03-09
Raftstore: support bucket split/merge -- #11759

tonyxuqqi:
ref -- #11759

Issue Number: ref #11759

What's Changed:

```
Support bucket split/merge per bucket size.
```

### Related changes

- https://github.com/tikv/tikv/pull/11763

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU when region_bucket_enabled is true

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4683,17 +4701,72 @@ where
...
+            let mut j = 0;
+            let buckets = buckets.drain(..);
+            for mut bucket in buckets {
+                while i < region_buckets_keys.len() {
+                    if region_buckets_keys[i] == bucket_ranges[j].0 {
+                        // the bucket size is small and does not have split keys,
+                        // then it should be merged with its left neighbor
+                        if bucket.keys.is_empty()
+                            && bucket.size
+                                <= self.ctx.coprocessor_host.cfg.region_bucket_merge_size.0
```
BusyJay, line 4724: So 128 buckets `[128MiB, 1MiB, ..., 1MiB]` will be merged as `[255MiB]` instead of `[128MiB, 128MiB]`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4683,17 +4701,72 @@ where
...
+                        {
+                            // i is not the last entry (which is end key)
+                            if region_buckets_keys.len() > 2 {
+                                region_buckets_keys.remove(i); // bucket is too small
+                                // the key is removed, to compensate the += 1 below.
+                                i -= 1;
```
BusyJay, line 4730: Why not just `continue`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4683,17 +4701,72 @@ where
...
+            }
+
+            region_buckets.set_keys(region_buckets_keys.into());
+            assert_eq!(j, bucket_ranges.len());
+            self.fsm.peer.region_buckets = region_buckets;
+        } else {
+            assert_eq!(buckets.len(), 1);
+            let bucket_keys = buckets.pop().unwrap().keys;
+            let mut region_buckets = Buckets::default();
+            let now = monotonic_raw_now();
```
BusyJay, line 4752: Should use system time instead of monotontic_raw_now. The latter is an arbitrary number the kernel choose.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: but as long as 255MiB increases to 256MiB, then it will split into [128MiB, 128MiB].  Here we have the cap of unbalance.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: Good catch. The previous version has some other code to run below, but this version can use 'continue'.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: Here all we need is a monotonic increasing number to differentiate Buckets right? System time can go backwards which can lead to confuse about which Bucket version is newer.

BusyJay: It needs to increase monotonic across machines.

components/raftstore/src/store/fsm/peer.rs
BusyJay: When will it be split? I don't see further check is done after the loop.

tonyxuqqi: > When will it be split? I don't see further check is done after the loop.

The next time the check split run.

tonyxuqqi: Refined a bit the logic. Now it will only do the merge if the combined bucket size does not exceed 2x of region bucket size. This is to avoid too big bucket size.
In your example, it's still 255MiB.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: cross machine comparison rely on region's epoch as today bucket version will compare epoch first.

components/pd_client/src/lib.rs
```diff
@@ -149,6 +164,36 @@ impl BucketStat {
             *bytes += key.len() as u64 + value_size;
         }
     }
+
+    pub fn split(&mut self, idx: usize) {
```
5kbpers, line 168: @bufferflies and I discussed previously that both split buckets should inherit the stats of the original bucket. Should we do this here as well?

components/raftstore/src/coprocessor/config.rs
```diff
@@ -87,6 +90,7 @@ impl Default for Config {
             enable_region_bucket: false,
             region_bucket_size: DEFAULT_BUCKET_SIZE,
             region_size_threshold_for_approximate: DEFAULT_LARGE_REGION_SIZE_THRESHOLD,
+            region_bucket_merge_size: DEFAULT_BUCKET_MERGE_SIZE,
```
5kbpers, line 93: Suggest that the default value of it is better to be calculated from `region_bucket_size` e.g. `region_bucket_size / 2`, otherwise it would be painful to tune the config.

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -99,9 +99,20 @@ impl<'a, E> Host<'a, E> {
...
-            return bucket_checker.approximate_split_keys(region, engine);
+            let result = bucket_checker.approximate_split_keys(region, engine);
+            if let Ok(keys) = result {
+                return Ok(Bucket {
+                    keys,
+                    size: region_size,
+                });
+            } else {
+                return Err(result.unwrap_err());
+            }
```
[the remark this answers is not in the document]
tonyxuqqi: Thanks. Good suggestion.

components/pd_client/src/lib.rs
tonyxuqqi: ok.

components/raftstore/src/coprocessor/config.rs
tonyxuqqi: Do you mean region_bucket_merge_size's default value is based on actual value of region_bucket_size (not the default DEFAULT_BUCKET_SIZE) right?
But apparently it cannot be done in default() because all the values here are hardcoded.

components/raftstore/src/store/msg.rs
```diff
@@ -39,6 +41,13 @@ pub struct WriteResponse {
     pub response: RaftCmdResponse,
 }
 
+// Peer's internal stat, for test purpose only
+#[derive(Debug)]
```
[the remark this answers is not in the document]
tonyxuqqi: it cannot because it will be used by components/test_raftstore/src/cluster.rs which is not under #[cfg(test)].
Also the code will be a bit ugly because  #[cfg(test)] will scatter among product code.

components/raftstore/src/coprocessor/config.rs
```diff
@@ -124,6 +128,16 @@ impl Config {
         if self.region_bucket_size.0 == 0 {
             return Err(box_err!("region_bucket size cannot be 0."));
         }
+        if self.region_bucket_merge_size.0 == 0 {
+            return Err(box_err!("region-bucket-merge-size cannot be 0."));
+        }
+        if self.region_bucket_merge_size.0 >= self.region_bucket_size.0 {
```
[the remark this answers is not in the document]
tonyxuqqi: Good point.  It's possible that two buckets are merged and then in next round it's split again if the new merged bucket has some write traffic and leads to the actual size grows to bigger than configured region_bucket, which is legitimate. 

But if the config of region_bucket_merge_size is more than half of configured region_bucket_size (user changes the default value), then merged bucket size could be larger than region_bucket_size which could be split in next round. This is not legitimate. 
So to prevent this happening, the check here should be if self.region_bucket_merge_size.0 >= self.region_bucket_size.0/2.

components/raftstore/src/store/msg.rs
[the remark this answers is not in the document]
BusyJay: You can guard it with testexport features.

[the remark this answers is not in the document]
tonyxuqqi: > You can guard it with testexport features.

good to know. Thanks.

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -124,11 +124,21 @@ where
     }
 }
 
+#[derive(Default, Clone, Debug)]
+pub struct BucketRange(pub Vec<u8>, pub Vec<u8>);
+
+#[derive(Default, Clone, Debug)]
+pub struct Bucket {
+    pub keys: Vec<Vec<u8>>,
+    pub size: u64,
```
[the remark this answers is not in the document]
tonyxuqqi: > It's bucket size or region size?

bucket size---as it's under Bucket struct.

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -187,8 +206,97 @@ where
...
+    ) {
+        let mut buckets: Vec<Bucket> = Vec::default();
+        let mut skip = false;
+        if let Some(ref bucket_ranges) = bucket_ranges {
+            skip = bucket_ranges.is_empty();
+            for bucket_range in bucket_ranges {
+                let mut bucket = region.clone();
+                bucket.set_start_key(bucket_range.0.clone());
+                bucket.set_end_key(bucket_range.1.clone());
+                let bucket_entry = match host.approximate_bucket_keys(&bucket, &self.engine) {
```
5kbpers, line 223: How about replacing the parameter `Region` of approximate_bucket_keys with two slices of `u8`  for reducing clone?

5kbpers (approved):

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -270,9 +385,22 @@ where
...
+        let empty_bucket = vec![];
+        let (mut skip_check_bucket, bucket_range_list) =
+            if let Some(ref bucket_range_list) = bucket_ranges {
+                (
+                    bucket_range_list.is_empty() || !host.enable_region_bucket(),
+                    bucket_range_list,
+                )
+            } else {
+                (!host.enable_region_bucket(), &empty_bucket)
+            };
```
youjiali1995, line 402: It looks strange. Why it skips checking bucket if `bucket_ranges` is `Some(vec![])` and may checks bucket if `bucket_ranges` is `None`.

tonyxuqqi: When specifying bucket_ranges with value, it means check_split_and_bucket will ONLY check bucket in bucket_ranges and thus it's very straightforward that if it's empty then nothing to check. 
On the other hand when bucket_ranges is not specified, it means checks the whole region. The reason to have None instead of a vec![Bucket{start, end key of region}] is that it's more explicit to mean the whole region.

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -187,8 +206,97 @@ where
...
+            let _ = self.router.send(
+                region.get_id(),
+                CasualMessage::RefreshRegionBuckets {
+                    region_epoch: region.get_region_epoch().clone(),
+                    buckets,
+                    bucket_ranges,
+                    cb: Callback::None,
+                },
+            );
+        }
```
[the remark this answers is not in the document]
tonyxuqqi: Thanks.

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -99,9 +99,17 @@ impl<'a, E> Host<'a, E> {
                 region_size / self.cfg.region_bucket_size.0,
                 CheckPolicy::Approximate,
             );
-            return bucket_checker.approximate_split_keys(region, engine);
+            return bucket_checker
+                .approximate_split_keys(region, engine)
+                .map(|keys| Bucket {
+                    keys,
+                    size: region_size,
+                });
```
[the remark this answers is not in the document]
tonyxuqqi: good point. It can simplified the caller's logic.

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -223,16 +285,34 @@ where
...
+                    if host.enable_region_bucket() {
+                        match self.approximate_check_bucket(region, &mut host, bucket_ranges) {
+                            Ok(()) => {}
+                            Err(e) => {
+                                error!(%e;
+                                    "approximate_check_bucket failed";
+                                    "region_id" => region_id,
+                                );
+                            }
+                        }
```
youjiali1995, line 298:
```suggestion
                        if let Err(e) =
                            self.approximate_check_bucket(region, &mut host, bucket_ranges)
                        {
                            error!(%e;
                                "approximate_check_bucket failed";
                                "region_id" => region_id,
                            );
                        }
```

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -391,7 +498,15 @@ where
...
-                    self.refresh_approximate_bucket_keys(&region, &mut host);
+                    match self.approximate_check_bucket(&region, &mut host, None) {
+                        Ok(()) => {}
+                        Err(e) => {
+                            error!(%e;
+                                "approximate_check_bucket failed";
+                                "region_id" => region.get_id(),
+                            );
+                        }
+                    }
```
youjiali1995, line 509:
```suggestion
                    if let Err(e) = self.approximate_check_bucket(&region, &mut host, None) {
                        error!(%e;
                            "approximate_check_bucket failed";
                            "region_id" => region.get_id(),
                        );
                    }
```

components/raftstore/src/store/worker/split_check.rs
```diff
@@ -284,28 +377,66 @@ where
...
+                                bucket.keys.push(origin_key.to_vec());
+                                bucket.size += bucket_size;
+                                bucket_size = 0;
+                            }
+                        }
                     }
                 }
             }
 
+            if buckets.len() < bucket_range_list.len()
```
youjiali1995, line 427: Is there the invariant that `buckets.len() == bucket_range_list.len()`?

youjiali1995: I saw the assertion at https://github.com/tikv/tikv/pull/12132/files#diff-7a4718b9cddd27a4ef7e563024d5283cf14b4ba3abf26edc33b974b87680a5fbR4822.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4749,27 +4791,132 @@ where
...
+            .region_buckets
+            .as_ref()
+            .map(|b| b.meta.version)
+            .unwrap_or_default();
+        let mut region_buckets: BucketStat;
+        if let Some(bucket_ranges) = bucket_ranges {
+            assert_eq!(buckets.len(), bucket_ranges.len());
+            let mut i = 0;
+            let mut j = 0;
+            region_buckets = self.fsm.peer.region_buckets.as_ref().unwrap().clone();
```
youjiali1995, line 4825:
```suggestion
            region_buckets = self.fsm.peer.region_buckets.clone().unwrap();
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4780,11 +4927,66 @@ where
...
+            // if the bucket's write_bytes exceed half of the configured region_bucket_size,
+            // add it to the bucket_ranges for checking update
+            let bucket_update_diff_size_threshold =
+                self.ctx.coprocessor_host.cfg.region_bucket_size.0 / 2;
+            if diff_in_bytes >= bucket_update_diff_size_threshold {
+                if i == stats.write_bytes.len() - 1 {
+                    bucket_ranges.push(BucketRange(keys[i].clone(), vec![]));
+                } else {
+                    bucket_ranges.push(BucketRange(keys[i].clone(), keys[i + 1].clone()));
+                }
```
youjiali1995, line 4978: `keys`'s length == `write_bytes`'s length + 1, and you shouldn't use `vec![]` as the end key because checker doesn't handle it properly.
```suggestion
                bucket_ranges.push(BucketRange(keys[i].clone(), keys[i + 1].clone()));
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4749,27 +4791,132 @@ where
...
+                    }
+                }
+                i += 1;
+                j += 1;
+            }
+            meta.region_epoch = region_epoch;
+            meta.version = gen_bucket_version(self.fsm.peer.term(), current_version);
+            region_buckets.meta = Arc::new(meta);
+            assert_eq!(j, bucket_ranges.len());
+        } else {
```
youjiali1995, line 4871:
```suggestion
        if let Some(bucket_ranges) = bucket_ranges {
            assert_eq!(buckets.len(), bucket_ranges.len());
            let mut i = 0;
            region_buckets = self.fsm.peer.region_buckets.clone().unwrap();
            let mut meta = (*region_buckets.meta).clone();
            for (bucket, bucket_range) in buckets.into_iter().zip(bucket_ranges) {
                while i < meta.keys.len() && meta.keys[i] != bucket_range.0 {
                    i += 1;
                }
                assert!(i != meta.keys.len());
                // the bucket size is small and does not have split keys,
                // then it should be merged with its left neighbor
                let region_bucket_merge_size =
                    self.ctx.coprocessor_host.cfg.region_bucket_merge_size_ratio
                        * (self.ctx.coprocessor_host.cfg.region_bucket_size.0 as f64);
                if bucket.keys.is_empty() && bucket.size <= (region_bucket_merge_size as u64) {
                    meta.sizes[i] = bucket.size;
                    // i is not the last entry (which is end key)
                    assert!(i < meta.keys.len() - 1);
                    // the region has more than one bucket
                    // and the left neighbor + current bucket size is not very big
                    if meta.keys.len() > 2
                        && i != 0
                        && meta.sizes[i - 1] + bucket.size
                            < self.ctx.coprocessor_host.cfg.region_bucket_size.0 * 2
                    {
                        // bucket is too small
                        region_buckets.left_merge(i);
                        meta.left_merge(i);
                        continue;
                    }
                } else {
                    // update size
                    meta.sizes[i] = bucket.size / (bucket.keys.len() + 1) as u64;
                    // insert new bucket keys (split the original bucket)
                    for bucket_key in bucket.keys {
                        i += 1;
                        region_buckets.split(i);
                        meta.split(i, bucket_key);
                    }
                }
                i += 1;
            }
            meta.region_epoch = region_epoch;
            meta.version = gen_bucket_version(self.fsm.peer.term(), current_version);
            region_buckets.meta = Arc::new(meta);
        } else {
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4749,27 +4791,132 @@ where
...
+                if bucket.keys.is_empty() && bucket.size <= (region_bucket_merge_size as u64) {
+                    meta.sizes[i] = bucket.size;
+                    // i is not the last entry (which is end key)
+                    assert!(i < meta.keys.len() - 1);
+                    // the region has more than one bucket
+                    // and the left neighbor + current bucket size is not very big
+                    if meta.keys.len() > 2
+                        && i != 0
+                        && meta.sizes[i - 1] + bucket.size
+                            < self.ctx.coprocessor_host.cfg.region_bucket_size.0 * 2
```
youjiali1995, line 4846: Why *2 here? Is it possible the bucket is splitted soon?

components/raftstore/src/store/worker/split_check.rs
tonyxuqqi: thanks.

components/raftstore/src/store/worker/split_check.rs
tonyxuqqi: > Is there the invariant that `buckets.len() == bucket_range_list.len()`?

When bucket_range_list is None, it would still return one bucket object that contains all the split keys for the whole region.

tonyxuqqi: > I saw the assertion at https://github.com/tikv/tikv/pull/12132/files#diff-7a4718b9cddd27a4ef7e563024d5283cf14b4ba3abf26edc33b974b87680a5fbR4822. It would be better to assert it here.

fair.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: > `keys`'s length == `write_bytes`'s length + 1, and you shouldn't use `vec![]` as the end key because checker doesn't handle it properly.

Good catch. Updated the checker to handle vec![] properly.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: it's not possible because when do split, we only allow bucket's size >= 2* region_bucket_size. 
The idea here is to reduce small size bucket as much as possible.

youjiali1995 (approved):

PR #8193, opened 2020-07-05
backup,config: set backup thread pool size via config instead of gRPC

kennytm:
### What problem does this PR solve?

Issue Number: close pingcap/br#394, close #8076. 

Problem Summary:

### What is changed and how it works?

This is the companion PR of pingcap/br#396. 

In #5553, the `backup` component gained ability to an adjustable thread pool size. The size is provided via the incoming gRPC request. Given [how the ThreadPool is implemented in TiKV](https://github.com/tikv/tikv/blob/b45e052df8fb5d66aa8b3a77b5c992ddbfbb79df/components/tikv_util/src/threadpool.rs#L204-L219), a user can easily (accidentally) DoS all TiKV servers by setting the thread pool size to a huge number like 2<sup>32</sup>-1. 

Furthermore, when we provide the thread pool size through gRPC, the backup client (BR) must determine the suitable concurrency number in some ad-hoc way. 

This PR solves both issue by moving the thread pool size back as a TiKV config:

```toml
[backup]
num-threads = 24
```

The default value is set to (0.75 × Number of CPUs) to better utilize the available resources. The value can still be dynamically configured, but given this more practical default value it should be much less needed.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test

Side effects

### Release note 

* A new configuration `backup.num-threads` is introduced to control the backup thread pool size.

PR #12141, opened 2022-03-10
config: limit backup workers cap

ethercflow:

ref #11804  

What's Changed:

```commit-message
config: limit backup workers cap
```

### Related changes

- None

### Check List 

Tests 

- Manual test,
   - use mysql client to test

Side effects
- None

### Release note 

```release-note
None.
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2234,8 +2234,9 @@ pub struct BackupConfig {
 
 impl BackupConfig {
     pub fn validate(&self) -> Result<(), Box<dyn Error>> {
-        if self.num_threads == 0 {
-            return Err("backup.num_threads cannot be 0".into());
+        let limit = SysQuota::cpu_cores_quota() as usize;
+        if self.num_threads == 0 || self.num_threads > limit {
+            return Err(format!("backup.num_threads cannot be 0 or larger than {}", limit).into());
         }
         if self.batch_size == 0 {
             return Err("backup.batch_size cannot be 0".into());


ethercflow:

src/config.rs
```diff
@@ -2234,8 +2234,9 @@ pub struct BackupConfig {
 
 impl BackupConfig {
     pub fn validate(&self) -> Result<(), Box<dyn Error>> {
-        if self.num_threads == 0 {
-            return Err("backup.num_threads cannot be 0".into());
+        let limit = SysQuota::cpu_cores_quota() as usize;
+        if self.num_threads == 0 || self.num_threads > limit {
```
NingLin-P, line 2238: Same concern as https://github.com/tikv/tikv/pull/12068#discussion_r825657740, also the actual number of threads used by backup is decided by the backup request, `num_threads` is used by a token limiter

ethercflow: I discussed it with @YuJuncen before, since backup is cpu intensive, so it doesn't make much sense to set the number of threads above 100% cpu quota.  And it has the feature of automatic current limiting when the load is high.  /cc @YuJuncen

YuJuncen: @NingLin-P 
> actual number of threads used by backup is decided by the backup request.
The number of threads isn't decided by the request more because the risk of misconfigure leading to panicking. (see https://github.com/tikv/tikv/pull/8193)

The step of saving backup files into external storage is now asynchronous, which means those threads would only be blocked by scanning the local MVCC(which is fast enough and hardly blocking, in some test cases, with the NVME disks,  CPU usage of those threads become up to 95%.). Most of CPUs were used for doing `zstd` compressing. So IMO, we cannot take much good from backing up with a thread pool huger than the CPU number size.

NingLin-P (approved):

5kbpers (approved):

PR #12145, opened 2022-03-10
cmd: link c++ explicitly

### LINKED ISSUE #12144: The prebuilt tikv-server requires stdc++ now
```
% ldd bin/tikv-server 
	linux-vdso.so.1 =>  (0x00007ffc64f45000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f7fda17c000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f7fd9da3000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f7fd9b87000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f7fd9971000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f7fd9769000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f7fd9467000)
	libc.so.6 => /usr/local/lib/libc.so.6 (0x00007f7fd9099000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7fddc34000)
```

The changes seems to be brought by #12098.

BusyJay:
Issue Number: Close #12144

What's Changed:

```commit-message
In the past, we rely on librocksdb-sys to include the required stdc++
library. This is not reliable after upgrading grpcio, which also needs
the same library. I suspect it doesn't include all the required symbols.

Anyway, this PR chooses to link to stdc++ explicitly in binary.
```

### Check List 

Tests 

### Release note 

```release-note
None
```

### DIFF
--- a/cmd/build.rs
+++ b/cmd/build.rs
@@ -1,8 +1,51 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
+fn link_cpp(tool: &cc::Tool) {
+    let stdlib = if tool.is_like_gnu() {
+        "libstdc++.a"
+    } else if tool.is_like_clang() {
+        "libc++.a"
+    } else {
+        // Don't link to c++ statically on windows.
+        return;
+    };
+    link_sys_lib(stdlib, &tool)
+}
+
+fn link_sys_lib(lib: &str, tool: &cc::Tool) {
+    let output = tool
+        .to_command()
+        .arg("--print-file-name")
+        .arg(lib)
+        .output()
+        .unwrap();
+    if !output.status.success() || output.stdout.is_empty() {
+        // fallback to dynamically
+        return;
+    }
+    let path = match std::str::from_utf8(&output.stdout) {
+        Ok(path) => std::path::PathBuf::from(path),
+        Err(_) => return,
+    };
+    if !path.is_absolute() {
+        return;
+    }
+    // remove lib prefix and .a postfix.
+    let libname = &lib[3..lib.len() - 2];
+    println!("cargo:rustc-link-lib=static={}", &libname);
+    println!(
+        "cargo:rustc-link-search=native={}",
+        path.parent().unwrap().display()
+    );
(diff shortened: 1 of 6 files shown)


Connor1996 (commented):
how about adding a check in `check-bins.py` to make sure some libs are linked statically.

Connor1996 (approved):
LGTM

scripts/check-bins.py
```diff
@@ -18,6 +18,8 @@
 
 JEMALLOC_SYMBOL = ["je_arena_boot", " malloc"]
 
+SYS_LIB = ["libstdc++"]
```
tabokie, line 21: Is "libc++" needed too?

cmd/build.rs
```diff
@@ -1,8 +1,51 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
+fn link_cpp(tool: &cc::Tool) {
+    let stdlib = if tool.is_like_gnu() {
+        "libstdc++.a"
+    } else if tool.is_like_clang() {
+        "libc++.a"
```
tabokie, line 7: I can't find "libc++.a" with `clang++`, found "libstdc++.a" though.

tabokie (approved):
Rest LG.

cmd/build.rs
sticnarf: I believe TiKV cannot be built with clang++ yet. There are quite a lot errors when compiling RocksDB with clang

BusyJay: Interesting, seems titan can't pass compilation
```
/git/checkouts/rust-rocksdb-a9a28e74c6ead8ef/1494bb5/librocksdb_sys/libtitan_sys/titan/src/db_impl.cc:105:14: error: variable 'file_size' set but not used [-Werror,-Wunused-but-set-variable]
      uint64_t file_size = 0;
               ^
  1 error generated.

```

PR #12151, opened 2022-03-14
Foreground Quota Limiter: limit the cpu, bandwidth to provide stable QPS

### LINKED ISSUE #11855: Provide predictable stable write QPS under specified environment

If we give enough pressure to the TiKV, the write QPS of TiKV is usually high at the begin and then drop as time pass by. It is not friendly to the user. If TiKV can provide predictable stable write QPS, it is a better experience for the user, although the stable QPS may lower than the max peak QPS.

Add CPU Quota and write bandwidth quota for TiKV to provide predictable stable write QPS when the environment is specified like 4vCPU with a 4K IOPS 100MB/s bandwidth disk.

hunterlxt:

Issue Number: Close #11855 #11856

What's Changed:

```commit-message
Add a global forefront quota limiter based speed limiter to record the CPU resource overhead occupied by various requests. When the limiter reaches the limit quota, the request will be forced to sleep for a period of time to compensate. Statistics are only done on the part of process and are therefore approximate.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

we can see after quota a small quota, QPS became more stable.
image

We have some discussion on sleep using sync or async. If use sync mode, the related thread will be blocked but this may be what this feature want, which means compensating for the cost cpu time to yield cpu resource to other worker. If use async mode, await makes the threads not blocked but the CPU limit is inaccurate because `await` is just a task to give up the CPU, the worker still can execute others. Below is some comparative tests

1. Using smaller refills incurs more thread switching overhead
2. Check that the return value of each consume_write/read does not exceed the upper limit, this behavior has no benefits and no disadvantages, so I think it will help not to delay some requests for too long
3. In async delay mode, the quota_delay of `kv_snapshot` should not be delayed, because `kv_snapshot` is often called by `with_tls_engine`, which is part of other requests and is very frequent. In async mode, the limiter's buckets have not been refilled, so a large number of `kv_snapshots` are slowed down.
4. In sync sleep mode, cpu resources will be released truely, so the throttled duration on the grafana panel are more accurate and much smaller. Since async only kicks out the current tasks, it needs to artificially limit a smaller quota value to achieve a similar cpu limit effect to sync sleep mode.
5. Using speed limiter for cpu usage is not a good choice, because the time used by cpu itself also occupies time buckets, so maybe it should be delayed = (quota_delay - cost time)

image
image
image
image
image
image

Side effects

- It has no effect if it is turned off by default. After turning on the limit, it may affect the maximum qps value.

### Release note 

```release-note
Add a global foreground quota limiter to control cpu and bandwidth usage, which will improve the stability of QPS.
```


tabokie:
Could you show us some test results? In specific is there any overhead when quota limits are disabled? And how does the incomplete CPU metering affect the accuracy of CPU limiting?

components/tidb_query_executors/src/runner.rs
```diff
@@ -606,7 +624,12 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
         }
 
         warnings.merge(&mut result.warnings);
-        Ok((is_drained, record_len))
+        let wait = self.quota_limiter.consume_read(
+            1,
+            record_len,
+            (thread_cpu_time.elapsed().as_micros()) as usize,
```
sticnarf, line 630: It is costy to fetch the thread cpu time. Now, we read thread cpu time no matter whether this feature is on or off. I believe this will cause performance regression, especially for point get.

components/tidb_query_executors/src/runner.rs
```diff
@@ -511,13 +525,16 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
-            let (drained, len) = self.internal_handle_request(
+            let (drained, len, quota_delay) = self.internal_handle_request(
                 true,
                 batch_size.min(self.stream_row_limit - record_len),
                 &mut current_chunk,
                 &mut warnings,
                 &mut ctx,
             )?;
+
+            std::thread::sleep(quota_delay);
```
sticnarf, line 536: Thread sleep is used here but an async sleep is used everywhere else. How do I know which to choose?

components/tidb_query_executors/src/runner.rs
hunterlxt: In `QuotaLimiter`, if it set to 0, limiter will jump and just return Duration::ZERO

components/tidb_query_executors/src/runner.rs
hunterlxt: use async sleep in aysnc code block and thread sleep in sync code block

components/tidb_query_executors/src/runner.rs
sticnarf: But most of the cost comes from `ThreadTime::now()` and `thread_cpu_time.elapsed()`, which are not filtered by the feature flag.

components/tidb_query_executors/src/runner.rs
sticnarf: But the effect is different. If an async sleep is used, the same thread pool can still handle other requests immediately. But a thread sleep will truly block the thread, making the thread unable to serve any task for a while. Which one is you really want?

hunterlxt: seems that we want a truly block

components/tidb_query_executors/src/runner.rs
hunterlxt: Got it

hunterlxt:
> Could you show us some test results? In specific is there any overhead when quota limits are disabled? And how does the incomplete CPU metering affect the accuracy of CPU limiting?

results are posted in description. 

when quota limits are disabled, you might worry about two parts of overhead: 1. access cpu time 2. quota limiter calculation. For 1, I added a unify method to get start timer, when cpu limiter diabled, it just use `CLOCK_MONOTONIC`; For 2, there is a field named `is_unlimited` in speed limiter, when quota is set to INFINITY, `is_unlimited` will be set to true and all consume operation will return immediately. 

incomplete CPU metering bring a negative effec that users can't think of this as a restriction like cgroups, it's more of a reference. And this feature is more used for cloud, of course, external experienced users should also be happy that there is at least one option to turn it on.

components/tikv_kv/src/lib.rs
```diff
@@ -456,10 +458,24 @@ pub unsafe fn destroy_tls_engine<E: Engine>() {
...
+    let start_time = if let Some(quota_limiter) = &quota_limiter {
+        quota_limiter.get_now_time()
+    } else {
+        Box::new(Instant::now())
+    };
+
     let (callback, future) =
         tikv_util::future::paired_must_called_future_callback(drop_snapshot_callback::<E>);
     let val = engine.async_snapshot(ctx, callback);
+    if let Some(quota_limiter) = quota_limiter {
```
tonyxuqqi, line 472: two issues here:
1) async_snapshot is not actually finished at L472, the duration collected is wrong. 
2) cannot use std::thread::sleep as it will make the whole thread sleep.

components/raftstore/src/store/config.rs
```diff
@@ -357,6 +359,8 @@ impl Default for Config {
             inspect_interval: ReadableDuration::millis(500),
             check_leader_lease_interval: ReadableDuration::secs(0),
             renew_leader_lease_advance_duration: ReadableDuration::secs(0),
+            // TODO: add comments @tonyxuqqi
+            max_batch_size: 64,
```
tonyxuqqi, line 363: change it to default 256

components/server/src/server.rs
```diff
@@ -256,6 +258,14 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         let latest_ts = block_on(pd_client.get_tso()).expect("failed to get timestamp from PD");
         let concurrency_manager = ConcurrencyManager::new(latest_ts);
 
+        let quota_limiter = Arc::new(QuotaLimiter::new(
+            config.quota.cpu,
+            config.quota.write_kvs,
```
tonyxuqqi, line 263: For simplicity, let's remove write_kvs, total_qps  for now.

components/tidb_query_executors/src/runner.rs
```diff
@@ -438,14 +443,17 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
+            let (drained, record_len, quota_delay) = self.internal_handle_request(
                 false,
                 batch_size,
                 &mut chunk,
                 &mut warnings,
                 &mut ctx,
             )?;
 
+            if !quota_delay.is_zero() {
+                std::thread::sleep(quota_delay);
```
tonyxuqqi, line 455: do not use thread::sleep.

components/tidb_query_executors/src/runner.rs
tonyxuqqi: I think it's better to use async style sleep. The reason is that it's very expensive to frequently call OS's sleep as it will do context switch. Second is that we may want to let some requests run as the thread may run different requests. Third,  by the test it turns out that async sleep works pretty well. 
The way the async sleep still works is to slow down the requests so that the caller side will slow down. Because the caller won't send unlimited requests to TiKV without waiting current ones back.

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,202 @@
...
+        };
+
+        let req_rate_dur = self.req_rate_limiter.consume_duration(req_cnt);
+
+        std::cmp::max(req_rate_dur, std::cmp::max(cpu_dur, bw_dur))
+    }
+
+    // If `cputime_limiter` is set to INFINITY, use `CLOCK_MONOTONIC` to save cost.
+    pub fn get_now_time(&self) -> Box<dyn QuotaTimer> {
+        if self.cputime_limiter.speed_limit() == f64::INFINITY {
```
tonyxuqqi, line 138: use is_finite instead.

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    pub fn consume_write(
+        &self,
+        req_cnt: usize,
+        kv_cnt: usize,
+        bytes: usize,
+        cpu_micro_secs: usize,
+    ) -> Duration {
+        let cpu_dur = self.cputime_limiter.consume_duration(cpu_micro_secs);
```
tonyxuqqi, line 91: add check to totally skip consume_duration for better performance when cpu limiter is infinity.

same as other limiters.

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,202 @@
...
+        )
+    }
+
+    pub fn consume_read(
+        &self,
+        req_cnt: usize,
+        read_bytes: usize,
+        cpu_micro_secs: usize,
+    ) -> Duration {
+        let cpu_dur = self.cputime_limiter.consume_duration(cpu_micro_secs);
```
tonyxuqqi, line 123: add check to totally skip consume_duration for better performance when cpu limiter is infinity.

same as other limiters.

components/tikv_kv/src/lib.rs
```diff
@@ -456,10 +458,24 @@ pub unsafe fn destroy_tls_engine<E: Engine>() {
 pub fn snapshot<E: Engine>(
     engine: &E,
     ctx: SnapContext<'_>,
+    quota_limiter: Option<Arc<QuotaLimiter>>,
```
tonyxuqqi, line 461: I think it might be cleaner to add quota_limiter to SnapContext

components/tikv_util/src/quota_limiter.rs
hunterlxt: explanation in https://github.com/tikv/tikv/pull/12151#issuecomment-1066823807

components/tikv_util/src/quota_limiter.rs
hunterlxt: same in https://github.com/tikv/tikv/pull/12151#issuecomment-1066823807

components/server/src/server.rs
hunterlxt: finished

components/raftstore/src/store/config.rs
```diff
@@ -573,10 +576,8 @@ impl Config {
                     "store-max-batch-size should be greater than 0 and less than or equal to 10240"
                 ));
             }
-        } else if self.hibernate_regions {
```
sticnarf, line 576: The default config is changed when hibernate regions is off?

components/tidb_query_executors/src/runner.rs
```diff
@@ -511,7 +528,8 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
         // record count less than batch size and is not drained
         while record_len < self.stream_row_limit && !is_drained {
             let mut current_chunk = Chunk::default();
-            let (drained, len) = self.internal_handle_request(
+            // for now `handle_streaming_request` don't take up significant cpu resources.
```
sticnarf, line 531: You can leave it as a TODO. Streaming coprocessor on TiKV is just not enabled in TiDB now.

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub fn get_now_timer(&self) -> Box<dyn Timer> {
+        if self.cputime_limiter.speed_limit().is_infinite() {
+            Box::new(Instant::now_coarse())
+        } else {
+            Box::new(ThreadTime::now())
+        }
+    }
+}
+
+pub trait Timer {
```
sticnarf, line 121: I prefer to just use an enum to avoid adding allocation for each `get_now_timer`.

components/raftstore/src/store/config.rs
```diff
@@ -264,6 +264,8 @@ pub struct Config {
     // Interval to inspect the latency of raftstore for slow store detection.
     pub inspect_interval: ReadableDuration,
 
+    pub max_batch_size: usize,
```
tabokie, line 267: What is this for?

src/config.rs
```diff
@@ -2487,6 +2523,9 @@ pub struct TiKvConfig {
     #[online_config(skip)]
     pub log: LogConfig,
 
+    #[online_config(skip)]
```
tabokie, line 2526: Should be configurable.

src/config.rs
```diff
@@ -2432,6 +2432,42 @@ impl LogConfig {
...
+            return Err("Max cpu quota is limited to 96000, it means 96vCPU"
+                .to_string()
+                .into());
+        }
+        if self.write_bandwidth.0 > ReadableSize::mb(200).0 {
+            return Err("Max write bandwidth is limited to 200MB".to_string().into());
+        }
+        if self.read_bandwidth.0 > ReadableSize::gb(2).0 {
+            return Err("Max read bandwidth is limited to 2GB".to_string().into());
+        }
```
tabokie, line 2466: I think these restrictions are unnecessary.

src/storage/txn/scheduler.rs
```diff
@@ -786,6 +795,18 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+        if !quota_delay.is_zero() {
+            TXN_COMMAND_THROTTLE_TIME_COUNTER_VEC_STATIC
+                .get(tag)
+                .inc_by(quota_delay.as_micros() as u64);
+            GLOBAL_TIMER_HANDLE
+                .delay(std::time::Instant::now() + quota_delay)
+                .compat()
+                .await
+                .unwrap();
+        }
```
sticnarf, line 808: This looks not so accurate.
1. The counted written bytes will be larger than actual because no data will be written if a pessimistic lock is encountered or pessimistic locks are written only into the memory.
2. The cpu time is not accurate either. If an error is returned in L786, it is not counted.

Both look not very important, but better to leave a comment here.

src/storage/mod.rs
```diff
@@ -936,6 +976,29 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
 
+                    let result_len = result
+                        .as_ref()
+                        .unwrap_or(&Vec::new())
+                        .iter()
+                        .fold(0, |acc, v| {
+                            acc + v.as_ref().unwrap_or(&(Vec::new(), Vec::new())).1.len()
+                        });
+
+                    let read_bytes = key_bytes + result_len;
```
sticnarf, line 988: Why not use the data in `stats`?

src/storage/txn/scheduler.rs
```diff
@@ -735,6 +738,8 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
     /// message if successful or a `FinishedWithErr` message back to the `Scheduler`.
     async fn process_write(self, snapshot: E::Snap, task: Task, statistics: &mut Statistics) {
         fail_point!("txn_before_process_write");
+        let quota_limiter = Arc::clone(&self.inner.quota_limiter);
```
tabokie, line 741: Why clone?

components/raftstore/src/store/config.rs
hunterlxt: ok I'll set 1024 for `hibernate_regions` disabled but config value for enabled.

components/tidb_query_executors/src/runner.rs
```diff
@@ -438,14 +445,24 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
-            let (drained, record_len) = self.internal_handle_request(
+            let (drained, record_len, quota_delay) = self.internal_handle_request(
                 false,
                 batch_size,
                 &mut chunk,
                 &mut warnings,
                 &mut ctx,
             )?;
 
+            if !quota_delay.is_zero() {
```
tabokie, line 456: IMO you should put these boilerplate codes inside the limiter. E.g. Returns a new `Throttle` object instead of raw duration. Maybe pass in a reference to the prometheus `IntCounter` to get around the metrics stuff.

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,176 @@
...
+        let should_delay = if max_dur > cpu_time {
+            max_dur - cpu_time
+        } else {
+            Duration::ZERO
+        };
+        std::cmp::min(MAX_QUOTA_DELAY, should_delay)
+    }
+
+    // If `cputime_limiter` is set to INFINITY, use `CLOCK_MONOTONIC_COARSE` to save cost.
+    pub fn get_now_timer(&self) -> Box<dyn Timer> {
```
tabokie, line 112: Could returning an `Option` be enough?

components/tikv_util/src/quota_limiter.rs
hunterlxt: Fixed this by new type `QuotaTimer`

components/tikv_util/src/quota_limiter.rs
hunterlxt: fixed in https://github.com/tikv/tikv/pull/12151#discussion_r827606925

src/config.rs
hunterlxt: cc @tonyxuqqi

src/config.rs
hunterlxt: you mean supporting online config?

tabokie: Yep

components/raftstore/src/store/config.rs
hunterlxt: In our bench tests, sometimes it's useful to set larger

tabokie: Put it in another PR then, that gives a cleaner commit history.

src/storage/mod.rs
hunterlxt: Good suggestion, fixed

src/storage/txn/scheduler.rs
hunterlxt: fixed

components/tidb_query_executors/src/runner.rs
hunterlxt: fixed your first part

components/raftstore/src/store/config.rs
hunterlxt: Ok, working on it

src/config.rs
hunterlxt: agree but maybe this will be another PR, let's make this small and easy to merge (exceeded 1k lines)

src/storage/txn/scheduler.rs
hunterlxt: For question 2, most of cpu resource will be used in the above block, so it's doesn't matter.

image

components/tidb_query_executors/src/runner.rs
tabokie: I was thinking something like this, the timing stuff is hidden:
```rust
let throttle = limiter.new_throttle();
/* do the work */
throttle.read_bytes(stats.read);
throttle.read_bytes(stats.write);
throttle.async_consume(&TXN_COMMAND_THROTTLE_TIME_COUNTER_VEC_STATIC.get(CMD)).await;
```

tonyxuqqi:
LGTM

components/raftstore/src/store/config.rs
hunterlxt: https://github.com/tikv/tikv/pull/12151#discussion_r827610789 here decided to remove it to another PR

components/tidb_query_executors/src/runner.rs
hunterlxt: most agree but just async_consume.await will bring unnecessary switch. How about like this: 
```
let sample = limiter.new_sample();
/* do the work */
sample.read_bytes(stats.read);
sample.write_bytes(stats.write);
if sample.is_throttled() {
  limiter.async_consume(sample, &TXN_COMMAND_THROTTLE_TIME_COUNTER_VEC_STATIC.get(CMD)).await;
}
```

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        let bw_dur = if bytes > 0 {
+            self.write_bandwidth_limiter.consume_duration(bytes)
+        } else {
+            Duration::ZERO
+        };
+
+        let max_dur = std::cmp::max(cpu_dur, bw_dur);
+        let should_delay = if max_dur > cpu_time {
+            max_dur - cpu_time
```
Connor1996, line 108: why

src/storage/txn/scheduler.rs
```diff
@@ -786,6 +794,17 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
         };
         SCHED_STAGE_COUNTER_VEC.get(tag).write.inc();
 
+        let throttle = self
+            .inner
+            .quota_limiter
+            .consume_write(write_bytes, cost_time);
```
Connor1996, line 800: if the result is a error, it's early return above, then the cost time is not recorded here

components/tidb_query_executors/src/runner.rs
tabokie: You can do the conditional wait inside `async_consume`.

src/storage/txn/scheduler.rs
tabokie: @hunterlxt I think we can solve this by implementing `Drop` for the sample in https://github.com/tikv/tikv/pull/12151#discussion_r827676224, in which the duration will be consumed from limiter (but no delay is triggered).

components/tikv_util/src/quota_limiter.rs
hunterlxt: this is a rate limiter to control bandwidth, but for cpu, we need small adjustment. If cpu used 500ms but quota to 1000ms, we don't need limit because cpu already waste real 500ms as its cost. So it only need to compensate for the insufficient buckets.

src/storage/txn/scheduler.rs
hunterlxt: ok i'll move this block earlier

components/tidb_query_executors/src/runner.rs
hunterlxt: after bench more voluntary switch if use this style

components/tikv_util/src/quota_limiter.rs
hunterlxt: this is the exact argument point between xuqi and me, for now only sync sleep actually waste real time and for async mode, I prefer call the `forefront-cpu-time` a indicator, it's not represent real cpu time. You can refer 6th pic in the description

hunterlxt: anyway in async mode both calculation methods are not fully correct. But current method is more in line with the description of the token bucket algorithm.

hunterlxt: btw, I would like to express that even use the old method, it is still impossible to compensate correctly in asynchronous mode, and now we are using a fixed number to multiply the config parameter, which is completely from our experience.

Connor1996: understand

hunterlxt:
@tabokie and I did an exploration to impl `Drop` for `Throttle` and wrap most logic inside to reduce redundant code. But after trying to change, I found that the delay of throttle will be enforced, even if sometimes we just want to count the time and not delay. In addition, metric code cannot be moved into `async_consume` because I record 2 different types of metric. Later we made some compromises to avoid exposing the timing method to the outside and unify consume code. This is why the code is what it looks like now.

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,225 @@
...
+// completion time of tasks through restrictions of different metrics.
+#[derive(Debug)]
+pub struct QuotaLimiter {
+    cputime_limiter: Limiter,
+    write_bandwidth_limiter: Limiter,
+    read_bandwidth_limiter: Limiter,
+}
+
+// Throttle must be consumed in quota limiter.
+pub struct Throttle {
```
tabokie, line 32: Considering the context,
```suggestion
pub struct Sample {
```

src/storage/mod.rs
```diff
@@ -608,6 +622,20 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
                         .get(CMD)
                         .observe(command_duration.saturating_elapsed_secs());
 
+                    let read_bytes = key.len()
+                        + result
+                            .as_ref()
+                            .unwrap_or(&None)
+                            .as_ref()
+                            .map_or(0, |v| v.len());
```
tabokie, line 630:
```suggestion
                            .map_or(0, |v| v.len());
```

etc/config-template.toml
```diff
@@ -42,6 +42,17 @@
...
+[quota]
+## Quota is use to add some limitation for the read write flow and then
+## gain predictable stable performance.
+## CPU quota for these front requests can use, default value is 0, it means unlimited.
+## The unit is millicpu which 1000 means 1vCPU and this config is approximate.
+# forefront-cpu-time = 0
+## Write bandwidth limitation for this TiKV instance, default value is 0 which means unlimited.
+# write-bandwidth = "0B"
+## Read bandwidth limitation for this TiKV instance, default value is 0 which means unlimited
+# read-bandwidth = "0B"
```
tabokie, line 54: I don't often see the word "forefront" in a technical context. How about all use the "foreground" prefix, or drop the prefix.

tabokie (commented):
Rest looks good.

src/storage/mod.rs
hunterlxt: I see errors like this: 
```
no method named `len` found for reference `&std::option::Option<std::vec::Vec<u8>>` in the current scope
method not found in `&std::option::Option<std::vec::Vec<u8>>
```

etc/config-template.toml
hunterlxt: fixed

src/storage/mod.rs
tabokie: Okay.

tabokie (approved):

components/tidb_query_executors/Cargo.toml
```diff
@@ -24,6 +24,7 @@ collections = { path = "../collections" }
 tipb = { git = "https://github.com/pingcap/tipb.git" }
 yatp = { git = "https://github.com/tikv/yatp.git", branch = "master" }
 itertools = "0.10"
+futures = { version = "0.3", features = ["thread-pool", "compat"] }
```
sticnarf, line 27: Where is the `thread-pool` feature used?

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,221 @@
...
+            enable_cpu_limit: !self.cputime_limiter.speed_limit().is_infinite(),
+        }
+    }
+
+    // To consume a sampler and return delayed duration.
+    // If the sampler is null, the speed limiter will just return ZERO.
+    pub async fn async_consume(&self, th: Sample) -> Duration {
+        let cpu_dur = self
+            .cputime_limiter
+            .consume_duration(th.cpu_time.as_micros() as usize);
```
sticnarf, line 139: Is it a good idea to check `cpu_time` first?

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,220 @@
...
+            read_bytes: 0,
+            write_bytes: 0,
+            cpu_time: Duration::ZERO,
+            enable_cpu_limit: !self.cputime_limiter.speed_limit().is_infinite(),
+        }
+    }
+
+    // To consume a sampler and return delayed duration.
+    // If the sampler is null, the speed limiter will just return ZERO.
+    pub async fn async_consume(&self, th: Sample) -> Duration {
```
sticnarf, line 136: What does the name `th` mean?

components/tikv_util/src/metrics/mod.rs
```diff
@@ -47,13 +48,37 @@ pub fn dump() -> String {
...
+}
+
 lazy_static! {
     pub static ref CRITICAL_ERROR: IntCounterVec = register_int_counter_vec!(
         "tikv_critical_error_total",
         "Counter of critical error.",
         &["type"]
     )
     .unwrap();
+    pub static ref NON_TXN_COMMAND_THROTTLE_TIME_COUNTER_VEC: IntCounterVec =
```
sticnarf, line 71: I suggest adding a comment about its unit (microsecond)

src/storage/txn/scheduler.rs
```diff
@@ -772,6 +779,15 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
                 .process_write(snapshot, context)
                 .map_err(StorageError::from)
         };
+
+        sample.add_write_bytes(write_bytes);
+        let quota_delay = quota_limiter.async_consume(sample).await;
```
sticnarf, line 784: As I commentted before, the write bytes can be a bit inaccurate due to error requests or in-memory pessimistic locks. Let's leave a todo comment here.

sticnarf (commented):
Mostly lgtm

components/tikv_util/src/quota_limiter.rs
```diff
@@ -0,0 +1,220 @@
...
+
+// To avoid long tail latency.
+const MAX_QUOTA_DELAY: Duration = Duration::from_secs(1);
+
+// Quota limiter allows users to obtain stable performance by increasing the
+// completion time of tasks through restrictions of different metrics.
+#[derive(Debug)]
+pub struct QuotaLimiter {
+    cputime_limiter: Limiter,
+    write_bandwidth_limiter: Limiter,
```
Connor1996, line 27: why do we need a write/read limiter here, I think io rate limiter is enough. /cc @tabokie

Connor1996 (commented):
why do we need a write/read limiter here, I think io rate limiter is enough. /cc @tabokie

components/tikv_util/src/quota_limiter.rs
hunterlxt: what's your point

components/tikv_util/src/quota_limiter.rs
hunterlxt: th for throttle, but in last comment it be suggested from `Throttle` to `Sample`, fixed

components/tikv_util/src/quota_limiter.rs
sticnarf: The frontend read bandwidth may be much larger than the real disk read because of the existence of block cache. I'm not sure if that's the cause of a separate read limit.

components/tikv_util/src/quota_limiter.rs
sticnarf: Do not call `consume_duration` if `cpu_time` is zero like read or write bytes.

src/storage/mod.rs
```diff
@@ -577,17 +587,19 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
-                        ctx.get_isolation_level(),
-                        !ctx.get_not_fill_cache(),
-                        bypass_locks,
-                        access_locks,
-                        false,
-                    );
-                    let result = snap_store
+                    let (result, delta) = {
+                        let _guard = sample.observe_cpu();
+                        let perf_statistics = PerfStatisticsInstant::new();
```
Connor1996, line 592: move it out of the brace

components/tikv_util/src/quota_limiter.rs
hunterlxt: fixed

src/storage/mod.rs
```diff
@@ -886,52 +915,56 @@ impl<E: Engine, L: LockManager> Storage<E, L> {
...
-                                })
-                                .collect();
-                            KV_COMMAND_KEYREAD_HISTOGRAM_STATIC
-                                .get(CMD)
-                                .observe(kv_pairs.len() as f64);
-                            kv_pairs
-                        });
-
-                    let delta = perf_statistics.delta();
+                    let (result, delta, stats) = {
```
Connor1996, line 922: a little verbose, how about using drop(_guard), so no need to use brace

hunterlxt: rust compiler won't let this happen

hunterlxt: send check failed

src/storage/mod.rs
hunterlxt:
```
 = help: within `InTags<impl futures::Future<Output = [async output]>>`, the trait `std::marker::Send` is not implemented for `tikv_kv::PerfStatisticsInstant`
note: future is not `Send` as this value is used across an await
    --> src/storage/mod.rs:981:74
     |
922  |                     let perf_statistics = PerfStatisticsInstant::new();
     |                         --------------- has type `tikv_kv::PerfStatisticsInstant` which is not `Send`
...
981  |                     let quota_delay = quota_limiter.async_consume(sample).await;
     |                                                                          ^^^^^^ await occurs here, with `perf_statistics` maybe used later
...
1011 |                 }
     |                 - `perf_statistics` is later dropped here
note: required by a bound in `ReadPoolHandle::spawn_handle`
    --> src/read_pool.rs:139:33
     |
132  |     pub fn spawn_handle<F, T>(
     |            ------------ required by a bound in this
...
139  |         F: Future<Output = T> + Send + 'static,
     |                                 ^^^^ required by this bound in `ReadPoolHandle::spawn_handle`
```

src/storage/txn/scheduler.rs
```diff
@@ -772,6 +779,16 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
                 .process_write(snapshot, context)
                 .map_err(StorageError::from)
         };
+
+        // TODO: write bytes can be a bit inaccurate due to error requests or in-memory pessimistic locks.
```
Connor1996, line 783: if the result is a error, no need to add write bytes. And do we need to add read bytes?

hunterlxt: here are process_write, don't need to add front read bytes in my view

components/tikv_util/src/quota_limiter.rs
tabokie: I do think bandwidth is somewhat misleading. When using the quota limiter, we don't really care about how many physical IOs (both buffered and on disk) are involved. We are just (blindly) tuning these knobs for better results. It's better to think of them as request size rather than IO resources.

src/storage/txn/scheduler.rs
sticnarf: Ah, that's a problem. In the scheduler itself, most of its work is reading the engine. We probably should record the read bytes comtributed by the scheduler

hunterlxt: Can you give a suggestion where to record

hunterlxt: fetch these part of data from `statistics`

hunterlxt: fixed

sticnarf (approved):

Connor1996 (approved):
LGTM

components/tikv_util/src/quota_limiter.rs
tonyxuqqi: > 

good point. The name of these two limiter may be renamed to write_request_size_limiter, read_response_size_limiter..

tonyxuqqi: > why do we need a write/read limiter here, I think io rate limiter is enough. /cc @tabokie

The IO rate limiter is more about eliminate low priority task's impact to high priority task, per my understand. Here the read/write limiter is more about control frontend traffic. 
This is just in case CPU is not the bottleneck (but IO is).

Issue #12088, opened 2022-03-03
tracker.rs doesn't log the slow request if it exceeds the deadline 


coderplay:
## Bug Report

If a coprocessor task takes longer than 1 minute, it can't be logged by tracker slowlog.

https://github.com/tikv/tikv/blob/master/src/coprocessor/tracker.rs#L256

### What did you expect?

We want to get more information from the slow log to see why it take more than 1 minute.

youjiali1995:
Good catch. We will improve it.

JayiceZ:
> Good catch. We will improve it.

Can I try to fix it?

youjiali1995:
@JayiceZ Of course.

sticnarf:
Per https://github.com/tikv/tikv/pull/12154#issuecomment-1067610392, personally I think it not a good idea to simply log the request as long as the request exceeds the deadline. This may cause log explosion under heavy load.

Instead, I'm thinking about returning enough information to TiDB and let TiDB aggregate and log the slow query. For example, how many requests are dropped due to exceeding the deadline, how much cpu time and i/o is wasted on these aborted executions.

coderplay:
Agreed on the aggregated summary to avoid the log flooding. However, why not do the aggregation in-place on the current  TiKV node?  There are basically multiple advantages doing it on TiKV node 1) saving a networking communications between TiKV and TiDB 2) in case there is no TiDB, user just run TiKV  3) in case a single node is busy while the others are idle because of data skew/hotspot,  per TiKV node level aggregation should be more useful.

sticnarf:
A node level aggregation can be done through the metrics. We will try improving the Grafana dashboards.

To aggregate query-level information, we have to return these statistics to TiDB, because a TiDB query consists of multiple coprocessor requests.

The tracker only works for the tidb_query coprocessor. It's not useful for KV users. Anyway, it is better to provide the similar statistics in KV interfaces.

PR #12154, opened 2022-03-14
copr:add slow log when task exceeded ddl

JayiceZ:

Issue Number: Close [#xxx](https://github.com/tikv/tikv/issues/12088)

What's Changed:

```commit-message
In the old implementation, the copr task that exceeded the deadline won't be logged in slow log. This PR can improve it.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```

### DIFF
--- a/src/coprocessor/endpoint.rs
+++ b/src/coprocessor/endpoint.rs
@@ -366,7 +366,7 @@ impl<E: Engine> Endpoint<E> {
         // When this function is being executed, it may be queued for a long time, so that
         // deadline may exceed.
         tracker.on_scheduled();
-        tracker.req_ctx.deadline.check()?;
+        tracker.check_deadline()?;
 
         // Safety: spawning this function using a `FuturePool` ensures that a TLS engine
         // exists.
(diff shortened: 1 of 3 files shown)


youjiali1995:
I just realized that we intended to just take the process time into account(https://github.com/tikv/tikv/pull/10858). PTAL, @sticnarf .

sticnarf:
> I just realized that we intended to just take the process time into account(#10858). PTAL, @sticnarf .

Yes, it was intended to only print slow logs for the requests that really take a long time to run.

If the TiKV is under extreme pressure, most of the requests will just exceed the deadline. Then, there can be so many slow logs to be printed that they even occupy the resource for execution, and it may cause an avalanche.

JayiceZ:
> I just realized that we intended to just take the process time into account(#10858). PTAL, @sticnarf .

Got it. Then I close this PR?😂

sticnarf:
> Got it. Then I close this PR?

The PR itself does not seem like a perfect solution in my opinion. Sorry for that.

Or, we can think about other ways to solve the original issue.

coderplay:
>Then, there can be so many slow logs to be printed that they even occupy the resource for execution, and it may cause an avalanche.

It is being flooded by the logs `Coprocessor task terminated due to exceeding the deadline` already, shouldn't be a big deal to add another one?

PR #12156, opened 2022-03-14
cdc: don't emit resolved timestamps before scan

hicqu:

Issue Number: Close https://github.com/pingcap/tiflow/issues/4782

CDC shouldn't emit resolved timestamps before incremental scan finished. https://github.com/tikv/tikv/pull/11098 tried to fix the problem but the implementation is wrong. This patch gives a correct solution.

```commit-message
cdc: don't emit resolved timestamps before scan
```

### Related changes

### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/cdc/src/delegate.rs
+++ b/components/cdc/src/delegate.rs
@@ -67,7 +67,7 @@ impl Default for DownstreamState {
 }
 
 /// Shold only be called when it's uninitialized or stopped. Return false if it's stopped.
-pub fn on_init_downstream(s: &AtomicCell<DownstreamState>) -> bool {
+pub(crate) fn on_init_downstream(s: &AtomicCell<DownstreamState>) -> bool {
     s.compare_exchange(
         DownstreamState::Uninitialized,
         DownstreamState::Initializing,
@@ -76,7 +76,7 @@ pub fn on_init_downstream(s: &AtomicCell<DownstreamState>) -> bool {
 }
 
 /// Shold only be called when it's initializing or stopped. Return false if it's stopped.
-pub fn post_init_downstream(s: &AtomicCell<DownstreamState>) -> bool {
+pub(crate) fn post_init_downstream(s: &AtomicCell<DownstreamState>) -> bool {
     s.compare_exchange(DownstreamState::Initializing, DownstreamState::Normal)
         .is_ok()
 }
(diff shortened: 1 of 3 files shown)


components/cdc/src/endpoint.rs
```diff
@@ -1253,23 +1257,24 @@ impl<E: KvEngine> Initializer<E> {
...
+                "conn_id" => ?conn_id);
+            Err(box_err!("scan canceled"))
+        };
 
         while !done {
             // When downstream_state is Stopped, it means the corresponding
             // delegate is stopped. The initialization can be safely canceled.
             if self.downstream_state.load() == DownstreamState::Stopped {
-                on_cancel!();
+                on_cancel()?;
```
overvenus, line 1277: What about `return on_cancel();`? So we know for sure that it is returned here.

components/cdc/tests/failpoints/test_endpoint.rs
```diff
@@ -251,52 +251,33 @@ fn test_no_resolved_ts_before_downstream_initialized() {
...
-        let (mut req_tx, event_feed, receive_event) =
-            new_event_feed(suite.get_region_cdc_client(region.id));
+        let (mut req_tx, event_feed, _) = new_event_feed(suite.get_region_cdc_client(region.id));
         let req = suite.new_changedata_request(region.id);
         block_on(req_tx.send((req, WriteFlags::default()))).unwrap();
         req_txs.push(req_tx);
         event_feeds.push(event_feed);
-        receive_events.push(receive_event);
         // Sleep a while to wait the capture has been initialized.
         thread::sleep(Duration::from_secs(1));
```
overvenus, line 268: After sleep could you wait a resolved ts event? So we known it is initialized.

components/cdc/src/delegate.rs
```diff
@@ -61,6 +66,40 @@ impl Default for DownstreamState {
     }
 }
 
+/// Shold only be called when it's uninitialized or stopped. Return false if it's stopped.
+pub fn on_init_downstream(s: &AtomicCell<DownstreamState>) -> bool {
```
hi-rustin, line 70: I think the visibility of these methods could be smaller, for example they should be pub(crate). Using smaller visibility would help us to find dead code.

hi-rustin (approved):

overvenus (approved):
LGTM

PR #12177, opened 2022-03-15
split_controller: check all_key_ranges_num to reduce the sample overhead

JmPotato:

In some cases, we may don't have enough key ranges to sample here. If we still use the RNG to randomly sample, it will cost some unnecessary overhead. This PR makes it check `all_key_ranges_num` and `sample_num` first. If the previous one is less, `sample` function will return all key ranges directly.

```commit-message
split_controller: check all_key_ranges_num to reduce the sample overhead
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```


AndreMouche (commented):
Please run `make clippy`

AndreMouche (commented):
LGTM

lhy1024 (approved):

lhy1024 (approved):

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -55,31 +55,51 @@ where
     .collect()
 }
 
+#[inline(always)]
+fn prefix_sum_mut<F, T>(iter: IterMut<'_, T>, read: F) -> Vec<usize>
```
lhy1024, line 59: Do we need to remove function `prefix_sum`?

JmPotato: No. `prefix_sum` is still used elsewhere.

AndreMouche (approved):
LGTM

PR #12181, opened 2022-03-16
do not force compact unreplicated entries

### LINKED ISSUE #12161: raft log lag is more and more when down one tikv for 38min
[2022/03/15 10:09:22.527 +08:00] [INFO] [client.go:376] ["Cluster version information"] [type=pd] [version=6.0.0-nightly] [git_hash=38ee0a8f7eeeaefbb43ccb708b1bb4a074d55594]
[2022/03/15 10:09:22.527 +08:00] [INFO] [client.go:376] ["Cluster version information"] [type=tikv] [version=6.0.0-alpha] [git_hash=b394b4d1e88b401c58e1255da043300c00cbe5c7]
[2022/03/15 10:09:22.527 +08:00] [INFO] [client.go:376] ["Cluster version information"] [type=tidb] [version=6.0.0-nightly] [git_hash=3f3a061afe93689ece19e43915a08f393ea83582]

8core、16G

### Steps to reproduce
1、[2022/03/15 10:09:22.647 +08:00] sysbench --mysql-user=root --db-driver=mysql --time=0 --report-interval=10 --mysql-db=test --mysql-host=tc-tidb.glh-master-64gs8 --mysql-port=4000 --threads=512 --tables=32 --table-size=10000000 --mysql-ignore-errors=2013,1213,1105,1205,8022,8027,8028,9004,9007 --rand-type=gaussian oltp_insert run
2、[2022/03/15 10:20:22.911 +08:00] [INFO] [chaos.go:358] ["fault will last for"] [duration=38m0s]
[2022/03/15 10:20:22.911 +08:00] [INFO] [util.go:59] ["Wait QPS recover after inject fault."] [wait=2m0s] [callBy=oltp_rel_001_006_test.go:79]
[2022/03/15 10:20:22.917 +08:00] [INFO] [chaos.go:86] ["Run chaos"] [name=failure] [selectors="[glh-master-64gs8/tc-tikv-2]"] [experiment="{\"Duration\":\"\",\"Scheduler\":null}"]

### What did you expect?
after fault recover，the raft log lag can be less and less and can be a little value in short time

### What did happened?
the raft log lag is more and more
image

tabokie:

Issue Number: Close #12161

What's Changed:

Do not force compact entries that aren't replicated yet,

### Related changes

### Check List 

Tests
- Unit test
- Integration test

Side effects

There will be more background writes when some TiKVs are down.

### Release note 

```release-note
None
```


Connor1996 (commented):
LGTM, let's if this resolves the issue

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4417,13 +4417,13 @@ where
 
         let first_idx = self.fsm.peer.get_store().first_index();
 
-        let mut compact_idx = if force_compact
-            // Too many logs between applied index and first index.
-            || (applied_idx > first_idx && applied_idx - first_idx >= self.ctx.cfg.raft_log_gc_count_limit)
-            // Raft log size ecceeds the limit.
+        let mut compact_idx = if force_compact && replicated_idx > first_idx {
```
tonyxuqqi, line 4420: For the lag log issue, why RocksDB works without this change?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4417,13 +4417,13 @@ where
...
+            replicated_idx
+        } else if (applied_idx > first_idx
```
tonyxuqqi, line 4422: When code hits this line, the compact_index will still be larger than replicated_idx. Will it still have problem for RaftEngine?

components/raftstore/src/store/fsm/peer.rs
tabokie: Only raft engine triggers force compaction, so there won’t be any unexpected  side effect.

components/raftstore/src/store/fsm/peer.rs
tabokie: What this PR does is to avoid frequent GC when some TiKVs are down. The GC triggered by size limit will be infrequent.

Connor1996 (approved):
LGTM

you06:
@hunterlxt PTAL

5kbpers (approved):

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: > Only raft engine triggers force compaction, so there won’t be any unexpected side effect.
IC. It will be better to add comment on parameter 'force_compact' that only raft engine uses this flag.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: > What this PR does is to avoid frequent GC when some TiKVs are down. The GC triggered by size limit will be infrequent.

Understand. Could you explain why frequent GC triggers the problem, but infrequent GC does not? Also It will be better to add the some brief explanation why the PR is going to fix the problem.

tabokie: Log lag happens when some log entries are not replicated to the follower. If those log entries are compacted in leader, then we are forced to use snapshot to synchronize log, which is costly. When snapshot transfer throughput cannot catch up with incoming foreground writes, log lag can never recover.

This PR makes sure that force compaction can never affect the Raft log catch up behavior, by not deleting any unreplicated entries.

tonyxuqqi: For infrequent GC due to its size, could it still delete unreplicated entries and lead to exactly same problem?
I know the probability of it is lower than force compaction, but still possible?

tabokie: That's the same behavior as before. It didn't cause problem then, won't cause problem now. To be more specific, it will only be an issue when snapshot sending rate is slower than GC rate, which isn't likely for size-induced GC.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4439,13 +4439,13 @@ where
...
+            && applied_idx - first_idx >= self.ctx.cfg.raft_log_gc_count_limit)
             || (self.fsm.peer.raft_log_size_hint >= self.ctx.cfg.raft_log_gc_size_limit.0)
         {
-            std::cmp::max(first_idx + (last_idx - first_idx) / 4, replicated_idx)
+            std::cmp::max(first_idx + (last_idx - first_idx) / 2, replicated_idx)
```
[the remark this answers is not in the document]
cosven: Fine, I see there is a check in `exec_compact_log`, which ensures the compact_idx is never large than the applied_idx.

PR #12186, opened 2022-03-16
raftstore:Fix memory metrics overflow on raft msgs

### LINKED ISSUE #12160: raftstore: Integer overflow problems on memory metrics

$./tikv-server --version
TiKV 
Release Version:   6.0.0-alpha
Edition:           Community
Git Commit Hash:   ce39ef6b52a050c594b2b01e8d60a6e11853faec
Git Commit Branch: fix-log-entry-oom-t
UTC Build Time:    2022-03-14 11:39:21
Enable Features:   jemalloc mem-profiling portable sse test-engines-rocksdb cloud-aws cloud-gcp cloud-azure
Profile:           release

### Steps to reproduce

1. sysbench insert
### What did you expect?
1.  no overflow.
### What did happened?
1. The metrics on raft messages overflow.
2. master of 0314 + trace log
0W6TTjgjYc
EkRt8BAfRe
feOg8VakKq

tier-cap:

Issue Number: Close #12160

What's Changed:

```commit-message
Fix raft msg memory metrics overflow.
```

### Related changes

### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
Solve the problem of raft msg memory metrics overflow, which will cause intermittent packet loss and oom.
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1745,6 +1745,16 @@ where
...
         defer!({
+            let do_check = || {
+                fail_point!("memtrace_raft_messages_overflow_check_peer_recv", |_| {
+                    true
+                });
+                false
+            };
+            if do_check() {
+                let prev = MEMTRACE_RAFT_MESSAGES.sum();
+                assert!(prev >= heap_size);
```
5kbpers, line 1756: Better to replace this with `debug_assert!`

components/raftstore/src/store/worker/read.rs
```diff
@@ -520,16 +524,26 @@ where
             // The local `ReadDelegate` is up to date
             Some(d) if !d.track_ver.any_new() => Some(Arc::clone(d)),
             _ => {
+                if self.get_must_track_version_changed_failpoint() {
+                    if let Some(d) = self.delegates.get(&region_id) {
+                        info!("must_track_version_changed: {}", d.track_ver.any_new());
```
5kbpers, line 529: What does this failpoint suppose to do? Can we just add `debug` logs?

NingLin-P, line 529: What is this log for? While entering this branch already show that the previous `track_ver.any_new()` is true,  call it again immediately usually return false

components/raftstore/src/store/worker/read.rs
```diff
@@ -520,16 +524,26 @@ where
...
                 let (meta_len, meta_reader) = {
                     let meta = self.store_meta.lock().unwrap();
                     (
                         meta.readers.len(),
                         meta.readers.get(&region_id).cloned().map(Arc::new),
                     )
                 };
+                if self.get_must_track_version_changed_failpoint() {
+                    info!("store meta lock acquire takes {:?}", t.saturating_elapsed());
+                }
```
NingLin-P, line 546: Can we use `slow_log` instead?

components/raftstore/src/store/fsm/peer.rs
NingLin-P, line 1756: Do you mean:
```rust
fail_point!(
    "memtrace_raft_messages_overflow_check_peer_recv",
    MEMTRACE_RAFT_MESSAGES.sum() < heap_size,
    |_| {}
);

// in test
fail::cfg("memtrace_raft_messages_overflow_check_peer_recv", "panic").unwrap();
```
?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -242,10 +242,17 @@ where
         }
         let peer_msg = PeerMsg::RaftMessage(InspectedRaftMessage { heap_size, msg });
         let event = TraceEvent::Add(heap_size);
+        let send_failed = Cell::new(true);
+
+        MEMTRACE_RAFT_MESSAGES.trace(event);
+        defer!(if send_failed.get() {
+            MEMTRACE_RAFT_MESSAGES.trace(TraceEvent::Sub(heap_size));
+        });
```
NingLin-P, line 250: What is the difference with the original code?

components/raftstore/src/store/worker/read.rs
tier-cap: This logic is not included, but only fix make clippy check.

tier-cap: This logic is not changed, but only fix for make clippy pass.

components/raftstore/src/store/worker/read.rs
tier-cap: This logic is not changed, but only fix for make clippy pass.

components/raftstore/src/store/fsm/peer.rs
tier-cap: Correct

components/raftstore/src/store/fsm/store.rs
tier-cap: Original code logic is:
    op1: send msg; op2: add size.
If the msg handle occurs between op1 and op2, then the sub size will happen before add size, which leads to overflow.

components/raftstore/src/store/fsm/peer.rs
tier-cap: We need the send msg failpoint to trigger this assert.

components/raftstore/src/store/worker/read.rs
```diff
@@ -520,16 +524,26 @@ where
...
+                if self.get_must_track_version_changed_failpoint() {
+                    if let Some(d) = self.delegates.get(&region_id) {
+                        info!("must_track_version_changed: {}", d.track_ver.any_new());
+                    }
+                }
                 debug!("update local read delegate"; "region_id" => region_id);
                 self.metrics.rejected_by_cache_miss += 1;
 
+                let t = Instant::now();
+
```
tier-cap, line 536: I haven't changed these lines, but why these diffs occur?

components/raftstore/src/store/fsm/peer.rs
NingLin-P: Why not adopt the above suggestion, I think it is cleaner

tests/failpoints/cases/test_metrics_overflow.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+use test_raftstore::*;
+
+#[test]
+fn test_memory_metrics_overflow() {
```
NingLin-P, line 6: The test can pass without the fix.

components/raftstore/src/store/fsm/store.rs
NingLin-P: Got it

components/raftstore/src/store/fsm/peer.rs
tier-cap: Done!

tests/failpoints/cases/test_metrics_overflow.rs
tier-cap: Cannot, keep the store loop recv and handle msgs before the metrics add, then the test must fail.

NingLin-P (approved):

5kbpers (approved):

Issue #12092, opened 2022-03-04
v5.3.1 version, run tpch workload, print lots of [advance.rs:273] ["check leader failed"] 


seiya-annie:
### What version of TiKV are you using?
v5.3.1

### What operating system and CPU are you using?
linux

### Steps to reproduce
1. install v5.3.1
2. /data-sdb/tools/go-tpc/bin/go-tpc tpch -H xxx.16.5.xxx -P 4000 -D tpch --sf 1 --analyze --tiflash prepare 

### What did you expect?
run normally

### What did happened?
[2022/03/04 12:35:43.011 +08:00] [ERROR] [advance.rs:273] ["check leader failed"] [to_store=56] [store_id=1] [error="RpcFailure(RpcStatus { code: 12-UNIMPLEMENTED, message: \"\", details: [] })"]
[2022/03/04 12:35:44.011 +08:00] [INFO] [<unknown>] ["New connected subchannel at 0x7f7efa22e690 for subchannel 0x7f7e88e1d1c0"]
[2022/03/04 12:35:44.011 +08:00] [ERROR] [advance.rs:273] ["check leader failed"] [to_store=56] [store_id=1] [error="RpcFailure(RpcStatus { code: 12-UNIMPLEMENTED, message: \"\", details: [] })"]
[2022/03/04 12:35:45.012 +08:00] [INFO] [<unknown>] ["New connected subchannel at 0x7f7e4a64c0f0 for subchannel 0x7f7e88e1d1c0"]
[2022/03/04 12:35:45.012 +08:00] [ERROR] [advance.rs:273] ["check leader failed"] [to_store=56] [store_id=1] [error="RpcFailure(RpcStatus { code: 12-UNIMPLEMENTED, message: \"\", details: [] })"]
[2022/03/04 12:35:46.013 +08:00] [INFO] [<unknown>] ["New connected subchannel at 0x7f7efbb97200 for subchannel 0x7f7e88e1d1c0"]
[2022/03/04 12:35:46.013 +08:00] [ERROR] [advance.rs:273] ["check leader failed"] [to_store=56] [store_id=1] [error="RpcFailure(RpcStatus { code: 12-UNIMPLEMENTED, message: \"\", details: [] })"]
[2022/03/04 12:35:47.015 +08:00] [INFO] [<unknown>] ["New connected subchannel at 0x7f7efb329ed0 for subchannel 0x7f7e88e1d1c0"]
[...]

PR #12190, opened 2022-03-17
resolved_ts: fix oom and add more metrics

### LINKED ISSUE #12159: TiKV OOM under TPCC workload when two tikv down for 10 minutes
## Bug Report

TiKV OOM.

```
$ /tikv-server -V 
TiKV 
Release Version:  6.0.0-alpha 
Edition:          Community 
Git Commit Hash:  8954a76e2b87575d80336f502a4d078e5da1508f 
Git Commit Branch: heads/refs/tags/v6.0.0-nightly 
UTC Build Time:   2022-03-09 18:10:26 
Rust Version:     rustc 1.60.0-nightly (1e12aef3f 2022-02-13) 
Enable Features:  jemalloc mem-profiling portable sse test-engines-rocksdb cloud-aws cloud-gcp cloud-azure 
Profile:          dist_release
```

### Steps to reproduce

0. create a cluster with 5 replicas
    ```
    [replication]
    max-replicas = 5
    ```
2. run tpcc workload for 10 minutes (go-tpc)
3. down 2 tikvs for about 10 minutes

After the 2 tikvs were down, the other 3 tikvs's memory kept increasing.

### What did you expect?
No OOM.

### What did happened?
TiKV OOM.
image
image

For more details, see https://pingcap.feishu.cn/docs/doccnTugaMBp2jB62ZDwWX0Vv2c# .

5kbpers:

Issue Number: Close #12159

What's Changed:

```commit-message
* Do not hold mutex during calling `get_store_async`
* Add more metrics
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/resolved_ts/src/advance.rs
+++ b/components/resolved_ts/src/advance.rs
@@ -374,7 +374,6 @@ async fn get_tikv_client(
     let store = box_try!(pd_client.get_store_async(store_id).await);
     let mut clients = tikv_clients.lock().await;
     let start = Instant::now_coarse();
-    // let store = box_try!(pd_client.get_store_async(store_id).await);
     // hack: so it's different args, grpc will always create a new connection.
     let cb = ChannelBuilder::new(env.clone()).raw_cfg_int(
         CString::new("random id").unwrap(),
--- a/components/resolved_ts/src/advance.rs
+++ b/components/resolved_ts/src/advance.rs
@@ -371,7 +371,10 @@ async fn get_tikv_client(
             return Ok(client);
         }
     }
-    let store = box_try!(pd_client.get_store_async(store_id).await);
+    let timeout = Duration::from_millis(DEFAULT_CHECK_LEADER_TIMEOUT_MILLISECONDS);
+    let store = box_try!(box_try!(
+        tokio::time::timeout(timeout, pd_client.get_store_async(store_id)).await
+    ));
     let mut clients = tikv_clients.lock().await;
     let start = Instant::now_coarse();
     // hack: so it's different args, grpc will always create a new connection.
--- a/components/resolved_ts/src/advance.rs
+++ b/components/resolved_ts/src/advance.rs
@@ -365,23 +365,24 @@ async fn get_tikv_client(
     env: Arc<Environment>,
     tikv_clients: Arc<Mutex<HashMap<u64, TikvClient>>>,
 ) -> Result<TikvClient> {
-    let mut clients = tikv_clients.lock().await;
-    let client = match clients.get(&store_id) {
-        Some(client) => client.clone(),
-        None => {
-            let start = Instant::now_coarse();
-            let store = box_try!(pd_client.get_store_async(store_id).await);
-            // hack: so it's different args, grpc will always create a new connection.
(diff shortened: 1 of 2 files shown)


components/resolved_ts/src/advance.rs
```diff
@@ -359,23 +365,24 @@ async fn get_tikv_client(
...
-            client
+    {
+        let clients = tikv_clients.lock().await;
+        if let Some(client) = clients.get(&store_id).cloned() {
+            return Ok(client);
         }
-    };
-    Ok(client)
+    }
+    let store = box_try!(pd_client.get_store_async(store_id).await);
```
overvenus, line 374: Could you add a unit test?

components/resolved_ts/src/advance.rs
```diff
@@ -359,23 +365,24 @@ async fn get_tikv_client(
...
+    let mut clients = tikv_clients.lock().await;
+    let start = Instant::now_coarse();
+    // let store = box_try!(pd_client.get_store_async(store_id).await);
```
hicqu, line 377: The line can be removed.

components/resolved_ts/src/advance.rs
5kbpers: We don't have a good way to inject a failure for `get_store_async` since `failpoint` doesn't support async/await

hicqu (approved):

components/resolved_ts/src/advance.rs
NingLin-P, line 374: Should we also add a timeout for `get_store_async`?

components/resolved_ts/src/advance.rs
```diff
@@ -359,23 +365,24 @@ async fn get_tikv_client(
...
+    {
+        let clients = tikv_clients.lock().await;
+        if let Some(client) = clients.get(&store_id).cloned() {
+            return Ok(client);
         }
-    };
-    Ok(client)
+    }
+    let store = box_try!(pd_client.get_store_async(store_id).await);
+    let mut clients = tikv_clients.lock().await;
```
NingLin-P, line 375: How about acquiring the lock after the client is created to shorten the time of holding the lock.

overvenus (approved):

components/resolved_ts/src/advance.rs
5kbpers: `TikvClient` has `!Send` so it cannot cross await point.

NingLin-P (approved):

tabokie:
This PR needs to be reverted, PD can't handle these amount of concurrent `get_store`. Before and after:

image

hehechen:
> 



> This PR needs to be reverted, PD can't handle these amount of concurrent `get_store`. Before and after:
> 
> image
> 
> Edit: Normally this would work fine, but for cluster with large number of tiflash nodes, the `check_leader_async` will return error and trigger another round of requesting. So instead of reverting this one, we should filter "unimplemented" error from `check_leader_async` RPC and avoid reconnection. It will also close #12092.

In fact, TiFlash has implemented check_leader RPC, the "unimplemented" error just because the [address is wrong.](https://github.com/tikv/tikv/issues/13310)

Issue #9765, opened 2021-03-08
Built-in profiling can cause a segmentation fault


sticnarf:
### What version of TiKV are you using?

Latest nightly: dc8ce2cf6a8904cb3dad556f71b11bac3531689b

### What operating system and CPU are you using?

Linux. It happens on both CentOS 7 (3.10.0-957.27.2.el7.x86_64) and Arch Linux (5.11.2-arch1-1).

### Steps to reproduce

Run YCSB workloada on a 3-node TiKV cluster. Meanwhile, get `http://{TiKV_IP}:20180/debug/pprof/profile?seconds=30`.

### What did you expect?

Get a CPU profiling flamegraph.

### What did happened?

A segmentation fault happens. Backtrace of the core:

```
Program terminated with signal SIGSEGV, Segmentation fault.
[Current thread is 1 (Thread 0x7ff7f83fe640 (LWP 162560))]
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /data/deploy/tikv-20160/bin/tikv-server.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) bt
#0  x86_64_fallback_frame_state (context=0x7ff7f83d6a50, context=0x7ff7f83d6a50, fs=0x7ff7f83d6b40) at ./md-unwind-support.h:63
#1  uw_frame_state_for (context=0x7ff7f83d6a50, fs=0x7ff7f83d6b40) at /build/gcc/src/gcc/libgcc/unwind-dw2.c:1271
#2  0x00007ff83337433b in _Unwind_Backtrace (trace=0x556d67bdcd00 <backtrace::backtrace::libunwind::trace::trace_fn>, trace_argument=0x7ff7f83d6f40) at /build/gcc/src/gcc/libgcc/unwind.inc:302
#3  0x0000556d688c5d97 in perf_signal_handler ()
#4  <signal handler called>
#5  0x0000556d6832d941 in crossbeam_deque::Stealer<T>::steal_batch_and_pop ()
#6  0x0000000000010830 in ?? ()
#7  0x0000000000000000 in ?? ()
```

Frame 5 has many possibilities, it may not be `crossbeam_deque::Stealer<T>::steal_batch_and_pop`.

sticnarf:
Probably due to incorrect DWARF info during stack probing.

pprof-rs issue: https://github.com/tikv/pprof-rs/issues/57
rust issue: https://github.com/rust-lang/rust/issues/83139

[...]

PR #12201, opened 2022-03-18
add vdso to the pprof blocklist

YangKeao:

Add `[vdso]` to the `pprof-rs` blocklist, we "guess" the problem would be the dwarf information in vdso is incorrect. This is a PR to test whether it would work well after blocking the `vdso`.

Issue Number: Close https://github.com/tikv/tikv/issues/9765

### Release note 

```release-note
Fix crash when profiling in Ubuntu 18.04.
```

### DIFF
--- a/src/server/status_server/profile.rs
+++ b/src/server/status_server/profile.rs
@@ -202,7 +202,7 @@ where
     let on_start = || {
         let guard = pprof::ProfilerGuardBuilder::default()
             .frequency(frequency)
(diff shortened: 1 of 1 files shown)


breeswish (approved):
I did not experience crash any more in 12 hours, while without this fix there are 10 crashes in 3 hours. The fix is very effective.

sticnarf (approved):
LGTM

Issue #12061, opened 2022-03-02
use clippy to disable `hex::encode_upper`


skyzh:
## Development Task

https://github.com/tikv/tikv/blob/master/scripts/check-redact-log

Previously, we are using grep to find `encode_hex` in our codebase, so as not to print hex-encoded keys into the log. This script has many limitations. For example, if users export the `encode_hex` methods somewhere, the `check-redact-log` script won't be able to spot such changes.

Indeed, we can use clippy to disallow some method use. For example,

```
disallowed-methods = [
    { path = "hex::encode_hex", reason = "Keys should not be printed to log. Consider using log_wrapper for logging." }
]
```

JayiceZ:
I wrote a demo impl in the following commit: https://github.com/JayiceZ/tikv/commit/ddddd68932b389c62ad74a0c29b5b440dd5792f8

But it's confusing that it doesn't work correctly somewhere.😭

For example: `src/read_pool.rs`(only in the commit,for testing) and `cmd/tikv-ctl` use `hex::encode_upper `but they didn't trigger the lint... :(

skyzh:
Thanks a lot! Would you please send a PR and cc me for review?

skyzh:
Regarding your question, which command and what arts are you using for clippy?

JayiceZ:
> Regarding your question, which command and what args are you using for clippy?

`make clippy`

skyzh:
https://github.com/tikv/tikv/blob/master/scripts/clippy

Maybe that's because the lint rule for disallowed methods are disabled here? Not sure about this...

JayiceZ:
> 

But it works somewhere. `log_wrapper` can trigger the lint...

skyzh:
Hmm... Then I suspect there's bug with this lint rule. Don't worry, you may send a PR and I can check it myself in my environment.

JayiceZ:
> 

Alright~

PR #12206, opened 2022-03-19
security:use clippy to disallow hex_encode

JayiceZ:
close https://github.com/tikv/tikv/issues/12061
Signed-off-by: Jayice <1185430411@qq.com>

Issue Number: Close https://github.com/tikv/tikv/issues/12061

What's Changed:

```commit-message
use clippy to disallow `hex:encode_upper`, replacing grep
```

### Check List 

Tests 

### Release note 

```release-note
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,7 +285,6 @@ pre-clippy: unset-override
 	@rustup component add clippy
 
 clippy: pre-clippy
-	@./scripts/check-redact-log
 	@./scripts/check-docker-build
 	@./scripts/clippy-all
 
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,3 @@
+disallowed-methods = [
+    { path = "hex::encode_upper", reason = "Keys should not be printed to log. Consider using log_wrapper for logging." },
+]


JayiceZ:
It needs more discussion. see https://github.com/tikv/tikv/issues/12061

JayiceZ:
/cc @skyzh  PTAL
I didn't add `#![allow(clippy::disallowed_methods)]` in both `cmd/tikv-ctl `and `log_wrapper`, But only `log_wrapper` will trigger the lint by calling `make clippy`

skyzh:
> For example: src/read_pool.rs(only in the commit,for testing) and cmd/tikv-ctl use hex::encode_upper but they didn't trigger the lint... And at the same time, it can work in some other where. :(

Another reason might be that `clippy` may stop further check once it emits warning from the log_wrappers crate. I've added a test `hex::encode_upper` in `cmd` and see if clippy will give warnings...

JayiceZ:
> > For example: src/read_pool.rs(only in the commit,for testing) and cmd/tikv-ctl use hex::encode_upper but they didn't trigger the lint... And at the same time, it can work in some other where. :(
> 
> Another reason might be that `clippy` may stop further check once it emits warning from the log_wrappers crate. I've added a test `hex::encode_upper` in `cmd` and see if clippy will give warnings...

It didn't work in my env. Btw, it seems like you forget add `hex` dep in `tikv-server/Cargo.toml`

skyzh:
`hex` is in tikv-ctl's dependency list.

https://github.com/tikv/tikv/blob/e687a823f2de794c471ecac968531a1b150cfa53/cmd/tikv-ctl/src/main.rs#L51-L52

I added hex::encode_upper, and it passes clippy... Seems weird.

JayiceZ:
> `hex` is in tikv-ctl's dependency list.
> 
> https://github.com/tikv/tikv/blob/e687a823f2de794c471ecac968531a1b150cfa53/cmd/tikv-ctl/src/main.rs#L51-L52
> 
> I added hex::encode_upper, and it passes clippy... Seems weird.

I suspect it's the lint rule's bug😢

skyzh:
Well... I'll see if I could reproduce this in a minimal environment, and file an issue to cargo-chippy.

PR #12219, opened 2022-03-21
Quota limiter: make method to delay cpu more precise

### LINKED ISSUE #12218: Quota limiter: correct the mehod to limit cpu

https://github.com/tikv/tikv/commit/9a43faf4da20389a4e2c262ee8ab8b369a3bfec4. After https://github.com/tikv/tikv/pull/12151

don't matter

### Steps to reproduce

don't matter

### What did you expect?
1. The `foreground-cpu-time` configured by the user will be reduced by a fixed factor, which is not necessary.
2. The result returned by cpu_limiter does not need to be processed additionally.

### What did happened?
original implement don't affect the correctness of the results.

hunterlxt:

Issue Number: Close #12218

What's Changed:

```commit-message
Use the results returned by the cpu limiter directly without extra processing
```

### Related changes

- Need to cherry-pick to the release 6.0 branch

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- No

### Release note 

```release-note
None
```

### DIFF
--- a/components/tikv_util/src/quota_limiter.rs
+++ b/components/tikv_util/src/quota_limiter.rs
@@ -9,13 +9,6 @@ use super::timer::GLOBAL_TIMER_HANDLE;
 use cpu_time::ThreadTime;
 use futures::compat::Future01CompatExt;
 
-// The cpu time is not a real statistics, only part of the processing logic is
-// taken into account, so it needs to be multiplied by a factor.
-// transfer milli cpu to micro cpu
-//
-// TODO: Don't adjusted based on experience.
-const CPU_TIME_FACTOR: f64 = 0.8;
-
 // To avoid long tail latency.
 const MAX_QUOTA_DELAY: Duration = Duration::from_secs(1);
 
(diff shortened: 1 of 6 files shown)


hunterlxt:
In recent internal bench tests, I found that refill duration can affect performance, and less refill duration can get less delay duration when the bucket value is used off, which make grpc p99 lower and qps higher. But less refill duration means that more requests will be delayed (Each delay is shortened), so context switch may increase. 

Here are some results:
pqHYfUHmsl
0XQfCsp8yN
6hPqAgG2Kp
N2MgarRbwK

hunterlxt:
In the lasted commit, I added a new config to allow user to set max delay for each single request, which is a result of discussion with @tonyxuqqi

components/tikv_util/src/quota_limiter.rs
```diff
@@ -95,11 +89,14 @@ impl QuotaLimiter {
         cpu_quota: usize,
         write_bandwidth: ReadableSize,
         read_bandwidth: ReadableSize,
+        max_delay_duration: ReadableDuration,
     ) -> Self {
         let cputime_limiter = if cpu_quota == 0 {
             Limiter::new(f64::INFINITY)
         } else {
-            Limiter::new(cpu_quota as f64 * CPU_TIME_FACTOR * 1000_f64)
+            <Limiter>::builder(cpu_quota as f64 * 1000_f64)
```
sticnarf, line 97: The `<>` looks unnecessary?

sticnarf (approved):

Connor1996 (approved):
LGTM

PR #12223, opened 2022-03-21
copr: limit the number of tasks blocked by the semaphore

sticnarf:
This is only a hotfix PR. 

It adds a config setting the max number of tasks blocked by the semaphore:

```toml
[server]
end-point-semaphore-waiter-limit = 1024
```

Tasks exceeding the limit will be aborted using a `MaxPendingTasksExceeded` error.

Slow log will be printed for all aborted coprocessor requests due to exceeding the deadline or the task number limit.

### DIFF
--- a/src/coprocessor/endpoint.rs
+++ b/src/coprocessor/endpoint.rs
@@ -67,6 +67,8 @@ pub struct Endpoint<E: Engine> {
     /// The soft time limit of handling Coprocessor requests.
     max_handle_duration: Duration,
 
+    semaphore_waiter_limit: usize,
+
     slow_log_threshold: Duration,
 
     _phantom: PhantomData<E>,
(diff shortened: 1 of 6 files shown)


src/coprocessor/endpoint.rs
```diff
@@ -392,12 +396,27 @@ impl<E: Engine> Endpoint<E> {
...
-            handle_request_future.await
+            handle_request_future
+                .await
+                .map_err(Error::from)
+                .and_then(|res| res)
         };
-        let result = deadline_res.map_err(Error::from).and_then(|res| res);
+        if matches!(
+            result,
+            Err(Error::DeadlineExceeded | Error::MaxPendingTasksExceeded)
```
BusyJay, line 416: I think only `Error::DeadlineExceeded` needs to be printed.

sticnarf: Do you think we just revert back the slow log behavior, print the slow log when `req_lifetime` > threshold?

BusyJay: If the task takes too long and be aborted, will it still show in the slow log?

sticnarf: If we use `req_lifetime` to check, it will.

PR #11896, opened 2022-01-21
cdc: capture old value from txn layer dynamically

hicqu:
### What is changed and how it works?

Close #10091

* remove stuff about the `enable_old_value` option, because it will be always `true` now;
* refactor `Delegate` to make it easiler to maintain;
* avoid to capture old vaue from txn layer if all downstreams are stopped.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
None
```

PR #12262, opened 2022-03-25
cdc: remove downstreams correctly when uninitialized

hicqu:

https://github.com/tikv/tikv/pull/11896#discussion_r834951841 introduces a critical bug for CDC, which can cause data lost or downstreams blocking forever. Here is how to trigger the bug:
1. multiple downstreams subscribe one region
2. cancel one downstream before the region delegate has been initialized

Then, `Delegate::stop_observing` will be called incorrectly and all change events will be lost.

This PR does
1. correct the bug
2. add a log for `stop_observing`
3. complete unit tests for `Delegate::subscribe` and `Delegate::unsubscribe`.

```commit-message
cdc: remove downstreams correctly when uninitialized
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/cdc/src/delegate.rs
+++ b/components/cdc/src/delegate.rs
@@ -762,6 +762,8 @@ impl Delegate {
         if let Some(index) = downstreams.iter().position(|x| x.id == id) {
             let downstream = downstreams.swap_remove(index);
             if downstreams.is_empty() {
+                // Stop observing when the last downstream is removed. Otherwise the observer
+                // will keep pushing events to the delegate.
                 self.stop_observing();
             }
             return Some(downstream);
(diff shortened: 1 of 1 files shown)


components/cdc/src/delegate.rs
```diff
@@ -760,7 +761,7 @@ impl Delegate {
         let downstreams = self.downstreams_mut();
         if let Some(index) = downstreams.iter().position(|x| x.id == id) {
             let downstream = downstreams.swap_remove(index);
-            if self.downstreams.is_empty() {
+            if downstreams.is_empty() {
                 self.stop_observing();
```
overvenus, line 765: Could you add some comments about when and why we need to `stop_observing`?

overvenus (approved):
Rest LGTM

hi-rustin (approved):

PR #12272, opened 2022-03-25
encryption: fix issue with opening plaintext files

### LINKED ISSUE #12162: Failed to create raft engine when encryption enabled.

```
Release Version:   6.0.0-alpha
Edition:           Community
Git Commit Hash:   1047b51285456e702b66253da26c7f14d582ff0c
Git Commit Branch: master
UTC Build Time:    2022-03-14 08:48:38
Enable Features:   jemalloc mem-profiling portable sse test-engines-rocksdb cloud-aws cloud-gcp cloud-azure
```

### Steps to reproduce

add some config in tikv.toml and then starts tikv-server
```
[security.encryption]
data-encryption-method = "aes256-ctr"

[security.encryption.master-key]
type = "file"
path = "/tmp/backup_restore_test/master-key-file"
```
### What did you expect?
tikv serves well.

### What did happened?
```
[2022/03/15 12:37:28.488 +08:00] [FATAL] [[server.rs:1372](http://server.rs:1372/)] ["Failed to create raft engine: \"[components/raft_log_engine/src/[engine.rs:440](http://engine.rs:440/)]: IO Error: Other error [components/encryption/src/[crypter.rs:83](http://crypter.rs:83/)]: Nonce + Counter must be 12/16 bytes, 0\""]
```

tabokie:

Issue Number: Close #12162

What's Changed:

When key manager doesn't have the key for certain data files, we should still be able to open it.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/encryption/src/manager/mod.rs
```diff
@@ -591,7 +593,12 @@ impl DataKeyManager {
             writer,
             crypter::encryption_method_from_db_encryption_method(file.method),
             &file.key,
-            Iv::from_slice(&file.iv)?,
+            if file.iv.is_empty() {
```
glorv, line 596: I'm concerned about adding this logic will hide bugs that forget to register file name and thus the target file will be not encrypted when encryption is enabled. Are there any case we actually need this fallback logic?

tabokie: That's a fair point. But right now RocksDB also allows opening plaintext file:
https://github.com/tikv/rocksdb/blob/21712fe4243a57a5e7854574b933e67d0fd227f6/encryption/encryption.cc#L262. The two should be consistent.

With this patch, we should be able to enable encryption on an un-encrypted dataset. I honestly don't like using encryption this way. But QA thinks it's useful for their testing.

tabokie:
Not needed now.

tabokie:
Ping @glorv @Connor1996

glorv (approved):

Connor1996 (approved):
LGTM

PR #11921, opened 2022-01-27
raftstore: Report is apply busy to PD

Connor1996:
### What is changed and how it works?

Part of #11409  

What's Changed:
- Report is apply busy(apply wait duration > 100ms) to PD

### Check List 

- Manual test (add detailed scripts or steps below)

### Release note 

```release-note
None
```

PR #12308, opened 2022-03-30
raftstore: Make unsafe recovery wait apply cover snapshot apply cases ref #10483

v01dstar:

Issue Number: ref #10483

What's Changed:

Raft snapshot apply has a different code path to update the apply index, compare to other types of Raft entry which update apply index in `raftstore/store/fsm/peer.rs:on_apply_res()`. This path was not covered in the initial implementation of "unsafe recovery wait apply", and this PR fixes it.

### Related changes

PR that adds the initial implementation of unsafe recovery wait apply: https://github.com/tikv/tikv/pull/11716

### Check List 

Tests 

- Integration test

Side effects

N/A

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -6,7 +6,7 @@ use std::cell::Cell;
 use std::collections::Bound::{Excluded, Unbounded};
 use std::collections::VecDeque;
 use std::iter::Iterator;
-use std::sync::{atomic::AtomicUsize, atomic::Ordering, Arc};
+use std::sync::{atomic::AtomicUsize, Arc};
 use std::time::Instant;
 use std::{cmp, mem, u64};
 
@@ -23,7 +23,7 @@ use kvproto::errorpb;
 use kvproto::import_sstpb::SwitchMode;
 use kvproto::kvrpcpb::DiskFullOpt;
 use kvproto::metapb::{self, Region, RegionEpoch};
-use kvproto::pdpb::{CheckPolicy, StoreStats};
+use kvproto::pdpb::CheckPolicy;
 use kvproto::raft_cmdpb::{
     AdminCmdType, AdminRequest, CmdType, PutRequest, RaftCmdRequest, RaftCmdResponse, Request,
     StatusCmdType, StatusResponse,
@@ -58,7 +58,7 @@ use crate::store::cmd_resp::{bind_term, new_error};
 use crate::store::fsm::store::{PollContext, StoreMeta};
 use crate::store::fsm::{
     apply, ApplyMetrics, ApplyTask, ApplyTaskRes, CatchUpLogs, ChangeObserver, ChangePeer,
-    ExecResult, StoreInfo,
+    ExecResult,
 };
 use crate::store::hibernate_state::{GroupState, HibernateState};
 use crate::store::local_metrics::RaftMetrics;
@@ -151,12 +151,6 @@ where
     /// Before actually destroying a peer, ensure all log gc tasks are finished, so we
     /// can start destroying without seeking.
     logs_gc_flushed: bool,
-
-    /// To make sure the reported store/peer meta is up to date, each peer has to wait for the log
-    /// at its target commit index to be applied. The last peer does so triggers the next procedure
-    /// which is reporting the store/peer meta to PD.
-    unsafe_recovery_target_commit_index: Option<u64>,
-    unsafe_recovery_wait_apply_counter: Option<Arc<AtomicUsize>>,
 }
 
 pub struct BatchRaftCmdRequestBuilder<E>
@@ -275,8 +269,6 @@ where
                 trace: PeerMemoryTrace::default(),
                 delayed_destroy: None,
                 logs_gc_flushed: false,
-                unsafe_recovery_target_commit_index: None,
-                unsafe_recovery_wait_apply_counter: None,
             }),
         ))
     }
@@ -331,8 +323,6 @@ where
                 trace: PeerMemoryTrace::default(),
                 delayed_destroy: None,
                 logs_gc_flushed: false,
-                unsafe_recovery_target_commit_index: None,
-                unsafe_recovery_wait_apply_counter: None,
             }),
         ))
     }
@@ -882,53 +872,10 @@ where
         self.register_raft_base_tick();
     }
 
-    fn finish_unsafe_recovery_wait_apply(&mut self) {
-        if self
-            .fsm
-            .unsafe_recovery_wait_apply_counter
-            .as_ref()
-            .unwrap()
-            .fetch_sub(1, Ordering::Relaxed)
-            == 1
-        {
-            let mut stats = StoreStats::default();
-            stats.set_store_id(self.store_id());
-            let store_info = StoreInfo {
-                kv_engine: self.ctx.engines.kv.clone(),
-                raft_engine: self.ctx.engines.raft.clone(),
-                capacity: self.ctx.cfg.capacity.0,
-            };
-            let task = PdTask::StoreHeartbeat {
-                stats,
-                store_info,
-                send_detailed_report: true,
-                dr_autosync_status: self
-                    .ctx
-                    .global_replication_state
-                    .lock()
-                    .unwrap()
-                    .store_dr_autosync_status(),
-            };
-            if let Err(e) = self.ctx.pd_scheduler.schedule(task) {
-                panic!("fail to send detailed report to pd {:?}", e);
-            }
-        }
-        self.fsm.unsafe_recovery_target_commit_index = None;
-        self.fsm.unsafe_recovery_wait_apply_counter = None;
-    }
-
     fn on_unsafe_recovery_wait_apply(&mut self, counter: Arc<AtomicUsize>) {
-        self.fsm.unsafe_recovery_target_commit_index =
-            Some(self.fsm.peer.raft_group.store().commit_index());
-        self.fsm.unsafe_recovery_wait_apply_counter = Some(counter);
-        // If the applied index equals to the commit index, there is nothing to wait for, proceeds
-        // to the next step immediately. If they are not equal, further checks will be performed in
-        // on_apply_res().
-        if self.fsm.stopped
-            || self.fsm.peer.raft_group.store().applied_index()
-                == self.fsm.peer.raft_group.store().commit_index()
-        {
-            self.finish_unsafe_recovery_wait_apply();
+        self.fsm.peer.unsafe_recovery_wait_apply(self.ctx, counter);
+        if self.fsm.stopped {
+            self.fsm.peer.unsafe_recovery_finish_wait_apply(self.ctx);
         }
     }
 
@@ -1952,9 +1899,13 @@ where
             }
         }
         // After a log has been applied, check if we need to trigger the unsafe recovery reporting procedure.
-        if let Some(target_commit_index) = self.fsm.unsafe_recovery_target_commit_index {
-            if self.fsm.peer.raft_group.store().applied_index() >= target_commit_index {
-                self.finish_unsafe_recovery_wait_apply();
+        if let Some(unsafe_recovery_state) = self.fsm.peer.unsafe_recovery_state.as_ref() {
+            if let Some(wait_apply_target_index) =
+                unsafe_recovery_state.wait_apply_target_index.as_ref()
+            {
+                if self.fsm.peer.raft_group.store().applied_index() >= *wait_apply_target_index {
+                    self.fsm.peer.unsafe_recovery_finish_wait_apply(self.ctx);
+                }
             }
         }
     }
@@ -3055,8 +3006,10 @@ where
         assert!(!self.fsm.peer.is_handling_snapshot());
 
         // No need to wait for the apply anymore.
-        if self.fsm.unsafe_recovery_target_commit_index.is_some() {
-            self.finish_unsafe_recovery_wait_apply();
+        if let Some(unsafe_recovery_state) = self.fsm.peer.unsafe_recovery_state.as_ref() {
+            if unsafe_recovery_state.wait_apply_target_index.is_some() {
+                self.fsm.peer.unsafe_recovery_finish_wait_apply(self.ctx);
+            }
         }
 
         let mut meta = self.ctx.store_meta.lock().unwrap();
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -875,7 +875,9 @@ where
     fn on_unsafe_recovery_wait_apply(&mut self, counter: Arc<AtomicUsize>) {
         self.fsm.peer.unsafe_recovery_wait_apply(self.ctx, counter);
         if self.fsm.stopped {
-            self.fsm.peer.unsafe_recovery_finish_wait_apply(self.ctx);
+            self.fsm
+                .peer
+                .unsafe_recovery_maybe_finish_wait_apply(self.ctx, /*force=*/ true);
         }
     }
 
@@ -1899,14 +1901,10 @@ where
             }
         }
         // After a log has been applied, check if we need to trigger the unsafe recovery reporting procedure.
-        if let Some(unsafe_recovery_state) = self.fsm.peer.unsafe_recovery_state.as_ref() {
-            if let Some(wait_apply_target_index) =
-                unsafe_recovery_state.wait_apply_target_index.as_ref()
-            {
-                if self.fsm.peer.raft_group.store().applied_index() >= *wait_apply_target_index {
-                    self.fsm.peer.unsafe_recovery_finish_wait_apply(self.ctx);
-                }
-            }
+        if self.fsm.peer.unsafe_recovery_state.is_some() {
+            self.fsm
+                .peer
+                .unsafe_recovery_maybe_finish_wait_apply(self.ctx, /*force=*/ false);
         }
     }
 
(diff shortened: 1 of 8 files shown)


components/raftstore/src/store/fsm/store.rs
```diff
@@ -623,6 +623,9 @@ impl<'a, EK: KvEngine + 'static, ER: RaftEngine + 'static, T: Transport>
                     inspector.record_store_wait(send_time.saturating_elapsed());
                     self.ctx.pending_latency_inspect.push(inspector);
                 }
+                StoreMsg::SendDetailedReportForUnsafeRecovery => {
```
Connor1996, line 626: no need this msg, you can make `store_heartbeat_pd` to be a static function and call it in peer_fsm directly, see https://github.com/tikv/tikv/pull/11921/files#diff-51096b677ec8dac18619cd087727985b354be3e529ffdcf5b32429a32b77caa3R1653

components/raftstore/src/store/peer.rs
```diff
@@ -2282,6 +2292,10 @@ where
             }
             // Pause `read_progress` to prevent serving stale read while applying snapshot
             self.read_progress.pause();
+
+            if self.unsafe_recovery_state.is_some() {
+                self.unsafe_recovery_maybe_finish_wait_apply(ctx, /*force=*/ false);
```
Connor1996, line 2297: I think this should be called after `advance_apply_to` in `check_snap_status`

Connor1996: here doesn't indicate the snapshot is applied. Seems the test isn't failed, please update the test to cover the case. Maybe use `cluster.get_engine()` to check the data.

components/raftstore/src/store/fsm/store.rs
v01dstar: Done.

components/raftstore/src/store/peer.rs
v01dstar: Updated. 
As for the test, I don't think we need to do any extra work, since, after the "check" logic is changed, `assert_eq!(pd_client.must_get_store_reported(&nodes[1]), 0);` already covers the case.

components/raftstore/src/store/peer.rs
```diff
@@ -454,6 +454,14 @@ pub struct ReadyResult {
     pub has_write_ready: bool,
 }
 
+pub struct UnsafeRecoveryState {
+    // During unsafe recovery, in order to make sure the report send to PD contains the up-to-date
+    // information, each peer has to make sure that certain entries are not only commited but also
+    // applied, and who ever is the last one needs to schedule the reporting.
+    pub wait_apply_target_index: Option<u64>,
```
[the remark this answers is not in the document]
Connor1996: you should make the state `Enum` instead of a lot of `Option`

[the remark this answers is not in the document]
v01dstar: Indeed. Done

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1412,6 +1407,7 @@ impl<EK: KvEngine, ER: RaftEngine> RaftBatchSystem<EK, ER> {
         let mut builder = RaftPollerBuilder {
             cfg,
             store: meta,
+            store_start_time: time::get_time(),
```
BusyJay, line 1410: I don't think it's the correct start time.

components/raftstore/src/store/peer.rs
```diff
@@ -4387,6 +4400,38 @@ where
...
+        force: bool,
+    ) {
+        if let Some(unsafe_recovery_state) = self.unsafe_recovery_state.as_ref() {
+            let UnsafeRecoveryState::WaitApply {
+                target_index,
+                task_counter,
+            } = unsafe_recovery_state;
+            if self.raft_group.raft.raft_log.applied >= *target_index || force {
+                if task_counter.fetch_sub(1, Ordering::Relaxed) == 1 {
+                    store_heartbeat_pd(ctx, /*send_detailed_report=*/ true);
```
BusyJay, line 4447: Why is a store heartbeat sent from a peer? What's the point of abstract then? 🤣

BusyJay: Better than sending to PD directly.

Connor1996: alright, let's revert the change 🤣

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -879,53 +869,12 @@ where
...
-        self.fsm.unsafe_recovery_wait_apply_counter = Some(counter);
-        // If the applied index equals to the commit index, there is nothing to wait for, proceeds
-        // to the next step immediately. If they are not equal, further checks will be performed in
-        // on_apply_res().
-        if self.fsm.stopped
-            || self.fsm.peer.raft_group.store().applied_index()
-                == self.fsm.peer.raft_group.store().commit_index()
-        {
-            self.finish_unsafe_recovery_wait_apply();
+        self.fsm.peer.unsafe_recovery_wait_apply(self.ctx, counter);
```
Connor1996, line 873: can we remove the `unsafe_recovery_wait_apply` and keep the logic here?

tests/failpoints/cases/test_unsafe_recovery.rs
```diff
@@ -92,3 +93,94 @@ fn test_unsafe_recover_send_report() {
...
+
+    // Blocks the raft snap apply process.
+    let apply_triggered_pair = Arc::new((Mutex::new(false), Condvar::new()));
+    let apply_triggered_pair2 = Arc::clone(&apply_triggered_pair);
+    let apply_released_pair = Arc::new((Mutex::new(false), Condvar::new()));
+    let apply_released_pair2 = Arc::clone(&apply_released_pair);
+    fail::cfg_callback("region_apply_snap", move || {
+        {
+            debug!("unsafe recovery test, snapshot apply triggered");
+            let (lock, cvar) = &*apply_triggered_pair2;
```
Connor1996, line 132: extract utill functions called `trigger_wait` and `wait_on`?

Connor1996 (commented):
rest LGTM

components/raftstore/src/store/fsm/store.rs
v01dstar: Reverted, not applicable anymore

components/raftstore/src/store/peer.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Do you mean keep the state variable inside `raftstore/src/store/peer.rs`, but have the logic inside `raftstore/src/store/fsm/peer.rs`? That will be odd, won't it? Isn't it better to have the variables and the functions manipulating those variables close to each other?

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: Done. PTAL

components/raftstore/src/store/fsm/peer.rs
Connor1996: I agree the boundary between `store/peer.rs` and `store/fsm/peer.rs` is not cleared. We visit states recorded in `store/peer.rs` directly in `store/fsm/peer.rs`, See other `on_xxxx` in the pattern `self.fsm.peer.xxx`. 

Given that, `on_unsafe_recovery_wait_apply` with a  `unsafe_recovery_wait_apply` is verbose.

components/raftstore/src/store/peer.rs
```diff
@@ -4387,6 +4401,58 @@ where
...
     }
+
+    pub fn unsafe_recovery_wait_apply<T: Transport>(
+        &mut self,
+        ctx: &PollContext<EK, ER, T>,
+        counter: Arc<AtomicUsize>,
+    ) {
+        info!(
+            "unsafe recovery wait apply";
+            "target_index" => self.raft_group.store().commit_index(),
```
BusyJay, line 4412: Why use store's version instead of raft state machine's?

BusyJay, line 4412: Why use store's version instead of raft state machine's?

components/raftstore/src/store/peer.rs
```diff
@@ -4387,6 +4401,58 @@ where
...
+        });
+        self.unsafe_recovery_maybe_finish_wait_apply(ctx, /*force=*/ false);
+    }
+
+    pub fn unsafe_recovery_maybe_finish_wait_apply<T: Transport>(
+        &mut self,
+        ctx: &PollContext<EK, ER, T>,
+        force: bool,
+    ) {
+        if let Some(unsafe_recovery_state) = self.unsafe_recovery_state.as_ref() {
```
BusyJay, line 4428:
```suggestion
        if let Some(unsafe_recovery_state) = &self.unsafe_recovery_state {
```

BusyJay, line 4428:
```suggestion
        if let Some(unsafe_recovery_state) = self.unsafe_recovery_state.as_ref() {
```
```suggestion
        if let Some(unsafe_recovery_state) = &self.unsafe_recovery_state {
```

components/raftstore/src/store/peer.rs
```diff
@@ -4387,6 +4401,58 @@ where
...
+            } = unsafe_recovery_state;
+            if self.raft_group.raft.raft_log.applied >= *target_index || force {
+                info!(
+                    "unsafe recovery finish wait apply";
+                    "target_index" => target_index,
+                    "applied" =>  self.raft_group.raft.raft_log.applied,
+                    "force" => force,
+                    "counter" =>  task_counter.load(Ordering::Relaxed)
+                );
+                if task_counter.fetch_sub(1, Ordering::Relaxed) == 1 {
```
BusyJay, line 4446: Should be `SeqCst`.

BusyJay, line 4446: Should be `SeqCst`.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2751,8 +2701,10 @@ where
         assert!(!self.fsm.peer.is_handling_snapshot());
 
         // No need to wait for the apply anymore.
-        if self.fsm.unsafe_recovery_target_commit_index.is_some() {
-            self.finish_unsafe_recovery_wait_apply();
+        if self.fsm.peer.unsafe_recovery_state.is_some() {
```
BusyJay, line 2704: What if the message is still in queue when the peer is about to destroy itself? The refcount may not go to zero in the end.

BusyJay, line 2704: What if the message is still in queue when the peer is about to destroy itself? The refcount may not go to zero in the end.

v01dstar: You are talking about the value stored inside the counter, not the ref count of the counter itself right? If the case happens, TiKV relies on PD to resend the plan again after a "timeout"

components/raftstore/src/store/peer.rs
v01dstar: Memory order does not matter here right, since we only do one read-modify-write operation per thread?

components/raftstore/src/store/peer.rs
```diff
@@ -4387,6 +4401,40 @@ where
...
         }
     }
+
+    pub fn unsafe_recovery_maybe_finish_wait_apply<T: Transport>(
+        &mut self,
+        ctx: &PollContext<EK, ER, T>,
+        force: bool,
+    ) {
+        if let Some(unsafe_recovery_state) = &self.unsafe_recovery_state {
+            debug!(
```
Connor1996, line 4411: can we remove it now?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -879,53 +870,24 @@ where
...
-        // If the applied index equals to the commit index, there is nothing to wait for, proceeds
-        // to the next step immediately. If they are not equal, further checks will be performed in
-        // on_apply_res().
-        if self.fsm.stopped
-            || self.fsm.peer.raft_group.store().applied_index()
-                == self.fsm.peer.raft_group.store().commit_index()
-        {
-            self.finish_unsafe_recovery_wait_apply();
+        info!(
+            "unsafe recovery wait apply";
```
Connor1996, line 875: print region id and peer id

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1662,10 +1624,11 @@ where
             }
         }
         // After a log has been applied, check if we need to trigger the unsafe recovery reporting procedure.
-        if let Some(target_commit_index) = self.fsm.unsafe_recovery_target_commit_index {
-            if self.fsm.peer.raft_group.store().applied_index() >= target_commit_index {
-                self.finish_unsafe_recovery_wait_apply();
-            }
+        if self.fsm.peer.unsafe_recovery_state.is_some() {
+            debug!("unsafe recovery finishes applying an entry");
```
Connor1996, line 1628: can we remove it?

components/raftstore/src/store/peer.rs
BusyJay: Sending control is also an atomic operation. Unless it doesn't show up in condition or there is further double check on the atomic, `Relaxed` should probably not be used.

components/raftstore/src/store/peer.rs
```diff
@@ -454,6 +454,13 @@ pub struct ReadyResult {
     pub has_write_ready: bool,
 }
 
+pub enum UnsafeRecoveryState {
+    WaitApply {
+        target_index: u64,
```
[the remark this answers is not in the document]
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Done

components/raftstore/src/store/fsm/peer.rs
v01dstar: Done

components/raftstore/src/store/peer.rs
```diff
@@ -4387,6 +4401,40 @@ where
...
+                "applied" =>  self.raft_group.raft.raft_log.applied,
+                "force" => force
+            );
+            let UnsafeRecoveryState::WaitApply {
+                target_index,
+                task_counter,
+            } = unsafe_recovery_state;
+            if self.raft_group.raft.raft_log.applied >= *target_index || force {
+                info!(
+                    "unsafe recovery finish wait apply";
```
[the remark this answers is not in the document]
v01dstar: Done

components/raftstore/src/store/peer.rs
v01dstar: Done.

components/raftstore/src/store/peer.rs
v01dstar: Control sending depends on the evaluation of the atomic operation. So the order is already assured?

BusyJay: For example, what if the heartbeat is sent with stale region info? Besides, it's not a performance sensitive path, using `Relaxed` doesn't introduce any noticeable improvement but always bring safety concerns.

v01dstar: Changed.

BusyJay: I can see it's still Relaxed.

tests/failpoints/cases/test_unsafe_recovery.rs
```diff
@@ -92,3 +86,71 @@ fn test_unsafe_recover_send_report() {
...
+    cluster.stop_node(nodes[1]);
+    // at least 4m data
+    (0..10).for_each(|_| cluster.must_put(b"random_k", b"random_v"));
+    // Sleep for a while to ensure all logs are compacted.
+    sleep_ms(100);
+    // Makes the group lose its quorum.
+    cluster.stop_node(nodes[2]);
+
+    // Blocks the raft snap apply process.
+    let apply_triggered_pair = Arc::new((Mutex::new(false), Condvar::new()));
```
BusyJay, line 118: Why not use channel?

v01dstar: `apply_triggered` is a multi-productions-one-consumption model, while `apply_released` is a one-production-multi-consumptions model. So, I think channel is not best option for both models here? Since the channel may be filled up or over consumed?

Connor1996 (approved):
LGTM

Connor1996 (commented):
BTW, Please change the `10` to `50` on `test_unsafe_recovery.rs::L365` to make the test more stable.

components/raftstore/src/store/peer.rs
v01dstar: Sorry, wrong place, changed now.

tests/failpoints/cases/test_unsafe_recovery.rs
BusyJay: You mean:
```
let (triggered_tx, triggered_rx) = mpsc::bounded(1);
let (released_tx, released_rx) = mpsc::bounded(1);
thread::spawn(move || {
let _ = triggered_tx.send(());
let _ = released_rx.recv(());
});

triggered_rx.recv_timeout().unwrap();
drop(release_tx);
```
?

tests/failpoints/cases/test_unsafe_recovery.rs
```diff
@@ -92,3 +72,71 @@ fn test_unsafe_recover_send_report() {
     assert_eq!(reported, true);
     fail::remove("on_handle_apply_store_1");
 }
+
+#[allow(clippy::mutex_atomic)]
```
BusyJay, line 76: The allow lint can be removed now.

BusyJay (approved):

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: Done

tests/failpoints/cases/test_unsafe_recovery.rs
v01dstar: Done

v01dstar:
The test failure is irrelevant to this PR.

BusyJay:
`test_unsafe_recover_wait_for_snapshot_apply` keep retrying will just bring more unstable case to master.

Connor1996:
> `test_unsafe_recover_wait_for_snapshot_apply` keep retrying will just bring more unstable case to master.

previous failure isn't due to `test_unsafe_recover_wait_for_snapshot_apply`

v01dstar:
> > `test_unsafe_recover_wait_for_snapshot_apply` keep retrying will just bring more unstable case to master.
> 
> previous failure isn't due to `test_unsafe_recover_wait_for_snapshot_apply`

I suspect that the test failed due to insufficient wait for the raft log GC, thus the snapshot path was not triggered at all. Added some new code to coordinate the data insertion and raft log GC, hope this can make the test bit more stable.

PR #12321, opened 2022-04-02
[DNM] *: replace PerfLevel with PerfFlags

sticnarf:

Issue Number: Close #xxx

What's Changed:

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/src/storage/mvcc/reader/point_getter.rs
+++ b/src/storage/mvcc/reader/point_getter.rs
@@ -692,19 +692,19 @@ mod tests {
         let mut getter = new_point_getter(&engine, TimeStamp::max());
         let perf_statistics = PerfStatisticsInstant::new();
         must_get_value(&mut getter, b"foo", b"bar");
-        assert_eq!(perf_statistics.delta().0.internal_delete_skipped_count, 0);
+        assert_eq!(perf_statistics.delta().internal_delete_skipped_count, 0);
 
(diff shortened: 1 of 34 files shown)


src/storage/mvcc/reader/point_getter.rs
```diff
@@ -692,19 +692,19 @@ mod tests {
         let mut getter = new_point_getter(&engine, TimeStamp::max());
         let perf_statistics = PerfStatisticsInstant::new();
         must_get_value(&mut getter, b"foo", b"bar");
-        assert_eq!(perf_statistics.delta().0.internal_delete_skipped_count, 0);
+        assert_eq!(perf_statistics.delta().internal_delete_skipped_count, 0);
```
breeswish, line 695: Looks like it is now possible to record critical per-request latency metrics for troubleshooting fs related jitters, like Mutex time and reading SST file time without much cost. Do you plan to add them later?

sticnarf: Yes. I'm going to record them.

sticnarf: Yes. I'm going to record them.

PR #12323, opened 2022-04-02
*: simplify setup

### LINKED ISSUE #12322: setup code has a lot of duplicated codes
## Development Task

There are a lot of duplicated code in `init_raw_engine`.

https://github.com/tikv/tikv/blob/379180b21cdbb2de55ba41cab6b084ec56b4de80/components/server/src/server.rs#L1305-L1308

And the setup of kv engine is complicated and need to be rewritten when try to initialized in other place like tests.

BusyJay:

Issue Number: Close #12322

What's Changed:

```commit-message
Use generic to reduce duplicated code and extract KvEngineFactory
for reuse in tests and future usage.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_traits/src/engine.rs
+++ b/components/engine_traits/src/engine.rs
@@ -57,3 +57,11 @@ pub trait KvEngine:
     /// It cannot be used forever.
     fn bad_downcast<T: 'static>(&self) -> &T;
 }
+
+/// A factory trait to create new engine.
+///
+// It should be named as `EngineFactory` for consistency, but we are about to rename
+// engine to tablet, so always use tablet for new traits/types.
+pub trait TabletFactory<EK> {
+    fn create_tablet(&self) -> Result<EK>;
+}
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -105,8 +105,8 @@ use crate::raft_engine_switch::*;
 use crate::{memory::*, setup::*, signal_handler};
 
 #[inline]
-fn run_impl<CEr: ConfiguredRaftEngine>(config: TiKvConfig) {
-    let mut tikv = TiKVServer::<CEr>::init(config);
+fn run_impl<CER: ConfiguredRaftEngine>(config: TiKvConfig) {
+    let mut tikv = TiKVServer::<CER>::init(config);
 
     // Must be called after `TiKVServer::init`.
     let memory_limit = tikv.config.memory_usage_limit.unwrap().0;
@@ -1363,11 +1363,11 @@ impl ConfiguredRaftEngine for RaftLogEngine {
     }
 }
 
-impl<CEr: ConfiguredRaftEngine> TiKVServer<CEr> {
+impl<CER: ConfiguredRaftEngine> TiKVServer<CER> {
     fn init_raw_engines(
         &mut self,
         flow_listener: engine_rocks::FlowListener,
-    ) -> (Engines<RocksEngine, CEr>, Arc<EnginesResourceInfo>) {
+    ) -> (Engines<RocksEngine, CER>, Arc<EnginesResourceInfo>) {
         let block_cache = self.config.storage.block_cache.build_shared_cache();
         let env = self
             .config
@@ -1375,7 +1375,7 @@ impl<CEr: ConfiguredRaftEngine> TiKVServer<CEr> {
             .unwrap();
 
         // Create raft engine
-        let raft_engine = CEr::build(self, &env, &block_cache);
+        let raft_engine = CER::build(self, &env, &block_cache);
 
         // Create kv engine.
         let mut builder = KvEngineFactoryBuilder::new(env, &self.config, &self.store_path)
@@ -1564,8 +1564,8 @@ pub struct EnginesResourceInfo {
 impl EnginesResourceInfo {
     const SCALE_FACTOR: u64 = 100;
 
-    fn new<CEr: ConfiguredRaftEngine>(
-        engines: &Engines<RocksEngine, CEr>,
+    fn new<CER: ConfiguredRaftEngine>(
+        engines: &Engines<RocksEngine, CER>,
         max_samples_to_preserve: usize,
     ) -> Self {
         let raft_engine = engines.raft.as_rocks_engine().cloned();
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -81,8 +81,8 @@ use tikv::{
         service::{DebugService, DiagnosticsService},
         status_server::StatusServer,
         ttl::TTLChecker,
-        KvEngineFactory, Node, RaftKv, Server, CPU_CORES_QUOTA_GAUGE, DEFAULT_CLUSTER_ID,
-        GRPC_THREAD_PREFIX,
+        KvEngineFactoryBuilder, Node, RaftKv, Server, CPU_CORES_QUOTA_GAUGE,
+        DEFAULT_CLUSTER_ID, GRPC_THREAD_PREFIX,
     },
     storage::{
         self, config_manager::StorageConfigManger, mvcc::MvccConsistencyCheckObserver,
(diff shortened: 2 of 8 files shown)


src/server/engine_factory.rs
```diff
@@ -0,0 +1,135 @@
...
+
+#[derive(Clone)]
+pub struct KvEngineFactory<ER: RaftEngine> {
+    inner: Arc<FactoryInner>,
+    router: Option<RaftRouter<RocksEngine, ER>>,
+}
+
+impl<ER: RaftEngine> KvEngineFactory<ER> {
+    #[inline]
+    pub fn new(
```
tabokie, line 32: That's a lot of `Option`. Could builder style be more readable?

tabokie (approved):

components/server/src/server.rs
```diff
@@ -1466,11 +1392,19 @@ impl TiKVServer<RaftLogEngine> {
                 self.config.storage.block_cache.shared,
             )),
         );
+        if let Some(e) = engines.raft.as_rocks_engine() {
```
Connor1996, line 1395: How about providing a func called `RegisterConfig` which accepts `cfg_controller` for `ConfiguredRaftEngine` instead of `as_rocks_engine`

components/server/src/server.rs
```diff
@@ -1433,28 +1351,36 @@ impl TiKVServer<RaftLogEngine> {
...
+    fn init_raw_engines(
+        &mut self,
+        flow_listener: engine_rocks::FlowListener,
+    ) -> (Engines<RocksEngine, Er>, Arc<EnginesResourceInfo>) {
+        let block_cache = self.config.storage.block_cache.build_shared_cache();
+        let env = self
+            .config
+            .build_shared_rocks_env(self.encryption_key_manager.clone(), get_io_rate_limiter())
+            .unwrap();
+
```
Connor1996, line 1368: // Create raft engine

components/server/src/server.rs
```diff
@@ -1433,28 +1351,36 @@ impl TiKVServer<RaftLogEngine> {
...
+        let raft_engine = Er::build(self, &env, &block_cache);
```
Connor1996, line 1369: how about adding a struct called `RaftEngineFactory` and use same manner as `KvEngineFactory`

components/engine_traits/src/engine.rs
```diff
@@ -57,3 +57,11 @@ pub trait KvEngine:
     /// It cannot be used forever.
     fn bad_downcast<T: 'static>(&self) -> &T;
 }
+
+/// A factory trait to create new engine.
+///
+// It should be named as `EngineFactory` for consistency, but we are about to rename
+// engine to tablet, so always use tablet for new traits/types.
```
Connor1996, line 64: I think `tablet` should replace the term —— `Engines` instead of `Engine`

components/server/src/server.rs
BusyJay: They are not the same. `KvEngineFactory` will be reused in the future when creating multiple rocksdb inside a TiKV process.

Connor1996: Oh, RaftEngine is shared

components/engine_traits/src/engine.rs
BusyJay: Actually, it should be
```
pub struct Engines<Tablet, RaftEngine> {
    raft_engine: RaftEngine,
    tablets: HashMap<u64, Tablet>,
}
```

components/server/src/server.rs
```diff
@@ -1433,28 +1359,36 @@ impl TiKVServer<RaftLogEngine> {
...
-            self.config.storage.api_version(),
-        );
-        let db_path = self.store_path.join(Path::new(DEFAULT_ROCKSDB_SUB_DIR));
-        let kv_engine = engine_rocks::raw_util::new_engine_opt(
-            db_path.to_str().unwrap(),
-            kv_db_opts,
-            kv_cfs_opts,
-        )
-        .unwrap_or_else(|s| fatal!("failed to create kv engine: {}", s));
+impl<Er: ConfiguredRaftEngine> TiKVServer<Er> {
```
Connor1996, line 1366: Er -> CR

components/server/src/server.rs
```diff
@@ -1466,11 +1400,12 @@ impl TiKVServer<RaftLogEngine> {
                 self.config.storage.block_cache.shared,
```
Connor1996, line 1400: I'd like to make it the same as raft engine, `engines.kv.register_config()`

components/server/src/server.rs
Connor1996: but `KvEngineFactory` should still focus on building a kv engine, we'd make them in the same manner

BusyJay: It seems over abstracted to me giving that it will not be used again in anywhere.

Connor1996: what I expected is 
```
fn init_raw_engines() {
    let env = xxx;
    let cfg_controller = xx;
    let engines = Engines::default();

    let raft_engine = RaftEngineBuilder::build();
    raft_engine.register_config(&cfg_controller);
    engines.raft = raft_engines;

    for tablet in tablets {
       let kv_engine = KvEngineBuilder::build();
       kv_engine.register_config(&cfg_controller);
       engines.push_kv(kv_engine);
    }

    engines
}
```

Connor1996: `RaftEngineFactory` doesn't introduce additional abstract, it is just a rewrite of `ConfiguredRaftEngine`

components/server/src/server.rs
BusyJay: I'm afraid it's impossible unless a similar `ConfiguredKvEngine` trait is definied. But it's just strange to do so as the only implementation is for RocksEngine, and only one method in the trait is `register_config`. Maybe it should be done when KvEngine is abstracted over RocksEngine, PanicsEngine and others.

Connor1996: let's revert the `register_config` change, and decuce if it is rocks engine by `config.raft_engine.enable` instead of `as_rocksb_engine`. Then `ConfiguredXXX` is not needed any more. You can simply unify `RaftEngine` and `KvEngine` in the builder pattern then.

components/engine_traits/src/engine.rs
Connor1996: can it be like this:
```
pub struct Engines<KvEngine, RaftEngine> {
    raft_engine: RaftEngine,
    tablets: HashMap<u64, KvEngine>,
}
```
tablet is like a type name, KvEngine should be still the trait name

components/server/src/server.rs
BusyJay: I don't get it. How can you make the compiler know it's `RocksEngine` without using `as_rocksdb_engine`?

Connor1996: if `self.config.raft_engine.enable` 🤣

BusyJay: `enable` is a runtime check, the type is inferred at compile time. It means you need to cast a generic to specific time at runtime.

Connor1996: but here doesn't need static check, `if raft.as_rocksdb_engine().is_some()` is a runtime check too.

BusyJay: I think there is some misunderstanding here. The point is not about runtime check, is about how to make compiler accept a generic and treat it as RocksEngine. This PR achieves this via`as_rocksdb_engine`, so compiler know how to convert a generic to a specific type. With `self.config.raft_engine.enable`, the generic is still a generic, it can't become `RocksEngine` all of a sudden.

It will be better explained by code samples to prove your point.

components/server/src/server.rs
BusyJay: So to implement `RaftEngineBuilder`, there has to be another trait `RaftEngineBuilder` and then implement `RaftEngineBuilder` to `RaftLogEngineBuilder` and `RocksEngineAsRaftEngineBuilder`.

To implement `register_config`, there should be an extra trait `RegisterConfigTrait` and implement it to RocksEngine and RaftLogEngine.

So there will be two traits and two other types compared to current implementation, which only requires only one trait.

components/server/src/server.rs
Connor1996: Oh, you mean `EnginesResourceInfo` needs a `RocksEngine`, but what we have is `RaftEngine`.

components/engine_traits/src/engine.rs
BusyJay: Tablet is not a type name, it's an variable name. The point of renaming it is `TabletFactory` will not just act as a `KvEngine` factory, but it also manages tablets that serve the queries to tablets and control their lifetime.

Connor1996 (approved):
LGTM

components/server/src/server.rs
```diff
@@ -1433,28 +1359,36 @@ impl TiKVServer<RaftLogEngine> {
...
-            self.config.storage.api_version(),
-        );
-        let db_path = self.store_path.join(Path::new(DEFAULT_ROCKSDB_SUB_DIR));
-        let kv_engine = engine_rocks::raw_util::new_engine_opt(
-            db_path.to_str().unwrap(),
-            kv_db_opts,
-            kv_cfs_opts,
-        )
-        .unwrap_or_else(|s| fatal!("failed to create kv engine: {}", s));
+impl<CEr: ConfiguredRaftEngine> TiKVServer<CEr> {
```
Connor1996, line 1366: the naming of `CEr` is weird, seems we use capital words throughout TiKV

PR #12326, opened 2022-04-06
status_server: unwind the backtrace through frame pointer

YangKeao:

This PR attempts to use the frame pointer implementation of `pprof-rs` to unwind the backtrace. During the implementation, I faced three problems:

1. Some functions in the rust standard library doesn't follow the rule of frame pointer. (I found this problem by reading the disassembly result of the frame in standard library, it locates in some codes related with `Debug`.)
2. The `libstdc++.so` in my PC is not compiled with `-fno-omit-frame-pointer`. (But as I'm using gentoo, I can simply compile it again, but I don't know whether the binary distributed in other linux distributions enabled the frame pointer)
3. `tikv` doesn't enable the `no-omit-frame-pointer` feature of the grpc.

**A relatively safe way to avoid these three problems is to compile the tikv in this branch with:**

```
RUSTFLAGS=-Cforce-frame-pointers=yes cargo -Z build-std build --target x86_64-unknown-linux-gnu  --release
``` 

It will recompile the standard library from source and enable the frame pointer. I have tested the tikv under `tpcc` benchmark with the continuous profiling enabled, it doesn't panic for longer than 12 hours. But it still needs a further test on Cent OS, ubuntu and other famous distributions.

The https://github.com/tikv/pprof-rs/pull/116 implementation on `pprof-rs` also protects the read/write of the memory. The read address will be checked though `write(fd, addr, 1)` to a pipe. If the memory is inaccessible, this function call will return an error.

```release-note
unwind the backtrace through frame pointer
```


YangKeao:
Close

As @mornyx is working on https://github.com/tikv/tikv/pull/12480

PR #12328, opened 2022-04-06
txn: remove unnecessary length check in SnapshotStore::batch_get

SpadeA-Tang:

Issue Number: Close #xxx

What's Changed:

```
Checking batch length == 1 is what we did when `multi` field of `PointGetter`was not removed and we set 'multi` be false in that case. Now, as `multi` field is removed, it is unnecessary for us to check the length.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.
```


SpadeA-Tang:
cc @youjiali1995 @sticnarf

youjiali1995 (approved):

sticnarf (approved):
LGTM

PR #12335, opened 2022-04-07
server: support dynamic config max-grpc-send-msg-len and raft-msg-max-batch-size

### LINKED ISSUE #12334: online config: support max-grpc-send-msg-len

support change grpc message size related configs as follows:
```
set config tikv `server.max-grpc-send-msg-len`=33554432;
set config tikv `server.raft-msg-max-batch-size`=1024;
set config tikv `server.grpc-memory-pool-quota` = 104857600;
set config tikv `raftstore.raft-entry-max-size` = "32MiB";
```

glorv:

Issue Number: Close #12334

What's Changed:

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


glorv:
@5kbpers  @Connor1996  PTAL

5kbpers (approved):

5kbpers:
CI failed

glorv:
> CI failed

All failures are caused by github connection timeout.😂️

src/server/raft_client.rs
```diff
@@ -201,14 +234,10 @@ impl Buffer for BatchMessageBuffer {
...
-        let mut msg_size = msg.start_key.len()
-            + msg.end_key.len()
-            + msg.get_message().context.len()
-            + msg.extra_ctx.len()
-            // index: 3, term: 2, data tag and size: 3, entry tag and size: 3
-            + 11 * msg.get_message().get_entries().len();
-        for entry in msg.get_message().get_entries() {
-            msg_size += entry.data.len();
+        let msg_size = Self::message_size(&msg);
+        // try refrech config before check
```
Connor1996, line 238: refrech -> refresh

src/server/server.rs
```diff
@@ -163,7 +163,7 @@ impl<T: RaftStoreRouter<E::Local> + Unpin, S: StoreAddrResolver + 'static, E: En
 
         let conn_builder = ConnectionBuilder::new(
             env.clone(),
-            Arc::new(cfg.value().clone()),
+            Arc::clone(cfg),
```
Connor1996, line 166: cfg.clone()

Connor1996 (commented):
rest LGTM

src/server/raft_client.rs
glorv, line 238:
```suggestion
        // try refresh config before check
```

glorv:
@Connor1996 PTAL again, thanks.

components/raftstore/src/store/config.rs
```diff
@@ -461,6 +460,12 @@ impl Config {
             ));
         }
 
+        if self.raft_entry_max_size.0 == 0 || self.raft_entry_max_size.0 > ReadableSize::gb(3).0 {
+            return Err(box_err!(
+                "raft entry max size should be greater than 0 and less than or equal to 3GiB"
```
BusyJay, line 465: I don't think we support 3GiB. Not even 2.01GiB.

src/server/raft_client.rs
```diff
@@ -201,14 +234,10 @@ impl Buffer for BatchMessageBuffer {
...
-            + msg.end_key.len()
-            + msg.get_message().context.len()
-            + msg.extra_ctx.len()
-            // index: 3, term: 2, data tag and size: 3, entry tag and size: 3
-            + 11 * msg.get_message().get_entries().len();
-        for entry in msg.get_message().get_entries() {
-            msg_size += entry.data.len();
+        let msg_size = Self::message_size(&msg);
+        // try refresh config before check
+        if let Some(new_cfg) = self.cfg_tracker.any_new() {
```
BusyJay, line 239: Have you bench the performance impact?

src/server/config.rs
```diff
@@ -83,7 +83,6 @@ pub struct Config {
     #[online_config(skip)]
     pub status_thread_pool_size: usize,
 
-    #[online_config(skip)]
```
BusyJay, line 86: I don't see the point of changing this value online. It's only for controlling the size of a batch.

src/server/raft_client.rs
glorv: In my benchmark, this is no notable impact after this change. In the most common case, this is only one extra atomic load.

src/server/config.rs
glorv: The use case is from tidb-cloud to minimize the memory usage in a minimal test cluster. This kind of scenario is not common. I think it is harmless to support it.

components/raftstore/src/store/config.rs
glorv: I set this to the vlaue as `raft-max-size-per-msg`, this is just an intuitive upper bound, do you mean the hard limit should be 2GB? Sorry I don't find the related code.

src/server/raft_client.rs
BusyJay: Actually the atomic load is way heavier than the original that just sums up several numbers. A better place to check for new version is when a message is sent.

src/server/config.rs
BusyJay: The batch is about 10MiB, and it only consumes about 10MiB * grpc_concurrency = 40MiB.

src/server/raft_client.rs
glorv: So do you mean that we move this check into `flush` call as the frequency of flush should be much lower than `push` especially when the lead is heavy.

BusyJay: Yes.

PR #12340, opened 2022-04-08
cdc: Add supporting rawkv cdc codes.

zeminzhou:
Add codes that can support rawkv cdc.

Issue Number: ref [#11965](https://github.com/tikv/tikv/issues/11965)

What's Changed:
1. components/cdc/src/initializer.rs;
2. components/cdc/src/delegate.rs;

### Check List 

Tests 

- Unit test
- Integration test


zeminzhou:
@pingyu PTAL~, thanks

components/cdc/src/endpoint.rs
```diff
@@ -595,6 +597,11 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
             return;
         }
 
+        if (kv_api == ChangeDataRequestKvApi::RawKv || kv_api == ChangeDataRequestKvApi::TxnKv) && api_version != 2 {
+            error!("only api version2 supportes the change data request of RawKv and TxnKv"; "api_version" => api_version);
```
pingyu, line 601: `downstream.sink_error_event` to notify CDC.

pingyu:
```suggestion
            error!("RawKv and TxnKv are supported by api-version 2 only"; "api_version" => api_version);
```

components/cdc/src/initializer.rs
```diff
@@ -224,6 +232,30 @@ impl<E: KvEngine> Initializer<E> {
...
+                .hint_min_ts(hint_min_ts)
+                .build_delta_scanner(self.checkpoint_ts, TxnExtraOp::ReadOldValue)
+                .unwrap();
+
+            scanner_iter = Some(ScannerIter::Scanner(scanner));
+        } else {
+            let mut iter_opt = IterOptions::default();
+            iter_opt.set_lower_bound(b"r", 1);
+            iter_opt.set_upper_bound(b"s", 1);
+            let mut iter = RawMvccSnapshot::from_snapshot(snap)
```
pingyu, line 249: For CDC, the most scenario is scan from a checkpoint timestamp to now. Should not scan all versions.

components/cdc/src/initializer.rs
```diff
@@ -224,6 +232,30 @@ impl<E: KvEngine> Initializer<E> {
...
+                .fill_cache(false)
+                .range(None, None)
+                .hint_min_ts(hint_min_ts)
+                .build_delta_scanner(self.checkpoint_ts, TxnExtraOp::ReadOldValue)
+                .unwrap();
+
+            scanner_iter = Some(ScannerIter::Scanner(scanner));
+        } else {
+            let mut iter_opt = IterOptions::default();
+            iter_opt.set_lower_bound(b"r", 1);
```
pingyu, line 247: Every puller from CDC request for data of only one region. So we should use KeySpan here to limit cost of scan.

components/cdc/src/initializer.rs
```diff
@@ -292,15 +324,33 @@ impl<E: KvEngine> Initializer<E> {
...
-                Some(mut entry) => {
-                    read_old_value(entry.old_value(), &mut stats)?;
-                    total_bytes += entry.size();
-                    entries.push(Some(entry));
-                }
-                None => {
-                    entries.push(None);
-                    break;
+            if let Some(scanner_iter) = scanner_iter {
+                match scanner_iter {
```
pingyu, line 328: Consider place the judgement outside the loop.

components/cdc/src/delegate.rs
```diff
@@ -887,6 +967,27 @@ fn decode_lock(key: Vec<u8>, lock: Lock, row: &mut EventRow) -> bool {
     false
 }
 
+fn decode_rawkv(key: Vec<u8>, value: Vec<u8>, row: &mut EventRow) {
```
pingyu, line 970: Add unit tests for `decode_rawkv`.

PR #12356, opened 2022-04-13
server: support gzip compression for metrics

### LINKED ISSUE #12355: tikv should support gzip compression for metrics response

glorv:

Issue Number: ref #12355

What's Changed:
Support write metrics response with gzip compression. If the request header contains `Accept-Encoding: gzip`, tikv server will return the response with gzip compressed; If not, still return no-compressed response as before.

In my test in a local environment, use gzip compression can reduce the body size to about 1/12.
```
$ curl -v -i --compressed -s 127.0.0.1:20180/metrics
...
< HTTP/1.1 200 OK
< content-encoding: gzip
< content-type: text/plain; charset=utf-8
< content-length: 118739
< date: Mon, 18 Apr 2022 08:59:22 GMT

$ curl -v -i -s 127.0.0.1:20180/metrics
< HTTP/1.1 200 OK
< content-type: text/plain; charset=utf-8
< content-length: 1430175
< date: Mon, 18 Apr 2022 08:58:24 GMT
```

Performance Impact:
The gzip compression will cost about 12ms extra cpu time, which should not be the bottleneck in a typical 8c environment.

In my benchmark with (1*tidb(40c), 1*pd, 3*tikv(8c,16gb)) with 3k tpcc workload. With a concurrency of 200, each tikv's cpu usage is 700+, compare between v6.0.0 and this commit(based on v6.0.0), there is no visiable performance impact for all of tmpC, P99, P999 metrics.  

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

### Release note 

```release-note
Support encode metrics response with gzip to reduce the http body size
```


BusyJay:
383k doesn't seem a large cost.

glorv:
> 383k doesn't seem a large cost.

Sorry for the misleading exmple. This example is from a newly bootstraped cluster. For a typical online cluster the body size of the metrics request is about 1.4MiB.  /cc @zhangjinpeng1987 Do you have a brief statictics of how the metrics traffic cost in our cloud environment?

Cargo.lock
```diff
@@ -2720,6 +2721,16 @@ version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
 
+[[package]]
+name = "miniz-sys"
```
BusyJay, line 2725: Why this is added?

BusyJay (commented):
What's the impact on Prometheus?

Cargo.lock
glorv: It's depended by the [flate2](https://crates.io/crates/flate2) package.

glorv:
> What's the impact on Prometheus?

There is no impact. Prometheus supports and prefers gzip compressed http response.  See: https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

BusyJay:
> There is no impact. Prometheus supports and prefers gzip compressed http response. See: https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

Decoding gzip can't have the same performance as reading plain text. I'm worry about the performance when there are a lot of nodes.

glorv:
> > There is no impact. Prometheus supports and prefers gzip compressed http response. See: https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format
> 
> Decoding gzip can't have the same performance as reading plain text. I'm worry about the performance when there are a lot of nodes.

In my benchmark at local environment, for a 1.4MB length metrics content , both gzip encode and decode cost less than 5ms cpu time. Thus, given a cluster with 900 tikv (with is likely to store more than 1PB data), the average metrics qps 900/15 = 60/s, so the extra cpu needed is 5ms * 60 = 300ms. So we only need an extra of 0.3 cores. I think this must not be a bottleneck. I'm not sure if there are some test clusters big enough to check this performance regression.

Cargo.lock
BusyJay: flate2 has been a part of dependency before but miniz-sys is added in this PR.

glorv: Thanks for the reminding. I change flate2 to use zlib instead of miniz instead.

Cargo.toml
```diff
@@ -90,6 +90,7 @@ engine_traits_tests = { path = "components/engine_traits_tests", default-feature
 error_code = { path = "components/error_code", default-features = false }
 fail = "0.5"
 file_system = { path = "components/file_system", default-features = false }
+flate2 = { version = "=1.0.11", default-features = false, features = ["zlib"] }
```
BusyJay, line 93: Why `=`?

src/server/status_server/mod.rs
```diff
@@ -495,6 +502,36 @@ where
         }
     }
 
+    fn handle_get_metrics(req: Request<Body>) -> hyper::Result<Response<Body>> {
+        // the following logic is port from prometheus's golang:
+        // https://github.com/prometheus/client_golang/blob/24172847e35ba46025c49d90b8846b59eb5d9ead/prometheus/promhttp/http.go#L155-L176
+        let encoding = req
```
BusyJay, line 508: This may be extracted as a function `client_accept_gzip`.

Cargo.toml
glorv: I just copy from `tidb_query_expr` component. Changed to `1.0`.

src/server/status_server/mod.rs
glorv: done

glorv:
@BusyJay @5kbpers  PTAL again, thanks~

src/server/status_server/mod.rs
```diff
@@ -1327,6 +1372,61 @@ mod tests {
...
 
+    #[test]
+    fn test_metrics() {
+        let _test_guard = TEST_PROFILE_MUTEX.lock().unwrap();
+        let mut status_server = StatusServer::new(
+            1,
+            ConfigController::default(),
+            Arc::new(SecurityConfig::default()),
+            MockRouter,
+            std::env::temp_dir(),
```
BusyJay, line 1383: Should use `tempfile::TempDir` instead, otherwise they won't be cleaned up.

src/storage/config.rs
```diff
@@ -381,7 +381,7 @@ mod tests {
         let mut cfg = Config::default();
         assert!(cfg.validate().is_ok());
 
-        let max_pool_size = std::cmp::max(1, SysQuota::cpu_cores_quota() as usize);
+        let max_pool_size = std::cmp::max(4, SysQuota::cpu_cores_quota() as usize);
```
BusyJay, line 384: Why the change?

glorv: the original test logic is wrong, change the logic to make the test logic right. 
See: https://github.com/tikv/tikv/blob/1fb8980ccab9ff40c1adc206df52952dab8e8ad8/src/storage/config.rs#L119-L128

src/server/status_server/mod.rs
glorv: done. Replaced all with `tempfile::TempDir`

glorv:
@BusyJay PTAL again, thanks

BusyJay (approved):

5kbpers (approved):

PR #12363, opened 2022-04-14
Evict leader applied entries when memory insufficient

tier-cap:

Issue Number: Close #xxx

What's Changed:

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4650,7 +4657,7 @@ where
         if !self.fsm.peer.get_store().is_cache_empty() || !self.ctx.cfg.hibernate_regions {
             self.register_raft_gc_log_tick();
         }
-        fail_point!("on_raft_log_gc_tick_1", self.fsm.peer_id() == 1, |_| {});
+        fail_point!("on_raft_gc_log_tick_1", self.fsm.peer_id() == 1, |_| {});
```
Connor1996, line 4660: a lot of spots use `raft_log_gc`, let's rename `raft_gc_log` to `raft_log_gc`

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4778,9 +4785,9 @@ where
 
     fn on_entry_cache_evict_tick(&mut self) {
         fail_point!("on_entry_cache_evict_tick", |_| {});
-        if needs_evict_entry_cache(self.ctx.cfg.evict_cache_on_memory_ratio) {
-            self.fsm.peer.mut_store().evict_cache(true);
-        }
+        // call evict directly here, no need to recheck the memory usage condition.
```
Connor1996, line 4788: why

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1404,14 +1405,11 @@ where
...
-    pub fn evict_cache(&mut self, half: bool) {
+    /// Evict applied entries to save more memory.
+    pub fn evict_cache(&mut self) {
         if !self.cache.cache.is_empty() {
             let cache = &mut self.cache;
-            let cache_len = cache.cache.len();
-            let drain_to = if half { cache_len / 2 } else { cache_len - 1 };
-            let idx = cache.cache[drain_to].index;
-            let mem_size_change = cache.compact_to(idx + 1);
+            let mem_size_change = cache.compact_to(self.apply_state.applied_index + 1);
```
Connor1996, line 1412: it may cause raft to fetch entries in disk more frequently, consider using replicated_idx which you can refer to `on_raft_gc_log_tick`

tier-cap: 1. currently will not work. because the leader proposal flow is not controled by this. If we use replicated index, then if one peer is slow, oom will happen easily.

components/raftstore/src/store/fsm/peer.rs
tier-cap: Because we only register the reclaim tick when memory is insufficient.

tier-cap:
@Connor1996 @5kbpers

components/raftstore/src/store/peer_storage.rs
tonyxuqqi: Is this evict_cache running in fsm thread----Will it block fsm message processing?
If yes, then one way to reduce disk IO is that we may check if replicated_idx is far away from applied_index. If it's not far away, then compact_to replicated_idx is enough to avoid additional IO read. But if it's far away, then compact_to applied_index may be inevitable to save enough memory.

components/raftstore/src/store/config.rs
```diff
@@ -345,7 +346,7 @@ impl Default for Config {
             dev_assert: false,
             apply_yield_duration: ReadableDuration::millis(500),
             perf_level: PerfLevel::Uninitialized,
-            evict_cache_on_memory_ratio: 0.0,
+            evict_cache_on_memory_ratio: 0.15,
```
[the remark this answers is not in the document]
tier-cap: I think it should be, and I also not want to leave this setting to users. I'll update it.

components/raftstore/src/store/config.rs
```diff
@@ -603,9 +604,11 @@ impl Config {
             );
         }
 
-        if self.evict_cache_on_memory_ratio < 0.0 {
+        if self.evict_cache_on_memory_ratio < 0.0 || self.evict_cache_on_memory_ratio > 0.3 {
```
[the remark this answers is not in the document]
tier-cap: Right, also generate a warning to tell users.

components/raftstore/src/store/peer_storage.rs
tier-cap: One point, the `far away` is not a good definition. Because this is a peer level, not the whole level. Even if each peer doesn't trigger this condition, then the whole memory can still takes huge.

tier-cap: Maybe, we can try a two-level like way to evict cache.
First to evict only the replicated logs. If not, then Second, evict the applied.

Connor1996 (approved):
LGTM

components/raftstore/src/store/config.rs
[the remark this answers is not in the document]
ethercflow: @tier-cap Will update it?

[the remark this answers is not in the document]
ethercflow: @tier-cap Will update this?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1807,12 +1807,19 @@ where
         }
         self.ctx.pending_count += 1;
         self.ctx.has_ready = true;
+
+        // follower will clean it as soon as applied.
```
ethercflow, line 1811: Hi, @Connor1996 . I have a question, will the witness raft log gc part we do conflict with the oom prevention mechanism of this PR?

Connor1996: this pr just clean the cache, not do the raft log gc

components/raftstore/src/store/fsm/peer.rs
ethercflow: @tier-cap I don't know the details, just a question: 

Is it possible that between registration and response, other parts of memory have been reclaimed, at this point there has been enough memory to not need evict?

components/raftstore/src/store/config.rs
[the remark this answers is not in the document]
tier-cap: Instead, add memory configure check log if configure not match such suggest.

components/raftstore/src/store/config.rs
```diff
@@ -83,6 +83,9 @@ pub struct Config {
     pub raft_engine_purge_interval: ReadableDuration,
     // When a peer is not responding for this time, leader will not keep entry cache for it.
     pub raft_entry_cache_life_time: ReadableDuration,
+    // when memory insufficient, shile some peer hearbeat missing for this time,
```
kevin-xianliu, line 86: shile?  TYPO?

src/config.rs
```diff
@@ -3140,6 +3140,54 @@ impl TiKvConfig {
...
+                warn!(
+                    "0.25 suggested for raftstore.evict-cache-on-memory-ratio \
+                    while TiKV memory limit {} greater than 128GB.",
+                    memory_limit
+                );
+            }
+        }
+
         Ok(())
     }
```
kevin-xianliu, line 3193: Suggest to refactor those magic numbers such as 0.15, 0.2, 0.25, 0.3 for better readability.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4947,12 +4954,6 @@ where
             .peer
             .mut_store()
             .maybe_gc_cache(alive_cache_idx, applied_idx);
-        if needs_evict_entry_cache(self.ctx.cfg.evict_cache_on_memory_ratio) {
-            self.fsm.peer.mut_store().evict_cache(true);
-            if !self.fsm.peer.get_store().cache_is_empty() {
-                self.register_entry_cache_evict_tick();
```
kevin-xianliu, line 4953: Cache entries won't be evicted even exceeding evict_cache_on_memory_ratio? Any side effect if we remove this?

components/raftstore/src/store/config.rs
tier-cap: Done!

components/raftstore/src/store/fsm/peer.rs
tier-cap: This function is called very low frequently, which is not enough to avoid oom. So, remove it is no impact.

tier-cap: This function is called very low frequently, which is not enough to avoid oom. So, remove it is no impact.

components/raftstore/src/store/fsm/peer.rs
tier-cap: right, we only evict at most applied entries, not all.

src/config.rs
tier-cap: Haven't no suitable candidate names yet.
Any good advice?

tier-cap: Haven't no suitable candidate names yet.
Any good advice?

PR #12364, opened 2022-04-15
raftstore: Wait ticks for hibernated peer when doing force leader

Connor1996:

Issue Number: Ref #10483

What's Changed:

```commit-message
Force leader is rejected on a peer who is already a leader. For the hibernated leader,
it doesn't step down to follower when quorum is missing due to not ticking election. 
So wait ticks in force leader process for hibernated peers to make sure election ticking
is performed.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1521,10 +1521,6 @@ where
         }) = &mut self.fsm.peer.force_leader
         {
             *ticks -= 1;
-            debug!(
-                "sub tick";
-                "ticks" => *ticks,
-            );
             if *ticks == 0 {
                 let s = mem::take(failed_stores);
                 self.on_enter_pre_force_leader(s);
@@ -1900,12 +1896,6 @@ where
             }
         }
 
-        debug!(
-            "tick raft group";
-            "region_id" => self.region_id(),
-            "peer_id" => self.fsm.peer_id(),
-            "election_elapsed" => self.fsm.peer.raft_group.raft.election_elapsed,
-        );
         // Tick the raft peer and update some states which can be changed in `tick`.
         if self.fsm.peer.raft_group.tick() {
             self.fsm.has_ready = true;
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1268,16 +1268,16 @@ where
             }
             // wait two rounds of election timeout to trigger check quorum to step down the leader
             Some(
-                self.ctx.cfg.raft_election_timeout_ticks * 2
-                    - self.fsm.peer.raft_group.raft.election_elapsed
+                (self.ctx.cfg.raft_election_timeout_ticks * 2)
+                    .saturating_sub(self.fsm.peer.raft_group.raft.election_elapsed)
                     + 1,
             )
+        // When election timeout is triggered, leader_id is set to INVALID_ID.
+        // But learner(not promotable) is a exception here as it wouldn't tick
+        // election.
         } else if self.fsm.peer.raft_group.raft.promotable()
             && self.fsm.peer.leader_id() != raft::INVALID_ID
         {
-            // When election timeout is triggered, leader_id is set to INVALID_ID.
-            // But learner(not promotable) is a exception here as it wouldn't tick
-            // election.
             if self.fsm.hibernate_state.group_state() == GroupState::Ordered {
                 warn!(
                     "reject pre force leader due to leader lease may not expired";
@@ -1288,8 +1288,10 @@ where
             }
             // wait one round of election timeout to make sure leader_id is invalid
             Some(
-                self.ctx.cfg.raft_election_timeout_ticks
-                    - self.fsm.peer.raft_group.raft.election_elapsed
+                self.ctx
+                    .cfg
+                    .raft_election_timeout_ticks
+                    .saturating_sub(self.fsm.peer.raft_group.raft.election_elapsed)
                     + 1,
             )
         } else {
@@ -1478,7 +1480,7 @@ where
             ticks,
         }) = &mut self.fsm.peer.force_leader
         {
-            *ticks -= 1;
+            *ticks = (*ticks).saturating_sub(1);
             if *ticks == 0 {
                 let s = mem::take(failed_stores);
                 self.on_enter_pre_force_leader(s);
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1267,10 +1267,10 @@ where
                 return;
             }
             // wait two rounds of election timeout to trigger check quorum to step down the leader
+            // note: check quorum is triggered every `election_timeout` instead of `randomized_election_timeout`
             Some(
-                (self.ctx.cfg.raft_election_timeout_ticks * 2)
-                    .saturating_sub(self.fsm.peer.raft_group.raft.election_elapsed)
-                    + 1,
+                self.fsm.peer.raft_group.raft.election_timeout() * 2
+                    - self.fsm.peer.raft_group.raft.election_elapsed,
             )
         // When election timeout is triggered, leader_id is set to INVALID_ID.
         // But learner(not promotable) is a exception here as it wouldn't tick
@@ -1278,7 +1278,9 @@ where
         } else if self.fsm.peer.raft_group.raft.promotable()
             && self.fsm.peer.leader_id() != raft::INVALID_ID
         {
-            if self.fsm.hibernate_state.group_state() == GroupState::Ordered {
+            if self.fsm.hibernate_state.group_state() == GroupState::Ordered
+                || self.fsm.hibernate_state.group_state() == GroupState::Chaos
+            {
                 warn!(
                     "reject pre force leader due to leader lease may not expired";
                     "region_id" => self.fsm.region_id(),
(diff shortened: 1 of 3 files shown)


Connor1996:
PTAL @v01dstar

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1481,6 +1515,23 @@ where
...
+        }) = &mut self.fsm.peer.force_leader
+        {
+            *ticks -= 1;
+            debug!(
+                "sub tick";
+                "ticks" => *ticks,
+            );
+            if *ticks == 0 {
+                let s = mem::take(failed_stores);
+                self.on_enter_pre_force_leader(s);
```
v01dstar, line 1530: Instead of calling back on `on_enter_pre_force_leader()` maybe it looks cleaner if we separate the `on_enter_pre_force_leader()` into several functions like

```
SignificantMsg::EnterForceLeaderState => {
    self.on_enter_force_leader();
}

fn on_enter_force_leader() {
    if state.is_some() {
        ...
    }
    if is_leader || （promotable && has_leader）{
        self.leader_absence_check()
    } else {
        self.voter_liveness_check()
    }
}

fn check_force_leader() {
    if leader_absence_check_passes {
        self.voter_liveness_check()
    }
    if voter_liveness_check_passes {
        ...
    }
}
```

Connor1996: After waiting ticks, we need to re-check all the conditions in `on_enter_pre_force_leader`. Reusing is straightforward.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1251,30 +1251,64 @@ where
...
+                    "reject pre force leader due to leader lease may not expired";
+                    "region_id" => self.fsm.region_id(),
+                    "peer_id" => self.fsm.peer_id(),
+                );
+                return;
+            }
+            // wait one round of election timeout to make sure leader_id is invalid
+            Some(
+                self.ctx.cfg.raft_election_timeout_ticks
+                    - self.fsm.peer.raft_group.raft.election_elapsed
```
BusyJay, line 1292: It can overflow.

components/raftstore/src/store/fsm/peer.rs
v01dstar: That's also something we can optimize here, instead of waiting for certain ticks, and then check the leader absence in the  `on_enter_pre_force_leader()`, maybe we can proactively check whether the leader is absent or not in each tick at `check_force_leader()`?

Connor1996: Can't. For leader, `CheckQuorum` needs to wait for a round to see if peers are recent active; for follower, only after an election timeout, it can make sure the leader lease is expired. Anyway, both of them need to wait for some ticks to get the correct information.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1251,30 +1251,66 @@ where
...
+                "enter wait ticks";
                 "region_id" => self.fsm.region_id(),
                 "peer_id" => self.fsm.peer_id(),
+                "ticks" => ticks,
             );
+            self.fsm.peer.force_leader = Some(ForceLeaderState::WaitTicks {
+                failed_stores,
+                ticks,
+            });
+            self.reset_raft_tick(GroupState::Ordered);
```
BusyJay, line 1312: Why use `Ordered`? What if it's `Chaos` before?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1439,6 +1475,19 @@ where
 
     #[inline]
     fn check_force_leader(&mut self) {
+        if let Some(ForceLeaderState::WaitTicks {
+            failed_stores,
+            ticks,
+        }) = &mut self.fsm.peer.force_leader
+        {
+            *ticks = (*ticks).saturating_sub(1);
```
BusyJay, line 1483: The defense is unnecessary as it's always checked at the very following line.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1251,30 +1251,66 @@ where
...
+        let ticks = if self.fsm.peer.is_leader() {
+            if self.fsm.hibernate_state.group_state() == GroupState::Ordered {
+                warn!(
+                    "reject pre force leader due to already being leader";
+                    "region_id" => self.fsm.region_id(),
+                    "peer_id" => self.fsm.peer_id(),
+                );
+                return;
+            }
+            // wait two rounds of election timeout to trigger check quorum to step down the leader
```
BusyJay, line 1269: It should check quorum the very time elapsed is equal to timeout.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1251,30 +1251,66 @@ where
...
-                "reject pre force leader due to leader lease may not expired";
+            if self.fsm.hibernate_state.group_state() == GroupState::Ordered {
+                warn!(
+                    "reject pre force leader due to leader lease may not expired";
+                    "region_id" => self.fsm.region_id(),
+                    "peer_id" => self.fsm.peer_id(),
+                );
+                return;
+            }
+            // wait one round of election timeout to make sure leader_id is invalid
```
BusyJay, line 1289: Leader is always valid until it starts election.

components/raftstore/src/store/fsm/peer.rs
Connor1996: Need to trigger `CheckQuorum` twice, the first round is to set `recent_active = false`, the second round is to step down the leader. The leader may not step down within one round, due to `recent_active` may be true

components/raftstore/src/store/fsm/peer.rs
Connor1996: Yes, what if `ticks` is initialized to `0` at first? Though it wouldn't happen in current codebase.

components/raftstore/src/store/fsm/peer.rs
Connor1996: Yes, so wait timeout to trigger election

Connor1996: What's your point here?

components/raftstore/src/store/fsm/peer.rs
Connor1996: I want to keep ticking now to make sure a possible election timeout or check quorum is triggered within a finite time span.

BusyJay: I know why you want to register tick, but I don't understand why change group state.

components/raftstore/src/store/fsm/peer.rs
BusyJay: Then you can check if it's 0 first and then do the minus. And there is no need to `+1` any more.

components/raftstore/src/store/fsm/peer.rs
Connor1996: because I want to exit hibernate state if any, otherwise ticking is paused  and`check_force_leader` won't be called, and force leader is blocked for a long time.

components/raftstore/src/store/fsm/peer.rs
BusyJay: elapsed passes election timeout doesn't mean it will start elections. Should check with double timeout.

Connor1996: oh, should be `randomized_election_timeout`

components/raftstore/src/store/fsm/peer.rs
BusyJay: It still doesn't explain why `Ordered` is used.

Connor1996: I don't understand. So which one should it use?

BusyJay: It should behave just like accepting a new request.

Connor1996: Seems it just sets to `Ordered` when proposing

BusyJay: No, when it's follower, it will set it to `Chaos`.

tests/integrations/raftstore/test_unsafe_recovery.rs
```diff
@@ -300,6 +300,54 @@ fn test_force_leader_for_learner() {
...
+    let region = cluster.get_region(b"k2");
+    let peer_on_store1 = find_peer(&region, 1).unwrap();
+    cluster.must_transfer_leader(region.get_id(), peer_on_store1.clone());
+
+    // wait a while to hibernate
+    std::thread::sleep(Duration::from_millis(
+        cluster.cfg.raft_store.raft_election_timeout_ticks as u64
+            * cluster.cfg.raft_store.raft_base_tick_interval.as_millis()
+            * 3,
+    ));
```
v01dstar, line 325: It would be better if we assert the target region is hibernated here. Something like `assert_eq!(is_hibernated(region.get_id()), ture);`

Connor1996: can't, it's not exported

BusyJay (approved):

v01dstar (approved):

tabokie (approved):

PR #12369, opened 2022-04-18
txn: make prewrite idempotent after a new lock is written

sticnarf:

Issue Number: Close https://github.com/pingcap/tidb/issues/34066

What's Changed:

```commit-message
If a retried prewrite encounters a newer lock, the result may be not
idempotent.

This happens when the first prewrite succeeds and the transaction
is committed by resolving async-commit locks. Then, TiKV returns
KeyIsLocked in the retry while the first prewrite succeeds.

If the client does not handle it carefully, the client may report a false
commit failure to the user.

This commit fixes the bug, making prewrite idempotent in this case.
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Fixes that successfully committed optimistic transactions may report false WriteConflict on network errors.
```

### DIFF
--- a/src/storage/txn/commands/prewrite.rs
+++ b/src/storage/txn/commands/prewrite.rs
@@ -853,7 +853,7 @@ mod tests {
         )
         .err()
         .unwrap();
-        assert_eq!(2, statistic.write.seek);
+        assert_eq!(3, statistic.write.seek);
         match e {
             Error(box ErrorInner::Mvcc(MvccError(box MvccErrorInner::KeyIsLocked(_)))) => (),
             _ => panic!("error type not match"),
@@ -866,7 +866,7 @@ mod tests {
             102,
         )
         .unwrap();
-        assert_eq!(2, statistic.write.seek);
+        assert_eq!(3, statistic.write.seek);
         let e = prewrite(
             &engine,
             &mut statistic,
--- a/src/storage/txn/commands/prewrite.rs
+++ b/src/storage/txn/commands/prewrite.rs
@@ -537,12 +537,11 @@ impl<K: PrewriteKind> Prewriter<K> {
                     txn.guards = Vec::new();
                     final_min_commit_ts = TimeStamp::zero();
                 }
-                e @ Err(MvccError(box MvccErrorInner::KeyIsLocked { .. })) => {
-                    locks.push(
-                        e.map(|_| ())
-                            .map_err(Error::from)
-                            .map_err(StorageError::from),
-                    );
+                Err(MvccError(box MvccErrorInner::KeyIsLocked { .. })) => {
+                    match check_committed_record_on_err(prewrite_result, txn, reader, &key) {
+                        Ok(res) => return Ok(res),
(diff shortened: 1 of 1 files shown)


src/storage/txn/commands/prewrite.rs
```diff
@@ -537,12 +537,11 @@ impl<K: PrewriteKind> Prewriter<K> {
...
-                            .map_err(Error::from)
-                            .map_err(StorageError::from),
-                    );
+                Err(MvccError(box MvccErrorInner::KeyIsLocked { .. })) => {
+                    match check_committed_record_on_err(prewrite_result, txn, reader, &key) {
+                        Ok(res) => return Ok(res),
+                        Err(e) => locks.push(Err(e.into())),
+                    }
                 }
                 Err(e @ MvccError(box MvccErrorInner::AssertionFailed { .. })) => {
```
cfzjywxk, line 546: @MyonKeminta @sticnarf 
Is it necessary to also cover the assertion check failures here for the possible retrying prewrite requests?

sticnarf: This also reminds me about the last branch, other errors. It can be anything strange, for example, disk IO errors. Furthermore, it can be still not idempotent and it may be even impossible to do `check_committed_record_on_err`. In this case, it may result in a false commit failure...

Maybe we only report failure to the user when the error is of known types. For unknown errors, we treat them as undetermined errors.

cfzjywxk: Agree, another thing that may need to be considered is how much overhead or impact would be introduced if we do more checks.

MyonKeminta: The only case where the assertion failure may be reported before any conflict is found is as stated in this issue: https://github.com/tikv/tikv/issues/12113 . It seems adding check here may avoid throwing assertion_failed error in cases where the current transaction is in committed state. However it looks not necessary to me, since we already have a fix to #12113 .

MyonKeminta: Will the idea about treating unknown errors as undetermined be done in this PR? By the way, do we have any injection for producing IO errors?

sticnarf: > Will the idea about treating unknown errors as undetermined be done in this PR? By the way, do we have any injection for producing IO errors?

I can't remember we inject any IO errors in regular tests (otherwise, I believe we should have been discovered something strange before).

Treating unknown errors as undetermined will not be included in _this_ PR. It should be done in the client part, I think. And that problem is not a new issue, it seems to exist from the very beginning. So, that problem may be not as severe and urgent as this one.

sticnarf:
@cfzjywxk @MyonKeminta
I think the PR itself is enough for now.
It seems not so necessary for me to care about the additional performance impact here, because it is not common and reasonable to run a real workload with _a lot of_ optimistic write conflicts.
Any other concerns?

youjiali1995 (approved):

MyonKeminta (approved):

Issue #12362, opened 2022-04-14
Performance Diagnosis Enhancements


sticnarf:
TiKV users and developers often care about performance, but it is never easy to diagnose performance issues or find performance bottlenecks.

- Latency of some parts in the critical path is not recorded, sometimes making the cause of performance issues only inferred indirectly.
- Metrics displayed in the Grafana dashboard are not very well organized. It's not easy to find the cause quickly.
- It's hard to investigate the cause of an individual slow request if the overall latency is normal. 

To improve the situation, we need to fill up the missing parts of the metrics, add more details to the metrics and try to improve the way we display the metrics. And, it's better to collect performance data in the request level and send back to the client if possible, enabling analyzing tail latency causes. 

## Data to be collected

Below are the missing data we are going to collect. "(R)" means the data may be collected for each request, and "(M)" means they are collected as metrics.

### Read

- [ ] (R) Whether `async_snapshot` is completed by the local reader
- [ ] (R) `async_snapshot` propose wait duration
- [ ] (R) `async_snapshot` read index wait duration
- [ ] (RM) RocksDB `GetSnapshot` duration
- [ ] (R) RocksDB seek count, total seek duration, max seek duration
- [ ] (R) RocksDB block read duration
- [ ] (RM) RocksDB DB mutex wait duration
- [ ] (RM) Read pool scheduling wait duration
- [ ] (RM) Read pool queue length
- [ ] (R) Priority in the read pool
- [ ] (RM) Semaphore wait duration
- [ ] (RM) Semaphore wait queue length
- [ ] (RM) Task yield count, total yield duration

### Write

- [ ] (RM) Latch wait queue length
- [ ] (R) Latch wait time
- [ ] (RM) Scheduler pool wait duration
- [ ] (RM) Scheduler pool wait queue length
- [ ] (R) Scheduler flow control throttle duration
- [ ] (R) Scheduler flow control throttle reason
- [ ] (R) Request `async_write` duration
- [ ] (R) Request propose wait duration
- [ ] (RM) Store loop queue length
- [ ] (RM) Store loop queue wait rounds
- [ ] (R) Store loop batch size
- [ ] (R) Store loop batch write size
- [ ] (R) Write Raft log duration (including RocksDB details) 
- [ ] (RM) Raft flow control duration
- [ ] (RM) Inter-TiKV RPC latency
- [ ] (R) Apply wait duration
- [ ] (R) Read entry cache duration (if happens)
- [ ] (RM) Apply loop queue length
[...]

PR #12373, opened 2022-04-19
engine_rocks: replace PerfLevel with PerfFlags in raftstore

sticnarf:

Issue Number: ref #12362

What's Changed:

```commit-message
PerfFlags provides more fine-grained PerfContext collecting. In this way, we
only collect the duration we really care about. Theoretically, it reduces the
overhead of collecting PerfContext compared to using PerfLevel::EnableTime.

This commit only changes the raftstore part. The read pool part will be done
later.

The default value of raftstore.perf-level is changed to Uninitialized. The
config is not documented before, so I think the change is not so important
to change the default value.
```

### Check List 

Tests 

- Manual test

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_rocks/src/perf_context_impl.rs
+++ b/components/engine_rocks/src/perf_context_impl.rs
@@ -3,10 +3,11 @@
 use crate::perf_context_metrics::{
     APPLY_PERF_CONTEXT_TIME_HISTOGRAM_STATIC, STORE_PERF_CONTEXT_TIME_HISTOGRAM_STATIC,
 };
-use crate::raw_util;
+use crate::{
+    raw_util, set_perf_flags, set_perf_level, PerfContext as RawPerfContext, PerfFlag, PerfFlags,
+};
 use engine_traits::{PerfContextKind, PerfLevel};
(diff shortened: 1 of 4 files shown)


tabokie (approved):

components/engine_rocks/src/perf_context_impl.rs
```diff
@@ -45,6 +46,18 @@ macro_rules! observe_perf_context_type {
     };
 }
 
+lazy_static! {
+    /// Default perf flags for a write operation.
+    static ref DEFAULT_WRITE_PERF_FLAGS: PerfFlags = PerfFlag::WriteWalTime
```
Connor1996, line 51: What's the difference with `EnableTime`

sticnarf: Not much.
One difference is that we don't record time for other operations like RocksDB read in the same thread. If necessary, we can choose what we need later.
And because we are going to record durations about reading in the read pool with PerfFlags. This change will make it more consistent in the whole workspace.

Connor1996 (approved):
LGTM

PR #12392, opened 2022-04-20
*: fix multiple unit test errors in local environments

### LINKED ISSUE #12390: Improve local test running experience
These are some problems I meet at my place when running tests locally. I intend to solve them for myself and hopefully other people who meet the same problems.

1. `server::service::diagnostics::sys::tests::test_load_info` cannot pass on new Linux kernels (4.19+). The OS may provide more IO load info than the assertion expects:
```
thread 'server::service::diagnostics::sys::tests::test_load_info' panicked at 'assertion failed: `(left == right)`
  left: `["read_io/s", "read_merges/s", "read_sectors/s", "read_ticks/s", "write_io/s", "write_merges/s", "write_sectors/s", "write_ticks/s", "in_flight/s", "io_ticks/s", "time_in_queue/s", "discard_io/s", "discard_merged/s", "discard_sectors/s", "discard_ticks/s"]`,
 right: `["read_io/s", "read_merges/s", "read_sectors/s", "read_ticks/s", "write_io/s", "write_merges/s", "write_sectors/s", "write_ticks/s", "in_flight/s", "io_ticks/s", "time_in_queue/s"]`', src/server/service/diagnostics/sys.rs:633:13
```

2. We need to specify the `testexport` feature explicitly in order to compile tests of TiKV, otherwise `Api::CLIENT_TAG` cannot be found:
```
error[E0599]: no associated item named `CLIENT_TAG` found for type parameter `Api` in the current scope
    --> src/storage/mod.rs:4476:31
     |
4476 |             api_version: Api::CLIENT_TAG,
     |                               ^^^^^^^^^^ associated item not found in `Api`
```
Because of this, VSCode users may fail to use the inline "Run Test" feature.

3. `plugin_registry` tests assumes the build directory is `./target/{debug, release}/deps`. It's possible that the working directory is not the cargo root, or the user changes the default target directory path.

sticnarf:

Issue Number: Close #12390

What's Changed:

```commit-message
1. Relax the IO load info assertion from "assert_eq" to assert "starts with".
So the test won't fail on Linux 4.19+ because of the extra fields.

2. Enable the `testexport` feature of `api_version` in the `dev-dependencies`
of `tikv`. Then, it is not necessary to specify the `testexport` feature manually
to compile tikv tests.

3.  For `plugin_registry` tests, search the dynamic library under the same
directory with the test binary. This avoids failures when people are running the
tests not under the cargo root or the default target directory path is changed. 
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/server/service/diagnostics/sys.rs
+++ b/src/server/service/diagnostics/sys.rs
@@ -630,22 +630,19 @@ mod tests {
                 .iter()
                 .map(|x| x.get_key())
                 .collect::<Vec<&str>>();
-            assert_eq!(
-                keys,
-                vec![
-                    "read_io/s",
-                    "read_merges/s",
-                    "read_sectors/s",
-                    "read_ticks/s",
-                    "write_io/s",
-                    "write_merges/s",
-                    "write_sectors/s",
-                    "write_ticks/s",
-                    "in_flight/s",
-                    "io_ticks/s",
-                    "time_in_queue/s",
-                ]
-            );
+            assert!(keys.starts_with(&[
+                "read_io/s",
+                "read_merges/s",
+                "read_sectors/s",
+                "read_ticks/s",
+                "write_io/s",
+                "write_merges/s",
+                "write_sectors/s",
+                "write_ticks/s",
+                "in_flight/s",
+                "io_ticks/s",
+                "time_in_queue/s",
+            ]));
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/server/service/diagnostics/sys.rs
```diff
@@ -630,22 +630,19 @@ mod tests {
...
-                    "write_io/s",
-                    "write_merges/s",
-                    "write_sectors/s",
-                    "write_ticks/s",
-                    "in_flight/s",
-                    "io_ticks/s",
-                    "time_in_queue/s",
-                ]
-            );
+            assert!(keys.starts_with(&[
```
BusyJay, line 633: How about also print the content of `keys`?

5kbpers (approved):

Issue #12394, opened 2022-04-20
Make naming follow rust convention


BusyJay:
## Development Task

There are a lot of places that name types not following rust convention. For example, `APIVersion`, `SSTImporter`, `DBVector`, etc.

Changing all of them can be a big changes. And some of them come from other crates. We may only fix the one defined in our codebase first. And all things related to rust-rocksdb should wait tirocks ready.

PR #12395, opened 2022-04-20
*: make several type names rusty

BusyJay:
Issue Number: Ref #12394

What's Changed:

```commit-message
Following types are renamed:
`APIVersion` -> `KvFormat`,
`SSTXxxx` -> `SstXxxx`,
`TTLXxxx` -> `TtlXxxx`,
`KVXxx` -> `KvXxxx`,
struct `Storage` -> `StorageHandler`,
trait `Engine` -> `Storage`.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/sst_importer/src/errors.rs
```diff
@@ -26,7 +26,7 @@ pub fn error_inc(type_: &str, err: &Error) {
         Error::ParseIntError(..) => "parse_int",
         Error::FileExists(..) => "file_exists",
         Error::FileCorrupted(..) => "file_corrupt",
-        Error::InvalidSSTPath(..) => "invalid_sst",
+        Error::InvalidSstPath(..) => "invalid_sst",
```
BusyJay, line 29: /cc @5kbpers Will it change the error code?

components/api_version/src/lib.rs
```diff
@@ -71,28 +71,47 @@ pub trait APIVersion: Clone + Copy + 'static + Send + Sync {
...
+pub struct ApiV1;
 #[derive(Default, Clone, Copy)]
-pub struct APIV1TTL;
+pub struct ApiV1Ttl;
 #[derive(Default, Clone, Copy)]
-pub struct APIV2;
+pub struct ApiV2;
+
+#[macro_export]
+macro_rules! check_kv_format {
```
pingyu, line 81: Suggest naming as something like `test_kv_format_impl`, means "test KvFormat implementations". `check_kv_format` would be confused with [`check_api_version`](https://github.com/tikv/tikv/blob/8b509a7415e137b3719b1a947d11dfbf20e3bbb5/src/storage/mod.rs#L409).

Rest LGTM~

Connor1996 (commented):
BTW, Would you like to rename `trait Engine` to something else? maybe `KvStorage`

components/sst_importer/src/errors.rs
5kbpers: https://github.com/tikv/tikv/blob/8b509a7415e137b3719b1a947d11dfbf20e3bbb5/components/error_code/src/sst_importer.rs#L14 Need to modify here and run `make error-code`

BusyJay:
> BTW, Would you like to rename `trait Engine` to something else? maybe `KvStorage`

There will be `Storage` from transaction layer, `KvStorage` from tikv_kv component, `KvEngine` from engine trait. I'm not sure if it's a good name.

@pingyu @cfzjywxk do you have suggestions?

Connor1996:
> > BTW, Would you like to rename `trait Engine` to something else? maybe `KvStorage`
> 
> There will be `Storage` from transaction layer, `KvStorage` from tikv_kv component, `KvEngine` from engine trait. I'm not sure if it's a good name.
> 
> @pingyu @cfzjywxk do you have suggestions?

`Storage` is not on transaction layer, it provides both `raw_get` and `txn_get`. I agree `KvStorage` is not a good enough name,  but have no better idea.

pingyu:
>

pingyu:
> > BTW, Would you like to rename `trait Engine` to something else? maybe `KvStorage`
> 
> There will be `Storage` from transaction layer, `KvStorage` from tikv_kv component, `KvEngine` from engine trait. I'm not sure if it's a good name.
> 
> @pingyu @cfzjywxk do you have suggestions?

How about "StorageEngine" ?
Then the dependent relationship would be `Storage` -> `StorageEngine` -> `KvEngine`.

(p.s. I'm sorry that I click the wrong button... )

Connor1996:
> > > BTW, Would you like to rename `trait Engine` to something else? maybe `KvStorage`
> > 
> > 
> > There will be `Storage` from transaction layer, `KvStorage` from tikv_kv component, `KvEngine` from engine trait. I'm not sure if it's a good name.
> > @pingyu @cfzjywxk do you have suggestions?
> 
> `Storage` is not on transaction layer, it provides both `raw_get` and `txn_get`. I agree `KvStorage` is not a good enough name, but have no better idea.

@BusyJay Another option is,  the `trait Engine` should be renamed to `trait Storage`, considering in grafana we already think `async_write` and `async_snapshot` are on the layer of `Storage`.

And rename `struct Storage` to something else, maybe `TiKVHandler`?

cfzjywxk:
> > > > BTW, Would you like to rename `trait Engine` to something else? maybe `KvStorage`
> > > 
> > > 
> > > There will be `Storage` from transaction layer, `KvStorage` from tikv_kv component, `KvEngine` from engine trait. I'm not sure if it's a good name.
> > > @pingyu @cfzjywxk do you have suggestions?
> > 
> > 
> > `Storage` is not on transaction layer, it provides both `raw_get` and `txn_get`. I agree `KvStorage` is not a good enough name, but have no better idea.
> 
> @BusyJay Another option is, the `trait Engine` should be renamed to `trait Storage`, considering in grafana we already think `async_write` and `async_snapshot` are on the layer of `Storage`.
> 
> And rename `struct Storage` to something else, maybe `TiKVHandler`?

This seems to be good. The current `struct storage` is the command handling entry, in MySQL, this module is named `handlerton`(https://dev.mysql.com/doc/internals/en/custom-engine.html and https://dev.mysql.com/doc/internals/en/api-reference.html).

The `storage` could mean the generalized storage which contains transaction and raw kv related modules, or it could mean the actual standalone storage engine. It should be used with caution to avoid misunderstanding.

BusyJay:
OK, then `Storage` is renamed to `StorageHandler` and trait `Engine` is renamed to `Stroage`. What do you think? @Connor1996 @pingyu @cfzjywxk

pingyu:
> OK, then `Storage` is renamed to `StorageHandler` and trait `Engine` is renamed to `Stroage`. What do you think? @Connor1996 @pingyu @cfzjywxk

LGTM~

cfzjywxk:
> OK, then `Storage` is renamed to `StorageHandler` and trait `Engine` is renamed to `Stroage`. What do you think? @Connor1996 @pingyu @cfzjywxk

LGTM

Connor1996:
> OK, then `Storage` is renamed to `StorageHandler` and trait `Engine` is renamed to `Stroage`. What do you think? @Connor1996 @pingyu @cfzjywxk

LGTM

components/tikv_kv/src/btree_engine.rs
```diff
@@ -69,7 +70,7 @@ impl Default for BTreeEngine {
     }
 }
 
-impl Engine for BTreeEngine {
+impl Storage for BTreeEngine {
```
Connor1996, line 73: rename these `XXEngine` to `XXStorage`

components/tikv_kv/src/lib.rs
```diff
@@ -1,8 +1,8 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
-//! There are multiple [`Engine`](kv::Engine) implementations, [`RaftKv`](crate::server::raftkv::RaftKv)
-//! is used by the [`Server`](crate::server::Server). The [`BTreeEngine`](kv::BTreeEngine) and
-//! [`RocksEngine`](RocksEngine) are used for testing only.
+//! There are multiple [`Storage`](kv::Storage) implementations, [`RaftKv`](crate::server::raftkv::RaftKv)
```
Connor1996, line 3: let's rename `RaftKv` to `RaftKvStorage` ?

components/tikv_kv/src/lib.rs
```diff
@@ -262,7 +262,7 @@ pub struct SnapContext<'a> {
 }
 
 /// Engine defines the common behaviour for a storage engine type.
-pub trait Engine: Send + Clone + 'static {
+pub trait Storage: Send + Clone + 'static {
     type Snap: Snapshot;
     type Local: LocalEngine;
```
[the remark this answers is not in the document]
BusyJay: I don't think it's appropriate. `Storage` itself can be distributed, and `Local` means the very engine that is used in the single node.

components/tikv_kv/src/lib.rs
Connor1996: And maybe rename the component from `tikv_kv` to `storage`, and rename `src/storage/kv` to `src/storage_handler/storage` ?

components/server/src/server.rs
```diff
@@ -197,7 +197,7 @@ struct TiKVServer<ER: RaftEngine> {
     quota_limiter: Arc<QuotaLimiter>,
 }
 
-struct TiKVEngines<EK: KvEngine, ER: RaftEngine> {
+struct TiKvEngines<EK: KvEngine, ER: RaftEngine> {
     engines: Engines<EK, ER>,
     store_meta: Arc<Mutex<StoreMeta>>,
     engine: RaftKv<EK, ServerRaftStoreRouter<EK, ER>>,
```
Connor1996, line 203:
```suggestion
    storage: RaftKvStorage<EK, ServerRaftStoreRouter<EK, ER>>,
```

components/tikv_kv/src/lib.rs
```diff
@@ -264,7 +262,7 @@ pub struct SnapContext<'a> {
 /// Engine defines the common behaviour for a storage engine type.
 pub trait Storage: Send + Clone + 'static {
     type Snap: Snapshot;
-    type Local: LocalEngine;
+    type Local: KvEngine;
```
Connor1996, line 265:
```suggestion
    type Engine: KvEngine;
```

components/tikv_kv/src/lib.rs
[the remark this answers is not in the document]
Connor1996: yes, so it shouldn't use `Local`

components/tikv_kv/src/lib.rs
BusyJay: There are a lot of changes and `tikv_kv` looks OK to me.

src/storage/mod.rs
```diff
@@ -2753,10 +2752,10 @@ impl FlowStatsReporter for DummyReporter {
     fn report_write_stats(&self, _write_stats: WriteStats) {}
 }
 
-impl<E: Engine, L: LockManager, Api: APIVersion> TestStorageBuilder<E, L, Api> {
+impl<E: Storage, L: LockManager, F: KvFormat> TestStorageBuilder<E, L, F> {
```
pingyu, line 2755: Suggest renaming to `TestStorageHandlerBuilder`.

components/backup/src/endpoint.rs
```diff
@@ -611,7 +611,7 @@ impl SoftLimitKeeper {
 /// The endpoint of backup.
 ///
 /// It coordinates backup tasks and dispatches them to different workers.
-pub struct Endpoint<E: Engine, R: RegionInfoProvider + Clone + 'static> {
+pub struct Endpoint<E: Storage, R: RegionInfoProvider + Clone + 'static> {
```
Connor1996, line 614:
```suggestion
pub struct Endpoint<S: Storage, R: RegionInfoProvider + Clone + 'static> {
```
replace all `E: Storage` with `S: Storage`

components/tikv_kv/src/lib.rs
```diff
@@ -565,7 +563,7 @@ pub unsafe fn destroy_tls_engine<E: Engine>() {
 }
 
 /// Get a snapshot of `engine`.
-pub fn snapshot<E: Engine>(
+pub fn snapshot<E: Storage>(
     engine: &E,
```
Connor1996, line 567:
```suggestion
    storage: &S,
```
same for other places

BusyJay: I'm afraid this may make most cherry picks after this PR conflict with the pass.

BusyJay:
I find that renaming `Engine` to `Storage` may break a lot of things, not just codes but also docs and comments. It may also make future cherry-pick harder. I propose to just keep the existing naming in this PR.

pingyu:
> I find that renaming `Engine` to `Storage` may break a lot of things, not just codes but also docs and comments. It may also make future cherry-pick harder. I propose to just keep the existing naming in this PR.

LGTM, and very thanks for your work ~
(_I suggest to leave a comment to inform others that, don't try to rename `Engine`_ 🐶 )

BusyJay:
> LGTM, and very thanks for your work ~ (_I suggest to leave a comment to inform others that, don't try to rename `Engine`_ 🐶 )

Actually, I think it's acceptable to me to rename it. However, because I'm not the one that do heavy cherry picks for transaction and storage change, so my opinion may not be significant. To do the refactor, I think community should get involved and get a consensus instead of just a small group of us. And it should be finished in a PR that take good care of all comments and docs.

Connor1996:
> > LGTM, and very thanks for your work ~ (_I suggest to leave a comment to inform others that, don't try to rename `Engine`_ 🐶 )
> 
> Actually, I think it's acceptable to me to rename it. However, because I'm not the one that do heavy cherry picks for transaction and storage change, so my opinion may not be significant. To do the refactor, I think community should get involved and get a consensus instead of just a small group of us. And it should be finished in a PR that take good care of all comments and docs.

The doc including these terms should be only source code readings, but they all are specified to a fixed version. And I believe cherry-pick won't suffer too much from these. The following version is 6.1 LTS, it's a good opportunity to say goodbye to these burdens. If we don't take the first step, it's even harder to change it later. 

So I suggest let's merge it now without further subtle naming changes, and the rest subtle mismatch naming doesn't affect much and they will be iterated over time.

Connor1996 (approved):
LGTM

pingyu:
> > LGTM, and very thanks for your work ~ (_I suggest to leave a comment to inform others that, don't try to rename `Engine`_ 🐶 )
> 
> Actually, I think it's acceptable to me to rename it. However, because I'm not the one that do heavy cherry picks for transaction and storage change, so my opinion may not be significant. To do the refactor, I think community should get involved and get a consensus instead of just a small group of us. And it should be finished in a PR that take good care of all comments and docs.

I agree. It would be better to have a more widespread discussion, if the change will affect most people of the community.

BusyJay:
@Connor1996 I reverted the `Engine` change. If you still think it's necessary, I suggest you to send a proposal to internals.tidb.io and see the feed back first.

PR #12400, opened 2022-04-21
cmd: make lock files accessible to all users.

### LINKED ISSUE #12399: File locks for working around port conflict are sometimes inaccessible
master

unix

### Steps to reproduce
1. run tikv-server in root mode
2. run tikv-server in user mode

### What did you expect?
nornal startup

### What did happened?
`[FATAL] [[server.rs:894](http://server.rs:894/)] ["failed to create lock at /tmp/1001_TIKV_LOCK_FILES/0.0.0.0_20160: Permission denied (os error 13)"]`

hunterlxt:

Issue Number: Close #12399

What's Changed:

```commit-message
Make lock files accessible to all users.
```

### Related changes

- None

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Fix the bug where sometimes tikv-server would not start because it didn't have access to lock files.
```

### DIFF
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -380,6 +380,7 @@ impl<ER: RaftEngine> TiKVServer<ER> {
         let search_base = env::temp_dir().join(&lock_dir);
         file_system::create_dir_all(&search_base)
             .unwrap_or_else(|_| panic!("create {} failed", search_base.display()));
+        set_path_all_permission(&search_base);
 
         for entry in file_system::read_dir(&search_base).unwrap().flatten() {
             if !entry.file_type().unwrap().is_file() {
@@ -1485,6 +1486,8 @@ fn try_lock_conflict_addr<P: AsRef<Path>>(path: P) -> File {
         )
     });
 
+    set_path_all_permission(&path);
+
     if f.try_lock_exclusive().is_err() {
         fatal!(
             "{} already in use, maybe another instance is binding with this address.",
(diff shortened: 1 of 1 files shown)


components/server/src/server.rs
```diff
@@ -1504,6 +1507,18 @@ fn get_lock_dir() -> String {
     "TIKV_LOCK_FILES".to_owned()
 }
 
+// Make the lock files be accessed by all users. Only support in unix platform.
+#[cfg(unix)]
+fn set_path_all_permission<P: AsRef<Path>>(path: P) {
+    use std::fs::{set_permissions, Permissions};
+    use std::os::unix::fs::PermissionsExt;
+    // this may failed when setting a file created by root.
+    let _ = set_permissions(path, Permissions::from_mode(0o777));
```
5kbpers, line 1516: Should we log this error?

5kbpers (approved):
rest LGTM

components/server/src/server.rs
hunterlxt: If tikv-server started by root, it will return error when starting by normal user. I don't think this behavior is a error?

sticnarf (approved):
LGTM

components/server/src/server.rs
5kbpers: Got it.

PR #12411, opened 2022-04-21
*: set health status to unknown when raftstore gets stuck

### LINKED ISSUE #12398: Make health service aware of raftstore unavailability
Now, the health service only checks the network connectivity between the TiKV client and server. The client uses this to determine whether it should still send requests to this TiKV or whether it should refresh related region cache.

But if the raftstore **alone** becomes unavailable because of IO hang or bugs, the health service still returns `Serving` status. This may mislead the TiKV client and increase the recover time even if the leader is already transferred to other TiKV instances.

Therefore, I propose we should update the serving status if we detect the raftstore has stopped working.

sticnarf:

Issue Number: Close #12398

What's Changed:

```commit-message
The client uses health service to determine whether it should
still send requests to this TiKV or whether it should refresh
related region cache.

But if the raftstore alone becomes unavailable because of IO hang
or bugs, the health service still returns Serving status. This may
mislead the TiKV client and increase the recover time even if the
leader is already transferred to other TiKV instances.

This commit reuses the mechanism of slow score calculation
to detect whether the raftstore is normally working. So, the client
can refresh its region cache in time.
```

### Check List 

Tests 

- Integration test

### Release note 

```release-note
Report bad health status if raftstore stops working.
```


sticnarf:
I'm not sure whether this is a good approach. If there are better ways to do this, I'm happy to change it.

BusyJay:
When will the slow score be recalculated? If the store stuck for 5 seconds and then recover, how can client notice the recovery in time?

sticnarf:
> When will the slow score be recalculated? If the store stuck for 5 seconds and then recover, how can client notice the recovery in time?

The slow score calculation is driven by PD worker. It sends `StoreMsg::LatencyInspect` at `inspect_interval` (500ms by default), and the slow score is updated every 30 ticks. So, it means, the health status will recover in 15 seconds.

As regards client-go, if it finds a store is not healthy, it will recheck its status every 1 second.

BusyJay:
So in this case, use health status will cause regression, right?

sticnarf:
> So in this case, use health status will cause regression, right?

Personally, I think it an accetable regression. The update interval is 15 seconds, so the health status does not change until the raftstore keeps stuck for more than 15 seconds. And if there is such an extreme situation where it has been stuck for 15 seconds, it seems not a big problem to wait for another 15 seconds to get recovered.

BusyJay:
How about if UNKNOWN is returned, schedule the requests at random fashion?

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1918,6 +1925,17 @@ where
...
+            // `last_update_time` is refreshed every round. We use it to maintain
+            // the health status of the TiKV.
+            let health_status =
+                if self.slow_score.last_record_time < self.slow_score.last_update_time {
+                    ServingStatus::ServiceUnknown
+                } else {
+                    ServingStatus::Serving
+                };
+            if let Some(health_service) = &self.health_service {
+                health_service.set_serving_status("", health_status);
```
5kbpers, line 1937: How about updating the health status according to the slow score or the ratio of timeout records? For instance, if the ratio of timeout records exceeds `ratio_thresh`, then set the status as `ServiceUnknown` as the same as the way we update the slow score.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1918,6 +1925,17 @@ where
         self.slow_score.last_tick_id += 1;
         self.slow_score.last_tick_finished = false;
         if self.slow_score.last_tick_id % self.slow_score.round_ticks == 0 {
+            // `last_update_time` is refreshed every round. We use it to maintain
+            // the health status of the TiKV.
+            let health_status =
+                if self.slow_score.last_record_time < self.slow_score.last_update_time {
```
5kbpers, line 1931: It seems to be not necessary since here is at the end of every round.

sticnarf:
> > , if the liveness of a store is not reachable, it will not filter out the store. If request forwarding is enabled, it will try forwarding
> 
> This seems wrong to me. The new behavior may cause one slow stores slow down other nodes as forwarding is not cost free.

Now, I change the unknown status to become "strict enter, easy exit".

The status is changed to unknown only if no `StoreMsg` is not handled within 15 seconds (in this case, the leader should have dropped). And as long as one tick does not timeout, the status will change back to serving.

components/raftstore/src/store/worker/pd.rs
sticnarf: I'm inclined to be conservative. This PR only intends to solve the problem when the raftstore is totally stuck instead of just being slow.

Because being slow does not mean the leader will certainly transfer, setting the status to unknown too easily will cause unexpected request forwarding.

And if the leader transfers due to just slowness, the TiKV can still return `NotLeader` to the client. It does not need the involvement of the health status.

BusyJay (approved):

5kbpers (approved):

PR #12415, opened 2022-04-22
raftstore: make snapshot gc in background thread

### LINKED ISSUE #11966: Snapshot gc needs background cleanup to eliminate time loss to store loop

[root@CentOS76_VM log]# ../bin/tikv-server --version
TiKV 
Release Version:   5.0.4
Edition:           Community
Git Commit Hash:   dbb3d3a7ea092bb3c7442c54bed19987caa26bd1
Git Commit Branch: release-5.0-20211105
UTC Build Time:    2022-01-27 10:13:25
Rust Version:      rustc 1.51.0-nightly (bc39d4d9c 2021-01-15)
Profile:           release

### What did you expect?
1. Snapshot gc needs  to be moved from the store loop thread to backgrounds.
2. All non-essential I/O should be removed from the Store Loop thread.

### What did happened?
1. When there are too many snapshot files, the store loop thread will be blocked by snap gc, causing the client request to time out. 
2. Currently, only Store Loop executes the Snap GC time log, and there is no precise time-consuming timing data. For the time being, it is considered that only Snap GC is being performed during this time period. The timing log will be added later.
FyxffzNXqT
Yk6ZQREUNC
blsmc6OaEl

tonyxuqqi:

Issue Number: Close #11966

What's Changed:
```
raftstore: make snapshot gc in background thread to unblock raftstore message loop.
```

### Check List 

Tests 

- No code

### Release note 
```release-note
None
```


components/raftstore/src/store/worker/cleanup_snapshot.rs
```diff
@@ -0,0 +1,176 @@
+// Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
```
Connor1996, line 1: 2022

components/raftstore/src/store/worker/cleanup_snapshot.rs
```diff
@@ -0,0 +1,176 @@
...
+
+pub struct Runner<EK, ER>
+where
+    EK: KvEngine,
+    ER: RaftEngine,
+{
+    store_id: u64,
+    router: RaftRouter<EK, ER>,
+    snap_mgr: SnapManager,
+    _engine: PhantomData<EK>,
```
Connor1996, line 43: why need this phantom

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1065,6 +1066,24 @@ where
     }
 
     fn on_gc_snap(&mut self, snaps: Vec<(SnapKey, bool)>) {
+        macro_rules! schedule_delete_snapshot_files {
```
Connor1996, line 1069: schedule_delete_snapshot_files = |key, snap| { ... }

components/raftstore/src/store/worker/cleanup_snapshot.rs
```diff
@@ -0,0 +1,176 @@
...
+        let (mut last_region_id, mut keys) = (0, vec![]);
+        let schedule_gc_snap = |region_id: u64, snaps| -> Result<()> {
+            debug!(
+                "schedule snap gc";
+                "store_id" => self.store_id,
+                "region_id" => region_id,
+            );
+
+            let gc_snap = PeerMsg::CasualMessage(CasualMessage::GcSnap { snaps });
+            match Deref::deref(&self.router).send(region_id, gc_snap) {
```
Connor1996, line 76: `self.router.send()` is not workable?

tonyxuqqi: It will have following error: ---the problem is that gc_snap is PeerMsg, not CasualMessage (so I cannot use CasualRouter). So it seems I have to call its defer API.  But looks (*self.router).send works fine. 

error[E0061]: this function takes 1 argument but 2 arguments were supplied
  --> components/raftstore/src/store/worker/cleanup_snapshot.rs:76:31
   |
76 |             match self.router.send(region_id, gc_snap) {
   |                               ^^^^ ---------  ------- supplied 2 arguments
   |                               |
   |                               expected 1 argument

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: ok. closure seems better than macro.

components/raftstore/src/store/worker/cleanup_snapshot.rs
tonyxuqqi: good catch.

components/raftstore/src/store/worker/cleanup_snapshot.rs
BusyJay: You mean `CasualRouter::send(&self.router, region_id, gc_snap)`?

tonyxuqqi: > 
Not CasualRouter because the msg here is PeerMessage, not CasualMessage. I need to call BatchRouter's send, which is done by deref self.router.

tonyxuqqi: The reason that it cannot use CasualRouter::send(&self, region_id: u64, msg: CasualMessage<EK>)  is because it's error processing does not meet the need and we have to use the underlying router's send, which only accept PeerMsg as message type.

components/raftstore/src/store/worker/cleanup_snapshot.rs
BusyJay: It's not necessary.

components/raftstore/src/store/worker/cleanup_snapshot.rs
```diff
@@ -0,0 +1,176 @@
...
+                }
+            }
+            Task::DeleteSnapshotFiles {
+                key,
+                mut snapshot,
+                check_entry,
+            } => {
+                let snapshot = snapshot.as_mut();
+                if let Err(e) = snapshot.load_snapshot_meta_if_necessary() {
+                    warn!(
```
BusyJay, line 163: Why still try to delete it when it fails to load meta?

tonyxuqqi: In case the meta file is corrupted, we still need to delete the ssts in the best effort. 
The delete_snapshot() will try to find possible sst candidates files to delete.

components/raftstore/src/store/worker/cleanup_snapshot.rs
tonyxuqqi: ok

BusyJay (approved):

Connor1996 (approved):
LGTM

PR #12419, opened 2022-04-22
split_controller: fix the bug that may cause the stats_monitor to fall into a dead loop

### LINKED ISSUE #12416: TiKV Short-term memory usage continues to rise

v6.0.0
2b4f7d57af9ca3e21bc07b91a9cfc466ed87734f

### What did happened?

image
image

Suspect the same problem as this one：https://github.com/tikv/tikv/issues/12229

JmPotato:

Issue Number: Close #12416

What's Changed:

```commit-message
split_controller: fix the bug that may cause the stats_monitor to fall into a dead loop
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Fix a bug that causes the TiKV thread to fall into a dead loop and then lead to OOM.
```

### DIFF
--- a/components/raftstore/src/store/worker/split_controller.rs
+++ b/components/raftstore/src/store/worker/split_controller.rs
@@ -1,6 +1,6 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::cmp::Ordering;
+use std::cmp::{min, Ordering};
 use std::collections::BinaryHeap;
 use std::collections::HashMap;
 use std::collections::HashSet;
@@ -15,13 +15,14 @@ use kvproto::metapb::{self, Peer};
 use kvproto::pdpb::QueryKind;
 use pd_client::{merge_bucket_stats, new_bucket_stats, BucketMeta, BucketStat};
 use tikv_util::config::Tracker;
-use tikv_util::{debug, info};
+use tikv_util::{debug, info, warn};
 
 use crate::store::metrics::*;
 use crate::store::worker::query_stats::{is_read_query, QueryStats};
 use crate::store::worker::split_config::get_sample_num;
 use crate::store::worker::{FlowStatistics, SplitConfig, SplitConfigManager};
 
+const DEFAULT_MAX_SAMPLE_LOOP_COUNT: usize = 10000;
 pub const TOP_N: usize = 10;
 
 // LOAD_BASE_SPLIT_EVENT metrics label definitions.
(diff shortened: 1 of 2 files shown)


lhy1024 (commented):
LGTM. But do we need to revert max_retry to avoid other similar problems?

nolouch:
cc @hundundm

nolouch (commented):
Can we try to add some fuzz test for the function?

JmPotato:
> Can we try to add some fuzz test for the function?

Yes, I'm working on it.

JmPotato:
> LGTM. But do we need to revert max_retry to avoid other similar problems?

After this PR, I believe there is no need to add this mechanism.

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -81,11 +81,15 @@ where
     F: Fn(&mut T) -> &mut Vec<KeyRange>,
 {
     let mut sampled_key_ranges = vec![];
+    // Retain the non-empty key ranges.
+    // `key_ranges_provider` may return an empty key ranges vector, which will cause
+    // the later sampling to fall into a dead loop. So we need to filter it out here.
+    key_ranges_providers
+        .retain_mut(|key_ranges_provider| !key_ranges_getter(key_ranges_provider).is_empty());
```
AndreMouche, line 88: So it seems the key_ranges_providers won't too big so we run the function `key_ranges_getter` directly on item? If yes, Could we store the result of `key_range_getter(key_ranges_provider)`  and use it directly on L122?

Or If the `key_ranges_providers` maybe very big, we could consider to control  `while` loop instead?

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -905,33 +933,54 @@ mod tests {
...
+                let all_key_ranges_num = *prefix_sum(key_ranges.iter(), Vec::len).last().unwrap();
+                assert_eq!(
+                    sampled_key_ranges.len(),
+                    std::cmp::min(sample_num, all_key_ranges_num)
+                );
+            }
+        }
+    }
+
+    fn check_sample_length_with_timeout(key_ranges: Vec<Vec<KeyRange>>, timeout: Duration) {
```
lhy1024, line 949: if we revert DEFAULT_MAX_SAMPLE_LOOP_COUNT, do we still need it?

nolouch (commented):
LGTM

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -121,6 +132,12 @@ where
             sampled_key_ranges.push(key_ranges.remove(j)); // Sampling without replacement
         }
     }
+    if sample_loop_count == 0 {
+        warn!("The number of sampled key ranges could be less than the sample_num, the sampling may fall into a dead loop before";
```
nolouch, line 136:
```suggestion
        warn!("the number of sampled key ranges could be less than the sample_num, the sampling may fall into a dead loop before";
```

nolouch (approved):

lhy1024 (approved):

AndreMouche (approved):
LGTM

PR #12426, opened 2022-04-24
cdc: add resolved ts for rawkv

zeminzhou:

Issue Number: ref #11965

What's Changed:
1. Add new resovled ts logic for rawkv cdc;

### Check List 

Tests 

- Unit test
- Integration test

```release-note
add resolved ts for rawkv cdc
```


components/cdc/src/endpoint.rs
```diff
@@ -347,10 +374,14 @@ pub struct Endpoint<T, E> {
     region_read_progress: RegionReadProgressRegistry,
 
     // Metrics and logging.
-    min_resolved_ts: TimeStamp,
-    min_ts_region_id: u64,
-    resolved_region_count: usize,
-    unresolved_region_count: usize,
+    txnkv_min_resolved_ts: TimeStamp,
```
pingyu, line 377: Maybe we can use the "keyspace" to handle metrics for different mode. Every metric would be hold in a map, key by the "keyspace".

components/cdc/src/ts_tracker.rs
```diff
@@ -0,0 +1,133 @@
...
+        self.subscribed_regions
+            .read()
+            .unwrap()
+            .get(&region_id)
+            .is_some()
+    }
+}
+
+impl TsTracker for CdcTsTracker {
+    fn track_ts(&self, region_id: u64, ts: TimeStamp) {
```
pingyu, line 47: Can we just let `cdc::Endpoint` implement `track_ts` ? Then we can reuse the region subscriptions of `cdc::Endpoint`.

pingyu:
This PR is re-implemented in https://github.com/tikv/tikv/pull/12866.

PR #12432, opened 2022-04-26
*: apply perf flags to record RocksDB perf context in coprocessor

sticnarf:

Issue Number: Ref #12362

What's Changed:

```commit-message
This commit enables some more perf flags about reading, such as
block read time, DB mutex wait time. So, it enables us to find out
the root cause of more performance issues.

It also includes some refactorings of unifying the functions of capturing
PerfContext in tikv_kv into engine_rocks because engine specific functions
should be hidden in engine_* as much as possible. This is the first step that
removes statistics from tikv_kv, and later code referencing directly to the
engine_rocks should be also removed.
```

### Check List 

Tests 

- Existing tests

Side effects

- Performance regression
    - Consumes more CPU

I tried TPC-H and sysbench select_random_ranges and find no obvious regression.

### Release note 

```release-note
None
```


you06:
What do you think of setting the read stats in the coprocessor response's execution details?

sticnarf:
> What do you think of setting the read stats in the coprocessor response's execution details?

Yes, they should be returned, but I'm not going to include them in this PR.

We are going to add a few more execution details, not only about coprocessor read, but also about write operations. I plan to think more about the protocol before including the statistics in the response.

you06 (approved):

components/engine_rocks/src/perf_context_impl.rs
```diff
@@ -102,25 +196,207 @@ impl PerfContextStatistics {
...
-                report_perf_context!(self, STORE_PERF_CONTEXT_TIME_HISTOGRAM_STATIC);
+                report_write_perf_context!(self, STORE_PERF_CONTEXT_TIME_HISTOGRAM_STATIC);
             }
+            PerfContextKind::GenericRead => {
+                // TODO: Currently, metrics about reading is reported in other ways.
+                // It is better to unify how to report the perf metrics.
+                //
+                // Here we only record the PerfContext data into the fields.
+                self.read = ReadPerfContext::capture();
+            }
```
zhongzc, line 216: make it generics?

sticnarf: Sorry, what do you want it to be generic about? This is under `engine_rocks`, so it needn't be generic over engine.

zhongzc: I hadn't thought clearly. Never mind...

So we will report these statistics as metrics in the future?

zhongzc: And those statistics will also return as coprocessor response...

I hope that there will be an explicit way to remind people of their multiple purposes.

sticnarf: We already do. Now, it's in the coprocessor mod where we report these statistics.
But I think we can report these metrics like raftstore statistics here as well. So,  it will look more consistent about reporting rocksdb statistics.

sticnarf: What about changing `report_metrics` to `stop_observe` and add another trait function to `PerfContext` to get the inner data?

zhongzc (approved):

PR #12437, opened 2022-04-26
storage: adjust num_shard_bits by capacity size

### LINKED ISSUE #12436: small shared cache size (such as 1GB)  with big region size could lead to extreme slow TPC-C prepare

v6.0

centos 7
### Steps to reproduce

set storage.block-cache.capacity with 1GB
set coprocessor.region-split-size with 10GB
set coprocessor.region-max-size with 15GB

run tpc-c 1500 warehouse prepare
### What did you expect?
it can be finished within 8 hours or less.
### What did happened?
it cannot finish in 24 hours and the qps is down to single digit. The scheduler worker CPU is high and filter insert throughput is very high. 
image
image

The reason is that under small block capacity and fix shard bits  6(the shard bits are not public configuration), each shard is only 16MB, which is not enough to hold filter block data when region is large---as the compaction guard could result in big SST file size under big region for big level compaction.  When one shard size is not enough to hold filter block data, it will try to load the filter block into cache repeatedly which consumes lots of CPU by calculating checksum.

tonyxuqqi:
Close #12436

Issue Number: Close #12436

What's Changed:

```
Adjust num_shard_bits by capacity size to ensure each shard meet the minimal size of 64MB. Too small shard size could lead to extreme slow insert performance under large region size. 
```

### Check List 

Tests 

- Unit test

Side effects

- None
### Release note 

```release-note
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -72,6 +72,8 @@ pub const DEFAULT_ROCKSDB_SUB_DIR: &str = "db";
 pub const BLOCK_CACHE_RATE: f64 = 0.45;
 /// By default, TiKV will try to limit memory usage to 75% of system memory.
 pub const MEMORY_USAGE_LIMIT_RATE: f64 = 0.75;
+/// Min block cache shard's size. If a shard is too small, the index/filter data may not fit one shard
+pub const MIN_BLOCK_CACHE_SHARD_SIZE: usize = 64 * MIB as usize;
 
 const LOCKCF_MIN_MEM: usize = 256 * MIB as usize;
 const LOCKCF_MAX_MEM: usize = GIB as usize;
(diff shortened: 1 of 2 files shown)


BusyJay (approved):

src/config.rs
```diff
@@ -72,6 +72,8 @@ pub const DEFAULT_ROCKSDB_SUB_DIR: &str = "db";
 pub const BLOCK_CACHE_RATE: f64 = 0.45;
 /// By default, TiKV will try to limit memory usage to 75% of system memory.
 pub const MEMORY_USAGE_LIMIT_RATE: f64 = 0.75;
+/// Min block cache shard's size. If a shard is too small, the index/filter data may not fit one shard
+pub const MIN_BLOCK_CACHE_SHARD_SIZE: usize = 128 * MIB as usize;
```
zhangjinpeng1987, line 76: maybe 128MB is still not enough?

PR #12442, opened 2022-04-28
*: fix thread name truncating issue

### LINKED ISSUE #12451: Thread name will be truncated if too long
## Development Task
As described in https://github.com/tikv/tikv/pull/5593 and https://github.com/tikv/tikv/pull/10004, we could not identify different threads in metrics by name when their names are truncated due to the thread name length limitation in Linux.

GanZiheng:

Issue Number: Close #12451

What's Changed:

```commit-message
Based on #5593, add a global hashmap recording relationship between
thread id and thread name; add some wrappers to spawn threads and
update the hashmap.

It's necessary to use `after_start_wrapper` and `before_stop_wrapper`
together. Otherwise it may cause reporting a wrong thread name if
a thread inserts its name to hashmap and doesn't remove it, while
another thread reuses the same tid and doesn't update the hashmap.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None.
```


Connor1996 (commented):
Please test it with real cluster to see if the thread name shows well in grafana

GanZiheng:
> Please test it with real cluster to see if the thread name shows well in grafana

thread_name

Seems like it works as wish.

skyzh (commented):
It seems that we now have a lot of functions to deny. I would suggest:

* Using clippy to deny functions https://github.com/tikv/tikv/pull/12206 (but seems that there're still bugs)
* Another example https://github.com/MaterializeInc/materialize/blob/main/clippy.toml

The check script is easy to be circumvented. Checking it on compiler level might be a better choice.

GanZiheng:
> It seems that we now have a lot of functions to deny. I would suggest:
> 
> * Using clippy to deny functions #12206 (but seems that there're still bugs)
> * Another example https://github.com/MaterializeInc/materialize/blob/main/clippy.toml
> 
> The check script is easy to be circumvented. Checking it on compiler level might be a better choice.

Thanks for your advice, I will take a shot.

GanZiheng:
> > It seems that we now have a lot of functions to deny. I would suggest:
> > 
> > * Using clippy to deny functions [security: use clippy to disallow hex_encode #12206](https://github.com/tikv/tikv/pull/12206) (but seems that there're still bugs)
> > * Another example https://github.com/MaterializeInc/materialize/blob/main/clippy.toml
> > 
> > The check script is easy to be circumvented. Checking it on compiler level might be a better choice.
> 
> Thanks for your advice, I will take a shot.

It looks like that clippy could not deny methods written in our local crates. https://github.com/rust-lang/rust-clippy/issues/7479 reported the same issue.

components/backup/src/utils.rs
```diff
@@ -90,11 +90,11 @@ pub fn create_tokio_runtime(thread_count: usize, thread_name: &str) -> TokioResu
         .thread_name(thread_name)
         .enable_io()
         .enable_time()
-        .on_thread_start(|| {
+        .after_start_wrapper(|| {
```
Connor1996, line 93: Let's use a warper to wrap around these builder, and override the definition of the `after_start` and `before_start`

Connor1996: and then utilize `disallowed_types` to avoid using the original builders

GanZiheng: I tried this and found a problem with it.
If we call a method not overridden, we could get the original builder inside the wrapper, and then we are unable to call the method which is overridden.

tonyxuqqi, line 93: minor: IMO, on_thread_start_wrapper/on_thread_stop_wrapper seem a better naming.

components/test_raftstore/src/server.rs
```diff
@@ -429,6 +430,8 @@ impl ServerCluster {
             TokioBuilder::new_multi_thread()
                 .thread_name(thd_name!("debugger"))
                 .worker_threads(1)
+                .after_start_wrapper(|| {})
+                .before_stop_wrapper(|| {})
```
tonyxuqqi, line 434: For this specific case, I'm wondering if we can simplify the code here by merge these two line into one, such as
.on_start_stop_wrapper().

components/backup/src/utils.rs
GanZiheng: `after_start_wrapper` and `before_stop_wrapper` comes from `ThreadBuildWrapper` trait. I implement this trait for different kinds of thread pool builder, so their wrapper function names are same.

components/file_system/src/io_stats/mod.rs
```diff
@@ -54,7 +55,7 @@ mod tests {
         let _ths = (0..8)
             .map(|_| {
                 let tx_clone = tx.clone();
-                std::thread::Builder::new().spawn(move || {
+                std::thread::spawn(move || {
```
Connor1996, line 58: should we disallow `std::thread::spawn`?

scripts/clippy
```diff
@@ -33,6 +33,7 @@ CLIPPY_LINTS=(-A clippy::module_inception  \
 	-A clippy::enum_variant_names \
         -W clippy::dbg_macro \
         -W clippy::todo \
+				-D clippy::disallowed-methods \
```
Connor1996, line 36: format ident

components/file_system/src/io_stats/mod.rs
GanZiheng: I think it is not necessary. Since this pr actually adds methods to set hooks for thread pool builder and we should deny regular set-hook methods only.

Connor1996 (approved):
LGTM

Connor1996:
PTAL @BusyJay

clippy.toml
```diff
@@ -0,0 +1,12 @@
+disallowed-methods = [
+    { path = "std::thread::Builder::spawn", reason = "Wrapper function should be used instead, refer to https://github.com/tikv/tikv/pull/12442 for more details." },
```
BusyJay, line 2: Better explain it directly. And should mention what `wrapper` is needed exactly.

components/tikv_util/src/metrics/mod.rs
```diff
@@ -113,3 +113,21 @@ pub fn convert_record_pairs(m: HashMap<String, u64>) -> RecordPairVec {
         })
         .collect()
 }
+
+pub trait StdThreadBuildWrapper {
```
BusyJay, line 117: This is not an appropriate place to define the trait. Better put it in sys/thread.

components/tikv_util/src/metrics/threads_dummy.rs
```diff
@@ -42,3 +49,81 @@ impl Default for ThreadInfoStatistics {
...
+            f();
+        })
+    }
+
+    fn before_stop_wrapper<F>(&mut self, f: F) -> &mut Self
+    where
+        F: Fn() + Send + Sync + 'static,
+    {
+        // #[allow(clippy::disallowed_methods)]
+        self.before_stop(move || {
```
BusyJay, line 81: Why not `self.before_stop(f)`?

components/tikv_util/src/metrics/threads_linux.rs
```diff
@@ -466,11 +474,123 @@ impl TidRetriever {
...
+        self.spawn(|| {
+            add_thread_name_to_map();
+            let res = f();
+            remove_thread_name_from_map();
+            res
+        })
+    }
+}
+
+impl<T: PoolTicker> ThreadBuildWrapper for YatpPoolBuilder<T> {
```
BusyJay, line 514: Even the `Pool` is defined by the `tikv_util`, why bother to add an extra layer? For example, why not just put the hooks to runner directly?

GanZiheng: I just modify the `after_start` and `before_stop` of `YatpPoolBuilder`, instead of implementing trait for it.

components/tikv_util/src/yatp_pool/mod.rs
```diff
@@ -231,15 +231,21 @@ impl<T: PoolTicker> YatpPoolBuilder<T> {
     where
         F: Fn() + Send + Sync + 'static,
     {
-        self.before_stop = Some(Arc::new(f));
+        self.before_stop = Some(Arc::new(move || {
```
BusyJay, line 234: This is unnecessary. It should be added just like `add_thread_memory_accessor`.

src/storage/txn/sched_pool.rs
```diff
@@ -69,8 +71,8 @@ impl SchedPool {
         let pool = YatpPoolBuilder::new(SchedTicker {reporter:reporter.clone()})
             .thread_count(1, pool_size, max_pool_size)
             .name_prefix(name_prefix)
-            // Safety: by setting `after_start` and `before_stop`, `FuturePool` ensures
-            // the tls_engine invariants.
+            // Safety: by setting `after_start_wrapper` and `before_stop_wrapper`,
```
BusyJay, line 74: Stale comment.

components/tikv_util/src/yatp_pool/mod.rs
GanZiheng: You mean we should not change `after_start` and `before_stop` of `YatpPoolBuilder`, but call them with `add_thread_name_to_map` and `remove_thread_name_from_map` every time?

BusyJay: I mean calling `add_thread_name_to_map` in the same function which calls `add_thread_memory_accessor`.

GanZiheng: I put them in the runner of yatp, please take a look again : ).

components/tikv_util/src/yatp_pool/mod.rs
```diff
@@ -107,7 +107,8 @@ impl<T: PoolTicker> Runner for YatpPoolRunner<T> {
         if let Some(f) = self.after_start.take() {
             f();
         }
-        tikv_alloc::add_thread_memory_accessor()
+        tikv_alloc::add_thread_memory_accessor();
+        crate::sys::thread::add_thread_name_to_map()
```
BusyJay, line 111: Better move it to the beginning of function.

components/tikv_util/src/yatp_pool/mod.rs
```diff
@@ -138,6 +139,7 @@ impl<T: PoolTicker> Runner for YatpPoolRunner<T> {
         }
         self.ticker.on_tick();
         self.inner.end(local);
+        crate::sys::thread::remove_thread_name_from_map();
```
BusyJay, line 142: Better move it to the end of function.

BusyJay (approved):
Rest LGTM

PR #12443, opened 2022-04-28
raftstore:turn to probe mode correctly when logappend rejected

tier-cap:

Issue Number: Close #xxx

What's Changed:

```commit-message
turn to probe mode correctly when logappend rejected
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
```


BusyJay:
You should not change raft code. Instead, a new message should be introduced to let leader call `report_unreachable`.

tier-cap:
> You should not change raft code. Instead, a new message should be introduced to let leader call `report_unreachable`.

Right, no change on raft code.
I just add a new message type on ExtraMessage.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2479,6 +2473,12 @@ where
             ExtraMessageType::MsgHibernateResponse => {
                 self.on_hibernate_response(msg.get_from_peer());
             }
+            ExtraMessageType::MsgRejectRaftLogCausedByMemoryUsage => {
```
BusyJay, line 2476: The name is too limited. It can be caused by any reasons.

tier-cap: It's only used by memory limit.

BusyJay: Why? For example, if disk is nearly full, I want to make leader not sending any entries, I can also use this extra message type.

BusyJay: Why not make them use one method?

tier-cap: When reject, to send a heartbeat msg is not a good way.

BusyJay: I don't understand, why reusing a message becomes a wrong way?

tier-cap: Not a wrong way.
Leader turn to Probe, which will be awaken by the follower heartbeat.

PR #12445, opened 2022-04-29
*: Sort dependencies in Cargo.toml in order automatically

### LINKED ISSUE #12444: Sort dependencies in Cargo.toml in order
## Development Task
Sort dependencies in Cargo.toml in order, it's all in the muddle now.

Connor1996:

Issue Number: Close #12444 

What's Changed: Utilize https://github.com/DevinR528/cargo-sort to sort the dependencies.

```commit-message
Sort dependencies in Cargo.toml in order automatically
```

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```


Makefile
```diff
@@ -271,10 +271,11 @@ unset-override:
 
 pre-format: unset-override
 	@rustup component add rustfmt
+	@cargo install -q cargo-sort 
 
 format: pre-format
-	@cargo fmt -- --check >/dev/null || \
-	cargo fmt
+	@cargo fmt -- --check >/dev/null || cargo fmt
```
tabokie, line 277: The first check seems useless.

tabokie (approved):

BusyJay:
What's the point of sorting?

Makefile
Connor1996: The first check is to do the actual format, and the second is to print out the error if there is something wrong with the `cargo fmt`

Connor1996:
> What's the point of sorting?

neater than unordered

Makefile
tabokie: No, the check won't do format. It prints unformatted code to stdout and leave it at that.

I think this line first scans the code in read-only mode, and then do the actual format if needed. But it's not like "format" has a big overhead. I don't see why we need to check in read-only mode first.

Connor1996: I see

BusyJay:
> neater than unordered

They have been sorted in the [past](https://github.com/tikv/tikv/pull/6079), and now they need to be sorted again. I don't think we need to bother change it. Or there should be tools to guarantee its order automatically.

Connor1996:
> > neater than unordered
> 
> They have been sorted in the [past](https://github.com/tikv/tikv/pull/6079), and now they need to be sorted again. I don't think we need to bother change it. Or there should be tools to guarantee its order automatically.

@BusyJay Please check the change of Makefile, it's guaranteed by cargo-sort when running `make format`

BornChanger (commented):
Just curious, if some IDE support automatic sorting?

innerr (approved):

BornChanger:
> make format

Saw your last comment. make format can sort automatically.

BornChanger (approved):

PR #12460, opened 2022-05-06
*: introduce general request tracker

sticnarf:

Issue Number: Ref #12362

What's Changed:

```commit-message
This commit is a large refactoring that moves collecting engine
PerfContext from storage and coprocessor to engine_rocks and
the tracker.

Now, the storage and coprocessor are mostly decoupled with a
specific engine (engine_rocks).

And it introduces a general trakcer mechanism to collect the metrics
of a request during its whole lifetime. It will help us collect more
performance critical data of a single request more easily.
```

### Related changes

### Check List 

Tests 

- Manual test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

**I'm still doing benchmarks with this PR. The `HashMap` in the `Tracker` provides flexibility but may bring overhead. If it has much performance impact, we can change it to a pre-defined struct.**

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_traits/src/perf_context.rs
+++ b/components/engine_traits/src/perf_context.rs
@@ -48,7 +48,9 @@ pub trait PerfContextExt {
 pub enum PerfContextKind {
     RaftstoreApply,
     RaftstoreStore,
+    /// Commands in tikv::storage, the inner str is the command tag.
     Storage(&'static str),
+    /// Coprocessor requests in tikv::coprocessor, the inner str is the request type.
     Coprocessor(&'static str),
 }
 
@@ -60,6 +62,6 @@ pub trait PerfContext: Send {
     /// Reinitializes statistics and the perf level
     fn start_observe(&mut self);
 
-    /// Reports the current collected metrics to prometheus
+    /// Reports the current collected metrics to prometheus and trackers
     fn report_metrics(&mut self, trackers: &[TrackerToken]);
 }
--- a/components/engine_traits/src/perf_context.rs
+++ b/components/engine_traits/src/perf_context.rs
@@ -1,5 +1,6 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 use tikv_util::numeric_enum_serializing_mod;
+use tracker::TrackerToken;
 
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum PerfLevel {
@@ -47,7 +48,8 @@ pub trait PerfContextExt {
 pub enum PerfContextKind {
     RaftstoreApply,
     RaftstoreStore,
-    GenericRead,
+    Storage(&'static str),
+    Coprocessor(&'static str),
 }
 
 /// Reports metrics to prometheus
@@ -59,5 +61,5 @@ pub trait PerfContext: Send {
     fn start_observe(&mut self);
 
     /// Reports the current collected metrics to prometheus
-    fn report_metrics(&mut self);
+    fn report_metrics(&mut self, trackers: &[TrackerToken]);
 }
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -525,7 +525,7 @@ where
             self.kv_wb().write_opt(&write_opts).unwrap_or_else(|e| {
                 panic!("failed to write to engine: {:?}", e);
             });
-            self.perf_context.report_metrics(&[]);
+            self.perf_context.report_metrics(&[]); // TODO: pass in request trackers
             self.sync_log_hint = false;
             let data_size = self.kv_wb().data_size();
             if data_size > APPLY_WB_SHRINK_SIZE {
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -525,7 +525,7 @@ where
             self.kv_wb().write_opt(&write_opts).unwrap_or_else(|e| {
                 panic!("failed to write to engine: {:?}", e);
             });
-            self.perf_context.report_metrics();
+            self.perf_context.report_metrics(&[]);
             self.sync_log_hint = false;
             let data_size = self.kv_wb().data_size();
             if data_size > APPLY_WB_SHRINK_SIZE {
--- a/components/tracker/src/slab.rs
+++ b/components/tracker/src/slab.rs
@@ -6,7 +6,7 @@ use lazy_static::lazy_static;
 use parking_lot::Mutex;
 use slab::Slab;
 
-use crate::Tracker;
+use crate::{metrics::*, Tracker};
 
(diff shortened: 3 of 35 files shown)


components/tracker/src/tls.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn set_tls_tracker(token: TrackerToken) {
+    TLS_TRACKER_TOKEN.with(|c| c.set(token))
+}
+
+pub fn unset_tls_tracker() {
+    TLS_TRACKER_TOKEN.with(|c| c.set(INVALID_TRACKER_TOKEN))
+}
+
+pub fn get_tls_tracker() -> TrackerToken {
+    TLS_TRACKER_TOKEN.with(|c| c.get())
```
you06, line 30: Can we add an assertion like `TLS_TRACKER_TOKEN.with(|c| c.get()) != INVALID_TRACKER_TOKEN` in debug mode?

src/server/service/kv.rs
```diff
@@ -1320,6 +1321,12 @@ fn future_get<E: Engine, L: LockManager, F: KvFormat>(
     storage: &Storage<E, L, F>,
     mut req: GetRequest,
 ) -> impl Future<Output = ServerResult<GetResponse>> {
+    let tracker = GLOBAL_TRACKERS.insert(Tracker::new(RequestInfo::new(
+        req.get_context(),
+        RequestType::KvGet,
+        req.get_version(),
+    )));
+    set_tls_tracker(tracker);
```
you06, line 1329: Where is the usage of this thread-local variable?

components/tracker/src/tls.rs
sticnarf: It's expected that `TLS_TRACKER_TOKEN` can be `INVALID_TRACKER_TOKEN`. To control the total memory usage, we only have limited active trackers. So, it is possible for a request to fail to create a tracker and then the token for this request is `INVALID_TRACKER_TOKEN`. But don't worry, it should not happen in normal cases.

src/server/service/kv.rs
sticnarf: I don't want to pass the token all over through function parameters. So, when the request just comes in, set it in the TLS storage, we can get the token any time later.

When we spawn a future from a context with a tls tracker, the `TrackedFuture` will maintain this tracker in the spawned futures.

components/tracker/src/tls.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use pin_project::pin_project;
+
+use crate::{slab::TrackerToken, Tracker, GLOBAL_TRACKERS, INVALID_TRACKER_TOKEN};
+
+thread_local! {
+    static TLS_TRACKER_TOKEN: Cell<TrackerToken> = Cell::new(INVALID_TRACKER_TOKEN);
+}
+
+pub fn set_tls_tracker(token: TrackerToken) {
```
cfzjywxk, line 18: How about adding a token suffix so it's clearer the targe is token but not tracker itsself.

components/tracker/src/tls.rs
```diff
@@ -0,0 +1,69 @@
...
+    TLS_TRACKER_TOKEN.with(|c| {
+        c.set(token);
+    })
+}
+
+pub fn clear_tls_tracker() {
```
cfzjywxk, line 24: Would there be a situation the tracker is not set appropriately but `with_tls_tracker` is called later so there's no valid tracker information, or if the previous tracker information is not cleared properly? How could we find these unpaired?

components/engine_traits/src/perf_context.rs
```diff
@@ -47,7 +48,8 @@ pub trait PerfContextExt {
 pub enum PerfContextKind {
     RaftstoreApply,
     RaftstoreStore,
-    GenericRead,
+    Storage(&'static str),
```
cfzjywxk, line 51: The two names are a bit confusing, is the `Storage` type for non-coprocessor read prcoessing?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -525,7 +525,7 @@ where
             self.kv_wb().write_opt(&write_opts).unwrap_or_else(|e| {
                 panic!("failed to write to engine: {:?}", e);
             });
-            self.perf_context.report_metrics();
+            self.perf_context.report_metrics(&[]);
```
cfzjywxk, line 528: Is the `&[]` a TODO?

sticnarf: Yes, it's a TODO. But with `&[]`, it will work like it works now.

In https://github.com/tikv/tikv/pull/12783/files#diff-ee7d1a7f274d36b83db93153af39a61497eaea746565e713d02b91d68f25a180, you can see how it is finally passed in.

components/tracker/src/tls.rs
sticnarf: I'm not sure... Integration tests about it would be great but I'm still wondering the proper way to write them.

sticnarf: It's much more difficult... There is little information along the way to check whether we are writing to the right tracker.

cfzjywxk: Yes, it's not easy to verify the collected information itself directly, maybe we could leave this to the latter tests.

you06 (approved):

components/tracker/src/tls.rs
sticnarf: A counter is now added.

cfzjywxk (approved):

components/tracker/src/slab.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+    CURRENT_SHARD_ID.with(|c| {
+        let shard_id = c.get();
+        c.set((shard_id + 1) % SLAB_SHARD_COUNT);
+        shard_id
+    })
+}
+
+pub struct ShardedSlab {
+    shards: [Mutex<TrackerSlab>; SLAB_SHARD_COUNT],
```
youjiali1995, line 32: `CachePadded`?

sticnarf: I'm not very sure if it's necessary. The critical section is very short, so race inside one cache line should be rare. I can add it in the next PR to relieve the possible impact.

src/coprocessor/endpoint.rs
```diff
@@ -468,17 +485,25 @@ impl<E: Engine> Endpoint<E> {
         req: coppb::Request,
         peer: Option<String>,
     ) -> impl Future<Output = MemoryTraceGuard<coppb::Response>> {
+        let tracker = GLOBAL_TRACKERS.insert(::tracker::Tracker::new(RequestInfo::new(
```
TonsnakeLin, line 488: why not alloc a global tracker pool, although the pool always occupy allocated memory which is not large, it will not alloc memory for every request.

Issue #11964, opened 2022-02-14
5.1.4 tikv restart in many testcases after start cpu profiling


mayjiang0203:
### What version of TiKV are you using?

[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Release Version:   5.1.3"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Edition:           Community"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Git Commit Hash:   937df38124b977518eb2b932d7ced8be5d4eb163"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Git Commit Branch: heads/refs/tags/v5.1.4"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["UTC Build Time:    Unknown (env var does not exist when building)"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Rust Version:      rustc 1.53.0-nightly (16bf626a3 2021-04-14)"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Enable Features:   jemalloc mem-profiling portable sse protobuf-codec test-engines-rocksdb cloud-aws cloud-gcp"]
[2022/02/11 22:07:30.025 +00:00] [INFO] [lib.rs:86] ["Profile:           dist_release"]

### Steps to reproduce

test plan : 603473
testbed: endless-oltp-tps-603473-1-940
testcase: [oltp_rel_008_001] , down one tidb during running sysbench read only workload.

### What did you expect?
No tikv restart.

### What did happened?
Tikv restart.

Logs before restart: 
[2022/02/11 21:57:28.872 +00:00] [INFO] [mod.rs:389] ["write report successfully"]
[2022/02/11 22:00:35.945 +00:00] [INFO] [util.rs:544] ["connecting to PD endpoint"] [endpoints=http://tc-pd-2.tc-pd-peer.endless-oltp-tps-603511-1-612.svc:2379]
[2022/02/11 22:00:35.947 +00:00] [INFO] [<unknown>] ["New connected subchannel at 0x7f06344b1fb0 for subchannel 0x7f04757b8700"]
[2022/02/11 22:07:17.775 +00:00] [INFO] [profiler.rs:296] ["starting cpu profiler"]
[...]

Issue #9957, opened 2021-04-05
Built-in CPU profiling causes a segmentation fault on macOS 11


sticnarf:
### What version of TiKV are you using?

```
Release Version:   5.0.0-rc.x
Edition:           Community
Git Commit Hash:   4807a3d610f2809bd6cc5b1e7645dbb01d5e9e52
Git Commit Branch: release-5.0
UTC Build Time:    2021-04-05 07:11:09
Rust Version:      rustc 1.51.0-nightly (bc39d4d9c 2021-01-15)
Enable Features:   jemalloc sse protobuf-codec test-engines-rocksdb cloud-aws cloud-gcp
```

### What operating system and CPU are you using?

macOS 11.2.3. Intel 4258U. 

### Steps to reproduce

Executing a write workload on TiKV (for example running TPC-C on TiDB). Then, use the built-in profiling tool to get a flame graph by `GET http://{TiKV_IP}:20180/debug/pprof/profile?seconds=30`

### What did you expect?

A CPU flame graph is generated.

### What did happened?

A segmentation fault occurs. 

Examining the core dump, we find the CFI parser in `libunwind.dylib` tries to access `0x0`:

```
(lldb) t 38
* thread #38, stop reason = signal SIGSTOP
    frame #0: 0x00007fff2a55e8b0 libunwind.dylib`libunwind::CFI_Parser<libunwind::LocalAddressSpace>::decodeFDE(libunwind::LocalAddressSpace&, unsigned long, libunwind::CFI_Parser<libunwind::LocalAddressSpace>::FDE_Info*, libunwind::CFI_Parser<libunwind::LocalAddressSpace>::CIE_Info*) + 30
libunwind.dylib`libunwind::CFI_Parser<libunwind::LocalAddressSpace>::decodeFDE:
->  0x7fff2a55e8b0 <+30>: movl   (%rsi), %r13d
    0x7fff2a55e8b3 <+33>: leaq   0x4(%rsi), %rbx
    0x7fff2a55e8b7 <+37>: movl   $0xffffffff, %eax         ; imm = 0xFFFFFFFF 
    0x7fff2a55e8bc <+42>: movq   %rbx, -0x30(%rbp)
(lldb) bt
* thread #38, stop reason = signal SIGSTOP
[...]

PR #12480, opened 2022-05-10
*: enable frame pointer & enable cpu profiling on aarch64

### LINKED ISSUE #10658: Fail to profile TiKV in aarch64

TiKV v5.1.0
tiup v1.5.2
libgcc_s-7.3.0-20190804.so.1

```
Linux localhost.localdomain 4.19.90-23.8.v2101.ky10.aarch64 #1 SMP Mon May 17 17:07:38 CST 2021 aarch64 aarch64 aarch64 GNU/Linux
```
```
processor       : 63
model name      : HUAWEI,Kunpeng 920
BogoMIPS        : 200.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma dcpop asimddp asimdfhm
CPU implementer : 0x48
CPU architecture: 8
CPU variant     : 0x1
CPU part        : 0xd01
CPU revision    : 0
```

### Steps to reproduce

Run a TiKV and use the http API to profile it.

### What did you expect?
Succeed in profiling.

### What did happened?

TiKV cores dump.
```
#0  0x0000fffd7b6aceb4 in ?? () from /lib64/libgcc_s.so.1
#1  0x0000fffd7b6ae534 in _Unwind_Backtrace () from /lib64/libgcc_s.so.1
#2  0x0000aaac01eedb58 in backtrace::backtrace::libunwind::trace (cb=...) at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/libunwind.rs:88
#3  backtrace::backtrace::trace_unsynchronized (cb=...) at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/mod.rs:66
#4  pprof::profiler::perf_signal_handler (_signal=<optimized out>) at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/pprof-0.4.2/src/profiler.rs:128
#5  <signal handler called>
```

mornyx:
Issue Number: close #10658

### What's changed

- Set environment variables in `Makefile` to enable frame pointer for TiKV.
- Add two features `pprof-dwarf` and `pprof-fp` to control compilation behavior.
- Re-enable CPU Profiling on ARM architecture.

The stack backtrack implementation of `pprof-rs` which TiKV currently depends on is based on `libunwind`, which may trigger segmentation faults in some scenarios (especially on ARM architecture).

Here are some related issues:

- https://github.com/tikv/tikv/issues/10658
- https://github.com/tikv/tikv/issues/9957
- https://github.com/tikv/tikv/issues/11964

This PR introduces a new version of `pprof-rs`, which provides stack backtracking based on **frame pointers**, and it also implements **memory access protection** to avoid unexpected segmentation faults.

> More tech info is in https://github.com/tikv/tikv/pull/12326 maintained by @YangKeao.

### Performance overhead

Enabling the frame pointer will use the `RBP/x29` register to hold the frame address, one less general purpose register means there may be a performance regression.

Here is a simple [measurement](https://github.com/tikv/pprof-rs/pull/116#issuecomment-1102156557). In general, the frame-pointer-based stack backtrace has no performance regression compared to the original DWARF-based stack backtrace.

In other words, for scenes with CPU Profiling (or Continuous Profiling) enabled, the overall overhead based on frame pointers is much lower than that based on DWARF. But for scenes that don't use profiling, there is a little extra overhead at all times for frame pointers.

### Default behavior and manual control

There are now one additional environment variable that control compilation behavior: `TIKV_FRAME_POINTER`. We set it to `1` by default in the `Makefile`, which means it's enabled by default.

If you want to disable frame-pointer, please manually set the environment variable `TIKV_FRAME_POINTER=0 make` (this will fallback to `libunwind`-based stack backtrace).

If you avoid run `make` and just use `cargo build`, the frame pointer will not be enabled and std will not be recompiled.

### Check List

Tests

- Unit test
- Integration test
- Manual test
    - Deployed TiKV on all platforms of `Linux+x86_64`, `Linux+aarch64`, `macOS+x86_64`, `macOS+aarch64` and verified the results of CPU Profiling.

### Release note

```release-note
None
```


Cargo.toml
```diff
@@ -123,6 +123,7 @@ paste = "1.0"
 pd_client = { path = "components/pd_client", default-features = false }
 pin-project = "1.0"
 pnet_datalink = "0.23"
+pprof = { git = "https://github.com/tikv/pprof-rs.git", rev = "63d3ff6780c49c2b0e14b399803a77a4c5145308", default-features = false, features = ["flamegraph", "frame-pointer", "protobuf-codec"] }
```
YangKeao, line 126: v0.9.0 has been released.

Makefile
```diff
@@ -34,6 +34,13 @@
 SHELL := bash
 ENABLE_FEATURES ?=
 
+# Rust & C/C++ compiler flags
+#
+# Enable frame pointers for stable CPU Profiling.
+export RUSTFLAGS := $(RUSTFLAGS) -Cforce-frame-pointers=yes
```
YangKeao, line 40: Maybe they should be modified in `etc/cargo.config.dist` like what I have done in https://github.com/tikv/tikv/pull/12326?

mornyx: Yes, though, we still need to set `build-std` in `cargo.config.dist`.

sticnarf:
Can we make it optional to use frame pointers? I don't hope to bother to build std on platforms that unwinding with dwarf works mostly fine.

mornyx:
> Can we make it optional to use frame pointers? I don't hope to bother to build std on platforms that unwinding with dwarf works mostly fine.

Sure, but segfaults are currently found on both x86 and ARM, and I haven't figured out when the frame-pointer needs to be disabled.. (unless user never use CPU Profiling)

sticnarf:
> Sure, but segfaults are currently found on both x86 and ARM, and I haven't figured out when the frame-pointer needs to be disabled.. (unless user never use CPU Profiling)

I remember there isn't any segfault on x86_64 Linux if we blocklists glibc and vdso (on some distributions).

mornyx:
> I remember there isn't any segfault on x86_64 Linux if we blocklists glibc (and vdso on some distributions).

A few days ago we discovered a new segfault stack in our internal cluster (Linux on x86), which was crashing a machine about every 10 hours:

origin_img_v2_4b406c0f-8045-401a-a7e3-343f447592cg

It looks like `SIGPROF` hits destructor code inserted by the Rust compiler, and then the segfault occurs when trying to backtrace at this place. I replaced the internal cluster's TiKV with a hotfix version with frame pointers and the problem was solved.

So at least for now, the frame pointer needs to be enabled for all supported platforms. (But we can reserve an option which is enabled by default.)

sticnarf:
> > I remember there isn't any segfault on x86_64 Linux if we blocklists glibc (and vdso on some distributions).
> 
> A few days ago we discovered a new segfault stack in our internal cluster (Linux on x86), which was crashing a machine about every 10 hours:
> 
> origin_img_v2_4b406c0f-8045-401a-a7e3-343f447592cg
> 
> It looks like `SIGPROF` hits destructor code inserted by the Rust compiler, and then the segfault occurs when trying to backtrace at this place. I replaced the internal cluster's TiKV with a hotfix version with frame pointers and the problem was solved.
> 
> So at least for now, the frame pointer needs to be enabled for all supported platforms. (But we can reserve an option which is enabled by default.)

I see. But I just never encounter such a problem during my daily development. Can you at least give me the option to disable it by making it a cargo feature?

mornyx:
> I see. But I just never encounter such a problem during my daily development. Can you at least give me the option to disable it by making it a cargo feature?

It is indeed necessary, I will add.

Makefile
mornyx: Since we need to control whether frame-pointer and build-std are enabled through env vars, writing them in the configuration file is no longer suitable.

sticnarf:
> It's difficult to control whether to enable frame pointers or whether to compile the standard library through cargo features, we provide environment variables, please see the `Default behavior and manual control` section described in this PR for more info.
> 
> cc @sticnarf

Can we add new features `pprof-dwarf` and `pprof-fp`? `pprof-dwarf` enables `pprof/backtrace-rs` and `pprof-fp` enables `pprof/frame-pointer`. So, we can still build version that uses dwarf to unwind and outputs results normally.

cmd/tikv-server/Cargo.toml
```diff
@@ -28,6 +28,9 @@ test-engines-panic = ["server/test-engines-panic"]
 
 nortcheck = ["server/nortcheck"]
 
+pprof-dwarf = ["server/pprof-dwarf"]
+pprof-fp = ["server/pprof-fp"]
```
sticnarf, line 32: Can we directly set `tikv/pprof-*` features here? The features in `server` crate don't seem necessary...

mornyx: @sticnarf Fixed, PTAL~

sticnarf (approved):
LGTM

sticnarf:
@BusyJay PTAL

Makefile
```diff
@@ -34,6 +34,32 @@
...
+endif
+
+# The Rust standard library is recompiled by default. (The purpose is to enable
+# frame pointers in std)
+#
+# If you want to avoid recompiling the Rust standard library, please manually
+# set the environment variable `TIKV_BUILD_STD=0 make` (this will lose CPU Profiling
+# samples related to Rust std functions).
+ifndef TIKV_BUILD_STD
+export TIKV_BUILD_STD=1
```
BusyJay, line 53: What's the build time difference?

mornyx: By measuring the execution time of `make dist_release`, on my laptop it took 26 minutes when build-std was off and 28.5 minutes when it was on.

BusyJay: > this will lose CPU Profiling samples related to Rust std functions

What does it mean? Will it be truncated or ignored?

mornyx: > > this will lose CPU Profiling samples related to Rust std functions
> 
> What does it mean? Will it be truncated or ignored?

If the function in std does not contain a frame pointer, it means that the location on the stack where the frame pointer should be saved will be unknown. When `pprof-rs` encounters such an illegal address during backtracking, the backtracking will be stopped. That means we will lose a lot of samples and the final flame graph will be distorted.

But it doesn't matter, we just need to close `TIKV_FRAME_POINTER` when closing `TIKV_BUILD_STD`, the backtracking will fallback to `libunwind`.

BusyJay: Then why not just use one switch?

mornyx: > Then why not just use one switch?

Although we rarely use one of these options alone, I think behaviorally these are two different things.
Although rare, it is safe to have `TIKV_FRAME_POINTER` on, and `TIKV_BUILD_STD` off, without even specifying `RUSTFLAGS=-Cforce-frame-pointers=yes`. This just gets us a call stack with a depth of 1, and reduces the performance overhead appropriately.

BusyJay: Don't over optimize, simple is better.

Makefile
```diff
@@ -123,6 +150,16 @@ export DOCKER_IMAGE_TAG ?= "latest"
...
 export CARGO_BUILD_PIPELINING=true
 
+# Compiler gave us the following error message when using a specific version of gcc on
+# aarch64 architecture and TIKV_BUILD_STD=1:
+#     .../atomic.rs: undefined reference to __aarch64_xxx
+# This is a temporary workaround.
+# See: https://github.com/rust-lang/rust/issues/93166
+#      https://bugzilla.redhat.com/show_bug.cgi?id=1830472
+ifeq ($(TIKV_BUILD_RUSTC_TARGET),aarch64-unknown-linux-gnu)
+export RUSTFLAGS := $(RUSTFLAGS) -Ctarget-feature=-outline-atomics
```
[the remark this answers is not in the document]
mornyx: See this article about what `outline-atomics` is: https://mysqlonarm.github.io/ARM-LSE-and-MySQL/

When we build the Rust standard library on aarch64, we will get linker errors:
https://ci.pingcap.net/blue/rest/organizations/jenkins/pipelines/build-common/runs/598/nodes/5/steps/20/log/?start=0

This appears to be a bug in the Rust compiler that occurs when outline atomics is enabled. Outline atomics were added with gcc 9.3.1 and turned on by default in gcc 10.1. `target-feature=-outline-atomics` means to turn off this option (note the **-** prefix).

I don't think this should have any effect, just some ARM specific compatibility options at the compile stage.

Makefile
mornyx: > Don't over optimize, simple is better.

What do you think about using a single `TIKV_FRAME_POINTER` to control both frame-pointer option and build-std option? I can change to this pattern.

mornyx: Updated, all behaviors can be controlled with just `TIKV_FRAME_POINTER` now. cc @BusyJay

BusyJay (approved):
To use this integrated perf library, we need to:
1. rebuild std and worsen build time(8%)
2. lose compatibility on some platform
3. frame pointer always on and hurt performance.

And it still can't provide always-on historical perf samples. I don't like this heavy integrations and the trade off. We should optimize for the future and choose the eBPF way.

That being said, I give one +1 for this as it's the quickest way to land on v6.1.

mornyx:
> To use this integrated perf library, we need to:
> 
> 1. rebuild std and worsen build time(8%)
> 2. lose compatibility on some platform
> 3. frame pointer always on and hurt performance.
> 
> And it still can't provide always-on historical perf samples. I don't like this heavy integrations and the trade off. We should optimize for the future and choose the eBPF way.
> 
> That being said, I give one +1 for this as it's the quickest way to land on v6.1.

Thank you Jay, I will continue to work on eBPF after v6.1 and try to release a PoC as soon as possible.

mornyx:
/assign @BusyJay @sticnarf 

Please help me /merge, thanks!

BusyJay:
After this PR, compilation is very easy to trigger oom-killer on 8 vcores 32GiB.

mornyx:
> After this PR, compilation is very easy to trigger oom-killer on 8 vcores 32GiB.

I created https://github.com/tikv/tikv/issues/12591 to track this.

PR #12484, opened 2022-05-10
raftstore: adjust default log gc limit automatically

### LINKED ISSUE #12418: raft_log_gc_count_limit should be adjusted according to the actual region size
## Development Task

Now we are suppose to support using dynamic region size #11515, we also need to change the gc count limit and size limit dynamically.

5kbpers:

Issue Number: Close #12418

What's Changed:

```commit-message
Adjust default raft log gc limit according to region split size.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Adjust default config of log gc limit according to config of region size automatically
```


src/config.rs
```diff
@@ -2966,6 +2965,22 @@ impl TiKvConfig {
             );
         }
 
+        // Raft log gc limit must be 75% region size at least;
+        if self.raft_store.raft_log_gc_size_limit < self.coprocessor.region_split_size * 3 / 4 {
+            warn!(
```
BusyJay, line 2970: How about making it `info`?

BusyJay (approved):

gengliqi (approved):

BusyJay:
Will changing `region-split-size` also adjust `region-max-size`?

tonyxuqqi (approved):

tonyxuqqi:
LGTM

5kbpers:
> Will changing `region-split-size` also adjust `region-max-size`?

No, it will throw an error if `region-max-size` < `region-split-size`. Need to adjust them automatically as well?

BusyJay:
> No, it will throw an error if `region-max-size` < `region-split-size`. Need to adjust them automatically as well?

Yes, otherwise it's difficult to adjust configurations.

5kbpers:
> > No, it will throw an error if `region-max-size` < `region-split-size`. Need to adjust them automatically as well?
> 
> Yes, otherwise it's difficult to adjust configurations.

Done, PTAL again.

components/raftstore/src/coprocessor/config.rs
```diff
@@ -98,16 +98,16 @@ impl Default for Config {
 impl Config {
     pub fn validate(&mut self) -> Result<()> {
         if self.region_max_size.0 < self.region_split_size.0 {
```
BusyJay, line 100: How about making `region_max_size` as `Option<ReadableSize>`? So if it's not set, we can give it a correct number.

components/raftstore/src/coprocessor/config.rs
```diff
@@ -98,16 +98,16 @@ impl Default for Config {
...
-            info!("region_max_size {:?} < region_split_size {:?}, adjust region_split_size to 125% region_max_size automatically",
-                self.region_max_size,
-                self.region_split_size,
+            info!(
+                "region_max_size {:?} < region_split_size {:?}, adjust region_split_size to 125% region_max_size automatically",
+                self.region_max_size, self.region_split_size,
             );
             self.region_max_size = self.region_split_size / 2 * 3;
```
BusyJay, line 105: It doesn't match the log.

5kbpers: Fixed

components/raftstore/src/coprocessor/config.rs
5kbpers: Changed, PTAL again.

src/config.rs
```diff
@@ -2980,9 +2979,30 @@ impl TiKvConfig {
...
 
+        #[cfg(not(any(test, feature = "textexport")))]
+        self.raft_log_gc_limit_adjust();
+
         Ok(())
     }
 
+    pub fn raft_log_gc_limit_adjust(&mut self) {
+        // Raft log gc limit must be 75% region size at least;
+        if self.raft_store.raft_log_gc_size_limit < self.coprocessor.region_split_size * 3 / 4 {
```
BusyJay, line 2990: How about also using `Option`? This should not be a hard limit.

5kbpers: Too many changes were introduced previously, will try again if it was needed.

BusyJay: User should be allowed to use a smaller value for aggressive disk space reclaim.

BusyJay (approved):

BusyJay:
You may also need to calculate default region-split-key based on region-split-size

5kbpers:
> You may also need to calculate default region-split-key based on region-split-size

Done, PTAL.

BusyJay (approved):
So many cases failed.

PR #12488, opened 2022-05-11
txn: do conflict check before returning assertion failure

### LINKED ISSUE #12487: Do constraint check before returning an assertion error
## Development Task
ref: 
https://github.com/tikv/client-go/pull/492
https://github.com/pingcap/tidb/issues/33641

Assertion may be false-positive for a secondary index key.
Briefly the case is :
1. txn-a prewrites key-1
2. txn-b prewrites index-1
3. txn-b prewrites key-1, and encounters lock
4. txn-b rolls back txn-a
5. txn-b prewrites key-1, and successfully commit
6. txn-a prewrites index-1, and succeeds (because there is no constraint check) 

We enforce a constraint check before the assertion error is returned. If the constraint check failed, return it instead of the assertion failure.

ekexium:
Signed-off-by: ekexium <ekexium@gmail.com>

Issue Number: Close #12487 

What's Changed:

We enforce a constraint check before the assertion error is returned. If the constraint check failed, return it instead of the assertion failure.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/commands/prewrite.rs
+++ b/src/storage/txn/commands/prewrite.rs
@@ -2189,7 +2189,7 @@ mod tests {
         );
         assert!(matches!(
             err,
-            MvccError(box MvccErrorInner::AssertionFailed { .. })
+            MvccError(box MvccErrorInner::WriteConflict { .. })
         ));
     }
 
--- a/src/storage/txn/actions/prewrite.rs
+++ b/src/storage/txn/actions/prewrite.rs
@@ -499,17 +499,25 @@ impl<'a> PrewriteMutation<'a> {
             MVCC_PREWRITE_ASSERTION_PERF_COUNTER_VEC.write_loaded.inc();
         }
 
-        match (self.assertion, write) {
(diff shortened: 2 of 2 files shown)


src/storage/txn/commands/prewrite.rs
```diff
@@ -2189,7 +2189,7 @@ mod tests {
         );
         assert!(matches!(
             err,
-            MvccError(box MvccErrorInner::AssertionFailed { .. })
+            MvccError(box MvccErrorInner::WriteConflict { .. })
```
sticnarf, line 2192: Is there a case when the assertion fails on the index key but there is no write conflict now?

ekexium: I can't think of such a case.
We've only encountered 2 false assertion failures, both of which should return write conflicts.

sticnarf (approved):
LGTM

MyonKeminta (approved):

PR #12491, opened 2022-05-11
kvengine: fix LZ4 compression

coocood:
Signed-off-by: Evan Zhou <coocood@gmail.com>

What's Changed:

After merge and lz4 update, the compress_to_buffer method is removed, so the length is not included in the compressed data, cause decompression failure, so we manually prepend the length to the buffer.

### Check List 

Tests 

  - make sure LZ4 is used in config.
  - load sysbench 1m records and check the table count.

### DIFF
--- a/components/kvengine/src/table/sstable/builder.rs
+++ b/components/kvengine/src/table/sstable/builder.rs
@@ -417,8 +417,11 @@ impl BlockBuilder {
         for i in 0..num_entries {
             self.block.build_entry(buf, i, common_prefix_len);
         }
-        if self.compression_tp != NO_COMPRESSION {
-            self.compress();
+        match self.compression_tp {
+            NO_COMPRESSION => (),
+            LZ4_COMPRESSION => self.compress_lz4(),
+            ZSTD_COMPRESSION => self.compress_zstd(),
+            _ => panic!("unexpected compression type {}", self.compression_tp),
         }
         let mut checksum = 0u32;
         if checksum_tp == CRC32_IEEE {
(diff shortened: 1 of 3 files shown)


components/kvengine/src/table/sstable/builder.rs
```diff
@@ -417,8 +417,11 @@ impl BlockBuilder {
         for i in 0..num_entries {
             self.block.build_entry(buf, i, common_prefix_len);
         }
-        if self.compression_tp != NO_COMPRESSION {
-            self.compress();
+        match self.compression_tp {
+            NO_COMPRESSION => (),
+            LZ4_COMPRESSION => self.compress_lz4(),
+            ZSTD_COMPRESSION => self.compress_zstd(),
+            _ => panic!("unexpected compression type {}", self.compression_tp),
```
youjiali1995, line 424: It's better to use enum to represent compression type. I can refactor it later.

components/kvengine/src/table/sstable/builder.rs
```diff
@@ -487,36 +490,38 @@ impl BlockBuilder {
         }
     }
 
-    fn compress(&mut self) {
+    fn compress_lz4(&mut self) {
         unsafe {
+            self.buf.put_u32_le(self.compression_buf.len() as u32);
```
youjiali1995, line 495:
```suggestion
```

components/kvengine/src/table/sstable/builder.rs
coocood: Yes, we can do it in another PR.

youjiali1995 (approved):

Issue #12492, opened 2022-05-11
config: TiKV should provide a command to output the description of all config fields


glorv:
### Is your feature request related to a problem? Please describe:

TiKV contains so many config fields and currently, user can only find them on the pingcap document website.

### Describe the feature you'd like:

It will be very convenient to provide a command to output the description of all user-visible config fields. Like following:
```
> ./tikv-server -h
FLAGS: 
...
    -C, --config <FILE>                      Set the configuration file
        --config-info                        Print the configuration information in json format
...
>
> ./tikv-server --config-info json --config ./tikv.toml
{
        "Component": "TiKV Server",
        "Version": "6.1.0",
        "Parameters": [
                {"Name": "abort-on-panic", "Type": "Boolean", "DefaultValue": false, "ValueInFile": false, "description":"xxx" },
                {"Name": "wal-recovery-mode", "Type": "String", "DefaultValue": "/tmp/tikv/store", "ValueInFile": null, "description":"xxx" },
                {"Name": "readpool.unified.max-tasks-per-worker", "Type": "Number", "MinValue": 2, "MaxValue": 1000000, "DefaultValue": 2000, "ValueInFile": null, "description":"xxx"},
                {"Name": "wal-recovery-mode", "Type": "Array", "ValueOptions": [0, 1, 2, 3], "DefaultValue": 2, "ValueInFile": null, "description":"xxx"},
                ...
        ]
}
```

BusyJay:
This is a big feature. I don't see any discussion on this. What's the motivation of the feature? Who is benefited from the feature? Why reading docs, and we are all supposed to read docs, is not enough?

BusyJay:
After discussion, the feature is to provide a way to make people access default value easily using command line tools. This can be convenient for checking configuration change during upgrading. Documentation is still the expected source of truth. For configuration explanation, users are expected to read documentations.

So the scope of the feature is to output default value, current value (reading from configuration file) and configuration name by command line.

PR #12493, opened 2022-05-11
config: add config_info procedure macro

glorv:

Issue Number: Ref #12492

What's Changed:

Add a new procedure macro `ConfigInfo` like `OnlineConfig` that auto-generates the config description base on the `config_info` attributes.

NOTE: this PR only contains the proc-macro implement, a following PR will apply it to TiKVConfig and implement the new command parameter `--config-info`

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/config_info/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+///
+///
+/// Example:
+/// Please refer the unit test at the bottom of this file.
+///
+pub trait ConfigInfo {
+    type Encoder: Serialize;
+    /// Get encoder that can be serialize with `serde::Serializer`
+    /// with the disappear of `#[config_info(skip)]` field
+    // NOTE: the `input` parameter is use to extract the value_in_file value,
```
tabokie, line 66: Why not always use `self`?

components/config_info/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+        /// This is a field that it's value bound depend on other field or the env variable.
+        #[config_info(
+            default_desc = "MAX(4, CPU * 0.8)",
+            min_desc = "value of `a`",
+            max_desc = "`a` * 5"
+        )]
+        g: u64,
+        /// This is a example of Vec type.
+        h: Vec<u64>,
+        /// This is a example of array type.
```
tabokie, line 195: Is it possible to parse other metadata from comment as well?

components/config_info/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+#[derive(Serialize)]
+pub enum FieldCfgType {
+    /// represent `[T; N]` and `Vec<T>` types.
+    Array,
+    /// represent `bool` type
+    Boolean,
+    /// represent map types `HashMap<K, V>` or `BTreeMap<K, V>`
+    Map,
```
tabokie, line 142: What config uses this type?

components/config_info/src/lib.rs
glorv: the `default` value is fetched from `self` , and the `value-in-file` is fetched from the `` which is the value parsed from the actual config when `-config tikv-server.toml` is provided. The current implement only return the `value-in-file` when its value is different from the default value.

components/config_info/src/lib.rs
glorv: I think so, If we define a special format for each attribute, but there seems no much benefit to do so.

components/config_info/src/lib.rs
glorv: The only config is `server.labels` by now.

components/config_info/src/lib.rs
tabokie: I think comment is cleaner and more readable. It would look cumbersome to add yet another modifier other than `online_config`.

components/config_info/src/lib.rs
tabokie: I presume the reason to not use `Default::default()` is because some default values are generated in `sanitize`? I actually prefer we print out those as well because sometimes it's hard to figure out what the `sanitize` outputs without looking at the code.

components/config_info/src/lib.rs
glorv: But document comment is syntax sugar for `#[log = "..."]`, put user-defined attributes to doc comments and manually parse them in the proc-macro looks a bit weird to me.

tabokie: I think parsing doc comment is actual acceptable. Doc tests are generated by parsing code snippet in doc comments. We can introduce our own "dialect" if needed.

In general, I just don't like the idea of adding `#[config_info(..)]` to every config items. Especially when you are putting meaningful information into it.

glorv:
#12517 is the draft PR to apply this new proc-marco to tikv-server

PR #12505, opened 2022-05-13
Raftstore: fix split skip bug due to incorrectly setting has_calculated_region_size

### LINKED ISSUE #12510: Raftstore: tikv does not split regions whose  size is larger than the max region size after restart with config change
6.0-alpha

CentOS 7/8

### Steps to reproduce
Set region max size/ region split size with 1GB and upload about 30GB data.  Waiting for the region count stable and then change the max size/region split size with 200MB, meanwhile enable region bucket and set region bucket size as 100MB. 
Restart tikvs.

### What did you expect?
TiKV'e regions will be splitted after the reboot.

### What did happened?
TiKV's regions will not be splitted. 

related bug: https://github.com/tikv/pd/issues/4926

tonyxuqqi:

Issue Number: Close #12510

What's Changed:
Related bug:  tikv/pd#4926

```
rename has_calculated_region_size to may_skip_split_check and only set it  true after split check task is scheduled. 
```

### Related changes

### Check List 

Tests 

- Manual test
- No code
- Unit test

Side effects

None

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -4880,8 +4880,8 @@ where
         // should work even if we change the region max size.
         // If peer says should update approximate size, update region size and check
         // whether the region should split.
-        // We assume that `may_skip_split_check` is only set true after the split check task is 
-        // scheduled. 
+        // We assume that `may_skip_split_check` is only set true after the split check task is
+        // scheduled.
         if self.fsm.peer.may_skip_split_check
             && self.fsm.peer.compaction_declined_bytes < self.ctx.cfg.region_split_check_diff.0
             && self.fsm.peer.size_diff_hint < self.ctx.cfg.region_split_check_diff.0
@@ -4937,7 +4937,7 @@ where
         self.fsm.peer.size_diff_hint = 0;
         self.fsm.peer.compaction_declined_bytes = 0;
         // the task is scheduled, next tick may skip it.
-        self.fsm.peer.may_skip_split_check = true;  
+        self.fsm.peer.may_skip_split_check = true;
     }
 
     fn on_prepare_split_region(
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -1149,7 +1149,6 @@ where
     fn on_clear_region_size(&mut self) {
         self.fsm.peer.approximate_size = None;
         self.fsm.peer.approximate_keys = None;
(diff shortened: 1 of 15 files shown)


components/raftstore/src/store/worker/split_check.rs
```diff
@@ -515,6 +515,8 @@ where
                             "region_id" => region.get_id(),
                         );
                     }
+                    let region_id = region.get_id();
```
BusyJay, line 518: Why not just also trigger split if necessary here?

nolouch:
Does https://github.com/tikv/pd/issues/4926 relative to this fix?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1149,7 +1149,6 @@ where
     fn on_clear_region_size(&mut self) {
         self.fsm.peer.approximate_size = None;
         self.fsm.peer.approximate_keys = None;
-        self.fsm.peer.has_calculated_region_size = false;
```
BusyJay, line 1152: Why remove this?

BusyJay (commented):
Any test case?

BusyJay:
> Does [tikv/pd#4926](https://github.com/tikv/pd/issues/4926) relative to this fix?

Yes.

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: It's moved into register tick.

tonyxuqqi:
> Any test case?

Should it be covered by existing test cases at least?

tests/integrations/raftstore/test_split_region.rs
```diff
@@ -610,6 +610,45 @@ fn test_node_split_region_diff_check() {
     test_split_region_diff_check(&mut cluster);
 }
 
+// Test steps
+// set max region size/split size 2000 and put data till 100
```
BusyJay, line 614: You mean `1000`?

BusyJay (approved):

BusyJay:
I think the issue number is wrong.

tonyxuqqi:
> I think the issue number is wrong.

You're right. It's PD's issue#. I just created one for TiKV.

tonyxuqqi:
> I think the issue number is wrong.

You're right it's PD#, it should be marked as tikv/pd# instead.

5kbpers (approved):

PR #12523, opened 2022-05-16
add RawKV GC metric versions count

ystaticy:

Issue Number: Ref https://github.com/tikv/tikv/issues/12535

What's Changed: 

```release-note 
Fix RawKV Compaction filter report total is aways 0
```

### Check List 

Tests 

Test steps as follows:
1. To set api version in tikv.toml as follows:
enable-ttl = true
api-version = 2
2. start tikv cluster.
3. write data until trigger compaction.
4. Observation log output ```total``` value is not always 0.

### DIFF
--- a/src/server/gc_worker/rawkv_compaction_filter.rs
+++ b/src/server/gc_worker/rawkv_compaction_filter.rs
@@ -196,6 +196,8 @@ impl RawCompactionFilter {
             if commit_ts.into_inner() >= self.safe_point {
                 return Ok(CompactionFilterDecision::Keep);
             }
+
+            self.versions += 1;
             let raw_value = ApiV2::decode_raw_value(value)?;
             // If it's the latest version, and it's deleted or expired, it needs to be sent to GCWorker to be processed asynchronously.
             if !raw_value.is_valid(self.current_ts) {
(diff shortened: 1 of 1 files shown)


src/server/gc_worker/rawkv_compaction_filter.rs
```diff
@@ -196,6 +196,8 @@ impl RawCompactionFilter {
             if commit_ts.into_inner() >= self.safe_point {
                 return Ok(CompactionFilterDecision::Keep);
             }
+
+            self.versions += 1;
```
pingyu, line 200: Should only entries with `!raw_value.is_valid()` be counted ?

ystaticy: In txn gc of compaction_filter.rs  records will ```self.versions += 1;```,
which is not match ```commit_ts > self.safe_point || value_type != CompactionFilterValueType::Value```

ystaticy: ```versions``` means: Total stale versions meet by compaction filters.
It is defined in ```CompactionFilterStats```
```
pub struct CompactionFilterStats {
    pub versions: Cell<usize>, // Total stale versions meet by compaction filters.
    pub filtered: Cell<usize>, // Filtered versions by compaction filters.
    pub last_report: Cell<Instant>,
}
```

pingyu (approved):
LGTM~

pingyu (approved):

MyonKeminta (approved):

PR #10386, opened 2021-06-17
statistic: more stable test for query test

lhy1024:
Signed-off-by: lhy1024 <admin@liudos.us>

### What problem does this PR solve?

Issue Number: close #10380 

Problem Summary:

query test is unstable, for write request statistics it will be occasionally inaccurate

### What is changed and how it works?

What's Changed:

Add more detections while being compatible with occasional redundant qps statistics

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
```
rm t.log
for i in {1..100}
do
    cargo test --package tests --test integrations -- raftstore::test_stats::test_query_stats --exact --nocapture &>> t.log
    sleep 1
done
cat t.log | grep panic
```
there is only panic with `No address added out of total 1 resolved`
image

### Release note 

```release-note
None
```

PR #12552, opened 2022-05-17
tests: replace `std::time::SystemTime` with `std::time::Instant`

lhy1024:

According to https://github.com/tikv/tikv/pull/10386/files#r874333838

Tests 

- Unit test

### Release note 

```release-note
None.

```


sticnarf (approved):

BusyJay (approved):

lhy1024:
@BusyJay @sticnarf

BusyJay:
I suggest to merge #12562 instead, which also makes some cases stable.

Issue #12557, opened 2022-05-17
RawKV GCTask need to return handled and wasted


ystaticy:
## Development Task
RawKV GCTask need to return handled and wasted

PR #12558, opened 2022-05-17
Rawkv gctask metrics

ystaticy:

Issue Number: Ref https://github.com/tikv/tikv/issues/12557

What's Changed: 

```release-note 
RawKV GCTask return handled and wasted count
```

### Check List 

Tests 

- Manual test 
If the data is the latest version is deleted or expired and to Compcation,it will trigger ```GcTask::RawGcKeys```
We can get ```tikv_gc_compaction_filter_mvcc_deletion_handled``` value is not ```0``` in grafana

### DIFF
--- a/src/server/gc_worker/gc_worker.rs
+++ b/src/server/gc_worker/gc_worker.rs
@@ -495,6 +495,7 @@ where
         let mut raw_modifies = MvccRaw::new();
         let mut keys = get_keys_in_regions(keys, regions_provider)?;
 
+        let (mut handled_keys, mut wasted_keys) = (0, 0);
         let mut gc_info = GcInfo::default();
         let mut next_gc_key = keys.next();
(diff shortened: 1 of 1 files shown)


MyonKeminta (approved):
LGTM. Not sure whether using the same metrics as transactional GC is a good idea though.

pingyu (approved):
LGTM

pingyu:
> LGTM. Not sure whether using the same metrics as transactional GC is a good idea though.

I think it would be better to separate metrics for different KV modes later.

ystaticy:
> > LGTM. Not sure whether using the same metrics as transactional GC is a good idea though.
> 
> I think it would be better to separate metrics for different KV modes later.

Ok, the metrics will be separate later.

iosmanthus (approved):
LGTM

PR #12576, opened 2022-05-19
cdc: make tso worker threads configuable

hicqu:

Issue Number: ref #12592

Add a configuration entry to custom cdc tso worker threads. With the config `cdc.min-ts-interval = 100ms`, test shows that 130K Region will utilize the thread about 70% usage. So single tso worker thread can be a bottle neck for some extreme cases.

By the way this PR also removes some noisy logs. And #12592 is also resolved.

```commit-message
 cdc: make tso worker threads configuable 
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```


src/config.rs
```diff
@@ -2416,6 +2423,7 @@ impl Default for CdcConfig {
             // is more than 500MB/s, so 128MB/s is enough.
             incremental_scan_speed_limit: ReadableSize::mb(128),
             incremental_scan_ts_filter_ratio: 0.2,
+            tso_worker_threads: 1,
```
zhangjinpeng1987, line 2426: better to add a range limitation for its value.

zhangjinpeng1987: How to change it on the Cloud?

hicqu: Do you mean dynamically change it in TiDB? It's not available now.

overvenus (approved):

Issue #12574, opened 2022-05-19
[Dynamic Region] tikv panic repeatly


Lily2025:
### What version of TiKV are you using?
/ # ./tikv-server -V
TiKV 
Release Version:   6.1.0-alpha
Edition:           Community
Git Commit Hash:   2f4b1a5ecefbfc15ceaa471f25dc4700117f1114
Git Commit Branch: heads/refs/tags/v6.1.0-nightly
UTC Build Time:    2022-05-18 11:05:16
Rust Version:      rustc 1.60.0-nightly (1e12aef3f 2022-02-13)
Enable Features:   jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
Profile:           dist_release

### What operating system and CPU are you using?
8core、16GB

### Steps to reproduce
    "coprocessor": {
        "split-region-on-table": false,
        "batch-split-limit": 10,
        "region-max-size": "1536MiB",
        "region-split-size": "1GiB",
        "region-max-keys": 144000000000,
        "region-split-keys": 96000000000,
        "consistency-check-method": "mvcc",
        "enable-region-bucket": true,
        "region-bucket-size": "128MiB",
        "region-size-threshold-for-approximate": "384MiB",
        "region-bucket-merge-size-ratio": 0.33
    },
[2022/05/18 16:30:36.018 +08:00] [INFO] [cmd.go:124] ["Start remote command"] [cmd="sysbench --mysql-user=root --db-driver=mysql --time=0 --report-interval=10 --mysql-db=sysbench_64_7000w --mysql-host=tc-tidb.glh-dynamic-regions-jw8pv --mysql-port=4000 --threads=512 --tables=64 --table-size=70000000 --mysql-ignore-errors=2013,1213,1105,1205,8022,8027,8028,9004,9007 --rand-type=gaussian oltp_read_only run"] [nodename=benchtoolset]

### What did you expect?
all tikv are normal

### What did happened?
some tikv panic several times

PR #12586, opened 2022-05-19
Raftstore: reset the bucket after the regions merge

tonyxuqqi:

Issue Number: Ref #12574

What's Changed:

```commit-message
Reset the buckets of target region after the merge, this is to refresh the buckets in-time. 
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -4118,8 +4118,6 @@ where
         source: metapb::Region,
     ) {
         self.register_split_region_check_tick();
-        self.fsm.peer.region_buckets = None;
-        self.fsm.peer.last_region_buckets = None;
         let mut meta = self.ctx.store_meta.lock().unwrap();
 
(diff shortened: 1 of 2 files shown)


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4118,6 +4118,8 @@ where
         source: metapb::Region,
     ) {
         self.register_split_region_check_tick();
+        self.fsm.peer.region_buckets = None;
```
BusyJay, line 4121: How about moving it to L4167? We may need to check all other places that `size_diff_hint` is reset.

BusyJay (approved):

components/raftstore/src/store/fsm/peer.rs
tonyxuqqi: done.  Good point.

5kbpers (approved):

Issue #12595, opened 2022-05-20
when enable api v2 , TiDB resolve lock meet error


ystaticy:
### What version of TiKV are you using?

```
[2022/05/20 15:37:26.939 +08:00] [INFO] [lib.rs:79] ["Welcome to TiKV"]
[2022/05/20 15:37:26.939 +08:00] [INFO] [lib.rs:84] ["Release Version:   6.1.0-alpha"]
[2022/05/20 15:37:26.939 +08:00] [INFO] [lib.rs:84] ["Edition:           Community"]
[2022/05/20 15:37:26.940 +08:00] [INFO] [lib.rs:84] ["Git Commit Hash:   1fb8980ccab9ff40c1adc206df52952dab8e8ad8"]
[2022/05/20 15:37:26.940 +08:00] [INFO] [lib.rs:84] ["Git Commit Branch: master"]
[2022/05/20 15:37:26.940 +08:00] [INFO] [lib.rs:84] ["UTC Build Time:    Unknown (env var does not exist when building)"]
[2022/05/20 15:37:26.940 +08:00] [INFO] [lib.rs:84] ["Rust Version:      rustc 1.60.0-nightly (1e12aef3f 2022-02-13)"]
[2022/05/20 15:37:26.940 +08:00] [INFO] [lib.rs:84] ["Enable Features:   jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raf
t-engine cloud-aws cloud-gcp cloud-azure"]
```

### What operating system and CPU are you using?

CentOS 7.6
Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz

### Steps to reproduce

tikv configuration is :
```
enable-ttl = true
api-version = 2
```

and start PD,TiKV,TiDB. wait TiDB to start update safepoint.

### What did you expect?
gc process can execute success.

### What did happened?

TiDB log:
```
[2022/05/20 15:47:34.821 +08:00] [INFO] [range_task.go:281] ["canceling range task because of error"] [name=resolve-locks-runner] ["failed startKey"=] ["failed endKey"=] [error="unexpected scanlock error: error:<abort:\"Error(InvalidKeyRangeMode { cmd: scan_lock, storage_api_version: V2, range: (None, Some(\\\"7480000000000000FF0500000000000000F8\\\")) })\" > "]
[...]

PR #12596, opened 2022-05-20
fixbug:when enable api v2 , TiDB resolve lock meet error

ystaticy:

Issue Number: Close https://github.com/tikv/tikv/issues/12595

What's Changed:

```
If cmd is scan_lock , it will not execute invalid_key_range_mode check.
```
### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
fix:when enable api v2 , TiDB resolve lock meet error
```

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -1263,8 +1263,8 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
                     .get(priority_tag)
                     .inc();
 
-                // Remove the check_api_version_ranges,
-                // because of resolve locks range from TiDB is invalid key range mode
+                // Do not check_api_version in scan_lock, to be compatible with TiDB gc-worker,
+                // which resolves locks on regions, and boundary of regions will be out of range of TiDB keys.
 
                 let command_duration = tikv_util::time::Instant::now_coarse();
 
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -485,9 +485,7 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
                         range.0.as_ref().map(AsRef::as_ref),
                         range.1.as_ref().map(AsRef::as_ref),
                     );
-                    if ApiV2::parse_range_mode(range) != KeyMode::TiDB
-                        && cmd != CommandKind::scan_lock
-                    {
+                    if ApiV2::parse_range_mode(range) != KeyMode::TiDB {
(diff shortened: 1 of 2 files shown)


src/storage/mod.rs
```diff
@@ -485,7 +485,9 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
                         range.0.as_ref().map(AsRef::as_ref),
                         range.1.as_ref().map(AsRef::as_ref),
                     );
-                    if ApiV2::parse_range_mode(range) != KeyMode::TiDB {
+                    if ApiV2::parse_range_mode(range) != KeyMode::TiDB
```
pingyu, line 488: Suggest to just remove `check_api_version_ranges` in `scan_lock` [here](https://github.com/tikv/tikv/blob/1fb8980ccab9ff40c1adc206df52952dab8e8ad8/src/storage/mod.rs#L1267). Also leave comments about why.

ystaticy: OK. I already fix it in ```scan_lock```, and use the old conditional judgment inside ```check_api_version_ranges```.

src/storage/mod.rs
```diff
@@ -1264,15 +1263,8 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
-                    api_version,
-                    ctx.api_version,
-                    CMD,
-                    [(
-                        start_key.as_ref().map(Key::as_encoded),
-                        end_key.as_ref().map(Key::as_encoded),
-                    )],
-                )?;
+                // Remove the check_api_version_ranges,
+                // because of resolve locks range from TiDB is invalid key range mode
```
pingyu, line 1267: Suggestion:
Do not `check_api_version` in `scan_lock`, to be compatible with TiDB gc-worker,
which resolves locks on regions, and boundary of regions will be out of range of TiDB keys.

pingyu (commented):
Rest LGTM~

MyonKeminta (approved):

iosmanthus (approved):
LGTM

tests/integrations/storage/test_storage.rs
```diff
@@ -966,7 +966,7 @@ fn test_txn_store_txnkv_api_version() {
                 store.batch_get_command_err(&[key, key, key], 10);
 
                 store.scan_err(key, None, 100, 10);
-                store.scan_locks_err(20, key, &end_key, 10);
+                store.scan_locks_ok(20, key, &end_key, 10, vec![]);
```
pingyu, line 969: Please comment that why it is ok here.

PR #12602, opened 2022-05-23
server: support return simplified metrics response

glorv:

Issue Number: ref #12355

What's Changed:

```commit-message
```

This PR is split from #12417 by only filter counter with 0 value and histogram with 0 samples. This will make it much easier to review and can be released in v6.1.

In my benchmark: By applying this simple logic, the /metrics response size can be reduced to about 35% with tpcc workload. When `simplify-metrics` is set to true, there is no much effect for the grafana charts, the only user-awareness change is some counter value show no-vaule instead of 0 value. 

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```


glorv:
@BusyJay @5kbpers PTAL thanks~

glorv:
@BusyJay @innerr @5kbpers PTAL, thanks~

innerr (approved):
LGTM, but the release note is not right for now (should be none?)

5kbpers:
Any side effects of it?

glorv:
> Any side effects of it?

The only side effect is that(as described in the PR desc): for counter-type charts, the displayed value will be non-value instead of 0 if `simplify-metrics` is set to true.

5kbpers (approved):

PR #12612, opened 2022-05-23
cdc: skip prewrite without value

hicqu:

Issue Number: ref https://github.com/tikv/tikv/issues/12717.
`
What's Changed:

```commit-message
cdc: skip prewrite without value
```

### Related changes

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/cdc/src/delegate.rs
+++ b/components/cdc/src/delegate.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use api_version::{ApiV2, KeyMode, KvFormat};
-use collections::HashMap;
+use collections::{HashMap, HashMapEntry};
 use crossbeam::atomic::AtomicCell;
 use kvproto::{
     cdcpb::{
@@ -721,7 +721,16 @@ impl Delegate {
                     );
                 }
 
-                rows.insert(row.key.clone(), (row, has_value));
+                match rows.entry(row.key.clone()) {
+                    HashMapEntry::Occupied(o) => {
+                        let o = o.into_mut();
+                        mem::swap(&mut o.0.value, &mut row.value);
+                        o.0 = row;
+                    }
+                    HashMapEntry::Vacant(v) => {
+                        v.insert((row, has_value));
+                    }
+                }
             }
             "lock" => {
                 let (mut row, mut has_value) = (EventRow::default(), false);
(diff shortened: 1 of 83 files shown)


components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -2316,3 +2316,39 @@ fn test_resolved_ts_with_learners() {
...
+
+#[test]
+fn test_prewrite_without_value() {
+    let cluster = new_server_cluster(0, 2);
+    cluster.pd_client.disable_default_operator();
+    let mut suite = TestSuiteBuilder::new().cluster(cluster).build();
+    let rid = suite.cluster.get_region(&[]).id;
+    let ctx = suite.get_context(rid);
+    let client = suite.get_tikv_client(rid).clone();
+    let large_value = vec![b'x'; 1024];
```
overvenus, line 2328:
```suggestion
    let large_value = vec![b'x'; 2*txn_types::SHORT_VALUE_MAX_LEN];
```

components/cdc/src/delegate.rs
```diff
@@ -780,16 +760,20 @@ impl Delegate {
...
                 }
 
-                *occupied = row;
+                let occupied = rows.entry(row.key.clone()).or_default();
+                if occupied.1 {
+                    assert!(!has_value);
+                    has_value = true;
+                    mem::swap(&mut occupied.0.value, &mut row.value);
+                }
+                *occupied = (row, has_value);
```
overvenus, line 769: How about `has_value = !row.value.is_empty()`? So we don't have to change `decode_write` and `decode_default`.

components/cdc/src/delegate.rs
```diff
@@ -481,83 +481,63 @@ impl Delegate {
         let mut rows = vec![Vec::with_capacity(entries_len)];
         let mut current_rows_size: usize = 0;
         for entry in entries {
+            let (mut row, mut _has_value) = (EventRow::default(), false);
```
hi-rustin, line 484: Why does `_has_value` have a `_` prefix?

components/cdc/src/delegate.rs
hi-rustin: This makes sense to me.

hi-rustin (commented):
rest LGTM

components/cdc/src/delegate.rs
hi-rustin: +1 👍

overvenus (approved):

components/cdc/src/delegate.rs
hicqu: We have discussed it offline. If there is an empty short value, `has_value` should also be `true` but `row.value` will be empty.

hi-rustin (approved):

Issue #12615, opened 2022-05-23
Pessimistic prewrite may be not idempotent on CommitTsTooLarge


sticnarf:
## Bug Report

Considering the following case:
1. All prewrites of a pessimistic async-commit transaction succeeds at first. But the client does not receive the responses of non-pessimistic keys.
2. Some keys of the transaction are committed because of async lock resolve.
3. In the retried prewrite for a non-pessimistic key, the calculated `commit_ts` exceeds `max_commit_ts`. Then, it returns async commit fallback to the client and successfully prewrites the lock.
4. The client executes a legacy percolator commit.

This may cause several possible problems.

- If the primary lock is committed in step 2, the final commit may fail (TxnLockNotFound). The end user may receive an error while the transaction commits successfully.
- If the non-pessimistic key is committed in step 2, the final commit may commit this key twice, causing different commit TS for a single transaction.

~~This might be related to https://github.com/pingcap/tidb/issues/34875.~~

PR #12616, opened 2022-05-23
txn: check committed record before async commit fallback

sticnarf:

Issue Number: Close https://github.com/tikv/tikv/issues/12615

What's Changed:

```commit-message
If a retried prewrite encounters CommitTsTooLarge, it may cause correctness
issues (#12615). So, like other errors, this commit adds a check for committed
records to make prewrite idempotent.
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression

It increases cost on `CommitTsTooLarge`. But this error should not happen very frequently in normal production.

### Release note 

```release-note
Fix possible duplicate commit record in async-commit pessimistic transactions.
```

### DIFF
--- a/src/storage/txn/commands/prewrite.rs
+++ b/src/storage/txn/commands/prewrite.rs
@@ -517,6 +517,19 @@ impl<K: PrewriteKind> Prewriter<K> {
                             .insert(key, (old_value, Some(mutation_type)));
                     }
                 }
+                Ok((..)) => {
+                    // If it needs min_commit_ts but min_commit_ts is zero, the lock
+                    // has been prewritten and has fallen back from async commit or 1PC.
+                    // We should let later keys prewrite in the old 2PC way.
+                    props.commit_kind = CommitKind::TwoPc;
+                    async_commit_pk = None;
+                    self.secondary_keys = None;
+                    self.try_one_pc = false;
+                    fallback_1pc_locks(txn);
+                    // release memory locks
+                    txn.guards = Vec::new();
+                    final_min_commit_ts = TimeStamp::zero();
+                }
                 Err(MvccError(box MvccErrorInner::WriteConflict {
                     start_ts,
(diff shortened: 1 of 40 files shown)


cfzjywxk (approved):

src/storage/txn/commands/prewrite.rs
```diff
@@ -528,6 +528,13 @@ impl<K: PrewriteKind> Prewriter<K> {
                     return check_committed_record_on_err(prewrite_result, txn, reader, &key);
                 }
                 Err(MvccError(box MvccErrorInner::CommitTsTooLarge { .. })) | Ok((..)) => {
+                    // The prewrite might be a retry and the record may have been committed.
+                    // So, we need to prevent the fallback to avoid duplicate commits.
+                    if let Ok(res) =
+                        check_committed_record_on_err(prewrite_result, txn, reader, &key)
```
youjiali1995, line 534: L530 matches `Ok((..))`. Is it expected? BTW, I forget why it matches `OK((..))` here.

sticnarf: Sadly, this definitely causes some problems... If it needs min_commit_ts but min_commit_ts is zero, the lock
has been prewritten and has fallen back from async commit or 1PC.

sticnarf:
@youjiali1995 Do you mind taking a look again?

src/storage/txn/commands/prewrite.rs
sticnarf: Yes.

sticnarf: > I'm a bit confused. It seems `CommitTsTooLarge` is consumed inside and this error would not be returned no matter if it's a retried request or not?

`CommitTsTooLarge` is always consumed and not returned to the client...

PR #12620, opened 2022-05-23
raftstore: bucket and region split bug fix

### LINKED ISSUE #12597: [Dynamic Region] Some large regions did not split as expected
./tikv-server -V
 TiKV 
Release Version:   6.1.0-alpha
Edition:           Community
Git Commit Hash:   a23fd680807d35d60828dc298cf99eca7448faa2
Git Commit Branch: HEAD
UTC Build Time:    2022-05-19 04:37:08
Enable Features:   jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
2022-05-19T18:30:12.351+0800	INFO	k8s/client.go:107	it should be noted that a long-running command will not be interrupted even the use case has ended. For more information, please refer to https://github.com/pingcap/test-infra/discussions/129
./pd-server -V
 Release Version: v6.1.0-nightly
Edition: Community
Git Commit Hash: 85918318d7a127a845af25e208b82b153820fc0d
Git Branch: heads/refs/tags/v6.1.0-nightly
UTC Build Time:  2022-05-18 10:57:30
2022-05-19T18:30:12.527+0800	INFO	k8s/client.go:107	it should be noted that a long-running command will not be interrupted even the use case has ended. For more information, please refer to https://github.com/pingcap/test-infra/discussions/129
./tidb-server -V
 Release Version: v6.1.0-nightly
Edition: Community
Git Commit Hash: 381e870c5ca018d9b5b6e89233d2cbd9a3b41cac
Git Branch: heads/refs/tags/v6.1.0-nightly
UTC Build Time: 2022-05-18 11:04:58
GoVersion: go1.18.2
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
Check Table Before Drop: false

8core、16GB

### Steps to reproduce
update config repeatedly while run tpcc
[2022/05/19 18:42:08.656 +08:00] [INFO] [cmd.go:124] ["Start remote command"] [cmd="go-tpc tpcc run -D tpcc --host tc-tidb.glh-dynamic-regions-jw8pv -P4000 --warehouses 1000 -T 64 --time 36000m --ignore-error '2013,1213,1105,1205,8022,8027,8028,9004,9007,9001'"] [nodename=benchtoolset]

### What did you expect?
big region can split

### What did happened?
current config
    "coprocessor": {
        "split-region-on-table": false,
        "batch-split-limit": 10,
        "region-max-size": "800MiB",
        "region-split-size": "512MiB",
        "region-max-keys": 8000000,
        "region-split-keys": 4800000,
        "consistency-check-method": "mvcc",
        "enable-region-bucket": true,
        "region-bucket-size": "128MiB",
        "region-size-threshold-for-approximate": "384MiB",
        "region-bucket-merge-size-ratio": 0.33
    },

    "max-merge-region-keys": 1000000,
    "max-merge-region-size": 100,

Some large regions did not split as expected
eGUffDgG8s
/pd-ctl region topsize 1
image

tonyxuqqi:

Issue Number: Close #12597

What's Changed:

```commit-message
1) originally the split by keys only support CheckPolicy::Scan. Add CheckPolicy::Approximate support. 
2) refine the way to use approximate size for split---calculate the actual split count per region size instead of use batch_split_limit
3) refine the bucket_version logic to make it monotonic----basically keep the region_buckets in last_region_buckets on split/merge 
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- None

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/coprocessor/config.rs
+++ b/components/raftstore/src/coprocessor/config.rs
@@ -89,7 +89,7 @@ impl Default for Config {
             perf_level: PerfLevel::Uninitialized,
             enable_region_bucket: false,
             region_bucket_size: DEFAULT_BUCKET_SIZE,
-            region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * 4,
+            region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * BATCH_SPLIT_LIMIT,
             region_bucket_merge_size_ratio: DEFAULT_REGION_BUCKET_MERGE_SIZE_RATIO,
         }
     }
--- a/components/raftstore/src/coprocessor/config.rs
+++ b/components/raftstore/src/coprocessor/config.rs
@@ -89,7 +89,7 @@ impl Default for Config {
             perf_level: PerfLevel::Uninitialized,
             enable_region_bucket: false,
             region_bucket_size: DEFAULT_BUCKET_SIZE,
-            region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * BATCH_SPLIT_LIMIT,
+            region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * BATCH_SPLIT_LIMIT / 2 * 3,
             region_bucket_merge_size_ratio: DEFAULT_REGION_BUCKET_MERGE_SIZE_RATIO,
         }
     }
--- a/components/raftstore/src/coprocessor/split_check/mod.rs
+++ b/components/raftstore/src/coprocessor/split_check/mod.rs
@@ -135,7 +135,7 @@ pub fn calc_split_keys_count(
     split_threshold: u64,
     batch_split_limit: u64,
 ) -> u64 {
-    let actual_split_limit = if count_per_region % split_threshold == 0 {
+    let actual_split_limit = if count_per_region % split_threshold < split_threshold / 2 {
         count_per_region / split_threshold
     } else {
         count_per_region / split_threshold + 1
--- a/components/raftstore/src/coprocessor/split_check/mod.rs
+++ b/components/raftstore/src/coprocessor/split_check/mod.rs
@@ -135,11 +135,10 @@ pub fn calc_split_keys_count(
     split_threshold: u64,
     batch_split_limit: u64,
 ) -> u64 {
-    let actual_split_limit = if count_per_region % split_threshold < split_threshold / 2 {
-        count_per_region / split_threshold
-    } else {
-        count_per_region / split_threshold + 1
-    };
     // split keys count is split count - 1
-    std::cmp::min(actual_split_limit.saturating_sub(1), batch_split_limit)
+    // if the count_per_region % split_threshold < split_threshold / 2, then split count is count_per_region/split_threshold
+    std::cmp::min(
+        count_per_region.saturating_sub(split_threshold / 2) / split_threshold,
+        batch_split_limit,
+    )
 }
--- a/components/raftstore/src/coprocessor/split_check/mod.rs
+++ b/components/raftstore/src/coprocessor/split_check/mod.rs
@@ -133,14 +133,24 @@ impl<'a, E> Host<'a, E> {
 pub fn calc_split_keys_count(
     count_per_region: u64,
     split_threshold: u64,
+    max_count_per_region: u64,
     batch_split_limit: u64,
 ) -> u64 {
+    if count_per_region < max_count_per_region {
+        return 0;
+    }
+
     // split keys count is split count - 1
-    // e.g. if split_threshold is 60, count_per_region is 60, return 0
-    //      if split_threshold is 60, count_per_region is 61, return 1
-    //      if split_threshold is 60, count_per_region is 121, return 2
+    // e.g. when split_threshold is 60 & max_count_per_region is 90
+    //      if count_per_region is 60, return 0
+    //      if count_per_region is 90, return 1
+    //      if count_per_region is 121, return 1
+    //      if count_per_region is 183, return 2
     std::cmp::min(
-        ((count_per_region as f64 / split_threshold as f64).ceil() as u64).saturating_sub(1),
+        std::cmp::max(
+            ((count_per_region as f64 / split_threshold as f64 + 0.5) as u64).saturating_sub(1),
+            count_per_region / max_count_per_region,
+        ),
         batch_split_limit,
     )
 }
--- a/components/raftstore/src/coprocessor/split_check/mod.rs
+++ b/components/raftstore/src/coprocessor/split_check/mod.rs
@@ -128,3 +128,15 @@ impl<'a, E> Host<'a, E> {
         self.cfg.region_bucket_size.0
     }
 }
+
+#[inline]
+pub fn calc_split_keys_count(count_per_region: u64, split_threshold: u64, batch_split_limit: u64) -> u64 {
+    let actual_split_limit = if count_per_region % split_threshold == 0 {
+        count_per_region / split_threshold
+    } else {
+        count_per_region / split_threshold + 1
+    };
+    
+    // split keys count is split count - 1
+    std::cmp::min(actual_split_limit - 1, batch_split_limit)
+}
--- a/components/raftstore/src/coprocessor/split_check/mod.rs
+++ b/components/raftstore/src/coprocessor/split_check/mod.rs
@@ -130,13 +130,16 @@ impl<'a, E> Host<'a, E> {
 }
 
 #[inline]
-pub fn calc_split_keys_count(count_per_region: u64, split_threshold: u64, batch_split_limit: u64) -> u64 {
+pub fn calc_split_keys_count(
+    count_per_region: u64,
+    split_threshold: u64,
+    batch_split_limit: u64,
+) -> u64 {
     let actual_split_limit = if count_per_region % split_threshold == 0 {
         count_per_region / split_threshold
     } else {
         count_per_region / split_threshold + 1
     };
-    
     // split keys count is split count - 1
     std::cmp::min(actual_split_limit - 1, batch_split_limit)
 }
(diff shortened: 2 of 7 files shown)


BusyJay:
I don't think it's correct. Size approximate split is only done when it exceeds a multiplier limit. If key approximate limit is not implemented in the past, I suggest to not land it in v6.1.

/cc @Connor1996, do you still remember why it's not implemented in the past?

components/raftstore/src/coprocessor/split_check/size.rs
```diff
@@ -224,7 +224,7 @@ pub fn get_region_approximate_size(
 }
 
 /// Get region approximate split keys based on default, write and lock cf.
-fn get_approximate_split_keys(
+pub fn get_approximate_split_keys(
```
Connor1996, line 227: it's based on size, shouldn't use it.

Connor1996:
> I don't think it's correct. Size approximate split is only done when it exceeds a multiplier limit. If key approximate limit is not implemented in the past, I suggest to not land it in v6.1.
> 
> /cc @Connor1996, do you still remember why it's not implemented in the past?

Seems, just not implement it with no reason

components/raftstore/src/coprocessor/split_check/size.rs
tonyxuqqi: > it's based on size, shouldn't use it.
to approximately split region by keys, there's no way to use key count.

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -128,3 +128,15 @@ impl<'a, E> Host<'a, E> {
...
+pub fn calc_split_keys_count(count_per_region: u64, split_threshold: u64, batch_split_limit: u64) -> u64 {
+    let actual_split_limit = if count_per_region % split_threshold == 0 {
+        count_per_region / split_threshold
+    } else {
+        count_per_region / split_threshold + 1
+    };
+    
+    // split keys count is split count - 1
+    std::cmp::min(actual_split_limit - 1, batch_split_limit)
+}
```
BusyJay, line 142: Missing EOL.

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -128,3 +128,15 @@ impl<'a, E> Host<'a, E> {
         self.cfg.region_bucket_size.0
     }
 }
+
+#[inline]
+pub fn calc_split_keys_count(count_per_region: u64, split_threshold: u64, batch_split_limit: u64) -> u64 {
+    let actual_split_limit = if count_per_region % split_threshold == 0 {
```
BusyJay, line 134: I think it's `cmp::min(count_per_region.saturating_sub(1) / split_threshold, batch_split_limit)`.

components/raftstore/src/store/peer.rs
```diff
@@ -3165,8 +3165,10 @@ where
         // Reset delete_keys_hint and size_diff_hint.
         self.delete_keys_hint = 0;
         self.size_diff_hint = 0;
-        self.region_buckets = None;
-        self.last_region_buckets = None;
+        if self.region_buckets.is_some() {
```
BusyJay, line 3168: This should be extracted as a method.

tests/integrations/raftstore/test_split_region.rs
```diff
@@ -1134,12 +1134,63 @@ fn test_refresh_region_bucket_keys() {
...
+    let bucket_version5 = cluster.refresh_region_bucket_keys(
         &region,
         buckets,
         Some(bucket_ranges),
         Some(expected_buckets.clone()),
     );
+
+    assert_eq!(bucket_version5, bucket_version4 + 1);
+
+    {
```
BusyJay, line 1146: This is not the TiKV test style.

components/raftstore/src/coprocessor/config.rs
```diff
@@ -89,7 +89,7 @@ impl Default for Config {
             perf_level: PerfLevel::Uninitialized,
             enable_region_bucket: false,
             region_bucket_size: DEFAULT_BUCKET_SIZE,
-            region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * 4,
+            region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * BATCH_SPLIT_LIMIT,
```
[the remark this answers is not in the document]
tonyxuqqi: good catch.

components/raftstore/src/coprocessor/split_check/mod.rs
tonyxuqqi: good catch on potential overflow.  But the "if block" is still needed

components/raftstore/src/store/peer.rs
tonyxuqqi: good point

tests/integrations/raftstore/test_split_region.rs
tonyxuqqi: ok

components/raftstore/src/coprocessor/split_check/mod.rs
BusyJay: Why?

components/raftstore/src/coprocessor/split_check/mod.rs
tonyxuqqi: fixed

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -128,3 +128,18 @@ impl<'a, E> Host<'a, E> {
...
     }
 }
+
+#[inline]
+pub fn calc_split_keys_count(
+    count_per_region: u64,
+    split_threshold: u64,
+    batch_split_limit: u64,
+) -> u64 {
+    let actual_split_limit = if count_per_region % split_threshold < split_threshold / 2 {
```
BusyJay, line 138: I think it should be `cmp::min(count_per_region.saturating_sub(split_threshold / 2) / split_threshold, batch_split_limit)`.

tonyxuqqi: done

components/raftstore/src/coprocessor/split_check/mod.rs
tonyxuqqi: ok. used the suggestion below.

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -128,3 +128,17 @@ impl<'a, E> Host<'a, E> {
         self.cfg.region_bucket_size.0
     }
 }
+
+#[inline]
+pub fn calc_split_keys_count(
```
BusyJay, line 133: Any test case?

BusyJay (commented):
If the bucket version overflow, any method to detect such failure?

components/raftstore/src/coprocessor/split_check/mod.rs
tonyxuqqi: the new added test_split_check_by_approximate_keys will call this method.

tonyxuqqi:
> If the bucket version overflow, any method to detect such failure?

it won't overflow, given the tick's running interval at 5s or 10s. It would require 1000 more years to overflow.

BusyJay:
> it won't overflow, given the tick's running interval at 5s or 10s. It would require 1000 more years to overflow.

I have seen it was set to hundreds of milliseconds in a production cluster. Though it's rare.

components/raftstore/src/coprocessor/split_check/mod.rs
BusyJay: I suggest to add a unit test case instead of depending on other cases.

tonyxuqqi: ok.

tonyxuqqi: > If the bucket version overflow, any method to detect such failure?

It's impossible to overflow practically,  even with 0.1s tick. Because to trigger refresh the buckets, it will need the enough write traffic to accumulate enough new data. 
But let's said the same term runs for 1000 years and it does overflow, then it will crash.  The restart will recover from it. 
On the other hand, if the region changes the leader every 0.1s tick (the region's term changed every 0.1s), then after  more than 13 years, the term << 32 will be back to 0. But practically it will be impossible.

BusyJay: Why do you think it's impossible for a cluster to run for 13 years? 🤣

components/raftstore/src/coprocessor/split_check/mod.rs
```diff
@@ -128,3 +128,29 @@ impl<'a, E> Host<'a, E> {
...
+
+    // split keys count is split count - 1
+    // e.g. when split_threshold is 60 & max_count_per_region is 90
+    //      if count_per_region is 60, return 0
+    //      if count_per_region is 90, return 1
+    //      if count_per_region is 121, return 1
+    //      if count_per_region is 183, return 2
+    std::cmp::min(
+        std::cmp::max(
+            ((count_per_region as f64 / split_threshold as f64 + 0.5) as u64).saturating_sub(1),
```
BusyJay, line 151: You mean `((count_per_region as f64 / split_threshold as f64).round() as u64).saturating_sub(1)`

BusyJay (approved):
Rest LGTM

components/raftstore/src/coprocessor/split_check/size.rs
```diff
@@ -51,6 +51,10 @@ where
     E: KvEngine,
 {
     fn on_kv(&mut self, _: &mut ObserverContext<'_>, entry: &KeyEntry) -> bool {
+        // If there's no need to check region split, skip it.
+        // Otherwise, the region whose keys > max region keys will not be splitted when batch_split_limit is zero,
+        // because eventually "over_limit && self.current_size + self.split_size >= self.max_size"
+        // will return true.
```
BusyJay, line 57: Should add a TODO to separate the bucket checker from split checker.

BusyJay (approved):

5kbpers (approved):

PR #12629, opened 2022-05-24
pd_client: Do not reconnect on store tombstone error

### LINKED ISSUE #12506: PD client keeps reconnecting on error StoreTombstone

master

No matter

### Steps to reproduce

Execute pd-ctl unsafe remove-failed-stores, and find 

### What did you expect?
No PD reconnection

### What did happened?\
PD client reconnects a lot of times
image
image

Connor1996:

Issue Number: close #12506 

What's Changed:

```commit-message
do not reconnect PD on store tombstone error
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Fix the issue that keep retrying reconnecting PD when encounter store tombstone error
```

### DIFF
--- a/components/pd_client/src/errors.rs
+++ b/components/pd_client/src/errors.rs
@@ -7,8 +7,6 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum Error {
-    #[error("{0}")]
-    Io(#[from] std::io::Error),
     #[error("cluster {0} is already bootstrapped")]
     ClusterBootstrapped(u64),
     #[error("cluster {0} is not bootstrapped")]
@@ -29,10 +27,22 @@ pub enum Error {
 
 pub type Result<T> = result::Result<T, Error>;
 
+impl Error {
+    pub fn retryable(&self) -> bool {
+        match self {
+            Error::Grpc(_) | Error::Other(_) | Error::ClusterNotBootstrapped(_) => true,
+            Error::RegionNotFound(_)
+            | Error::StoreTombstone(_)
+            | Error::GlobalConfigNotFound(_)
+            | Error::ClusterBootstrapped(_)
+            | Error::Incompatible => false,
+        }
+    }
+}
+
 impl ErrorCodeExt for Error {
     fn error_code(&self) -> ErrorCode {
         match self {
-            Error::Io(_) => error_code::pd::IO,
             Error::ClusterBootstrapped(_) => error_code::pd::CLUSTER_BOOTSTRAPPED,
             Error::ClusterNotBootstrapped(_) => error_code::pd::CLUSTER_NOT_BOOTSTRAPPED,
             Error::Incompatible => error_code::pd::INCOMPATIBLE,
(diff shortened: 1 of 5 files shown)


components/resolved_ts/src/advance.rs
```diff
@@ -1,389 +0,0 @@
-// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
```
5kbpers, line 1: Why remove this?

Connor1996: PTAL again, it's removed by mistake

components/resolved_ts/src/advance.rs
```diff
@@ -234,7 +234,10 @@ pub async fn region_resolved_ts_store(
                 let client =
                     get_tikv_client(to_store, pd_client, security_mgr, env, tikv_clients.clone())
                         .await
-                        .map_err(|e| (to_store, true, format!("[get tikv client] {}", e)))?;
+                        .map_err(|e| {
+                            let reconnect = !matches!(e, pd_client::Error::StoreTombstone(..));
```
BusyJay, line 238: Why not reused `should_not_retry`?

components/pd_client/src/util.rs
```diff
@@ -441,8 +441,13 @@ where
     fn should_not_retry(resp: &Result<Resp>) -> bool {
         match resp {
             Ok(_) => true,
-            // Error::Incompatible is returned by response header from PD, no need to retry
-            Err(Error::Incompatible) => true,
+            // these errors are not caused by network, no need to retry
+            Err(
+                Error::Incompatible
```
BusyJay, line 446: How about make the matching as a method of `Error`?

components/pd_client/src/errors.rs
```diff
@@ -29,10 +27,22 @@ pub enum Error {
 
 pub type Result<T> = result::Result<T, Error>;
 
+impl Error {
+    pub fn retryable(&self) -> bool {
+        match self {
+            Error::Grpc(_) | Error::Other(_) | Error::ClusterNotBootstrapped(_) => true,
+            Error::RegionNotFound(_)
```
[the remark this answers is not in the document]
Connor1996: call `get_region_and_leader` but the region heartbeat hasn't been received yet

components/pd_client/src/errors.rs
```diff
@@ -29,10 +27,22 @@ pub enum Error {
...
+            | Error::StoreTombstone(_)
+            | Error::GlobalConfigNotFound(_)
```
BusyJay, line 36: Why this doesn't need retry?

Connor1996: the config should be wrong, it doesn't the result after retry
/cc @glorv

glorv: `GlobalConfigNotFound` is returned when the target key does not exist. So I think it is reasonable to not retry this error as the retry is likely to return the same error.

Though, In the current implement, the tikv side code just ignores this error with an error log, so there is no chance to propgate this error the outer level error handling code.

https://github.com/tikv/tikv/blob/c461cd829fc6e06f071441f70645c8403e7453a5/components/pd_client/src/client.rs#L309-L318

BusyJay (approved):

Issue #12664, opened 2022-05-26
raftstore: tikv panicked with `no entry found for key`


zyguan:
### What version of TiKV are you using?

master @ c3c9707faed1cf9ef9023f820dd6d1b28285e85f

### What operating system and CPU are you using?

kubernetes

### Steps to reproduce

Run jepsen monotonic with the following options
```
{:auto-retry-limit :default,
 :ssh-private-key "jepsen.pem",
 :isolation :repeatable-read,
 :concurrency 15,
 :predicate-read false,
 :tarball-url
 "http://minio.pingcap.net:9000/tp-team/tests/jepsen/tidb-master-linux-amd64.tar.gz",
 :txn-mode "mixed",
 :force-reinstall true,
 :update-in-place false,
 :ssh
 {:username "root",
  :password "root",
  :strict-host-key-checking false,
  :private-key-path "jepsen.pem"},
 :nemesis-long-recovery false,
 :follower-read false,
 :nemesis
 {:interval 10,
  :schedules true,
  :partition-pd-leader true,
  :partition-half true,
  :partition-ring true},
 :nodes
 ["node-0.node-peer.jepsen-tps-810956-1-62"
  "node-1.node-peer.jepsen-tps-810956-1-62"
  "node-2.node-peer.jepsen-tps-810956-1-62"
  "node-3.node-peer.jepsen-tps-810956-1-62"
  "node-4.node-peer.jepsen-tps-810956-1-62"],
 :test-count 1,
 :read-lock "FOR UPDATE",
 :use-index true,
 :os #object[tidb.core.Image 0xe62319f "tidb.core.Image@e62319f"],
 :time-limit 300,
 :recovery-time 10,
 :version "master",
 :auto-retry :default,
 :workload :monotonic,
 :init-txn-sql nil,
 :init-sql
 ("set @@tidb_enable_mutation_checker=1, @@tidb_txn_assertion_level=strict")}
```

### What did you expect?

Test passed.

### What did happened?

Test failed because tikv-server panicked. The full logs can be accessed [here](http://rms.pingcap.net:31714/artifacts/testground/plan-exec-810956/plan-exec-810956-3671234100/archive).
```
[...]

PR #12669, opened 2022-05-26
add logs for entry missing

BusyJay:
Issue Number: Ref #12664 

What's Changed:

```commit-message
...So we can know what is actually happening.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/store.rs
+++ b/components/raftstore/src/store/fsm/store.rs
@@ -2035,11 +2035,20 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
 
         let mut is_overlapped = false;
         let mut regions_to_destroy = vec![];
-        for (_, id) in meta.region_ranges.range((
+        for (key, id) in meta.region_ranges.range((
             Excluded(data_key(msg.get_start_key())),
             Unbounded::<Vec<u8>>,
         )) {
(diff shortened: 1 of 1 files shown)


zhangjinpeng1987 (approved):
LGTM

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2035,11 +2035,20 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
 
         let mut is_overlapped = false;
         let mut regions_to_destroy = vec![];
-        for (_, id) in meta.region_ranges.range((
+        for (key, id) in meta.region_ranges.range((
             Excluded(data_key(msg.get_start_key())),
             Unbounded::<Vec<u8>>,
         )) {
-            let exist_region = &meta.regions[id];
+            let exist_region = match meta.regions.get(id) {
```
5kbpers, line 2042: We have many other places using `meta.regions[..]`. How about extracting to a function for it?

BusyJay: Thought about it, but those places were less frequently visited and context everywhere might be too verbose.

5kbpers (approved):

BornChanger (approved):

PR #12678, opened 2022-05-27
raftstore: use approximate size to generate bucket unless split region check is needed and scan is used for it.

tonyxuqqi:

Issue Number: ref #12597

What's Changed:

```commit-message
When bucket is enabled,  using CheckPolicy::Scan would lead to large amount of read IO after tikv restart. Before this PR, the Scan is used unless the region size reaches 1.5G, which is very rare for a 96 ~ 256MB's region-split-size. After this change, generating bucket won't introduce new scan unless the scan is necessary for splitting region. This can significantly reduce the read IO.
Also refine the logic for the fix of 12597. 
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/coprocessor/config.rs
+++ b/components/raftstore/src/coprocessor/config.rs
@@ -50,6 +50,8 @@ pub struct Config {
     pub region_bucket_size: ReadableSize,
     // region size threshold for using approximate size instead of scan
     pub region_size_threshold_for_approximate: ReadableSize,
+    #[online_config(skip)]
+    pub prefer_gen_bucket_by_approximate: bool,
     // ratio of region_bucket_size. (0, 0.5)
     // The region_bucket_merge_size_ratio * region_bucket_size is threshold to merge with its left neighbor bucket
     pub region_bucket_merge_size_ratio: f64,
@@ -91,6 +93,7 @@ impl Default for Config {
             region_bucket_size: DEFAULT_BUCKET_SIZE,
             region_size_threshold_for_approximate: DEFAULT_BUCKET_SIZE * BATCH_SPLIT_LIMIT / 2 * 3,
             region_bucket_merge_size_ratio: DEFAULT_REGION_BUCKET_MERGE_SIZE_RATIO,
+            prefer_gen_bucket_by_approximate: true,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


components/raftstore/src/coprocessor/config.rs
```diff
@@ -50,6 +50,8 @@ pub struct Config {
     pub region_bucket_size: ReadableSize,
     // region size threshold for using approximate size instead of scan
     pub region_size_threshold_for_approximate: ReadableSize,
+    #[online_config(skip)]
+    pub prefer_gen_bucket_by_approximate: bool,
```
5kbpers, line 54: I don't think we need a switch here if the performance was verified.

components/raftstore/src/coprocessor/split_check/size.rs
```diff
@@ -51,13 +51,6 @@ where
     E: KvEngine,
 {
     fn on_kv(&mut self, _: &mut ObserverContext<'_>, entry: &KeyEntry) -> bool {
-        // If there's no need to check region split, skip it.
```
BusyJay, line 54: Why delete?

tonyxuqqi: because we want to leverage the scan (if it happened) to check if the region needs the split. So no need to skip it.

tonyxuqqi: The original bug that deleted change fixed no longer exists because the batch_split_limit won't be 0 anymore.

components/raftstore/src/coprocessor/config.rs
tonyxuqqi: > I don't think we need a switch here if the performance was verified.

By adding an undocumented config it's useful for performance troubleshooting in the future to verify if imbalanced buckets play a role in some scenario.  We need a way to generate absolutely accurate buckets and this flag can help do that.

BusyJay, line 54: `prefer_approximate_bucket` seems more concise.

BusyJay (approved):

components/raftstore/src/coprocessor/config.rs
tonyxuqqi: Good point.

5kbpers (approved):

Issue #5049, opened 2019-07-09
Line number isn't displayed in backtrace for tests


Connor1996:
## Bug Report

In https://github.com/tikv/tikv/pull/4689, it turns debuginfo completely off for the dev, test, and bench profiles. But we set `[profile.test] debug=1` to keep the line information for tests. 

When I running a test, the backtrace shows without line number for codes of `libtikv`, see https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/tikv_ghpr_test/runs/6483/nodes/110/steps/660/log/?start=0
```
 10:     0x55ea2bb7b202 - core::result::unwrap_failed::hf29614f40702c601
  11:     0x55ea2bbe04dc - core::result::Result<T,E>::unwrap::h2860fd56f6b605d4
  12:     0x55ea2a5291e7 - tikv::raftstore::store::fsm::batch::BatchSystem<N,C>::shutdown::h1b0a5ecdcf7dc7c1
  13:     0x55ea2abcb5e8 - tikv::raftstore::store::fsm::store::RaftBatchSystem::shutdown::h0a5eed9d8a5f9227
  14:     0x55ea29eecd12 - tikv::server::node::Node<C>::stop_store::h81766c31dce70bfb
  15:     0x55ea29ef395d - tikv::server::node::Node<C>::stop::ha0bec93fd452b2e0
  16:     0x55ea29fa390f - <test_raftstore::node::NodeCluster as test_raftstore::cluster::Simulator>::stop_node::h776f806e9624f6fd
  17:     0x55ea29c0c934 - test_raftstore::cluster::Cluster<T>::stop_node::h2b6c255358df1246
                               at /home/jenkins/tikv-master/build/components/test_raftstore/src/cluster.rs:215
  18:     0x55ea29c0c356 - test_raftstore::cluster::Cluster<T>::shutdown::hfed4a9d244b48252
                               at /home/jenkins/tikv-master/build/components/test_raftstore/src/cluster.rs:523
  19:     0x55ea29c0d006 - <test_raftstore::cluster::Cluster<T> as core::ops::drop::Drop>::drop::h2ab1b063808ba94f
                               at /home/jenkins/tikv-master/build/components/test_raftstore/src/cluster.rs:1019
  20:     0x55ea29b7fd71 - core::ptr::real_drop_in_place::h56b2740450cdc9cd
[...]

PR #12708, opened 2022-05-31
*: optimize debug build

### LINKED ISSUE #12707: optimize size of target directory
## Development Task

Currently, debug build will generate about 40GiB data in target. And the generated binaries are also very large, and occupies the CI cache very quickly.

We can optimize the directory size by disabling debug info.

BusyJay:
Issue Number: Close #12707 

What's Changed:

```commit-message
This PR optimize debug build by disabling all debuginfo excepts tests
itself. So that the generated artifacts will be smaller and also speed
up compile time a little.
```

My local tests show that target directory changes from 45GiB to about 9.2GiB. And build time on Linux VM (8 vcores32GiB memory 2200MHz) is changing from 17m22s to 13m46s. The improvement probably comes from writing less data to disk.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -277,9 +277,25 @@ members = [
 ]
 default-members = ["cmd/tikv-server", "cmd/tikv-ctl"]
 
+[profile.dev.package.grpcio-sys]
+debug = false
+opt-level = 1
+
+[profile.dev.package.librocksdb_sys]
+debug = false
+opt-level = 1
+
+[profile.dev.package.libtitan_sys]
+debug = false
+opt-level = 1
+
+[profile.dev.package.tests]
+debug = 1
+opt-level = 1
+
 [profile.dev]
 opt-level = 0
-debug = true
+debug = 0
 codegen-units = 4
 lto = false
 incremental = true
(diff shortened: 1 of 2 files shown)


BusyJay:
This is also a workaround for #4711.

Cargo.toml
```diff
@@ -277,9 +277,25 @@ members = [
...
+opt-level = 1
+
+[profile.dev.package.tests]
+debug = 1
+opt-level = 1
+
 [profile.dev]
 opt-level = 0
-debug = true
+debug = 0
```
Connor1996, line 298: I think it's handy for debugging, let's make it to 1 as least? It's enabled on purpose before https://github.com/tikv/tikv/issues/5049

Connor1996 (commented):
Should update the `build issues` part in `CONTRIBUTING.md` as well

Cargo.toml
BusyJay: > I think it's handy for debugging

I set the tests to 1 for the exact purpose. So that we can see the exact panic point in tests. Setting global to 1 will make the target directory grow to 20GiB.

Connor1996 (approved):
LGTM

sticnarf (approved):

PR #12732, opened 2022-06-01
server: further simplify metrics

glorv:

Issue Number: ref #12355

What's Changed:

```commit-message
```
This pr adds new options to support further simplifying TiKV's metrics from two dimensions and each provides 3 different simplification levels:
- Metrics Level (like log level, a higher level outputs fewer data). We split all the metrics into two groups, "high-priority" metrics(explicitly registry with a separated registry) and normal metrics (use the default registry). 
 - All. the default level, returns all data.
 - ReduceFrequency.  Always return the high priority metrics but only return the normal metrics by every other request. In other words, all the normal metrics' fetch frequency is reduced by half. 
 - OnlyHighPriority. Only return the high-priority metrics.
- Metrics Compact Level. This dimension describes how to deal with each metric data.
 - NoCompact. the default level, do nothing to the metrics data.
 - LoseLess. As implemented in #12602, do not return counters with 0 value and histograms will 0 samples as they are the same as the *default* value, so we reduce the data with no information loss.
 - Aggressive. Try to reduce the metrics as much as possible. In addition to level `LoseLess`, this PR will also reduce histogram vector type by filtering all histograms whose sample count < max sample count * 0.01. We may also do further operations(in the future), e.g. we can apply the ideas in #12417 to compress some data types when we come up with some consensus.

NOTE: for development conveniency, the "high priority" metrics are the union of all metrics that exist in one of tikv_summary.json, tikv_fast_tun.json, performance_read.json, performance_write.json, tikv_trobule_shooting.json, alertmanager/*.json. There may be some important metrics that are missing.

This PR introduces two new config fields: `server.metrics-level` and `server.metrics-compact-level` to control these two dimensions. 

Test result: The following table shows the data size return with combining of different `server.metrics-level` and `server.metrics-compact-level`. (NOTE: the value if server.metrics-level = 1 is the expected avg value.)

| Metrics Level\Metrics Compact Level | 0 | 1 | 2 |
| --- | ---| --- | --- | 
| 0 |  1407K | 521K/37.0% | 400K/28.4% |
| 1 | 1037K/73.7%  | 377K/26.8% | 274K/19.5% |
| 2 | 670K/47.6% | 234K/16.6% | 148K/10.5% |

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/engine_rocks/src/rocks_metrics.rs
```diff
@@ -1219,66 +1224,74 @@ lazy_static! {
...
     ).unwrap();
     pub static ref STORE_ENGINE_CACHE_EFFICIENCY: EngineTickerMetrics =
         auto_flush_from!(STORE_ENGINE_CACHE_EFFICIENCY_VEC, EngineTickerMetrics);
 
-    pub static ref STORE_ENGINE_MEMTABLE_EFFICIENCY_VEC: IntCounterVec = register_int_counter_vec!(
+    pub static ref STORE_ENGINE_MEMTABLE_EFFICIENCY_VEC: IntCounterVec = register_int_counter_vec_with_registry!(
         "tikv_engine_memtable_efficiency",
         "Hit and miss of memtable",
-        &["db", "type"]
+        &["db", "type"],
```
zhangjinpeng1987, line 1239: Seems we rarely using this metric to diagnose issues.

components/engine_rocks/src/rocks_metrics.rs
```diff
@@ -1219,66 +1224,74 @@ lazy_static! {
...
     ).unwrap();
     pub static ref STORE_ENGINE_WRITE_SERVED: EngineTickerMetrics =
         auto_flush_from!(STORE_ENGINE_WRITE_SERVED_VEC, EngineTickerMetrics);
 
-    pub static ref STORE_ENGINE_BLOOM_EFFICIENCY_VEC: IntCounterVec = register_int_counter_vec!(
+    pub static ref STORE_ENGINE_BLOOM_EFFICIENCY_VEC: IntCounterVec = register_int_counter_vec_with_registry!(
         "tikv_engine_bloom_efficiency",
         "Efficiency of rocksdb's bloom filter",
-        &["db", "type"]
+        &["db", "type"],
```
zhangjinpeng1987, line 1266: Rarely used when diagnosing issues.

components/engine_rocks/src/rocks_metrics.rs
```diff
@@ -1307,16 +1320,18 @@ lazy_static! {
...
-        linear_buckets(0.0, 1.0, 7).unwrap()
+        linear_buckets(0.0, 1.0, 7).unwrap(),
+        HIGH_PRIORITY_REGISTRY
     ).unwrap();
-    pub static ref STORE_ENGINE_LOCATE_VEC: IntCounterVec = register_int_counter_vec!(
+    pub static ref STORE_ENGINE_LOCATE_VEC: IntCounterVec = register_int_counter_vec_with_registry!(
         "tikv_engine_locate",
         "Number of calls to seek/next/prev",
-        &["db", "type"]
+        &["db", "type"],
```
zhangjinpeng1987, line 1333: Rarely used when diagnosing issues.

components/raftstore/src/store/metrics.rs
```diff
@@ -313,34 +319,39 @@ lazy_static! {
...
+            exponential_buckets(0.00001, 2.0, 26).unwrap(),
+            HIGH_PRIORITY_REGISTRY
         ).unwrap();
     pub static ref STORE_WF_WRITE_END_DURATION_HISTOGRAM: Histogram =
-        register_histogram!(
+        register_histogram_with_registry!(
             "tikv_raftstore_store_wf_write_end_duration_seconds",
             "Bucketed histogram of proposals' write db end duration.",
-            exponential_buckets(0.00001, 2.0, 26).unwrap()
+            exponential_buckets(0.00001, 2.0, 26).unwrap(),
```
zhangjinpeng1987, line 353: What is the difference between "write db end duration" and "write kv db end duration"?

components/tikv_util/src/metrics/mod.rs
```diff
@@ -36,16 +46,63 @@ mod metrics_reader;
...
+
+static METRICS_COMPACT_LEVZEL: AtomicU32 = AtomicU32::new(MetricsCompactLevel::NoCompact as u32);
+static METRICS_LEVEL: AtomicU32 = AtomicU32::new(MetricsLevel::All as u32);
+
+// reduce the return frequency of normal metrics factor
+const NORMAL_METRICS_REDUCE_FACTOR: u64 = 2;
+static METRICS_REQUEST_COUNTER: AtomicU64 = AtomicU64::new(0);
+
+pub fn set_metrics_compact_level(level: MetricsCompactLevel) {
+    METRICS_COMPACT_LEVZEL.store(level as u32, Ordering::Relaxed);
```
zhangjinpeng1987, line 62: maybe Ordering::Release?

components/tikv_util/src/metrics/mod.rs
```diff
@@ -36,16 +46,63 @@ mod metrics_reader;
...
+}
+
+pub fn get_metrics_compact_level() -> MetricsCompactLevel {
+    METRICS_COMPACT_LEVZEL.load(Ordering::Relaxed).into()
```
zhangjinpeng1987, line 66: maybe Ordering::Acquire?

components/tikv_util/src/metrics/mod.rs
```diff
@@ -36,16 +46,63 @@ mod metrics_reader;
...
+lazy_static! {
+    // the registry for high priority metrics.
+    pub static ref HIGH_PRIORITY_REGISTRY: Registry = Registry::new();
+}
+
+static METRICS_COMPACT_LEVZEL: AtomicU32 = AtomicU32::new(MetricsCompactLevel::NoCompact as u32);
+static METRICS_LEVEL: AtomicU32 = AtomicU32::new(MetricsLevel::All as u32);
+
+// reduce the return frequency of normal metrics factor
+const NORMAL_METRICS_REDUCE_FACTOR: u64 = 2;
```
zhangjinpeng1987, line 58: Can the factor changed?

components/tikv_util/src/metrics/mod.rs
```diff
@@ -59,7 +116,20 @@ pub fn dump_to(w: &mut impl Write, should_simplify: bool) {
             MetricType::COUNTER => {
                 metrics.retain(|m| m.get_counter().get_value() > 0.0);
             }
-            MetricType::HISTOGRAM => metrics.retain(|m| m.get_histogram().get_sample_count() > 0),
+            MetricType::HISTOGRAM => {
+                let threshold = if simplify_level == MetricsCompactLevel::LoseLess {
```
zhangjinpeng1987, line 120: The simplify_level is always `NoCompact` since L105 has checked.

components/tikv_util/src/metrics/mod.rs
```diff
@@ -36,16 +46,63 @@ mod metrics_reader;
...
+    }
+}
+
+pub fn dump_metrics_to(w: &mut impl Write, metric_families: Vec<MetricFamily>) {
+    let simplify_level = get_metrics_compact_level();
+
     let encoder = TextEncoder::new();
-    let metric_families = prometheus::gather();
-    if !should_simplify {
+    if simplify_level == MetricsCompactLevel::NoCompact {
```
zhangjinpeng1987, line 105: Why need NoCompact? seems this also may ignore high-priority-metrics when the simplify_level is other values?

src/server/config.rs
```diff
@@ -252,6 +261,8 @@ impl Default for Config {
             // Go tikv client uses 4 as well.
             forward_max_connections_per_address: 4,
             simplify_metrics: false,
+            metrics_compact_level: MetricsCompactLevel::NoCompact,
+            metrics_level: MetricsLevel::All,
```
zhangjinpeng1987, line 265: I'm a bit of confused by these two `level`, is it possible to unify these two parameters?

zhangjinpeng1987 (commented):
rest LGTM

components/tikv_util/src/metrics/mod.rs
glorv: As most of the function aggregation window size are 60s(some are 30s), thus, if the factor is bigger than 2, the charts in grafana will output abnormal output. A factor of 3 will result in abnormal output because some window contains 2 sample and others are only 1. Facotr >=4 will make most charts empty(in this case we can manually increase the aggregation window size.)  So, I'm not sure if there are scenario that this factor can be more than 2.

components/tikv_util/src/metrics/mod.rs
glorv: I originally thought the order is not essential here because we don't need that strong consistency. I will change to `Release` order to make it more  ergonomic.

components/tikv_util/src/metrics/mod.rs
glorv: `NoCompact` means just returning the original data, in this branch, it calls `encoder.encode` for all metrics at L106.

components/tikv_util/src/metrics/mod.rs
glorv: if the level is `NoCompact` it will return at L109.

src/server/config.rs
glorv: In my original design, I also combine these two configs into one and only provide 4 levels. @kevin-xianliu
 thinks it's better and more flexible to use two configs because they are two different dimensions.

kevin-xianliu: In my mind, there are two dimensions to reduce metrics traffic . One is adjusting the accuracy of metrics, the other is adjusting the total number of active metrics.  If we use one parameter to control above two dimensions, it might be hard to extend in the future?

kevin-xianliu: Although, I do think we can use clearer names instead of  metrics_compact_level and metrics_level.

zhangjinpeng1987: better to use a clearer name.

components/tikv_util/src/metrics/mod.rs
```diff
@@ -115,6 +186,100 @@ pub fn convert_record_pairs(m: HashMap<String, u64>) -> RecordPairVec {
...
+impl From<&ConfigValue> for MetricsCompactLevel {
+    fn from(c: &ConfigValue) -> Self {
+        c.clone().into()
+    }
+}
+
+// MetricsLevel controls how much metrics can be returned. A higher level means less data.
+#[derive(Serialize_repr, Deserialize_repr, Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(u32)]
+pub enum MetricsLevel {
```
kevin-xianliu, line 241: I think we would better not using both frequency and priority in the levels, which might make people confused. 
One possible proposal:
pub enum MetricLevel {
    None,     // return nothing (reserved)
    HighPriority,   // return high priority metrics
    MiddlePriority,   //return middle and high priority metrics (reserved)
    LowPriority,  // return low, middle and high priority metrics (reserved)
    ALL  // return all  metrics
}

components/tikv_util/src/metrics/mod.rs
```diff
@@ -115,6 +186,100 @@ pub fn convert_record_pairs(m: HashMap<String, u64>) -> RecordPairVec {
         .collect()
 }
 
+/// MetricsCompactLevel defines the level of compact metrics sample data, a higher level of
+/// means smaller data size and more information loss.
+#[derive(Serialize_repr, Deserialize_repr, Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(u32)]
+pub enum MetricsCompactLevel {
```
kevin-xianliu, line 193: MetricCompactPolicy might be a better choice ? For instance:
pub enum MetricCompactPolicy{
    NoCompaction = 0,   
    LoseLessCompaction = 1, 
    LossyCompaction = 2,
}

components/raftstore/src/store/metrics.rs
glorv: "write db end duration" = "write kv db end duration" + "write raft db end duration"

glorv:
@zhangjinpeng1987 @kevin-xianliu  PTAL again, thank you~

src/server/config.rs
```diff
@@ -171,8 +172,16 @@ pub struct Config {
 
     // whether to compact metrics or not.
     #[doc(hidden)]
+    // deprecated. use metrics_compact_level and metrics_level instead.
     pub simplify_metrics: bool,
 
+    // the level of metrics compression
```
glorv, line 178:
```suggestion
    // the policy of metrics compaction
```

glorv:
@zhangjinpeng1987 @kevin-xianliu  @BusyJay Could you please take a look again~

BusyJay:
I feel it's a bit complicated. @breezewish do you have ideas on this?

zhangjinpeng1987 (approved):
LGTM

kevin-xianliu (approved):

breezewish:
@BusyJay Thanks for the mentioning. I do feel it too complicated, but I would like to know some background first. I have raised my questions in the referenced issue https://github.com/tikv/tikv/issues/12355

PR #12756, opened 2022-06-03
engine_traits: refactor Engines to support both tablet and global kvdb

tonyxuqqi:

Issue Number: Ref #12772

What's Changed:

```commit-message
add tablet related function into Engines API.
The idea is that for region's KV data read/write, using tablet() instead of engines.kv regardless multirocks is enabled or not. 
```

### Check List 

Tests 

- No code (UT TBD)

Side effects

### Release note 

```release-note
None
```


components/engine_traits/src/engines.rs
```diff
@@ -1,21 +1,43 @@
...
     pub raft: R,
+    tablet_kv: Option<K>,
+    pub tablets: Option<Box<dyn TabletFactory<K> + Send>>,
 }
 
 impl<K: KvEngine, R: RaftEngine> Engines<K, R> {
     pub fn new(kv_engine: K, raft_engine: R) -> Self {
         Engines {
-            kv: kv_engine,
+            kv: kv_engine.clone(),
```
zhangjinpeng1987, line 23: Why clone() here?

components/engine_traits/src/engines.rs
```diff
@@ -1,21 +1,43 @@
...
+            tablet_kv: None,
+            tablets: None,
+        }
+    }
+
+    pub fn new_with_tablets(
+        kv_engine: K,
+        raft_engine: R,
+        tablet: K,
+        tablets: Box<dyn TabletFactory<K> + Send>,
```
zhangjinpeng1987, line 34: What's the difference between `tablet` and `tablets` here?

components/engine_traits/src/engines.rs
```diff
@@ -30,4 +52,47 @@ impl<K: KvEngine, R: RaftEngine> Engines<K, R> {
...
+    }
+}
+
+impl<K: Clone, R: Clone> Clone for Engines<K, R> {
+    #[inline]
+    fn clone(&self) -> Engines<K, R> {
+        Engines {
+            kv: self.kv.clone(),
+            raft: self.raft.clone(),
+            tablets: if let Some(tablets) = &self.tablets {
```
zhangjinpeng1987, line 77: you can use `tablets.map(|x| x.clone())` to simplify the code here.

zhangjinpeng1987:
@tonyxuqqi Please fix compile errors.

components/engine_traits/src/engines.rs
tonyxuqqi: good catch. It's not needed. The initial code needs it, but not now.

components/engine_traits/src/engines.rs
tonyxuqqi: Tablet is for single region's KV engine. Tablets is the factory to create any region's tablet. 
In split scenario, we need the ability to create a new region's tablet, so we need the tablets here.

components/engine_traits/src/engines.rs
tonyxuqqi: Thanks.

tonyxuqqi:
> @tonyxuqqi Please fix compile errors.

I fixed some test code, but this PR will need the dependent PR for complete compilation: https://github.com/tikv/rust-rocksdb/pull/693

components/engine_traits/src/engine.rs
```diff
@@ -63,5 +66,98 @@ pub trait KvEngine:
...
+    fn open_tablet_raw(&self, path: &Path, readonly: bool) -> Result<EK>;
+    fn create_root_db(&self) -> Result<EK>;
+    #[inline]
+    fn exists(&self, id: u64, suffix: u64) -> bool {
+        self.exists_raw(&self.tablet_path(id, suffix))
+    }
+    fn exists_raw(&self, path: &Path) -> bool;
+    fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf;
+    fn tablets_path(&self) -> PathBuf;
+    fn clone(&self) -> Box<dyn TabletFactory<EK> + Send>;
```
tabokie, line 92: Why not `pub trait TabletFactory<EK>: Clone`?

components/engine_traits/src/engine.rs
```diff
@@ -62,6 +65,86 @@ pub trait KvEngine:
 ///
 // It should be named as `EngineFactory` for consistency, but we are about to rename
 // engine to tablet, so always use tablet for new traits/types.
-pub trait TabletFactory<EK> {
-    fn create_tablet(&self) -> Result<EK>;
+pub trait TabletFactory<EK>: Clone {
```
BusyJay, line 68: This is not a good idea. It makes `Factory` not object safe. `Factory` is not in performance critical path, making it object safe can bring a lot of convenience.

components/engine_traits/src/engines.rs
```diff
@@ -31,3 +32,7 @@ impl<K: KvEngine, R: RaftEngine> Engines<K, R> {
         self.kv.sync()
     }
 }
+
+pub trait EnginesFactory<K, R> {
```
BusyJay, line 36: What's the purpose?

components/engine_traits/src/engine.rs
tonyxuqqi: @BusyJay agree. I realized it won't work for dyn TabletFactory

src/server/engine_factoryv2.rs
```diff
@@ -0,0 +1,275 @@
...
+        if let Some(s) = path.file_name().map(|s| s.to_string_lossy()) {
+            let mut split = s.split('_');
+            tablet_id = split.next().and_then(|s| s.parse().ok()).unwrap_or(0);
+            tablet_suffix = split.next().and_then(|s| s.parse().ok()).unwrap_or(1);
+        }
+        self.create_tablet(tablet_id, tablet_suffix)
+    }
+
+    #[inline]
+    fn create_root_db(&self) -> Result<RocksEngine> {
```
BusyJay, line 201: Perhaps we can change it to `create_share_db`. In v2, there will be no root db at all.

src/server/engine_factory.rs
```diff
@@ -171,8 +173,35 @@ impl<ER: RaftEngine> KvEngineFactory<ER> {
 
 impl<ER: RaftEngine> TabletFactory<RocksEngine> for KvEngineFactory<ER> {
     #[inline]
-    fn create_tablet(&self) -> Result<RocksEngine> {
+    fn create_root_db(&self) -> Result<RocksEngine> {
         let root_path = self.kv_engine_path();
-        self.create_tablet(&root_path)
+        let tablet = self.create_tablet(&root_path)?;
+        let mut root_db = self.inner.root_db.lock().unwrap();
```
BusyJay, line 179: I don't get it. For v1, it's pointless or even wrong to call create_root_db multiple times.

src/server/engine_factory.rs
```diff
@@ -171,8 +173,35 @@ impl<ER: RaftEngine> KvEngineFactory<ER> {
...
+        root_db.replace(tablet.clone());
+        Ok(tablet)
+    }
+
+    fn create_tablet(&self, _id: u64, _suffix: u64) -> Result<RocksEngine> {
```
BusyJay, line 184: All the following methods are useless for v1 factory as it should not be used at all.

src/server/engine_factoryv2.rs
BusyJay: share makes more sense to me as it means all regions share one kv db.

src/server/engine_factory.rs
tonyxuqqi: The whole point of abstraction is to unify the db creation API so that the caller don't need to care if it's a single db or tablet. Under the hood, the root_db (shared_db) is only created once in v1. 

Otherwise what's the point of Trait TabletFactory.

src/server/engine_factory.rs
tonyxuqqi: Once we finished the raftstore2, let's revisit them if they're truly useless, I will separate TabletFactory into two, one is for v1 and the other is for v2. 
Right now let's unify them as it could be used potentially to share the raftstore code.

components/engine_traits/src/engines.rs
zhangjinpeng1987: Maybe rename tablets to tablet_creator or other name can describe its meaning.

components/engine_traits/src/engine.rs
```diff
@@ -63,5 +66,101 @@ pub trait KvEngine:
...
+    fn tablets_path(&self) -> PathBuf;
+    fn clone(&self) -> Box<dyn TabletFactory<EK> + Send>;
+    fn load_tablet(&self, _path: &Path, id: u64, suffix: u64) -> Result<EK> {
+        self.open_tablet(id, suffix)
+    }
+    fn mark_tombstone(&self, _region_id: u64, _suffix: u64) {}
+    fn is_tombstoned(&self, _region_id: u64, _suffix: u64) -> bool {
+        false
+    }
+}
```
tabokie, line 103: I think we should introduce new method only when it's truly used.

components/engine_traits/src/engine.rs
```diff
@@ -63,5 +66,101 @@ pub trait KvEngine:
...
+    }
+    fn exists_raw(&self, path: &Path) -> bool;
+    fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf;
+    fn tablets_path(&self) -> PathBuf;
+    fn clone(&self) -> Box<dyn TabletFactory<EK> + Send>;
+    fn load_tablet(&self, _path: &Path, id: u64, suffix: u64) -> Result<EK> {
+        self.open_tablet(id, suffix)
+    }
+    fn mark_tombstone(&self, _region_id: u64, _suffix: u64) {}
+    fn is_tombstoned(&self, _region_id: u64, _suffix: u64) -> bool {
```
tabokie, line 100: The user would expect `is_tombstoned` following a `mark_tombstone` returns true. I suggest leave it unimplemented or blanket fill it with `unimplemented!()`.

components/engine_traits/src/engine.rs
tonyxuqqi: per the prototype we did, these methods should be used.
The tombstone is used for releasing the tablet.

tabokie: I mean adding it in the PR that uses it. I think it makes the context clearer and won't add dev overhead.

components/engine_traits/src/engine.rs
tonyxuqqi: fair.

components/engine_traits/src/engines.rs
tonyxuqqi: it's used to provide a unified way for creating engines. Per discussion offline, I can remove it for now. will revisit it if it's needed later.

src/server/engine_factory.rs
tonyxuqqi: > I don't get it. For v1, it's pointless or even wrong to call create_root_db multiple times.

For create_shared_db, the behavior is same. It will fail if created by multiple times. 
But here the change is just to cache the clone, so that next time for create_tablet, it can return it directly---essentially it has the same API with v2, but with v1's implementation. The purpose is to potentially share the code. If in the end it's not used, I will remove it. But I believe it's helpful, at least in some test cases.

components/engine_traits/src/engine.rs
tonyxuqqi: So the approach is we define the engine function first and then will implement other components. These functions can be used by different PRs in the same time. 
Because we're implementing a new version of raftstore and so it will be more efficient to define the APIs first. Otherwise, there's no need for this PR at all.

src/config.rs
```diff
@@ -178,7 +178,6 @@ impl TitanCfConfig {
         opts.set_level_merge(self.level_merge);
         opts.set_range_merge(self.range_merge);
         opts.set_max_sorted_runs(self.max_sorted_runs);
-        opts.set_gc_merge_rewrite(self.gc_merge_rewrite);
```
tabokie, line 181: Emit a warning in `validate`.

components/engine_traits/src/engine.rs
tabokie: In that case I presume you want it to be close the final version? How about use doc comments and order the methods properly (and consistently).

components/engine_traits/src/engine.rs
```diff
@@ -63,5 +66,120 @@ pub trait KvEngine:
 // It should be named as `EngineFactory` for consistency, but we are about to rename
 // engine to tablet, so always use tablet for new traits/types.
```
tabokie, line 67: Pls also explain a bit about id and suffix and cache.

components/engine_traits/src/engine.rs
```diff
@@ -63,5 +66,120 @@ pub trait KvEngine:
...
 pub trait TabletFactory<EK> {
```
tabokie, line 68: Would it be better to use a `TabletFactoryV2` instead? I understand it contains disjoint sets of interfaces. (e.g. old raftstore only use `create_shared_db` and should have access to other methods).

tonyxuqqi: No, old raftstore will likely use the new APIs in some code to share with V2.

components/engine_traits/src/engine.rs
tonyxuqqi: fair.

tabokie (approved):

zhangjinpeng1987:
@tonyxuqqi please fix the conficts

PR #12759, opened 2022-06-06
Reserve key space id encoding in backup convert

### LINKED ISSUE #12758: RawKV API V2 supports "key space" encoded in key

The [multi-tenant](https://en.wikipedia.org/wiki/Multitenancy) is an essential feature for cloud user to reduce cost. And we need the "key space" as a fundamental to implement multi-tenant.

RawKV API V2 support "key space" encoded in key.
As the key encoding is implemented in client (see https://github.com/tikv/client-java/issues/555), the only feature need to be implemented in TiKV now is the cross api version backup. See [backup/endpoint.rs:213](https://github.com/tikv/tikv/blob/da16e5eca3e44f8a987acdd6aad31acad3f5c05e/components/backup/src/endpoint.rs#L213), [backup/endpoint.rs:464](https://github.com/tikv/tikv/blob/da16e5eca3e44f8a987acdd6aad31acad3f5c05e/components/backup/src/endpoint.rs#L464),[backup/endpoint.rs:470](https://github.com/tikv/tikv/blob/da16e5eca3e44f8a987acdd6aad31acad3f5c05e/components/backup/src/endpoint.rs#L470).

haojinming:

Issue Number: Close #12758 

What's Changed:

Add encoding of key space id in apiv2 key when backup convert.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

```release-note

Add key space id encoding in apiv2 key when backup convert.
```

### DIFF
--- a/components/api_version/src/api_v1.rs
+++ b/components/api_version/src/api_v1.rs
@@ -1,6 +1,6 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
-use engine_traits::Error;
+use tikv_util::box_err;
 
 use super::*;
 
@@ -45,9 +45,7 @@ impl KvFormat for ApiV1 {
     ) -> Result<Key> {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok(Key::from_encoded_slice(key)),
-            ApiVersion::V2 => Err(Error::Engine(String::from(
-                "unsupported conversion from v2 to v1",
-            ))), // reject apiv2 -> apiv1 conversion
+            ApiVersion::V2 => Err(box_err!("unsupported conversion from v2 to v1")), // reject apiv2 -> apiv1 conversion
         }
     }
 
@@ -58,9 +56,7 @@ impl KvFormat for ApiV1 {
     ) -> Result<(Vec<u8>, Vec<u8>)> {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok((start_key, end_key)),
-            ApiVersion::V2 => Err(Error::Engine(String::from(
-                "unsupported conversion from v2 to v1",
-            ))), // reject apiv2 -> apiv1 conversion
+            ApiVersion::V2 => Err(box_err!("unsupported conversion from v2 to v1")), // reject apiv2 -> apiv1 conversion
         }
     }
 }
--- a/components/api_version/src/api_v1.rs
+++ b/components/api_version/src/api_v1.rs
@@ -1,6 +1,8 @@
 // Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
 
 use super::*;
+use tikv_util::box_err;
+use engine_traits::Error as EngineError;
 
 impl KvFormat for ApiV1 {
     const TAG: ApiVersion = ApiVersion::V1;
@@ -44,26 +46,20 @@ impl KvFormat for ApiV1 {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok(Key::from_encoded_slice(key)),
             ApiVersion::V2 => {
-                debug_assert_eq!(ApiV2::parse_key_mode(key), KeyMode::Raw);
-                let (user_key, _) = ApiV2::decode_raw_key(&Key::from_encoded_slice(key), true)?;
-                let raw_key = ApiV2::remove_prefix(user_key)?;
-                Ok(Self::encode_raw_key_owned(raw_key, None))
+                Err(EngineError::Other(box_err!("unsupported conversion")))
             }
         }
     }
 
     fn convert_raw_user_key_range_version_from(
         src_api: ApiVersion,
-        mut start_key: Vec<u8>,
-        mut end_key: Vec<u8>,
+        start_key: Vec<u8>,
+        end_key: Vec<u8>,
     ) -> Result<(Vec<u8>, Vec<u8>)> {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok((start_key, end_key)),
             ApiVersion::V2 => {
-                // TODO: check raw key range after check_api_version_range is refactored.
-                start_key = ApiV2::remove_prefix(start_key)?;
-                end_key = ApiV2::remove_prefix(end_key)?;
-                Ok((start_key, end_key))
+                Err(EngineError::Other(box_err!("unsupported conversion")))
             }
         }
     }
--- a/components/api_version/src/api_v1.rs
+++ b/components/api_version/src/api_v1.rs
@@ -45,9 +45,9 @@ impl KvFormat for ApiV1 {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok(Key::from_encoded_slice(key)),
             ApiVersion::V2 => {
                 debug_assert_eq!(ApiV2::parse_key_mode(key), KeyMode::Raw);
-                let (mut user_key, _) = ApiV2::decode_raw_key(&Key::from_encoded_slice(key), true)?;
-                user_key.remove(0); // remove first byte `RAW_KEY_PREFIX`
-                Ok(Self::encode_raw_key_owned(user_key, None))
+                let (user_key, _) = ApiV2::decode_raw_key(&Key::from_encoded_slice(key), true)?;
+                let raw_key = ApiV2::remove_prefix(user_key)?;
+                Ok(Self::encode_raw_key_owned(raw_key, None))
             }
         }
     }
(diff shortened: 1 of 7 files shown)


components/api_version/src/api_v1.rs
```diff
@@ -44,26 +46,20 @@ impl KvFormat for ApiV1 {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok(Key::from_encoded_slice(key)),
             ApiVersion::V2 => {
-                debug_assert_eq!(ApiV2::parse_key_mode(key), KeyMode::Raw);
-                let (mut user_key, _) = ApiV2::decode_raw_key(&Key::from_encoded_slice(key), true)?;
-                user_key.remove(0); // remove first byte `RAW_KEY_PREFIX`
-                Ok(Self::encode_raw_key_owned(user_key, None))
+                Err(EngineError::Other(box_err!("unsupported conversion")))
```
pingyu, line 49:
```suggestion
                Err(box_err!("unsupported conversion"))
```
The same to other three similar.

components/api_version/src/api_v2.rs
```diff
@@ -235,6 +235,16 @@ impl ApiV2 {
         Ok(Key::split_on_ts_for(key)?)
     }
 
+    pub fn add_prefix(key: &[u8]) -> Vec<u8> {
+        let mut apiv2_key = Vec::with_capacity(ApiV2::get_encode_len(key.len() + 2));
+        apiv2_key.push(RAW_KEY_PREFIX);
+        let mut tmp_buf = [0; MAX_VARINT64_LENGTH];
+        let written = NumberCodec::encode_var_u64(&mut tmp_buf, DEFAULT_KEY_SPACE_ID);
```
pingyu, line 242: Suggest to try making written be constant. Or just write a 0 and verify it's always equal to var_u64 in unit test.

components/api_version/src/api_v2.rs
```diff
@@ -235,6 +235,16 @@ impl ApiV2 {
         Ok(Key::split_on_ts_for(key)?)
     }
 
+    pub fn add_prefix(key: &[u8]) -> Vec<u8> {
+        let mut apiv2_key = Vec::with_capacity(ApiV2::get_encode_len(key.len() + 2));
```
pingyu, line 239: Reserve capacity after encoded keyspace id is got and avoid hard coding the length.

components/api_version/src/api_v1.rs
pingyu: Maybe just panic ? The conversion from V2 to V1/V1TTL should be rejected by TiKV-BR, and when receive backup request. Other than the moment to convert key version.

components/backup/src/endpoint.rs
```diff
@@ -206,20 +206,31 @@ async fn save_backup_file_worker(
         let files = if msg.files.need_flush_keys() {
             match msg.files.save(&storage).await {
                 Ok(mut split_files) => {
+                    let mut has_err = false;
                     for file in split_files.iter_mut() {
                         // In the case that backup from v1 and restore to v2,
                         // the file range need be encoded as v2 format.
                         // And range in response keep in v1 format.
-                        let (start, end) = codec.convert_key_range_to_dst_version(
+                        let ret = codec.convert_key_range_to_dst_version(
```
pingyu, line 214: Reject V2 -> V1/V1TTL earlier, e.g. `KeyValueCodec::check_backup_api_version`.

components/api_version/src/api_v1.rs
haojinming: done, mark apiv2->apiv1 as unreachable

components/backup/src/endpoint.rs
haojinming: done

components/api_version/src/api_v2.rs
haojinming: done

components/api_version/src/api_v2.rs
sunxiaoguang: Let's make key space fixed.

sunxiaoguang (approved):

components/api_version/src/api_v2.rs
haojinming: done

components/api_version/src/api_v1.rs
```diff
@@ -43,28 +43,18 @@ impl KvFormat for ApiV1 {
     ) -> Result<Key> {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => Ok(Key::from_encoded_slice(key)),
-            ApiVersion::V2 => {
-                debug_assert_eq!(ApiV2::parse_key_mode(key), KeyMode::Raw);
-                let (mut user_key, _) = ApiV2::decode_raw_key(&Key::from_encoded_slice(key), true)?;
-                user_key.remove(0); // remove first byte `RAW_KEY_PREFIX`
-                Ok(Self::encode_raw_key_owned(user_key, None))
-            }
+            ApiVersion::V2 => unreachable!(), // reject apiv2 -> apiv1 conversion
```
overvenus, line 46: It will panic TiKV if the arm matches. What about returning an error?

components/api_version/src/api_v2.rs
```diff
@@ -198,11 +197,11 @@ impl KvFormat for ApiV2 {
     ) -> (Vec<u8>, Vec<u8>) {
         match src_api {
             ApiVersion::V1 | ApiVersion::V1ttl => {
-                start_key.insert(0, RAW_KEY_PREFIX);
+                start_key = ApiV2::add_prefix(&start_key);
                 if end_key.is_empty() {
                     end_key.insert(0, RAW_KEY_PREFIX_END);
```
iosmanthus, line 202: The EMPTY end key should be mapped to `r001` or `x001`?.

iosmanthus: Another question, when do we need this function while executing backup and restore.

components/api_version/src/api_v1.rs
haojinming: backup process do check the apiversion in https://github.com/tikv/tikv/blob/ba391ff506c8b7f4b0cd7c9ef0b9f04ce87c3d7e/components/backup/src/utils.rs#L127-L133 to ensure that this branch is unreachable.  Handle error will make backup process a little complicated here https://github.com/tikv/tikv/blob/ba391ff506c8b7f4b0cd7c9ef0b9f04ce87c3d7e/components/backup/src/endpoint.rs#L213-L216. And we want take this pr in v6.1.0 release, so we need make min changes.

components/api_version/src/api_v2.rs
haojinming: This is a conversion scenario, no key space exists in apiv1, so I think we need set 'r' + 1 as the whole end. Anyway, 'r' + 1 and `r001` both works here. @pingyu How do you think?

haojinming: This function is needed when return backup response, range in `BackupResponse.File` need to be the apiv2 format.

pingyu: En... I suggest `r001` . It is a reasonble range start & end for default key space. Although both are ok here.

haojinming: After consideration, i think use `r001` as the end is better. `r001` as end can avoid some useless download and ingest during restore if target tikv cluster is not empty. I have changed the end as `r001`. PLAT~ @iosmanthus @pingyu

pingyu (approved):
LGTM~

components/api_version/src/api_v1.rs
BusyJay: Agree should return error instead of just panicking.

haojinming: @overvenus @BusyJay error handling is added, PTAL~ thanks.

components/api_version/src/api_v2.rs
```diff
@@ -16,6 +16,8 @@ pub const RAW_KEY_PREFIX_END: u8 = RAW_KEY_PREFIX + 1;
 pub const TXN_KEY_PREFIX: u8 = b'x';
 pub const TIDB_META_KEY_PREFIX: u8 = b'm';
 pub const TIDB_TABLE_KEY_PREFIX: u8 = b't';
+pub const DEFAULT_KEY_SPACE_ID: [u8; 3] = [0, 0, 0]; // reserve 3 bytes for key space id.
```
BusyJay, line 19: Why? I thought it should be just one byte?

haojinming: We plan to use 3 bytes fixed length to encode key space id, support more than 16 million key space. And we does not use var int encoding incase there are some issues we may miss.

Connor1996 (approved):
LGTM

PR #12781, opened 2022-06-08
check config change after flush

### LINKED ISSUE #12780: Optimize the config change check in raft_client MessageBuffer
## Development Task

#12335 introduced server config version check in `MessageBuffer::push`, but since this is a hot patch, this extra check(with atomic operations) may affect the performance.

glorv:

Issue Number: Close #12780

What's Changed:

```
Move the online config change from `push` to `flush` for better performance.
```

Benchmark:
I use the `bench_client_buffer_push` to test this change. Because it's hard to mock the grpc connect in `flush`, I use the `clear` method to mock flush that only clean the buffer and check the response. This bench doesn't do config change simultaneously as the config change is rural.

I run each bench for 10 times and choose the middle result, the middle result is almost equal to the avg result.
before:
```
running 1 test
test server::raft_client::tests::bench_buffer_push       ... bench:       2,171 ns/iter (+/- 762)
```

after:
```
running 1 test
test server::raft_client::tests::bench_buffer_push       ... bench:       2,102 ns/iter (+/- 514)
```

The bench result shows there is about **3%** performance improvement. In theory, the cost of an atomic load is less than 10ns, so the theoretical extra cost of the check is about 5%. Thus, the bench result matches the expectation.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression

### Release note 

```release-note
None
```

### DIFF
--- a/src/server/raft_client.rs
+++ b/src/server/raft_client.rs
@@ -1202,6 +1202,21 @@ mod tests {
         assert!(msg_buf.full());
     }
 
+    fn new_test_msg(size: usize) -> RaftMessage {
+        let mut msg = RaftMessage::default();
+        msg.set_region_id(1);
+        let mut region_epoch = RegionEpoch::default();
+        region_epoch.conf_ver = 1;
+        region_epoch.version = 0x123456;
+        msg.set_region_epoch(region_epoch);
+        msg.set_start_key(vec![0; size]);
+        msg.set_end_key(vec![]);
+        msg.mut_message().set_snapshot(Snapshot::default());
+        msg.mut_message().set_commit(0);
+        assert_eq!(BatchMessageBuffer::message_size(&msg), size);
+        msg
+    }
+
     #[test]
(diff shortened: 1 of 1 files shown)


src/server/raft_client.rs
```diff
@@ -1233,4 +1245,23 @@ mod tests {
         msg_buf.push(make_msg(2));
         assert!(msg_buf.full());
     }
+
+    #[bench]
+    fn bench_client_buffer_push(b: &mut test::Bencher) {
```
zhangjinpeng1987, line 1250: Better to move into tests/benches?

zhangjinpeng1987 (commented):
rest LGTM

src/server/raft_client.rs
zhangjinpeng1987: benches dir first.

zhangjinpeng1987 (approved):
LGTM

BusyJay (approved):

PR #12783, opened 2022-06-08
raftstore: record write durations into tracker

sticnarf:

Issue Number: Ref https://github.com/tikv/tikv/issues/12362

What's Changed:

```commit-message
This commit replaces the request_times in the raftstore callback with a
tracker token. Then, the waterfall metrics of a raft command will be
recorded into the tracker.
```

### Check List 

Tests 

### Release note 

```release-note
None
```


components/raftstore/src/store/async_io/write.rs
```diff
@@ -535,7 +535,13 @@ where
...
                     );
                 });
-            self.perf_context.report_metrics(&[]); // TODO: pass in request trackers
+            let trackers: Vec<_> = self
+                .batch
+                .tasks
+                .iter()
+                .flat_map(|task| task.trackers.iter().flat_map(|t| t.as_tracker_token()))
+                .collect();
+            self.perf_context.report_metrics(&trackers);
```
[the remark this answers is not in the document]
sticnarf: Well, this is some mess now. Raft engine does not have perf contexts, so I just use RocksDB's perf context now. If the user still uses RocksDB as the raft engine, this will still record related information.

I'm trying to add perf context to raft-engine: https://github.com/tikv/raft-engine/pull/227

components/raftstore/src/store/peer.rs
```diff
@@ -1585,7 +1582,7 @@ where
                 // network partition from the new leader.
                 // For lease safety during leader transfer, transit `leader_lease`
                 // to suspect.
-                self.leader_lease.suspect(monotonic_raw_now());
+                self.leader_lease.suspect(now);
```
cfzjywxk, line 1585: Why change this to use `now` above?

sticnarf: I thought they are CPU operations to handling these messages and should be very fast. So, it is not quite necessary to use different `now` for different messages and it should reduce some unncessary cost...

The change is not necessary, I revert it to still get a new time for each message.

components/raftstore/src/store/peer.rs
```diff
@@ -1610,22 +1607,31 @@ where
...
                 .iter()
-                .map(|e| e.get_index())
+                .map(|e| (e.get_term(), e.get_index()))
             {
-                while let Some((propose_idx, instant)) = self.pending_propose_instants.front() {
-                    if index == *propose_idx {
+                if let Ok(idx) = self
+                    .proposals
+                    .queue
+                    .binary_search_by_key(&index, |p: &Proposal<_>| p.index)
```
cfzjywxk, line 1619: Do we need to measure the new overhead introduced here since this is a crucial path?
Besides, do we need to record the duration used in the `RaftClient` as the message is just put into the buffer here?

components/raftstore/src/store/peer.rs
cfzjywxk: Make sense, maybe we could optimize this seperately.

components/raftstore/src/store/peer.rs
sticnarf: The overhead of the binary search here should be okay.. `find_trackers` (which was find_propose_time` before) also uses binary search in critical paths already. Anyway, I'll measure the overall performance affect of the whole PR.

> Besides, do we need to record the duration used in the RaftClient as the message is just put into the buffer here?

It would be good to have. The time between `send` and `flush` is missing and counted as part of `commit_log_duration`. But I don't have a good idea to record it in `RaftClient`. This may change too much of the interfaces.

cfzjywxk (approved):
LGTM

sticnarf:
Overall CPU cost is 0.3%. Looks acceptable...

tikv

you06 (approved):

PR #12791, opened 2022-06-09
gRPC: use gzip level-2 compression by default

zkkxu:
Issue Number:  Ref #12929

this pr only sacrifice 5% CPU to cut off 50% cross az network traffic between tikvs

What's Changed:
```commit-message
add initial arguments for gzip compression:
gzip_compression_level: represent gzip compression level, the origin gzip compression level is 6 and hardcoding;
compression_lower_bound: this represent gzip will compress the data only larger than this
```

### Check List 

Tests 
- No code

Side effects
- Performance regression
    - Only cost more 5% CPU

### Release note 

```release-note
none
```


src/server/config.rs
```diff
@@ -26,6 +26,8 @@ const DEFAULT_GRPC_CONCURRENT_STREAM: i32 = 1024;
 const DEFAULT_GRPC_RAFT_CONN_NUM: usize = 1;
 const DEFAULT_GRPC_MEMORY_POOL_QUOTA: u64 = isize::MAX as u64;
 const DEFAULT_GRPC_STREAM_INITIAL_WINDOW_SIZE: u64 = 2 * 1024 * 1024;
+const DEFAULT_GRPC_GZIP_COMPRESSION_LEVEL: usize = 6;
```
kevin-xianliu, line 29: Should us set the default level 2 instead of 6?

src/server/config.rs
```diff
@@ -213,6 +219,8 @@ impl Default for Config {
             raft_client_queue_size: 8192,
             raft_msg_max_batch_size: 128,
             grpc_compression_type: GrpcCompressionType::None,
+            grpc_gzip_compression_level: 6,
```
kevin-xianliu, line 222: Better to use  DEFAULT_GRPC_GZIP_COMPRESSION_LEVEL and DEFAULT_GRPC_COMPRESSION_LOWER_BOUND.

src/server/config.rs
zhangjinpeng1987: agree

knightXun: nice hit

BusyJay (commented):
Do connections between tikv and tidb need compression?

zkkxu:
> Do connections between tikv and tidb need compression?

that depends on tidb client

BusyJay:
If client enables gzip, what level will tikv use?

zkkxu:
> If client enables gzip, what level will tikv use?

tikv only decompress the data. as for the returned data, it depends on tikv grpc client

BusyJay:
> as for the returned data, it depends on tikv grpc client

If client accepts gzip, what level will be used?

zkkxu:
> > as for the returned data, it depends on tikv grpc client
> 
> If client accepts gzip, what level will be used?

If size less than grpc_min_message_size_to_compress, will use grpc_gzip_compression_level, default 2.

BusyJay:
> If size larger than grpc_min_message_size_to_compress, will use grpc_gzip_compression_level, default 2. otherwise no compression

How does server know it should use level 2? I don't see it's configured anywhere.

zkkxu:
> > If size larger than grpc_min_message_size_to_compress, will use grpc_gzip_compression_level, default 2. otherwise no compression
> 
> How does server know it should use level 2? I don't see it's configured anywhere.
grpc message has a metadata, the level will be writed into the metadata

BusyJay:
> U can image the grpc message has a metadata, the level will be writed into the metadata

I fully understand how it work. But I don't see it's configured in this PR. There is no way for grpc know the level should be 2 instead of 3, 4, 5, 6 when responding to client.

zkkxu:
> > U can image the grpc message has a metadata, the level will be writed into the metadata
> 
> I fully understand how it work. But I don't see it's configured in this PR. There is no way for grpc know the level should be 2 instead of 3, 4, 5, 6 when responding to client.

oh, there's no compression between tikv and tidb. this pr only compress data between tikvs

BusyJay (approved):

zhangjinpeng1987 (approved):
LGTM

PR #12805, opened 2022-06-13
tests: ignore env var dependent tests at runtime

### LINKED ISSUE #12804: all ignored tests are run in `make test`

N/A

N/A

### Steps to reproduce

`make test`

### What did happened?

At this line:
https://github.com/tikv/tikv/blob/d6d6d6ee725639d05ce81577a71e4a76c14152ce/scripts/test-all#L23

`--package tikv_alloc` argument is ignored when `--workspace` is present. So all ignored tests other than those in `tikv_alloc` are run. This is not expected.

tabokie:

Issue Number: Close #12804

What's Changed:

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/tikv_alloc/src/jemalloc.rs
+++ b/components/tikv_alloc/src/jemalloc.rs
@@ -191,11 +191,12 @@ mod profiling {
         //
         // TODO: need a test for the dump_prof(None) case, but
         // the cleanup afterward is not simple.
+        #[allow(non_snake_case)]
         #[test]
-        #[ignore]
-        fn test_profiling_memory() {
+        fn test_profiling_memory_where_MALLOC_CONF() {
+            std::env::set_var("MALLOC_CONF", "prof::true");
             // Make sure somebody has turned on profiling
-            assert!(is_profiling_on(), r#"Set MALLOC_CONF="prof:true""#);
+            assert!(is_profiling_on(), r#"Set MALLOC_CONF=prof:true"#);
 
             let dir = Builder::new()
                 .prefix("test_profiling_memory")
--- a/components/tikv_alloc/src/lib.rs
+++ b/components/tikv_alloc/src/lib.rs
@@ -82,6 +82,10 @@
 //! `--features=mem-profiling` to cargo for eather `tikv_alloc` or
 //! `tikv`.
 
+#![cfg_attr(test, feature(test))]
+#![cfg_attr(test, feature(custom_test_frameworks))]
+#![cfg_attr(test, test_runner(runner::run_env_conditional_tests))]
+
 #[cfg(feature = "jemalloc")]
 #[macro_use]
 extern crate lazy_static;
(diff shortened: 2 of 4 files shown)


components/tikv_alloc/src/jemalloc.rs
```diff
@@ -191,11 +191,11 @@ mod profiling {
         //
         // TODO: need a test for the dump_prof(None) case, but
         // the cleanup afterward is not simple.
+        #[allow(non_snake_case)]
         #[test]
-        #[ignore]
-        fn test_profiling_memory() {
+        fn test_profiling_memory_where_MALLOC_CONF() {
             // Make sure somebody has turned on profiling
-            assert!(is_profiling_on(), r#"Set MALLOC_CONF="prof:true""#);
```
Connor1996, line 198: why remove the inner quotation

components/tikv_alloc/src/lib.rs
```diff
@@ -124,3 +128,44 @@ pub use crate::{imp::*, trace::*};
 
 #[global_allocator]
 static ALLOC: imp::Allocator = imp::allocator();
+
+#[cfg(test)]
+mod runner {
+    extern crate test;
+    use test::*;
+
+    pub fn run_env_conditional_tests(cases: &[&TestDescAndFn]) {
```
Connor1996, line 137: depict what it does

components/tikv_alloc/src/lib.rs
```diff
@@ -124,3 +128,37 @@ pub use crate::{imp::*, trace::*};
...
+            .map(|case| {
+                let mut desc = case.desc.clone();
+                let testfn = match case.testfn {
+                    TestFn::StaticTestFn(f) => TestFn::StaticTestFn(f),
+                    TestFn::StaticBenchFn(f) => TestFn::StaticBenchFn(f),
+                    ref f => panic!("unexpected testfn {:?}", f),
+                };
+                if let Some(msg) = desc.ignore_message {
+                    let keyword = "#ifdef";
+                    if let Some(var_name) = msg.strip_prefix(keyword) {
```
sticnarf, line 151: After stripping keyword, `var_name` still contains a space at the beginning.

sticnarf (commented):
Do you know why CI fails because tikv-ctl depends on libstdc++?

tabokie:
> Do you know why CI fails because tikv-ctl depends on libstdc++?

I do not. Upstream reports https://github.com/rust-lang/rust/issues/94983 but it didn't fix our issue. I'm trying to reproduce it in a smaller repo and bisect the cause.

cmd/build.rs
```diff
@@ -32,7 +32,7 @@ fn link_sys_lib(lib: &str, tool: &cc::Tool) {
     }
     // remove lib prefix and .a postfix.
     let libname = &lib[3..lib.len() - 2];
-    println!("cargo:rustc-link-lib=static={}", &libname);
+    println!("cargo:rustc-link-lib=static:+whole-archive={}", &libname);
```
Connor1996, line 35: what's this

tabokie: From what I can read, whole-archive flag will force using static symbols when there're both a static and a dynamic version available. It is enabled by default before. The fact that we must enable it to pass the check means we somehow linked multiple versions of libstdc++.

sticnarf (approved):
LGTM

cmd/build.rs
BusyJay: How can that happen? I think it should report duplicated symbol error.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2022-02-14
+nightly-2022-05-01
```
BusyJay, line 1: Why use 05-01?

cmd/build.rs
tabokie: I meant to say that multiple versions are available for the linker. So it's up to the implementation to choose one of them. And FYI during the debugging I have ruled out rust-rocksdb by setting the `static_libcpp` feature.

rust-toolchain
tabokie: https://github.com/rust-lang/rust/issues/96132 is fixed in 04-17.

Connor1996 (approved):
LGTM

PR #12806, opened 2022-06-13
server: change TiKvServer to TiKVServer

zhangjinpeng1987:
Signed-off-by: zhangjinpeng1987 <zhangjinpeng@pingcap.com>

What's Changed:

```commit-message
Typo: change TiKvServer to TiKVServer
```

### Related changes

- No

### Check List 

Tests 

- No code

Side effects

No

### Release note 

```release-note
None
```

### DIFF
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -8,7 +8,7 @@
 //! The entry point is `run_tikv`.
 //!
 //! Components are often used to initialize other components, and/or must be explicitly stopped.
-//! We keep these components in the `TiKvServer` struct.
+//! We keep these components in the `TiKVServer` struct.
 
 use std::{
     cmp,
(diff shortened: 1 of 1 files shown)


sticnarf:
Per https://rust-lang.github.io/api-guidelines/naming.html, if we consider TiKV as a single word, the camel case should be `TikvServer` instead of `TiKVServer`

zhangjinpeng1987:
> Per https://rust-lang.github.io/api-guidelines/naming.html, if we consider TiKV as a single word, the camel case should be `TikvServer` instead of `TiKVServer`

@sticnarf TiKV is a “Distinguished Name”

sticnarf:
> @sticnarf TiKV is a “Distinguished Name”

That doesn't account for using `TiKV` in camel case. If we have to keep the original cases in camel case, why we do not in snake case, for example it's certainly  `modify_tikv_config` instead of `modify_TiKV_config`.

BusyJay:
It's named `TiKv` on purpose to follow rust naming convention. https://github.com/tikv/tikv/pull/12395

zhangjinpeng1987:
> > @sticnarf TiKV is a “Distinguished Name”
> 
> That doesn't account for using `TiKV` in camel case. If we have to keep the original cases in camel case, why we do not in snake case, for example it's certainly `modify_tikv_config` instead of `modify_TiKV_config`.

both TiKV and tikv are acceptable distinguished name, but TiKv or Tikv is not.

zhangjinpeng1987:
> It's named `TiKv` on purpose to follow rust naming convention. #12395

I think TiKv or Tikv is very ugly.

zhangjinpeng1987:
Close this PR since above oppositions.

PR #12818, opened 2022-06-14
*: set health status to unknown when raftstore gets stuck (#12411)

### LINKED ISSUE #12398: Make health service aware of raftstore unavailability

ti-srebot:
cherry-pick #12411 to release-5.4
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/12818
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/12818:release-5.4-2552144c31b4
```

---

Issue Number: Close #12398

What's Changed:

```commit-message
The client uses health service to determine whether it should
still send requests to this TiKV or whether it should refresh
related region cache.

But if the raftstore alone becomes unavailable because of IO hang
or bugs, the health service still returns Serving status. This may
mislead the TiKV client and increase the recover time even if the
leader is already transferred to other TiKV instances.

This commit reuses the mechanism of slow score calculation
to detect whether the raftstore is normally working. So, the client
can refresh its region cache in time.
```

### Check List 

Tests 

- Integration test

### Release note 

```release-note
Report bad health status if raftstore stops working.
```


components/raftstore/Cargo.toml
```diff
@@ -40,6 +40,11 @@ file_system = { path = "../file_system", default-features = false }
 fs2 = "0.4"
 futures = "0.3"
 futures-util = { version = "0.3.1", default-features = false, features = ["io"] }
+<<<<<<< HEAD
```
zhangjinpeng1987, line 43: @sticnarf Please fix conflicts.

sticnarf (approved):

PR #12836, opened 2022-06-15
DNM: Test ART

cosven:


tabokie:
@cosven I think you will need to port rust bindings to enable ART (like https://github.com/tikv/rust-rocksdb/pull/339). Or set it as the default in RocksDB (while disabling tikv enable_doubly_skiplist).

cosven:
> @cosven I think you will need to port rust bindings to enable ART (like [tikv/rust-rocksdb#339](https://github.com/tikv/rust-rocksdb/pull/339)). Or set it as the default in RocksDB (while disabling tikv enable_doubly_skiplist).

@tabokie Thanks for your reminder. I'll take a look.

PR #12837, opened 2022-06-16
raftstore: record metrics of proposal send wait duration

sticnarf:

Issue Number: Ref #12362 

What's Changed:

```commit-message
Our raft implementation has its flow control mechanism to limit the inflight
message number. But we're not able to know whether and when we are
throttled. Then, it's hard for us to know whether we should adjust the
max-inflight-msgs config.

This commit tries to improve this case. It's complex to add hooks to raft-rs
to know how long a message is throttled, but we can do it in the raftstore.
We record the propose time, and consume it and record it in the histogram
when the messages is send through the transport. If flow control takes effect,
the ready will be smaller than all under-replicated logs. So, we can know if
flow control takes effect.
```

Grafana panel looks like:

image

### Check List 

Tests 

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

This should be very minor and not noticable.

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/peer.rs
+++ b/components/raftstore/src/store/peer.rs
@@ -724,6 +724,9 @@ where
     #[getset(get = "pub")]
     leader_lease: Lease,
     pending_reads: ReadIndexQueue<EK::Snapshot>,
+    /// Record the propose instants to calculate the wait duration before
+    /// the proposal is sent through the Raft client.
+    pending_propose_instants: VecDeque<(u64, Instant)>,
 
     /// If it fails to send messages to leader.
     pub leader_unreachable: bool,
@@ -924,6 +927,7 @@ where
             raft_max_inflight_msgs: cfg.raft_max_inflight_msgs,
             proposals: ProposalQueue::new(tag.clone()),
             pending_reads: Default::default(),
+            pending_propose_instants: Default::default(),
             peer_cache: RefCell::new(HashMap::default()),
             peer_heartbeats: HashMap::default(),
             peers_start_pending_time: vec![],
@@ -1570,6 +1574,7 @@ where
         ctx: &mut PollContext<EK, ER, T>,
         msgs: Vec<RaftMessage>,
     ) {
+        // let now = Instant::now();
         for msg in msgs {
             let msg_type = msg.get_message().get_msg_type();
             if msg_type == MessageType::MsgTimeoutNow && self.is_leader() {
@@ -1594,6 +1599,26 @@ where
                 "to" => to_peer_id,
                 "disk_usage" => ?msg.get_disk_usage(),
             );
+
+            for index in msg
+                .get_message()
+                .get_entries()
+                .iter()
+                .map(|e| e.get_index())
+            {
+                while let Some((propose_idx, instant)) = self.pending_propose_instants.front() {
+                    if index == *propose_idx {
+                        ctx.raft_metrics
+                            .proposal_send_wait
+                            .observe(instant.elapsed().as_secs_f64());
+                    }
+                    if index >= *propose_idx {
+                        self.pending_propose_instants.pop_front();
+                    } else {
+                        break;
+                    }
+                }
+            }
 
             if let Err(e) = ctx.trans.send(msg) {
                 // We use metrics to observe failure on production.
@@ -2047,6 +2072,7 @@ where
                     self.mut_store().cancel_generating_snap(None);
                     self.clear_disk_full_peers(ctx);
                     self.clear_in_memory_pessimistic_locks();
+                    self.pending_propose_instants.clear();
                 }
                 _ => {}
             }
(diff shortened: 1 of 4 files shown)


components/raftstore/src/store/peer.rs
```diff
@@ -4269,6 +4295,9 @@ where
             }
         }
 
+        self.pending_propose_instants
+            .push_back((propose_index, Instant::now()));
```
Connor1996, line 4299: use `coarse_now()`

components/raftstore/src/store/peer.rs
```diff
@@ -1570,6 +1574,7 @@ where
         ctx: &mut PollContext<EK, ER, T>,
         msgs: Vec<RaftMessage>,
     ) {
+        // let now = Instant::now();
```
Connor1996, line 1577: remove

Connor1996 (commented):
How about just visiting the inflighs in raft progress, we can get the number of inflight msgs directly

Connor1996:
> > How about just visiting the `inflights` in `raft progress`, we can get the number of inflight msgs directly
> 
> I was thinking about how to show the metrics. From the inflight msgs in progress, we can know whether flow control happens, but it may be hard to tell how much it contributes to the final latency... Any good ideas?

Emmm... Seems both are needed

cfzjywxk:
Seems the `Proposal send wait duration` is ok to show the duration impact of inflight flow control. Maybe we should also record this duration in the request tracker and in slow log?

components/raftstore/src/store/peer.rs
sticnarf: I think `Instant::now()` is fast enough. It occupies <0.05% CPU on the flamegraph in my testing. I worry `coarse_now()` may not have enough accuracy (the accuracy can be 10ms on some Linux distributions).

Connor1996 (approved):
LGTM

components/raftstore/src/store/peer.rs
```diff
@@ -1595,6 +1600,26 @@ where
...
+                .get_message()
+                .get_entries()
+                .iter()
+                .map(|e| e.get_index())
+            {
+                while let Some((propose_idx, instant)) = self.pending_propose_instants.front() {
+                    if index == *propose_idx {
+                        ctx.raft_metrics
+                            .proposal_send_wait
+                            .observe(now.saturating_duration_since(*instant).as_secs_f64());
```
zhangjinpeng1987, line 1613: Does this can affect the performance since it is in the hot path?

zhangjinpeng1987:
I remember we have the `propose_wait` metrics, can we reuse or merge it?

components/raftstore/src/store/peer.rs
sticnarf: In my benchmark, the overhead is barely noticable on the flamegraph. It's recorded once per proposal, just like other metrics.

sticnarf:
> I remember we have the `propose_wait` metrics, can we reuse or merge it?

I think they are different metrics and it's better for us to distinguish.

`propose_wait` reflects how busy the store loop is. If it is long, we will either use async io, or increase the store pool size if cpu is exhausted.

This `propose_send_wait` duration reflects the effect of flow control. It will lead us to investigate network issues or increase max-inflight-msgs.

They are indications of different problems, so we need to separate them.

sticnarf:
> Seems the `Proposal send wait duration` is ok to show the duration impact of inflight flow control. Maybe we should also record this duration in the request tracker and in the slow log?

I can integrate it into #12783 later. This PR is easier to review, so let's merge it first.

cfzjywxk (approved):
LGTM

components/raftstore/src/store/peer.rs
TonsnakeLin: Here has two layers of loops, can we reduce it to one layer? e.g. define pending_propose_instants as a hashmap<propose_index, instant>, then we can access it directly.

sticnarf: Messages are sent in the same order with proposing. So, this nested loop will not cause quadratic complexity. Therefore, using `VecDeque` (of which the inner implementation is a ring buffer) is actually cheaper than a `HashMap`.

BusyJay: Note other metrics use waterfall to reduce performance loss. In the previous benchmark, it can cause about several percents drop. I think you can use poll_ctx.current_time to record the time.

components/raftstore/src/store/peer.rs
```diff
@@ -725,6 +725,9 @@ where
     #[getset(get = "pub")]
     leader_lease: Lease,
     pending_reads: ReadIndexQueue<EK::Snapshot>,
+    /// Record the propose instants to calculate the wait duration before
+    /// the proposal is sent through the Raft client.
+    pending_propose_instants: VecDeque<(u64, Instant)>,
```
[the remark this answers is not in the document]
sticnarf: Now this field has been removed. The propose time in `Proposal` is used instead.

PR #12839, opened 2022-06-16
cdc: set min-ts-interval to 200ms to reduce latency

### LINKED ISSUE #12840: reduce cdc latency
## Development Task

Cdc latency is determined by several factors. On the TiKV side the key is `cdc.min-ts-interval`, which is `1s` by default. If we can set it to a small value, it will be helpful to reduce cdc latency.

However some issues should be resolved:
* whether it will utilize too much CPU resource or not; tso thread count can be a bottleneck;
* whether it will utilize too much CPU resource on the CDC side or not.

hicqu:

Set cdc.min-ts-interval to 200ms to reduce latency.

Issue Number: Close #12840 

```commit-message
cdc: set min-ts-interval to 200ms to reduce latency
```

### Related changes

### Check List 

Tests 

- Unit test
- Manual test
  https://github.com/pingcap/tiflow/pull/5440 and https://github.com/tikv/tikv/pull/12576 shows that after the it's changed, TiKV and TiCDC won't utilize too much CPU resource and latency can be reduced obviously. 

Side effects

### Release note 

```release-note
change cdc.min-ts-interval from 1s to 200ms to reduce latency
```


zhangjinpeng1987:
Better to provide some performance test under heavy write like sysbench write workload. Also provide the flamegraph that show the side-effect of enlarge the frequency by 5 times.

overvenus (approved):

3pointer (approved):

PR #12841, opened 2022-06-16
copr: Support paging for executors

wshwsh12:

Issue Number: ref https://github.com/tikv/tikv/issues/12848

What's Changed:
The modify of each executor:
1. Aggregate: Check the group length is bigger than the paging_size or not. If true, the aggregate should return immediately to avoid memory growing.
2. TopN: Check the paging_size is bigger than N. If true, use the original logic of TopN. If false, we return the result from the children executor directly.
3. Limit: Use the original logic of Limit is ok.

```commit-message
Support paging for Aggregate/Limit/TopN executors.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/tidb_query_executors/src/runner.rs
```diff
@@ -172,6 +172,7 @@ pub fn build_executors<S: Storage + 'static>(
     ranges: Vec<KeyRange>,
     config: Arc<EvalConfig>,
     is_scanned_range_aware: bool,
+    paging_size: Option<u64>,
```
sticnarf, line 175: You can try putting these inside of `EvalConfig`.

zhangjinpeng1987:
@wshwsh12 Please add a related issue or rfc to describe what problem this PR want to solve.

components/tidb_query_executors/src/runner.rs
```diff
@@ -378,6 +378,8 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
     ) -> Result<Self> {
         let executors_len = req.get_executors().len();
         let collect_exec_summary = req.get_collect_execution_summaries();
+        let mut config = EvalConfig::from_request(&req)?;
+        config.paging_size = paging_size;
         let config = Arc::new(EvalConfig::from_request(&req)?);
```
sticnarf, line 383: You should use the `config` created above instead of creating an `EvalConfig` again

components/tidb_query_executors/src/top_n_executor.rs
```diff
@@ -309,6 +338,15 @@ impl<Src: BatchExecutor> BatchExecutor for BatchTopNExecutor<Src> {
             };
         }
 
+        match self.context.cfg.paging_size {
+            None => {}
```
sticnarf, line 342: Use `if let Some(paging_size) = self.context.cfg.paging_size`

sticnarf:
Now, we don't need to set `is_scanned_range_aware` when paging is on?

wshwsh12:
> Now, we don't need to set `is_scanned_range_aware` when paging is on?

@sticnarf  No. We need set `is_scanned_range_aware` to tell ScanExecutor to record the scanned range.

sticnarf (approved):

wshwsh12:
> But now, `update_scanned_range_from_scanned_row` is kept the same as before. It is updated every row. Will you improve this later?
I will improve it in the next PR

components/tidb_query_executors/src/top_n_executor.rs
```diff
@@ -309,6 +338,12 @@ impl<Src: BatchExecutor> BatchExecutor for BatchTopNExecutor<Src> {
             };
         }
 
+        if let Some(paging_size) = self.context.cfg.paging_size {
+            if self.n > paging_size as usize {
```
you06, line 342: if `n > paging_size > batch_size`, only one batch will be scanned. The cop loop then scans the executor again because `record_all < paging_size` calls the topn executor and reads more data and concatenates them together. I'm worried if the total order can be guaranteed in the concatenated chunk.

you06 (approved):

PR #12843, opened 2022-06-16
raftstorev2: add basic layout

### LINKED ISSUE #12842: Support physical isolation between regions
RFC: tikv/rfcs#93

Summary:

Tasks:

- [ ] raftstore v2
    - [x] implement read
    - [x] implement write
    - [x] implement split
    - [ ] implement merge
    - [x] implement conf change
    - [x] coprocessor hooks
    - [x] memory trace
    - [x] flow control
    - [x] peer lifetime trace
- [x] introduce tablet
- [x] implement snapshot

More to be added.

BusyJay:
Issue Number: Ref #12842

What's Changed:

```commit-message
This is an attempt to reimplement raftstore using the new assumptions
that peer's range can be overlapped.

Currently, compatability is not considered, though we may think about
how to migrate from old version by the end of this year.

No concrete implementations is added yet, we may choose reuse
implementation from v1 or implementing new logic base on actual
requirement. The principle is
1. do not introduce history debt while reusing code as much as possible.
2. do not change the current implementations.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


zhangjinpeng1987 (approved):
LGTM

components/raftstore-v2/src/router/message.rs
```diff
@@ -0,0 +1,293 @@
+// Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
```
tonyxuqqi, line 1: 2022

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -0,0 +1,56 @@
...
+};
+use slog::Logger;
+
+/// A storage for raft.
+pub struct Storage<ER> {
+    engine: ER,
+    logger: Logger,
+}
+
+impl<ER> Storage<ER> {
```
tonyxuqqi, line 16: for Raft Storage,  what's the expecting difference between v2 and v1?

components/raftstore-v2/src/router/message.rs
```diff
@@ -0,0 +1,293 @@
...
+            ResponseChannel::None => write!(fmt, "Callback::None"),
+            ResponseChannel::Read(_) => write!(fmt, "Callback::Read(..)"),
+            ResponseChannel::Write { .. } => write!(fmt, "Callback::Write(..)"),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
+pub enum PeerTick {
```
tonyxuqqi, line 137: why not use V1's implementation as they're identical?

components/raftstore-v2/src/router/message.rs
```diff
@@ -0,0 +1,293 @@
...
+            PeerTick::CheckLeaderLease,
+            PeerTick::ReactivateMemoryLock,
+            PeerTick::ReportBuckets,
+        ];
+        TICKS
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum StoreTick {
```
tonyxuqqi, line 187: Same here. It seems identical to v1.

components/raftstore-v2/src/router/message.rs
```diff
@@ -0,0 +1,293 @@
...
+        RaftCommand {
+            request,
+            ch,
+            send_time: Instant::now(),
+        }
+    }
+}
+
+/// Message that can be sent to a peer.
+pub enum PeerMsg<EK: KvEngine> {
```
tonyxuqqi, line 228: Comparing with V1, it does not have CasualMessage, HeartbeatPd, Destroy.   So are they not needed anymore in v2?

components/raftstore-v2/src/raft/storage.rs
BusyJay: The initialization is different.

components/raftstore-v2/src/router/message.rs
BusyJay: No, they are different. Several ticks are removed.

components/raftstore-v2/src/router/message.rs
BusyJay: Several ticks will be moved from `StoreTick` to `PeerTick`.

components/raftstore-v2/src/router/message.rs
BusyJay: No, I delete them on purpose and add them as needed in future development.

tonyxuqqi:
Also regarding the assumptions:
1) peer's range can be overlapped: do you mean the actual tablet's data range or the region's range. 
2) "Currently, compatability is not considered, though we may think about
how to migrate from old version by the end of this year",  does it mean the TiKV won't work for both version until this end of the year? (Or do you mean we will essentially two version of tikv?)

BusyJay:
>     1. peer's range can be overlapped: do you mean the actual tablet's data range or the region's range.

Both can be overlapped.
 
> does it mean the TiKV won't work for both version until this end of the year? (Or do you mean we will essentially two version of tikv?)

Yes, it will work with either one but not both at the same time. There will be only one version TiKV while can start with different version of raftstore.

tonyxuqqi (approved):

components/raftstore-v2/src/router/message.rs
tisonkun: ASF suggests using the interception year of the project consistently for all its files. But of course it's not a requirement.

* https://github.com/apache/skywalking/issues/10223
* https://issues.apache.org/jira/browse/LEGAL-629

PR #12845, opened 2022-06-16
Storage: Support regional flow controller

### LINKED ISSUE #12844: support regional flow controller in multi-rocksdb version
## Development Task
Today the granularity of flow controller is per TiKV node and technically this is because of the single RocksDB instance. 
In multi-rocksdb instance version, since all the RocksDB metrics is per region/tablet,  it's very straightforward to support flow controller per region. 
From functional perspective, regional flow controller also has the benefit to control the traffic more precisely and get ready for future expansion such as multi-tenancy support.  
The flow control caller should be same for both single rocksdb version or multi-rocksdb version. As a result, the FlowControl trait will be designed for the latter, but the current implementation will just fit the new interface.

tonyxuqqi:
Issue Number: Close #12844

What's Changed:

```commit-message
Make FlowController as trait and implement two versions for single rocksdb version and multi-rocksdb version
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None
```


BusyJay:
The change is too large, please split into smaller ones.

tonyxuqqi:
> The change is too large, please split into smaller ones.

it cannot be split. Most of the code is interface change's adapter code. The real changes are in two files.

src/storage/txn/flow_controller.rs
```diff
@@ -488,6 +517,100 @@ impl<E: CFNamesExt + FlowControlFactorsExt + Send + 'static> FlowChecker<E> {
...
             wait_for_destroy_range_finish: false,
+            tablet_suffix,
+        }
+    }
+
+    pub fn get_tablet_suffix(&self) -> u64 {
+        self.tablet_suffix
+    }
+
+    pub fn set_tablet_suffix(&mut self, tablet_suffix: u64) {
```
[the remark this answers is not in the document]
tonyxuqqi: Good to know thanks.
Fixed.

src/storage/txn/flow_controller.rs
```diff
@@ -66,21 +67,37 @@ enum Trend {
 ///
 /// Here is a brief flow showing where the mechanism works:
 /// grpc -> check should drop(discardable ratio) -> limiter -> async write to raftstore
-pub struct FlowController {
+
+pub trait FlowController {
```
Connor1996, line 71: no need to introduce this level of abstraction. We can just use TabletFlowController, and consume there is only one tablet when dynamic region is disabled.

src/storage/txn/tablet_flow_controller.rs
```diff
@@ -0,0 +1,727 @@
...
+                                checker.on_flow_info_msg(enabled, msg);
+                            }
+                        }
+                        Ok(FlowInfo::Created(region_id, suffix)) => {
+                            let mut checkers = flow_checkers.as_ref().write().unwrap();
+                            let checker = checkers
+                                .entry(region_id)
+                                .or_insert_with(|| insert_limiter_and_checker(region_id, suffix));
+                            // check if the checker's engine is exactly (region_id, suffix)
+                            // if  checker.suffix < suffix, it means its tablet is old and needs the refresh
```
zhangjinpeng1987, line 166: Why use "suffix" to indicate the freshness?

src/storage/txn/tablet_flow_controller.rs
```diff
@@ -0,0 +1,727 @@
...
+mod tests {
+    use engine_rocks::FlowInfo;
+    use engine_traits::DummyFactory;
+
+    use super::{
+        super::flow_controller::{tests::*, *},
+        *,
+    };
+
+    fn test_tablet_flow_controller()
```
zhangjinpeng1987, line 291: should rename to `make_test_tablet_flowxxx` or similar

src/storage/txn/tablet_flow_controller.rs
```diff
@@ -0,0 +1,727 @@
...
+            region_id,
+            tablet_suffix,
+        ))
+        .unwrap();
+        assert_eq!(flow_controller.should_drop(region_id), false);
+        assert_eq!(flow_controller.is_unlimited(region_id), false);
+        assert_ne!(flow_controller.consume(region_id, 2000), Duration::ZERO);
+
+        // not throttle once the number of memtables falls below the threshold
+        stub.0.num_memtable_files.store(1, Ordering::Relaxed);
```
zhangjinpeng1987: num_memtables?

src/storage/txn/scheduler.rs
```diff
@@ -199,7 +199,7 @@ struct SchedulerInner<L: LockManager> {
     // used to control write flow
     running_write_bytes: CachePadded<AtomicUsize>,
 
-    flow_controller: Arc<FlowController>,
+    flow_controller: Arc<dyn FlowController + Send + Sync>,
```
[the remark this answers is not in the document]
zhangjinpeng1987: +1

src/storage/txn/flow_controller.rs
tonyxuqqi: > no need to introduce this level of abstraction. We can just use TabletFlowController, and assume there is only one tablet when dynamic region is disabled.

There's a performance overhead for TabletFlowController, which will need a Hashmap and the lock for each access. 
Also logically we simply cannot use the same implementation, because it will get the tablet by hashmap's lookup on region_id, which won't really work for single rocksdb version------unless introduce some ugly if/else logic for single rocksdb version. 
That's why I'd rather to have this abstraction to make the implementation cleaner for each version.

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
tonyxuqqi: Generic will introduce too much code change (As it will add additional generic parameter for all places that use this class). Just like TabletFactory, I prefer dynamic dispatch here. The performance difference is very very little between two.

src/storage/txn/tablet_flow_controller.rs
tonyxuqqi: It's the design of tablet---basically the suffix is the raft log index. So higher suffix, the refresher the tablet.

src/storage/txn/tablet_flow_controller.rs
tonyxuqqi: That's the existing code, but i agree with you.

src/storage/txn/tablet_flow_controller.rs
tonyxuqqi: good point.

zhangjinpeng1987 (approved):
LGTM

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
Connor1996: we already have a generic LockManager, the change won't be too breaking

zhangjinpeng1987:
@Connor1996 any other comments?

src/storage/txn/scheduler.rs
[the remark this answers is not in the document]
tonyxuqqi: There's other places using Arc<FlowController>,  such as  node.rs, config.rs, config_manager.rs, mod.rs.  The performance difference of dyn and static binding is negligible in this case, the actual dispatch impact is at most a few ns per call.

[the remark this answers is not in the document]
BusyJay: Perhaps we can use `enum` in this case. Flow control should work on per tablet level, so if there is only one tablet, then it can use singleton, otherwise, it's upgraded to multiple tablets. This is also how `mpsc` implemented in std lib and other channel libraries.

Connor1996 (approved):
LGTM

src/storage/txn/singleton_flow_controller.rs
```diff
@@ -0,0 +1,1584 @@
...
+
+// CFFlowChecker records some statistics and states related to one CF.
+// These statistics fall into five categories:
+//   * memtable
+//   * L0 files
+//   * L0 production flow (flush flow)
+//   * L0 consumption flow (compaction read flow of L0)
+//   * pending compaction bytes
+// And all of them are collected from the hook of RocksDB's event listener.
+struct CFFlowChecker {
```
BusyJay, line 379: It should be `CfFlowChecker`.

src/storage/txn/flow_controller.rs
```diff
@@ -1,1340 +1,83 @@
...
-const MAX_THROTTLE_SPEED: f64 = 200.0 * 1024.0 * 1024.0; // 200MB
-
-const EMA_FACTOR: f64 = 0.6; // EMA stands for Exponential Moving Average
-
-#[derive(Eq, PartialEq, Debug)]
-enum Trend {
-    Increasing,
-    Decreasing,
-    NoTrend,
+pub enum FlowControlType {
```
BusyJay, line 9: Why not name it `FlowControl` directly? So initializing it won't need a double calls.

BusyJay (commented):
Generally LGTM per layout and style part.

src/storage/txn/singleton_flow_controller.rs
tonyxuqqi: singleton_flow_controller.rs is just the git mv from flow_controller.rs. That's the existing code. But I think you're right.

Issue #12849, opened 2022-06-20
Support TiKV using different KvEngines


CalvinNeo:
## Development Task

We need to support TiKV using a KvEngine other than engine_rocks.
1. Add some new coprosessors as complement
2. Add new features to trait engine

#### Observe execution
In AdminObserver and QueryObserver:

1. `pre_exec_admin`/`pre_exec_query` #12868
    Not all KvEngines can handle all of the commands. For example, if a KvEngine is not based on RocksDB, it may not be able to handle ComputeHash/VerifyHash. This observer allows pre-handling these commands, and filtering non-executable commands by returning false. 
3. `post_exec_admin`/`post_exec_query` #12850
    Informs results and modifications(to states) of the command, immediately after the command is executed. It returns true if one of the observes think a persistence is needed to our specific KvEngine, and we will call commit to do such persistence.
4. on_empty_cmd #12851
    Empty commands will be proposed when a new leader starts working, or processing a read-index request during leadership changing. We need to observe this, to ensure `KvEngine` can always has the up-to-date apply state. 
5. Support delay deletion of IngestSst #13061
    Some KvEngine can't actually ingest sst, since they do not use RocksDB as a backend. So they may decode the sst, and store them in memory. If we delete the sst file immediately, we may lost data if we have not persist our data in KvEngine.
    However, we can't do a persistence for every ingest sst request, this just cost too much. So we introduce this PR to allow a delay deletion of sst files, controlled by post_exec.

#### Observer applying snapshots
In ApplySnapshotObserver:
1. `pre_apply_snapshot` #12889
    It is called immediately when RegionRunner accepts one Task::Apply task.
    Since there are already `pre_apply` observer for commands, a `pre_apply` observer for applying snapshots is reasonable. We can do some statistics, transform snapshot files, batch snapshots by this observer, mostly in background threads.
2. `post_apply_snapshot` also #12889
    It is called when the snapshot is applied.
    Currently, there are only observers to inform the after when we handle a cf file. We need another observer when the whole snapshot is applied.

#### Observe RegionChange
1. `pre_commit` #12957
    Before commit and finish_for, if the observer decides it is not a proper time to do actual persist, it can skip this persistence.
    This observer is aim to disable persistence of advanced `applied_index`, since some KvEngine may not persist replicated data when `commit` and `finish_for`, so we can persist advanced `applied_index` either. 
[...]

PR #12850, opened 2022-06-20
raftstore: Implement coprocessor observer post_exec_admin(query)

CalvinNeo:

Issue Number: ref #12849

What's Changed:

According to #12849, we need to support some new observers, including post_exec_admin(query) which is implemented in this PR.
```commit-message
Support new observers post_exec_admin(query).
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1323,6 +1346,10 @@ where
             }
         }
 
+        if should_write {
+            info!("persist data and apply state"; "region_id" => self.region_id(), "peer_id" => self.id(), "state" => ?self.apply_state);
+            ctx.commit(self);
```
CalvinNeo, line 1351: Should we move this into an outer function?

CalvinNeo:
> Add some test for changes in apply.rs

Will do that after pre_exec_query(admin) are merged.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -5491,6 +5550,102 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_exec_observer() {
```
tonyxuqqi, line 5554: add some comments for the test steps and the logic to verify.

tonyxuqqi (approved):

components/raftstore/src/store/fsm/apply.rs
CalvinNeo: Will do after pre_exec is merged

components/cdc/src/delegate.rs
```diff
@@ -446,6 +446,7 @@ impl Delegate {
         for cmd in batch.into_iter(self.region_id) {
             let Cmd {
                 index,
+                term: _,
```
BusyJay, line 449: What's this?

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -416,6 +416,33 @@ impl<E: KvEngine> CoprocessorHost<E> {
         }
     }
 
+    pub fn post_exec(
```
BusyJay, line 419: Why not use `post_apply`?

components/cdc/src/delegate.rs
CalvinNeo: We don't use `term` here.

components/raftstore/src/coprocessor/dispatcher.rs
CalvinNeo: The existing `post_apply` observer is not enough. Only `pre_exec` can observe more necessary execution information we need, at a earlier time than `post_apply`.

BusyJay: Please add some comment to describe the difference.

CalvinNeo: addressed

components/raftstore/src/store/fsm/apply.rs
BusyJay: Yes, otherwise the response won't be returned for the last command.

BusyJay (commented):
I think we need to bench the performance impact after adding two observers. sysbench should be enough.

CalvinNeo:
> I think we need to bench the performance impact after adding two observers. sysbench should be enough.
It is tested that the performance is quite close to each other.
VHH72SaHAz

BusyJay (approved):

PR #12851, opened 2022-06-20
raftstore: Implement coprocessor observer on_empty_cmd

CalvinNeo:

Issue Number: ref #12849

What's Changed:

```commit-message
Support new observers on_empty_cmd.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


BusyJay:
Any test case?

CalvinNeo:
> Any test case?

I have tests in `test_trigger_right_hook`. Since `on_empty_cmd` will not affect any procedures, so I think we only need to ensure it is called at the right time.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1077,7 +1077,9 @@ where
 
             return self.process_raft_cmd(apply_ctx, index, term, cmd);
         }
-        // TOOD(cdc): should we observe empty cmd, aka leader change?
+
+        // we should observe empty cmd, aka leader change
```
BusyJay, line 1081: Inaccurate comment. Empty command doesn't have to be associated with leadership change.

CalvinNeo: try to fix

zhangjinpeng1987 (approved):
LGTM

BusyJay (approved):

PR #12866, opened 2022-06-21
cdc: Resolved-ts for RawKV

haojinming:

Issue Number: Ref #11965 

What's Changed:

Impl resolve ts for rawkv cdc process.
1. Use `CdcObserver` as `RawTsTracker`.
2. Reuse `resolver` to resolve rawkv apiv2  timestamp(non-decreasing for one same region). 
3. Add `RawTrackLock` & `RawUntrackLock` for rawkv, which use BinaryHeap<Reverse<Timestamp>>, untrack will unlock all tiemstamp before input ts. 
4. `RawTrackLock` in  CausalObserver::pre_propose && `RawUntrackLock`  in `sink_rawkv_downstream`.
5. Add enum `PendingLock::RawTrack` & `PendingLock::RawUntrack` to store the pending lock info for rawkv.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/causal_ts/src/observer.rs
```diff
@@ -95,14 +100,17 @@ impl<Ts: CausalTsProvider> QueryObserver for CausalObserver<Ts> {
             }
 
             ApiV2::append_ts_on_encoded_bytes(req.mut_put().mut_key(), ts.unwrap());
+            self.ts_tracker
```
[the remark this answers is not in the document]
haojinming: New implementation is 
1. define new `raw_track_lock` & `raw_untrack_lock` which use a new BinaryMap<TimeStamp>. 
2. In `CausalObserver::pre_propose_query`, track the first ts in one batch.
3. In `Delegate::sink_data`, untrack the last ts in one batch.
4. This implementation depends on the assumption that timestamp in one same region is non-decreasing, which is  guranteed now.
5. For the corner case, we use lock ttl(default 60s), if the resolved_ts in one region is not advanced within 60s, untrack the ts before "min_ts - 60s".

PTAL~ @pingyu

components/cdc/src/delegate.rs
```diff
@@ -613,12 +606,33 @@ impl Delegate {
...
+        self.sink_raw_downstream(raw_rows, index)?;
 
         Ok(())
     }
 
+    fn sink_raw_downstream(&mut self, entries: Vec<EventRow>, index: u64) -> Result<()> {
+        if entries.is_empty() {
+            return Ok(());
+        }
+        let max_raw_ts = entries
```
pingyu, line 618: The last entry should have the `max_raw_ts`. So we don't need to iterate all entries.

components/resolved_ts/src/resolver.rs
```diff
@@ -169,6 +200,19 @@ impl Resolver {
                 .inc();
         }
 
+        let min_raw_ts = self
```
pingyu, line 203: Should be placed before `if self.resolved_ts >= new_resolved_ts {` to make `RTS_RESOLVED_FAIL_ADVANCE_VEC` metrics correct.

Suggest to add metrics for the force untrack.
Suggest to make `RAW_LOCK_TS_TTL` configurable.

components/cdc/src/endpoint.rs
```diff
@@ -1019,6 +1028,12 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
                 }
                 min_ts_min_lock = min_mem_lock_ts;
             }
+            // If flush_causal_timestamp fails, cannot schedule RegisterMinTsEvent task
+            // as new coming raw data may use timestamp smaller than min_ts
+            if let Err(e) = observer.flush_causal_timestamp() {
+                error!("cdc flush causal timestamp failed"; "err" => ?e);
```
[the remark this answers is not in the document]
haojinming: done

components/cdc/src/delegate.rs
haojinming: done

components/cdc/src/endpoint.rs
```diff
@@ -1122,6 +1138,16 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
     fn on_open_conn(&mut self, conn: Conn) {
         self.connections.insert(conn.get_id(), conn);
     }
+
+    fn on_raw_track_ts(&mut self, region_id: u64, ts: TimeStamp) {
+        if let Some(ref mut delegate) = self.capture_regions.get_mut(&region_id) {
+            if let Some(resolver) = delegate.resolver.as_mut() {
```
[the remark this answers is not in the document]
haojinming: sure, it makes more sense to be in `delagate`.

components/resolved_ts/src/resolver.rs
```diff
@@ -176,7 +196,7 @@ impl Resolver {
     ///
     /// `min_ts` advances the resolver even if there is no write.
     /// Return None means the resolver is not initialized.
-    pub fn resolve(&mut self, min_ts: TimeStamp) -> TimeStamp {
+    pub fn resolve(&mut self, min_ts: TimeStamp) -> ResolvedTs {
```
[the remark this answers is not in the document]
haojinming: I got the point. The hard thing is how to detect the dead lock in single region's `Resolver.resolve` or new added `raw_resolve` func with Box_plot algorithm, which needs all regions' resolved_ts info.

haojinming:
Return `ResolvedTs` in `Resolver.resolve` and use extreme outlier algorithm to detect dead lock in raw regions. PTAL~ @pingyu @overvenus , thanks.

components/resolved_ts/src/resolver.rs
```diff
@@ -1,13 +1,40 @@
...
+
+impl ResolvedTs {
+    pub fn default() -> ResolvedTs {
+        ResolvedTs {
+            raw_ts: TimeStamp::zero(),
+            txn_ts: TimeStamp::zero(),
+        }
+    }
+    pub fn get_min(&self) -> TimeStamp {
+        cmp::min(self.raw_ts, self.txn_ts)
```
overvenus, line 30: Does it need to skip `TimeStamp::zero()`?

components/server/src/server.rs
```diff
@@ -814,6 +797,23 @@ impl<ER: RaftEngine> TiKvServer<ER> {
...
 
+        // Register causal observer for RawKV API V2
+        if let ApiVersion::V2 = F::TAG {
+            let tso = block_on(causal_ts::BatchTsoProvider::new_opt(
+                self.pd_client.clone(),
+                self.config.causal_ts.renew_interval.0,
+                self.config.causal_ts.renew_batch_min_size,
+            ));
+            if let Err(e) = tso {
+                panic!("Causal timestamp provider initialize failed: {:?}", e);
```
overvenus, line 808:
```suggestion
                fatal!("Causal timestamp provider initialize failed: {:?}", e);
```

components/cdc/src/endpoint.rs
```diff
@@ -306,6 +319,47 @@ impl ResolvedRegionHeap {
...
+            {
+                return vec![self.heap.peek().unwrap().0.region_id];
+            }
+        } else {
+            let ts_vec: Vec<TimeStamp> = self.heap.iter().map(|iter| iter.0.resolved_ts).collect();
+            let size = ts_vec.len();
+            let q1_ts = ts_vec[(size + 1) / 4];
+            let q3_ts = ts_vec[3 * (size + 1) / 4];
+            let delta = q3_ts.physical().saturating_sub(q1_ts.physical());
+            let outliers: Vec<u64> = self
```
overvenus, line 343: Is there any test covers the algorithm?

haojinming: yes, test cases are added in test_resolved_region_heap.

components/resolved_ts/src/resolver.rs
haojinming: `raw_ts` and `txn_ts` are initialized as `TimeStamp::zero()` and updated every MinTs task, they all never decrease. So I think they must not be `zero` returned from 'Resolver.resolve'.

components/cdc/src/endpoint.rs
```diff
@@ -306,6 +319,47 @@ impl ResolvedRegionHeap {
...
+        } else if self.heap.len() < 4 {
+            // if count is smaller than 4, just return one outlier if exists.
+            let resolved = self.heap.peek().unwrap().0.resolved_ts;
+            if Duration::from_millis(min_ts.physical().saturating_sub(resolved.physical()))
+                > RAW_RESOLVED_TS_OUTLIER_THRESHOLD
+            {
+                return vec![self.heap.peek().unwrap().0.region_id];
+            }
+        } else {
+            let ts_vec: Vec<TimeStamp> = self.heap.iter().map(|iter| iter.0.resolved_ts).collect();
```
pingyu, line 338: It would be more efficient to gather all raw resolved-ts and sort, than use a heap.

components/cdc/src/endpoint.rs
```diff
@@ -306,6 +319,47 @@ impl ResolvedRegionHeap {
         self.clear();
         self.heap.shrink_to(min_capacity);
     }
+
+    // extreme outier match the following two conditions:
+    // 1. https://en.wikipedia.org/wiki/Box_plot
+    // 2. the gap with min_ts is larger than RAW_RESOLVED_TS_OUTLIER_THRESHOLD.
+    fn get_extreme_outlier(&self, min_ts: TimeStamp) -> Vec<u64> {
```
pingyu, line 326: I think we can only judge the smallest timestamp, to reduce code complexity. The deadlock is rare.

components/cdc/src/endpoint.rs
```diff
@@ -819,10 +869,37 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
...
+                let observe_id = delegate.handle.id;
+                let deregister = Deregister::Delegate {
+                    region_id,
+                    observe_id,
+                    err: Error::Other(box_err!("raw region dead lock")),
+                };
+                if let Err(e) = self.scheduler.schedule(Task::Deregister(deregister)) {
+                    error!("cdc schedule cdc task failed"; "error" => ?e);
+                }
+                CDC_RAW_REGION_OUTLIER_COUNT.inc();
```
pingyu, line 890: Suggest to also report the duration histogram of outlier raw resolved-ts.

components/cdc/src/endpoint.rs
```diff
@@ -840,13 +917,16 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
...
                 if let Some(resolved_ts) = delegate.on_min_ts(min_ts) {
-                    if resolved_ts < self.min_resolved_ts {
-                        self.min_resolved_ts = resolved_ts;
+                    if resolved_ts.get_min() < self.min_resolved_ts {
+                        self.min_resolved_ts = resolved_ts.get_min();
                         self.min_ts_region_id = region_id;
                     }
-                    resolved_regions.push(region_id, resolved_ts);
+                    resolved_regions.push(region_id, resolved_ts.get_min());
+                    if resolved_ts.is_min_ts_from_raw() {
```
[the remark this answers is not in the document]
haojinming: Actually at max one region can have mixed `raw/tidb` data. If we add tidb region's `raw_ts` into heap, tidb's region maybe regarded as outlier as its transaction maybe takes much time. So the `is_min_ts_from_raw` is used to seperate the region is from `raw` or `tidb`.

components/cdc/src/endpoint.rs
haojinming: yes, `ResolvedRegionHeap` is reused to store raw region resolved_ts info here.

components/cdc/src/endpoint.rs
```diff
@@ -306,6 +319,47 @@ impl ResolvedRegionHeap {
...
     }
+
+    // extreme outier match the following two conditions:
+    // 1. https://en.wikipedia.org/wiki/Box_plot
+    // 2. the gap with min_ts is larger than RAW_RESOLVED_TS_OUTLIER_THRESHOLD.
+    fn get_extreme_outlier(&self, min_ts: TimeStamp) -> Vec<u64> {
+        if self.heap.is_empty() {
+            return vec![];
+        } else if self.heap.len() < 4 {
+            // if count is smaller than 4, just return one outlier if exists.
```
pingyu, line 330: I suggest not to detect the outlier when number of regions less than 4 or a threshold (maybe 10 ?). When the number is small, the confidence of outlier detection is low.
As there is no good method to do so now, just don't do it.
Moreover, this scenario should not happen in a production environment.

components/cdc/src/endpoint.rs
haojinming: added.

components/cdc/src/endpoint.rs
haojinming: ok, just detect when region number is more than 10.

components/causal_ts/src/lib.rs
```diff
@@ -55,4 +59,19 @@ pub mod tests {
             Ok(self.ts.fetch_add(1, Ordering::Relaxed).into())
         }
     }
+
+    #[derive(Clone)]
```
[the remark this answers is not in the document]
haojinming: done

components/cdc/src/endpoint.rs
```diff
@@ -70,6 +70,7 @@ const METRICS_FLUSH_INTERVAL: u64 = 10_000; // 10s
 const WARN_RESOLVED_TS_LAG_THRESHOLD: Duration = Duration::from_secs(600);
 // Suppress repeat resolved ts lag warning.
 const WARN_RESOLVED_TS_COUNT_THRESHOLD: usize = 10;
+const RAW_RESOLVED_TS_OUTLIER_THRESHOLD: Duration = Duration::from_secs(60);
```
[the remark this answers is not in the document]
haojinming: I'm not sure it's suitable to make it visible in tikv config file as it's an inner paramemter. User may not understand what happend when this parameter changes.

components/causal_ts/src/lib.rs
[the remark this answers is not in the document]
haojinming: done

components/resolved_ts/src/resolver.rs
```diff
@@ -1,13 +1,40 @@
...
+}
+
+impl ResolvedTs {
+    pub fn default() -> ResolvedTs {
+        ResolvedTs {
+            raw_ts: TimeStamp::zero(),
+            txn_ts: TimeStamp::zero(),
+        }
+    }
+    pub fn get_min(&self) -> TimeStamp {
```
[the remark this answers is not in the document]
haojinming: done

components/cdc/src/endpoint.rs
haojinming: Added ResolvedRegionVec to improve efficiency.

components/cdc/src/endpoint.rs
haojinming: Done, return one maximally. As MinTs task will run periodically, all regions can also be resolved even if more than one regions have dead lock.

components/backup-stream/src/subscription_track.rs
```diff
@@ -304,7 +304,7 @@ impl TwoPhaseResolver {
             return min_ts.min(stable_ts);
         }
 
-        self.resolver.resolve(min_ts)
+        self.resolver.resolve(min_ts).get_min()
```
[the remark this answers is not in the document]
YuJuncen: LGTM

components/cdc/src/endpoint.rs
```diff
@@ -819,10 +874,35 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
         }
     }
 
+    // detect outlier raw regions, schedule deregister for outlier raw regions.
+    fn handle_raw_outlier_regions(
+        &self,
+        raw_resolved_regions: &mut ResolvedRegionVec,
+        min_ts: TimeStamp,
+    ) {
+        if let Some(region) = raw_resolved_regions.get_extreme_outlier(min_ts) {
```
overvenus, line 883: Could you add some log when it's going to deregister a region?

components/cdc/src/endpoint.rs
```diff
@@ -819,10 +874,35 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
...
+            if let Some(delegate) = self.capture_regions.get(&region.region_id) {
+                let observe_id = delegate.handle.id;
+                let deregister = Deregister::Delegate {
+                    region_id: region.region_id,
+                    observe_id,
+                    err: Error::Other(box_err!("raw region dead lock")),
```
overvenus, line 889: Could you test if TiCDC recovers automatically when a region is deregister by this error?

components/cdc/src/endpoint.rs
```diff
@@ -819,10 +874,39 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
...
+                };
+                warn!(
+                    "deregister raw region {} as resolved_ts has much lag, dead lock may occurs.",
```
overvenus, line 892:
```suggestion
                    "cdc deregister raw region {} as resolved_ts has much lag, dead lock may occurs.",
```

components/cdc/src/endpoint.rs
haojinming: I trigger the deregister process manually, TiCDC can register the region after deregister and has no effect on resolved_ts. All process is working well.

components/cdc/src/endpoint.rs
haojinming: done

components/cdc/src/delegate.rs
```diff
@@ -613,10 +608,41 @@ impl Delegate {
...
+            return Ok(());
+        }
+        // the entry's timestamp is non-decreasing, the last has the max ts.
+        let max_raw_ts = TimeStamp::from(entries.last().unwrap().commit_ts);
+        match self.resolver {
+            Some(ref mut resolver) => {
+                resolver.raw_untrack_lock(max_raw_ts);
+            }
+            None => {
+                assert!(self.pending.is_some(), "region resolver not ready");
```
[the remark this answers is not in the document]
haojinming: add pendinglock::rawtrackts test case: [`test_raw_pending_lock`](https://github.com/haojinming/tikv/blob/4c33fe74cfe898b683e347bc2a8bd711517e380d/components/cdc/src/endpoint.rs#L2048), RawUntrackLock is a bit hard to test, I would like to add it later in next pr, this pr is a little large.

components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -216,6 +216,9 @@ fn test_cdc_rawkv_basic() {
             Box::new(|delegate| {
                 let d = delegate.unwrap();
                 assert_eq!(d.downstreams().len(), 1);
+                let resolver = d.resolver.as_ref().unwrap();
+                let lock_keys = resolver.locks().keys();
+                assert_eq!(lock_keys.len(), 0);
```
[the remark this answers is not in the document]
haojinming: this case is not suitable to add judgement here as raw does not use `resolver.locks()`, test cases are added in `test_raw_track_ts` and `test_raw_dead_lock`

components/cdc/src/endpoint.rs
```diff
@@ -1837,6 +1950,191 @@ mod tests {
...
+        suite.run(Task::MinTS {
+            regions: region_ids,
+            min_ts: u64::MAX.into(),
+        });
+        let task_recv = suite
+            .task_rx
+            .recv_timeout(Duration::from_millis(500))
+            .unwrap()
+            .unwrap();
+        if let Task::Deregister(Deregister::Delegate {
```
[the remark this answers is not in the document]
haojinming: great suggestion, thanks.

components/causal_ts/src/observer.rs
```diff
@@ -97,8 +97,9 @@ impl<Ts: CausalTsProvider, Tk: RawTsTracker> QueryObserver for CausalObserver<Ts
                 ts = Some(self.causal_ts_provider.get_ts().map_err(|err| {
                     coprocessor::Error::Other(box_err!("Get causal timestamp error: {:?}", err))
                 })?);
+                // use prev ts as `resolved_ts` means the data with smaller or equal ts has already sink to cdc.
                 self.ts_tracker
-                    .track_ts(region_id, ts.unwrap())
+                    .track_ts(region_id, ts.unwrap().prev())
```
pingyu, line 102: Suggest to add test case for this change.

haojinming: I will add test in later PR with metric json, it's not easy to test this scenario.

overvenus (approved):

PR #12884, opened 2022-06-23
TiKV: support tablet in DBConfigManager

### LINKED ISSUE #12883: TiKV online config change adaption for MultiRocksDB
## Development Task
Now we have Tablet abstraction,  the DBConfigManger should work with tablets as well as single kv engine.

tonyxuqqi:

Issue Number: Close #12883

What's Changed:

```commit-message
DBConfigManager should work with tablets, meanwhile compatible with current single kv engine. 
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


src/config.rs
```diff
@@ -1542,33 +1565,78 @@ impl DBConfigManger {
...
+                        error!(
+                            "set_block_cache_size {} failed on tablet {} {}. err:{:?}",
+                            size.as_mb(),
+                            region_id,
+                            suffix,
+                            &result
+                        );
+                    }
+                } else {
+                    result = Err(Box::from(r.err().unwrap()));
```
BornChanger, line 1585: Do we need to print error log for failure get_options_cf() ?

tonyxuqqi: good catch. I think so.

src/config.rs
```diff
@@ -1542,33 +1565,82 @@ impl DBConfigManger {
...
+        self.tablet_factory
+            .loop_tablet_cache(Box::new(|region_id, suffix, db: &RocksEngine| {
+                let r = db.get_options_cf(cf);
+                if let Ok(opt) = r {
+                    let r = opt.set_block_cache_capacity(size.0);
+                    if r.is_err() {
+                        result = Err(Box::from(r.err().unwrap()));
+                        error!(
+                            "opt.set_block_cache_capacity {} failed on tablet {}_{}. err:{:?}",
+                            size.as_mb(),
```
ethercflow, line 1578: Do we need to print error log with `cf` field?

ethercflow (approved):
LGTM

src/config.rs
```diff
@@ -1542,33 +1565,83 @@ impl DBConfigManger {
...
+        result
     }
 
     fn set_rate_bytes_per_sec(&self, rate_bytes_per_sec: i64) -> Result<(), Box<dyn Error>> {
-        let mut opt = self.db.as_inner().get_db_options();
-        opt.set_rate_bytes_per_sec(rate_bytes_per_sec)?;
-        Ok(())
+        let mut result = Ok(());
+        self.tablet_factory
+            .loop_tablet_cache(Box::new(|region_id, suffix, db: &RocksEngine| {
```
BusyJay, line 1603: rate limiter should be shared, so this is unnecessary to loop.

src/config.rs
```diff
@@ -1542,33 +1565,83 @@ impl DBConfigManger {
                  block-cache.capacity in storage module instead"
                 .into());
         }
-        let opt = self.db.get_options_cf(cf)?;
-        opt.set_block_cache_capacity(size.0)?;
+        let mut result = Ok(());
+        self.tablet_factory
```
BusyJay, line 1569: Cache is also shared.

src/config.rs
tonyxuqqi: The current implementation of rate limiter is not shared. It's per DB. Once it's changed to the shared one, then we may change the code here as well.

src/config.rs
tonyxuqqi: When the cache is shared, it won't go to this code path ( there's a shortcut above)

BusyJay: Cache must be shared.  Making every tablet use its owned cache will not work in practice. `shared_block_cache` is used for controlling cache between raft db and kv db in the past, not for tablets.

src/config.rs
BusyJay: Rate limiter should be shared, every db has its own rate limiter actually means no limit.

BusyJay:
The whole point of extracting factory is to make sharing easier. 🤣

src/config.rs
tonyxuqqi: As discussed offline, if multirocks enabled, the shared cache has to be enabled (which is the default now)---we would need to add the validation once the new flag is ready.
So the loop_tablet_cache code here only will run for single rocks db, which is totally fine.

components/engine_rocks/src/engine.rs
```diff
@@ -187,6 +194,72 @@ impl SyncMutable for RocksEngine {
...
+    }
+    fn destroy_tablet(&self, _id: u64, _suffix: u64) -> Result<()> {
+        Ok(())
+    }
+    fn exists_raw(&self, _path: &Path) -> bool {
+        true
+    }
+    fn tablet_path(&self, _id: u64, _suffix: u64) -> PathBuf {
+        PathBuf::from(&self.root_path)
+    }
```
zhangjinpeng1987, line 220: Please add blank line between methods.

components/engine_rocks/src/engine.rs
```diff
@@ -187,6 +194,72 @@ impl SyncMutable for RocksEngine {
...
+
+    fn loop_tablet_cache(&self, mut f: Box<dyn FnMut(u64, u64, &RocksEngine) + '_>) {
+        if let Some(engine) = &self.engine {
+            f(0, 0, engine);
+        }
+    }
+
+    fn set_shared_db_block_cache_size(&self, cf: &str, size: ReadableSize) -> Result<()> {
+        let opt = self.engine.as_ref().unwrap().get_options_cf(cf)?;
+        opt.set_block_cache_capacity(size.0)
```
zhangjinpeng1987, line 246: Seems you don't set the opt back to the engine?

components/engine_rocks/src/engine.rs
```diff
@@ -187,6 +194,72 @@ impl SyncMutable for RocksEngine {
...
+                root_path: self.root_path.clone(),
+            });
+        }
+        Box::<DummyRocksEngineFactory>::new(DummyRocksEngineFactory {
+            engine: Some(self.engine.as_ref().unwrap().clone()),
+            root_path: self.root_path.clone(),
+        })
+    }
+
+    fn loop_tablet_cache(&self, mut f: Box<dyn FnMut(u64, u64, &RocksEngine) + '_>) {
```
zhangjinpeng1987, line 238: Maybe "foreach_tablet" is more clear?

src/config.rs
```diff
@@ -1552,23 +1576,50 @@ impl DBConfigManger {
...
 
     fn set_rate_bytes_per_sec(&self, rate_bytes_per_sec: i64) -> Result<(), Box<dyn Error>> {
-        let mut opt = self.db.as_inner().get_db_options();
-        opt.set_rate_bytes_per_sec(rate_bytes_per_sec)?;
-        Ok(())
+        let mut result = Ok(());
+        self.tablet_factory
+            .loop_tablet_cache(Box::new(|region_id, suffix, db: &RocksEngine| {
+                let mut opt = db.as_inner().get_db_options();
+                let r = opt.set_rate_bytes_per_sec(rate_bytes_per_sec);
```
zhangjinpeng1987, line 1583: forget to set back to db?

components/engine_rocks/src/engine.rs
tonyxuqqi: it's the ref and no need to setback.

components/engine_rocks/src/engine.rs
tonyxuqqi: ok. Either is fine to me.

src/config.rs
tonyxuqqi: Shared rate limiter API is not ready in rust-rocks. I'd like to postpone the changes to rust-rocks due to the development of tirocks.

src/config.rs
```diff
@@ -1494,45 +1493,70 @@ pub enum DBType {
...
-            if let Ok(v) = cfg_value {
-                CONFIG_ROCKSDB_GAUGE
-                    .with_label_values(&[cf, cfg_name])
-                    .set(v);
+        self.tablet_factory.for_each_opened_tablet(Box::new(
+            |region_id, suffix, db: &RocksEngine| {
+                let r = db.set_options_cf(cf, opts);
+                if r.is_err() {
+                    result = Err(Box::from(r.err().unwrap()));
+                    error!(
```
BusyJay, line 1536: If it fails on all db, then it will write region count lines of logs? Doesn't seem to be reasonable.

src/config.rs
```diff
@@ -1552,23 +1577,52 @@ impl DBConfigManger {
...
-            return Err("fail to set rate_limiter_auto_tuned".into());
-        }
-        Ok(())
+        let mut result = Ok(());
+        self.tablet_factory.for_each_opened_tablet(Box::new(
+            |region_id, suffix, db: &RocksEngine| {
+                let mut opt = db.as_inner().get_db_options();
+                let r = opt.set_auto_tuned(rate_limiter_auto_tuned);
+                if r.is_err() {
+                    result = Err(Box::from(r.err().unwrap()));
```
BusyJay, line 1607: It fails to set the configuration already, seems meaningless to double check later.

components/server/src/server.rs
```diff
@@ -1531,7 +1531,14 @@ impl ConfiguredRaftEngine for RocksEngine {
     fn register_config(&self, cfg_controller: &mut ConfigController, share_cache: bool) {
         cfg_controller.register(
             tikv::config::Module::Raftdb,
-            Box::new(DBConfigManger::new(self.clone(), DBType::Raft, share_cache)),
+            Box::new(DBConfigManger::new(
+                Arc::new(DummyRocksEngineFactory {
```
BusyJay, line 1535: Why not use v1 factory?

src/config.rs
tonyxuqqi: These APIs call should never fail in legitimate scenarios.  I would rather see the the error! itself have some mechanism to ignore the error messages instead of doing some limit in the caller side everywhere.

BusyJay: > the error! itself have some mechanism to ignore the error messages 

How?

tonyxuqqi: Basically we need to know all the failed tablet/cfs that do not apply the new config.

BusyJay: Do we really need to know that? Isn't know the failure count and some of errors sufficient?

src/config.rs
```diff
@@ -1494,45 +1494,68 @@ pub enum DBType {
...
+                    error!(
+                        "set_cf_config {} failed on tablet {}_{}. err:{:?}",
+                        cf, region_id, suffix, &result
+                    );
+                }
+            }));
+
+        if result.is_ok() {
```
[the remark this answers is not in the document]
tonyxuqqi: i use result?; initially , but it will run into this error.
error[E0283]: type annotations needed
    --> src/config.rs:1535:34
     |
1535 |                     result = Err(Box::from(r.err().unwrap()));
     |                                  ^^^^^^^^^ cannot infer type for struct `Box<_, _>`
     |
     = note: multiple `impl`s satisfying `Box<_, _>: From<engine_traits::Error>` found in the following crates: `alloc`, `std`:
             - impl<'a, E> From<E> for Box<(dyn StdError + 'a)>
               where E: 'a, E: StdError;
             - impl<'a, E> From<E> for Box<(dyn StdError + std::marker::Send + std::marker::Sync + 'a)>
               where E: 'a, E: StdError, E: std::marker::Send, E: std::marker::Sync;
             - impl<T> From<T> for Box<T>;

error[E0283]: type annotations needed
    --> src/config.rs:1544:15
     |
1544 |         result?;
     |               ^ cannot infer type for struct `Box<_, _>`

src/config.rs
tonyxuqqi: good point.

components/server/src/server.rs
tonyxuqqi: v1 factory is simply cannot be built in this context. It requires KvEngineFactoryBuilder which will need other structs that are only available during initialization. 
On the other hand, v1 factory is designed to really create a db. But here the db already exists, and we just need a wrapper around it to unify the code.

BusyJay: Factory also provides a way to query actual tablet, so accessing engine through factory is OK. I think it can be passed from initialization to config.

tonyxuqqi: Discussed offline. The best way is to extract a dedicate trait TabletAccessor for accessing tablets, either from single engine or multi-engines.

components/engine_traits/src/engine.rs
```diff
@@ -61,11 +61,24 @@ pub trait KvEngine:
     fn bad_downcast<T: 'static>(&self) -> &T;
 }
 
+/// TabletAccessor is the trait to access all the tablets with provided accessor
+///
+/// For single rocksdb instance, it essentially accesses the global kvdb with the accessor
+/// For multi rocksdb instances, it accesses all the tablets with the accessor
+pub trait TabletAccessor<EK> {
+    /// Loop visit all opened tablets by the specified function.
+    fn for_each_opened_tablet(&self, _f: Box<dyn FnMut(u64, u64, &EK) + '_>);
```
BusyJay, line 70: How about `impl FnMut(u64, u64, &EK)`?

tonyxuqqi: no, &EK is from the callee, not the caller. But the FnMut is from the caller's behavior and it can be anything.

components/engine_traits/src/engine.rs
```diff
@@ -61,11 +66,92 @@ pub trait KvEngine:
...
+
+    /// return true if it's single engine;
+    /// return false if it's a multi-tablet factory;
+    fn is_single_engine(&self) -> bool;
+}
+
+/// max error count to log
+const MAX_ERROR_COUNT: u32 = 5;
+
+pub struct TabletErrorCollector {
```
BusyJay, line 85: Should add some comment explaining what it does.

BusyJay (approved):

Issue #12848, opened 2022-06-20
Improve the paging protocol


wshwsh12:
## Development Task

Now tidb supports using paging protocol to process index scan requests, and look up table early to reduce query latency.
But now only support index scan and not enabled by default. We can improve the paging implement to bring more benefits.

## Goal
1. Solve the **OOM problem** of reading a large amount of data such as full table scan through the paging protocol
2. Paging supports common scenarios and becomes the default protocol
3. Only one implementation is maintained. The current coprocessor protocol degenerates into a special subset of the paging protocol

## Todo list (in tikv part)

- Executors support paging protocal. We need to make some modifications to the implementation of the executors to ensure correctness.
    - [ ] Support aggregate executor
    - [ ] Support limit executor
    - [ ] Support topn executor
- Modify the way to get the end key for the read data executor. Now we need update end key for each row. So the performance is terrible.

PR #12886, opened 2022-06-23
copr: update scanned range for each batch when paging is enable

wshwsh12:

Issue Number: ref https://github.com/tikv/tikv/issues/12848

What's Changed: 
Now we need update end key for each row. The performance is terrible.
I modify the update frequency, so update the scanned range for each batch.

```commit-message
Update scanned range for each batch when paging is enable
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -65,6 +65,15 @@ impl<T: Storage> RangesScanner<T> {
     // Note: This is not implemented over `Iterator` since it can fail.
     // TODO: Change to use reference to avoid allocation and copy.
     pub fn next(&mut self) -> Result<Option<OwnedKvPair>, StorageError> {
+        self.next_row(true)
```
sticnarf, line 68: Why just pass `true` instead of `self.is_scanned_range_aware`?

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -65,6 +65,15 @@ impl<T: Storage> RangesScanner<T> {
...
+    }
+
+    /// Fetches next row.
+    /// Note: `update_scanned_range` can control whether update the scanned range when `is_scanned_range_aware` is true.
+    pub fn next_row(
```
sticnarf, line 73: The name difference between `next` and `next_row` may be confusing. I don't have a good idea, but at least `next_opt` is better?

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -666,4 +675,72 @@ mod tests {
...
+            scan_backward_in_range: false,
+            is_key_only: false,
+            is_scanned_range_aware: true,
+        });
+
+        assert_eq!(&scanner.next_row(false).unwrap().unwrap().0, b"foo");
+        // Only lower_inclusive is updated.
+        assert_eq!(&scanner.working_range_begin_key, b"foo");
+        assert_eq!(&scanner.working_range_end_key, b"");
+        assert_eq!(&scanner.next_row(true).unwrap().unwrap().0, b"foo_2");
```
sticnarf, line 696: I recommend adding a blank line before this line to spearate each case.

sticnarf (approved):

you06:
Is it possible to reduce the call of `update_scanned_range_from_new_point`?

https://github.com/tikv/tikv/blob/fb2330abcd2325041a95ec4dff9aebe7dd86b63f/components/tidb_query_common/src/storage/scanner.rs#L82

wshwsh12:
> Is it possible to reduce the call of `update_scanned_range_from_new_point`?
> 
> https://github.com/tikv/tikv/blob/fb2330abcd2325041a95ec4dff9aebe7dd86b63f/components/tidb_query_common/src/storage/scanner.rs#L82

Yes, but not necessary. One update for each point range is acceptable, I think.

sticnarf:
> Yes, but not necessary. One update for each point range is acceptable, I think.

Emm, index lookup can include lots of point ranges. Maybe it's better to optimise it.

wshwsh12:
Now `update_scanned_range_from_xxx` only update `current_range` for the first range and the last range. PTAL again. @sticnarf @you06

you06 (approved):

PR #12889, opened 2022-06-23
raftstore: Implement coprocessor observer pre(post)_apply_snapshot

CalvinNeo:

Issue Number: ref #12849

What's Changed:

```commit-message
Support new observers pre(post)_apply_snapshot.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/store/worker/region.rs
```diff
@@ -620,6 +638,44 @@ where
...
+                status,
+                peer_id,
+            } => (region_id, status.clone(), peer_id),
+            _ => panic!("invalid apply snapshot task"),
+        };
+
+        let region_state = self.get_region_state(*region_id)?;
+        let apply_state = self.get_apply_state(*region_id)?;
+
+        let abort = status.clone();
```
tonyxuqqi, line 655: why status.clone() here.

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -674,6 +711,28 @@ mod tests {
...
         }
+
+        fn pre_apply_snapshot(
+            &self,
+            ctx: &mut ObserverContext<'_>,
+            _: u64,
+            _: &SnapKey,
+            _: Option<&Snapshot>,
+        ) {
+            self.called.fetch_add(17, Ordering::SeqCst);
```
tonyxuqqi, line 722: any comments for this odd 17 number?

CalvinNeo: There is an unique number for each observer in this mock `TestCoprocessor`. 15/16 is for pre(post)_exec_query(admin) in other PRs.

components/raftstore/src/store/worker/region.rs
```diff
@@ -620,6 +635,46 @@ where
...
+            &snap_key,
+            Some(&s),
+        );
+        info!(
+            "pre apply snapshot";
+            "region_id" => region_id,
+            "peer_id" => peer_id,
+            "state" => ?apply_state,
+            "time_takes" => ?timer.saturating_elapsed(),
+        );
```
tonyxuqqi, line 675: does it need to be info!?
I think maybe debug! and add a pre-snapshot duration metrics makes more sense.

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -168,6 +168,28 @@ pub trait ApplySnapshotObserver: Coprocessor {
     /// Hook to call after applying sst file. Currently the content of the
     /// snapshot can't be passed to the observer.
     fn apply_sst(&self, _: &mut ObserverContext<'_>, _: CfName, _path: &str) {}
+
+    /// Hook when receiving Task::Apply.
+    /// Should pass valid _snapshot, the option is only for testing.
```
BusyJay, line 173: Why underscore?

components/raftstore/src/store/mod.rs
```diff
@@ -21,7 +21,7 @@ mod peer_storage;
 mod read_queue;
 mod region_snapshot;
 mod replication_mode;
-mod snap;
+pub mod snap;
```
BusyJay, line 24: Why?

components/raftstore/src/store/worker/region.rs
```diff
@@ -363,6 +360,31 @@ where
...
+                        log_wrappers::Value::key(&state_key)
+                    ));
+                }
+            };
+        Ok(apply_state)
+    }
+
+    /// Applies snapshot data of the Region.
+    fn apply_snap(&mut self, region_id: u64, abort: Arc<AtomicUsize>, peer_id: u64) -> Result<()> {
+        info!("begin apply snap data"; "region_id" => region_id);
```
BusyJay, line 383: Please also log the peer id.

components/raftstore/src/store/worker/region.rs
BusyJay: If tiflash has its own metrics, then I think we don't need another metrics.

components/raftstore/src/store/worker/region.rs
CalvinNeo: fixed

CalvinNeo: fixed

components/raftstore/src/store/mod.rs
CalvinNeo: We need some functions inside to manipulate snapshots, including `snap::snap_io::get_decrypter_reader` and `snap::plain_file_used`

components/raftstore/src/store/worker/region.rs
```diff
@@ -692,8 +747,13 @@ where
...
             }
-            if let Some(Task::Apply { region_id, status }) = self.pending_applies.pop_front() {
-                self.ctx.handle_apply(region_id, status);
+            if let Some(Task::Apply {
+                region_id,
+                status,
+                peer_id,
+            }) = self.pending_applies.pop_front()
+            {
+                self.ctx.handle_apply(region_id, status, peer_id);
```
BusyJay, line 756: The order should be `region_id, peer_id, status`.

components/raftstore/src/store/worker/region.rs
```diff
@@ -627,6 +642,46 @@ where
...
+
+        let term = apply_state.get_truncated_state().get_term();
+        let idx = apply_state.get_truncated_state().get_index();
+        let snap_key = SnapKey::new(*region_id, term, idx);
+        let s = box_try!(self.mgr.get_snapshot_for_applying(&snap_key));
+        if !s.exists() {
+            return Err(box_err!("missing snapshot file {}", s.path()));
+        }
+        check_abort(&abort)?;
+        self.coprocessor_host.pre_apply_snapshot(
```
BusyJay, line 670: Why not put it in `apply_snap`? So all these states and snapshot don't need to be fetched again.

components/raftstore/src/store/worker/region.rs
CalvinNeo: OK, and it will also affects `apply_snap`

components/raftstore/src/store/worker/region.rs
CalvinNeo: If we do `pre_apply_snapshot` in `apply_snap`, which means eventually popped from pending task queue by `handle_pending_applies`, that will be too late. 
TiFlash need this because we can do `pre_apply_snapshot` for multiple snapshots in parallel.
If this procedure brings extra costs, I think I can introduce a feature `need_pre_apply_snapshot` in engine_trait as a switch.

BusyJay (approved):
Rest LGTM

components/raftstore/src/store/worker/region.rs
```diff
@@ -763,6 +817,14 @@ where
             }
             task @ Task::Apply { .. } => {
                 fail_point!("on_region_worker_apply", true, |_| {});
+                if self.ctx.coprocessor_host.should_pre_apply_snapshot() {
+                    if let Err(e) = self.ctx.pre_apply_snapshot(&task) {
+                        info!(
```
breezewish, line 822: maybe warn?

components/raftstore/src/store/worker/region.rs
```diff
@@ -763,6 +817,14 @@ where
             }
             task @ Task::Apply { .. } => {
                 fail_point!("on_region_worker_apply", true, |_| {});
+                if self.ctx.coprocessor_host.should_pre_apply_snapshot() {
+                    if let Err(e) = self.ctx.pre_apply_snapshot(&task) {
```
breezewish, line 821: these pre-apply-snapshots are handled in serial instead of concurrently?

breezewish (approved):
the rest LGTM

components/raftstore/src/store/worker/region.rs
CalvinNeo: They are handled in parallel

components/raftstore/src/store/worker/region.rs
CalvinNeo: I think maybe we can just remove this log. Since:
1. If the error is raised by check_abort, which means we just quit applying this snapshot, it is in expectation.
2. If the error is because we have not finished receiving/persisting snapshot for now, it is also in expectation.

For the rear case, we can have a `pre_apply_snapshot(.., None)` to notify, so we can print if we need in proxy code, though currently this is not necessary.

breezewish (approved):

PR #12900, opened 2022-06-24
*: set write and time details in RPC responses 

sticnarf:

Issue Number: Ref https://github.com/tikv/tikv/issues/12362

Depends on https://github.com/pingcap/kvproto/pull/931

What's Changed:

```commit-message
kvproto#931 adds a few more details about the time used
by an RPC request. We are filling these additional information
to the responses in this commit.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/tracker/src/lib.rs
+++ b/components/tracker/src/lib.rs
@@ -34,9 +34,38 @@ impl Tracker {
     pub fn write_scan_detail(&self, detail_v2: &mut pb::ScanDetailV2) {
         detail_v2.set_rocksdb_block_read_byte(self.metrics.block_read_byte);
         detail_v2.set_rocksdb_block_read_count(self.metrics.block_read_count);
+        detail_v2.set_rocksdb_block_read_nanos(self.metrics.block_read_nanos);
         detail_v2.set_rocksdb_block_cache_hit_count(self.metrics.block_cache_hit_count);
         detail_v2.set_rocksdb_key_skipped_count(self.metrics.internal_key_skipped_count);
         detail_v2.set_rocksdb_delete_skipped_count(self.metrics.deleted_key_skipped_count);
+        detail_v2.set_get_snapshot_nanos(self.metrics.get_snapshot_nanos);
+    }
+
+    pub fn write_write_detail(&self, detail: &mut pb::WriteDetail) {
+        detail.set_store_batch_wait_nanos(self.metrics.wf_batch_wait_nanos);
+        detail.set_propose_send_wait_nanos(self.metrics.propose_send_wait_nanos);
+        detail.set_persist_log_nanos(
+            self.metrics.wf_persist_log_nanos - self.metrics.wf_send_to_queue_nanos,
+        );
+        detail.set_raft_db_write_leader_wait_nanos(
+            self.metrics.store_mutex_lock_nanos + self.metrics.store_thread_wait_nanos,
+        );
+        detail.set_raft_db_sync_log_nanos(self.metrics.store_write_wal_nanos);
+        detail.set_raft_db_write_memtable_nanos(self.metrics.store_write_memtable_nanos);
+        // It's an approximation considering generating proposal is fast CPU operation.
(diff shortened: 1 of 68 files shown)


sticnarf:
@you06 @cfzjywxk I think this PR is reviewable but some fields are added in https://github.com/pingcap/kvproto/pull/931. And you can also refer to https://github.com/tikv/client-go/pull/536 and https://github.com/pingcap/tidb/pull/35814 for using them in the TiDB slow log.

src/server/service/kv.rs
```diff
@@ -369,7 +400,7 @@ impl<T: RaftStoreRouter<E::Local> + 'static, E: Engine, L: LockManager, F: KvFor
 
     fn coprocessor(&mut self, ctx: RpcContext<'_>, req: Request, sink: UnarySink<Response>) {
         forward_unary!(self.proxy, coprocessor, ctx, req, sink);
-        let begin_instant = Instant::now_coarse();
+        let begin_instant = Instant::now();
```
cfzjywxk, line 403: Why change this here? It seems the usages are not consistent for example the snapshot duration calculation.

sticnarf: now_coarse can only provides 1~10ms precision (depending on the kernel config), which is sometimes not accurate enough for us.

sticnarf:
The total rpc time is also returned in the response. So, we can know the exact time used in the RPC layer (including the network latency).

image

src/server/service/batch.rs
```diff
@@ -35,7 +35,7 @@ pub struct ReqBatcher {
 
 impl ReqBatcher {
     pub fn new(batch_size: usize) -> ReqBatcher {
-        let begin_instant = Instant::now_coarse();
+        let begin_instant = Instant::now();
```
siddontang, line 38: may using now here affect the performance in the hot path?

sticnarf: `now` takes about 20ns while `now_coarse` takes about 4ns.

As our fastest RPC takes tens of microseconds, such increase is ignorable.

you06:
> Maybe we could add a new kv time graph to show these different layers somewhere together:

Maybe it's good to display such metrics in performance overview dashboard.

components/tracker/src/lib.rs
```diff
@@ -34,9 +34,38 @@ impl Tracker {
     pub fn write_scan_detail(&self, detail_v2: &mut pb::ScanDetailV2) {
         detail_v2.set_rocksdb_block_read_byte(self.metrics.block_read_byte);
         detail_v2.set_rocksdb_block_read_count(self.metrics.block_read_count);
+        detail_v2.set_rocksdb_block_read_nanos(self.metrics.block_read_nanos);
```
you06, line 37: Also need to update `metrics.block_read_nanos` when reporting metrics to the tracker?

components/tracker/src/lib.rs
```diff
@@ -34,9 +34,38 @@ impl Tracker {
...
         detail_v2.set_rocksdb_block_cache_hit_count(self.metrics.block_cache_hit_count);
         detail_v2.set_rocksdb_key_skipped_count(self.metrics.internal_key_skipped_count);
         detail_v2.set_rocksdb_delete_skipped_count(self.metrics.deleted_key_skipped_count);
+        detail_v2.set_get_snapshot_nanos(self.metrics.get_snapshot_nanos);
+    }
+
+    pub fn write_write_detail(&self, detail: &mut pb::WriteDetail) {
+        detail.set_store_batch_wait_nanos(self.metrics.wf_batch_wait_nanos);
```
cfzjywxk, line 45: Do we need to link the performance map in the comments to explain the relationships?

cfzjywxk (approved):

you06 (approved):

components/tracker/src/lib.rs
sticnarf: Let me the change some bits of the performance to match the metrics first.

PR #12910, opened 2022-06-24
online_config: allow return error when update config

### LINKED ISSUE #12909: replace Into with TryInto when convert ConfigValue

glorv:

Issue Number: Close #12909

What's Changed:

```
Replace Into::into with TryInto::try_info to simplify the type conversion of ConfigValue
```

Currently, OnlineConfig depends on `From`/`Into` to convert between `ConfigValue` and the actual config type. For enum types that only accepts some specific string or integer values we have to define a enum variant like `ConfigValue::IOPriority(String)`, `ConfigValue::BlobRunMode(String)`. And the variants count increase linearly with new enum types. 

By changing the type conversion interface from `Into` to `TryInfo`, we can simplify this kind of type into `ConfigValue::String`. 

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

### Release note 

```release-note
None
```

### DIFF
--- a/components/cdc/src/endpoint.rs
+++ b/components/cdc/src/endpoint.rs
@@ -447,7 +447,10 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
     fn on_change_cfg(&mut self, change: ConfigChange) {
         // Validate first.
         let mut validate_cfg = self.config.clone();
-        validate_cfg.update(change);
+        if let Err(e) = validate_cfg.update(change) {
+            warn!("cdc config update failed"; "error" => ?e);
+            return;
+        }
         if let Err(e) = validate_cfg.validate() {
             warn!("cdc config update failed"; "error" => ?e);
             return;
@@ -459,7 +462,7 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
             "change" => ?change
         );
         // Update the config here. The following adjustments will all use the new values.
-        self.config.update(change.clone());
+        self.config.update(change.clone()).unwrap();
 
         // Maybe the cache will be lost due to smaller capacity,
         // but it is acceptable.
--- a/components/engine_rocks/src/config.rs
+++ b/components/engine_rocks/src/config.rs
@@ -1,6 +1,6 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::str::FromStr;
+use std::{convert::TryFrom, str::FromStr};
 
 use online_config::ConfigValue;
 use rocksdb::{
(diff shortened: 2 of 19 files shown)


glorv:
@tabokie  @BusyJay  PTAL

components/tikv_util/src/config.rs
```diff
@@ -1117,6 +1117,15 @@ impl<T> VersionTrack<T> {
         self.version.fetch_add(1, Ordering::Release);
     }
 
+    pub fn try_update<F>(&self, f: F) -> Result<(), Box<dyn std::error::Error>>
```
tabokie, line 1120: Why not change the `update`?

components/resolved_ts/src/endpoint.rs
```diff
@@ -579,7 +579,10 @@ where
     fn handle_change_config(&mut self, change: ConfigChange) {
         let prev = format!("{:?}", self.cfg);
         let prev_advance_ts_interval = self.cfg.advance_ts_interval;
-        self.cfg.update(change);
+        if let Err(e) = self.cfg.update(change).unwrap() {
```
tabokie, line 582: What's with this `unwrap`?

tabokie (approved):

zhangjinpeng1987 (approved):
LGTM

zhangjinpeng1987:
@glorv please check the ci.

components/engine_rocks/src/config.rs
```diff
@@ -225,21 +225,22 @@ pub enum BlobRunMode {
...
-                m => panic!("expect: kNormal, kReadOnly or kFallback, got: {:?}", m),
-            }
+impl TryFrom<ConfigValue> for BlobRunMode {
+    type Error = String;
+    fn try_from(c: ConfigValue) -> Result<BlobRunMode, Self::Error> {
+        if let ConfigValue::String(s) = c {
+            Self::from_str(&s)
         } else {
-            panic!("expect: ConfigValue::BlobRunMode, got: {:?}", c);
+            panic!("expect: ConfigValue::String, got: {:?}", c);
```
BusyJay, line 243: Why not return error instead?

components/tikv_util/src/config.rs
```diff
@@ -1130,6 +1133,24 @@ impl<T> VersionTrack<T> {
     }
 }
 
+/// CfgChanged is used to determine after call `VersionTrack::update`,
+/// whether the config is changed.
+pub trait CfgChanged: 'static {
+    fn changed(&self) -> bool;
+}
+
+impl CfgChanged for () {
```
BusyJay, line 1142: This is very strange. Why not make the function always return `Result`?

components/engine_rocks/src/config.rs
glorv: Normal logic can't enter this branch, so I think it's ok to just panic because this must be a bug then.

BusyJay: Is it possible that TiDB configures it with an int by SQL? Or can this be triggered by invalid client requests?

glorv: No. The input (either toml or json) will first be changed to a ConfigValue with the same type, if the type is not match, the first step will return an error. See: https://github.com/tikv/tikv/blob/a9c3e56552c803642f640e9b3fa8725aa3072400/src/config.rs#L3573-L3590

glorv:
@BusyJay PTAL, thanks

BusyJay (approved):

components/cdc/src/endpoint.rs
```diff
@@ -459,7 +462,7 @@ impl<T: 'static + RaftStoreRouter<E>, E: KvEngine> Endpoint<T, E> {
             "change" => ?change
         );
         // Update the config here. The following adjustments will all use the new values.
-        self.config.update(change.clone());
+        self.config.update(change.clone()).unwrap();
```
overvenus, line 465: Seems like it may panic.

PR #12892, opened 2022-06-23
*: mitigations for Raft Engine encryption key issue

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

### What is changed and how it works?

Issue Number: Close #12890

What's Changed:

- Update Raft Engine and implement new trait methods.
  - `delete` ensures encryption keys of new files can be properly cleaned up.
  - `delete_metadata` helps Raft Engine cleans up stale encryption keys left from older version of TiKV.
- Update tikv-ctl to support scanning and purging encryption keys. If the server is shutdown due to OOM or disk full, this operation can still be executed to free up space.

### Related changes

https://github.com/tikv/raft-engine/pull/229

- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Fix encryption keys not cleaned up when Raft Engine is enabled
```

PR #12915, opened 2022-06-27
raftstore: separate raft and kv perf contexts

sticnarf:

Issue Number: Ref #12362

What's Changed:

```commit-message
We used to record perf contexts for Raft RocksDB and KV RocksDB with
the same PerfContext. But we also have raft-engine now. So, we will miss
perf contexts if we still use RocksDB perf contexts.

This commit adds PerfContext support to RaftEngine and distinguish it
from the perf context used for applying. Then, we'll record correct perf
statistics for both raft engine and KV DB.

Updated raft-engine to include tikv/raft-engine#227 
```

Updated raft-engine to include tikv/raft-engine#227 

### Check List 

Tests 

### Release note 

```release-note
None
```


components/raft_log_engine/src/engine.rs
```diff
@@ -222,6 +224,11 @@ impl FileSystem for ManagedFileSystem {
             })
         }
     }
+
+    fn delete<P: AsRef<Path>>(&self, _path: P) -> IoResult<()> {
+        // TODO(tabokie)
+        Ok(())
```
cfzjywxk, line 230: Do we need to mark it `todo!()` to avoid side effects by now?

components/raft_log_engine/src/perf_context.rs
```diff
@@ -0,0 +1,29 @@
...
+        let perf_context = get_perf_context();
+        for token in trackers {
+            GLOBAL_TRACKERS.with_tracker(*token, |t| {
+                t.metrics.store_thread_wait_nanos =
+                    perf_context.write_wait_duration.as_nanos() as u64;
+                t.metrics.store_write_wal_nanos = (perf_context.log_write_duration
+                    + perf_context.log_sync_duration
+                    + perf_context.log_rotate_duration)
+                    .as_nanos() as u64;
+                t.metrics.store_write_memtable_nanos =
```
cfzjywxk, line 24: This is a bit confusing setting `write_memtable` to `log_rotate`, maybe add some comments or link to check their relations?

components/raftstore/src/store/peer.rs
```diff
@@ -1221,7 +1221,7 @@ where
     pub fn destroy(
         &mut self,
         engines: &Engines<EK, ER>,
-        perf_context: &mut EK::PerfContext,
+        perf_context: &mut ER::PerfContext,
```
you06, line 1224: But no trackers are passed when calling `perf_context.report_metrics`, will this observation take any effect? And I'm not sure if recording the perf context in destroy is required, is it frequently happening?

you06 (commented):
Also need to remove the todo comment.

https://github.com/tikv/tikv/blob/bc84cb1ab370a6c21b18c38885be63f16c519ac6/components/raftstore/src/store/async_io/write.rs#L544

components/raftstore/src/store/peer.rs
sticnarf: `detroy` is not bounded to any user request, so it will not be recorded to the tracker.

components/raft_log_engine/src/perf_context.rs
sticnarf: Sorry, it's a mistake. `apply_duration` should be used here.

components/raft_log_engine/src/engine.rs
sticnarf: I propogate it to the inner `delete`, and should work now. cc @tabokie

sticnarf: After all, this will be replaced with the content in https://github.com/tikv/tikv/pull/12892.. So, just let it work...

cfzjywxk (approved):
LGTM

you06 (approved):

PR #12924, opened 2022-06-28
raftstore: Implement engine trait can_apply_snapshot

CalvinNeo:

Issue Number: ref #12849

What's Changed:

```commit-message
Support new engine trait can_apply_snapshot
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/engine_rocks/src/engine.rs
```diff
@@ -23,8 +23,8 @@ use crate::{
 
 #[derive(Clone, Debug)]
 pub struct RocksEngine {
-    db: Arc<DB>,
-    shared_block_cache: bool,
+    pub db: Arc<DB>,
```
tonyxuqqi, line 26: we already have way to return the DB, no need to make it public.

components/engine_rocks/src/engine.rs
```diff
@@ -23,8 +23,8 @@ use crate::{
...
+    pub shared_block_cache: bool,
```
tonyxuqqi, line 27: same here. prefer to have get method for this.

components/engine_traits/src/engine.rs
```diff
@@ -59,6 +59,11 @@ pub trait KvEngine:
     /// This only exists as a temporary hack during refactoring.
     /// It cannot be used forever.
     fn bad_downcast<T: 'static>(&self) -> &T;
+
+    /// Can apply snapshot now
+    fn can_apply_snapshot(&self) -> bool {
```
tonyxuqqi, line 64: when should can_apply_snapshot return false

components/engine_rocks/src/raw_util.rs
```diff
@@ -14,8 +14,8 @@ use rocksdb::{
 use tikv_util::warn;
 
 pub struct CFOptions<'a> {
-    cf: &'a str,
-    options: ColumnFamilyOptions,
+    pub cf: &'a str,
```
tonyxuqqi, line 17: why make it pub

components/engine_traits/src/engine.rs
zhangjinpeng1987: Agree, better to add comment for when it will return false.

components/raftstore/src/store/worker/region.rs
```diff
@@ -675,6 +675,14 @@ where
         }
     }
 
+    /// Called when received new ApplyTask
+    fn handle_new_pending_applies(&mut self) {
```
zhangjinpeng1987, line 679: Please give a comment to explain when should call 'handle_new_pending_applies' and when should call 'handle_pending_applies'.

components/raftstore/src/store/worker/region.rs
```diff
@@ -1267,5 +1275,15 @@ mod tests {
...
         thread::sleep(Duration::from_millis(PENDING_APPLY_CHECK_INTERVAL * 2));
         assert!(!check_region_exist(6));
+
+        #[cfg(feature = "failpoints")]
+        {
+            engine.kv.compact_files_in_range(None, None, None).unwrap();
+            fail::cfg("handle_new_pending_applies", "return").unwrap();
+            gen_and_apply_snap(7);
+            thread::sleep(Duration::from_millis(PENDING_APPLY_CHECK_INTERVAL * 2));
+            wait_apply_finish(&[7]);
```
zhangjinpeng1987, line 1285: Any assertion to check?

CalvinNeo: If the snapshot is not applied, wait_apply_finish shall fire with Timeout

components/raftstore/src/store/worker/region.rs
CalvinNeo: This function is kind of ambiguous, I am trying to remove that.

components/engine_rocks/src/engine.rs
```diff
@@ -66,6 +66,10 @@ impl RocksEngine {
         self.shared_block_cache = enable;
     }
 
+    pub fn get_shared_block_cache(&self) -> bool {
```
BusyJay, line 69: Should be `shared_block_cache`. Note we will remove the field in the future and cache will be always shared.

components/engine_traits/src/engine.rs
```diff
@@ -64,6 +64,14 @@ pub trait KvEngine:
     /// This only exists as a temporary hack during refactoring.
     /// It cannot be used forever.
     fn bad_downcast<T: 'static>(&self) -> &T;
+
+    /// Returns false if KvEngine can't apply snapshot for this region now.
+    /// Some KvEngines need to do some transforms before apply data from
+    /// snapshot. These procedures can be batched in background if there are
+    /// more than one incoming snapshots, thus not blocking applying thread.
+    fn can_apply_snapshot(&self, _is_timeout: bool, _region_id: u64) -> bool {
```
BusyJay, line 72: Why need `is_timeout`? How can it make any difference?

CalvinNeo: If there is no incoming applying snapshot for a long time, `handle_pending_applies` can be triggered by `on_timeout`, thus proxy can do it directly in foreground.

BusyJay: Then why bother calling `can_apply_snapshot` again?

CalvinNeo: `handle_pending_applies` will apply multiple snapshots as long as it will not cause write stall. However, in proxy we won't handle the more than one snapshot if the following snapshot is still pre-handling.

components/raftstore/src/store/config.rs
```diff
@@ -31,6 +31,21 @@ lazy_static! {
...
 }
 
+// used to periodically check whether we should delete a stale peer's range in
+// region runner
+
+#[cfg(test)]
+pub const STALE_PEER_CHECK_TICK: usize = 1; // 1000 milliseconds
+
+#[cfg(not(test))]
+pub const STALE_PEER_CHECK_TICK: usize = 10; // 10000 milliseconds
```
BusyJay, line 41: You should set it explicitly in test.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1557,6 +1560,8 @@ pub mod tests {
             s.engines.kv.clone(),
             mgr,
             0,
+            PENDING_APPLY_CHECK_INTERVAL,
```
BusyJay, line 1563: Why not pass a `cfg` instead?

components/raftstore/src/store/worker/region.rs
```diff
@@ -427,8 +412,15 @@ where
             coprocessor_host: self.coprocessor_host.clone(),
         };
         s.apply(options)?;
-        self.coprocessor_host
-            .post_apply_snapshot(&region, peer_id, &snap_key, Some(&s));
+        match self
```
BusyJay, line 415: Why not `if let Err(e) = `?

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -21,7 +21,7 @@ use raft::{eraftpb, StateRole};
 pub mod config;
 mod consistency_check;
 pub mod dispatcher;
-mod error;
+pub mod error;
```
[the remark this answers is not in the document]
CalvinNeo: We need to use coprocessor's `error::Error`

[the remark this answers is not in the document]
BusyJay: Then why not just reexport the type?

components/raftstore/src/store/worker/region.rs
CalvinNeo: solved

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -539,15 +539,19 @@ impl<E: KvEngine> CoprocessorHost<E> {
...
-            peer_id,
-            snap_key,
-            snap,
-        );
+    ) -> Result<()> {
+        let mut ctx = ObserverContext::new(region);
+        for observer in &self.registry.apply_snapshot_observers {
+            let observer = observer.observer.inner();
+            let res = observer.post_apply_snapshot(&mut ctx, peer_id, snap_key, snap);
+            #[allow(clippy::all)]
```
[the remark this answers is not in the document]
CalvinNeo: fixed

components/engine_rocks/src/lib.rs
```diff
@@ -32,7 +32,7 @@ mod compact;
 pub use crate::compact::*;
 mod db_options;
 pub use crate::db_options::*;
-mod db_vector;
+pub mod db_vector;
```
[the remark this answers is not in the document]
CalvinNeo: fixed

components/raftstore/src/coprocessor/mod.rs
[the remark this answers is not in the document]
CalvinNeo: fixed

components/raftstore/src/store/config.rs
```diff
@@ -137,6 +140,14 @@ pub struct Config {
     #[online_config(skip)]
     pub snap_apply_batch_size: ReadableSize,
 
+    #[doc(hidden)]
+    #[online_config(skip)]
+    pub region_worker_tick_interval: ReadableDuration,
+
+    #[doc(hidden)]
+    #[online_config(skip)]
+    pub clean_stale_tick_max: usize,
```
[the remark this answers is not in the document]
CalvinNeo: changed to stale_peer_check_tick

components/raftstore/src/store/peer_storage.rs
CalvinNeo: fixed

components/raftstore/src/store/config.rs
```diff
@@ -137,6 +140,14 @@ pub struct Config {
     #[online_config(skip)]
     pub snap_apply_batch_size: ReadableSize,
 
+    #[doc(hidden)]
+    #[online_config(skip)]
+    pub region_worker_tick_interval: ReadableDuration,
+
+    #[doc(hidden)]
+    #[online_config(skip)]
+    pub stale_peer_check_tick: usize,
```
BusyJay, line 149: I think it's ambiguous. There is also `peer_stale_state_check_interval` in the same struct. Hard to tell the difference.

components/raftstore/src/store/snap.rs
```diff
@@ -1200,6 +1200,10 @@ impl Snapshot {
         self.hold_tmp_files = false;
         Ok(())
     }
+
+    pub fn cf_files(&self) -> &[CfFile] {
+        self.cf_files.as_slice()
```
BusyJay, line 1205: You mean `&self.cf_files`?

components/raftstore/src/store/worker/region.rs
```diff
@@ -427,8 +414,12 @@ where
             coprocessor_host: self.coprocessor_host.clone(),
         };
         s.apply(options)?;
-        self.coprocessor_host
-            .post_apply_snapshot(&region, peer_id, &snap_key, Some(&s));
+        if let Err(e) =
+            self.coprocessor_host
+                .post_apply_snapshot(&region, peer_id, &snap_key, Some(&s))
+        {
+            return Err(box_err!("post apply snapshot error {:?}", e));
```
BusyJay, line 421: Returning error can make the applying snapshot in undetermined state. I'm afraid it can be misused.

components/tikv_util/src/timer.rs
```diff
@@ -114,6 +115,10 @@ fn start_global_timer() -> Handle {
     rx.recv().unwrap()
 }
 
+fn start_global_timer() -> Handle {
```
BusyJay, line 118: Instead of creating a new function, I suggest to change the place that use `start_global_timer`.

components/raftstore/src/store/config.rs
CalvinNeo: The previous author use constant `STALE_PEER_CHECK_TICK`, so I use `stale_peer_check_tick `

components/raftstore/src/store/snap.rs
CalvinNeo: fixed

components/raftstore/src/store/config.rs
BusyJay: That's because the constant is defined in region worker, it has context.

components/tikv_util/src/timer.rs
CalvinNeo: fixed

components/raftstore/src/store/worker/region.rs
CalvinNeo: we log an error here instead

components/engine_rocks/src/encryption.rs
```diff
@@ -12,7 +12,7 @@ use rocksdb::{
 use crate::{r2e, raw::Env};
 
 // Use engine::Env directly since Env is not abstracted.
-pub(crate) fn get_env(
```
CalvinNeo, line 15: Exposed to parse SST files.

BusyJay: You should use `engine_rocks::get_env` instead.

CalvinNeo: fixed

components/raftstore/src/store/worker/region.rs
```diff
@@ -745,22 +732,35 @@ where
     }
 
     /// Tries to apply pending tasks if there is some.
-    fn handle_pending_applies(&mut self) {
+    fn handle_pending_applies(&mut self, is_timeout: bool) {
         fail_point!("apply_pending_snapshot", |_| {});
+        let mut is_first = true;
```
BusyJay, line 737: Perhaps should be named `new_batch`.

BusyJay (approved):
Rest LGTM

components/raftstore/src/store/worker/region.rs
CalvinNeo: fixed

components/raftstore/src/store/config.rs
```diff
@@ -137,6 +137,17 @@ pub struct Config {
     #[online_config(skip)]
     pub snap_apply_batch_size: ReadableSize,
 
+    // used to periodically check whether schedule pending applies in region runner
+    #[doc(hidden)]
+    #[online_config(skip)]
+    pub region_worker_tick_interval: ReadableDuration,
```
breezewish, line 143: Seems that the name is changed from "pending_apply_check_xxx". The previous name is more obervious to me IMO, could you explain your idea behind this name change?

breezewish:
The rest looks good to me.  @tonyxuqqi PTAL, thanks!

components/raftstore/src/store/config.rs
CalvinNeo: We already have this configuration in proxy long ago, so this is more like a backport to TiKV

breezewish: @busyjay seems that this name has already been used in ser/deser config items. Do you have some suggestions?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1144,9 +1147,24 @@ pub mod tests {
...
                 FetchedLogs, LogFetchedNotifier, RaftlogFetchRunner, RegionRunner, RegionTask,
             },
+            Config,
         },
     };
 
+    const PENDING_APPLY_CHECK_INTERVAL: u64 = 200;
+    const STALE_PEER_CHECK_TICK: usize = 1;
+
+    fn make_raftstore_cfg(use_delete_range: bool) -> Arc<VersionTrack<Config>> {
```
tonyxuqqi, line 1157: all these configuration except use_delete_range parameter are hardcoded. So what's the point of creating a config instead of hardcode in the RegionRunner ?

tonyxuqqi: Also please add a comment to clarify what's the consumer of this Config. It may be easily confused as the config used by raftstore itself.

tonyxuqqi (approved):

PR #12928, opened 2022-06-29
server: support lossy metrics compaction

glorv:

Issue Number: ref #12355

What's Changed:

```
add a new config to support lossy metrics compaction(reduce histogram size by filter out variants with less than 1% samples), this can further reduce the metrics size by about 10%.
```

This PR is split from #12732 by keeping the `metrics-policy` part:

Add a new config `server.metrics-compact-policy` with three options:
- "no"(default). do nothing, return all the raw metrics data.
- "lossless". As implemented in https://github.com/tikv/tikv/pull/12602, do not return counters with 0 value and histograms will 0 samples as they are the same as the default value, so we reduce the data with no information loss.
- "lossy". Try to reduce the metrics as much as possible. In addition to level LoseLess, this PR will also reduce histogram vector type by filtering all histograms whose sample count < max sample count * 0.01. We may also do further operations(in the future), e.g. we can apply the ideas in https://github.com/tikv/tikv/pull/12417 to compress some data types when we come up with some consensus.

Test Result:
| Metrics Compaction Policy | no | lossless | lossy |
| -- | -- | -- | -- |
| | 1407K | 521K(37.0%) | 400K(28.4%) |
 
### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/components/tikv_util/src/metrics/mod.rs
+++ b/components/tikv_util/src/metrics/mod.rs
@@ -1,11 +1,18 @@
 // Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::{collections::HashMap, io::Write};
+use std::{
+    collections::HashMap,
+    convert::TryFrom,
+    io::Write,
+    sync::atomic::{AtomicU8, Ordering},
+};
 
 use kvproto::pdpb;
 use lazy_static::lazy_static;
+use online_config::ConfigValue;
 use prometheus::{proto::MetricType, *};
 use prometheus_static_metric::*;
+use serde::{Deserialize, Serialize};
 
 #[cfg(target_os = "linux")]
 mod threads_linux;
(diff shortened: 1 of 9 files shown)


glorv:
@BusyJay @breezewish @zhangjinpeng1987  @kevin-xianliu  PTAL, thank you~

components/tikv_util/src/metrics/mod.rs
```diff
@@ -59,7 +77,20 @@ pub fn dump_to(w: &mut impl Write, should_simplify: bool) {
             MetricType::COUNTER => {
                 metrics.retain(|m| m.get_counter().get_value() > 0.0);
             }
-            MetricType::HISTOGRAM => metrics.retain(|m| m.get_histogram().get_sample_count() > 0),
+            MetricType::HISTOGRAM => {
+                let threshold = if compact_policy == MetricsCompactPolicy::Lossless {
+                    0
+                } else {
+                    // only retain histogram that the sample count > 0.01 * max_sample_count
```
BusyJay, line 84: This can be a big problem for analyzing tail latency.

glorv: Yes. So this level should be only used when metrics data size is significant.

BusyJay: I suggest to keep the highest data point at least.

glorv:
Closed as we don't reach an agreement on whether it is proper to reduce that much metrics.

Issue #12935, opened 2022-06-29
Add check conditions before compaction filter in  RawKV GC 


ystaticy:
## Development Task
Add some conditions before new RawKV GC compaction filter :
1. If rocksdb is write stall or stopped, it will not execute rawkv compaction filter.
2. If enable_compaction_filter=false or (need version check && version >= 5.0.0), it will not execute rawkv compaction filter.
3. If check_need_gc return false, it will not execute rawkv compaction filter.

<img width="946" alt="image" src="https://user-images.githubusercontent.com/3659421/176421875-86e1850f-634a-427d-8247-523af18465ec.png">

The pseudo code of ```check_props``` is :
```
    let check_props = |props: &MvccProperties| -> (bool, bool /*skip_more_checks*/) {
        if props.min_ts > safe_point {
            return (false, false);
        }
        if ratio_threshold < 1.0 || context.is_bottommost_level() {
            // According to our tests, `split_ts` on keys and `parse_write` on values
            // won't utilize much CPU. So always perform GC at the bottommost level
            // to avoid garbage accumulation.
            return (true, true);
        }

        if props.num_versions as f64 > props.num_rows as f64 * ratio_threshold {
            // It means versions is too much
            return (true, false);
        }

        if props.num_deletes as f64 > props.num_puts as f64 * ratio_threshold {
            // It means deletes version is much more than puts
            return (true, false);
        }

        if props.min_ttl_ts > safe_point {
            // It means all records ttl is large than safe_point
            return (false, false);
        }

        (props.max_row_versions > 1024, false)
    };


```

PR #12938, opened 2022-06-30
Add check conditions before compaction filter in RawKV GC

ystaticy:

Issue Number: Close https://github.com/tikv/tikv/issues/12935

What's Changed:

```
Add check conditions before compaction filter in RawKV GC
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Add check conditions before compaction filter in RawKV GC
```


components/engine_rocks/src/properties.rs
```diff
@@ -493,6 +494,120 @@ impl TablePropertiesCollector for MvccPropertiesCollector {
...
+        if self.row_versions > self.props.max_row_versions {
+            self.props.max_row_versions = self.row_versions;
+        }
+
+        let decode_raw_value = ApiV2::decode_raw_value(value);
+        if let Ok(raw_value) = decode_raw_value {
+            if raw_value.is_valid(current_ts) {
+                self.props.num_puts += 1;
+                if !raw_value.is_delete {
+                    let expire_ts = raw_value.expire_ts.unwrap();
```
pingyu, line 575: Suggest not to assume that a not deleted entry must have `expire_ts`. We may add more fields in the future.
Also, `expire_ts == 0` means never expired, should be ignored.

src/server/gc_worker/rawkv_compaction_filter.rs
```diff
@@ -136,13 +180,71 @@ impl CompactionFilter for RawCompactionFilter {
...
+fn do_check_allowed(enable: bool, skip_vcheck: bool, feature_gate: &FeatureGate) -> bool {
+    enable && (skip_vcheck || feature_gate.can_enable(COMPACTION_FILTER_RAWKV_GC_FEATURE))
+}
+
+fn check_need_gc(
+    safe_point: TimeStamp,
+    ratio_threshold: f64,
+    context: &CompactionFilterContext,
+) -> bool {
+    println!("======check_need_gc01======");
```
pingyu, line 192: Remove debug codes.

src/server/gc_worker/rawkv_compaction_filter.rs
```diff
@@ -136,13 +180,71 @@ impl CompactionFilter for RawCompactionFilter {
     }
 }
 
+fn do_check_allowed(enable: bool, skip_vcheck: bool, feature_gate: &FeatureGate) -> bool {
+    enable && (skip_vcheck || feature_gate.can_enable(COMPACTION_FILTER_RAWKV_GC_FEATURE))
+}
+
+fn check_need_gc(
```
pingyu, line 187: Suggest to reuse `check_need_gc` of TiDB's compaction filter, which can adapt RawKV's MvccProperties as `props.min_ttl_ts` is always 0.

ystaticy: Now, It reuse check_need_gc.

And ```props.min_ttl_ts``` default value is ```TimeStamp::max()```.
So,if props.min_ttl_ts != TimeStamp::max() && props.min_ttl_ts > safe_point, then it will return (false,false)
If ```props.min_ttl_ts = TimeStamp::max()``` mean there are no writes with ttl, the return value depends on other conditions.

components/engine_rocks/src/properties.rs
ystaticy: I modified the logic, If expire_ts is None, it will not affect min_ttl_ts

components/engine_rocks/src/properties.rs
```diff
@@ -394,6 +395,7 @@ pub struct MvccPropertiesCollector {
     row_versions: u64,
     cur_index_handle: IndexHandle,
     row_index_handles: IndexHandles,
+    key_mode: KeyMode,
```
pingyu, line 398: How about use `GcKeyMode` ? Then there are only two enums and will not be confused with the difference of TiDB & Txn.

components/engine_rocks/src/properties.rs
```diff
@@ -423,13 +438,32 @@ impl TablePropertiesCollector for MvccPropertiesCollector {
             return;
         }
 
-        let (k, ts) = match Key::split_on_ts_for(key) {
-            Ok((k, ts)) => (k, ts),
-            Err(_) => {
-                self.num_errors += 1;
+        let current_ts = ttl_current_ts();
```
pingyu, line 441: Can `current_ts` be cached in `self` ?

src/server/gc_worker/compaction_filter.rs
```diff
@@ -699,13 +701,22 @@ fn check_need_gc(
...
+        if key_mode == KeyMode::Txn || key_mode == KeyMode::Tidb {
+            // When comparing `num_versions` with `num_puts`, trait internal levels
+            // specially because MVCC-deletion marks can't be handled at those
+            // levels.
+            let num_rollback_and_locks = (props.num_versions - props.num_deletes) as f64;
+            if num_rollback_and_locks > props.num_puts as f64 * ratio_threshold {
+                return (true, false);
+            }
+        } else if props.num_deletes as f64 > props.num_puts as f64 * ratio_threshold {
             return (true, false);
```
pingyu, line 713: We handle GC of deletion at bottom most level only. So we should ignore `num_deletes` here ?

And according to the codes of Txn above, `num_deletes` is also considered to be no need to GC in Txn.

components/engine_rocks/src/properties.rs
```diff
@@ -452,18 +486,29 @@ impl TablePropertiesCollector for MvccPropertiesCollector {
...
         }
 
-        let write_type = match Write::parse_type(value) {
-            Ok(v) => v,
-            Err(_) => {
-                self.num_errors += 1;
-                return;
+        if self.key_mode == KeyMode::Raw {
+            let decode_raw_value = ApiV2::decode_raw_value(value);
+            if let Ok(raw_value) = decode_raw_value {
```
pingyu, line 491: Suggest to add `num_errors` and return on error.

src/server/gc_worker/rawkv_compaction_filter.rs
```diff
@@ -24,15 +25,22 @@ use txn_types::Key;
...
+            GC_COMPACTION_FAILURE, GC_COMPACTION_FILTERED, GC_COMPACTION_FILTER_MVCC_DELETION_MET,
+            GC_COMPACTION_FILTER_ORPHAN_VERSIONS, GC_COMPACTION_FILTER_PERFORM,
+            GC_COMPACTION_FILTER_SKIP, GC_CONTEXT,
         },
         GcTask, STAT_RAW_KEYMODE,
     },
     storage::mvcc::{GC_DELETE_VERSIONS_HISTOGRAM, MVCC_VERSIONS_HISTOGRAM},
 };
 
+// The default version that can enable compaction filter for GC. This is
```
pingyu, line 38: It's confused. And we add the RawKV compaction filter since v6.1.0, so why we need this feature gate ?

components/engine_rocks/src/properties.rs
ystaticy: If use ```GcKeyMode```, it  will cause the problem of circular dependency. So I use ```KeyMode ``` here.
And add a comments to indicate that ```KeyMode::Txn``` is used in both TiDB and Txn scenarios

tests/failpoints/cases/test_table_properties.rs
```diff
@@ -0,0 +1,233 @@
...
+        .build_with_cfg(&cfg)
+        .unwrap();
+    let raw_engine = engine.get_rocksdb();
+    let mut gc_runner = TestGcRunner::new(0);
+
+    do_write(&engine, false, 5);
+
+    // Check init value
+    assert_eq!(
+        GC_COMPACTION_FILTER_PERFORM
```
pingyu, line 48: Need reset the metrics at the beginning ?

tests/failpoints/cases/test_table_properties.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    do_write(&engine, false, 5);
+    engine.get_rocksdb().flush_cfs(true).unwrap();
+
+    // Set ratio_threshold, let (props.num_versions as f64 > props.num_rows as
+    // f64 * ratio_threshold) return false
+    gc_runner.ratio_threshold = Option::Some(f64::MAX);
+
+    // is_bottommost_level = false
+    do_gc(&raw_engine, 1, &mut gc_runner, &dir);
```
pingyu, line 217: Either the two conditions (`props.num_versions > props.num_rows * ratio_threshold`, `! is_bottommost_level`) will skip the compaction. Suggest to test them alone.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -687,6 +687,10 @@ fn check_need_gc(
         if props.min_ts > safe_point {
             return (false, false);
         }
+        println!(
```
pingyu, line 690: Suggest to remove the debug codes.

src/config.rs
```diff
@@ -679,6 +680,10 @@ impl DefaultCfConfig {
             prop_size_index_distance: self.prop_size_index_distance,
             prop_keys_index_distance: self.prop_keys_index_distance,
         };
+        cf_opts.add_table_properties_collector_factory(
+            "tikv.mvcc-properties-collector",
```
pingyu, line 684: Should we use another name being different with `MvccPropertiesCollectorFactory` ?

tests/failpoints/cases/test_table_properties.rs
```diff
@@ -0,0 +1,239 @@
...
+        3
+    );
+    assert_eq!(
+        GC_COMPACTION_FILTER_SKIP
+            .with_label_values(&[STAT_RAW_KEYMODE])
+            .get(),
+        0
+    );
+
+    GC_COMPACTION_FILTER_PERFORM.reset();
```
pingyu, line 109: Why not reset at the beginning ? Otherwise it would be interfered by others.

tests/failpoints/cases/test_table_properties.rs
ystaticy: I want to check_need_gc return false, so the following conditions need to be met  here:
1.  ```ratio_threshold < 1.0 || context.is_bottommost_level()``` return false.
2.  ```props.num_versions as f64 > props.num_rows as f64 * ratio_threshold``` return false.
3.  ```num_rollback_and_locks > props.num_puts as f64 * ratio_threshold``` return false.

so I make a compation is_bottommost_level=false ,and set ratio_threshold=```f64::MAX```
then the above three conditions can be met at the same time.

tests/failpoints/cases/test_table_properties.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    // The check_need_gc return false, GC_COMPACTION_FILTER_SKIP will add 1.
+    assert_eq!(
+        GC_COMPACTION_FILTER_SKIP
+            .with_label_values(&[STAT_RAW_KEYMODE])
+            .get(),
+        2
+    );
+
+    GC_COMPACTION_FILTER_PERFORM.reset();
```
pingyu, line 243: Nit: these two are not necessary.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -678,7 +678,7 @@ fn do_check_allowed(enable: bool, skip_vcheck: bool, feature_gate: &FeatureGate)
     enable && (skip_vcheck || feature_gate.can_enable(COMPACTION_FILTER_GC_FEATURE))
 }
 
-fn check_need_gc(
+pub fn check_need_gc(
```
cfzjywxk, line 681: There're two `check_need_gc` methods in the `gc_worker` mod, as this `check_need_gc` would be used by both the txn and the raw KV, it's better to distinguish their names and usages in comments. Maybe the refactor could be done in later PRs.

cfzjywxk (approved):

pingyu:
Hold until v6.3.0 code freeze.

guoxiangCN:
@ystaticy  Hi~  Why did this PR add is_ Bottomost_level  limitations, can you explain your original intention to me ？
the change:
```
if !raw_value.is_valid(self.current_ts) {
  self.raw_handle_delete();
}
```
to
```
if !raw_value.is_valid(self.current_ts) && self.is_bottommost_level {
  self.raw_handle_delete();
}
```

PR #12940, opened 2022-06-30
add TestTabletFactory and refine TabletFactory trait

### LINKED ISSUE #12939: engine_traits: refine the TabletFactory trait to remove the dependency on raftstore v1's RaftRouter
## Development Task
Today the TabletFactory's implementation directly depends on raftstore v1's RaftRouter to send notification back for compaction events. This will bring problem when raftstore v2's RaftRouter counterpart is another structure.
From code layer perspective, it does not make sense to let engine's factory directly depend on the upper layer. Instead engine should provide a abstraction and upper layer can implement it, so that no hard dependency between raftstore v1 and tablet factory.

Also we need a test tablet factory for multi-rocks test.

tonyxuqqi:

Issue Number: Close #12939

What's Changed:

```commit-message
Add TestTabletFactory for testing raftstorev2 with real multi-rocksdb. 
Before this change there's no way to create a tablet factory in raftstore's test code. 
And also this PR refines TabletFactory trait so that its implementation code will not hard wire with raftstore v1's code. 
This will make Tabletfactory's implementation work both with raftstore v1 and raftstore v2. 
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/Cargo.toml
+++ b/components/raftstore/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 publish = false
 
 [features]
-default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
+default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "engine_rocks"]
 failpoints = ["fail/failpoints"]
 testexport = []
 test-engine-kv-rocksdb = [
@@ -38,7 +38,7 @@ crc32fast = "1.2"
 crossbeam = "0.8"
 derivative = "2"
 encryption = { path = "../encryption", default-features = false }
-engine_rocks = { path = "../engine_rocks", default-features = false } 
+engine_rocks = { path = "../engine_rocks", default-features = false, optional = true } 
 
 # Should be [dev-dependencies] but we need to control the features
 # https://github.com/rust-lang/cargo/issues/6915
--- a/components/raftstore/Cargo.toml
+++ b/components/raftstore/Cargo.toml
@@ -38,6 +38,7 @@ crc32fast = "1.2"
 crossbeam = "0.8"
 derivative = "2"
 encryption = { path = "../encryption", default-features = false }
+engine_rocks = { path = "../engine_rocks", default-features = false } 
 
 # Should be [dev-dependencies] but we need to control the features
 # https://github.com/rust-lang/cargo/issues/6915
--- a/components/server/src/lib.rs
+++ b/components/server/src/lib.rs
@@ -7,5 +7,6 @@ extern crate tikv_util;
 pub mod setup;
 pub mod memory;
 pub mod raft_engine_switch;
+pub mod raftstore_v1;
 pub mod server;
 pub mod signal_handler;
--- a/components/server/src/lib.rs
+++ b/components/server/src/lib.rs
@@ -7,6 +7,5 @@ extern crate tikv_util;
 pub mod setup;
 pub mod memory;
 pub mod raft_engine_switch;
-pub mod raftstore_v1;
 pub mod server;
 pub mod signal_handler;
--- a/src/server/engine_factory.rs
+++ b/src/server/engine_factory.rs
@@ -1,23 +1,21 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
 use std::{
+    ops::Deref,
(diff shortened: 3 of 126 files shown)


src/server/engine_factory.rs
```diff
@@ -34,12 +32,12 @@ struct FactoryInner {
     root_db: Mutex<Option<RocksEngine>>,
 }
 
-pub struct KvEngineFactoryBuilder<ER: RaftEngine> {
+pub struct KvEngineFactoryBuilder {
     inner: FactoryInner,
-    router: Option<RaftRouter<RocksEngine, ER>>,
+    compact_event_sender: Option<Arc<dyn CompactedEventSender + Send + Sync>>,
```
BusyJay, line 37: Why `Arc`?

components/server/src/lib.rs
```diff
@@ -7,5 +7,6 @@ extern crate tikv_util;
 pub mod setup;
 pub mod memory;
 pub mod raft_engine_switch;
+pub mod raftstore_v1;
```
BusyJay, line 10: It's very strange to have a raftstore module in server.

src/server/engine_factory.rs
tonyxuqqi: Well because this CompactedEventSender is supposed to be thread safe, so no need to use Box instead? 
It is much easier to handle clone() for Arc than for Box.

components/server/src/lib.rs
tonyxuqqi: it's meant to put raftstore_v1 related implementation.  For now there's only one struct, the submodule seems an overkill. 
I don't like to see too many raftstore v1 and raftstore v2 implementation.  But with today v2's interface change,  perhaps we have to do this.

src/server/engine_factory.rs
BusyJay: Then why there is a `clone` method in `CompactedEventSender`?

components/server/src/lib.rs
BusyJay: Perhaps we should move these implementation to raftstore component instead.

src/server/engine_factory.rs
tonyxuqqi: So in builder, we can use Arc's clone as it's more lightweight. But for TabletFactory, we use CompactedEventSender's clone so that each tablet can have its own Mutex for better performance.

components/server/src/lib.rs
tonyxuqqi: good point

tonyxuqqi: It has a problem in raftstore----raftstore is supposed to be KvEngine agnostic, but here we're dealing with specific RockDBEngine.

components/raftstore/Cargo.toml
```diff
@@ -38,6 +38,7 @@ crc32fast = "1.2"
 crossbeam = "0.8"
 derivative = "2"
 encryption = { path = "../encryption", default-features = false }
+engine_rocks = { path = "../engine_rocks", default-features = false }
```
BusyJay, line 41: Should be optional and enabled only when engine rocks is chosen.

src/server/engine_factory.rs
```diff
@@ -77,40 +74,31 @@ impl<ER: RaftEngine> KvEngineFactoryBuilder<ER> {
...
-
-impl<ER: RaftEngine> KvEngineFactory<ER> {
+impl KvEngineFactory {
     pub fn create_raftstore_compaction_listener(&self) -> Option<CompactionListener> {
-        let router = self.router.lock().unwrap();
-        let ch = match &*router {
-            Some(r) => Mutex::new(r.clone()),
-            None => return None,
-        };
+        self.compact_event_sender.as_ref()?;
```
BusyJay, line 101: What's the point?

tonyxuqqi: if compact_event_sender is None, then return immediately.

components/raftstore/Cargo.toml
tonyxuqqi: how? I don't see any current reference of engine_rocks is optional.

BusyJay (approved):

5kbpers (approved):

5kbpers (approved):

PR #12593, opened 2022-05-20
*: improve Load Base Split by introducing CPU statistical info

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

### What is changed and how it works?

Issue Number: Part of https://github.com/tikv/tikv/issues/12063.

What's Changed:

```commit-message
Improve Load Base Split by introducing CPU statistical info.
```

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 

```release-note
None.
```

PR #12944, opened 2022-06-30
raftstore: support to split the region on half with the given key range

### LINKED ISSUE #12943: Support to split the region on half with the given key range
## Development Task

For now, `check_split_and_bucket` only supports splitting a region by its start key and end key. It is necessary to support splitting the region on half with the given key range if we want to provide features like CPU-based Load Base Split to split a region by its workload flexibly.

JmPotato:

Issue Number: Close #12943. This is a sub-PR of https://github.com/tikv/tikv/pull/12593.

What's Changed:

```commit-message
Support to split the region on half with the given key range.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/keys/src/lib.rs
+++ b/components/keys/src/lib.rs
@@ -241,11 +241,11 @@ pub fn enc_end_key(region: &Region) -> Vec<u8> {
 }
 
 #[inline]
-pub fn data_end_key(region_end_key: &[u8]) -> Vec<u8> {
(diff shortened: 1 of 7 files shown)


lhy1024 (commented):
Do we need to add any tests?

lhy1024 (approved):

5kbpers (approved):

PR #12948, opened 2022-06-30
raftstore: Implement observer on_compute_engine_size

CalvinNeo:

Issue Number: ref #12849

What's Changed:

```commit-message
Implement observer on_compute_engine_size
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/coprocessor/mod.rs
```diff
@@ -151,6 +151,18 @@ pub trait SplitCheckObserver<E>: Coprocessor {
     );
 }
 
+#[derive(Debug, Default)]
+pub struct EngineSize {
```
tonyxuqqi, line 155: Is it better to rename EngineSize as StoreSizeInfo?

tonyxuqqi (approved):

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -203,6 +203,18 @@ pub trait SplitCheckObserver<E>: Coprocessor {
     );
 }
 
+#[derive(Debug, Default)]
+pub struct StoreSizeInfo {
+    pub capacity: u64,
```
BusyJay, line 208: Please add some comments.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -2304,6 +2284,49 @@ fn collect_report_read_peer_stats(
     stats
 }
 
+fn collect_engine_size<EK: KvEngine, ER: RaftEngine>(
+    coprocessor_host: &CoprocessorHost<EK>,
+    store_info: Option<&StoreInfo<EK, ER>>,
+    snap_mgr_size: u64,
+) -> Option<(u64, u64, u64)> {
+    if let Some(engine_size) = coprocessor_host.on_compute_engine_size() {
+        Some((engine_size.capacity, engine_size.used, engine_size.avail))
```
BusyJay, line 2293: How about returning? So there will be no indentation in other branch.

components/raftstore/src/coprocessor/mod.rs
BusyJay: And what's the meaning of all these fields? Why there is a `used` when the other two are available?

BusyJay (approved):

PR #12954, opened 2022-07-04
server: collect count and duration by request source

sticnarf:

Issue Number: Ref #12362

What's Changed:

```commit-message
TiKV client can pass request_source through Context. It is
useful for us to know how many requests there are from each
source.

So, this commit collects the count and the total duration by
request source. The source label is not added to the command
type in order to avoid creating too many label combinations.
```
The grafana panel looks like

grpc

### Check List 

Tests 

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2623,7 +2623,7 @@ dependencies = [
 [[package]]
 name = "kvproto"
 version = "0.0.2"
-source = "git+https://github.com/pingcap/kvproto.git#0e2f26c0a46ae7d666d6ca4410046a39e0c96f36"
+source = "git+https://github.com/pingcap/kvproto.git#acfe326c7cb2bdcdbfc991cada1973a68f34836f"
 dependencies = [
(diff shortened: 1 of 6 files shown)


you06 (approved):

cfzjywxk (approved):

cfzjywxk:
Another part I think is about the raftstore/engine/io usage sources, some of them could be introduced by scheduling and scaling. For example, heartbeat processing would use much raftstore resources and brings much overhead for the user performance but it could not be observed directly.

For the io usage, there's a panel displaying the io flow source, maybe it could be integrated with the request source in the new tracking information?

sticnarf:
> For example, heartbeat processing would use many raftstore resources and brings much overhead for the user performance but it could not be observed directly.

It's hard to quantify and compare the influence of different kinds of raft messages. Flamegraphs may be a better alternative.

Foreground IO flow statistics by source may be possible, but the overall IO is hard due to compaction.

PR #12957, opened 2022-07-05
raftstore: Implement coprocessor observer pre_persist

CalvinNeo:

Issue Number: ref #12849

What's Changed:

Before commit and finish_for, if the observer decides it is not a proper time to do actual persist, it can skip this persistence.
This observer is aim to disable persistence of advanced `applied_index`, since some KvEngine may not persist replicated data when `commit` and `finish_for`, so we can persist advanced `applied_index` either. 
This can be easily done if we just disable `write_apply_state`. However, TiKV assert that data and meta should be persist as a whole, so if we don't want to persist advanced `applied_index`, we should not persist data(which is actually empty, except some pending deleting ssts) either.
We don't worry too much about rubbish data. Since commit can also be triggered by `post_exec`, or at least `ApplyContext::end`. We don't call `write_apply_sate` in `ApplyContext::end`.

```commit-message
Support coprocessor observer pre_commit
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -635,6 +635,26 @@ impl<E: KvEngine> CoprocessorHost<E> {
         );
     }
 
+    /// `pre_commit` is called we we want to persist data or meta for a region.
+    /// For example, in `finish_for` and `commit`,
+    /// we will separately call `pre_commit` with is_finished = true/false.
+    /// By returning false, we reject this persistence.
+    pub fn pre_commit(
```
BusyJay, line 642: Better give it a better name without leaking the implementation details. It will be changed in v2.

components/test_raftstore/src/node.rs
```diff
@@ -269,6 +269,7 @@ impl Simulator for NodeCluster {
                 .max_total_size(cfg.server.snap_max_total_size.0)
                 .encryption_key_manager(key_manager)
                 .max_per_file_size(cfg.raft_store.max_snapshot_file_raw_size.0)
+                .enable_multi_snapshot_files(true)
```
BusyJay, line 272: Why the change?

CalvinNeo: Because I find test cases like `test_server_huge_snapshot_multi_files` may not check expected behavior if `enable_multi_snapshot_files` is set to false. 
These tests are to test if snapshot works fine when we enable multi file snapshot. However, we may not generate multi file snapshot if this switch is closed, so these tests will actually test single file snapshot.

tonyxuqqi (approved):

BusyJay (approved):

PR #12962, opened 2022-07-05
tikv: refactor StorageConfigManger to support multi-rocksdb

### LINKED ISSUE #12961: StorageConfigManger: support multi-rocksdb
## Development Task
Today the StorageConfigManger only works with single rocksdb instance. The StorageConfigManger needs to support multi-rocksdb instance as well. 
The idea is to add set_shared_block_cache_capacity into TabletFactory trait so that it can work for both single and multi-rocks db.

tonyxuqqi:

Issue Number: Close #12961

What's Changed:

```commit-message
refactor StorageConfigManger to support multi-rocksdb
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/server/src/server.rs
```diff
@@ -1618,6 +1620,7 @@ impl<CER: ConfiguredRaftEngine> TiKvServer<CER> {
             builder = builder.block_cache(cache);
         }
         let factory = builder.build();
```
BornChanger, line 1622: seems redundant assignment here

BornChanger (changes requested):

components/server/src/server.rs
BornChanger: How about rewriting as ` let factory = Arc::new(builder.build());`

ethercflow: +1

ethercflow: +1

tonyxuqqi: done

ethercflow:
@tonyxuqqi CI failed, please check.

components/engine_test/src/lib.rs
```diff
@@ -282,6 +282,19 @@ pub mod kv {
         fn clone(&self) -> Box<dyn TabletFactory<KvTestEngine> + Send> {
```
BusyJay, line 282: Interesting. In the prototype, a value local hashmap is used as cache to speed up access of tablet. That will make it !Sync. So `clone` is introduced to address the problem. Now that the implementation doesn't have a cache, I don't think `clone` is necessary anymore.

tonyxuqqi:
> @tonyxuqqi CI failed, please check.

done

components/engine_test/src/lib.rs
tonyxuqqi: We still have the local hashmap in TabletFactoryV2 and it's still Sync. But now we're using Arc<dyn TabletFactory> instead of Box<dyn TabletFactory>, so the customized clone() is not needed anymore.

components/server/src/server.rs
```diff
@@ -1659,11 +1661,12 @@ impl<CER: ConfiguredRaftEngine> TiKvServer<CER> {
         cfg_controller.register(
             tikv::config::Module::Rocksdb,
             Box::new(DBConfigManger::new(
-                Arc::new(factory),
+                std::clone::Clone::clone(&factory),
```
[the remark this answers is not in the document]
tonyxuqqi: TabletFactory had its customized version of clone(). But now it's removed and we can use clone() here directly.

components/engine_test/src/lib.rs
BusyJay: You mean `Arc<Mutex<HashMap<>>>`? Never thought accessing local cache would require lock..

src/storage/config_manager.rs
```diff
@@ -78,18 +73,26 @@ impl<EK: Engine, L: LockManager> ConfigManager for StorageConfigManger<EK, L> {
             if let Some(v) = flow_control.remove("enable") {
                 let enable: bool = v.into();
                 if enable {
-                    for cf in self.kvdb.cf_names() {
-                        self.kvdb
-                            .set_options_cf(cf, &[("disable_write_stall", "true")])
-                            .unwrap();
-                    }
+                    self.tablet_factory.for_each_opened_tablet(
```
BusyJay, line 76: If you can compute the string value ahead of time, then you can only write one loop.

tonyxuqqi: > If you can compute the string value ahead of time, then you can only write one loop.
Good point. I followed the original code but it indeed can be simplified.

BusyJay (approved):

ethercflow (approved):
LGTM

PR #12970, opened 2022-07-06
causal_ts: Implement TSO batch list

### LINKED ISSUE #12794: Workload Error when killing pd Leader in HA test

6.1.0-alpha   5c6018e5711bec3d13f9cdf39d3a84ed9de27809

### Steps to reproduce

1. Deploy RawKV Cluster, API-Version =2 , storage.enable-ttl: true
2. Start Workload R-W
3. Kill PD Leader

### What did you expect?
Workload runs normally.

### What did happened?
Some R-W actions failed.

pingyu:

Issue Number: Close #12794

What's Changed:

```commit-message
Implement TSO batch list to improve tolerance to TSO service fault.
```

Before this PR, TSO is cached by `TsoBatch`, which can only keep at most 262144 TSO as logical clock is 18 bits. This will be a limitation for TSO service fault tolerance. See issue [#12794](https://github.com/tikv/tikv/issues/12794#issuecomment-1152054826).

This PR introduce the `TSO batch list`, which is an ordered list of `TsoBatch`. It aims to:

1. Cache more number of TSO to improve high availability.

2. Fully utilize cached TSO when some regions require latest TSO (e.g. in the scenario of leader transfer, as @BusyJay suggest [here](https://github.com/tikv/tikv/pull/12099#discussion_r832232629)). Other regions without the requirement can still use older TSO cache.

Please note that to limit the change size, the following logics are separated to another PR(s):
* Remove expired batch in an asynchronous manner, to reduce latency.
* Implementation of `get_ts()` with minimal ts requirement in upper `BatchTsoProvider`.

Also, the screenshot of added metrics:
image

### Related changes

No.
No.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Implement TSO batch list to improve tolerance to TSO service fault.
```


pingyu:
> /cc @iosmanthus @haojinming @zeminzhou

PTAL, thanks~

components/causal_ts/src/tso.rs
```diff
@@ -194,50 +353,60 @@ impl<C: PdClient + 'static> BatchTsoProvider<C> {
...
                         }
                         e
                     })?;
-                    debug!("BatchTsoProvider::renew_tso_batch"; "batch renew" => ?batch, "ts" => ?ts);
-                }
-                TS_PROVIDER_TSO_BATCH_SIZE.set(new_batch_size as i64);
+                debug!("BatchTsoProvider::renew_tso_batch";
+                    "tso_batch_list.tso_count" => tso_batch_list.tso_count(), "ts" => ?ts);
+
+                // Should only be invoked after successful renew. Otherwise the TSO usage will is lost,
```
haojinming, line 392:
```suggestion
                // Should only be invoked after successful renew. Otherwise the TSO usage will be lost,
```

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+    }
+}
+
+impl TsoBatchList {
+    pub fn new(capacity: u32) -> Self {
+        Self {
+            inner: Default::default(),
+            tso_count: Default::default(),
+            tso_usage: Default::default(),
+            capacity,
```
haojinming, line 162: suggest `Self { capacity, ..Default::default() }`

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+}
+
+// Inner data structure of batch list.
+// The reasons why `crossbeam_skiplist::SkipMap` is not chosen:
+// 1. In `flush()` procedure, a reader of `SkipMap` can still acquire a batch after the it is removed,
+//    which would violate the causality requirement. The `RwLock<BTreeMap>` avoid this scenario
+//    by lock synchronization.
+// 2. It is a scenario with much more reads than writes. The `RwLock` would be no less efficient
+//    than lock free implementation.
+type TsoBatchListInner = BTreeMap<u64, TsoBatch>;
```
haojinming, line 140: Why not choose LinkedList?  I think the timestamp in batch list is ordered and always pop the min ts and new allocated batch can push to back.

pingyu: To support the `after_ts` argument of `pop()`. Which is used in scenario (such as leader transfer) that some regions need latest TSO, but other regions can still use cached ones.

components/causal_ts/src/config.rs
```diff
@@ -38,6 +45,7 @@ impl Default for Config {
         Self {
             renew_interval: ReadableDuration::millis(crate::tso::TSO_BATCH_RENEW_INTERVAL_DEFAULT),
             renew_batch_min_size: crate::tso::TSO_BATCH_MIN_SIZE_DEFAULT,
+            batch_list_capacity: crate::tso::TSO_BATCH_LIST_CAPACITY,
```
iosmanthus, line 48: TSO_BATCH_LIST_CAPACITY
```suggestion
            batch_list_capacity: crate::tso::DEFAULT_TSO_BATCH_LIST_CAPACITY,
```

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+        Self {
+            size: batch_size,
+            physical,
+            logical_end,
+            logical_start: AtomicU64::new(logical_start),
+        }
+    }
+
+    /// Count of available TSO in the batch.
+    pub fn count(&self) -> u32 {
```
iosmanthus, line 98:
```suggestion
    pub fn remain(&self) -> u32 {
```

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+    /// The original start timestamp in the batch.
+    pub fn original_start(&self) -> TimeStamp {
+        TimeStamp::compose(self.physical, self.logical_end - self.size as u64)
+    }
+
+    /// The excluded end timestamp after the last in batch.
+    pub fn excluded_end(&self) -> TimeStamp {
+        TimeStamp::compose(self.physical, self.logical_end)
+    }
+}
```
iosmanthus, line 112: Since these two functions are twinned, could we rename theme as `start` and `end`?

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+            tso_usage: Default::default(),
+            capacity,
+        }
+    }
+
+    pub fn tso_count(&self) -> u32 {
+        std::cmp::max(self.tso_count.load(Ordering::Relaxed), 0) as u32
+    }
+
+    pub fn tso_usage(&self) -> u32 {
```
iosmanthus, line 170:
```suggestion
    pub fn usage(&self) -> u32 {
```

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+        self.tso_usage.load(Ordering::Relaxed)
+    }
 
-        if physical < self.physical
-            || (physical == self.physical && logical_start < self.logical_end)
+    pub fn take_and_report_tso_usage(&self) -> u32 {
```
iosmanthus, line 174: ditto, since the method is in the TsoBatchList, we may ignore the prefix `tso`.

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+        let usage = self.tso_usage.swap(0, Ordering::Relaxed);
+        TS_PROVIDER_TSO_BATCH_LIST_COUNTING
+            .with_label_values(&["tso_usage"])
+            .observe(usage as f64);
+        usage
+    }
+
+    // TODO: make async
+    fn remove_batch(&self, key: u64) {
+        if let Some(batch) = self.inner.write().remove(&key) {
```
iosmanthus, line 184: This is a potential bottleneck here, since the lock is a sync op. Should we add a metrics to track down the time to remove batch?

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
 
-        self.size = batch_size;
-        self.physical = physical;
-        self.logical_end = logical;
-        self.logical_start.store(logical_start, Ordering::Relaxed);
-        Ok(())
+        // remove items out of capacity limitation.
+        // TODO: make it async
+        if self.inner.read().len() > self.capacity as usize {
+            if let Some((_, batch)) = self.inner.write().pop_first() {
```
iosmanthus, line 248: How can we be sure that the current len won't exceed the cap too much?

components/causal_ts/src/tso.rs
```diff
@@ -62,55 +69,205 @@ impl TsoBatch {
...
+/// 1. Cache more number of TSO to improve high availability. See issue #12794.
+///    `TsoBatch` can only cache at most 262144 TSO as logical clock is 18 bits.
+/// 2. Fully utilize cached TSO when some regions require latest TSO (e.g. in the scenario of leader transfer).
+///    Other regions without the requirement can still use older TSO cache.
+struct TsoBatchList {
+    inner: RwLock<TsoBatchListInner>,
+
+    /// Number of available TSO.
+    /// Using signed integer for avoiding a wrap around huge value as it's not precisely counted.
+    tso_count: AtomicI32,
```
iosmanthus, line 124: Why should we need this field, since we only use this field to update `tso_usage`?

components/causal_ts/src/tso.rs
pingyu: I'm afraid that `start` & `end` is ambiguous. The reader will not know whether it is the start of original batch, or start of current available. Similar to the `end`, one must read the codes to know it is inclusive or exclusive.

components/causal_ts/src/tso.rs
pingyu: `remove_batch` will be changed to an asynchronous method in next PR.

components/causal_ts/src/tso.rs
pingyu: The push is the only method to append a list node, and it will be append more that one node in echo call. So I think it is enough to check only once.

pingyu:
@haojinming @iosmanthus 
All comments are addressed or replied. And I also improve the algorithm for calculating new batch size, please refer to comments at the header of "tso.rs".
PTAL, thanks~

components/causal_ts/src/metrics.rs
```diff
@@ -20,7 +20,14 @@ lazy_static! {
         "tikv_causal_ts_provider_tso_batch_renew_duration_seconds",
         "Histogram of the duration of TSO batch renew",
         &["result", "reason"],
-        exponential_buckets(1e-6, 2.0, 20).unwrap() // 1us ~ 1s
+        exponential_buckets(1e-4, 2.0, 20).unwrap() // 0.1ms ~ 104s
```
iosmanthus, line 23: The last upper bound of the bucket should be `1e-4 * 2**19 == 52.4288`?

haojinming (approved):
LGTM~

pingyu:
> /cc @BusyJay @Connor1996

PTAL, thanks ~

components/causal_ts/src/tso.rs
```diff
@@ -63,56 +93,203 @@ impl TsoBatch {
...
+
+    /// Length of batch list. It is used to limit size for efficiency, and keep
+    /// batches fresh.
+    capacity: u32,
+}
+
+// Inner data structure of batch list.
+// The reasons why `crossbeam_skiplist::SkipMap` is not chosen:
+// 1. In `flush()` procedure, a reader of `SkipMap` can still acquire a batch
+// after the it is removed,    which would violate the causality requirement.
```
Connor1996, line 164: bad format, same for other comments

components/causal_ts/src/tso.rs
```diff
@@ -63,56 +93,203 @@ impl TsoBatch {
...
+        };
+        while let Some((key, batch)) = iter.next() {
+            if let Some((ts, is_used_up)) = batch.pop() {
+                let key = *key;
+                drop(inner);
+                self.tso_usage.fetch_add(1, Ordering::Relaxed);
+                self.tso_remain.fetch_sub(1, Ordering::Relaxed);
+                if is_used_up {
+                    // TODO: make it async
+                    self.remove_batch(key);
```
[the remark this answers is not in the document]
pingyu: The previous batches will be gradually released in [`TsoBatchList::push()`](https://github.com/pingyu/tikv/blob/3c5f1e785eb4d0474108bf3f5a7499d864153353/components/causal_ts/src/tso.rs#L275), when the number of batches in `TsoBatchList` excceed the [`TsoBatchList::capacity`](https://github.com/pingyu/tikv/blob/3c5f1e785eb4d0474108bf3f5a7499d864153353/components/causal_ts/src/tso.rs#L158).

components/causal_ts/src/tso.rs
pingyu: Done, PTAL~

components/causal_ts/src/tso.rs
```diff
@@ -197,51 +409,61 @@ impl<C: PdClient + 'static> BatchTsoProvider<C> {
...
-        match pd_client.batch_get_tso(new_batch_size).await {
+        let tso_remain = tso_batch_list.remain();
+        let new_batch_size =
+            Self::calc_new_batch_size(tso_batch_list.clone(), renew_parameter, need_flush);
+
+        TS_PROVIDER_TSO_BATCH_LIST_COUNTING
+            .with_label_values(&["tso_remain"])
+            .observe(tso_remain as f64);
+        TS_PROVIDER_TSO_BATCH_LIST_COUNTING
+            .with_label_values(&["new_batch_size"])
```
Connor1996, line 424: use static metrics

components/causal_ts/src/tso.rs
```diff
@@ -63,56 +95,207 @@ impl TsoBatch {
...
+    /// It is used for the scenario that some regions have causality
+    /// requirement (e.g. after transfer, the next timestamp of new leader
+    /// should be larger than the store where it is transferred from).
+    /// `after_ts` is included.
+    pub fn pop(&self, after_ts: Option<TimeStamp>) -> Option<TimeStamp> {
+        let inner = self.inner.read();
+        let mut iter = match after_ts {
+            Some(after_ts) => TsoBatchListIter::Range(inner.range((
+                Included(&after_ts.into_inner()),
+                Included(&TimeStamp::max().into_inner()),
```
Connor1996, line 233: Unbounded?

components/causal_ts/src/tso.rs
```diff
@@ -63,56 +95,207 @@ impl TsoBatch {
...
+            ))),
+            None => TsoBatchListIter::Iter(inner.iter()),
```
Connor1996, line 235: use `inner.range(..)` to avoid `TsoBatchListIter`?

pingyu:
@Connor1996 All comments are addressed. PTAL, thanks ~

components/causal_ts/src/tso.rs
```diff
@@ -227,14 +207,11 @@ impl TsoBatchList {
...
         let inner = self.inner.read();
-        let mut iter = match after_ts {
-            Some(after_ts) => TsoBatchListIter::Range(inner.range((
-                Included(&after_ts.into_inner()),
-                Included(&TimeStamp::max().into_inner()),
-            ))),
-            None => TsoBatchListIter::Iter(inner.iter()),
-        };
-        while let Some((key, batch)) = iter.next() {
+        let range = after_ts.map_or_else(
```
Connor1996, line 210: match is more readiable

components/causal_ts/src/tso.rs
```diff
@@ -377,14 +354,14 @@ impl<C: PdClient + 'static> BatchTsoProvider<C> {
...
         Self::renew_tso_batch_internal(self.renew_request_tx.clone(), need_flush, reason).await
     }
 
     async fn renew_tso_batch_internal(
         renew_request_tx: Sender<RenewRequest>,
         need_flush: bool,
-        reason: &str,
+        reason: TsoBatchRenewReason,
     ) -> Result<()> {
         let start = Instant::now();
```
Connor1996, line 366: use corase_now

components/causal_ts/src/tso.rs
```diff
@@ -400,9 +377,9 @@ impl<C: PdClient + 'static> BatchTsoProvider<C> {
             .map_err(|_| box_err!("renew response channel is dropped"))
             .and_then(|r| r.map_err(|err| Error::BatchRenew(err)));
 
-        let label = if res.is_ok() { "ok" } else { "err" };
-        TS_PROVIDER_TSO_BATCH_RENEW_DURATION
-            .with_label_values(&[label, reason])
+        TS_PROVIDER_TSO_BATCH_RENEW_DURATION_STATIC
```
Connor1996, line 380: Do we really need to distinguish the types for the duration?

pingyu: The purpose is as following:

1. To improve the caching algorithm and reduce the chances of TSO "used up", as "used up" will prolong duration of `get_ts()`. The metric can tell us how often "used-up" happens.
2. To judge that how seriously "flush" affects performance of `get_ts()`. It will also help to evaluate the effect that we add `after_ts` to `get_ts()`.

components/causal_ts/src/tso.rs
pingyu: Done, PTAL~

components/causal_ts/src/tso.rs
pingyu: Done, PTAL~

Connor1996 (approved):
LGTM

Issue #12876, opened 2022-06-21
Support 2 Replicas With 1 Log Only Mechanism


Connor1996:
## Development Task

On cloud env, TiKV is equipped with AWS EBS storage for a single node which provides the durability of 99.99%. Moreover, TiKV organizes 3 Raft replicas on top of that, which may be overkill for durability. To reduce cost, we can introduce a mechanism named "2 Replicas With 1 Log Only" instead of 3 replicas to save 30% storage resources and cut down about 100%(1 core) CPU usage empirically on the cloud. Compared with the original 3 replicas architecture, the "1 Log Only" replica stores raft logs but does not apply it, which still ensures data consistency through the Raft protocol.

## Task Breakdown

- TiKV
  - [x] modify kvproto, add witness field @Connor1996 
    - https://github.com/pingcap/kvproto/pull/928
    - https://github.com/pingcap/kvproto/pull/975
  - [x] Precedent code clean
    - https://github.com/tikv/tikv/pull/13244
    - https://github.com/tikv/tikv/pull/13307
    - https://github.com/tikv/tikv/pull/13377
  - [x] raftstore process to support witness @Connor1996 https://github.com/tikv/tikv/pull/12972
    - [x] forbid write and read for witness
    - [x] reject transfer leader command for witness
    - [x] skip apply write command for witness
    - [x] reject generating snapshot when witness is leader
  - [x] witness skip raft log gc if has lagging peer @ethercflow https://github.com/tikv/tikv/pull/13869
    - [x] add replicated index in CompactLog command https://github.com/pingcap/kvproto/pull/1023
    - [x] fix panic https://github.com/tikv/tikv/pull/14054
  - [x] witness take low priority election @ethercflow https://github.com/tikv/tikv/pull/13491
    - [x] Adjust raft-rs impl to solve compatibility issue https://github.com/tikv/raft-rs/pull/495
  - [x] witness conf change 
    - [x] adaption to snapshot process, witness use empty snap @Connor1996  https://github.com/tikv/tikv/pull/12972
    - [x] kvproto add a new admin command https://github.com/pingcap/kvproto/pull/1004   
    - [x] add switch witness operator @ethercflow https://github.com/tikv/pd/pull/5634
    - [x] support non-witness -> witness @ethercflow https://github.com/tikv/tikv/pull/13491
    - [x] support witness -> non-witness @ethercflow https://github.com/tikv/tikv/pull/13491
[...]

PR #12972, opened 2022-07-06
raftstore: Introduce witness peer

Connor1996:

Issue Number: Ref https://github.com/tikv/tikv/issues/12876

What's Changed:

```commit-message
Introduce witness peer
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

```release-note
Introduce witness peer
```


Connor1996:
More tests are on the way, you can review the overall process first. @BusyJay @ethercflow

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1797,23 +1799,22 @@ where
...
+            ConfChangeType::RemoveNode => &PEER_ADMIN_CMD_COUNTER.remove_peer,
+            ConfChangeType::AddLearnerNode => &PEER_ADMIN_CMD_COUNTER.add_learner,
+        };
+        metrics.all.inc();
     }
 
     pub fn inc_success(cct: ConfChangeType) {
-        write_metric(cct, "success")
+        let metrics = match cct {
+            ConfChangeType::AddNode => &PEER_ADMIN_CMD_COUNTER.add_peer,
```
BusyJay, line 1813: How about making these metrics changes in a separate PR?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4719,6 +4748,25 @@ where
...
+        if let Err(e) = util::check_peer_id(msg, self.fsm.peer.peer_id()) {
+            self.ctx
+                .raft_metrics
+                .invalid_proposal
+                .mismatch_peer_id
+                .inc();
+            return Err(e);
+        }
+
+        // Forbid reads and writes when it's a witness except transfer leader
```
BusyJay, line 4761: Technically speaking, witness can write. Though not supporting read is useless when txn kv is used.

components/raftstore/src/store/peer.rs
```diff
@@ -229,6 +230,7 @@ bitflags! {
         const SPLIT          = 0b0000_0010;
         const PREPARE_MERGE  = 0b0000_0100;
         const COMMIT_MERGE   = 0b0000_1000;
+        const NO_ADMIN       = 0b0001_0000;
```
BusyJay, line 233: `split`, `prepare_merge` and `commit_merge` are exclusive. How about define the struct as a struct instead of bitflags? So we can avoid running out of bits.

components/raftstore/src/store/fsm/peer.rs
Connor1996: Yes, and stop write can accelerate the transfer leader speed

components/raftstore/src/store/fsm/apply.rs
Connor1996: it is just a 10-line simple change, with no need for a separate PR.

components/raftstore/src/store/peer.rs
Connor1996: Do you mean to define the struct in protobuf? I think no need. We can just extend the bits type to u32 when it's exhausted and it doesn't break compatibility.

BusyJay: Just in Rust:
```
pub struct ProposalContext {
    admin_type: AdminType,
    sync_log: bool,
}

pub enum AdminType {
    Split = 0b0000_0010,
    PREPARE_MERGE = 0b0000_0100,
    COMMIT_MERGE = 0b0000_1000,
}
```

components/raftstore/src/store/peer.rs
```diff
@@ -4343,6 +4515,7 @@ where
         }
 
         if !req.has_admin_request() {
+            ctx.insert(ProposalContext::NO_ADMIN);
```
BusyJay, line 4518: I don't think it's worthy just for an uncommon role to add a flag for all normal requests. You can check for encoding instead. Technically, if it has admin requests, it must contain admin tag.

Connor1996: It's costly. I must walk through the previous bytes to check the admin tag, which is close to the overhead of decoding. 
image

components/raftstore/src/store/peer.rs
Connor1996: What's the difference? We still need to encode it as bitflags to keep compatibility.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4941,15 +4983,17 @@ where
...
-                .inc();
-            return Err(e);
+
+        // Forbid reads and writes when it's a witness except transfer leader
+        if self.fsm.peer.is_witness() {
+            if !(msg.has_admin_request()
+                && msg.get_admin_request().get_cmd_type() == AdminCmdType::TransferLeader)
+            {
+                self.ctx.raft_metrics.invalid_proposal.witness.inc();
+                return Err(Error::RecoveryInProgress(self.region_id()));
```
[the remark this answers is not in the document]
Connor1996: yeah, will add a dedicated error, leave a todo here.

components/raftstore/src/store/peer.rs
BusyJay: That's not true. Generally, it only needs to decode 3 varints. The first varint is header field, the second varint is the header length. The third varint is request tag. It can return if the third tag is requests, admin_request or status_request.

components/raftstore/src/store/peer.rs
BusyJay: It needs to be encoded as a u8 for compatibility. But we can now assign 0b0000_0011 to new value.

components/raftstore/src/store/peer.rs
Connor1996: I see, the payload size can be gotten from the byte right after the tag.

Connor1996: done

components/raftstore/src/store/peer.rs
```diff
@@ -4195,6 +4351,36 @@ where
...
+                ));
+            }
+            if s.get_role() != t.get_role() {
+                return Err(box_err!(
+                    "mismatch peer role, source peer {:?}, target peer {:?}",
+                    s,
+                    t
+                ));
+            }
+            if s.get_is_witness() != t.get_is_witness() {
```
rleungx, line 4375: We also need to consider it in PD.

Connor1996: Sure

Cargo.toml
```diff
@@ -338,7 +338,7 @@ codegen-units = 4
 
 [profile.test]
 opt-level = 0
-debug = 0
+debug = 0
```
BusyJay, line 341: Why the change?

components/batch-system/Cargo.toml
```diff
@@ -16,6 +16,7 @@ file_system = { path = "../file_system", default-features = false }
 lazy_static = "1.3"
 online_config = { path = "../online_config" }
 prometheus = { version = "0.13", default-features = false, features = ["nightly"] }
+prometheus-static-metric = "0.5"
```
BusyJay, line 19: Why?

Cargo.toml
ethercflow, line 341: Looks like an extra whitespace

components/tikv_util/src/time.rs
```diff
@@ -284,6 +284,12 @@ pub enum Instant {
     MonotonicCoarse(Timespec),
 }
 
+impl Default for Instant {
```
BusyJay, line 287: Why not just call `Instant::now()`? Note it's also not implemented for std::time::Instant.

components/raftstore/src/store/peer.rs
```diff
@@ -3823,6 +3903,83 @@ where
...
+        }
+
+        let prs = self.raft_group.raft.prs();
+        // find other peers that are not down
+        let (_, peer) = self
+            .region()
+            .get_peers()
+            .iter()
+            .filter(|peer| peer.id != self.peer.id)
+            .filter(|peer| !peer.is_witness)
```
ethercflow, line 3930: after https://github.com/tikv/tikv/pull/13209 has been merged, should we add a filter to ignore non_witnesses who are waiting for data?

Connor1996: yes, and seems should filter out learner as well

Connor1996: add a todo here

components/raftstore/src/store/peer.rs
```diff
@@ -2980,6 +3042,14 @@ where
...
             );
             let commit_term = self.get_store().term(commit_index).unwrap();
+
+            if self.is_witness() {
+                committed_entries.retain(|e| has_admin_request(e.get_data()));
+                if committed_entries.is_empty() {
+                    return;
+                }
+            }
+
```
rleungx, line 3052: Can it be prejudged?

Connor1996: prejudge it in where?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5248,6 +5305,15 @@ where
             replicated_idx
         };
         assert!(compact_idx >= first_idx);
+
+        if let Some(idx) = witness_replicated_idx {
+            // do not compact the log not replicated to witness, except witness is the most
```
ethercflow, line 5310: I hava a question: 

I see that only one witness is considered here.

``` rust
let witness_peer = self.region().get_peers().iter().find(|p| p.is_witness);
```
 So if there are two witnesses, the first is the most lagged, but the second is not, should we also compact the log and why?

components/raftstore/src/store/peer.rs
```diff
@@ -2275,6 +2325,7 @@ where
                             "region_id" => self.region_id,
                         );
                     }
+                    self.maybe_witness_transfer_leader(ctx);
```
ethercflow, line 2328: I have a question:

As the witness is forbidden to read and write, and if it is a witness, the transfer leader will be executed immediately. So the call to this function can be brought to the front of `StateRole::Leader => {}` to bypass renew leader lease?

components/raftstore/src/store/peer.rs
```diff
@@ -3735,9 +3811,21 @@ where
                         cp
                     ));
                 }
-                (ConfChangeType::RemoveNode, _)
-                | (ConfChangeType::AddNode, PeerRole::Voter)
-                | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {}
+                (ConfChangeType::RemoveNode, _) => {}
+                (ConfChangeType::AddNode, PeerRole::Voter)
+                | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {
+                    if peer.get_id() == self.peer_id()
```
ethercflow, line 3817: When the witness is converted to non-witness and the non-witness is waiting for apply snapshot, the operator of pd may push witness -> non-witness conf change again, should we filter out this conf change here? Otherwise, the apply thread will print an error message, may cause misjudgment?

Cargo.toml
```diff
@@ -211,7 +211,7 @@ procinfo = { git = "https://github.com/tikv/procinfo-rs", rev = "6599eb9dca74229
 # When you modify TiKV cooperatively with kvproto, this will be useful to submit the PR to TiKV and the PR to
 # kvproto at the same time.
 # After the PR to kvproto is merged, remember to comment this out and run `cargo update -p kvproto`.
-[patch.'https://github.com/pingcap/kvproto']
+# [patch.'https://github.com/pingcap/kvproto']
```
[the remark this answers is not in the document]
Connor1996: It's been commented out before. Someone has done it mistakenly, so I just revert it.

components/raftstore/src/store/peer.rs
Connor1996: Nice catch, it should be checked in `should_renew_lease`.

Connor1996: On second thought, it's unnecessary. It rejects' witness's read in the local reader.

components/raftstore/src/store/fsm/peer.rs
Connor1996: nice catch

components/raftstore/src/store/peer.rs
Connor1996: Yes, I will add a check to ignore duplicated operators.

ethercflow: yes :)

components/raftstore/src/store/peer.rs
```diff
@@ -3823,6 +3909,84 @@ where
...
+                    let elapsed = instant.saturating_elapsed();
+                    if elapsed < ctx.cfg.max_peer_down_duration.0 {
+                        return true;
+                    }
+                } else {
+                    info!("can't find peer heartbeat"; "peer_id" => peer.get_id());
+                }
+                false
+            })
+            .fold((0, None), |(max_matched, chosen), peer| {
```
ethercflow, line 3949: I have question:

Is it possible that after the selected peer is elected, when the witness leader replicates the log to this peer, the peer reject and prompts the witness leader to send a snapshot (but witness leader cann't gen snapshot)?  I'm not sure can the current filter condition + max_matched filter out such peers?

Connor1996: why would the peer prompt send a snapshot?

ethercflow: I mean is it possible that the elected peer is still more behind than the witness leader? So when it receives `MsgAppend` from the witness leader, it is rejected.

ethercflow: Or am I misunderstanding the implementation of raft transfer leader?

Connor1996: `the elected peer is still more behind than the witness leader`, it won't happen as raft log gc logic promises the witness would have the log.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2220,6 +2227,10 @@ where
                 if self.fsm.stopped {
                     return;
                 }
+                if res.apply_state.applied_index < self.fsm.peer.get_store().applied_index() {
```
ethercflow, line 2230: In what scenario does this trigger?

ethercflow (approved):
LGTM

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -6,10 +6,17 @@ use crossbeam::atomic::AtomicCell;
...
+use raft::{RawNode, StateRole, INVALID_ID};
+use raftstore::{
+    store::{
+        fsm::Proposal,
+        metrics::PEER_PROPOSE_LOG_SIZE_HISTOGRAM,
+        util::{Lease, RegionReadProgress},
+        Config, EntryStorage, ProposalQueue, RaftlogFetchTask, ReadDelegate, ReadIndexQueue,
+        ReadIndexRequest, TrackVer, Transport, TxnExt, WriteRouter,
+    },
+    Error,
```
BusyJay, line 19: Why the change?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1029,15 +1029,19 @@ where
...
-            let expect_index = self.apply_state.get_applied_index() + 1;
-            if expect_index != entry.get_index() {
-                panic!(
-                    "{} expect index {}, but got {}, ctx {}",
-                    self.tag,
-                    expect_index,
-                    entry.get_index(),
-                    apply_ctx.tag,
-                );
+            let peer = find_peer_by_id(&self.region, self.id).unwrap();
```
BusyJay, line 1032: How about caching peer or `is_witness` in `ApplyDelegate`? So it doesn't need to search again in every loop.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3073,6 +3097,11 @@ pub struct Apply<C> {
     pub entries_size: usize,
     pub cbs: Vec<Proposal<C>>,
     pub bucket_meta: Option<Arc<BucketMeta>>,
+    // Used for witness, as witness would skip applying write commands, so the applied index in the
```
BusyJay, line 3100: Why not always update the applied index?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -436,16 +438,42 @@ where
     /// Gets a snapshot. Returns `SnapshotTemporarilyUnavailable` if there is no
     /// unavailable snapshot.
     pub fn snapshot(&self, request_index: u64, to: u64) -> raft::Result<Snapshot> {
+        let peer = find_peer_by_id(&self.region, self.peer_id).unwrap();
```
BusyJay, line 441: How about setting a flag instead of searching?

components/raftstore/src/store/fsm/apply.rs
Connor1996: the write commands are not sent to apply worker if it's a witness so the apply system can save CPU time to poll that fsm.

components/raftstore/src/store/fsm/peer.rs
Connor1996: seems it's stale, remove it

components/raftstore-v2/src/raft/peer.rs
Connor1996: by mistake, revert it

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2194,18 +2198,19 @@ where
                     let (role, exist_id, incoming_id) =
                         (exist_peer.get_role(), exist_peer.get_id(), peer.get_id());
 
+                    let is_witness_change = exist_peer.is_witness != peer.is_witness;
                     if exist_id != incoming_id // Add peer with different id to the same store
                             // The peer is already the requested role
-                            || (role, change_type) == (PeerRole::Voter, ConfChangeType::AddNode)
-                            || (role, change_type) == (PeerRole::Learner, ConfChangeType::AddLearnerNode)
+                            || ((role, change_type) == (PeerRole::Voter, ConfChangeType::AddNode) && !is_witness_change)
```
BusyJay, line 2204: What could go wrong if `add_witness` is added as a new command instead of mixing with conf change?

Because changing between witness and non-witness have nothing to do with configuration change, and doesn't even need to utilize joint consensus. Using a separate new command seems more clear to me.

components/raftstore/src/store/fsm/apply.rs
Connor1996: done

components/raftstore/src/store/peer_storage.rs
Connor1996: done

components/raftstore/src/store/fsm/apply.rs
BusyJay: But there is nothing to be changed in conf change function, right? Only those conversions between witness and non-witness go to other command.

Connor1996: I agree it's clearer. But the change in the conf change function is small, using a separate command would introduce a lot of changes. Not benefit shows that it is worth doing that.

BusyJay: > a lot of changes

Why? It should be less than 10 lines of code in apply.

```
fn apply_switch_witness(&self, changes) {
    for peer in changes {
        for p in self.region_mut().mut_peers() {
            if p.id == peer.id {
                *p = peer;
                break;
            }
        }
    }
}
```

Connor1996: need to change all along the path of PD dispatching operator, and the epoch should be increased as well.

components/raftstore/src/store/peer.rs
ethercflow: I think maybe leader should put future non-witness into `wait_data_peers`, as the non-witness peer may `request_snapshot` before leader finish apply `witness->non-witness` conf change. Then leader will send a witness snapshot to the non-witness peer.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -436,16 +444,41 @@ where
...
     /// unavailable snapshot.
     pub fn snapshot(&self, request_index: u64, to: u64) -> raft::Result<Snapshot> {
+        if self.peer.as_ref().unwrap().is_witness {
+            // witness could be the leader for a while, do not generate snapshot now
+            return Err(raft::Error::Store(
+                raft::StorageError::SnapshotTemporarilyUnavailable,
+            ));
+        }
+
+        let for_witness = find_peer_by_id(&self.region, to).unwrap().is_witness;
```
ethercflow, line 454: I think may need to add a judgment if the peer is in `wait_data_peers` See: https://github.com/tikv/tikv/pull/12972/files#r1000411516

components/raftstore/src/store/fsm/apply.rs
BusyJay: > Need to change all along the path of PD dispatching operator

Should it just a function? The algorithm is the same, the only difference is dispatching conf change or the new admin command. Besides, how to tell if a conversion is finished is also different from conf change on PD side. PD needs to check if it's in wait data list. And with a new command, you don't even need to enforce all conf changes have to be done by v2.

Note the change also needs to be ported to v2, using a new command can also make the process less headache.

Connor1996: > Should it just a function? The algorithm is the same, the only difference is dispatching conf change or the new admin command. Besides, how to tell if a conversion is finished is also different from conf change on PD side. PD needs to check if it's in wait data list. 

Actually, it's simple to change on PD side, the structure is extensible enough. What I mean is the process on TiKV side when receiving a new kind of operator and checks when proposing. It does not hard to change, but still needs more effort than changing conf change directly. Though I agree using a new command can also make the process less headache when porting to v2.

Will remove the support of switching witness by conf change in this PR, and use a new command to support in the next PR. Please move on to review the rest of the code.

Connor1996: > And with a new command, you don't even need to enforce all conf changes have to be done by v2.
BTW, it's already been done, all conf changes are handled by change_peer_v2.

Connor1996: /cc @ethercflow

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -899,6 +899,8 @@ where
     term: u64,
     /// The Region information of the peer.
     region: Region,
+    /// The Peer information.
+    peer: metapb::Peer,
```
BusyJay, line 903: Seems `id` field can be removed.

components/raftstore/src/store/peer.rs
```diff
@@ -3828,6 +3941,84 @@ where
...
+        }
+        fail_point!("witness_transfer_leader", |_| {});
+        info!(
+            "witness try to transfer leader";
+            "region_id" => self.region_id,
+            "peer_id" => self.peer.get_id(),
+        );
+
+        if self.raft_group.raft.lead_transferee.is_some() {
+            info!("witness already transferring leader");
```
BusyJay, line 3956: Useless log.

components/raftstore/src/store/peer.rs
```diff
@@ -3828,6 +3941,84 @@ where
...
+            // already being in transfer leader
+            return;
+        }
+
+        let prs = self.raft_group.raft.prs();
```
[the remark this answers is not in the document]
Connor1996: It's not responsive enough for PD to transfer leader for witness, as PD checks the regions one by one periodically and has a transfer-leader scheduling limit which avoids the operator from being added in time.

components/raftstore/src/store/peer_storage.rs
Connor1996: We can validate the snapshot on receiving and reject it to resend a new snapshot.

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
BusyJay: > not responsive enough

Any tests to show the difference? Usually the limit is not the bottleneck but the time of transfer leader's cost is.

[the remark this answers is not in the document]
Connor1996: No test. It's decided by the PD's mechanism. Polling regions a round needs hundreds of milliseconds or even more. It never gives promise that the operator is scheduled at once. And what if there is an existing operator on that region?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -418,6 +418,12 @@ where
     }
 
     fn can_batch(&self, cfg: &Config, req: &RaftCmdRequest, req_size: u32) -> bool {
+        // Do not batch admin request with other requests in one entry, so witness can
+        // just skip applying the entries not having admin request.
+        if req.has_admin_request() {
```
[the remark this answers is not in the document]
Connor1996: It doesn't check the admin request. Note that `get_requests` just returns normal requests excluding admin request.

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
BusyJay: > And what if there is an existing operator on that region?

It doesn't matter. Just like adding an evict leader on a store, all leaders will be evicted right away.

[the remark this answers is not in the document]
Connor1996: But it's not fast enough. What do you think @rleungx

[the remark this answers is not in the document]
BusyJay: What the speed requirement? I prefer to let PD to handle all the leader transfer because PD is supposed to be the component where scheduling decision is made. Every components try to transfer leader will make the system hard to determine and break configured constraints.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
BusyJay: Yes, but `requests` is not empty means `admin_request` is ignored.

components/raftstore/src/store/peer.rs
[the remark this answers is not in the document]
rleungx: I agree that all scheduling decisions should be made by PD. But currently, when the cluster is large, it usually takes several minutes to traverse all regions. Is that accepted?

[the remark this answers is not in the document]
Connor1996: several minutes is really unacceptable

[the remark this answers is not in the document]
Connor1996: @BusyJay I agree the target TiKV chooses may break configured constraints. But placement rule's following operator would schedule it to the right place.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
Connor1996: I see

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1029,15 +1029,18 @@ where
...
-            if expect_index != entry.get_index() {
-                panic!(
-                    "{} expect index {}, but got {}, ctx {}",
-                    self.tag,
-                    expect_index,
-                    entry.get_index(),
-                    apply_ctx.tag,
-                );
+            // the applied index of witness is not continuous, skip index check
+            if !self.peer.is_witness {
```
BusyJay, line 1033: If there are multiple admin commands, and the first one changes it from witness to non-witness, will it panic at L1035?

tests/integrations/server/lock_manager.rs
```diff
@@ -159,13 +159,11 @@ fn must_merge_region(
     cluster.must_put(target_region_key, b"v");
 }
 
-fn find_peer_of_store(region: &Region, store_id: u64) -> Peer {
+fn find_peer(region: &Region, store_id: u64) -> Option<&Peer> {
```
BusyJay, line 162: Not as clear as before. When seeing a function name `find_peer` that accepts a id, it's usually assumed the id is peer id.

components/raftstore/src/store/worker/read.rs
```diff
@@ -782,6 +783,15 @@ where
             return Err(e);
         }
 
+        // Check witness
+        if find_peer_by_id(&delegate.region, delegate.peer_id)
```
BusyJay, line 787: What if it's non-witness, and then change to witness? Is it possible the check pass but before taking a snapshot, data is destroyed?

Connor1996: Seems possible, should recheck the `trackver` after the snapshot is taken. How do we avoid that for requesting a stale read while the peer is removed?

components/raftstore/src/store/fsm/apply.rs
Connor1996: It should be considered in the switch witness PR having a field to indicate the peer is `unavailable`. Add a TODO here.

components/raftstore/src/store/worker/read.rs
BusyJay: It will double check safe_ts.

components/raftstore/src/store/fsm/apply.rs
ethercflow: Yes, this situation has been dealt with in another PR

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1116,7 +1155,7 @@ where
         let term = entry.get_term();
         let data = entry.get_data();
 
-        if !data.is_empty() {
+        if (self.peer.is_witness && !can_witness_skip(entry)) || !data.is_empty() {
```
BusyJay, line 1158: I believe it should be `(!self.peer.is_witness || !can_witness_skip(entry)) && !data.is_empty()`.

BusyJay, line 1158: If it's witness before, but a command changes it to non-witness, and starts applying all following command. It can cause problem if the command depends on previous side effect.

Connor1996: Oh, stupid brain at night

Connor1996: When it changes to non-witness, it should be marked with `unavailable` and stop applying the following commands until the snapshot is applied. This is considered in next PR.

BusyJay: It may be conflict with existing mechanism. Currently, applying snapshot requires applied index is updated to last scheduled index. If applying is stopped, applying snapshot is also blocked.

Connor1996: Add a note to consider `ready_to_handle_pending_snap` in https://github.com/tikv/tikv/pull/13491

components/raftstore/src/store/worker/read.rs
BusyJay: Any update?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -436,16 +444,41 @@ where
...
+        if for_witness {
+            // generate an empty snapshot for witness directly
+            let mut snapshot = Snapshot::default();
```
BusyJay, line 457: Can this be extracted as a function? It can be reused for normal snapshot and v2.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -631,7 +672,15 @@ where
             "state" => ?self.apply_state(),
         );
 
-        Ok(region)
+        Ok(HandleReadyResult::Snapshot(Box::new(
```
BusyJay, line 675: Why not return `(region, for_witness)`? So we can keep `apply_snapshot` itself not tied to handle result and easier to test.

components/raftstore/src/store/worker/read.rs
Connor1996: fixed

Connor1996:
PTAL @BusyJay

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5237,9 +5287,26 @@ where
         let first_idx = self.fsm.peer.get_store().first_index();
         let last_idx = self.fsm.peer.get_store().last_index();
 
+        // replicated_idx excludes witnesses
         let (mut replicated_idx, mut alive_cache_idx) = (last_idx, last_idx);
+        let mut witness_replicated_idx = None;
```
BusyJay, line 5292: Why not just make it `u64::MAX` and assign it with `cmp::min(idx, matched)`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5237,9 +5287,26 @@ where
         let first_idx = self.fsm.peer.get_store().first_index();
         let last_idx = self.fsm.peer.get_store().last_index();
 
+        // replicated_idx excludes witnesses
```
BusyJay, line 5290: Why?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5311,6 +5378,17 @@ where
...
         };
         assert!(compact_idx >= first_idx);
+
+        if let Some(idx) = witness_replicated_idx {
+            // 1. do not compact the log not replicated to witness, except witness is the
+            // most lagging peer
+            // 2. if there is a lagging behind follower, should not compact log, otherwise
+            // witness can't help the follower to catch up if the leader is down
+            if idx != replicated_idx {
+                compact_idx = std::cmp::min(compact_idx, replicated_idx);
```
BusyJay, line 5388: I'm afraid the problem solved by adjusting compact log index will all be brought back by witness.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5311,6 +5378,17 @@ where
...
+            }
+        }
+
         // Have no idea why subtract 1 here, but original code did this by magic.
         compact_idx -= 1;
```
BusyJay, line 5393: It will underflow if replicated index is 0.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5311,6 +5378,17 @@ where
             replicated_idx
         };
         assert!(compact_idx >= first_idx);
+
+        if let Some(idx) = witness_replicated_idx {
+            // 1. do not compact the log not replicated to witness, except witness is the
+            // most lagging peer
+            // 2. if there is a lagging behind follower, should not compact log, otherwise
+            // witness can't help the follower to catch up if the leader is down
+            if idx != replicated_idx {
```
BusyJay, line 5387: This is too strict. You can't expect different nodes have the same replication speed. So they are likely to be different in a heavy write group.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5311,6 +5378,17 @@ where
             replicated_idx
         };
         assert!(compact_idx >= first_idx);
+
+        if let Some(idx) = witness_replicated_idx {
+            // 1. do not compact the log not replicated to witness, except witness is the
+            // most lagging peer
+            // 2. if there is a lagging behind follower, should not compact log, otherwise
```
BusyJay, line 5385: How about delaying witness's compact log? We can include replicated index in compact log request. Witness delete logs only when replicated index > last compact log index.

components/raftstore/src/store/fsm/peer.rs
Connor1996: what's the problem

components/raftstore/src/store/fsm/peer.rs
Connor1996: It's complicated
- If there is no new write, replicated index is update but compact log won't be triggered again, then the witness's log won't be compacted in time
- How does the leader know whether the witness has too many logs

components/raftstore/src/store/fsm/peer.rs
Connor1996: handy for checking witness case below

components/raftstore/src/store/fsm/peer.rs
BusyJay: Like OOM, log gc not in time, etc.

components/raftstore/src/store/fsm/peer.rs
BusyJay: If hibernate is enabled, there is a negotiate stage of hibernate, replicated index can be set in the message. Or the witness can query the lagging peers whether they have catch up logs and then delete its local logs.

components/raftstore/src/store/fsm/peer.rs
Connor1996: It won't have more accumulated logs than before when witness is enabled. When the log gap reaches gc count max, PD would promote the witness to non-witness, then force gc can be performed.

BusyJay: The process of generating snapshot, sending snapshot, applying snapshot can take minutes to finish in v1. I don't think "more accumulated logs than before" is true.

Connor1996:
After discussion, the raft log gc logic is renewed for witness. Remove it from this PR, would add it in the next PR.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1126,43 +1165,44 @@ where
...
-                apply_ctx.commit(self);
-                if self.metrics.written_bytes >= apply_ctx.yield_msg_size
-                    || self
-                        .handle_start
-                        .as_ref()
-                        .map_or(Duration::ZERO, Instant::saturating_elapsed)
-                        >= apply_ctx.yield_duration
-                {
-                    return ApplyResult::Yield;
+            if !(self.peer.is_witness && can_witness_skip(entry)) {
```
BusyJay, line 1168: It's hard to read, `!is_witness || !can_skip` is more readable.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4992,15 +5040,17 @@ where
...
-        // peer_id must be the same as peer's.
-        if let Err(e) = util::check_peer_id(msg, self.fsm.peer.peer_id()) {
-            self.ctx
-                .raft_metrics
-                .invalid_proposal
-                .mismatch_peer_id
-                .inc();
-            return Err(e);
+
+        // Forbid reads and writes when it's a witness except transfer leader
```
BusyJay, line 5044:
```suggestion
        // Forbid requests when it's a witness unless it's transfer leader
```

components/raftstore/src/store/util.rs
```diff
@@ -125,6 +125,29 @@ pub fn is_initial_msg(msg: &eraftpb::Message) -> bool {
...
+    applied_term: u64,
+    for_witness: bool,
+) -> Snapshot {
+    let mut snapshot = Snapshot::default();
+    snapshot.mut_metadata().set_index(applied_index);
+    snapshot.mut_metadata().set_term(applied_term);
+    snapshot
+        .mut_metadata()
+        .set_conf_state(conf_state_from_region(&region));
+    snapshot.set_data({
```
BusyJay, line 140: Strange style. Better use blank line
```

let snap_data = xxx;
snapshot.set_data(snap_data.write_to_bytes().into());
```

components/raftstore/src/store/util.rs
```diff
@@ -907,9 +930,19 @@ pub fn check_conf_change(
             (ConfChangeType::RemoveNode, PeerRole::Voter) if kind != ConfChangeKind::Simple => {
                 return Err(box_err!("{:?}: can not remove voter directly", cp));
             }
-            (ConfChangeType::RemoveNode, _)
-            | (ConfChangeType::AddNode, PeerRole::Voter)
-            | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {}
+            (ConfChangeType::RemoveNode, _) => {}
+            (ConfChangeType::AddNode, PeerRole::Voter)
+            | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {
+                if peer.get_id() == leader.get_id()
```
BusyJay, line 936: Since we use a standalone method for witness, so here we should check whether conf change is used for updating the witness.

components/raftstore/src/store/peer.rs
```diff
@@ -4552,6 +4606,7 @@ where
             1 => peers.get(0).unwrap(),
             _ => peers.choose(&mut rand::thread_rng()).unwrap(),
         };
+        info!("propose transfer leader"; "region_id" => self.region_id, "peer_id" => self.peer.get_id(), "peer" => ?peer);
```
BusyJay, line 4609: Useless log. Transfer leader already triggers more than 6 logs a time.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -597,24 +621,29 @@ where
         for r in destroy_regions {
             write_peer_state(kv_wb, r, PeerState::Tombstone, None)?;
         }
-        write_peer_state(kv_wb, &region, PeerState::Applying, None)?;
 
-        let last_index = snap.get_metadata().get_index();
+        // Witness snapshot is applied atomically as no async applying operation to
+        // region worker, so no need to set the peer state
```
BusyJay, line 626: It doesn't need to set the state to `Applying`, but it needs to persist the region info.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -942,7 +974,14 @@ where
             }
         }
 
-        self.schedule_applying_snapshot();
+        if !res.for_witness {
+            self.schedule_applying_snapshot();
+        } else {
+            // Bypass apply snapshot process for witness as the snapshot is empty, so mark
+            // status as finished directly here
+            let status = Arc::new(AtomicUsize::new(JOB_STATUS_FINISHED));
```
BusyJay, line 982: has_ready should also need to be set to true, otherwise, it will wait for next tick to know whether snapshot is finished.

Connor1996: check L2878 in `fn handle_raft_ready_append()`, it sends a `SnapshotApplied` msg to trigger next round

components/raftstore/src/store/util.rs
```diff
@@ -907,9 +929,22 @@ pub fn check_conf_change(
             (ConfChangeType::RemoveNode, PeerRole::Voter) if kind != ConfChangeKind::Simple => {
                 return Err(box_err!("{:?}: can not remove voter directly", cp));
             }
-            (ConfChangeType::RemoveNode, _)
-            | (ConfChangeType::AddNode, PeerRole::Voter)
-            | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {}
+            (ConfChangeType::RemoveNode, _) => {}
+            (ConfChangeType::AddNode, PeerRole::Voter)
+            | (ConfChangeType::AddLearnerNode, PeerRole::Learner) => {
+                if let Some(p) = region
```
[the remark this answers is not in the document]
Connor1996: I don't get it

components/raftstore/src/store/peer.rs
```diff
@@ -1076,6 +1079,7 @@ where
             skip_bcast_commit: true,
             pre_vote: cfg.prevote,
             max_committed_size_per_ready: MAX_COMMITTED_SIZE_PER_READY,
+            priority: if peer.is_witness { 0 } else { 1 },
```
BusyJay, line 1082: I'm afraid this make election fail. For example, there are 5 replicas, and in the loop of rolling update. 3 of them are using the new binary, so their priority is 1. the remaining are 0. If PD transfer leader to one of the remaining node, the operation will fail and cause unavailability.

components/raftstore/src/store/util.rs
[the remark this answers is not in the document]
BusyJay: It can be implemented the same way as `check_dup`.

components/raftstore/src/store/peer.rs
Connor1996: Seems the only way to solve that is changing the raft-rs definition, lower number should mean higher priority

Connor1996: PTAL https://github.com/tikv/raft-rs/pull/493

components/raftstore/src/store/util.rs
```diff
@@ -915,6 +942,19 @@ pub fn check_conf_change(
             }
         }
 
+        if let Some(p) = region
```
BusyJay, line 945:
```
if let Some(p) = expr && p.get_is_witness() != peer.get_is_witness() {
}
```

components/raftstore/src/store/peer.rs
```diff
@@ -4070,6 +4090,36 @@ where
             }
         }
 
+        // compare peers' location and role
+        let target_region = req.get_admin_request().get_prepare_merge().get_target();
```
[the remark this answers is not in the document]
Connor1996: removed

components/raftstore/src/store/peer.rs
Connor1996: remove setting priority in this PR, will support in next PRs

Connor1996: PTAL

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1126,43 +1165,44 @@ where
...
+            // we should observe empty cmd, aka leader change,
+            // read index during confchange, or other situations.
+            apply_ctx.host.on_empty_cmd(&self.region, index, term);
         }
 
-        // we should observe empty cmd, aka leader change,
-        // read index during confchange, or other situations.
-        apply_ctx.host.on_empty_cmd(&self.region, index, term);
-
         self.apply_state.set_applied_index(index);
```
BusyJay, line 1206: Clearing callback are executed on the assumption of empty command.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1986,7 +2030,7 @@ where
                 let add_ndoe_fp = || {
                     fail_point!(
                         "apply_on_add_node_1_2",
-                        self.id == 2 && self.region_id() == 1,
+                        self.id() == 2 && self.region_id() == 1,
```
BusyJay, line 2033: Should check the witness won't be changed through conf change.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2263,6 +2308,19 @@ where
                             exist_peer
                         ));
                     }
+                    if exist_peer.get_is_witness() != peer.get_is_witness() {
```
BusyJay, line 2311: How about merging with above condition? Same store with different ID may not be duplicate anyway.

components/raftstore/src/store/util.rs
```diff
@@ -915,6 +942,18 @@ pub fn check_conf_change(
             }
         }
 
+        if region
```
BusyJay, line 945: Any test case?

tests/integrations/raftstore/test_witness.rs
```diff
@@ -0,0 +1,425 @@
...
+
+use std::{iter::FromIterator, sync::Arc, time::Duration};
+
+use futures::executor::block_on;
+use kvproto::metapb;
+use pd_client::PdClient;
+use test_raftstore::*;
+use tikv_util::store::find_peer;
+
+fn become_witness(cluster: &Cluster<ServerCluster>, region_id: u64, peer: &mut metapb::Peer) {
```
BusyJay, line 11: If the implementation is changed from add/remove to the new command, there will be no case covering add/remove.

tests/integrations/raftstore/test_witness.rs
```diff
@@ -0,0 +1,425 @@
...
+    let peer_on_store1 = find_peer(&region, nodes[0]).unwrap().clone();
+    cluster.must_transfer_leader(region.get_id(), peer_on_store1);
+
+    let mut peer_on_store2 = find_peer(&region, nodes[1]).unwrap().clone();
+    // nonwitness -> witness
+    become_witness(&cluster, region.get_id(), &mut peer_on_store2);
+
+    // the other follower is isolated
+    cluster.add_send_filter(IsolationFilterFactory::new(3));
+    for i in 1..100 {
```
BusyJay, line 398: This could last quite a long time.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2956,16 +2958,55 @@ where
...
+        } else {
+            None
+        };
+
+        // If the index of snapshot is not newer than peer's apply index, it
+        // is possibly because there is witness -> non-witness switch, and the peer
+        // requests snapshot from leader but leader doesn't applies the switch yet.
+        // In that case, the snapshot is a witness snapshot whereas non-witness snapshot
+        // is expected.
+        if snap.get_metadata().get_index() < self.fsm.peer.get_store().applied_index()
```
BusyJay, line 2990: Any test case?

components/tikv_util/src/store/region.rs
```diff
@@ -32,9 +32,11 @@ pub fn region_on_same_stores(lhs: &Region, rhs: &Region) -> bool {
     // Because every store can only have one replica for the same region,
     // so just one round check is enough.
     lhs.get_peers().iter().all(|lp| {
-        rhs.get_peers()
-            .iter()
-            .any(|rp| rp.get_store_id() == lp.get_store_id() && rp.get_role() == lp.get_role())
+        rhs.get_peers().iter().any(|rp| {
```
BusyJay, line 35: Any test case?

BusyJay (commented):
There should be test case to cover:
- Adding a witness will not apply any data even though its apply index is increased;
- Although witness will not have any data, but admin commands like split/merge/confchange/compactlog are all executed and epoch is updated as expected;
- Remove witness should create correct tombstone.

Rest LGTM

components/raftstore/src/store/fsm/peer.rs
Connor1996: Covered in next PR. This PR doesn't support switch witness

tests/integrations/raftstore/test_witness.rs
Connor1996: check out  `add a new witness peer` in `test_witness_conf_change`

BusyJay: Why conf change depends on switch witness? 🤣

components/test_raftstore/src/util.rs
```diff
@@ -95,6 +95,27 @@ pub fn must_get_cf_none(engine: &RocksEngine, cf: &str, key: &[u8]) {
     must_get(engine, cf, key, None);
 }
 
+pub fn must_peer_latest_applied<T: Simulator>(
```
BusyJay, line 98: Why not use `wait_applied_index`?

tests/integrations/raftstore/test_witness.rs
```diff
@@ -48,17 +68,112 @@ fn test_witness_split_merge() {
...
+    cp.set_peer(peer);
+    let req = new_admin_request(
+        region.get_id(),
+        region.get_region_epoch(),
+        new_change_peer_v2_request(vec![cp]),
+    );
+    let resp = cluster
+        .call_command_on_leader(req, Duration::from_millis(100))
+        .unwrap();
+    assert!(resp.get_header().has_error());
```
BusyJay, line 140: Should be more specific.

components/test_raftstore/src/util.rs
BusyJay: Seems can be removed.

Issue #12901, opened 2022-06-24
Remove KV WAL


5kbpers:
## Development Task
Disable KV WAL completely and move all states from KV DB to raft engine.

See more details in https://github.com/tikv/rfcs/pull/94.

PR #12978, opened 2022-07-07
raftstore: set seqno in ApplyRes

5kbpers:

Issue Number: Ref #12901 

What's Changed:

```commit-message
Set sequence number to `ApplyRes`
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/tikv_util/src/sequence_number.rs
```diff
@@ -0,0 +1,155 @@
...
+    pending_start_counter: VecDeque<bool>,
+    // counter start from 1, so 0 means no start counter received.
+    last_ack_counter: u64,
+    // (end_counter, sequence number)
+    pending_sequence: BTreeMap<u64, SequenceNumber>,
+    last_sequence: Option<SequenceNumber>,
+}
+
+impl SequenceNumberWindow {
+    pub fn push(&mut self, sn: SequenceNumber) -> Option<SequenceNumber> {
```
tonyxuqqi, line 65: please add the comment for the meaning of the return value.

components/tikv_util/src/sequence_number.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[derive(Default)]
+pub struct SequenceNumberWindow {
+    // The sequence number doesn't be received in order, we need a ordered set to
+    // store received start counters which are bigger than last_start_counter + 1.
+    pending_start_counter: VecDeque<bool>,
+    // counter start from 1, so 0 means no start counter received.
+    last_ack_counter: u64,
+    // (end_counter, sequence number)
+    pending_sequence: BTreeMap<u64, SequenceNumber>,
```
tonyxuqqi, line 60: can it be simplified with Vec instead. Because we can use end_counter - last_ack_counter -1 as index

components/tikv_util/src/sequence_number.rs
```diff
@@ -0,0 +1,155 @@
...
+                version: 0
+            })
+        );
+        let mut sn2 = SequenceNumber::start();
+        let mut sn3 = SequenceNumber::start();
+        let mut sn4 = SequenceNumber::start();
+        sn2.end(4);
+        sn3.end(3);
+        sn4.end(2);
+        assert_eq!(window.push(sn2), None);
```
tonyxuqqi, line 132: you may add a following test to verify the behavior as well
sn2.end(6);
sn3.end(5);
sn4.end(7);
assert_eq!(push(sn2), none));
assert_eq!(push(sn3), Some(..)));
assert_eq!(push(sn4), Some(..)));

components/tikv_util/src/sequence_number.rs
tonyxuqqi: Is it doable?

components/engine_rocks/src/write_batch.rs
```diff
@@ -99,15 +99,17 @@ impl RocksWriteBatchVec {
 }
 
 impl engine_traits::WriteBatch for RocksWriteBatchVec {
-    fn write_opt(&mut self, opts: &WriteOptions) -> Result<()> {
+    fn write_opt(&mut self, opts: &WriteOptions) -> Result<u64> {
         let opt: RocksWriteOptions = opts.into();
         if self.support_write_batch_vec {
```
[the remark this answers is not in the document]
5kbpers: submitted a PR for this and it will be merged soon https://github.com/tikv/rocksdb/pull/307, now we can just leave a comment here and resolve it after rocksdb is updated.

components/tikv_util/src/sequence_number.rs
```diff
@@ -0,0 +1,160 @@
...
+        sn5.end(3);
+        sn2.end(5);
+        sn3.end(2);
+        sn4.end(4);
+        window.push(sn2);
+        assert_eq!(window.committed_seqno(), 1);
+        window.push(sn5);
+        assert_eq!(window.committed_seqno(), 1);
+        window.push(sn3);
+        assert_eq!(window.committed_seqno(), 1);
```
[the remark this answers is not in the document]
5kbpers: Sequence numbers are corresponding to end counters, so here it needs to wait until `sn5` is committed.

components/tikv_util/src/sequence_number.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: Why put it in util?

5kbpers: Other files also have this header. 🤔

components/engine_traits_tests/src/scenario_writes.rs
```diff
@@ -42,17 +42,20 @@ impl WriteScenarioEngine {
...
-                wb.write()
+                wb.write()?;
+                Ok(())
             }
             WriteBatchOtherCf => {
                 let mut wb = self.db.engine.write_batch();
                 wb.put_cf(CF_WRITE, key, value)?;
-                wb.write()
+                wb.write()?;
+                Ok(())
```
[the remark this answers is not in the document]
5kbpers: Other patterns like `NoCf` and `DefaultCf` return a Result.

components/tikv_util/src/sequence_number.rs
BusyJay: What I mean is why put `sequence_number` module in tikv_util? It seems rocksdb related.

5kbpers: Moved to engine_traits/util

components/engine_traits/src/util.rs
```diff
@@ -21,3 +29,192 @@ pub fn check_key_in_range(
         })
     }
 }
+
+lazy_static! {
+    static ref SEQUENCE_NUMBER_COUNTER_ALLOCATOR: AtomicU64 = AtomicU64::new(0);
```
BusyJay, line 34: Why need lazy_static to make it static?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +29,192 @@ pub fn check_key_in_range(
...
+}
+
+pub trait Notifier: Send {
+    fn notify_memtable_sealed(&self, seqno: u64);
+    fn notify_memtable_flushed(&self, seqno: u64);
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SequenceNumber {
+    pub number: u64,
```
BusyJay, line 48: Better use accessor instead of making it fully public.

components/engine_traits/src/util.rs
```diff
@@ -1,5 +1,13 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 2: Can you add more docs to describe how it works and how should it be used?

components/engine_traits/src/util.rs
5kbpers: Previously some variables were inited with CFs, forgot to remove lazy_static after changing the init way.

components/engine_traits/src/util.rs
5kbpers: Added some comments for `SequenceNumber` and `SequenceNumberWindow`

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,209 @@ pub fn check_key_in_range(
...
+    }
+
+    pub fn get_version(&self) -> u64 {
+        self.version
+    }
+}
+
+/// Receive all seqno and their counters, check the last committed seqno (a
+/// seqno is considered committed if all seqno before it was received), and
+/// return the largest sequence number received.
```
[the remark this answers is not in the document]
5kbpers: The number of uncommitted seqno

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,209 @@ pub fn check_key_in_range(
...
+#[derive(Default)]
+pub struct SequenceNumberWindow {
```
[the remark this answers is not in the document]
tabokie: So it's a 消消乐 algorithm👀

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,212 @@ pub fn check_key_in_range(
...
+/// seqno is considered committed if all `start_counter` before its
+/// `end_counter` was received), and return the largest sequence number
+/// received.
+#[derive(Default)]
+pub struct SequenceNumberWindow {
+    // The sequence number doesn't be received in order, we need a buffer to
+    // store received start counters which are bigger than ack_counter.
+    pending_start_counter: VecDeque<bool>,
+    // counter start from 1, so ack_counter as 0 means no start counter received.
+    ack_counter: u64,
```
tabokie, line 101:
```suggestion
    // writes with start_counter <= ack_start_counter are all committed.
    ack_start_counter: u64,
```

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,212 @@ pub fn check_key_in_range(
...
+
+/// Receive all seqno and their counters, check the last committed seqno (a
+/// seqno is considered committed if all `start_counter` before its
+/// `end_counter` was received), and return the largest sequence number
+/// received.
+#[derive(Default)]
+pub struct SequenceNumberWindow {
+    // The sequence number doesn't be received in order, we need a buffer to
+    // store received start counters which are bigger than ack_counter.
+    pending_start_counter: VecDeque<bool>,
```
tabokie, line 99:
```suggestion
    // Status of writes with start_counter starting from ack_start_counter+1.
    write_status_windows: VecDeque<bool>,
```

components/engine_traits/src/util.rs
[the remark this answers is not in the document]
BusyJay: Will it be OOM?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,209 @@ pub fn check_key_in_range(
         })
     }
 }
+
+static SEQUENCE_NUMBER_COUNTER_ALLOCATOR: AtomicU64 = AtomicU64::new(0);
+/// Everytime active memtable switched, this version should be increased.
+pub static VERSION_COUNTER_ALLOCATOR: AtomicU64 = AtomicU64::new(0);
+//. Max sequence number that was synced and persisted.
+pub static SYNCED_MAX_SEQUENCE_NUMBER: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
BusyJay: They should access by methods. For example, what's the point of reducing `VERSION_COUNTER_ALLOCATOR`?

components/engine_traits/src/util.rs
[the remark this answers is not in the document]
5kbpers: Probably, the next PR will introduce metrics to observe its size.

components/engine_traits/src/util.rs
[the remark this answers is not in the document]
5kbpers: Got it

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,221 @@ pub fn check_key_in_range(
...
+        self.memtable_version
+    }
+}
+
+/// Receive all seqno and their counters, check the last committed seqno (a
+/// seqno is considered committed if all `start_counter` before its
+/// `end_counter` was received), and return the largest sequence number
+/// received.
+#[derive(Default)]
+pub struct SequenceNumberWindow {
```
tabokie, line 104: Maybe call it collector?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,221 @@ pub fn check_key_in_range(
...
+    max_received_seqno: u64,
+}
+
+impl SequenceNumberWindow {
+    pub fn push(&mut self, sn: SequenceNumber) {
+        // start_delta - 1 is the index of `write_status_window`.
+        let start_delta = match sn.start_counter.checked_sub(self.ack_start_counter) {
+            Some(delta) if delta > 0 => delta as usize,
+            _ => {
+                assert!(sn.number <= self.max_received_seqno);
```
tabokie, line 122: Each start_counter uniquely maps to a write. When could it be equal?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,221 @@ pub fn check_key_in_range(
...
+            }
+        }
+        self.write_status_window.drain(..acks);
+        self.ack_start_counter += acks as u64;
+        let mut sequences = self
+            .pending_sequence
+            .split_off(&(self.ack_start_counter + 1));
+        std::mem::swap(&mut sequences, &mut self.pending_sequence);
+        if let Some(sequence) = sequences.values().max() {
+            self.committed_seqno = sequence.number;
```
tabokie, line 156: It seems difficult to reason if `pending_sequence` is a monotonic-increase sequence. Maybe just take a `max` here to be safe.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -913,6 +954,8 @@ where
     trace: ApplyMemoryTrace,
 
     buckets: Option<BucketStat>,
+
+    last_write_seqno: Vec<SequenceNumber>,
```
tabokie, line 958: I know we already have fields like `last_flush_applied_index`, but `last_` is really hard to understand. How about `unfinished_write_seqno`?

components/engine_traits/src/util.rs
5kbpers: Window is a better name for me because what it does is similar to tcp window

components/engine_traits/src/util.rs
5kbpers: Sometimes a seqno may be corresponding to multiple `ApplyRes`, so it could be duplicated in the receiving end.

components/engine_traits/src/util.rs
5kbpers: Yes `pending_sequence` is not monotonic-increased.

components/raftstore/src/store/fsm/apply.rs
5kbpers: Good idea

components/engine_traits/src/util.rs
tabokie: So it should be `self.committed_seqno = std::cmp::max(self.committed_seqno, sequence.number);`?

5kbpers: `committed_seqno` is updated after a `start_counter` was committed, `start_counter` is the allocated counter **before** a write and `end_counter` is the value of `WRITE_COUNTER_ALLOCATER` **after** a write. It means the corresponding seqno of a `end_counter` (assume it is 10) is smaller or equal than the seqno at the point of time 10 is allocated as a `start_counter`. Then if a `start_counter` was committed and all `end_counter` smaller than the `start_counter` was pushed into `pending_seqno`, we can ensure that the max value in `pending_seqno` is bigger than current `committed_seqno`, so `comitted_seqno` should never fall back.

5kbpers: Added an assertion here

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -410,6 +410,17 @@ where
...
+
+    // Whether to disable WAL.
+    disable_wal: bool,
+
+    /// A general apply progress for a delegate is:
+    /// `prepare_for` -> `commit` [-> `commit` ...] -> `finish_for`.
+    /// Sometimes an `ApplyRes` was created with an applied_index, but data
+    /// before the applied index still not be written to kvdb, let's call this
+    /// `ApplyRes` uncommitted. Data will finally be written to kvdb in
+    /// `flush`. Here is the count of uncommmitted `ApplyRes`.
```
BusyJay, line 422:
```suggestion
    /// A general apply progress for a delegate is:
    /// `prepare_for` -> `commit` [-> `commit` ...] -> `finish_for`.
    /// Sometimes an `ApplyRes` is created with an applied_index, but data
    /// before the applied index is still not written to kvdb. Let's call the
    /// `ApplyRes` uncommitted. Data will finally be written to kvdb in
    /// `flush`.
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -522,9 +543,17 @@ where
             self.perf_context.start_observe();
             let mut write_opts = engine_traits::WriteOptions::new();
             write_opts.set_sync(need_sync);
-            self.kv_wb_mut().write_opt(&write_opts).unwrap_or_else(|e| {
+            write_opts.set_disable_wal(self.disable_wal);
```
BusyJay, line 546: Note, if WAL is disabled, then it needs to be disabled in all places. Otherwise, after restart, it may cause corruption.

components/engine_traits/src/util.rs
tabokie: How about this case:
```
T1: start=1,                 seqno=2, end=2
T2:         start=2, seqno=1,              end=2
```
When the write in T1 has finished, the `ack_start_counter` should be 1, and `committed_seqno` be 2. When later T2 finished, the `committed_seqno` will jump back to 1.

tabokie: Oh forget it, I got the first step wrong.

tabokie: A better way of reasoning would be 反证法:
Assuming we have an incoming write `A` with `A.seqno < self.committed_seqno`, we have the event order: `A.pre_write -> A.seqno -> X.seqno -> X.post_write`, where `X` is an acknowledged write. This event order implies `A.start_counter <= X.end_counter`, leading to contradition.

tabokie (approved):

components/engine_traits/src/util.rs
BusyJay: Check https://github.com/tikv/rfcs/pull/94/files#diff-0a9e70252aa1bfaf482c5f0bd69e22ea6b627d109ceb5aad8c9742060e5f94e7R106.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -630,11 +665,17 @@ where
...
         // if power failure happen, raft WAL may synced to disk, but kv WAL may not.
         // so we use sync-log flag here.
-        let is_synced = self.write_to_db();
+        let (is_synced, seqno) = self.write_to_db();
 
         if !self.apply_res.is_empty() {
             fail_point!("before_nofity_apply_res");
-            let apply_res = mem::take(&mut self.apply_res);
+            let mut apply_res = mem::take(&mut self.apply_res);
+            if let Some(seqno) = seqno {
```
BusyJay, line 673: Why not make this as part of `write_to_db`?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,227 @@ pub fn check_key_in_range(
...
+        }
+    }
+
+    pub fn post_write(&mut self, number: u64) {
+        self.number = number;
+        self.end_counter = WRITE_COUNTER_ALLOCATOR.load(Ordering::SeqCst);
+    }
+
+    pub fn max(left: Self, right: Self) -> Self {
+        match left.number.cmp(&right.number) {
```
BusyJay, line 84: You mean
```
cmp::max_by_key(left, right, |s| s.number)
```
?

components/engine_traits/src/util.rs
```diff
@@ -21,3 +27,227 @@ pub fn check_key_in_range(
...
+        self.pending_sequence
+            .entry(sn.end_counter)
+            .and_modify(|value| {
+                *value = SequenceNumber::max(*value, sn);
+            })
+            .or_insert(sn);
+        self.write_status_window[start_delta - 1] = true;
+        // Commit seqno of the counter which all smaller counter were received.
+        let mut acks = 0;
+        for received in self.write_status_window.iter() {
```
BusyJay, line 142: Can early return if `start_delta != 1`.

BusyJay (approved):

PR #12980, opened 2022-07-08
Allow undo tombstone, add region force recreation

v01dstar:

Issue Number: Close #xxx

What's Changed:
Enrich tikv-ctl, so that we can **undo** a tombstone, or forcibly recreate a region. 

```commit-message
```

### Related changes

- N/A

### Check List 

Tests 

- Unit test

Side effects

N/A

### Release note 

```release-note
None
```

### DIFF
--- a/cmd/tikv-ctl/src/cmd.rs
+++ b/cmd/tikv-ctl/src/cmd.rs
@@ -337,6 +337,10 @@ pub enum Cmd {
         #[structopt(long)]
         /// force execute without pd
         force: bool,
+
+        #[structopt(long)]
+        /// undo tombstone
+        undo: bool,
     },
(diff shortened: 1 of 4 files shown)


tonyxuqqi (approved):

BusyJay:
Since master branch has online recovery, I don't think these functions are still necessary.

tonyxuqqi:
> Since master branch has online recovery, I don't think these functions are still necessary.

But we probably need to backport this change to older version. So it's still necessary?

BusyJay:
We can backport this, but I don't think we need to land it on master. These commands are quite dangerous, if users mis-use these commands, it may lead to unexpected result.

PR #12984, opened 2022-07-11
tablet: load_tablet should remove the old tablet in the cache.

SpadeA-Tang:

Issue Number: Close #xxx

What's Changed:

```
Remove old tablet in the registry in `load_tablet`.
```

### Related changes

### Check List 

Tests 

- Unit test

### DIFF
--- a/src/server/engine_factory_v2.rs
+++ b/src/server/engine_factory_v2.rs
@@ -19,6 +19,17 @@ pub struct KvEngineFactoryV2<ER: RaftEngine> {
     registry: Arc<Mutex<HashMap<(u64, u64), RocksEngine>>>,
 }
 
+// Extract tablet id and tablet suffix from the path.
+fn get_id_and_suffix_from_path(path: &Path) -> (u64, u64) {
+    let (mut tablet_id, mut tablet_suffix) = (0, 1);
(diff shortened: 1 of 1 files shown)


SpadeA-Tang:
PTAL @tonyxuqqi @5kbpers

tonyxuqqi (approved):

tabokie (commented):
I didn't know we allow randomly changing path like this before. Knowing that I don't think this code is very robust.

I can immediately think of a few cases:
-  `load_tablet(random_path, random_id, random_suffix)` will drop the root db (0, 1) instead of returning an error.
- if `load_tablet` fails at `rename`, the region can't use the old id to get the engine anymore.

SpadeA-Tang:
> I didn't know we allow randomly changing path like this before. Knowing that I don't think this code is very robust.
> 
> I can immediately think of a few cases:
> 
> * `load_tablet(random_path, random_id, random_suffix)` will drop the root db (0, 1) instead of returning an error.
> * if `load_tablet` fails at `rename`, the region can't use the old id to get the engine anymore.

I think this is mainly used in apply snapshot where we will change the suffix to the log index of the snapshot.
Maybe I should only remove the old one if `open_tablet_raw` succeeds.

tabokie (approved):

PR #12993, opened 2022-07-11
raftstore: change send proposal time to waterfall metrics

sticnarf:

Issue Number: Ref #12362

What's Changed:

```commit-message
Durations related to a single query are recorded as waterfall metrics,
which means it records the duration from the very beginning to the
instant when the event happens.

Previously, proposal_send_wait_nanos was an exception. So, this
commit makes this consistent with other metrics.

This commit also adds a Grafana Raft waterfall panel for it.
```
image

### Check List 

Tests 

### Release note 

```release-note
None
```


you06 (commented):
Still 2 usages of removed `tikv_raftstore_proposal_send_wait_duration_seconds` in L14007 and L14098 of `tikv_details.json`. Sorry, I can't provide a link to it because this file is too large to show on the webpage.

sticnarf:
> Still 2 usages of removed `tikv_raftstore_proposal_send_wait_duration_seconds` in L14007 and L14098 of `tikv_details.json`. Sorry, I can't provide a link to it because this file is too large to show on the webpage.

I remove these two panels because it will not lead to meaningful result by just minusing two waterfall metrics

you06 (approved):

PR #12995, opened 2022-07-11
raftstore: Avoid printing error log in case sending CaptureChange message failed

### LINKED ISSUE #12996: Avoid printing error log in case sending SignificantMsg::CaptureChange message failed
## Development Task

The `SignificantMsg::CaptureChange` message is used by CDC and resolved_ts modules. Sometimes it may be sent to stale region due to region change, which may produce an error log. It's not a severe problem, however the error log may make users confused. It would be better if we print a warning log instead of error.

MyonKeminta:

Issue Number: Close #12996

What's Changed:

Print warning log instead of error log when sending CaptureChange message failed.

```commit-message
```

### Related changes

### Check List 

Tests 

-

Side effects

-

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/msg.rs
+++ b/components/raftstore/src/store/msg.rs
@@ -635,6 +635,17 @@ impl<EK: KvEngine> fmt::Debug for PeerMsg<EK> {
     }
 }
 
+impl<EK: KvEngine> PeerMsg<EK> {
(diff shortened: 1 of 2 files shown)


5kbpers (approved):

zhangjinpeng1987 (approved):
LGTM

PR #13000, opened 2022-07-12
Add query processor metrics

### LINKED ISSUE #13027: Track memory usage of query executors
## Development Task
Track the memory usage of Coprocessor executors.  Add a new metric that tracks query executor memory usage, which can be exported to Prometheus and used by Grafana. 

1. The query executors tracked are:

    pub label_enum MemTrace {
         aggr_fast_hash,
         aggr_slow_hash,
         aggr_stream,
         aggr_simple,
         top_n,
         index_scan,
         table_scan,
         limit,
         projection,
         selection,
     }

2. Add a new counter of type IntGauge that can be polled by Prometheus: 
     
    tikv_query_executor_memory_usage

    The global counter per executor is  MEMTRACE_QUERY_EXECUTOR.

3. Add a new trait to components/tidb_query_common/src/metrics.rs4. 
   This trait is implemented by the executors that are sub-types of a common base type. e.g,  the aggregation executors.

    pub trait MemoryTrace {
        fn alloc_trace(&mut self, len: size);
    }

5. Add a new method to BatchExecutor.

    fn alloc_trace(&mut self, len: usize);

6. Add a new data member to all the executors that counts the number of bytes allocated.

   n_bytes: size

7. Implement  the Drop trait for the  executors to track memory release.

The data allocation is tracked mostly in components/tidb_query_executors/src/runner.rs. There is one thing to note, to reduce scope, the memory used by tipb::Chunk is tracked by the executor that produced the data. This means that the lifetime of the tipb::Chunk data counter  is tied to the executor lifetime.

subains:
Signed-off-by: Sunny Bains <sunny.bains@pingcap.com>

Issue Number: Close #13027 

What's Changed:

Add metrics to track coprocessor query executor memory usage.

### Related changes

### Check List 

Tests 

  Start cluster using tiup
  Add the Prometheus query from tikv/metrics/grafana/tikv_details.json to the coprocessor details dashboard
  Connect to the cluster using the MySQL client  and run the SQL.

   create table t (c1 int, c2 int, c3 int);
   insert into t values(-1, null, null), (-10, null, -5), (null, -1, -1), (-10, -10, 0), (null, null, 2), (null, null, 2), (-10, -10, 3), (null, 1, 4);
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  insert into t select * from t;
  select * from t group by c2,c1,c3;
  select * from t order by c2 limit 10;

  You should be able to see the memory usage go up for the relevant query processor and then go down to zero once the query completes.

Side effects

None

### Release note 

```release-note
Support tracking of  Coprocessor Query Executor memory, this should help in diagnosing memory consumption.
```


Cargo.toml
```diff
@@ -299,7 +299,7 @@ opt-level = 1
 
 [profile.dev]
 opt-level = 0
-debug = 0
+debug = 2
```
zhangjinpeng1987, line 302: Why 2 here?

components/tidb_query_executors/src/selection_executor.rs
```diff
@@ -207,6 +214,12 @@ impl<Src: BatchExecutor> BatchExecutor for BatchSelectionExecutor<Src> {
     fn can_be_cached(&self) -> bool {
         self.src.can_be_cached()
     }
+
+    #[inline]
+    fn alloc_trace(&mut self, len: usize) {
+        self.n_bytes += len;
+        MEMTRACE_QUERY_EXECUTOR.limit.add(len as i64);
```
zhangjinpeng1987, line 221: s/limit/selection

components/tidb_query_executors/src/simple_aggr_executor.rs
```diff
@@ -153,6 +177,13 @@ impl<Src: BatchExecutor> AggregationExecutorImpl<Src> for SimpleAggregationImpl
         let rows_len = input_logical_rows.len();
         self.has_input_rows |= rows_len > 0;
 
+        let n_bytes = self.n_bytes;
+
+        // FIXME: Are we counting this multiple times?
+        if rows_len > 0 && self.n_bytes == 0 {
+            self.n_bytes += rows_len * size_of::<usize>();
```
zhangjinpeng1987, line 184: where the new memory allocated?

components/tidb_query_executors/src/selection_executor.rs
subains: Removed.

components/tidb_query_executors/src/simple_aggr_executor.rs
subains: This is the vector of offsets, in this case the input_logical_rows.

tonyxuqqi:
need to specify the issue# in the PR

Cargo.toml
subains: Removed

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
```diff
@@ -338,6 +369,8 @@ impl<Src: BatchExecutor> AggregationExecutorImpl<Src> for FastHashAggregationImp
             }
         };
 
+        self.alloc_trace(self.states_offset_each_logical_row.len() * size_of::<usize>());
```
sticnarf, line 372: Is it counted multiple times? Every time we call `process_batch_input`, we add its size to the total consumed memory. But it's cleared at the first of each `process_batch_input` call.

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
```diff
@@ -237,6 +243,24 @@ pub struct FastHashAggregationImpl {
     group_by_exp: RpnExpression,
     group_by_field_type: FieldType,
     states_offset_each_logical_row: Vec<usize>,
+    /// Number of bytes allocated by the executor.
+    n_bytes: usize,
```
sticnarf, line 247: Where is `n_bytes` maintained when some entries are inserted into the `groups`?

components/tidb_query_executors/src/runner.rs
```diff
@@ -584,56 +584,64 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
             }
-            record_len += result.logical_rows.len();
+        } else {
+            data_len = 0;
         }
 
+        // The data_len is from tipb::Chunk. We track it here, this means that
+        // the the metrics will not track it for the lifetime of tipb::Chunk
+        // but the lifetime of self.outmost_executor.
+        self.out_most_executor
```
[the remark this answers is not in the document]
subains: I've pushed a change that tracks the chunk allocation separately.

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
subains: It subtracts the previous size now.

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
```diff
@@ -416,20 +464,24 @@ where
...
             }
             None => {
                 // Group does not exist, prepare groups.
                 let offset = states.len();
                 states_offset_each_logical_row.push(offset);
                 group.insert(val, offset);
+                n_bytes += size_of::<Result<Option<S>>>() + size_of::<usize>();
                 for aggr_fn in aggr_fns {
                     states.push(aggr_fn.create_state());
+                    n_bytes += size_of::<Box<dyn AggrFunctionState>>();
```
sticnarf, line 478: `size_of::<Box<dyn AggrFunctionState>>` does not include its size on the heap. `std::mem::size_of_val(&*state)` can be used to get the heap space.

subains: I've included the heap length too in the calculation.

components/tidb_query_datatype/src/codec/batch/lazy_column.rs
```diff
@@ -203,6 +203,8 @@ impl LazyBatchColumn {
             }
         }
 
+        ctx.n_bytes += raw_vec_len;
```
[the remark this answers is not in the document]
subains: The raw vector is not added to the memory usage, the memory is counted only when the raw is decoded into columns.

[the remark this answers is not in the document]
sticnarf: Seems okay although it may be not so accurate to just use the raw vector length.

sticnarf (approved):

sticnarf:
Please format the code according to the CI failure.

components/tidb_query_datatype/src/codec/batch/lazy_column.rs
```diff
@@ -208,6 +208,8 @@ impl LazyBatchColumn {
...
```
tonyxuqqi, line 211: is the n_bytes supposed to count the memory of raw_vec or the vec in decoded_column?
In anycase, why the memory usage is just same as `raw_vec_len`

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
```diff
@@ -252,24 +286,34 @@ impl<Src: BatchExecutor> AggregationExecutorImpl<Src> for FastHashAggregationImp
         mut input_physical_columns: LazyBatchColumnVec,
         input_logical_rows: &[usize],
     ) -> Result<()> {
+        self.free_trace(self.states_offset_each_logical_row.len() * size_of::<usize>());
         // 1. Calculate which group each src row belongs to.
         self.states_offset_each_logical_row.clear();
+
+        let rows_len = input_logical_rows.len();
+
+        if rows_len > 0 && self.n_bytes == 0 {
```
tonyxuqqi, line 295: question: why the alloc_trace is called only when n_bytes == 0

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
```diff
@@ -237,6 +248,29 @@ pub struct FastHashAggregationImpl {
...
+            .aggr_fast_hash
+            .sub(self.n_bytes as i64);
+    }
+}
+
+impl MemoryTrace for FastHashAggregationImpl {
+    #[inline]
+    fn alloc_trace(&mut self, len: usize) {
+        self.n_bytes += len;
+        MEMTRACE_QUERY_EXECUTOR.aggr_fast_hash.add(len as i64);
```
tonyxuqqi, line 267: a general question: how is the prometheus gauge implemented internally to avoid the performance issue. Does it have sophisticated thread-local vars or it's just global variable?
If it's the latter, did you try to the run performance comparison with and without this change?

tonyxuqqi (commented):
Did you run the test to see if the memory trace can reflect the memory usage correctly?
The rest LGTM.

components/tidb_query_datatype/src/codec/batch/lazy_column.rs
subains: ctx.n_bytes  in this function counts only the row bytes.

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
subains: It's not a thread local variable. Not sure who has to do the tests. Do we have some documentation about the HW, Config, types of tests etc. to run?

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
subains: It's a one time  allocation, so we only count it the first time.

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
tonyxuqqi: we can use benchbot to launch the test comparison. Manual: https://pingcap.feishu.cn/wiki/wikcn72geTuUtMaJ6VC79IjaAbb#
Tips: use translation function of Feishu Doc to translate it into pure English version:)

Atomic operation is OK for ops around a 100K or so, but will be a bottleneck if the ops can reach 1M/s. 
I guess we're probably fine here. There may be very tiny performance overhead. < 0.5%

components/tidb_query_datatype/src/codec/batch/lazy_column.rs
subains: Fixed.

components/tidb_query_executors/src/fast_hash_aggr_executor.rs
subains: Submitted.

tonyxuqqi (approved):

Cargo.toml
```diff
@@ -338,7 +338,7 @@ codegen-units = 4
 
 [profile.test]
 opt-level = 0
-debug = 0
+debug = 2
```
Connor1996, line 341: why change it

components/tidb_query_common/src/metrics.rs
```diff
@@ -29,16 +30,45 @@ make_auto_flush_static_metric! {
...
     )
     .unwrap();
-}
-
-lazy_static::lazy_static! {
     pub static ref EXECUTOR_COUNT_METRICS: LocalCoprExecutorCount =
         auto_flush_from!(COPR_EXECUTOR_COUNT, LocalCoprExecutorCount);
 }
+
+pub trait MemoryTrace {
```
tabokie, line 71: Shouldn't put this in "metrics.rs".

Connor1996, line 71: It's weird to define `MemoryTrace` trait in `metrics.rs`

metrics/grafana/tikv_details.json
```diff
@@ -27263,6 +27263,104 @@
             "alignLevel": null
           }
         },
+	{
```
tabokie, line 27266: Wrong indent.

components/tidb_query_common/src/metrics.rs
```diff
@@ -29,16 +30,45 @@ make_auto_flush_static_metric! {
...
+
+    pub struct MemTraceGauge : IntGauge {
+        "type" => MemTrace,
+    }
+}
+
+lazy_static! {
+    pub static ref MEMTRACE_QUERY_EXECUTOR: MemTraceGauge = register_static_int_gauge_vec!(
+        MemTraceGauge,
+        "tikv_query_executor_memory_usage",
```
Connor1996, line 56: why not record the memory usage in current memory trace metrics

components/tidb_query_common/src/metrics.rs
subains: Where do you suggest it should go?

subains: Sure, where should it go and for future reference, why do you think it's weird?

Cargo.toml
subains: This was for testing and debugging, it sneaked in by mistake.

gengliqi: You can debug the test if you do the same change.😉

metrics/grafana/tikv_details.json
```diff
@@ -27263,6 +27263,104 @@
...
+          "aliasColors": {},
```
5kbpers, line 27267: Better to put this panel under tikv-detail -> Server.

components/tidb_query_common/src/metrics.rs
tabokie: "component/tikv_alloc_src/trace.rs", it's where the old tracing stuff is put.

components/tidb_query_common/src/metrics.rs
subains: I did go down that path first, the reason I moved away from it are:

- There is no hierarchy, it's a single level counter
- Without any clear benefit it seemed like over engineering

In case I've missed or misunderstood something, can you please elaborate.

components/tidb_query_common/src/metrics.rs
subains: I can move it, but the trace.rs file refers to a different MemoryTrace infrastructure.

components/tidb_query_common/src/metrics.rs
Connor1996: It's hierarchical by `sub_trace`, check `MEMTRACE_ANALYZE` for reference. I think we can make `index_scan/table_scan/...` be the sub-trace for `MEMTRACE_QUERY` which is a sub-trace to `MEMTRACE_ROOT` of coprocessor mod

components/tidb_query_common/src/metrics.rs
Connor1996: Why can't use the existing MemoryTrace infrastructure? I think it's better to unify.

Connor1996:
Any update? It would be helpful to investigate high memory usage problems.

Issue #12063, opened 2022-03-02
Tracking Issue for Load Base Split Improvement


JmPotato:
As mentioned in https://github.com/tikv/pd/issues/4588, Load Base Split can not work well in most or pure scan scenarios. We need to introduce more statistical dimensions to the Load Base Split decision process, such as statistics related to the read CPU usage. This is the tracking issue for some related work.

## Steps

- [x] Add key ranges in read CPU usage collecting (https://github.com/tikv/tikv/issues/11988)
- [x] CPU usage collecting
	- [x] ~gRPC Poll CPU utilization collecting (https://github.com/tikv/tikv/issues/12139)~
- [x] CPU usage-based split process (https://github.com/tikv/tikv/pull/12593)
	- [x] https://github.com/tikv/tikv/issues/12937
	- [x] https://github.com/tikv/tikv/pull/12942
	- [x] https://github.com/tikv/tikv/issues/12943
	- [x] https://github.com/tikv/tikv/pull/12955
	- [x] https://github.com/tikv/tikv/pull/12969

## Unresolved Questions

- [ ] **Observer & Metrics**: We need to settle a way to observe the Load Base Split behavior to help us debug and check the split results.
- [x] ~**Cost Model**: A cost model is needed to evaluate and determine the split effect between the read pool and gRPC poll CPU.~
- [x] **Switch & Mechanism**: What kind of switch or mechanism it should be like to control the CPU usage collecting?
- [ ] **Feedback Functionality**: tikv/pd#5267

PR #13005, opened 2022-07-12
raftstore: make the gRPC poll busy check to consider the average value

JmPotato:

Issue Number: ref #12063.

What's Changed:

```commit-message
Make the gRPC poll busy check to consider the average value to make sure the check is accurate.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/raftstore/src/store/worker/split_controller.rs
+++ b/components/raftstore/src/store/worker/split_controller.rs
@@ -589,6 +589,7 @@ pub struct AutoSplitController {
     max_grpc_thread_count: usize,
     max_unified_read_pool_thread_count: usize,
     unified_read_pool_scale_receiver: Option<Receiver<usize>>,
+    grpc_thread_usage_vec: Vec<f64>,
 }
 
 impl AutoSplitController {
@@ -605,6 +606,7 @@ impl AutoSplitController {
             max_grpc_thread_count,
             max_unified_read_pool_thread_count,
             unified_read_pool_scale_receiver,
+            grpc_thread_usage_vec: vec![],
         }
     }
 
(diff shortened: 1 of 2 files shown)


components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -606,26 +607,50 @@ impl AutoSplitController {
...
             return false;
         }
+        if self.cfg.grpc_thread_cpu_overload_threshold_ratio <= 0.0 {
+            return true;
+        }
+        let avg_grpc_thread_usage = self.get_avg_grpc_thread_usage();
         let grpc_thread_cpu_overload_threshold =
             self.max_grpc_thread_count as f64 * self.cfg.grpc_thread_cpu_overload_threshold_ratio;
-        grpc_thread_usage > 0.0 && grpc_thread_usage >= grpc_thread_cpu_overload_threshold
+        avg_grpc_thread_usage > 0.0 && avg_grpc_thread_usage >= grpc_thread_cpu_overload_threshold
```
AndreMouche, line 653: It seems that `grpc_thread_cpu_overload_threshold` should always positive? 
```suggestion
        avg_grpc_thread_usage >= grpc_thread_cpu_overload_threshold
```

components/raftstore/src/store/worker/split_controller.rs
```diff
@@ -606,26 +607,50 @@ impl AutoSplitController {
...
     }
 
-    fn is_grpc_poll_busy(&self, grpc_thread_usage: f64) -> bool {
+    fn is_grpc_poll_busy(&self) -> bool {
         #[cfg(feature = "failpoints")]
         fail_point!("mock_grpc_poll_is_not_busy", |_| { false });
         if self.max_grpc_thread_count == 0 {
             return false;
         }
+        if self.cfg.grpc_thread_cpu_overload_threshold_ratio <= 0.0 {
```
zhangjinpeng1987, line 647: when will hit this condition?

JmPotato: This config field could be set to zero theoretically, so this is to handle the zero case especially.

components/raftstore/src/store/worker/split_controller.rs
JmPotato: Since it's calculated by `self.max_grpc_thread_count as f64 * self.cfg.grpc_thread_cpu_overload_threshold_ratio`, it could be zero when `self.max_grpc_thread_count` is zero or `self.cfg.grpc_thread_cpu_overload_threshold_ratio` is too small. I think we better make sure the coner case is fully covered here.

AndreMouche: It seems they have been checked on the previous line L643 and L646?

zhangjinpeng1987 (approved):
LGTM

AndreMouche (approved):
LGTM

Issue #13008, opened 2022-07-13
flaky test `test_check_data_dir`


tabokie:
## Bug Report

Details TBD.

tabokie:
Seems to be fixed already.

PR #13009, opened 2022-07-13
tests: do not use `assert!(r.is_ok())` in tests

tabokie:

Issue Number: Ref #13008

What's Changed:

I'm trying to fix a flaky test, but the error isn't printed due to the use of `assert!(r.is_ok())`.

```commit-message
None
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


BusyJay:
Cool, can it be a lint?

tabokie:
Disallowing `is_ok` entirely is an overkill, and clippy doesn't support configuring for complex things like `assert!(..)`. I filed an issue to clippy about this specific lint.

components/tikv_util/src/config.rs
```diff
@@ -1944,7 +1943,7 @@ mod tests {
         let ret = check_data_dir_empty(tmp_path.to_str().unwrap(), "txt");
         assert!(ret.is_err());
         let ret = check_data_dir_empty(tmp_path.to_str().unwrap(), "xt");
-        assert!(ret.is_ok());
+        ret.unwrap();
```
BusyJay, line 1946: Why not merge with the previous line?

components/tikv_util/src/yatp_pool/future_pool.rs
```diff
@@ -472,10 +472,10 @@ mod tests {
         // full
         assert!(spawn_long_time_future(&read_pool, 8, 100).is_err());
 
-        assert!(rx.recv().is_ok());
-        assert!(rx.recv().is_ok());
-        assert!(rx.recv().is_ok());
-        assert!(rx.recv().is_ok());
+        rx.recv().unwrap().unwrap();
```
BusyJay, line 475: Why unwrap twice?

tabokie: I think it should check the task succeed.

components/tikv_util/src/config.rs
tabokie: I didn't bother to type really.

BusyJay: Any updates?

BusyJay (approved):

MyonKeminta (approved):

PR #13010, opened 2022-07-13
server: raise error when bootstrap with a zero store-id

### LINKED ISSUE #13011: server: raise error when bootstrap with a zero store-id
## Development Task

We found that [pd can return 0 when alloc id](https://github.com/tikv/pd/blob/813c88199d8eb531f9ae303ced095ff12e793337/server/id/id.go#L65), usually because of a failure to update base.

A zero store-id is weird, I think we should at least return a error rather than debug here.

CalvinNeo:

Issue Number: Close #13011

What's Changed:

We found that a zero-store id can be generated when alloc id.

A zero store-id is weird, I think we should at least return a error rather than debug here. 

```commit-message
raise error when bootstrap with a zero store-id
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


BusyJay:
I don't get it. If error is returned, it should check the error and return RPC error right?

CalvinNeo:
> I don't get it. If error is returned, it should check the error and return RPC error right?

If pd meets a error in `rebaseLocked`, it will return a err. However, it caller `Alloc` returns `id = 0` instead of forward this err as an RPC error. I don't know why pd does so, however, I think we should print out at lease a error in TiKV side.

BusyJay:
> If pd meets an error in `rebaseLocked`, it will return a err. However, its caller `Alloc` returns `id = 0` instead of forward this err as an RPC error. I don't know why pd did so, however, I think we should print out at least a error on TiKV side when such thing happen.

I think it deserves an issue in PD repository. It seems a bug. /cc @nolouch

src/server/node.rs
```diff
@@ -167,7 +167,11 @@ where
         let mut store_id = self.check_store(&engines)?;
         if store_id == INVALID_ID {
             store_id = self.alloc_id()?;
-            debug!("alloc store id"; "store_id" => store_id);
+            if store_id == 0 {
```
BusyJay, line 170: It should return error instead. `store_id` should not be 0.

CalvinNeo: addressed

BusyJay:
https://github.com/tikv/tikv/blob/master/components/raftstore-v2/src/bootstrap.rs#L75 also needs to be checked. How about changing the implementation of `alloc_id` and check if the returned value is zero?

BusyJay (approved):

CalvinNeo:
> Where is the PD issue?

After dig a little deeper, I found may be the zero store id is not caused by pd's returning value.

breezewish (approved):

PR #13013, opened 2022-07-13
raftstorev2: add apply batch system

BusyJay:
Issue Number: Ref #12842

What's Changed:

This is to add another building block for supporting basic write flow.
```commit-message
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/src/tablet.rs
```diff
@@ -0,0 +1,93 @@
...
+
+struct LatestTablet<EK> {
+    data: Mutex<Option<EK>>,
+    version: AtomicU64,
+}
+
+/// Tablet may change during split, merge and applying snapshot. So we need a shared value to
+/// reflect the latest tablet. `CachedTablet` provide cache that can speed up common access.
+#[derive(Clone)]
+pub struct CachedTablet<EK> {
```
tonyxuqqi, line 16: tablet is already cached in TabletFactoryV2. And we can add a method in TabletFactoryV2 to get latest tablet by region id.   I think having all cached tablet in TabletFactory makes it more clear.

BusyJay: No, they are different. `CachedTablet` is supposed to be shared between `Peer`, `Apply` and `ReadDelegate`. Using `TabletFactoryV2` is a bad idea as it needs to lock and hash get for every reads/writes.

tonyxuqqi (approved):

components/raftstore-v2/src/tablet.rs
```diff
@@ -0,0 +1,93 @@
...
+    atomic::{AtomicU64, Ordering},
+    Arc, Mutex,
+};
+
+struct LatestTablet<EK> {
+    data: Mutex<Option<EK>>,
+    version: AtomicU64,
+}
+
+/// Tablet may change during split, merge and applying snapshot. So we need a shared value to
```
Connor1996, line 13: When can it use a stale tablet rather than latest

BusyJay: For example, it there is no on-going admin commands, raftstore can safely assume cached tablet is latest, so no need to refresh cache.

Connor1996: oh, it's just to avoid calling `lock()`

Connor1996 (approved):
LGTM

Issue #12986, opened 2022-07-11
Not able to configure TiKV log level online


zhaopei05:
## Bug Report
Not able to configure TiKV log level online

### What version of TiKV are you using?
v5.4.0

### What operating system and CPU are you using?
RHEL 7.9, X86-64

### Steps to reproduce
Run sql "set config tikv `log.level`='debug';" then "show config where type='tikv' and name='log.level';"
Or run command "tiup ctl tikv --host 127.0.0.1:20160 modify-tikv-config -n log.level -v debug --ignore-version"

### What did you expect?
TiKV log level set to 'debug' successfully.

### What actually happened?
Log level kept 'info' when queried back through sql or got below error with tiup ctl command:
DebugClient::modify_tikv_config: RpcFailure: 2-UNKNOWN "failed to update config, err: \"config log can not be changed\""
Error: exit status 255

### Addition:
Can we also update underlying RocksDB log level with the fix?

glorv:
@zhaopei05 Thanks for the reporting. This is not a bug as we do not support dynamically change config `log.*` currently. You can submit a **Feature Request** and we may support it later. You can refer to [Modify TiKV configuration online](https://docs.pingcap.com/tidb/stable/dynamic-config#modify-tikv-configuration-online) to see all the configs that are supported to change via sql currently.

zhaopei05:
Thanks @glorv for the comments.

zhaopei05:
@glorv BTW, is there particular reason why TiKV doesn't support dynamic log level change? This is supposed to be a very helpful function when we want to grab more debug info w/o interrupting what's already happening by restarting the process.

glorv:
> 

I briefly looked into the code yesterday. Currently, `log.level` controls the log level both for tikv log and rocksdb log. It's not hard to let tikv log support dynamic change. But the log level in rocksdb can't be changed after initialized currently.   @tabokie Do you know if it is possible to let rocksdb also support change log level dynamically?

tabokie:
@glorv RocksDB log level is always set to info even if TiKV is not. No need to consider it here.

glorv:
@zhaopei05  Though we do not change tikv's log trough sql, you can set it through http. e.g.:
```
curl -X PUT -H "Content-Type: application/json" -d '{"log-level":"debug"}' 127.0.0.1:20180/log-level
```
This can change the log level of a single tikv-server. Hope this can make your debug easier. 

[...]

PR #13019, opened 2022-07-14
log: support dynamically change log level via sql

glorv:

Issue Number: ref #12986

What's Changed:

```
- support dynamically change log level via sql
- fix a bug that can not change log level to a lower level(e.g. info --> debug)
```

The bug was introduced in #4935 as it forgot to remove the redundant call of `filter_level()` with a static level parameter.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,7 @@
 use std::{
     cmp,
     collections::HashMap,
+    convert::TryFrom,
     error::Error,
     fs, i32,
     io::{Error as IoError, ErrorKind, Write},
@@ -21,7 +22,7 @@ use api_version::ApiV1Ttl;
 use causal_ts::Config as CausalTsConfig;
 use encryption_export::DataKeyManager;
 use engine_rocks::{
-    config::{self as rocks_config, BlobRunMode, CompressionType, LogLevel},
+    config::{self as rocks_config, BlobRunMode, CompressionType, LogLevel as RocksLogLevel},
     get_env,
     properties::MvccPropertiesCollectorFactory,
     raw::{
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -2690,21 +2669,86 @@ impl Default for File {
...
+    pub level: LogLevel,
+    #[online_config(skip)]
     pub format: LogFormat,
+    #[online_config(skip)]
     pub enable_timestamp: bool,
+    #[online_config(skip)]
     pub file: File,
 }
 
+/// LogLevel is a wrapper type of `slog::Level`
```
tabokie, line 2685: Why?

glorv: to implement some traits like From<ConfigValue>

glorv: Maybe we can implement a new attribue like serde's `with=..` to avoid this kind of boilerplate

tabokie: Only `impl From<LogLevel> for ConfigValue` is required? `impl TryFrom<ConfigValue> for LogLevel` seems to be optional (can be implemented as a function).

Is it possible to avoid it by blanket implement this:
```
impl From<T> for ConfigValue where T: std::string::ToString
```

glorv: `std::string::ToString` is implemented for every type that implements `std::fmt::Display` implicitly. So we may unexpected implement `ConfigValue` for non-string types here.

See: https://github.com/rust-lang/rust/blob/4603ac31b0655793a82f110f544dc1c6abc57bb7/library/alloc/src/string.rs#L2479-L2493

tabokie: Could you give an example case where we don't want it to use `std::fmt::Display`?

glorv: e.g ReadableDuration, it implement `Display` but has its own ConfigValue variant `ConfigValue::Duration`, See:
https://github.com/tikv/tikv/blob/0dff1be50281c72c55c5464751cd733032115ce4/components/tikv_util/src/config.rs#L439

tabokie: I see.

tabokie (approved):

youjiali1995 (approved):

sticnarf:
I'm suspecting this PR causes some regression.

After the PR, I find the slogger thread uses a lot of CPU while there are only very few logs. It seems to me that the `debug!` logs are now formatted and sent to the slogger thread no matter the current log level is.

tabokie:
@glorv Shall we revert it first? Today is code freezing day for v6.2.

sticnarf:
@tabokie I think https://github.com/tikv/tikv/pull/13080 solves the regression. Do you have any other worries?

tabokie:
Oh great, didn't see it.

PR #13043, opened 2022-07-18
log-backup: don't run the cond function if the entry state is removal

### LINKED ISSUE #13044: Br log backup checkpoint lag doesn't move foward
/ # /tikv-server -V
TiKV
Release Version:   6.2.0-alpha
Edition:           Community
Git Commit Hash:   23588581613251bdec283764a604197dd00ca377
Git Commit Branch: heads/refs/tags/v6.2.0-alpha
UTC Build Time:    2022-07-14 11:03:14

Cent OS

### Steps to reproduce
1. Run log backups 

### What did you expect?
1. Log backup checkpoint should move forward normally

### What did happened?
1. Log backup checkpoint doesn't move forward for 10+ hourds
M2hBVeOUv6

YuJuncen:
Signed-off-by: Yu Juncen <yujuncen@pingcap.com>

Issue Number: Close #13044 

What's Changed:

Now, `SubscriptionManager::deregister_region_if` won't call the `cond` argument if the target has been removed by setting the state to `Removal`.

```commit-message
`SubscriptionManager::deregister_region_if` won't call the `cond` argument 
if the target has been removed by setting the state to `Removal`.
This would fix some patterns like:
let mut exists = false;
subs.deregister_region_if(42,|_, _| exists = true);
do_with(exists);
```

### Check List 

Tests 

- Integration test

### Release note 

```release-note
Fixed a bug that caused checkpoint cannot be advanced 
```

### DIFF
--- a/components/backup-stream/src/subscription_manager.rs
+++ b/components/backup-stream/src/subscription_manager.rs
@@ -598,6 +598,15 @@ where
     }
 
     async fn get_last_checkpoint_of(&self, task: &str, region: &Region) -> Result<TimeStamp> {
+        #[cfg(feature = "failpoints")]
+        fail::fail_point!("get_last_checkpoint_of", |hint| Err(Error::Other(
+            box_err!(
+                "get_last_checkpoint_of({}, {:?}) failed because {:?}",
+                task,
+                region,
+                hint
+            )
+        )));
         let meta_cli = self.meta_cli.clone();
         let cp = meta_cli.get_region_checkpoint(task, region).await?;
         debug!("got region checkpoint"; "region_id" => %region.get_id(), "checkpoint" => ?cp);
--- a/components/backup-stream/src/subscription_track.rs
+++ b/components/backup-stream/src/subscription_track.rs
@@ -164,14 +164,26 @@ impl SubscriptionTracer {
         let region_id = region.get_id();
         let remove_result = self.0.get_mut(&region_id);
         match remove_result {
-            Some(o) if o.state == SubscriptionState::Removal => false,
-            Some(mut o) if if_cond(o.value(), region) => {
-                TRACK_REGION.dec();
-                o.value_mut().stop();
-                info!("stop listen stream from store"; "observer" => ?o.value(), "region_id"=> %region_id);
-                true
+            Some(o) => {
+                // If the state is 'removal', we should act as if the region subscription
+                // has been removed: the callback should not be called because somebody may
+                // use this method to check whether a key exists:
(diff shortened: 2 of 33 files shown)


components/backup-stream/src/subscription_track.rs
```diff
@@ -164,10 +164,9 @@ impl SubscriptionTracer {
         let region_id = region.get_id();
         let remove_result = self.0.get_mut(&region_id);
         match remove_result {
+            Some(o) if o.state == SubscriptionState::Removal => false,
             Some(mut o) if if_cond(o.value(), region) => {
```
overvenus, line 168: Please do not add condition in match arm, it's very easy to make mistake, IMO. What about 
```rust
match remove_result {
  Some(mut o) => {
    if o.state == SubscriptionState::Removal {
      // ...
    }
    if if_cond(o.value(), region) {
      // ...
    }
  }
} 
```

components/backup-stream/src/subscription_track.rs
```diff
@@ -164,10 +164,9 @@ impl SubscriptionTracer {
         let region_id = region.get_id();
         let remove_result = self.0.get_mut(&region_id);
         match remove_result {
+            Some(o) if o.state == SubscriptionState::Removal => false,
```
overvenus, line 167: Please add some comments.

overvenus (approved):

components/backup-stream/src/subscription_manager.rs
```diff
@@ -598,6 +598,15 @@ where
     }
 
     async fn get_last_checkpoint_of(&self, task: &str, region: &Region) -> Result<TimeStamp> {
+        #[cfg(feature = "failpoints")]
+        fail::fail_point!("get_last_checkpoint_of", |hint| Err(Error::Other(
```
overvenus, line 602: Failpoint is only enabled when `failpoint` feature is on. So no need to add `failpoints`. 
```suggestion
        fail::fail_point!("get_last_checkpoint_of", |hint| Err(Error::Other(
```

YuJuncen: `fail` in the crate is an optional dependency which would only be required when `failpoints` feature enabled. I am afraid if we remove the config it cannot compile...

YuJuncen: See:
https://github.com/tikv/tikv/blob/21e3bd64e8de517e09836cda4991b446914f3f75/components/backup-stream/Cargo.toml#L11

Issue #13038, opened 2022-07-18
tikv cannot start when open titan


Lily2025:
### What version of TiKV are you using?
./tikv-server -V
 TiKV 
Release Version:   6.2.0-alpha
Edition:           Community
Git Commit Hash:   23588581613251bdec283764a604197dd00ca377
Git Commit Branch: heads/refs/tags/v6.2.0-alpha
UTC Build Time:    2022-07-14 11:03:14
Rust Version:      rustc 1.62.0-nightly (7c4b47696 2022-04-30)
Enable Features:   pprof-fp jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
Profile:           dist_release

### Steps to reproduce
1、open titan
2、run workload
3、restart one tikv
4、tikv cannot start after restarting tikv（upgrade image which include the pr commit pipeline write）

### What did you expect?
tikv can start when open titan

### What did happened?
[...]

PR #12899, opened 2022-06-23
Optimize Commit pipeline performance

ethercflow:
close tikv/tikv#12898

Signed-off-by: Wenbo Zhang <ethercflow@gmail.com>

### What is changed and how it works?

Issue Number: Close #12898 

What's Changed:

```commit-message
Optimize Commit pipeline performance
```

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test
- Manual test
   - sysbench
   - ycsb 

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Optimize Commit pipeline performance
```

PR #13051, opened 2022-07-19
Fix panic when enable titan

ethercflow:
Issue Number: close tikv/tikv#13038

Tests 

- Manual test 
  - sysbench prepare 

Side effects

- Performance regression
    - None
   - None 

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2748,7 +2748,7 @@ dependencies = [
 [[package]]
 name = "librocksdb_sys"
 version = "0.1.0"
-source = "git+https://github.com/tikv/rust-rocksdb.git#2e00e78b945194e8a672e8e078b6c73956e9ace0"
+source = "git+https://github.com/tikv/rust-rocksdb.git#827a5df22cd59dc708c4c6a87dd8735a2312773d"
 dependencies = [
  "bindgen 0.57.0",
  "bzip2-sys",
@@ -2767,7 +2767,7 @@ dependencies = [
 [[package]]
 name = "libtitan_sys"
 version = "0.0.1"
-source = "git+https://github.com/tikv/rust-rocksdb.git#2e00e78b945194e8a672e8e078b6c73956e9ace0"
+source = "git+https://github.com/tikv/rust-rocksdb.git#827a5df22cd59dc708c4c6a87dd8735a2312773d"
 dependencies = [
  "bzip2-sys",
  "cc",
@@ -4584,7 +4584,7 @@ dependencies = [
 [[package]]
 name = "rocksdb"
 version = "0.3.0"
-source = "git+https://github.com/tikv/rust-rocksdb.git#2e00e78b945194e8a672e8e078b6c73956e9ace0"
+source = "git+https://github.com/tikv/rust-rocksdb.git#827a5df22cd59dc708c4c6a87dd8735a2312773d"
 dependencies = [
  "libc 0.2.125",
  "librocksdb_sys",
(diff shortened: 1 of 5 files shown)


tabokie (approved):

tabokie:
Please fix test in test_titan to use multi batch write.

ethercflow:
> Please fix test in test_titan to use multi batch write.

Sure

ethercflow:
> Please fix test in test_titan to use multi batch write.

@tabokie I suddenly realized that multi batch write is enabled by default, and I make sure it at IDC without this PR.

Connor1996 (commented):
Need a case to cover titan write

ethercflow:
> Need a case to cover titan write. Maybe enable titan in `test_raftstore` integration tests

@Connor1996 @tabokie Add a write wb test in `test_titan`, as only [more than one wb](https://github.com/tikv/tikv/pull/12899/files#diff-3be953d50746ae8872ed880f72d3c7226cda7e6470b82b9a5e77bfd2e337711dR101 )  will call `multi_batch_write` directly,  otherwise it will call `write_opt`, that's why existing titan tests are not painic with `enable_multi_write_batch` enabled.

tabokie (commented):
Why not change the `index>0` to use `support_multi_batch_write`? And eventually we can remove the flag in RocksDB.

ethercflow:
> Why not change the `index>0` to use `support_multi_batch_write`? And eventually we can remove the flag in RocksDB.

Because `write_opt` receives `WriteBatch` as a parameter, and `multi_write_batch` receives an `vec`, I think it is more straightforward to judge which interface to use by `index`. (As when we call `db.put`, `db.put` will also call `write` and use `flag` to judge which `write impl` to use).

If use `support_multi_batch_write`, the as_inner need changed to:

``` rust
pub fn as_inner(&self) -> &[RawWriteBatch] {
    if self.index > 0 {
        &self.wbs[0..self.index]
    } else {
         &[self.wbs[0]]
    }
}
```

Write once, judge twice

tabokie:
@ethercflow It's always `&self.wbs[0..=self.index]`. No need to add an if.

ethercflow:
> @ethercflow It's always `&self.wbs[0..=self.index]`. No need to add an if.

@tabokie `&self.wbs[0..0]` will return an empty `[]`.

tabokie:
@ethercflow It's `&self.wbs[0..=0]`, please check.

ethercflow:
> @ethercflow It's `&self.wbs[0..=0]`, please check.

Oh, you're right, I left `=` :(

tests/integrations/storage/test_titan.rs
```diff
@@ -431,3 +431,59 @@ fn test_delete_files_in_range_for_titan() {
         Some((Key::from_raw(b"b"), b"b_value".to_vec())),
     );
 }
+
+#[test]
+fn test_writebatch_for_titan() {
```
Connor1996, line 436: it easily misses the new API after changes in raftstore. Better just choose some basic tests under `tests/integrations/raftstore` with titan enabled

ethercflow: Got it, I'll choose some one, thanks!

ethercflow:
> Why not change the `index>0` to use `support_multi_batch_write`? And eventually we can remove the flag in RocksDB.

@tabokie changed to `support_write_batch_vec`, PTAL again, thanks!

tests/integrations/storage/test_titan.rs
ethercflow: > it easily misses the new API after changes in raftstore. Better just choose some basic tests under `tests/integrations/raftstore` with titan enabled

@Connor1996 I chose a few tests with `titan` enabled  include `put`, `delete`,`snap` interface. PTAL if it's enough to cover, thanks!

PR #13056, opened 2022-07-19
server: do not update background cpu limit when auto-tune is off

### LINKED ISSUE #13055: ThreadTime is collected accidentally even if QuotaLimiter auto-tune is disabled

Latest nightly, after https://github.com/tikv/tikv/pull/12679

### Steps to reproduce

Launch TiKV, with default config (auto-tune is off).

### What did you expect?

Thread cpu time should not be collect in critical paths (e.g. `Storage::get` and `Scheduler::process_write`) when we don't set any cpu limit and turn off auto-tune.

### What did happened?

I can observe large portion of time used for getting thread cpu time from the flamegraph, in process write of the txn scheduler.

image

sticnarf:

Issue Number: Close #13055

What's Changed:

```commit-message
ThreadTime is collected when the cpu limit in the QuotaLimiter
is not infinity.

#12679 updated the background cpu limit even if auto-tune is off,
which is unnecessary. So, that causes additional cost of collecting
thread CPU time in some critical paths.

This commit sets cpu_time_limit of the QuotaLimiter only if auto-tune
is enabled, so we don't waste effort to collect CPU time when
auto-tune is not enabled.
```

### Check List 

Tests 

`ThreadTime` does not appear on the flamegraph now.

### Release note 

```release-note
None
```

### DIFF
--- a/components/server/src/server.rs
+++ b/components/server/src/server.rs
@@ -1292,20 +1292,17 @@ impl<ER: RaftEngine> TiKvServer<ER> {
         };
 
         // Determine the base cpu quota
-        let base_cpu_quota = {
+        let base_cpu_quota =
             // if cpu quota is not specified, start from optimistic case
             if quota_limiter.cputime_limiter(false).is_infinite() {
-                let quota = 1000_f64
+                1000_f64
                     * f64::max(
                         BACKGROUND_REQUEST_CORE_LOWER_BOUND,
                         SysQuota::cpu_cores_quota() * BACKGROUND_REQUEST_CORE_DEFAULT_RATIO,
-                    );
-                quota_limiter.set_cpu_time_limit(quota as usize, false);
-                quota
+                    )
             } else {
                 quota_limiter.cputime_limiter(false) / 1000_f64
-            }
-        };
+            };
 
         // Calculate the celling and floor quota
         let celling_quota = f64::min(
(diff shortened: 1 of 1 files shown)


sticnarf:
cc @BornChanger

Connor1996 (commented):
seems quota limiter won't work if auto-tune is disabled?

sticnarf:
> seems quota limiter won't work if auto-tune is disabled?

If quota is set explicitly (through `foreground_xxx_limit` or `background_xxx_limit`), it will work.

components/server/src/server.rs
```diff
@@ -1292,20 +1292,17 @@ impl<ER: RaftEngine> TiKvServer<ER> {
         };
 
         // Determine the base cpu quota
-        let base_cpu_quota = {
+        let base_cpu_quota =
```
tabokie, line 1295: Very confused about these 1000_f64. L1298 is multiplied, but L1304 is divided.

tabokie: @BornChanger Help us out here

BornChanger: > @BornChanger Help us out here

quota_limiter.cputime_limiter() gives the value in million, while quota for set_cpu_time() accepts value in thousands.

BornChanger: i.e., to express 1 vcpu, quota_limiter.cputime_limiter() is 1,000,000, set_cpu_time() needs the parameter of 1,000. It's the reality currently and makes code ugly.

BornChanger:
I got your point now.  The redundant setting's side effect hurts performance. LGTM~~

BornChanger (approved):
LGTM~~

components/server/src/server.rs
```diff
@@ -1292,20 +1292,17 @@ impl<ER: RaftEngine> TiKvServer<ER> {
...
             // if cpu quota is not specified, start from optimistic case
             if quota_limiter.cputime_limiter(false).is_infinite() {
-                let quota = 1000_f64
+                1000_f64
                     * f64::max(
                         BACKGROUND_REQUEST_CORE_LOWER_BOUND,
                         SysQuota::cpu_cores_quota() * BACKGROUND_REQUEST_CORE_DEFAULT_RATIO,
-                    );
-                quota_limiter.set_cpu_time_limit(quota as usize, false);
```
BornChanger, line 1303: this line of code makes trouble.

sticnarf:
> Please update the issue description, Rest LGTM~

OK. I've clarified in the issue that the regression is because additional thread time collecting happening in `get` or `process_write`.

tabokie (approved):

Connor1996 (approved):
LGTM

Issue #13058, opened 2022-07-19
Use tirocks


BusyJay:
## Development Task

[tirocks](https://github.com/busyjay/tirocks) is the new rocksdb wrapper, it provides more rusty and concise abstractions compared to existing tikv/rust-rocksdb.

Now the development of tirocks is about to finish, we can start migrate tikv to use tirocks.

PR #13059, opened 2022-07-19
engine_trait: introduce status error

BusyJay:
Issue Number: Ref #13058 

What's Changed:

```commit-message
This PR is the first step to prepare for tirocks, the new rocksdb wrapper.

The status error is introduced and iterator trait is refactored to keep
consistent with tirocks.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


tabokie (approved):

components/engine_rocks/src/status.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+pub fn from_engine_traits_error(s: engine_traits::Error) -> String {
+    format!("{:?}", s)
+}
+
+/// A macro that will transform a rocksdb error to engine trait error.
+///
+/// r stands for rocksdb, e stands for engine_trait.
+macro_rules! r2e {
```
Connor1996, line 18: can we implement `From` for error and avoid using macro

tonyxuqqi, line 18: what about rocks2engine_error so it's no need to see the definition to understand what it does.

components/engine_rocks/src/status.rs
```diff
@@ -0,0 +1,31 @@
...
+    ($res:expr) => {{ ($res).map_err($crate::status::from_rocksdb_error) }};
+}
+
+pub(crate) use r2e;
+
+/// A trait that will transform a engine trait error to rocksdb error.
+///
+/// r stands for rocksdb, e stands for engine_trait.
+macro_rules! e2r {
```
tonyxuqqi, line 27: same here, engine2rocks_error?

components/engine_rocks/src/status.rs
BusyJay: The name is too long to be written everywhere.

Connor1996: how about adding intermediate type in engine_rocks

BusyJay: Then the type needs to be constructed every where. You can't build up such connections that just using `?` to convert two foreign types.

Connor1996: how about shortening `from_rocksdb_error` to `r2e`. Writing `xxx().map_err(r2e)` is more readible than `r2e!(xxx())`

Connor1996 (approved):
LGTM

PR #13061, opened 2022-07-20
raftstore: allow exec observers delay deletion of applied ssts

CalvinNeo:

Issue Number: ref #12849

What's Changed:

Some KvEngine can't actually ingest sst, since they do not use RocksDB as a backend. So they may decode the sst, and store them in memory. If we delete the sst file immediately, we may lost data if we have not persist our data in KvEngine.

However, we can't do a persistence for every ingest sst request, this just cost too much. So we introduce this PR to allow a delay deletion of sst files, controlled by post_exec.

```commit-message
allow exec observers delay deletion of applied ssts
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -446,26 +446,27 @@ impl<E: KvEngine> CoprocessorHost<E> {
     /// execution of the next command, including req/resp, apply state,
     /// modified region state, etc. Return true observers think a
     /// persistence is necessary.
-    pub fn post_exec(
+    pub fn post_exec<'a>(
```
BusyJay, line 449: You don't need to specify it explicitly.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -4977,15 +5004,51 @@ mod tests {
         fn post_apply_query(&self, _: &mut ObserverContext<'_>, _: &Cmd) {
             self.post_query_count.fetch_add(1, Ordering::SeqCst);
         }
+
+        fn post_exec_query<'a>(
```
BusyJay, line 5008: Unnecessary lifetime.

BusyJay (approved):

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -4965,6 +4987,10 @@ mod tests {
         cmd_sink: Option<Arc<Mutex<Sender<CmdBatch>>>>,
         filter_compact_log: Arc<AtomicBool>,
         filter_consistency_check: Arc<AtomicBool>,
+        delay_remove_ssts: Arc<AtomicBool>,
+        last_delete_sst_count: Arc<AtomicU64>,
+        last_pending_clean_sst_count: Arc<AtomicU64>,
+        last_pending_handle_sst_count: Arc<AtomicU64>,
```
breezewish, line 4993: What do these atomics used for?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -408,6 +409,11 @@ where
     /// never apply again at first, then we can delete the ssts files.
     delete_ssts: Vec<SstMetaInfo>,
 
+    /// A self-defined engine may be slow to ingest ssts.
+    /// It may move some elements of `delete_ssts` into `pending_clean_ssts` to
+    /// delay deletion. Otherwise we may lost data.
+    pending_clean_ssts: Vec<SstMetaInfo>,
```
breezewish, line 415: Is it more correct to be named as `pending_delete_ssts`?

breezewish (commented):
Generally looks good, but still need a RaftStore module reviewer / committer for one more approval.

components/raftstore/src/store/fsm/apply.rs
CalvinNeo: I think these three fields can be modified by test thread and apply thread, so I'd like use atomics for safety.

components/raftstore/src/coprocessor/mod.rs
```diff
@@ -75,12 +75,21 @@ impl<'a> ObserverContext<'a> {
...
     pub peer_id: u64,
     pub pending_remove: bool,
     pub modified_region: Option<Region>,
 }
 
+/// Context for exec observers of mutation to be applied to ApplyContext.
+pub struct ApplyCtxInfo<'a> {
+    pub pending_handle_ssts: &'a mut Option<Vec<SstMetaInfo>>,
+    pub delete_ssts: &'a mut Vec<SstMetaInfo>,
+    pub pending_delete_ssts: &'a mut Vec<SstMetaInfo>,
```
tonyxuqqi, line 90: where's the code that consumes pending_delete_ssts and delete the SST files?
How can we ensure that these SST files are finally deleted?

components/raftstore/src/coprocessor/dispatcher.rs
```diff
@@ -693,6 +694,19 @@ mod tests {
...
+
+        fn post_exec_admin(
+            &self,
+            ctx: &mut ObserverContext<'_>,
+            _: &Cmd,
+            _: &RaftApplyState,
+            _: &RegionState,
+            _: &mut ApplyCtxInfo<'_>,
+        ) -> bool {
+            self.called.fetch_add(18, Ordering::SeqCst);
```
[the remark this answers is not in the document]
CalvinNeo: In this test, we assign unique index for each observer functions so that we can test  all of the observer functions we want to be called are called. All numbers below 18 are used for other functions, so we use 18 for post_exec.

components/raftstore/src/coprocessor/mod.rs
CalvinNeo: In TiKV's view:
`post_exec` will consumes `pending_delete_ssts`, and it will return true to suggest a immediate write to db. 
In TiFlash's view:
When all data referred in the ssts is write into TiFlash's storage, TiFlash will consume `pending_delete_ssts` and suggest a write.
If TiFlash&Proxy restart before persistence, the recovered applied_index is before IngestSst command. So we can replay this command, and thus ensuring we will not lost track of sst files.

tonyxuqqi: > In TiKV's view: `post_exec` will consumes `pending_delete_ssts`, and it will return true to suggest a immediate write to db. In TiFlash's view: When all data referred in the ssts is write into TiFlash's storage, TiFlash will consume `pending_delete_ssts` and suggest a write. If TiFlash&Proxy restart before persistence, the recovered applied_index is before IngestSst command. So we can replay this command, and thus ensuring we will not lost track of sst files.

IC. Then is there any metrics for pending_delete_ssts? We need to have a way to figure this out.

CalvinNeo: We have a tikv_raft_applying_sst added
image

tonyxuqqi (approved):

breezewish (approved):

Issue #13057, opened 2022-07-19
test_stale_read_while_applying_snapshot always fails in local environment


SpadeA-Tang:
## Bug Report

In my local environment (ubuntu 22 and mac os), test_stale_read_while_applying_snapshot always fails after https://github.com/tikv/tikv/pull/12533. It can pass before this PR. 

Fail message:
```
thread 'cases::test_replica_stale_read::test_stale_read_while_applying_snapshot' panicked at 'message: "peer 2 is not ready, safe_ts 0, region 1" data_is_not_ready { region_id: 1 peer_id: 2 }'
```

SpadeA-Tang:
@tabokie

SpadeA-Tang:
@NingLin-P

tabokie:
@5kbpers Mind take a look?

Issue #12207, opened 2022-03-20
storage: Simplify API version related codes


pingyu:
### Is your feature request related to a problem? Please describe:

1. [`check_api_version`](https://github.com/tikv/tikv/blob/9a43faf4da20389a4e2c262ee8ab8b369a3bfec4/src/storage/mod.rs#L396) & [`check_api_version_range`](https://github.com/tikv/tikv/blob/9a43faf4da20389a4e2c262ee8ab8b369a3bfec4/src/storage/mod.rs#L455) are a bit cumbersome. As client SDK will always insert the correct key prefix for end-users (see [client-go](https://github.com/tikv/client-go/issues/452) & [client-java](https://github.com/tikv/client-java/issues/555)), we can depend on conversion with client SDK, and just do some defensive light-weight verifications. Check just one key in requests will be enough.

2. Many condition statements with API version (see [`raw_scan`](https://github.com/pingyu/tikv/blob/e922d6e2fea54969c336169669b104edf4af5fd9/src/storage/mod.rs#L2085) as example). In raw interfaces, API version related methods include `check_api_version`, `rawkv_cf`, `encode_raw_key`, `encode_raw_value`, and `decode_raw_key`. As API version of storage is determine by configuration, we can earlier bind the type of API version to generic parameter, to reduce conditional jump & code complexity.

### Describe the feature you'd like:

1. Change `check_api_version` & `check_api_version_range` to verifying the FIRST KEY in requests only.

2. Refactor storage as following:
  2.1 A new trait `StorageTrait` providing all storage methods for kv service.
  2.2 Add generic paramter `<API: APIVersion>` to [`Storage`](https://github.com/tikv/tikv/blob/9a43faf4da20389a4e2c262ee8ab8b369a3bfec4/src/storage/mod.rs#L136) struct.
  2.3 `Storage` is instantiated according to configuration item `api_version` in [ctor](https://github.com/tikv/tikv/blob/9a43faf4da20389a4e2c262ee8ab8b369a3bfec4/src/storage/mod.rs#L219), and is hold as `Box<StorageTrait>`.

### _About performance regression of dynamic dispatch `Box<StorageTrait>`_ ###
_By a [quite benchmark](https://gist.github.com/rust-play/9d597e7ee651cffa81bfdcd6fc74cf01), for every request, dynamic dispatch once do not has significant overhead. Benchmark result is as following:_
```sh
$ cargo bench 
...
[...]

PR #13067, opened 2022-07-20
storage: fix API v2 range check for reverse scan

### LINKED ISSUE #13021: API v2 range check will fail in reverse scan
This check might fail in `scan` or `raw_scan` while the reverse flag is set since with reverse scan, the end key is less than the start key (swapped position). 

https://github.com/tikv/tikv/blob/62545b0c5c854b4e42bf37d03dddfab2099ce20c/components/api_version/src/api_v2.rs#L58-L71

For example, if the client with maximum keyspace id: `0xff_ff_ff` tries to `scan("key", "")` reversely, it should encode the request into the protobuf format like:

```rust
ScanRequest {
	reverse: true,
	start_key: [ b'y', 0, 0, 0 ],
	end_key: [ b'x', 0xff, 0xff, 0xff ] ++ "key"
}
```

This check will fail, but the request is valid by the protocol. We should swap the `start_key` and `end_key` before checking the range on the server side.

iosmanthus:
Signed-off-by: iosmanthus <myosmanthustree@gmail.com>

### What is changed and how does it work?

Issue Number: Close #13021

What's Changed:

```commit-message
storage: fix API v2 range check for reverse scan
```

### Related changes

- Need to cherry-pick the release branch

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Fix API v2 range check for reverse scan
```

### DIFF
--- a/components/cdc/tests/integrations/test_cdc.rs
+++ b/components/cdc/tests/integrations/test_cdc.rs
@@ -19,8 +19,7 @@ use crate::{new_event_feed, TestSuite, TestSuiteBuilder};
 
 #[test]
 fn test_cdc_basic() {
-    test_cdc_basic_impl::<api_version::ApiV2>();
-    // test_kv_format_impl!(test_cdc_basic_impl<ApiV1 ApiV2>);
+    test_kv_format_impl!(test_cdc_basic_impl<ApiV1 ApiV2>);
 }
 
 fn test_cdc_basic_impl<F: KvFormat>() {


components/api_version/src/api_v2.rs
```diff
@@ -36,6 +37,14 @@ bitflags::bitflags! {
     }
 }
 
+fn infinite_end_of(prefix: &[u8]) -> [u8; 4] {
+    debug_assert!(prefix.len() >= KEY_PREFIX_SIZE);
+
+    let end_bytes = [prefix[0], prefix[1], prefix[2], prefix[3]];
```
ystaticy, line 43: This operation will be often used later, and it can be encapsulated as a special function to obtain the first four bits of the array.

components/api_version/src/lib.rs
```diff
@@ -261,11 +261,21 @@ mod tests {
         assert_eq!(ApiV1::parse_key_mode(&b"t_a"[..]), KeyMode::Unknown);
         assert_eq!(ApiV1Ttl::parse_key_mode(&b"ot"[..]), KeyMode::Raw);
         assert_eq!(
-            ApiV2::parse_key_mode(&[RAW_KEY_PREFIX, b'a', b'b']),
+            ApiV2::parse_key_mode(&[RAW_KEY_PREFIX, 0, 0, 0, b'a', b'b']),
```
ystaticy, line 264: Suggest use ```DEFAULT_KEY_SPACE_ID``` instead of ```0, 0, 0```

components/api_version/src/lib.rs
```diff
@@ -299,19 +309,19 @@ mod tests {
             KeyMode::TiDB
         );
         assert_eq!(
-            ApiV2::parse_range_mode((Some(b"x\0a"), Some(b"x\0z"))),
+            ApiV2::parse_range_mode((Some(b"x\x00\x00\x00a"), Some(b"x\x00\x00\x00z"))),
```
ystaticy, line 312: Suggest to use variables instead of direct use ```\x00\x00\x00``` and ```\xff\xff\xff```

components/api_version/src/api_v2.rs
iosmanthus: Make sense.

components/api_version/src/lib.rs
iosmanthus: This requires us to add many syntax noises to build literal test cases. The existed prefix `x` or `r` are hard-coded in every case, I think keeping the test case KISS should be enough.

components/api_version/src/lib.rs
iosmanthus: Maybe I should expand the slice to `r\x00\x00\x00ab`.

iosmanthus:
The integration test is blocked by https://github.com/tikv/tikv/issues/13057, the `test_stale_read_while_applying_snapshot` fails stably.

components/cdc/src/initializer.rs
```diff
@@ -237,9 +237,9 @@ impl<E: KvEngine> Initializer<E> {
             Scanner::TxnKvScanner(txnkv_scanner)
         } else {
             let mut iter_opt = IterOptions::default();
-            let (raw_key_prefix, raw_key_prefix_end) = ApiV2::get_rawkv_range();
-            iter_opt.set_lower_bound(&[raw_key_prefix], DATA_KEY_PREFIX_LEN);
-            iter_opt.set_upper_bound(&[raw_key_prefix_end], DATA_KEY_PREFIX_LEN);
+            let (raw_keyspace_start, raw_keyspace_end) = ApiV2::get_rawkv_range();
```
iosmanthus, line 240: @zeminzhou PTAL if this change will break the existed hypothesis.

components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -220,8 +220,8 @@ fn test_cdc_rawkv_basic() {
         )))
         .unwrap();
 
-    // If tikv enable ApiV2, raw key needs to start with 'r';
-    let (k, v) = (b"rkey1".to_vec(), b"value".to_vec());
+    // If tikv enable ApiV2, raw key needs to start with 'r' + keyspace_id;
+    let (k, v) = (b"r\x00\x00\x00key1".to_vec(), b"value".to_vec());
```
zeminzhou, line 224: Test keys starting with 'r' or 'x' are not only in one place, please check other test cases.

iosmanthus: All tests of the CDC module are passed.

components/cdc/tests/integrations/test_cdc.rs
```diff
@@ -19,7 +19,8 @@ use crate::{new_event_feed, TestSuite, TestSuiteBuilder};
 
 #[test]
 fn test_cdc_basic() {
-    test_kv_format_impl!(test_cdc_basic_impl<ApiV1 ApiV2>);
+    test_cdc_basic_impl::<api_version::ApiV2>();
+    // test_kv_format_impl!(test_cdc_basic_impl<ApiV1 ApiV2>);
```
pingyu, line 23: `ApiV1` ought to be tested.

src/storage/mod.rs
```diff
@@ -477,12 +477,23 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
-        ranges: impl IntoIterator<Item = (Option<impl AsRef<[u8]>>, Option<impl AsRef<[u8]>>)>,
-    ) -> Result<()> {
+        ranges: I,
+        reverse: bool,
+    ) -> Result<()>
+    where
+        I: IntoIterator<Item = (Option<K>, Option<K>)>,
+        K: AsRef<[u8]>,
+    {
+        let mut ranges: Box<dyn iter::Iterator<Item = (Option<K>, Option<K>)>> =
```
pingyu, line 491: It seems too complicated. Shall we just check the first item for vector of ranges ?
Please also refer to https://github.com/tikv/tikv/issues/12207.

components/cdc/tests/integrations/test_cdc.rs
iosmanthus: Sorry, I forget to cancel the comments.

src/storage/mod.rs
iosmanthus: The simplification might be placed into another pull request. I think this pull request should preserve the previous behaviour of the TiKV server.

components/api_version/src/api_v2.rs
```diff
@@ -48,26 +51,31 @@ impl KvFormat for ApiV2 {
...
+        if let (Some(start), Some(end)) = range {
+            if start.is_empty() || end.is_empty() {
+                return KeyMode::Unknown;
+            }
+
+            let mode = Self::parse_key_mode(start);
+
+            if mode == KeyMode::Unknown
+                || start[0] == end[0]
+                || (mode == KeyMode::TiDB && end == [start[0] + 1])
```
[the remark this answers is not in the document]
iosmanthus: If we combine them into one function, requiring that `infinite_end_of` could return both `[u8; 1] or [u8; 4]`, this might result in allocating a `Vec` instead of a fix-sized array.

Issue #12618, opened 2022-05-23
checks for snapshot last index is wrong


BusyJay:
## Bug Report

https://github.com/tikv/tikv/blob/1fb8980ccab9ff40c1adc206df52952dab8e8ad8/components/raftstore/src/store/peer_storage.rs#L467-L475

Supposing a leader append a lot of logs without replicating any of them. After several elections, new leader compacts all previous logs. If the old leader rejoin the group, new leader will send it a snapshot. If old leader exits before updating raft state, then the condition will be false and result in corrupted state.

LintianShi:
The root cause of this bug is that both apply state and raft state are updated during snapshot applying. However, apply state and raft state are seperately persisted in kvdb and raftdb. So we cannot write apply state and raft state atomically. If the instance crashes during snapshot applying, raft state may not be persisted successfully while apply state is persisted. When the instance recovers from the snapshot-applying state, the raft state is possible to be inconsistent with apply state.  

The current solution is that we saves a copy of raft state into kvdb which is named `snapshot_raft_state`. When recovering from the snapshot-applying state, we check the `snapshot_raft_state` in kvdb with raft state in raftdb. If the last index of raft state is behind the snapshot_raft_state, we use `snapshot_raft_state` as the correct raft state. Unluckily, the above solution still has a corner case mentioned by Jay. 

To ensure the correctness of recovering, we can use `snapshot_raft_state` to recover the raft state in every situation. The drawback of this solution is that the log entries appended during snapshot applying may be dropped after recovery. However, appending log entries during snapshot applying occurs with low possibility. So this solution may not hurt the performance a lot.

BusyJay:
How about checking `commit` field instead?

LintianShi:
I find weird that `commit` field in RaftLocalState.hard_state is not updated in `fn apply_snapshot`. Is it expected?

BusyJay:
It should be updated in the end. You can verify it with a test case.

LintianShi:
Yes, `commit` is updated out of `fn apply_snapshot`. It is updated in `fn handle_raft_ready`.

LintianShi:
As the `commit` of raft state always increases monotonically, we can use `commit` to check snapshot raft state in kvdb with raft state in raftdb. 
[...]

PR #13088, opened 2022-07-21
raftstore: fix checking for snapshot last index

LintianShi:
Signed-off-by: LintianShi <lintian.shi@pingcap.com>

Issue Number: Close https://github.com/tikv/tikv/issues/12618

What's Changed:

```commit-message
using commit instead of last_index to check gap between existing raft logs and snapshot when recovering from applying state.
```

### Related changes

### Check List 

Tests 

- Integration test

Side effects

- None

### Release note 

```release-note
None
```


BusyJay:
Any test case?

LintianShi:
I've tested `test_snapshot_recover_from_raft_write_failure` in test_snap.rs. I'm trying to create a special test case for this condition.

tests/failpoints/cases/test_snap.rs
```diff
@@ -785,3 +785,109 @@ fn test_snapshot_recover_from_raft_write_failure() {
...
+            .msg_type(MessageType::MsgAppend)
+            .direction(Direction::Recv),
+    ));
+    cluster.add_send_filter(CloneFilterFactory(
+        RegionPacketFilter::new(1, 3)
+            .msg_type(MessageType::MsgAppend)
+            .direction(Direction::Recv),
+    ));
+
+    // Reer 1 appends entries which is never committed.
```
5kbpers, line 824:
```suggestion
    // Peer 1 appends entries which is never committed.
```

5kbpers (approved):
rest LGTM

tonyxuqqi (approved):

components/raftstore/src/store/peer_storage.rs
```diff
@@ -464,13 +464,13 @@ pub fn recover_from_applying_state<EK: KvEngine, ER: RaftEngine>(
...
-    // if we recv append log when applying snapshot, last_index in raft_local_state will
-    // larger than snapshot_index. since raft_local_state is written to raft engine, and
-    // raft write_batch is written after kv write_batch, raft_local_state may wrong if
+    // since raft_local_state is written to raft engine, and
+    // raft write_batch is written after kv write_batch. raft_local_state may wrong if
     // restart happen between the two write. so we copy raft_local_state to kv engine
-    // (snapshot_raft_state), and set snapshot_raft_state.last_index = snapshot_index.
-    // after restart, we need check last_index.
-    if last_index(&snapshot_raft_state) > last_index(&raft_state) {
+    // (snapshot_raft_state), and set snapshot_raft_state.hard_state.commit = snapshot_index.
```
BusyJay, line 470: The comment is wrong, we don't set `hard_state.commit` explicitly.

tests/failpoints/cases/test_snap.rs
```diff
@@ -785,3 +785,109 @@ fn test_snapshot_recover_from_raft_write_failure() {
...
+    cluster.must_transfer_leader(1, new_peer(1, 1));
+
+    cluster.must_put(b"k1", b"v1");
+    for i in 1..4 {
+        must_get_equal(&cluster.get_engine(i), b"k1", b"v1");
+    }
+
+    // Guarantee that peer 2 and 3 won't receive any entries,
+    // so these entries cannot be committed.
+    cluster.add_send_filter(CloneFilterFactory(
```
BusyJay, line 813: Why not just
```
RegionPacketFilter::new(1, 1)
            .msg_type(MessageType::MsgAppend)
            .direction(Direction::Send)
```

components/raftstore/src/store/peer_storage.rs
LintianShi: In `save_snapshot_raft_state_to`, we set commit of hard_state in snapshot_raft_state to snapshot_index. And then we persist snapshot_raft_state into kvdb.
```
let mut snapshot_raft_state = self.raft_state.clone();
        snapshot_raft_state
            .mut_hard_state()
            .set_commit(snapshot_index);
```

BusyJay: Got it.

BusyJay:
The test case can pass without the fix on my local machine.

LintianShi:
In my dev machine, it panicked at line 878:` cluster.run_node(1).unwrap();`, when peer 1 recovers from crash.
```
running 1 test
test cases::test_snap::test_snapshot_recover_from_raft_write_failure_with_uncommitted_log ... FAILED

failures:

---- cases::test_snap::test_snapshot_recover_from_raft_write_failure_with_uncommitted_log stdout ----
thread 'cases::test_snap::test_snapshot_recover_from_raft_write_failure_with_uncommitted_log' panicked at 'called `Result::unwrap()` 
on an `Err` value: RaftServer(Other("[components/raftstore/src/store/peer_storage.rs:774]: [region 1] 1 validate state fail: 
Other(\"[components/raftstore/src/store/peer_storage.rs:612]: applied index > max(commit index, recorded commit index), 
region 1, raft state hard_state { term: 6 vote: 1 commit: 7 } last_index: 857, apply state applied_index: 33 commit_index: 7 
commit_term: 6 truncated_state { index: 33 term: 8 }\")"))', tests/failpoints/cases/test_snap.rs:878:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

tests/failpoints/cases/test_snap.rs
```diff
@@ -785,3 +785,109 @@ fn test_snapshot_recover_from_raft_write_failure() {
         cluster.must_put(format!("k1{}", i).as_bytes(), b"v1");
     }
 }
+
+#[test]
+fn test_snapshot_recover_from_raft_write_failure_with_uncommitted_log() {
```
BusyJay, line 790: Please add comment to explain the case.

tests/failpoints/cases/test_snap.rs
```diff
@@ -785,3 +785,106 @@ fn test_snapshot_recover_from_raft_write_failure() {
         cluster.must_put(format!("k1{}", i).as_bytes(), b"v1");
     }
 }
+
+/// Redo snapshot apply after restart when kvdb state is updated but raftdb state is not.
+/// In this test case, the peer has an extra long uncommitted log when applying snasphot.
```
BusyJay, line 790:
```suggestion
/// Test whether applying snapshot is resumed properly when last_index before applying snapshot
/// is larger than the snapshot index and applying is aborted between kv write and raft write.
```

BusyJay (approved):

PR #13091, opened 2022-07-21
*: solve the problem that test_stale_read_while_applying_snapshot is unstable

### LINKED ISSUE #13057: test_stale_read_while_applying_snapshot always fails in local environment

SpadeA-Tang:

Issue Number: Close #13057

What's Changed:

The test `test_stale_read_while_applying_snapshot` is fairly unstable. Althought CI can pass sometimes, In my local environment (ubuntu 20 and mac os), it often fails. This PR solves it.

### Failure reason:
The reason that it fails is that the config variables of `level0_file_num_compaction_trigger` and 
`level0_slowdown_writes_trigger` are both 4 for write cf of the kv db (Default config set None for `level0_slowdown_writes_trigger` and 4 for `level0_slowdown_writes_trigger`. None means `level0_slowdown_writes_trigger` will be set to 0 and then be set to 4 as shown below). 
image

Due to snapshot appy, the number of level0 files of write cf can reach at most to 3 in this test which trigers flow control as 3 > 4/2 (the code shown below) and skip `handle_apply`.  
image

What makes it even worse is that files number 3 will not triger compaction. So it will never call `handle_apply`. It means `safe_ts` will be 0 forever which makes the test fail (`safe_ts` will be 0 when applying snapshot).

So, `level0_slowdown_writes_trigger` / 2 shold be larger or equal to `level0_file_num_compaction_trigger` - 1.

### Solution

To be simple, set an appropriate value for `level0_slowdown_writes_trigger`.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -375,6 +375,13 @@ macro_rules! cf_config {
                     return Err("format-version larger than 5 is unsupported".into());
                 }
                 self.titan.validate()?;
+                if self.level0_slowdown_writes_trigger.unwrap() / 2 < self.level0_file_num_compaction_trigger - 1 {
+                    return Err(format!(
+                        "level0-slowdown-writes-trigger {} be at least 2 times of (level0-file-num-compaction-trigger {} - 1)",
+                        self.level0_slowdown_writes_trigger.unwrap(),
+                        self.level0_file_num_compaction_trigger,
+                    ).into());
+                }
                 Ok(())
             }
         }
@@ -635,7 +642,7 @@ impl Default for DefaultCfConfig {
             max_bytes_for_level_base: ReadableSize::mb(512),
             target_file_size_base: ReadableSize::mb(8),
             level0_file_num_compaction_trigger: 4,
-            level0_slowdown_writes_trigger: None,
+            level0_slowdown_writes_trigger: Some(20),
             level0_stop_writes_trigger: None,
             max_compaction_bytes: ReadableSize::gb(2),
             compaction_pri: CompactionPriority::MinOverlappingRatio,
@@ -747,7 +754,7 @@ impl Default for WriteCfConfig {
             max_bytes_for_level_base: ReadableSize::mb(512),
             target_file_size_base: ReadableSize::mb(8),
             level0_file_num_compaction_trigger: 4,
-            level0_slowdown_writes_trigger: None,
+            level0_slowdown_writes_trigger: Some(20),
             level0_stop_writes_trigger: None,
             max_compaction_bytes: ReadableSize::gb(2),
             compaction_pri: CompactionPriority::MinOverlappingRatio,
@@ -845,7 +852,7 @@ impl Default for LockCfConfig {
             max_bytes_for_level_base: ReadableSize::mb(128),
             target_file_size_base: ReadableSize::mb(8),
             level0_file_num_compaction_trigger: 1,
-            level0_slowdown_writes_trigger: None,
+            level0_slowdown_writes_trigger: Some(2),
             level0_stop_writes_trigger: None,
             max_compaction_bytes: ReadableSize::gb(2),
             compaction_pri: CompactionPriority::ByCompensatedSize,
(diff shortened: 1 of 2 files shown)


SpadeA-Tang:
> level0_slowdown_writes_trigger

Its default value is None which set it to 0. And the code in figure one set it to 4.

SpadeA-Tang:
> why `level0_slowdown_writes_trigger` is 4

Its default value is None which set it to 0. And the code in figure one set it to 4.

SpadeA-Tang:
> Will this problem also be triggered in a real cluster?

Maybe not, the default value of `level0_slowdown_writes_trigger` will be 20 while the `level0_file_num_compaction_trigger` is 4. But I guess if the users set inapropriate values for them, similar problems will happen.

tonyxuqqi:
> level0_slowdown_writes_trigger
I'm wondering if the l0_files_threshold is specified in the test. According the code below, the level0_slowdown_writes_trigger will be overwritten as l0_files_threshold if it's not specified. 

if let Some(v) = &mut $cf_opts.level0_slowdown_writes_trigger {
                    if $cfg.enable && *v > $cfg.l0_files_threshold as i32 {
                        warn!(
                            "{}.level0-slowdown-writes-trigger is too large. Setting it to \
                            storage.flow-control.l0-files-threshold ({})",
                            stringify!($cf_opts), $cfg.l0_files_threshold
                        );
                        *v = $cfg.l0_files_threshold as i32;
                    }
                } else {
                    $cf_opts.level0_slowdown_writes_trigger =
                        Some($cfg.l0_files_threshold as i32);
                }

BornChanger:
Is this problem a configuration problem then? If so, can we add sanity check in the configuration manager instead of changing the default value of some some parameters?

src/config.rs
```diff
@@ -635,7 +642,7 @@ impl Default for DefaultCfConfig {
             max_bytes_for_level_base: ReadableSize::mb(512),
             target_file_size_base: ReadableSize::mb(8),
             level0_file_num_compaction_trigger: 4,
-            level0_slowdown_writes_trigger: None,
+            level0_slowdown_writes_trigger: Some(20),
```
BornChanger, line 645: How are those default value numbers determined? Do we have such a document? I am afraid simply change the default value could bring some unexpected behavior when customers upgrades to newer version.

SpadeA-Tang:
> Is this issue caused by improper configuration then? If so, can we add sanity check in the configuration manager instead of changing the default value for some parameters?

The `Validate` will not be called by this mock server creation process. Maybe I should add this for mock too.

src/config.rs
SpadeA-Tang: The default value for this config parameter is 20 which is written in the TiKV config documentation (which is the default value set by rocksdb). I just set it temporarily, and I will configure a better way for it.

SpadeA-Tang:
> > level0_slowdown_writes_trigger
> 
> I'm wondering if the l0_files_threshold is specified in the test. According the code below, the level0_slowdown_writes_trigger will be overwritten as l0_files_threshold if it's not specified.
> 
> if let Some(v) = &mut $cf_opts.level0_slowdown_writes_trigger { if $cfg.enable && *v > $cfg.l0_files_threshold as i32 { warn!( "{}.level0-slowdown-writes-trigger is too large. Setting it to storage.flow-control.l0-files-threshold ({})", stringify!($cf_opts), $cfg.l0_files_threshold ); *v = $cfg.l0_files_threshold as i32; } } else { $cf_opts.level0_slowdown_writes_trigger = Some($cfg.l0_files_threshold as i32); }

These codes are in `Validate` method which will not be called in mock now.

tabokie (commented):
You should change the ingest condition to `> level0_file_num_compaction_trigger / 2 && >= level0_file_num_compaction_trigger`. Imposing a 2x restriction to user config doesn't make sense to me.

SpadeA-Tang:
> You should change the ingest condition to `> level0_file_num_compaction_trigger / 2 && >= level0_file_num_compaction_trigger`. Imposing a 2x restriction to user config doesn't make sense to me.

Good suggestion.

tabokie (approved):

youjiali1995 (approved):

Issue #12534, opened 2022-05-16
log-backup: add metrics for log backup


YuJuncen:
## Development Task
As title

PR #13105, opened 2022-07-22
log-backup: unify namespace for log backup metrics

YuJuncen:
Signed-off-by: Yu Juncen <yujuncen@pingcap.com>

Issue Number: Ref #12534 

What's Changed:
We have renamed the feature `backup stream` into `log backup`, however some of metrics keep the old name.
This PR has unified all the metrics of `backup-stream` with the namespace `log_bakcup`.

```commit-message
Renamed metrics in `backup-stream`:
tikv_stream_(.*) => tikv_log_backup_$1
tikv_pending_initial_scan => tikv_log_backup_initial_scan
```

### Related changes

- PR to update `tidb`:
(TBD)

### Check List 

Tests 

- No code

### Release note 

```release-note
All log backup related series has the prefix `tikv_log_backup` now.
```

### DIFF
--- a/components/backup-stream/src/metrics.rs
+++ b/components/backup-stream/src/metrics.rs
@@ -44,31 +44,31 @@ lazy_static! {
     )
     .unwrap();
     pub static ref HANDLE_EVENT_DURATION_HISTOGRAM: HistogramVec = register_histogram_vec!(
-        "tikv_stream_event_handle_duration_sec",
+        "tikv_log_backup_event_handle_duration_sec",
         "The duration of handling an cmd batch.",
         &["stage"],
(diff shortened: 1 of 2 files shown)


overvenus (commented):
It will be a breaking change for older grafana dashboard. Is it ok?

YuJuncen:
> It will be a breaking change for older grafana dashboard. Is it ok?

That would be OK, because the old grafana dashboard only has been released in the nightly build.

YuJuncen:
(In fact, once the 6.2.0 get released with old metric names, things would get too complex and probably we would have no chance to unify those series...)

overvenus (approved):

PR #13115, opened 2022-07-23
raftstore: extract EntryStorage

BusyJay:
Issue Number: ref #12842 

What's Changed:

```commit-message
This PR extract part of `PeerStorage` as `EntryStorage`, which only
serves entry access. It will be reused by raftstorev2.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


BusyJay:
The diff may not be easy to review. Basically what it does is copy paste functions from `PeerStorage` to `EntryStorage`.

components/raftstore/src/store/peer_storage.rs
```diff
@@ -417,601 +417,238 @@ impl Drop for EntryCache {
...
-        last_first_index: u64,
-    },
-    NoIOTask,
-}
-
-pub fn recover_from_applying_state<EK: KvEngine, ER: RaftEngine>(
-    engines: &Engines<EK, ER>,
-    raft_wb: &mut ER::LogBatch,
+/// A subset of `PeerStorage` that focus on accessing log entries.
+pub struct EntryStorage<ER> {
```
Connor1996, line 421: can we move it into a new file?

Connor1996 (approved):
LGTM

components/raftstore/src/store/peer_storage.rs
```diff
@@ -642,72 +325,31 @@ where
 
     peer_id: u64,
     region: metapb::Region,
-    raft_state: RaftLocalState,
-    apply_state: RaftApplyState,
     applied_index_term: u64,
```
Connor1996, line 328: move into `entry_storage`?

components/raftstore/src/store/entry_storage.rs
```diff
@@ -0,0 +1,1414 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
tonyxuqqi, line 1: Can this file be moved to raftstore/src/common/store instead so that it's straightforward that code is shared, as it's already moved anyway?
And I suppose there's no code change in this file comparing to the original peer_storage?

BusyJay: Yes, no logic is changed.

BusyJay: I will add a doc string to clarify it's shared as discussed.

tonyxuqqi (approved):

PR #13127, opened 2022-07-26
raftstorev2: fix cached tablet bug

### LINKED ISSUE #12842: Support physical isolation between regions

tabokie:

Issue Number: Ref #12842

What's Changed:

Some cleaning-ups while reading the code.
- Fix the issue that `CachedTablet::latest()` always acquires lock.
- Initialize buffer capacity when creating a `ApplyPoller`.
- Remove util function `check_or_prepare_bootstrap_first_region` for more clarity.
- Minor improvements to function comments.

```commit-message
None
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore-v2/src/batch/apply.rs
+++ b/components/raftstore-v2/src/batch/apply.rs
@@ -50,15 +50,17 @@ pub struct ApplyPoller {
 
 impl ApplyPoller {
     pub fn new(apply_ctx: ApplyContext, cfg_tracker: Tracker<Config>) -> ApplyPoller {
-        ApplyPoller {
+        let mut poller = ApplyPoller {
             apply_task_buf: Vec::new(),
             pending_latency_inspect: Vec::new(),
             apply_ctx,
             cfg_tracker,
-        }
+        };
+        poller.apply_buf_capacity();
+        poller
     }
 
-    /// Updates the internal buffer to latest capacity.
+    /// Updates the internal buffer to match the latest configuration.
     fn apply_buf_capacity(&mut self) {
         let new_cap = self.messages_per_tick();
         tikv_util::set_vec_capacity(&mut self.apply_task_buf, new_cap);
--- a/components/raftstore-v2/src/batch/apply.rs
+++ b/components/raftstore-v2/src/batch/apply.rs
@@ -50,14 +50,12 @@ pub struct ApplyPoller {
 
 impl ApplyPoller {
     pub fn new(apply_ctx: ApplyContext, cfg_tracker: Tracker<Config>) -> ApplyPoller {
-        let mut poller = ApplyPoller {
+        ApplyPoller {
             apply_task_buf: Vec::new(),
             pending_latency_inspect: Vec::new(),
             apply_ctx,
             cfg_tracker,
-        };
-        poller.apply_buf_capacity();
-        poller
+        }
     }
 
     /// Updates the internal buffer to match the latest configuration.
--- a/components/raftstore-v2/src/batch/apply.rs
+++ b/components/raftstore-v2/src/batch/apply.rs
@@ -50,7 +50,7 @@ pub struct ApplyPoller {
 
 impl ApplyPoller {
     pub fn new(apply_ctx: ApplyContext, cfg_tracker: Tracker<Config>) -> ApplyPoller {
-        ApplyPoller {
+        Self {
             apply_task_buf: Vec::new(),
             pending_latency_inspect: Vec::new(),
             apply_ctx,
--- a/components/raftstore-v2/src/batch/store.rs
+++ b/components/raftstore-v2/src/batch/store.rs
@@ -64,7 +64,7 @@ struct StorePoller<EK: KvEngine, T> {
 
 impl<EK: KvEngine, T> StorePoller<EK, T> {
     pub fn new(poll_ctx: StoreContext<T>, cfg_tracker: Tracker<Config>) -> Self {
-        StorePoller {
+        Self {
             store_msg_buf: Vec::new(),
             peer_msg_buf: Vec::new(),
             poll_ctx,
--- a/components/raftstore-v2/src/batch/store.rs
+++ b/components/raftstore-v2/src/batch/store.rs
@@ -64,16 +64,14 @@ struct StorePoller<EK: KvEngine, T> {
 
 impl<EK: KvEngine, T> StorePoller<EK, T> {
     pub fn new(poll_ctx: StoreContext<T>, cfg_tracker: Tracker<Config>) -> Self {
-        let mut poller = Self {
+        StorePoller {
             store_msg_buf: Vec::new(),
             peer_msg_buf: Vec::new(),
             poll_ctx,
             cfg_tracker,
             last_flush_time: TiInstant::now(),
             need_flush_events: false,
-        };
-        poller.apply_buf_capacity();
-        poller
+        }
     }
 
     /// Updates the internal buffer to match the latest configuration.
--- a/components/raftstore-v2/src/batch/store.rs
+++ b/components/raftstore-v2/src/batch/store.rs
@@ -26,7 +26,7 @@ use crate::{
     Error, PeerMsg, PeerTick, Result, StoreMsg,
 };
 
-/// A per thread context used for handling raft messages.
+/// A per-thread context used for handling raft messages.
 pub struct StoreContext<T> {
     /// A logger without any KV. It's clean for creating new PeerFSM.
     pub logger: Logger,
@@ -76,7 +76,7 @@ impl<EK: KvEngine, T> StorePoller<EK, T> {
         poller
     }
 
-    /// Updates the internal buffer to latest capacity.
+    /// Updates the internal buffer to match the latest configuration.
     fn apply_buf_capacity(&mut self) {
         let new_cap = self.messages_per_tick();
         tikv_util::set_vec_capacity(&mut self.store_msg_buf, new_cap);
@@ -119,6 +119,7 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport + 'static> PollHandler<PeerFsm<E
     }
 
     fn handle_control(&mut self, store: &mut StoreFsm) -> Option<usize> {
+        debug_assert!(self.store_msg_buf.is_empty());
         let received_cnt = store.recv(&mut self.store_msg_buf);
         let expected_msg_count = if received_cnt == self.messages_per_tick() {
             None
@@ -134,6 +135,7 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport + 'static> PollHandler<PeerFsm<E
         &mut self,
         peer: &mut impl DerefMut<Target = PeerFsm<EK, ER>>,
     ) -> HandleResult {
+        debug_assert!(self.peer_msg_buf.is_empty());
         let received_cnt = peer.recv(&mut self.peer_msg_buf);
         let handle_result = if received_cnt == self.messages_per_tick() {
             HandleResult::KeepProcessing
@@ -203,7 +205,7 @@ impl<EK: KvEngine, ER: RaftEngine, T> StorePollerBuilder<EK, ER, T> {
         }
     }
 
-    /// Init all the existing raft machine and cleanup stale tablets.
+    /// Initializes all the existing raft machines and cleanup stale tablets.
     fn init(&self) -> Result<HashMap<u64, SenderFsmPair<EK, ER>>> {
         let mut regions = HashMap::default();
         let cfg = self.cfg.value();
@@ -328,7 +330,7 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
 
 pub type StoreRouter<EK, ER> = BatchRouter<PeerFsm<EK, ER>, StoreFsm>;
 
-/// Create the batch system for polling raft activities.
+/// Creates the batch system for polling raft activities.
 pub fn create_store_batch_system<EK, ER>(
     cfg: &Config,
     store: Store,
--- a/components/raftstore-v2/src/bootstrap.rs
+++ b/components/raftstore-v2/src/bootstrap.rs
@@ -33,8 +33,8 @@ pub struct Bootstrap<'a, ER: RaftEngine> {
     logger: Logger,
 }
 
-// Although all methods won't change internal state, but they still receive `&mut self` as it's
-// not thread safe to bootstrap concurrently.
+// Although all methods won't change internal state, but they still receive
+// `&mut self` as it's not thread safe to bootstrap concurrently.
 impl<'a, ER: RaftEngine> Bootstrap<'a, ER> {
     pub fn new(
         engine: &'a ER,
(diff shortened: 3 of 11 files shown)


components/raftstore-v2/src/bootstrap.rs
```diff
@@ -71,7 +83,9 @@ impl<'a, ER: RaftEngine> Bootstrap<'a, ER> {
         Ok(Some(ident.get_store_id()))
     }
 
-    fn inner_bootstrap_store(&mut self) -> Result<u64> {
+    /// Allocates a store ID from PD and bootstraps the engine with it. Returns
+    /// the store ID if success.
+    fn bootstrap_store_imp(&mut self) -> Result<u64> {
```
Connor1996, line 88: seems unnecessary to extract a impl

components/raftstore-v2/src/bootstrap.rs
```diff
@@ -33,8 +33,8 @@ pub struct Bootstrap<'a, ER: RaftEngine> {
     logger: Logger,
 }
 
-// Although all methods won't change internal state, but they still receive `&mut self` as it's
-// not thread safe to bootstrap concurrently.
+// Although all methods won't change internal state, but they still receive
```
Connor1996, line 36: Maybe sketch the process of bootstrap here?

components/raftstore-v2/src/bootstrap.rs
tabokie: It's renamed from `inner_bootstrap_store` to `bootstrap_store_imp`.

Connor1996: I know, but it can be merged into `bootstrap_store` directly

components/raftstore-v2/src/bootstrap.rs
tabokie: Honestly I don't know what is "first region", could you help clarify?

Connor1996: the initial region whose start key and end key are both empty to cover the whole keyspace range. Other regions would be split from this region after performing some writes

components/raftstore-v2/src/bootstrap.rs
```diff
@@ -52,19 +57,33 @@ impl<'a, ER: RaftEngine> Bootstrap<'a, ER> {
...
-        };
-
-        Ok(store_id)
+        if let Some(id) = self.check_engine_store_id()? {
+            return Ok(id);
+        }
+        let id = self.pd_client.alloc_id()?;
+        debug!(self.logger, "alloc store id"; "store_id" => id);
+        let mut ident = StoreIdent::default();
+        if !self.engine.is_empty()? {
```
Connor1996, line 71: move to L68

components/raftstore-v2/src/bootstrap.rs
```diff
@@ -50,9 +55,35 @@ impl<'a, ER: RaftEngine> Bootstrap<'a, ER> {
...
+        self.engine.sync()?;
+        fail_point!("node_after_bootstrap_store", |_| Err(box_err!(
+            "injected error: node_after_bootstrap_store"
+        )));
+        Ok(id)
+    }
+
+    /// Gets and validates the store ID from engine if it's already
+    /// bootstrapped.
+    fn check_engine_store_id(&mut self) -> Result<Option<u64>> {
```
Connor1996, line 86: how about `check_store_id_in_engine`

components/raftstore-v2/src/bootstrap.rs
```diff
@@ -71,33 +102,115 @@ impl<'a, ER: RaftEngine> Bootstrap<'a, ER> {
...
+                        }
+                        Err(e) => {
+                            warn!(self.logger, "get the first region failed"; "err" => ?e);
+                        }
+                    }
+                }
+                Err(e) => {
+                    error!(
+                        self.logger,
+                        "bootstrap cluster";
```
Connor1996, line 187: failed to bootstrap cluster

Connor1996 (commented):
rest LGTM

components/raftstore-v2/src/tablet.rs
```diff
@@ -76,7 +76,14 @@ mod tests {
         // Setting tablet will refresh cache immediately.
         cached_tablet.set(2);
         assert_eq!(cached_tablet.cache().cloned(), Some(2));
-        assert_eq!(cached_tablet.latest().cloned(), Some(2));
+        // Test `latest()` will use cache.
+        {
```
BusyJay, line 80: This is not the TiKV style.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -58,8 +58,8 @@ pub struct Storage<ER> {
 impl<ER: RaftEngine> Storage<ER> {
     /// Creates a new storage.
     ///
-    /// All metadata should be initialized before calling this method. If the region is destroyed
-    /// `None` will be returned.
+    /// All metadata should be initialized before calling this method. If the
```
BusyJay, line 61: Why the change?

components/raftstore-v2/src/bootstrap.rs
```diff
@@ -167,75 +255,4 @@ impl<'a, ER: RaftEngine> Bootstrap<'a, ER> {
         box_try!(self.engine.consume(&mut wb, true));
         Ok(())
     }
-
-    fn inner_bootstrap_first_region(
```
BusyJay, line 171: Why move them around?

components/raftstore-v2/src/batch/apply.rs
```diff
@@ -50,15 +50,17 @@ pub struct ApplyPoller {
...
     pub fn new(apply_ctx: ApplyContext, cfg_tracker: Tracker<Config>) -> ApplyPoller {
-        ApplyPoller {
+        let mut poller = ApplyPoller {
             apply_task_buf: Vec::new(),
             pending_latency_inspect: Vec::new(),
             apply_ctx,
             cfg_tracker,
-        }
+        };
+        poller.apply_buf_capacity();
```
BusyJay, line 59: I don't think it's necessary, it's always applied in `begin`.

tabokie: Okay, I just copied it from store poller.

components/raftstore-v2/src/raft/storage.rs
tabokie: 80c limit.

BusyJay: In Rust, it's 99.

components/raftstore-v2/src/tablet.rs
tabokie: What is not? Be a little more specific?

components/raftstore-v2/src/bootstrap.rs
tabokie: Which part are you referring to? This function is removed because @Connor1996 suggested it and I agree it looks better. Or are you referring to the ordering between public function and private sub-routines? My habit is to read public functions first, private methods later (just like C++ code). If you have a better reason suggesting the opposite I'm willing to change it back.

components/raftstore-v2/src/tablet.rs
BusyJay: Wrapping test steps in a pair of brackets. We usually use a blank line.

components/raftstore-v2/src/raft/storage.rs
tabokie: I don't think so. Rust fmt default configures it as 80: https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#comment_width, and rust source code uses 80 too.

components/raftstore-v2/src/tablet.rs
tabokie: Okay. The intention really is to scope the unsafe modifications of `data`.

components/raftstore-v2/src/bootstrap.rs
BusyJay: I refer to the ordering of functions. Everyone has their habits. To me, related functions should be put together no matter what access level they are. However, I don't think it's a good idea to reorder them again after they are committed to the code base unless for common sense.

components/raftstore-v2/src/raft/storage.rs
BusyJay: Got it, never expect length of comment line is different from source code.

BusyJay: Turns out I'm not the only one has the confusion: https://github.com/rust-lang/rustfmt/issues/3349#issuecomment-622522913

tabokie: Having looked closer, apparently rust source code uses different comment width as well.. Personally I think 80c comment is perfect, whilst source code is easier to digest because they are structured.

components/raftstore-v2/src/bootstrap.rs
Connor1996: `inner_bootstrap_first_region` is extracted without any sense, it can be just placed into `bootstrap_first_region`. Moreover, `inner_bootstrap_first_region` disturbs the ordering for reading code, you need to read back and forth. Now, it's clean and neat.

tabokie: So we can all agree related functions should be grouped together. As for the ordering, I believe it's favorable to arrange code as we read them. In the old days this is mostly impossible because of name resolution constraints, but with Rust I think we should arrange things naturally whenever we can (which is top down, reading flow, in contrary to writing flow).

BusyJay: I think reading several small functions with meaningful names is a lot easier than reading a long functions with a lot of comments. But this function is not too long to be split, so I'm OK with the change.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -64,19 +64,17 @@ struct StorePoller<EK: KvEngine, T> {
 
 impl<EK: KvEngine, T> StorePoller<EK, T> {
     pub fn new(poll_ctx: StoreContext<T>, cfg_tracker: Tracker<Config>) -> Self {
-        let mut poller = Self {
+        StorePoller {
```
BusyJay, line 67: I use `Self` on purpose. I have found it very convenient when the name is changed several times during development. Though I'm OK with either way.

components/raftstore-v2/src/bootstrap.rs
BusyJay: > I think we should arrange things naturally whenever we can

I don't think so as there is no consensus on what's natural. For example, for peer struct, what's the natural way to order all those the functions?

And when reading using vscode/vim/ides, there is no much difference between top/down or jumping using builtin support tools.

I consider reordering functions evil as it's hard to review. If the suggestion is proposed the first time PR is sent, it's acceptable.

Connor1996 (approved):
LGTM

components/raftstore-v2/src/bootstrap.rs
tabokie: > If the suggestion is proposed the first time PR is sent, it's acceptable.

Clearly we need some basic style guides here, at least some sense of what is considered more readable. I can relate that ordering is not very important. Especially when the member functions keep growing and get further intertwined.

> I think reading several small functions with meaningful names is a lot easier than reading a long functions with a lot of comments.

However, I want to fight a little on this (even though it's not the point of this PR). Splitting a function creates cognitive load for the reader. When reading the code, one has to keep the control flow context inside the (human) memory, the same way that a computer would keep stack variables around. So I want to make sure we can all agree that function calling depth should be kept low if possible.

Furthermore for raftstore code, most functions mutate some internal states (let it be engine data, memory states, network states). Creating sub-routines make the state mutation hard to track.

Both of these two points can't be addressed with IDE. And since we are rebuilding the raftstore code, I think it makes sense to push for this and make the code easier to read than v1.

For reference, I love this post and some related HN discussions: https://hn.algolia.com/?dateRange=all&page=0&prefix=true&query=inlined%20code&sort=byPopularity&type=story

components/raftstore-v2/src/raft/storage.rs
Connor1996: Let's limit the comment width automatically by rustfmt, then it's no need to discuss it again and again. https://github.com/tikv/tikv/pull/13139 some comments should be adjusted after auto wrap, will work on fixing it once we get an agreement on the comment width.

components/raftstore-v2/src/bootstrap.rs
BusyJay: > I can relate that ordering is not very important.

Yes, to me it's not important. If the comment is sent back in the time, I will happily change it. The most important thing is:
1. Don't randomly change the order of committed code as it's hard to read and it causes trouble of future cherry-picks of potential bugs;
2. Functions with related purpose should be put together to ease context switch. That's the purpose why v2 has `operations` module to split code by operations.

> Splitting a function creates cognitive load for the reader. 

Note the prerequisite is **small functions with meaningful names**. As long as the function name makes sense, all we need to look to is whether the implementation represents its name. So there will be no such requirement to keep tracking variables. For example:
```
if store.needs_continue_bootstrapping() {
    store.bootstrap();
}
```
One can easily know what the flow is by name. Though I'm not suggesting the original naming is precise and correct, just suggesting reusing codes is not the only purpose of extracting functions.

BusyJay (approved):

components/raftstore-v2/src/bootstrap.rs
tabokie: Yes, it's a good point. Though many would argue that using function as a way to "group code" isn't really better than just writing some comments. Because to understand `needs_continue_bootstrapping`, you still need to do a context switch. It's against human's nature to put the questions aside and take the function name as its implementation.

In our situation, I am advising strongly on calling depth and state mutations. Good function names can only do as much. And take a second thought before doing a refactor. Veteran devs such as yourself sometimes just get too comfortable with sub-routines and treating them like "common sense", meanwhile for newcomer it would take a long while to get around.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -64,19 +64,17 @@ struct StorePoller<EK: KvEngine, T> {
...
-        let mut poller = Self {
+        Self {
             store_msg_buf: Vec::new(),
             peer_msg_buf: Vec::new(),
             poll_ctx,
             cfg_tracker,
             last_flush_time: TiInstant::now(),
             need_flush_events: false,
-        };
-        poller.apply_buf_capacity();
```
BusyJay, line 75: Receiving messages depends on the capacity. Without initializing the capacity, it will never receive any messages.

tabokie: I deleted it because of your https://github.com/tikv/tikv/pull/13127#discussion_r930583660. But yes, apparently I wasn't paying attention either.

BusyJay: I have another PR fixing it by passing an argument explicitly. Using capacity is too obscure.

PR #13132, opened 2022-07-27
*: remove engine_rocks raw_util

BusyJay:

Issue Number: Ref #13058 

What's Changed:

```commit-message
raw_util is duplicated with util. By using util, this PR also makes `DB`
type only in use in engine_rocks, which is helpful for adapting tirocks
later.

In addition, this PR fixes an unsound transforms between `Arc<DB>` and
`RocksEngine`.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/backup/src/endpoint.rs
```diff
@@ -505,7 +505,7 @@ impl BackupRange {
     async fn backup_raw_kv_to_file<E: Engine>(
         &self,
         engine: E,
-        db: Arc<DB>,
+        db: RocksEngine,
```
Connor1996, line 508: why it's generic on Engine, but still need to pass a rocksdb?

BusyJay: /cc @3pointer

Connor1996 (approved):
rest LGTM

components/engine_rocks/src/cf_options.rs
```diff
@@ -27,28 +29,40 @@ impl CFOptionsExt for RocksEngine {
     }
 }
 
-#[derive(Clone)]
-pub struct RocksColumnFamilyOptions(RawCFOptions);
+#[derive(Default, Clone)]
+pub struct RocksCfOptions(RawCfOptions);
```
tabokie, line 33: Bit weird to have both `RocksCfOptions` and `CFOptionsExt`/`RocksDBOptions`.

tabokie (approved):

components/engine_rocks/src/cf_options.rs
BusyJay: I will rename all those struct in a separate PR.

components/backup/src/endpoint.rs
BusyJay: I merge it now, and solve issues further if there is any.

3pointer: We need build `RocksSstWriterBuilder` from `RocksEngine` during backup. but it seems we cannot get `RocksEngine` from generic Engine 🤔

Connor1996: maybe you don't need to call `set_db` for `RocksSstWriterBuilder`

3pointer: if we don't call  `set_db`,  it seems we cannot get `cf_options` of current rocksdb. this makes backup sst different with normal sst files.

PR #13138, opened 2022-07-27
txn: add lock wait duration and read exeution duration track for pessimistic lock

cfzjywxk:

Issue Number: Ref #12362 

What's Changed:

```commit-message
The transaction command execution contains mainly two parts the read phase and the write phase, some 
commands like `acquire_pessimistic_lock` may have a wait and wakeup phase.
The tracker utility enables us to track the whole path information for a given command, maybe we 
could track more information about:
- The read phase for each transaction command, sometimes slow-read could slow the whole process.
- The in-queue lock wait and wake-up details for pessimistic lock request processing. Especially 
after the optimization of the lock wait/wakeup model which is developed by @MyonKeminta now, 
there would be not much pessimistic retry but the lock wait could become much longer. 

```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/tracker/src/lib.rs
+++ b/components/tracker/src/lib.rs
@@ -69,6 +69,11 @@ impl Tracker {
         detail.set_apply_write_wal_nanos(self.metrics.apply_wait_nanos);
         detail.set_apply_write_memtable_nanos(self.metrics.apply_write_memtable_nanos);
     }
+
+    pub fn write_time_detail(&self, detail: &mut pb::TimeDetail) {
+        detail.set_wait_wall_time_ms(self.metrics.wait_wall_time_ms);
+        detail.set_kv_read_wall_time_ms(self.metrics.kv_read_wall_time_ms);
+    }
 }
 
 #[derive(Debug, Default)]
@@ -148,4 +153,8 @@ pub struct RequestMetrics {
     pub apply_thread_wait_nanos: u64,
     pub apply_write_wal_nanos: u64,
     pub apply_write_memtable_nanos: u64,
+
+    // Some execution details about the `TimeDetail` fields in kv request responses.
+    pub wait_wall_time_ms: u64,
+    pub kv_read_wall_time_ms: u64,
 }
--- a/src/server/lock_manager/waiter_manager.rs
+++ b/src/server/lock_manager/waiter_manager.rs
@@ -26,6 +26,7 @@ use tikv_util::{
     worker::{FutureRunnable, FutureScheduler, Stopped},
 };
 use tokio::task::spawn_local;
+use tracker::GLOBAL_TRACKERS;
 
 use super::{config::Config, deadlock::Scheduler as DetectorScheduler, metrics::*};
 use crate::storage::{
@@ -180,6 +181,7 @@ pub(crate) struct Waiter {
     pub diag_ctx: DiagnosticContext,
     delay: Delay,
     _lifetime_timer: HistogramTimer,
+    enqueue_time: Instant,
 }
 
 impl Waiter {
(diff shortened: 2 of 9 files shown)


components/tracker/src/lib.rs
```diff
@@ -69,6 +69,11 @@ impl Tracker {
         detail.set_apply_write_wal_nanos(self.metrics.apply_wait_nanos);
         detail.set_apply_write_memtable_nanos(self.metrics.apply_write_memtable_nanos);
     }
+
+    pub fn write_time_detail(&self, detail: &mut pb::TimeDetail) {
+        detail.set_wait_wall_time_ms(self.metrics.wait_wall_time_ms);
+        detail.set_kv_read_wall_time_ms(self.metrics.kv_read_wall_time_ms);
```
sticnarf, line 75: Perhaps these durations can be broken down better. 

I'm going to further add these durations about the duration before `async_write`: https://github.com/pingcap/kvproto/pull/965

sticnarf (commented):
Although the fields in `TimeDetail` is not well-defined, it's ok to put them in `TimeDetail` for now.

components/tracker/src/lib.rs
cfzjywxk: Got it, I'm thinking about the dimension to split the durations, one way is to separate them by the `read` and `write` path just like the fast-tuning does.

sticnarf: I add the wait time in the waiter manager in https://github.com/pingcap/kvproto/pull/965 as part of `WriteDetail`. The newly added fields are specific to write commands, which might extend the original semantics of `WriteDetail` (which only includes the raftstore stuff).
Do you think it a good place to put the wait time of pessimistic locks?

src/server/lock_manager/waiter_manager.rs
```diff
@@ -222,6 +225,13 @@ impl Waiter {
     /// `Notify` consumes the `Waiter` to notify the corresponding transaction
     /// going on.
     fn notify(self) {
+        // Record the in queue wait time.
+        GLOBAL_TRACKERS.with_tracker(self.diag_ctx.tracker_token, |tracker| {
+            let notify_time = Instant::now();
+            tracker.metrics.wait_wall_time_ms = notify_time
```
you06, line 231: Just use `self.enqueue_time.elapsed()`.

components/tracker/src/lib.rs
cfzjywxk: How about splitting a command into `read` and `write` parts, the `read` part is the processing before async write which may include things like conflict check, existence check, value fetch, assertion checks, etc. All the durations in the async write are put into the `write` part. A possible way is:
- The `TimeDetail` could be used to represent the `read` part which coprocessor dose in the before.
- The `WriteDetail` could be used to represent the `write` part, and it includes raftstore metrics already.

Them the pessimistic lock wait duration could be put into the `TimeDetail` as it's in the `read` part.

cfzjywxk: @you06 @MyonKeminta 
Any suggestions about how to represent different durations of different stags in the [proto](https://github.com/pingcap/kvproto/pull/965)?

MyonKeminta: Personally I prefer considering things like pessimistic lock waiting and latch waiting as part of writing instead of reading. But since `TimeDetail` doesn't seem to intend to representing stuff related to reading only, I think I don't have any strong reason to prefer one to another.

you06: Looks `TimeDetail` is the general time of both read and write requests, and the lock related durations are more close to write requests, suppose there may be the write-push-down feature in the future TiKV, I prefer to place them in `WriteDetail`.

cfzjywxk: Another option is to keep the meaning of `WriteDetail` to represent raftstore metrics, and put other durations into `TimeDetail`, so the `WriteDetail` would not become too large or complex.

MyonKeminta:
It seems we need to take some extra efforts to adapt it if we are going to merge the pessimistic lock waiting behavior optimization to master? 🤔

PR #13139, opened 2022-07-27
*: limit comment width by rustfmt

Connor1996:

Issue Number: Close https://github.com/tikv/tikv/issues/13150

What's Changed:

```commit-message
limit comment width to 80 by rustfmt
```

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```


rustfmt.toml
```diff
@@ -1,6 +1,13 @@
 version = "Two"
 unstable_features = true
 
+comment_width = 100
```
Connor1996, line 4: @BusyJay @tabokie

Connor1996: maybe we can set it to 120 as the resolution of screen is large enough nowadays? `100` width is not handy to write in a single line sometimes.

Connor1996: for reference, risingwave use a width of 120 https://github.com/singularity-data/risingwave/blob/main/rustfmt.toml#L1

tabokie: What's the good of fitting it in one line? IMO the point is not the screen but the human reading experience.

E.g. I take a look at WSJ, NYT, Bloomberg. Their websites use line width ranging from 70 to 85. So I'll vote for 80 default here.

Connor1996: Consider the comment for code in a deep nest, the indentation may already occupy 30 width. Then the space is too narrow to write comments. Though I'm not insisting on writing in one line. 

I don't vote for 80, as the code width is 100. we'd make the comment width at least the same as code width.

tabokie: I think different width makes sense in two ways:
- It looks good. Down to the bottom it's because we read code in structured tokens, but read text as unstructured words. So we have a higher tolerance for code width, but lower for text.
- Long-form comments always fill the width, whereas code rarely hits the limit. When setting the comment width the same as code, you will get bulky comments that visually stand out.

cmd/tikv-ctl/src/main.rs
```diff
@@ -796,7 +797,9 @@ fn print_bad_ssts(data_dir: &str, manifest: Option<&str>, pd_client: RpcClient,
         println!("\nsst meta:");
         // The output may like this:
         // --------------- Column family "write"  (ID 2) --------------
-        // 63:132906243[3555338 .. 3555338]['7A311B40EFCC2CB4C5911ECF3937D728DED26AE53FA5E61BE04F23F2BE54EACC73' seq:3555338, type:1 .. '7A313030302E25CD5F57252E' seq:3555338, type:1] at level 0
+        // 63:132906243[3555338 ..
```
BusyJay, line 800: Bad format.

components/backup-stream/src/endpoint.rs
```diff
@@ -429,11 +430,13 @@ where
         // Let L be the instant some key locked, U be the instant it unlocked,
         // +---------*-------L-----------U--*-------------+
         //           ^   ^----(1)----^      ^ We get the snapshot for initial scanning at here.
-        //           +- If we issue refresh resolver at here, and the cmd batch (1) is the last cmd batch of the first observing.
-        //              ...the background initial scanning may keep running, and the lock would be sent to the scanning.
-        //              ...note that (1) is the last cmd batch of first observing, so the unlock event would never be sent to us.
-        //              ...then the lock would get an eternal life in the resolver :|
-        //                 (Before we refreshing the resolver for this region again)
+        //           +- If we issue refresh resolver at here, and the cmd batch (1) is the last cmd
```
BusyJay, line 433: Also bad format.

BusyJay (commented):
auto-wrap doesn't seem to perform well.

cmd/tikv-ctl/src/main.rs
Connor1996: yes, will fix it once we have a agreement on the comment width

components/backup-stream/src/endpoint.rs
Connor1996: should use `-` to indicate list item, will fix

rustfmt.toml
Connor1996: The second point hits me, okay let's stick to rust default comment width. What do you think @BusyJay

BusyJay: Note 80 doesn't count indentation, but 100 counts. So when using 80/100, it means:
1. The total length is within 100.
2. The length of comment itself is within 80.

I suggest to keep the default settings as most projects do.

Connor1996: where says `The length of comment itself is within 80.`

BusyJay: > Source lines which are entirely a comment should be limited to 80 characters in length (including comment sigils, but excluding indentation)

https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md

Connor1996: Oh, it gives a quite reasonable answer to why the default comment width is not the same as the max line width

Connor1996:
I walk through the changes one by one and fix all the bad formats. PTAL again @BusyJay @tabokie

tabokie (approved):
CI is failing.

BusyJay (approved):
It's impossible to review 500 files.

PR #13157, opened 2022-07-28
raftstore: refactor async write to allow write all states to raft engine

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
In v2, all states are moved to raft engine, so it doesn't need to write to kv db anymore.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore/src/store/peer_storage.rs
```diff
@@ -411,23 +299,18 @@ where
...
+        let raft_state = init_raft_state(&engines, region)?;
         let apply_state = init_apply_state(&engines, region)?;
-        if let Err(e) = validate_states(region.get_id(), &engines, &mut raft_state, &apply_state) {
-            return Err(box_err!("{} validate state fail: {:?}", tag, e));
-        }
-        let last_term = init_last_term(&engines, region, &raft_state, &apply_state)?;
-        let applied_term = init_applied_term(&engines, region, &apply_state)?;
+
         let entry_storage = EntryStorage::new(
             region.id,
```
Connor1996, line 306: no need to pass it now

components/raftstore/src/store/async_io/write.rs
```diff
@@ -81,6 +81,79 @@ where
     }
 }
 
+/// Extra writes besides raft engine.
+///
+/// For now, applying snapshot needs to persist some extra states. For v1,
+/// these states are written to KvEngine. For v2, they are written to raft
```
Connor1996, line 87: raft engine -> RaftEngine

components/raftstore/src/store/async_io/write.rs
```diff
@@ -189,16 +262,55 @@ where
     }
 }
 
+/// These states are set only in raftstore V2.
+#[derive(Default)]
+pub struct ExtraStates {
```
Connor1996, line 267: why not extend it and let v1 reuse it

BusyJay: Actually there is a TODO https://github.com/tikv/tikv/pull/13157/files#diff-2b07a6df127983bc0dac285d2ca79b6762ae50749697441f11420fafe7ebd21bR89. But it's not a goal of the PR and I don't want to add more complexity to V1.

BusyJay: Yes, but the complexity is restricted inside writer and and the old code path will still be the same. And after wards, it's free to mutate the v2 implementations and add more asynchronous implementations like async destroy.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -298,6 +445,20 @@ where
         for (region_id, state) in self.raft_states.drain() {
             self.raft_wb.put_raft_state(region_id, &state).unwrap();
         }
+        if let ExtraBatchWrite::V2(extra_states_map) = &mut self.extra_batch_write {
+            if !extra_states_map.is_empty() {
```
tonyxuqqi, line 449: is this line necessary?

tonyxuqqi (approved):

components/raftstore/src/store/async_io/write.rs
```diff
@@ -53,11 +52,11 @@ const RAFT_WB_SHRINK_SIZE: usize = 10 * 1024 * 1024;
 const RAFT_WB_DEFAULT_SIZE: usize = 256 * 1024;
 
 /// Notify the event to the specified region.
-pub trait Notifier: Clone + Send + 'static {
+pub trait PersistedNotifier: Clone + Send + 'static {
     fn notify_persisted(&self, region_id: u64, peer_id: u64, ready_number: u64);
 }
 
-impl<EK, ER> Notifier for RaftRouter<EK, ER>
+impl<EK, ER> PersistedNotifier for RaftRouter<EK, ER>
```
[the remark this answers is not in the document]
BusyJay: `Notifier` will be exported and the name is too generic. I agree `notify_persisted` can be renamed to `notify`.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -491,29 +651,37 @@ where
...
-            // TODO: Add perf context
-            self.batch.kv_wb.write_opt(&write_opts).unwrap_or_else(|e| {
-                panic!(
-                    "store {}: {} failed to write to kv engine: {:?}",
-                    self.store_id, self.tag, e
-                );
-            });
-            if self.batch.kv_wb.data_size() > KV_WB_SHRINK_SIZE {
-                self.batch.kv_wb = self.engines.kv.write_batch_with_cap(KV_WB_DEFAULT_SIZE);
+        if let ExtraBatchWrite::V1(kv_wb) = &mut self.batch.extra_batch_write {
```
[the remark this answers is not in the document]
BusyJay: v2 writes all things to raft_wb, which is done in `before_write_to_db`, there is  no `kv_wb` in v2.

components/raftstore/src/store/async_io/write.rs
```diff
@@ -249,6 +362,40 @@ where
             }
         }
 
+        match mem::replace(&mut task.extra_write, ExtraWrite::None) {
```
Connor1996, line 365: extract to the impl of `ExtraBatchWrite`?

Connor1996 (approved):
LGTM

PR #13159, opened 2022-07-29
storage: record and return latch wait and process time

sticnarf:

Issue Number: ref #12362

Depends on https://github.com/pingcap/kvproto/pull/965. There is still some controversy about how to organize the returned metrics in the protobuf. 

What's Changed:

```commit-message
This commit also records the time breakdown before async_write,
including the latch wait time and process and throttle time in the
scheduler.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```


components/tracker/src/lib.rs
```diff
@@ -147,3 +153,36 @@ pub struct RequestMetrics {
...
+            field,
+        }
+    }
+}
+
+impl<M> Drop for TrackGuard<M>
+where
+    M: FnMut(&mut RequestMetrics) -> &mut u64,
+{
+    fn drop(&mut self) {
```
OneSizeFitsQuorum, line 183: We can do this only for values that are not 0, reducing the lock acquisition overhead for metrics that are often 0, such as TOTAL_THROTTLE_DURATION

src/storage/txn/scheduler.rs
```diff
@@ -166,9 +168,15 @@ impl TaskContext {
     }
 
     fn on_schedule(&mut self) {
+        let elapsed = self.latch_timer.saturating_elapsed();
```
OneSizeFitsQuorum, line 171: can we use `TrackGuard` to make code style consistent?

src/storage/txn/scheduler.rs
```diff
@@ -820,12 +832,14 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
             + statistics.cf_statistics(CF_WRITE).flow_stats.read_bytes;
         sample.add_read_bytes(read_bytes);
         let quota_delay = quota_limiter.consume_sample(sample, true).await;
+        let mut total_throttle_duration =
+            TrackGuard::new(tracker, |m| &mut m.scheduler_throttle_nanos);
         if !quota_delay.is_zero() {
+            total_throttle_duration.value += quota_delay.as_nanos() as u64;
```
OneSizeFitsQuorum, line 838: I'm curious about where we really execute this delay? Can we confirm that the number here is the actual delay time? Is it possible that Runtime will call for a longer period of time? 

For example, in the flow_controller department below, the specific scheduling time can be calculated explicitly.

src/storage/txn/scheduler.rs
sticnarf: My thought is that `TrackGuard` is only used when it's hard to know when we will quit the current function. It may be simpler to get and mutate the tracker directly if it's possible.

components/tracker/src/lib.rs
```diff
@@ -64,7 +67,7 @@ impl Tracker {
         detail.set_apply_log_nanos(self.metrics.apply_time_nanos - self.metrics.apply_wait_nanos);
         detail.set_apply_mutex_lock_nanos(self.metrics.apply_mutex_lock_nanos);
         detail.set_apply_write_leader_wait_nanos(self.metrics.apply_thread_wait_nanos);
-        detail.set_apply_write_wal_nanos(self.metrics.apply_wait_nanos);
```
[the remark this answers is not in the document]
sticnarf: Yes, I did it...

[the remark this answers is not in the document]
OneSizeFitsQuorum: Got it)

src/storage/txn/scheduler.rs
OneSizeFitsQuorum: Got it~ I'm OK with both.

src/storage/txn/scheduler.rs
sticnarf: Good idea. I believe they are approximately the same. But it's absolutely more accurate to record it ourselves.

components/tracker/src/lib.rs
sticnarf: Done

OneSizeFitsQuorum (approved):
LGTM

Issue #13095, opened 2022-07-21
commit b558d0 causing 7%~14% performance degradation for sysbench oltp_update_index workload


dbsid:
### Steps to reproduce

compare sysbench oltp_update_index workload with https://github.com/tikv/tikv/commit/b558d0bffc554c4c0094a483c65d95a50a4141cf
 and https://github.com/tikv/tikv/commit/126da29086da78bad3171975d4269ba283d133bd

### What did you expect?
no performance degradation

### What did happened?
commit b558d0 cause 7% ~ 14% performance degradation.

![image](https://user-images.githubusercontent.com/2316425/180250567-14cc2748-69f0-4ba2-b3a2-9acda4c7ef75.png)

jebter:
can it be closed？

tabokie:
It doesn't affect 6.2 anymore, I removed the tag.

PR #13161, opened 2022-07-29
engine: revert write with multi wb to one wb

ethercflow:
Issue Number: Ref https://github.com/tikv/tikv/issues/13095

What's Changed:

```
engine: revert write with multi wb to one wb
```

### Related changes

- None

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - None

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_rocks/src/engine.rs
+++ b/components/engine_rocks/src/engine.rs
@@ -32,7 +32,7 @@ pub struct RocksEngine {
 impl RocksEngine {
     pub fn from_db(db: Arc<DB>) -> Self {
         RocksEngine {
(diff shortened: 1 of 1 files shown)


components/engine_rocks/src/engine.rs
```diff
@@ -34,7 +34,7 @@ impl RocksEngine {
         RocksEngine {
             db: db.clone(),
```
tabokie, line 35:
```suggestion
            db,
```

Connor1996 (approved):

PR #13164, opened 2022-07-29
cdc: load old value from data CF if necessary

### LINKED ISSUE #10256: CDC values in the default CF should also be read in txn layer when old value is enabled
## Bug Report

Values in the default CF should also be read in txn layer when old value is enabled. Otherwise, CDC endpoint CPU might be full when TiDB batch inserts rows that row size is larger than 255 bytes.

Short values (<=255 bytes) are inline in the Lock struct during the prewrite stage.
https://github.com/tikv/tikv/blob/cd905a208293c2b0c1fca9aa2756d5c26392b38c/src/storage/txn/actions/prewrite.rs#L349

And they are inline in the Write struct during the commit stage and stored in Write CF.
https://github.com/tikv/tikv/blob/cd905a208293c2b0c1fca9aa2756d5c26392b38c/src/storage/txn/actions/commit.rs#L86

For now, reader only reads old values that are inlined in the Write struct.
https://github.com/tikv/tikv/blob/cd905a208293c2b0c1fca9aa2756d5c26392b38c/src/storage/mvcc/reader/reader.rs#L487-L489

If the value is larger than 255 bytes, it is restored in the Default CF, and reader reads a timestamp in write cf, CDC endpoint get old value by the timestamp. This may take lots of CPU during batch insert. Also it can cause OOM by block CDC endpoint.

master.

### Steps to reproduce

Batch insert rows that large than 255 byts.

hicqu:

Issue Number: Close #10256 

What's Changed:

```commit-message
cdc: load old value from data CF if necessary
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - [x] Consumes more CPU
    - [ ]Consumes more MEM

### Release note 

```release-note
None
```


hicqu:
This PR can be closed because fetching old-values in transaction layer increases transactin latency. It's inacceptable. However, the issue could still exist for the cases with very heavy insert load, although we didn't neet it for a long while. We can try another way to resolve it: fetching old-values in some workers in cdc module, or introducing a multi-thread CDC endpoint loop.

PR #13165, opened 2022-07-29
tablet: fix the potential dead lock in open_tablet

### LINKED ISSUE #13213: `open_tablet` may cause deadlock
open_tablet will acquire the lock of registry. When the tablet is not in registry, it will call open_tablet_raw which may call create_tablet which will acquire the lock of registry again.
So deadlock may happen

SpadeA-Tang:

Issue Number: Close #13213

What's Changed:

`open_tablet` will acquire the lock of `registry`. When the tablet is not in `registry`, it will call `open_tablet_raw` which may call `create_tablet` which will acquire the lock of `registry` again.
So deadlock may happen.

Besides handling with this deadlock problem, this PR refactor the open/create related methods in TabletFactory. 
There are too many open_xxx/create_xxx methods currently, this PR uses one open_tablet + OpenOptions to replace them.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```


tonyxuqqi (approved):
One more thing might improve on this struct is the pattern : acquire lock; open/create db; release lock; as the lock hold time can be long due to the intensive io in open/create db. Instead we can acquire lock check reg and release; open db; acquire lock add reg and release. This way the lock hold time is significantly reduced to improve the parallelism

SpadeA-Tang:
> One more thing might can improve on this struct is the pattern : acquire lock; open/create db; release lock; as the lock hold time can be long due to the intensive io in open/create db. Instead we can acquire lock check reg and release; open db; acquire lock add reg and release. This way the lock hold time is significantly reduced to improve the parallelism

I think this way may involve race condition problem. For example: thread A acquire lock, discover db not exist，release lock,    create db and done,  just after it thread B acquire lock, discover db not exist, release lock, create db. This will cause Error.

tonyxuqqi:
> > One more thing might can improve on this struct is the pattern : acquire lock; open/create db; release lock; as the lock hold time can be long due to the intensive io in open/create db. Instead we can acquire lock check reg and release; open db; acquire lock add reg and release. This way the lock hold time is significantly reduced to improve the parallelism
> 
> I think this way may involve race condition problem. For example: thread A acquire lock, discover db not exist，release lock, create db and done, just after it thread B acquire lock, discover db not exist, release lock, create db. This will cause Error.

For a specific tablet, there's only one thread handling it in the caller side. So the race condition won't happen.  Generally speaking, doing IO work while holding a lock is one of common cause of bad P99.

BusyJay:
> For a specific tablet, there's only one thread handling it in the caller side. So the race condition won't happen. Generally speaking, doing IO work while holding a lock is one of common cause of bad P99.

There can be multiple places opening a tablet. For example, when applying a snapshot, when splitting. Nevertheless, a function should be implemented correctly (either panic, marked unsafe or no data race) despite the caller's behavior.

src/server/engine_factory_v2.rs
```diff
@@ -30,6 +30,16 @@ fn get_id_and_suffix_from_path(path: &Path) -> (u64, u64) {
     (tablet_id, tablet_suffix)
 }
 
+impl KvEngineFactoryV2 {
+    fn create_tablet_helper(&self, id: u64, suffix: u64) -> Result<RocksEngine> {
```
BusyJay, line 34: Why not inline it to `open_tablet_raw` can make all other functions call `open_tablet_raw` instead?

BusyJay:
There are too many `open_xxx` methods in the trait, I suggest to add an `OpenOptions` for tablet factory and use one open method.

components/engine_traits/src/engine.rs
```diff
@@ -153,6 +153,27 @@ impl Drop for TabletErrorCollector {
     }
 }
 
+#[derive(Default)]
+pub struct OpenOptions {
+    pub read_only: bool,
```
BusyJay, line 158: Bad style

BusyJay:
Those `open_xxx` and `create_xxx` can be removed, only keeping one `open`.

components/engine_traits/src/engine.rs
```diff
@@ -153,32 +153,79 @@ impl Drop for TabletErrorCollector {
...
+
+    pub fn read_only(&self) -> bool {
+        self.read_only
+    }
+
+    pub fn cache_only(&self) -> bool {
+        self.cache_only
+    }
+
+    pub fn validate(&self) -> Result<()> {
```
BusyJay, line 206: How about overwriting instead of validating? For example, if ignore `cache_only` if `create` is set.

components/engine_traits/src/engine.rs
```diff
@@ -153,32 +153,79 @@ impl Drop for TabletErrorCollector {
     }
 }
 
+#[derive(Default)]
+pub struct OpenOptions {
```
BusyJay, line 157: You may need a `skip_cache` as open_raw should not update cache.

BusyJay: Or perhaps just keep `open` and `open_raw` as `open_raw` needs a `Path` instead of `region_id` and `suffix`.

components/engine_traits/src/engine.rs
SpadeA-Tang: Fine.

components/engine_traits/src/engine.rs
SpadeA-Tang: Okay.

components/engine_traits/src/engine.rs
```diff
@@ -153,32 +153,85 @@ impl Drop for TabletErrorCollector {
...
+    // create tablet if non-exist
+    create: bool,
+    create_new: bool,
+    read_only: bool,
+    cache_only: bool,
+    skip_cache: bool,
+}
+
+impl OpenOptions {
+    /// Sets the option to create a new file, or open it if it already exists.
```
BusyJay, line 168: Wrong comment, it's not for file..

components/engine_traits/src/engine.rs
```diff
@@ -153,32 +153,85 @@ impl Drop for TabletErrorCollector {
...
+    }
+}
+
 /// A factory trait to create new engine.
 // It should be named as `EngineFactory` for consistency, but we are about to
 // rename engine to tablet, so always use tablet for new traits/types.
 pub trait TabletFactory<EK>: TabletAccessor<EK> {
-    /// Create an tablet by id and suffix. If the tablet exists, it will fail.
+    /// Open the tablet with id and suffix according to the OpenOptions. The
+    /// implementation for v1 and v2 are fairly different. See the comments
```
BusyJay, line 224: It's inappropriate to state the implementation details in a trait's doc.

components/engine_traits/src/engine.rs
```diff
@@ -153,32 +153,85 @@ impl Drop for TabletErrorCollector {
...
-    /// be opened.
-    fn open_tablet_cache(&self, id: u64, suffix: u64) -> Option<EK>;
-
-    /// Open a tablet by id and any suffix from cache
-    fn open_tablet_cache_any(&self, id: u64) -> Option<EK>;
-
-    /// Open tablet by path and readonly flag
-    fn open_tablet_raw(&self, path: &Path, readonly: bool) -> Result<EK>;
+    /// Open tablet by raw path without updating cache.
+    fn open_tablet_raw(&self, path: &Path) -> Result<EK>;
```
BusyJay, line 234: I think it also needs a `OpenOption`.

components/engine_test/src/lib.rs
```diff
@@ -243,60 +247,80 @@ pub mod kv {
...
-        fn open_tablet_cache_any(&self, id: u64) -> Option<KvTestEngine> {
-            let reg = self.registry.lock().unwrap();
-            if let Some(k) = reg.keys().find(|k| k.0 == id) {
-                return Some(reg.get(k).unwrap().clone());
-            }
-            None
-        }
-
-        fn open_tablet_raw(&self, path: &Path, _readonly: bool) -> Result<KvTestEngine> {
+        fn open_tablet_raw(&self, path: &Path) -> Result<KvTestEngine> {
```
BusyJay, line 306: The path may not be named correctly. For example, it may be just a temporary path.

components/engine_traits/src/engine.rs
SpadeA-Tang: Done.

components/engine_traits/src/engine.rs
SpadeA-Tang: Good suggestion.

components/engine_test/src/lib.rs
SpadeA-Tang: Good catch.

components/engine_test/src/lib.rs
```diff
@@ -243,60 +253,80 @@ pub mod kv {
...
                 return Err(box_err!(
                     "path {} does not have db",
                     path.to_str().unwrap_or_default()
                 ));
-            }
-            let (tablet_id, tablet_suffix) = get_id_and_suffix_from_path(path);
-            self.create_tablet(tablet_id, tablet_suffix)
+            };
+
+            self.inner.create_tablet(path)
```
BusyJay, line 324: What if `options.create_new` is true?

SpadeA-Tang: We should return error if exists even in open_tablet_raw?

BusyJay: Yes, I think everything checked in `open_tablet` should be also checked in `open_tablet_raw`. An simple implementation is let `open_tablet` call `open_tablet_raw`.

youjiali1995 (approved):

Issue #13160, opened 2022-07-29
raftstore: apply res may be dropped silently


5kbpers:
### What version of TiKV are you using?

Versions after 4.0.

### What did happened?
In https://github.com/tikv/tikv/pull/8487 we changed the way of sending ApplyRes to `try_send`
https://github.com/tikv/tikv/blob/84654c87d6def968a197d47babad8e08acdf685a/components/raftstore/src/store/fsm/store.rs#L293-L302

In extreme cases, the mailbox of PeerFsms may be full and some of ApplyRes may be lost.
Then some regions that was supposed to created by split can not be actually created here.
https://github.com/tikv/tikv/blob/84654c87d6def968a197d47babad8e08acdf685a/components/raftstore/src/store/fsm/peer.rs#L3596-L3601

And as only part of ApplyRes was lost, the range of the original region could still be changed by other admin commands. Then a region that is actually overlapped with current regions can still be created.
https://github.com/tikv/tikv/blob/84654c87d6def968a197d47babad8e08acdf685a/components/raftstore/src/store/fsm/store.rs#L2058-L2063

Here is an example: 
* Region A [1, 100) was splitted into region B [1, 50) and region C [50, 100), wrote peer states of B and C into KVDB, and the corresponding ApplyRes was lost so region C was not created. 
* Then region C was splitted into region D [50, 75) and region E [75, 100)
* Region D can be created in the store with the first message, it received a snapshot and wrote its peer state to KVDB
* Then store restarted, it would continuously panic at `clear_stale_data` https://github.com/tikv/tikv/blob/84654c87d6def968a197d47babad8e08acdf685a/components/raftstore/src/store/fsm/store.rs#L1217-L1228

PR #13168, opened 2022-07-29
raftstore: use force_send to send ApplyRes

5kbpers:

Issue Number: Close #13160

What's Changed:

```commit-message
Use force_send to send ApplyRes
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more MEM

### Release note 

```release-note
Fix a bug that regions may be overlapped if raftstore is too busy
```

### DIFF
--- a/components/raftstore/src/store/fsm/store.rs
+++ b/components/raftstore/src/store/fsm/store.rs
@@ -292,16 +292,21 @@ where
 {
     fn notify(&self, apply_res: Vec<ApplyRes<EK::Snapshot>>) {
         for r in apply_res {
-            self.router.try_send(
-                r.region_id,
+            let region_id = r.region_id;
+            if let Err(e) = self.router.force_send(
+                region_id,
                 PeerMsg::ApplyRes {
                     res: ApplyTaskRes::Apply(r),
                 },
-            );
+            ) {
+                error!("failed to send apply result"; "region_id" => region_id, "err" => ?e);
+            }
         }
     }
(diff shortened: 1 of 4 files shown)


Connor1996 (approved):
LGTM

components/raftstore/src/store/fsm/store.rs
```diff
@@ -292,16 +292,21 @@ where
...
-                r.region_id,
+            let region_id = r.region_id;
+            if let Err(e) = self.router.force_send(
+                region_id,
                 PeerMsg::ApplyRes {
                     res: ApplyTaskRes::Apply(r),
                 },
-            );
+            ) {
+                error!("failed to send apply result"; "region_id" => region_id, "err" => ?e);
```
BusyJay, line 302: What if it's shutting down?

5kbpers: Already checked here https://github.com/tikv/tikv/blob/1e13ddf3bf12c00afb4d049d05978fecae9a6067/components/batch-system/src/router.rs#L263

gengliqi, line 302: Seems the error log is printed only when the channel is closed. 
So is it necessary to add this error log?

gengliqi (approved):
LGTM

components/raftstore/src/store/fsm/store.rs
5kbpers: A little is better than none. 🤣

Issue #13186, opened 2022-07-29
DoubleLock in components/engine_test and server


BurtonQin:
### Bug 1:
In components/engine_test/src/lib.rs,
`registry` is a std::sync::Mutex
The first lock is on L170, `open_tablet_raw` is called on L176, `create_tablet` is called on L205, and the second lock is on L150.
https://github.com/tikv/tikv/blob/b22be438650b914231c6ea4b0afb9c72538044c9/components/engine_test/src/lib.rs#L169-L176
https://github.com/tikv/tikv/blob/b22be438650b914231c6ea4b0afb9c72538044c9/components/engine_test/src/lib.rs#L205
https://github.com/tikv/tikv/blob/b22be438650b914231c6ea4b0afb9c72538044c9/components/engine_test/src/lib.rs#L149-L150
### Bug 2:
Similar to Bug 1.
In src/server/engine_factory_v2.rs,
The first lock is on L52, `open_tablet_raw` is called on L58, `create_table` is called on L89, and the second lock is on L35.

### What version of TiKV are you using?
commit: b22be438650b914231c6ea4b0afb9c72538044c9

### Steps to reproduce

Found by code analysis

BurtonQin:
I found that Bug 2 is fixed by https://github.com/tikv/tikv/pull/13165

PR #13187, opened 2022-07-29
components/engine_test: fix double-lock in open_tablet

BurtonQin:

Issue Number: Ref #13186

What's Changed:

```
Restrict `registry.lock()` in a scope so that it cannot double-lock with another `registry.lock()` in `open_tablet_raw`->`create_table`
Remove the redundant `reg.insert`, which has been called in `create_tablet`.
```

### Related changes

### Check List 

Tests 

- No code

Side effects

### Release note 

```release-note
Fix a double-lock bug in components/engine_test
```


components/engine_test/src/lib.rs
```diff
@@ -261,9 +261,11 @@ pub mod kv {
         }
 
         fn open_tablet(&self, id: u64, suffix: u64) -> Result<KvTestEngine> {
-            let reg = self.registry.lock().unwrap();
-            if let Some(db) = reg.get(&(id, suffix)) {
-                return Ok(db.clone());
+            {
+                let reg = self.registry.lock().unwrap();
+                if let Some(db) = reg.get(&(id, suffix)) {
```
Connor1996, line 266:
```suggestion
                if let Some(db) = self.registry.lock().unwrap().get(&(id, suffix)) {
```

Connor1996 (approved):
Thanks for your finding. LGTM

tonyxuqqi (approved):

PR #13188, opened 2022-07-29
components/pd_client: fix double-read-lock in client

### LINKED ISSUE #12933: components/pd_client: potential deadlocks caused by double-readlock in call_option
## Bug Report

In components/pd_client/src/client.rs, 
`Client.inner` is a `std::sync::RwLock<Inner>`
The first readlock is on L705, `call_option` is called on L709, and the second readlock is on L195.
https://github.com/tikv/tikv/blob/d356be1d051f38b9cd9c9239468073c05ccc6c03/components/pd_client/src/client.rs#L704-L709
https://github.com/tikv/tikv/blob/d356be1d051f38b9cd9c9239468073c05ccc6c03/components/pd_client/src/client.rs#L194-L197

Similarly, there are other 9 cases in client.rs:
1. L217, L221, L195
2. L256, L260, L195
3. L499, L503, L195
4. L537, L541, L195
5. L670, L674, L195
6. L749, L753, L195
7. L783, L787, L195
8. L924, L928, L195
9. L962, L966, L195

master branch

### Steps to reproduce

Found by code analysis
### What did you expect?

Cause deadlock if interleaved by a writelock. This is true for both std::sync::RwLock and parking_lot::RwLock.
For more details, see
https://www.reddit.com/r/rust/comments/urnqz8/different_behaviors_of_recursive_read_locks_in/

I think one possible solution is to add a new function `call_option_inner`, which takes `&Inner` as the parameter instead of `&Client`. Then there is no need to lock inside `call_option_inner`.

BurtonQin:
close tikv/tikv#12933

Signed-off-by: Burton Qin <bobbqqin@gmail.com>

Issue Number: Close #12933

What's Changed:

```
The first read lock `client.inner.rl()` is followed by `call_option(&Client)`.
The second read lock is inside `call_option(&Client)`.
To avoid the second read lock, the PR adds a new function `call_option_inner(&Inner)` to substitute `call_option(&Client)` where deadlock may happen.
A read lockguard binds explicitly to var `inner` and is reused by `call_option_inner(&Inner)`.

```

### Related changes

### Check List 

Tests 

- No code

Side effects

### Release note 

```release-note
Fix potential deadlock in `RpcClient` when two read locks are interleaved by a write lock.
```


youjiali1995 (approved):

bufferflies:
`get_gc_safe_point` may double read lock , plz fix it .
https://github.com/tikv/tikv/blob/5d5da83315af15ccc15c26b3af32e3615c749b42/components/pd_client/src/client.rs#L850-L865

BurtonQin:
This is not a double-read-lock because the read-lock in `call_option` has been released before calling `client.inner.rl()`.
However, it may cause atomicity violation if `call_option` and `client.inner.rl()` are interleaved by a write from another thread,
since the two operations are not protected by the same lock.

bufferflies:
plz fix DCO

components/pd_client/src/client.rs
```diff
@@ -192,9 +192,12 @@ impl RpcClient {
...
-            .inner
-            .rl()
+        Self::call_option_inner(&client.inner.rl())
+    }
+
+    #[inline]
+    fn call_option_inner(inner: &Inner) -> CallOption {
+        inner
             .target_info()
             .call_option()
```
Connor1996, line 202: As `target_info` is fixed, seems we can cache it and avoid building `call_option` everytime until `target_info` is changed after reconnection. /cc @BusyJay

AndreMouche (approved):
LGTM

Issue #12091, opened 2022-03-04
TiKV binary is not debuggable


coderplay:
## Bug Report

Rustc generates invalid DWARF when turning on LTO 
see
* https://github.com/rust-lang/rust/issues/41775
* https://github.com/rust-lang/rust/issues/66118

When I tried to debug tikv-sever, I got this error:
```
(gdb) info function Deadline
Dwarf Error: Cannot find DIE at 0x719bf5 referenced from DIE at 0x8b3ddf [in module /Users/xxxx/tikv-server]
```

rust-lldb also has other issues with the binary, not sure if it's related to LTO.

### What version of TiKV are you using?
My TiKV is 5.2.2, this bug should applies to other versions

### What operating system and CPU are you using?
Linux/MacOS

### Steps to reproduce
Download the `tikv-server` release binary, run below commands

```
% rust-gdb ./tikv-server

(gdb) info function Deadline
```

### What did you expect?
The binary should work with rust-gdb

### What did happened?
gdb threw errors.

coderplay:
I suggest to 

https://github.com/tikv/tikv/blob/master/Cargo.toml#L283

coderplay:
We should disable `LTO` in  https://github.com/tikv/tikv/blob/master/Cargo.toml#L283

Issue #10906, opened 2021-09-06
DWARF info of dist_release binary is incorrect


sticnarf:
### What version of TiKV are you using?

Not really related. It affects at least TiKV 4.0 ~ 5.2.

### What operating system and CPU are you using?

I can confirm the problem on Linux x86-64 and Linux aarch64.

### Steps to reproduce

Run `make dist_release`. And use `dwz` on the binary, it reports:

```
$ objcopy --decompress-debug-sections bin/tikv-server
$ dwz bin/tikv-server
dwz: bin/tikv-server: Couldn't find DIE at [d4bb7] referenced by DW_AT_abstract_origin from DIE at [17c2a3]
```

[dwarf-validate](https://github.com/gimli-rs/gimli/blob/master/examples/dwarf-validate.rs) also reports dwarf errors:

```
DWARF error in bin/tikv-server: Invalid intra-unit reference in unit 0xb4181 from DIE 0xc8122 to 0x20a36
DWARF error in bin/tikv-server: Invalid intra-unit reference in unit 0xb4181 from DIE 0xc813f to 0x208ea
DWARF error in bin/tikv-server: Invalid intra-unit reference in unit 0xb4181 from DIE 0xc8156 to 0x208c0
DWARF error in bin/tikv-server: Invalid intra-unit reference in unit 0xb4181 from DIE 0xc81b9 to 0x59c
...
```

### What did you expect?

There should not be dwarf errors.

### What did happened?

Dwarf tools report errors.

sticnarf:
This issue only happens if **full LTO** is used. The dwarf errors disappear if only using thin LTO.

Some upstream bug reports:

https://github.com/rust-lang/rust/issues/66118

Malformed dwarf information may cause some trouble to debuggers. GDB sometimes gets stuck or crashes when it encounters these incorrect dwarf info.

BusyJay:
Is it still an issue?

sticnarf:
Yes, until we upgrade the toolchain. (Although the new toolchain will produce a bit incomplete debug info.)

sticnarf:
Closed by https://github.com/tikv/tikv/pull/11875

PR #13190, opened 2022-08-01
*: update toolchain to 2022-07-31

tabokie:

Issue Number: Ref #13008

What's Changed:

As a follow-up on #13009, enable new lint `assertions_on_result_states` to deny `assert!(r.is_ok())` pattern in the codebase.

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/concurrency_manager/src/lib.rs
```diff
@@ -135,9 +135,9 @@ mod tests {
     #[tokio::test]
     async fn test_lock_keys_order() {
         let concurrency_manager = ConcurrencyManager::new(1.into());
-        let keys: Vec<_> = [b"c", b"a", b"b"]
-            .iter()
-            .map(|k| Key::from_raw(*k))
+        let keys: Vec<_> = vec![b"c", b"a", b"b"]
+            .into_iter()
+            .map(|k| Key::from_raw(k))
```
sticnarf, line 140: Umm, this change introduces an additional allocation for `vec![]` but brings no benefit.

As it's in a test, it's okay to just make clippy happy. But it's curious to know what does clippy complain about the old code?

sticnarf (commented):
Please help upgrade `pin-project` to fix the warning raised by CI.

components/concurrency_manager/src/lib.rs
tabokie: Check [this](https://github.com/rust-lang/rust-clippy/issues/9101).

components/engine_rocks/src/engine.rs
```diff
@@ -250,7 +250,7 @@ mod tests {
         engine.put_cf(cf, b"k1", b"v2").unwrap();
 
         assert_eq!(&*engine.get_value(b"k1").unwrap().unwrap(), b"v1");
-        assert!(engine.get_value_cf("foo", b"k1").is_err());
+        engine.get_value_cf("foo", b"k1").unwrap_err();
```
Connor1996, line 253: why

components/concurrency_manager/src/lib.rs
sticnarf: Let's use `[b"c", b"a", b"b"].iter().copied()` just like other cases in this PR.

sticnarf (commented):
Note that https://github.com/rust-lang/rust/pull/95685 will reopen https://github.com/tikv/tikv/issues/10906. As we are using frame pointers for the built-in profiling, the toolchain update will not affect profiling. But it may potentially make tikv not debuggable by gdb again (reported by https://github.com/tikv/tikv/issues/12091)

sticnarf:
The rest lgtm

sticnarf (approved):

src/storage/types.rs
```diff
@@ -80,7 +80,7 @@ impl MvccInfo {
 }
 
 /// Represents the status of a transaction.
-#[derive(PartialEq, Debug)]
+#[derive(PartialEq, Eq, Debug)]
```
BusyJay, line 83: Why need this?

tabokie: It’s the derive_partial_eq_without_eq lint.

BusyJay: Interesting. Will derive useless trait slow down compilation?

tabokie: I think it would: https://github.com/rust-lang/rust/issues/80118, I'll see if I can disable it.

tabokie: Updated and removed a bunch of unneeded `Eq`s.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -1166,8 +1166,8 @@ mod tests {
         ];
 
         for index in indices {
-            for order in orders {
-                test_split_impl(*index, *order);
+            for &order in orders {
```
BusyJay, line 1169: `*order` is a prefered style than `&order` matching.

scripts/clippy
```diff
@@ -15,7 +15,12 @@ if [[ -n "$SHELL_DEBUG" ]] ; then
   set -x
 fi
 
-CLIPPY_LINTS=(-A clippy::module_inception  \
+# Notes:
+# - Enables `significant_drop_in_scrutinee` after
+#   https://github.com/rust-lang/rust-clippy/issues/8963 is fixed.
+# - `derive_partial_eq_without_eq` has compilation overhead.
+CLIPPY_LINTS=(
```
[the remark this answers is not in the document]
tabokie: It's only to keep diff short when adding a new lint in the future.

components/raftstore/src/coprocessor/region_info_accessor.rs
tabokie: I prefer this than putting a `[allow(explicit_auto_deref)]`.

BusyJay: Then it's a false alarm from clippy as "deref which would be done by auto-deref" is false in this case. You can get around the problem by `order.as_slice()` or give orders a explicit type `orders: &[&[usize]]`.

components/engine_traits/src/cf_defs.rs
```diff
@@ -14,11 +14,5 @@ pub fn name_to_cf(name: &str) -> Option<CfName> {
...
         return Some(CF_DEFAULT);
     }
-    for c in ALL_CFS {
-        if name == *c {
-            return Some(c);
-        }
-    }
-
-    None
+    ALL_CFS.iter().find(|&c| name == *c).copied()
```
BusyJay, line 17: `copied` can also be used in this case.

BusyJay (approved):
Rest LGTM

components/engine_traits/src/cf_defs.rs
tabokie: Which case are you referring to? This LoC already uses `copied`.

sticnarf: I think what BusyJay expects is like `ALL_CFS.iter().copied().find(|c| name == *c)`, so there is no reference matching.

tabokie: Oh, right.

PR #13193, opened 2022-08-01
raftstore: make `UNREACHABLE_BACKOFF` configurable

### LINKED ISSUE #13054: Raftstore may broadcast too many messages after one peer store become unreachable
## Bug Report

After one store disconnected, StoreFsm would broadcast a `StoreUnreachable` message to all regions to mark unreachable peers. https://github.com/tikv/tikv/blob/0dff1be50281c72c55c5464751cd733032115ce4/components/raftstore/src/store/fsm/store.rs#L378

But if TiKV has too many regions, handling these `StoreUnreachable` messages could eat raftstore CPU and block all other messages, making TiKV getting stucked.

5kbpers:

Issue Number: Close #13054

What's Changed:

```commit-message
make `UNREACHABLE_BACKOFF` configurable.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Add a new option to make unreachable_backoff of raftstore configurable.
```


BusyJay (commented):
Any stats to show the improvement?

BusyJay:
Maintaining a new field in `StoreMeta` is too complicated and very easy to make mistake. I don't have the confidence to get it merged quickly without introducing new problems. Giving that it's not a common case, I suggest to make broadcast backoff configurable. This should be easy enough and ease the problem.

5kbpers:
> Maintaining a new field in `StoreMeta` is too complicated and very easy to make mistake. I don't have the confidence to get it merged quickly without introducing new problems. Giving that it's not a common case, I suggest to make broadcast backoff configurable. This should be easy enough and ease the problem.

Unified `regions` and `regions_on_store` into one struct to make sure they are modified together. IMO adding a new field to record relations between stores and regions is the best way to fix this issue fundamentally, it can reduce unnecessary messages to peers.

BusyJay:
> Unified `regions` and `regions_on_store` into one struct to make sure they are modified together. IMO adding a new field to record relations between stores and regions is the best way to fix this issue fundamentally, it can reduce unnecessary messages to peers.

The fundamental cause of the issue is too many regions in one node. Filter out unrelated regions is just an optimization instead of a fix.

5kbpers:
> > Unified `regions` and `regions_on_store` into one struct to make sure they are modified together. IMO adding a new field to record relations between stores and regions is the best way to fix this issue fundamentally, it can reduce unnecessary messages to peers.
> 
> The fundamental cause of the issue is too many regions in one node. Filter out unrelated regions is just an optimization instead of a fix.

In this issue these failed tiflash have only several tables, so there is actually a small part of regions that need to be notified that tiflash is unreachable.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -115,13 +114,77 @@ pub struct StoreInfo<EK, ER> {
     pub capacity: u64,
 }
 
+#[derive(Default)]
+pub struct StoreMetaRegions {
```
gengliqi, line 118: How about adding some unit tests to verify the consistency between `regions` and `regions_on_store`?

BusyJay:
> In this issue these failed tiflash have only several tables, so there is actually a small part of regions that need to be notified that tiflash is unreachable.

Too many regions is already not a normal case, only few regions are affected is another rare situation. We should not optimize for rare cases when there is already an easy solution.

5kbpers:
> > In this issue these failed tiflash have only several tables, so there is actually a small part of regions that need to be notified that tiflash is unreachable.
> 
> Too many regions is already not a normal case, only few regions are affected is another rare situation. We should not optimize for rare cases when there is already an easy solution.

OK, already reverted those changes, PTAL again.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2682,13 +2681,14 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
...
         let now = Instant::now();
+        let unreachable_backoff = self.ctx.cfg.unreachable_backoff.0;
         if self
             .fsm
             .store
             .last_unreachable_report
             .get(&store_id)
-            .map_or(UNREACHABLE_BACKOFF, |t| now.saturating_duration_since(*t))
-            < UNREACHABLE_BACKOFF
+            .map_or(unreachable_backoff, |t| now.saturating_duration_since(*t))
```
BusyJay, line 2690: Perhaps the time should be inserted after iterating all regions just in case iterating takes too long to finish.

components/raftstore/src/store/worker/check_leader.rs
```diff
@@ -83,7 +83,7 @@ impl Runner {
                 meta.region_ranges
                 // get overlapped regions
                 .range((Excluded(start_key), Unbounded))
-                .take_while(|(_, id)| end_key > enc_start_key(&meta.regions[id]))
+                .take_while(|(_, id)| end_key > enc_start_key(&meta.regions[*id]))
```
BusyJay, line 86: Why the change?

BusyJay (approved):
Rest LGTM

components/raftstore/src/store/worker/check_leader.rs
5kbpers: Missed these comments.....Will change this back in the next PR.

Issue #5875, opened 2019-11-13
Replica read may cause tikv panic


youjiali1995:
## Bug Report
A newly created peer has empty region info until it applies snapshot. If a read request is sent to it, it will cause panic at https://github.com/tikv/tikv/blob/master/src/raftstore/store/fsm/peer.rs#L2425.

Replica read should skip these peers because it will take a long time to catch up with the leader.

5kbpers:
Good catch, I will investigate it.

5kbpers:
Could you give me a more concrete detail? Like the backtrace.

5kbpers:
Have investigated this issue, the panic would happen during the time between the store creates a `PeerFsm` and the peer handles `ExecResult::ChangePeer`.
This case should be covered by our long-running test, I will address that after fixing this bug.

PR #13202, opened 2022-08-02
raftstore-v2: support lease read

tonyxuqqi:
Issue Number: Ref #12842

What's Changed:

```commit-message
Raftstore v2's read index, lease renew and local read.
Local read will not actually read the data in raftstore, it just returns a response so that the frontend local read can run. 
Stale read will not go through raftstore v2 at all and all the check is done in local read.  
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore/src/store/peer.rs
```diff
@@ -702,6 +724,253 @@ pub enum UnsafeRecoveryState {
     Destroy(UnsafeRecoveryExecutePlanSyncer),
 }
 
+// #[RaftstoreCommon], copied from Peer
+/// Peer's trait
+pub trait RaftPeer<EK, ER>
```
BusyJay, line 729: How about reimplementing the logic instead of using trait? And we should think whether the flow is still valid or whether can be improved.

tonyxuqqi: I think for the read function, its core logic does not have real difference for multi-rocks and single rocksdb. Do you have an example which part is invalid or can be improved specifically to multi-rocks version.

components/raftstore-v2/Cargo.toml
```diff
@@ -46,6 +46,7 @@ raft-proto = { version = "0.7.0" }
 raftstore = { path = "../raftstore" }
 slog = "2.3"
 smallvec = "1.4"
+tikv_alloc = { path = "../tikv_alloc" }
```
BusyJay, line 49: Not necessary anymore.

components/raftstore/src/store/peer.rs
```diff
@@ -704,6 +726,326 @@ pub enum UnsafeRecoveryState {
...
+    fn leader_id(&self) -> u64;
+    fn region_id(&self) -> u64;
+    fn tag(&self) -> &String;
+    fn txn_ext(&self) -> Arc<TxnExt>;
+    fn read_progress(&self) -> Arc<RegionReadProgress>;
+    fn peer(&self) -> &metapb::Peer;
+    fn bucket_meta(&self) -> Option<Arc<BucketMeta>>;
+    fn txn_extra_op(&self) -> Arc<AtomicCell<TxnExtraOp>>;
+    fn mut_pending_reads(&mut self) -> &mut ReadIndexQueue<Self::Callback>;
+    fn mut_bcast_wake_up_time(&mut self) -> &mut Option<TiInstant>;
```
BusyJay, line 754: We are increasing region size in v2, so hibernate regions will probably be removed. There is no need to wake up anyone.

components/raftstore/src/store/peer.rs
```diff
@@ -704,6 +726,326 @@ pub enum UnsafeRecoveryState {
...
+            } else {
+                Err(box_err!(
+                    "{} can not read due to injected failure",
+                    self.tag()
+                ))
+            }
+        );
+
+        // See more in ready_to_handle_read().
+        if self.is_splitting() {
```
BusyJay, line 778: You don't need to consider this ATM. This should be added when split/merge is supported. And they may not be the same as v1.

components/raftstore/src/store/peer.rs
```diff
@@ -704,6 +726,326 @@ pub enum UnsafeRecoveryState {
...
+
+    /// Try to renew leader lease.
+    fn maybe_renew_leader_lease(
+        &mut self,
+        ts: Timespec,
+        store_meta: &mut Arc<Mutex<StoreMeta>>,
+        progress: Option<ReadProgress>,
+    ) {
+        // A nonleader peer should never has leader lease.
+        let read_progress = if !self.is_leader() {
```
BusyJay, line 938: I suggest to integrate them with `Lease` struct, so all the checks can be hidden, and probably be more efficient because no need to check every condition but just a bool flag like:
```
fn on_propose_split() {
    self.lease.prevent_renew();
}

fn maybe_renew_leader_lease() {
    if let Some(lease) = self.lease.renew() {
       ...
   }
}
```

components/raftstore/src/store/peer.rs
```diff
@@ -704,6 +726,326 @@ pub enum UnsafeRecoveryState {
...
+        // problem if the leader applies fewer values than the follower, the follower
+        // read could get a newer value, and after that, the leader may read a stale
+        // value, which violates linearizability.
+        self.store_applied_index() >= read_index
+            // If it is in pending merge state(i.e. applied PrepareMerge), the data may be stale.
+            // TODO: Add a test to cover this case
+            && !self.has_pending_merge_state()
+            // a peer which is applying snapshot will clean up its data and ingest a snapshot file,
+            // during between the two operations a replica read could read empty data.
+            && !self.is_handling_snapshot()
```
BusyJay, line 1010: This is not true in v2. Because old data will be kept until all references are dropped.

components/raftstore/src/store/peer.rs
```diff
@@ -704,6 +726,326 @@ pub enum UnsafeRecoveryState {
...
+    fn maybe_inject_propose_error(
+        &self,
+        #[allow(unused_variables)] req: &RaftCmdRequest,
+    ) -> Result<()> {
+        // The return value format is {req_type}:{store_id}
+        // Request matching the format will fail to be proposed.
+        // Empty `req_type` means matching all kinds of requests.
+        // ":{store_id}" can be omitted, meaning matching all stores.
+        fail_point!("raft_propose", |r| {
+            r.map_or(Ok(()), |s| {
```
BusyJay, line 1023: These rules are not used anymore.

components/raftstore/src/store/peer.rs
tonyxuqqi: I don't think it has big advantage, if it has any advantage at all. Because you have to introduce these hook APIs like on_propose_split, on_propose_merge, and on_force_leader and call them properly. if it missed any call, then it will be a bug. And what's the return of this change?  It reduces two bool comparison! Is it a big deal?

components/raftstore/src/store/peer.rs
tonyxuqqi: They're used. see test_merge.rs in tests/failpoints/cases.

components/raftstore-v2/src/operation/query/follower_read_index.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// return true if it's proposed.
+    pub(crate) fn follower_read_index<T: Transport>(
+        &mut self,
+        poll_ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        mut err_resp: RaftCmdResponse,
+        ch: QueryResChannel,
+        now: Timespec,
+    ) -> bool {
+        if !self.is_leader() && self.leader_id() == INVALID_ID {
```
SpadeA-Tang, line 38: !self.is_leader() may be redundant here.

tonyxuqqi: good catch.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -436,6 +454,22 @@ impl<EK: KvEngine, ER: RaftEngine> StoreRouter<EK, ER> {
     pub fn logger(&self) -> &Logger {
         &self.logger
     }
+
+    #[inline]
+    pub fn send_read_command(
```
BusyJay, line 459: I don't think this abstraction is useful.

components/raftstore-v2/src/lib.rs
```diff
@@ -26,7 +26,7 @@
 
 mod batch;
 mod bootstrap;
-mod fsm;
+pub mod fsm;
```
BusyJay, line 29: It should be private.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -23,13 +27,31 @@ use crate::{
     router::{QueryResChannel, QueryResult},
 };
 
+mod follower_read_index;
```
BusyJay, line 30: Should update the module docs for these new modules.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -100,6 +89,16 @@ impl<ER> Storage<ER> {
     }
 }
 
+impl<ER: RaftEngine> Debug for Storage<ER> {
```
BusyJay, line 92: Why move it around?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -184,6 +183,21 @@ impl<ER: RaftEngine> Storage<ER> {
     pub fn is_initialized(&self) -> bool {
         self.region_state.get_tablet_index() != 0
     }
+
+    #[inline]
+    pub fn applied_index(&self) -> u64 {
```
BusyJay, line 188: This doesn't seem to be a useful abstraction.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -241,4 +406,82 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+    }
+
+    #[inline]
+    pub fn has_pending_merge_state(&self) -> bool {
+        // TODOTODO
+        false
+    }
+}
+
+impl<EK, ER> RequestInspector for Peer<EK, ER>
```
BusyJay, line 450: This should be put into `operation/query/read_index.rs`.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -241,4 +406,82 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+        }
+        state
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn init_peer<EK: KvEngine, ER: RaftEngine>() -> Result<Option<Peer<EK, ER>>> {
```
BusyJay, line 484: Why?

BusyJay:
Can you update the PR title to a more meaningful one?

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -132,6 +189,115 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+        if self.next_proposal_index() == propose_index {
+            // The message is dropped silently, this usually due to leader absence
+            // or transferring leader. Both cases can be considered as NotLeader error.
+            return Err(Error::NotLeader(self.region_id(), None));
+        }
+
+        Ok(Either::Left(propose_index))
+    }
+
+    pub(crate) fn post_propose<T>(
```
BusyJay, line 287: You can skip this for now. I will add it after proposal is supported.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -23,13 +27,31 @@ use crate::{
...
+    PeerFsmDelegate<'a, EK, ER, T>
+{
     #[inline]
     pub fn on_query(&mut self, req: RaftCmdRequest, ch: QueryResChannel) {
         if !req.has_status_request() {
-            unimplemented!();
+            let mut resp = RaftCmdResponse::default();
+            let term = self.fsm.peer().term();
+            cmd_resp::bind_term(&mut resp, term);
+            let policy = self.fsm.peer_mut().inspect(&req);
```
BusyJay, line 44: I don't think `inspect` is necessary anymore. Because we are expecting this code path is for read only. So a lot of checks are unnecessary or can return early.

components/raftstore-v2/src/batch/store.rs
tonyxuqqi: how do you reuse the code then? It's called in multiple places. 
BTW, there's the similar one in v1.

components/raftstore-v2/src/lib.rs
tonyxuqqi: The raftstore-v2/tests code needs it.

components/raftstore-v2/src/raft/storage.rs
tonyxuqqi: it's used in store_applied_index() without exposing the internal fields of Storage.

components/raftstore-v2/src/operation/query/mod.rs
tonyxuqqi: But still we need to differentiate ReadStale, Get, ReadIndex.  To me it's better to create a new inspect and copy some of the code from it.

components/raftstore-v2/src/batch/store.rs
BusyJay: Like where? It should only be called once.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -336,9 +339,23 @@ impl WriteCallback for CmdResChannel {
 /// need to be a field of the struct.
 #[derive(Clone, PartialEq, Debug)]
 pub struct ReadResponse {
+    pub response: RaftCmdResponse,
```
BusyJay, line 342: u64 should be enough for returning a read index.

components/raftstore/src/store/peer.rs
```diff
@@ -3848,79 +3906,33 @@ where
...
-                    error!(
-                        "failed to notify pd";
-                        "region_id" => self.region_id,
-                        "peer_id" => self.peer_id(),
-                        "err" => %e,
-                    )
-                }
-            }
-            self.should_wake_up = true;
+            self.wake_up_leader(poll_ctx);
```
BusyJay, line 3935: Should not be exacted to a function, at least not in this PR.

components/raftstore-v2/src/raft/storage.rs
BusyJay: `store_applied_index` is another useless abstraction. To query applied_index, just use `self.storage().entry_storage().applied_index()`. And there should be at most two places that needs to query the information.

components/raftstore-v2/src/batch/store.rs
tonyxuqqi: It's to hide some internal details like PeerMsg::RaftQuery or PeerMsg::RaftCommand. Also it's more convenient to use.

components/raftstore/src/store/mod.rs
```diff
@@ -16,15 +19,12 @@ mod async_io;
...
-pub mod local_metrics;
-mod peer;
 mod peer_storage;
-mod read_queue;
 mod region_snapshot;
 mod replication_mode;
 pub mod snap;
 mod txn_ext;
-mod worker;
+pub mod worker;
```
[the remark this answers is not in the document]
tonyxuqqi: not needed. good catch

components/raftstore-v2/src/batch/store.rs
```diff
@@ -1,40 +1,46 @@
...
+use crossbeam::channel::{Sender, TrySendError};
 use engine_traits::{Engines, KvEngine, RaftEngine, TabletFactory};
-use futures::{compat::Future01CompatExt, FutureExt};
 use kvproto::{metapb::Store, raft_serverpb::PeerState};
 use raft::INVALID_ID;
-use raftstore::store::{
-    fsm::store::PeerTickBatch, local_metrics::RaftMetrics, Config, RaftlogFetchRunner,
-    RaftlogFetchTask, StoreWriters, Transport, WriteMsg, WriteSenders,
+use raftstore::{
+    bytes_capacity,
```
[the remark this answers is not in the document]
tonyxuqqi: how to make clippy work in raftstore-v2?

components/raftstore-v2/src/router/response_channel.rs
tonyxuqqi: makes sense.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -394,6 +407,11 @@ impl ReadCallback for QueryResChannel {
 }
 
 pub type QueryResSubscriber = BaseSubscriber<QueryResult>;
+impl fmt::Debug for QueryResChannel {
```
BusyJay, line 410: Need blank line.

components/raftstore-v2/src/router/mod.rs
```diff
@@ -2,14 +2,15 @@
 
 mod imp;
 mod internal_message;
-mod message;
-mod response_channel;
+pub mod message;
```
BusyJay, line 5: Should not be public.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -184,6 +184,16 @@ impl<ER: RaftEngine> Storage<ER> {
     pub fn is_initialized(&self) -> bool {
         self.region_state.get_tablet_index() != 0
     }
+
+    #[inline]
+    pub fn applied_term(&self) -> u64 {
```
BusyJay, line 189: I don't think it's necessary. If `applied_term` needs to be wrapped, then every access method in `EntryStorage` also needs to be wrapped. It makes no sense.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -1,24 +1,52 @@
...
     tablet::{self, CachedTablet},
     Result,
 };
 
+const REGION_READ_PROGRESS_CAP: usize = 128;
+
 /// A peer that delegates commands between state machine and raft.
 pub struct Peer<EK: KvEngine, ER: RaftEngine> {
-    raft_group: RawNode<Storage<ER>>,
+    pub(crate) raft_group: RawNode<Storage<ER>>,
```
BusyJay, line 49: Why?

BusyJay (commented):
v1 doesn't pay much attention on access control, I hope v2 don't step into the same river.

components/raftstore-v2/src/raft/storage.rs
tonyxuqqi: then why raft_state and apply_state are wrapped above ?

tonyxuqqi: the reason is applied_term  will be very frequently used, just like in v1.

components/raftstore-v2/src/router/mod.rs
tonyxuqqi: every message is to be consumed by other module's code. You have to pub use every single definition in message if it's not pub.

components/raftstore-v2/src/raft/peer.rs
tonyxuqqi: just like async_write, it needs to be used in other files  (such as read_index.rs) but under Peer struct.

BusyJay: Then it can be accessed by `raft_group()` or `raft_group_mut()`.

components/raftstore/src/store/peer.rs
```diff
@@ -277,19 +277,19 @@ pub struct CheckTickResult {
     reason: &'static str,
 }
 
-pub struct ProposedAdminCmd<S: Snapshot> {
+pub struct ProposedAdminCmd<C: ErrorCallback> {
```
BusyJay, line 280: Why need to add `ErrorCallback`?

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -27,6 +52,12 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
     pub(crate) async_writer: AsyncWriter<EK, ER>,
     has_ready: bool,
     pub(crate) logger: Logger,
+    pub(crate) pending_reads: ReadIndexQueue<QueryResChannel>,
```
BusyJay, line 55: Should use accessor just like other fields instead of making them public. There are several fields are marked as `pub(crate)` because they can lead to borrow check failure.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -55,6 +58,9 @@ pub struct StoreContext<EK: KvEngine, ER: RaftEngine, T> {
     /// The precise timer for scheduling tick.
     pub timer: SteadyTimer,
     pub write_senders: WriteSenders<EK, ER>,
+    /// store meta
+    pub store_meta: Arc<Mutex<StoreMeta<EK>>>,
+    pub current_time: Option<Timespec>,
```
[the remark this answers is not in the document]
tonyxuqqi: it's used in read_index.rs.
self.maybe_renew_leader_lease(propose_time, &mut ctx.store_meta, None);

components/raftstore/src/store/peer.rs
tonyxuqqi: Same logic as you did for ReadIndexQueue, ReadIndexRequest etc. 
CmdEpochChecker and ProposedAdminCmd can be reused in v2.

components/raftstore-v2/src/raft/peer.rs
tonyxuqqi: > There are several fields are marked as pub(crate) because they can lead to borrow check failure.
Though they have valid reason, still they look very inconsistent.  And in fact, pending_reads also has a valid reason.
self.pending_reads_mut().advance_leader_reads(self.tag(), states)---it won't compile. 

The bigger problem I think in fact is splitting Peer's implementation to different files are not very right. These functions are very tight to the core of peer.rs and thus they need to access every private field of Peer's definition. But now they have to access them through accessors, which essentially make these field pub. It breaks the encapsulation principle already.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -8,28 +8,75 @@
...
+            cmd_resp::bind_term(&mut resp, term);
+            let policy = self.inspect_read(&req);
+            match policy {
+                Ok(RequestPolicy::ReadIndex) => {
+                    self.fsm
+                        .peer_mut()
+                        .read_index(self.store_ctx, req, resp, ch);
+                }
+                Ok(RequestPolicy::ReadLocal) => {
+                    let read_resp = ReadResponse::new(0);
```
SpadeA-Tang, line 70: Why returns 0 here? I guess it just notify the sender that the leader is in lease now, so the sender can retry the read using LocalReader again?

components/raftstore/src/store/peer.rs
BusyJay: I added it because they were used. It doesn't seem to be used in `ProposeAdminCmd`.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -27,6 +52,12 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
...
+    read_progress: Arc<RegionReadProgress>,
+    pub(crate) tag: String,
+
+    leader_lease: Lease,
+    pending_remove: bool,
```
BusyJay, line 60: Should be delete, there is no pending remove in v2.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -27,6 +52,12 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
     pub(crate) async_writer: AsyncWriter<EK, ER>,
     has_ready: bool,
     pub(crate) logger: Logger,
+    pub(crate) pending_reads: ReadIndexQueue<QueryResChannel>,
+    read_progress: Arc<RegionReadProgress>,
+    pub(crate) tag: String,
```
BusyJay, line 57: Should be removed, logger is used in v2.

components/raftstore-v2/src/raft/peer.rs
BusyJay: `region_id` and `peer_id` should be used in `ReadIndexQueue`.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -184,6 +184,11 @@ impl<ER: RaftEngine> Storage<ER> {
     pub fn is_initialized(&self) -> bool {
         self.region_state.get_tablet_index() != 0
     }
+
+    #[inline]
+    pub fn commit_index(&self) -> u64 {
```
BusyJay, line 189: This also seem to be only used once.

components/raftstore/src/store/peer.rs
```diff
@@ -556,6 +556,85 @@ pub fn start_unsafe_recovery_report<EK: KvEngine, ER: RaftEngine>(
...
+
+// check if the request can be amended to the last pending read?
+// return true if it can.
+pub fn can_amend_read<I: RequestInspector, C>(
+    last_pending_read: Option<&ReadIndexRequest<C>>,
+    req: &RaftCmdRequest,
+    lease_state: LeaseState,
+    max_lease: TimeDuration,
+) -> bool {
+    let now = monotonic_raw_now();
```
BusyJay, line 605: This abstraction makes one more time query, which is a syscall, compared to the past.

components/raftstore-v2/tests/integrations/test_read.rs
```diff
@@ -0,0 +1,85 @@
...
+use raftstore::store::util::new_peer;
+use tikv_util::{
+    codec::number::NumberEncoder,
+    thread_group::{set_properties, GroupProperties},
+};
+use txn_types::WriteBatchFlags;
+
+#[test]
+fn test_read_index() {
+    set_properties(Some(GroupProperties::default()));
```
BusyJay, line 19: Properties is used for detect shutdown when clearing callbacks. In v2, we can have other way to detect such failure. I suggest not to include it.

tonyxuqqi: Without this line, it will panic.

components/raftstore-v2/src/operation/query/mod.rs
tonyxuqqi: > Why returns 0 here? I guess it just notifies the sender that the leader is in lease now, so the sender can retry the read using LocalReader again?

Yes. Generally speaking I think as long as it returns QueryResult::Read() regardless the read index value, then it means the local read can be done.

components/raftstore/src/store/peer.rs
tonyxuqqi: Good catch. Add the now as parameter.

components/raftstore-v2/src/raft/peer.rs
tonyxuqqi: used tag to make it compatible with v1's code. As discussed, v2's peer does not have tag field.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -8,28 +8,75 @@
...
+            }
+        }
+    }
 
-impl<'a, EK: KvEngine, ER: RaftEngine, T> PeerFsmDelegate<'a, EK, ER, T> {
     #[inline]
     pub fn on_query(&mut self, req: RaftCmdRequest, ch: QueryResChannel) {
         if !req.has_status_request() {
-            unimplemented!();
+            let mut resp = RaftCmdResponse::default();
```
BusyJay, line 59: Why not let the processing create `Response` by demand?

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -8,28 +8,75 @@
...
 
+mod follower_read_index;
 mod local;
+mod read_index;
+mod replica_read;
+
+impl<'a, EK: KvEngine, ER: RaftEngine, T: raftstore::store::Transport>
+    PeerFsmDelegate<'a, EK, ER, T>
+{
+    fn inspect_read(&mut self, req: &RaftCmdRequest) -> Result<RequestPolicy> {
```
BusyJay, line 41: Although we expect there should be only read requests, but we need to check if it's misused.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -8,28 +8,75 @@
...
+        let flags = WriteBatchFlags::from_bits_check(req.get_header().get_flags());
+        if flags.contains(WriteBatchFlags::STALE_READ) {
+            return Ok(RequestPolicy::StaleRead);
+        }
+
```
BusyJay, line 46: Why not check `header.read_quorum`?

BusyJay: And should add TODO for `applied_to_current_term`

components/raftstore-v2/src/operation/query/replica_read.rs
```diff
@@ -0,0 +1,111 @@
...
+            } else {
+                // The request could be proposed when the peer was leader.
+                // TODO: figure out that it's necessary to notify stale or not.
+                let term = self.term();
+                notify_stale_req(term, ch);
+            }
+        }
+    }
+
+    pub(crate) fn can_replica_read(
```
BusyJay, line 60: I don't think it needs to be `pub(crate)`.

components/raftstore-v2/src/operation/query/replica_read.rs
```diff
@@ -0,0 +1,111 @@
...
+        &self,
+        req: RaftCmdRequest,
+        check_epoch: bool,
+        read_index: Option<u64>,
+    ) -> QueryResult {
+        let region = self.region().clone();
```
BusyJay, line 66: Missing metrics.

components/raftstore-v2/src/operation/query/replica_read.rs
```diff
@@ -0,0 +1,111 @@
...
+        if check_epoch {
+            if let Err(e) = check_region_epoch(&req, &region, true) {
+                debug!(self.logger, "epoch not match"; "err" => ?e);
```
BusyJay, line 69: No need to log.

components/raftstore-v2/src/operation/query/replica_read.rs
```diff
@@ -0,0 +1,111 @@
...
+                let mut response = cmd_resp::new_error(e);
+                cmd_resp::bind_term(&mut response, self.term());
+                return QueryResult::Response(response);
```
BusyJay, line 72: This is actually a mis-usage. `QueryResult::Response` is for status command. In this case, it should return an error by `ch.report_error(resp)`.

components/raftstore-v2/src/operation/query/replica_read.rs
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+        let flags = WriteBatchFlags::from_bits_check(req.get_header().get_flags());
+        if flags.contains(WriteBatchFlags::STALE_READ) {
```
BusyJay, line 76: This is checked for stale read already. I suggest to use a standalone function for stale read, so it doesn't need to be checked again and again.

components/raftstore-v2/src/operation/query/replica_read.rs
tonyxuqqi: > I don't think it needs to be `pub(crate)`.

It needs to be called in mod.rs.

components/raftstore-v2/src/operation/query/replica_read.rs
tonyxuqqi: What specific metrics are you refering? Do you mean "metrics.read_index_confirm_wait_nanos "?

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,34 +2,112 @@
 
 //! There are two types of Query: KV read and status query.
 //!
-//! KV Read is implemented in local module and lease module (not implemented
-//! yet). Read will be executed in callee thread if in lease, which is
+//! KV Read is implemented in local module and query module.
```
BusyJay, line 5: `local` module is in `query` module.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,34 +2,112 @@
...
+//! Read will be executed in callee thread if in lease, which is
 //! implemented in local module. If lease is expired, it will extend the lease
 //! first. Lease maintainance is implemented in lease module.
 //!
 //! Status query is implemented in the root module directly.
+//! Follower's read index and replica read is implemenented replica.rs.
```
BusyJay, line 11: Inconsistent comment. It's called xxx module instead of xxx.rs in the above sentences.

components/raftstore-v2/src/operation/query/mod.rs
BusyJay: All read related checks from `pre_propose_raft_command` should also be checked here.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,352 @@
...
+    pub fn read_index<T: Transport>(
+        &mut self,
+        poll_ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: QueryResChannel,
+    ) -> bool {
+        // TODO: add pre_read_index to handle splitting or merging
+        let now = monotonic_raw_now();
+        if self.is_leader() {
+            self.leader_read_index(poll_ctx, req, ch, now)
```
BusyJay, line 145: Should be named `read_index_leader`.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,352 @@
...
+        true
+    }
+
+    fn leader_read_index<T: Transport>(
+        &mut self,
+        poll_ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: QueryResChannel,
+        now: Timespec,
+    ) -> bool {
```
BusyJay, line 84: Why need to return `bool`?

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,352 @@
...
+    // 2. The message is stale and dropped by the Raft group internally;
+    // 3. There is already a read request proposed in the current lease;
+    pub fn read_index<T: Transport>(
+        &mut self,
+        poll_ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: QueryResChannel,
+    ) -> bool {
+        // TODO: add pre_read_index to handle splitting or merging
+        let now = monotonic_raw_now();
```
BusyJay, line 143: Better call on demand.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,352 @@
...
+        };
+        let mut resp = RaftCmdResponse::default();
+        resp.mut_header().set_error(err);
+        read_ch.report_error(resp);
+    }
+
+    /// response the read index request
+    ///
+    /// awake the read tasks waiting in frontend (such as unified thread pool)
+    pub(crate) fn response_read<T>(
```
BusyJay, line 181: Should be `respond_read`.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    fn leader_read_index<T: Transport>(
+        &mut self,
+        poll_ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: QueryResChannel,
+        now: Timespec,
+    ) -> bool {
+        let lease_state = self.inspect_lease();
```
BusyJay, line 85: Isn't the lease inspected before?

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,352 @@
...
+        debug!(
+            self.logger,
+            "update read progress";
+            "progress" => ?progress,
+        );
+        reader.update(progress);
+    }
+}
+
+impl<EK, ER> RequestInspector for Peer<EK, ER>
```
BusyJay, line 324: The trait seems useless on v2 as we use `inspect_read` to validate read requests, so no only the implemented functions are used. And they can be implemented as methods instead.

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: > Should be `respond_read`.

This is to be consistent with v1's code:)

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: read_index is bool (same as v1)----I expect we will return bool as well in on_query() once the propose is done, as in v1 it will need the return value to set the has_ready flag.

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: not necessary. it may be skipped in some code path.

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: This is to save one system call, as you suggested before. Original version is put this line to the following two read_index fns.

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: But implementing this trait has better flexibility and it does not have additional cost.  We may call inspect in the future.

components/raftstore-v2/src/operation/query/lease.rs
BusyJay: We don't need to, let's make code more readable.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,344 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+use std::sync::{Arc, Mutex};
```
BusyJay, line 2: Missing blank line above.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,41 +2,176 @@
...
+            return Err(e);
+        }
+
+        // TODO: add flashback_state check
+
+        // Check whether the store has the right peer to handle the request.
+        let leader_id = self.leader_id();
+        let request = msg.get_requests();
+
+        // TODO: add false leader
```
BusyJay, line 135:
```suggestion
        // TODO: add force leader
```

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,41 +2,176 @@
...
             self.fsm.peer_mut().on_query_status(&req, ch);
         }
     }
 }
 
 impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    fn validate_query_msg<T: Transport>(
+        &mut self,
+        msg: &RaftCmdRequest,
+        ctx: &mut StoreContext<EK, ER, T>,
```
BusyJay, line 103: You can only pass `raft_metrics`, which should reduce code.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,344 @@
...
+        // }
+        // }
+    }
+
+    // Returns a boolean to indicate whether the `read` is proposed or not.
+    // For these cases it won't be proposed:
+    // 1. The region is in merging or splitting;
+    // 2. The message is stale and dropped by the Raft group internally;
+    // 3. There is already a read request proposed in the current lease;
+    pub fn read_index<T: Transport>(
```
BusyJay, line 134: How about inlining the function giving it actually has only 5 lines.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,344 @@
...
+            "request_id" => ?id,
+            "is_leader" => is_leader,
+        );
+
+        true
+    }
+
+    fn read_index_leader<T: Transport>(
+        &mut self,
+        poll_ctx: &mut StoreContext<EK, ER, T>,
```
BusyJay, line 79: It's named `ctx` in v2 normally.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,344 @@
...
+    ) -> bool {
+        poll_ctx.raft_metrics.propose.read_index.inc();
+
+        let request = req
+            .mut_requests()
+            .get_mut(0)
+            .filter(|req| req.has_read_index())
+            .map(|req| req.take_read_index());
+        let (id, dropped) = propose_read_index(self.raft_group_mut(), request.as_ref(), None);
+        if dropped && is_leader {
```
BusyJay, line 57: I think `is_leader` is checked already in `validate_query_msg`, it has to be true all the time.

components/raftstore-v2/src/operation/query/replica.rs
```diff
@@ -0,0 +1,161 @@
...
+                .inc();
+            let mut err_resp = RaftCmdResponse::default();
+            let term = self.term();
+            cmd_resp::bind_term(&mut err_resp, term);
+            cmd_resp::bind_error(&mut err_resp, Error::NotLeader(self.region_id(), None));
+            ch.report_error(err_resp);
+            return;
+        }
+
+        if self.propose_read_index(poll_ctx, req, self.is_leader(), ch, now) {
```
BusyJay, line 54: I suggest to inline the `propose_read_index` implementation to make it clear. Otherwise it's weird to let replica use the same method from lease module and lease module needs to check whether it's leader.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,344 @@
...
+    ) {
+        // TODO: add pre_read_index to handle splitting or merging
+        if self.is_leader() {
+            self.read_index_leader(poll_ctx, req, ch);
+        } else {
+            self.read_index_follower(poll_ctx, req, ch);
+        }
+    }
+
+    pub(crate) fn send_read_command<T>(
```
BusyJay, line 148: It seems only be used in replica module.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,344 @@
...
+        if !self.raft_group().raft.in_lease() {
+            return LeaseState::Suspect;
+        }
+        // None means now.
+        let state = self.leader_lease_mut().inspect(None);
+        if LeaseState::Expired == state {
+            debug!(
+                self.logger,
+                "leader lease is expired, region_id {}, peer_id {}, lease {:?}",
+                self.region_id(),
```
BusyJay, line 335: They are included in logger already.

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: The caller entry is too long to look good when it has repeated more than once. Also it has a TODO which will add more code into it later.  I prefer the current way. 
 Ok(RequestPolicy::ReadIndex) => {
                    self.fsm.peer_mut().read_index(self.store_ctx, req, ch);
                }
                vs.
                // TODO: add pre_read_index to handle splitting or merging
                 if self.fsm.peer_mut().is_leader() {
                      self.fsm.peer_mut().read_index_leader(poll_ctx, req, ch);
                  } else {
                      self.fsm.peer_mut().read_index_follower(poll_ctx, req, ch);
                  }

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: > 
No, read_index_follower will call this too.

components/raftstore-v2/src/operation/query/replica.rs
```diff
@@ -0,0 +1,199 @@
...
+                cmd.ch
+            }
+            _ => unreachable!(),
+        };
+        let mut resp = RaftCmdResponse::default();
+        resp.mut_header().set_error(err);
+        read_ch.report_error(resp);
+    }
+
+    fn response_replica_read<T>(
```
BusyJay, line 138:
```suggestion
    fn respond_read_replica<T>(
```

Note, it doesn't has to be replica_read.

components/raftstore-v2/src/operation/query/replica.rs
```diff
@@ -0,0 +1,199 @@
...
+                continue;
+            }
+            if req.get_header().get_replica_read() {
+                // We should check epoch since the range could be changed.
+                let region = self.region().clone();
+                if let Err(e) = check_region_epoch(&req, &region, true) {
+                    let mut response = cmd_resp::new_error(e);
+                    cmd_resp::bind_term(&mut response, self.term());
+                    ch.report_error(response);
+                }
```
BusyJay, line 177: Why not respond to replica read?

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,301 @@
...
+        } else {
+            self.read_index_follower(ctx, req, ch);
+        }
+    }
+
+    /// response the read index request
+    ///
+    /// awake the read tasks waiting in frontend (such as unified thread pool)
+    /// In v1, it's named as response_read.
+    pub(crate) fn respond_read<T>(
```
BusyJay, line 138: It seems `respond_read_index` is a better name as it only handles read index requests.

components/raftstore-v2/src/operation/query/stale.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use crate::{
+    batch::StoreContext,
+    raft::Peer,
+    router::{QueryResChannel, QueryResult, ReadResponse},
+    Result,
+};
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    pub(crate) fn can_stale_read(
```
BusyJay, line 23:
```suggestion
    pub(crate) fn respond_stale_read(
```

components/raftstore-v2/src/operation/query/stale.rs
```diff
@@ -0,0 +1,62 @@
...
+                let mut response = cmd_resp::new_error(e);
+                cmd_resp::bind_term(&mut response, self.term());
+                ch.report_error(response);
+                return;
+            }
+        }
+        let read_ts = decode_u64(&mut req.get_header().get_flag_data()).unwrap();
+        let safe_ts = self.read_progress().safe_ts();
+        if safe_ts < read_ts {
+            warn!(
```
BusyJay, line 42: I don't think it needs to be a warning.

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: but still it will have more code here. I think it's better to have a short and clear call in the switch.

BusyJay: It won't. those code should be put into `read_index_leader` instead of the switch.

tonyxuqqi: As discussed, put the read_index to mod.rs.

components/raftstore-v2/src/operation/query/replica.rs
tonyxuqqi: good catch.

components/raftstore-v2/src/operation/query/mod.rs
tonyxuqqi: it's checked in validate_query_msg.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -149,7 +215,6 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
         &self.tablet
     }
 
-    #[inline]
```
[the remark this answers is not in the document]
tonyxuqqi: I think it's a merge error.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,41 +2,318 @@
...
+            return Err(Error::NotLeader(self.region_id(), None));
+        }
+
+        // peer_id must be the same as peer's.
+        if let Err(e) = util::check_peer_id(msg, self.peer_id()) {
+            raft_metrics.invalid_proposal.mismatch_peer_id.inc();
+            return Err(e);
+        }
+        // check whether the peer is initialized.
+        if !self.storage().is_initialized() {
```
BusyJay, line 165: I don't think we need to check this on v2. If it's lease read, then it can't be uninitialized; if it's a follower read, then it must wait for apply, so it will be initialized in the end.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,230 @@
...
+    fn maybe_update_read_progress(&self, reader: &mut ReadDelegate, progress: ReadProgress) {
+        debug!(
+            self.logger,
+            "update read progress";
+            "progress" => ?progress,
+        );
+        reader.update(progress);
+    }
+
+    pub(crate) fn has_applied_to_current_term(&mut self) -> bool {
```
BusyJay, line 209: Should be in raft/peer.rs.

components/raftstore-v2/src/operation/query/replica.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// read index on follower
+    ///
+    /// call set_has_ready if it's proposed.
+    pub(crate) fn read_index_follower<T: Transport>(
+        &mut self,
+        ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: QueryResChannel,
+    ) {
+        let now = monotonic_raw_now();
```
BusyJay, line 39: This can be moved to L57.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,230 @@
...
+                    t.metrics.read_index_confirm_wait_nanos = (time - read_index_req.propose_time)
+                        .to_std()
+                        .unwrap()
+                        .as_nanos()
+                        as u64;
+                })
+            });
+
+            // leader reports key is locked
+            if let Some(locked) = read_index_req.locked.take() {
```
BusyJay, line 137: It can't be executed on Leader. I suggest to an assert `None` instead.

BusyJay (commented):
Rest LGTM

components/raftstore-v2/src/operation/query/mod.rs
SpadeA-Tang: So, why v1 need this check?

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,41 +2,318 @@
...
+
+        // TODO: add force leader
+
+        // ReadIndex can be processed on the replicas.
+        let is_read_index_request =
+            request.len() == 1 && request[0].get_cmd_type() == CmdType::ReadIndex;
+
+        let allow_replica_read = msg.get_header().get_replica_read();
+        let flags = WriteBatchFlags::from_bits_check(msg.get_header().get_flags());
+        let allow_stale_read = flags.contains(WriteBatchFlags::STALE_READ);
```
SpadeA-Tang, line 153: I think this is unnecessary as if it is stale read it will return error above.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,41 +2,318 @@
...
+            return Ok(RequestPolicy::ReadIndex);
+        }
+
+        // If applied index's term is differ from current raft's term, leader transfer
+        // must happened, if read locally, we may read old value.
+        if !self.fsm.peer_mut().has_applied_to_current_term() {
+            return Ok(RequestPolicy::ReadIndex);
+        }
+
+        match self.fsm.peer_mut().inspect_lease() {
```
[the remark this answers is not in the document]
tonyxuqqi: inspect_lease needs mut.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -2,41 +2,318 @@
...
+                Ok(RequestPolicy::ReadIndex) => {
+                    self.fsm.peer_mut().read_index(self.store_ctx, req, ch);
+                }
+                Ok(RequestPolicy::ReadLocal) => {
+                    self.store_ctx.raft_metrics.propose.local_read.inc();
+                    let read_resp = ReadResponse::new(0);
+                    ch.set_result(QueryResult::Read(read_resp));
+                }
+                _ => {
+                    unimplemented!();
```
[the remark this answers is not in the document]
tonyxuqqi: actually it should never happen and can panic!

components/raftstore-v2/src/operation/query/mod.rs
tonyxuqqi: good catch. The code above is added after this line. It should be removed.

components/raftstore-v2/src/operation/query/mod.rs
tonyxuqqi: > I don't think we need to check this on v2. If it's lease read, then it can't be uninitialized; if it's a follower read, then it must wait for apply, so it will be initialized in the end.

So we should add assert instead then.  Also why it needs the check in v1?

components/raftstore-v2/src/operation/query/lease.rs
tonyxuqqi: > It can't be executed on Leader. I suggest to an assert `None` instead.

Good Catch.

components/raftstore-v2/tests/integrations/test_read.rs
```diff
@@ -0,0 +1,213 @@
...
+    region_epoch.set_version(region_epoch.get_version() + 1);
+    req.mut_header().set_region_epoch(region_epoch.clone());
+    req.mut_header().set_term(6);
+    let res = router.query(region_id, req.clone()).unwrap();
+    let error_resp = res.response().unwrap();
+    assert!(error_resp.get_header().has_error());
+}
+
+#[test]
+fn test_read_stale() {
```
BusyJay, line 181: Why not merge it with test_snap_with_invalid_parameter?

components/raftstore-v2/src/operation/query/mod.rs
tonyxuqqi: So in v1 it's mainly for replica read scenario, which is not needed in v2's raftstore. 
https://github.com/tikv/tikv/issues/5875

components/raftstore-v2/tests/integrations/test_read.rs
tonyxuqqi: good point.

components/raftstore-v2/tests/integrations/test_read.rs
```diff
@@ -0,0 +1,188 @@
...
+    assert!(error_resp.get_header().has_error());
+
+    // run again, with incorrect peer_id
+    req.mut_header().set_peer(new_peer(1, 4));
+    let res = router.query(region_id, req.clone()).unwrap();
+    let error_resp = res.response().unwrap();
+    assert!(error_resp.get_header().has_error());
+
+    // run with stale term
+    req.mut_header().set_term(1);
```
BusyJay, line 165: If you keep one correct request, and then clone + modify to test the wrong case, this test can be simplified.

BusyJay (approved):

components/raftstore-v2/tests/integrations/test_read.rs
tonyxuqqi: Good point. Updated.

components/raftstore-v2/src/operation/query/lease.rs
```diff
@@ -0,0 +1,246 @@
...
+};
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    pub(crate) fn read_index_leader<T: Transport>(
+        &mut self,
+        ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: QueryResChannel,
+    ) {
+        let now = monotonic_raw_now();
```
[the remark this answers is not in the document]
tonyxuqqi: which function to reuse this "now"?

5kbpers (approved):

PR #13206, opened 2022-08-02
server: make EnginesResourcesInfo be compatible with Multi-Rocks DB version.

### LINKED ISSUE #13214: EnginesResourcesInfo should be compatible with Multi-Rocks DB version.
## Development Task
Now, EnginesResourcesInfo fetches the pending bytes from the single global kv rocks db. When Multi-rocks DB configuration is set, it is not suitable.

SpadeA-Tang:

Issue Number: Close #13214

What's Changed:

`IoRateLimiter`, a server level IO rate limiter, is used to control the total budgets allowed for different type of operations according to their priority.  It recalculates total budgets for each priority according to to the current flow condition.
When the IO rate mode is `WriteOnly`, the total budget is further adjusted by `EnginesResourceInfo` for **Low priority** operations according to the maximum values of the compaction pending bytes of kv engine (including default, lock, and write CFs) and raft engine with the main purpose of avoiding write stall.

Now, in Multi-Rocks DB settings, each region has it's own kv engine ( called tablet now ), so this PR modifies the relevant method of `EnginesResourceInfo` to consider the tablets of all regions.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```


components/server/src/server.rs
```diff
@@ -1874,9 +1877,27 @@ impl EnginesResourceInfo {
...
+                    }
+                }
+                collections::HashMapEntry::Vacant(slot) => {
+                    slot.insert((suffix, db.clone()));
+                }
+            },
+        );
+        for (_, (_, tablet)) in tabltes {
+            for cf in &[CF_DEFAULT, CF_WRITE, CF_LOCK] {
+                fetch_engine_cf(&tablet, cf, &mut normalized_pending_bytes);
```
tonyxuqqi, line 1897: is it possible that the tablet can be destroyed while being referenced here?
I think it's probably fine here if we won't destroy the tablet once the region peer is moved to other TiKV, otherwise there's race condition here that once for_each_opened_tablet is done,  some tablet is deleted already. 
To make it more robust,  I think we may add a flag in TabletFactory to mark it being used so that the data files or the underlying memory structure won't be deleted.

SpadeA-Tang: It indeed has the race condition problem. Let me come up with a way to handle with it.

components/server/src/server.rs
```diff
@@ -1874,9 +1877,31 @@ impl EnginesResourceInfo {
         if let Some(raft_engine) = &self.raft_engine {
             fetch_engine_cf(raft_engine, CF_DEFAULT, &mut normalized_pending_bytes);
         }
-        for cf in &[CF_DEFAULT, CF_WRITE, CF_LOCK] {
-            fetch_engine_cf(&self.kv_engine, cf, &mut normalized_pending_bytes);
+
+        // Record the compaction pending bytes of the tablet with the latest suffix for
+        // each region
+        let mut pending_bytes_record = HashMap::<_, (_, _)>::new();
```
tabokie, line 1883: Put the hashmap inside `EnginesResourceInfo` so that we don't need to reallocate every few seconds.

SpadeA-Tang: If one tablet is destroyed, how can we remove it from the hashmap?

tabokie: clear it before doing the scan.

components/server/src/server.rs
```diff
@@ -1831,23 +1834,26 @@ impl<EK: KvEngine, R: RaftEngine> EngineMetricsManager<EK, R> {
 }
 
 pub struct EnginesResourceInfo {
-    kv_engine: RocksEngine,
+    tablet_factory: Arc<KvEngineFactory>,
     raft_engine: Option<RocksEngine>,
+    // region_id -> (suffix, normalized_pending_bytes)
+    normalized_pending_baytes_record: Arc<Mutex<HashMap<u64, (u64, u32)>>>,
```
tabokie, line 1840:
```suggestion
    cache_latest_kv_pending_bytes: Arc<Mutex<HashMap<u64, (u64, u32)>>>,
```

tabokie (approved):

components/server/src/server.rs
```diff
@@ -1874,9 +1880,34 @@ impl EnginesResourceInfo {
...
-            fetch_engine_cf(&self.kv_engine, cf, &mut normalized_pending_bytes);
+
+        let mut cached_latest_kv_pending_bytes =
+            self.cached_latest_kv_pending_bytes.as_ref().lock().unwrap();
+        // cached_latest_kv_pending_bytes records the compaction pending bytes of the
+        // tablet with the latest suffix for each region in each regular update. Clear
+        // ensures that detroyed tablet will not be in it forever.
+        cached_latest_kv_pending_bytes.clear();
+        self.tablet_factory
+            .for_each_opened_tablet(&mut |id, suffix, db: &RocksEngine| {
```
tonyxuqqi, line 1891: The for_each_opened_tablet will acquire the lock, which blocks other operations such as split or apply a snapshot. 
So the key here is to figure out the total time of this function when we have 1000 instances, for example. If it takes seconds, then this approach will greatly impact P99 latency.

SpadeA-Tang: Yes. Using for_each_opend_tablet to fetch pending bytes will occupy the lock for a long time. Here's a bench comparison between fetching pending bytes and just cloning them. The difference is so noticable.
Kx2H1VCJoP
rftpc1ZzDL

tonyxuqqi (approved):

components/server/src/server.rs
```diff
@@ -1841,23 +1844,29 @@ impl<EK: KvEngine, R: RaftEngine> EngineMetricsManager<EK, R> {
...
 
 pub struct EnginesResourceInfo {
-    kv_engine: RocksEngine,
+    tablet_factory: Arc<KvEngineFactory>,
     raft_engine: Option<RocksEngine>,
+    // region_id -> (suffix, tablet)
+    // `update` is called perodically which needs this map for recording the latest tablet for each
+    // region and cached_latest_tablets is used to avoid memory allocation each time when
+    // calling `update`.
+    cached_latest_tablets: Arc<Mutex<HashMap<u64, (u64, RocksEngine)>>>,
```
BusyJay, line 1853: Why not make `update` accept a map? So lock is not required.

SpadeA-Tang: So who passes the map to `update` ? I think acquiring lock here is very fast and no contention will be involved.

BusyJay:
```
let engines_info_clone = engines_info.clone();
let mut map = HashMap::default();
self.background_worker
    .spawn_interval_task(DEFAULT_METRICS_FLUSH_INTERVAL, move || {
        let now = Instant::now();
        engine_metrics.flush(now);
        io_metrics.flush(now);
        engines_info_clone.update(now, &mut map);
    });
```

If lock is not used for solving conflicts, it means the implementation is wrong. Think twice before giving up.

SpadeA-Tang: Okay, good suggestion. I will propose another PR for it.

PR #13209, opened 2022-08-02
trace peers' availability info on leader side

ethercflow:
Issue Number: ref tikv/tikv#12876

Issue Number: Close #xxx

What's Changed:

```
trace peers' availability info on leader side
```

### Related changes

- No

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
trace peers' availability info on leader side
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2470,6 +2472,43 @@ where
...
+            unreachable!();
+        }
+        if self
+            .fsm
+            .peer
+            .region()
+            .get_peers()
+            .iter()
+            .all(|p| p.get_id() != from.get_id())
+        {
```
ethercflow, line 2503: may need remove from `applying_state` before return?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3481,6 +3529,22 @@ where
                             .raft_group
                             .raft
                             .adjust_max_inflight_msgs(peer_id, self.ctx.cfg.raft_max_inflight_msgs);
+                        // TODO: if witness -> non-witness
```
Connor1996, line 3532: This shouldn't be coupled with witness feature, no matter whether it has a witness, it needs trace apply index

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2470,6 +2472,43 @@ where
...
+        if self.fsm.peer.is_leader() {
+            unreachable!();
+        }
+        let mut msg = ExtraMessage::default();
+        msg.set_type(ExtraMessageType::MsgCollectPeerApplyIndexProgressResponse);
+        // FIXME: add a new field in ExtraMessage or use ctx?
+        msg.premerge_commit = self.fsm.peer.last_applying_idx;
+        self.fsm
+            .peer
+            .send_extra_message(msg, &mut self.ctx.trans, from);
```
Connor1996, line 2485: Send apply index with raft normal messages(heartbeat or append), no need to send it separately.

components/raftstore/src/store/fsm/peer.rs
Connor1996: e.g. pre-transfer leader check can leverage it as well.

ethercflow: Got it, thanks!

components/raftstore/src/store/peer.rs
```diff
@@ -5161,6 +5161,15 @@ where
             send_msg.set_end_key(region.get_end_key().to_vec());
         }
 
+        match msg.get_msg_type() {
+            MessageType::MsgAppend | MessageType::MsgHeartbeat => {
```
Connor1996, line 5165: Need to consider hibernate region as well. If in hibernate state and apply index is updated, send an extra message explicitly.

ethercflow: Fixed, thanks!

components/raftstore/src/store/peer.rs
```diff
@@ -729,6 +729,8 @@ where
     peer_cache: RefCell<HashMap<u64, metapb::Peer>>,
     /// Record the last instant of each peer's heartbeat response.
     pub peer_heartbeats: HashMap<u64, Instant>,
+    /// Record the applied index of each follower peer.
```
hehechen, line 732: And learner peer.

ethercflow: Sure, thanks!

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2470,6 +2488,46 @@ where
...
+        let prev_applying_idx = self
+            .fsm
+            .peer
+            .peer_applied_indices
+            .entry(from.get_id())
+            .or_insert(applying_idx);
+        *prev_applying_idx = applying_idx;
+    }
+
     fn on_extra_message(&mut self, mut msg: RaftMessage) {
```
Connor1996, line 2531: If there is an extra message, it would return directly after handling the extra message which leaves the raft message unresolved. Should change the spot of `has_extra_message` as well

ethercflow: Oh indeed. How about we modify the `on_extra_message` function signature to add a return value? 
- Return `false`, indicating only extra msg and no raft msg to process, return directly as before
- Return `true`, indicating there is also raft msg to be processed and cannot be returned directly

ethercflow: Oh indeed. How about we modify the signature of `on_extra_message` to add a return value?

- return `false`, it means that there is only extra msg without raft msg or snapshot, etc., it would return directly as before.
- return `true`, indicating that there are still raft msg or snapshot, etc. to process. 

As most `ExtraMessageType` does not need to execute the code behind `on_extra_message`.

Connor1996: we simply do not return after `on_extra_message`, and check `has_message` below

ethercflow: Sure, fixed thanks!

BusyJay:
What's the purpose of this PR?

ethercflow:
>   What's the purpose of this PR?

It'll be used to judge pending peer: whose (except witness) apply index is smaller than leader's truncated index

BusyJay:
> It'll be used to judge pending peer: whose (except witness) apply index is smaller than leader's truncated index

Why it has to be apply index? Why can't matched be used?

Connor1996:
> What's the purpose of this PR?

1. For witness -> non-witness conf-change, we need to know when the snapshot has been applied. The witness or uninitialized peer would always report apply index of 0. And once the snapshot is applied, it reports the latest apply index then PD gets to know the operator is finished.
2. It's a general feature that can benefit other features, 
 - Tiflash has a demand to detect slow apply tiflash peer
 - Pre-transfer-leader doesn't need to send a message to get the target peer's apply index

BusyJay:
>     * For witness -> non-witness conf-change, we need to know when the snapshot has been applied. The witness or uninitialized peer would always report apply index of 0. And once the snapshot is applied, it reports the latest apply index then PD gets to know the operator is finished.

Instead of querying index, I suggest to reply with the exact information that whether it's applying snapshot or whether it's in healthy state that can serve at least stale read.

>       * Tiflash has a demand to detect slow apply tiflash peer

Index less than truncated index doesn't mean it's slow. Leader is free to truncate logs whenever it wants despite other peers' apply progress. Slow apply should be monitored by time. The requirement is false.

>       * Pre-transfer-leader doesn't need to send a message to get the target peer's apply index

pre-transfer-leader doesn't just check for apply index. A message is always required.

Connor1996:
> Instead of querying index, I suggest to reply with the exact information that whether it's applying snapshot or whether it's in healthy state that can serve at least stale read.

I don't think exact information is general enough, what if we want to do flow control based on slow apply, it still needs to trace apply index progress. 

> Index less than truncated index doesn't mean it's slow. Leader is free to truncate logs whenever it wants despite other peers' apply progress. Slow apply should be monitored by time. The requirement is false.

Yes, it may be false-positive when compared to the truncated index, but it can compare to the leader's apply index.

> pre-transfer-leader doesn't just check for apply index. A message is always required.

I know, but it can be a quick return when the apply index is lagged.

BusyJay:
> what if we want to do flow control based on slow apply, it still needs to trace apply index progress.

Slow apply should not be measured by index. Instead, it should be measured by memory, CPU and disk capacity/IO. Only the follower knows whether it's overloaded and suggest leader to do a flow control. We already have such control by actively reject leader's MsgAppend.

> Yes, it may be false-positive when compared to the truncated index, but it can compare to the leader's apply index.

What's the point? They are doing different work and have different access pattern.

Connor1996:
> Slow apply should not be measured by index. Instead, it should be measured by memory, CPU and disk capacity/IO. Only the follower knows whether it's overloaded and suggest leader to do a flow control. We already have such control by actively reject leader's MsgAppend.

Not want to deep into the details of slow apply. My point is the apply index fits the witness requirement, and it also can be used by other features. So why bother to pass exact information dedicated to witness instead of a general one -- apply index.

> What's the point? They are doing different work and have different access pattern.

The point is that the slow Tiflash can be detected by the tiflash peer's apply index with leader apply index, it's not a false requirement. This issue of tiflash isn't related with time, it needs to know the index to check if the tiflash is ready. see https://pingcap.feishu.cn/docx/doxcnxQfXMgxnLV0pQqNxP9mQad

BusyJay:
> Not want to deep into the details of slow apply. My point is the apply index fits the witness requirement, and it also can be used by other features. So why bother to pass exact information dedicated to witness instead of a general one -- apply index?

"whether it's applying snapshot or whether it's in healthy state that can serve at least stale read" is not an exact information, it's a very generic information that can be used in many cases.

> The point is that the slow Tiflash can be detected by the tiflash peer's apply index with leader apply index, it's not a false requirement. This issue of tiflash isn't related with time, it needs to know the index to check if the tiflash is ready, see https://pingcap.feishu.cn/docx/doxcnxQfXMgxnLV0pQqNxP9mQad

A tiflash is ready means it can serve reads on certain timestamp. So the check if it's ready, it should check whether the safe read ts is lagging behind start ts. Checking two apply index won't solve the problem. For example, even missing 1 index can make the tiflash not ready for any read. It may even lead to more confusions. For example, 100 maybe too large for TiKV instance, but too small for TiFlash instance. How can user find a good value of pending log gap?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2130,6 +2130,26 @@ where
...
+                } else {
+                    // If the follower is in hibernate state and apply index is updated, the
+                    // `safe_ts` may have been updated, so send an extra message explicitly.
+                    if self.fsm.hibernate_state.group_state() == GroupState::Idle {
+                        let leader_id = self.fsm.peer.leader_id();
+                        let leader = self.fsm.peer.get_peer_from_cache(leader_id);
+                        if let Some(leader) = leader {
+                            let mut msg = ExtraMessage::default();
+                            msg.set_type(ExtraMessageType::MsgTracePeerAvailabilityInfo);
+                            msg.miss_data = self.fsm.peer.peer.get_is_witness();
```
Connor1996, line 2142: is witness or is uninitilized

components/raftstore/src/store/peer.rs
```diff
@@ -729,6 +729,12 @@ where
     peer_cache: RefCell<HashMap<u64, metapb::Peer>>,
     /// Record the last instant of each peer's heartbeat response.
     pub peer_heartbeats: HashMap<u64, Instant>,
+    // TODO: merge below two hashmaps?
```
Connor1996, line 732: can we merge these? maybe even with `peer_heartbeats`

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2470,6 +2490,27 @@ where
...
+        {
+            self.fsm.peer.peers_safe_ts.remove(&peer_id);
+            self.fsm.peer.peers_miss_data.remove(&peer_id);
+            return;
+        }
+        self.fsm.peer.peers_safe_ts.insert(peer_id, msg.safe_ts);
+        self.fsm.peer.peers_miss_data.insert(peer_id, msg.miss_data);
+    }
+
     fn on_extra_message(&mut self, mut msg: RaftMessage) {
```
Connor1996, line 2514: `has_extra_message` is not changed, it returns directly

Connor1996 (commented):
Please add test

components/raftstore/src/store/peer.rs
```diff
@@ -5161,6 +5169,21 @@ where
...
+            MessageType::MsgAppendResponse | MessageType::MsgHeartbeatResponse => {
+                let mut ext_msg = ExtraMessage::default();
+                ext_msg.set_type(ExtraMessageType::MsgTracePeerAvailabilityInfo);
+                ext_msg.miss_data = self.peer.get_is_witness();
+                ext_msg.safe_ts = if ext_msg.miss_data {
+                    0
+                } else {
+                    self.read_progress.safe_ts()
+                };
+                send_msg.set_extra_msg(ext_msg);
```
hehechen, line 5182: If the MsgHeartbeatResponse message has extra_msg, https://github.com/tikv/tikv/blob/6dcd0b3d66945583a133e60838533a5d12b3a487/components/raftstore/src/store/fsm/peer.rs#L2320 will return OK and not [execute the `step`](https://github.com/tikv/tikv/blob/6dcd0b3d66945583a133e60838533a5d12b3a487/components/raftstore/src/store/fsm/peer.rs#L2365).

ethercflow: Sorry fixed before, but may be lost by one force push :(

hehechen:
Why not trace safe_ts in check_leader response?

components/raftstore/src/store/peer.rs
```diff
@@ -753,6 +756,13 @@ where
     ///   target peer.
     /// - all read requests must be rejected.
     pub pending_remove: bool,
+    /// Currently it's used to indicate whether the witness -> non-witess
+    /// convertion operation is complete. The meaning of completion is that
+    /// this peer must contain the applied data, then PD can consider that
+    /// the conversion operation is complete, and can continue to schedule
+    /// other operators to prevent the existence of multiple witnesses in
+    /// the same time period.
+    pub unavailable: Arc<AtomicBool>,
```
Connor1996, line 765: why use `Arc`

ethercflow: Because I need to notify the apply thread to modify the state after the apply snapshot is successful to process the committed message correctly. Is Arc<bool> expensive? Or pass it when I pass create ApplyTask ?

Connor1996: pass it by ApplyRes

components/raftstore/src/store/peer.rs
```diff
@@ -919,7 +919,7 @@ where
     /// the conversion operation is complete, and can continue to schedule
     /// other operators to prevent the existence of multiple witnesses in
     /// the same time period.
-    pub unavailable: Arc<AtomicBool>,
+    pub unavailable: AtomicBool,
```
Connor1996, line 922: bool is enough

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3624,6 +3659,15 @@ where
             let (store_id, peer_id) = (peer.get_store_id(), peer.get_id());
             match change_type {
                 ConfChangeType::AddNode | ConfChangeType::AddLearnerNode => {
+                    if self.fsm.peer.is_leader() {
```
Connor1996, line 3662: move to L3688

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,7 +2631,34 @@ where
         self.fsm.hibernate_state.count_vote(from.get_id());
     }
 
-    fn on_extra_message(&mut self, mut msg: RaftMessage) {
+    fn on_trace_peer_availability_info(&mut self, from: &metapb::Peer, msg: &ExtraMessage) {
+        if !self.fsm.peer.is_leader() {
+            let mut resp = ExtraMessage::default();
+            resp.set_type(ExtraMessageType::MsgTracePeerAvailabilityInfo);
+            resp.miss_data = self.fsm.peer.unavailable;
```
Connor1996, line 2638: should send it actively when the snapshot is finished applying.

components/raftstore/src/store/peer.rs
```diff
@@ -888,6 +888,9 @@ where
     peer_cache: RefCell<HashMap<u64, metapb::Peer>>,
     /// Record the last instant of each peer's heartbeat response.
     pub peer_heartbeats: HashMap<u64, Instant>,
+    /// Record the data status of each follower or learner peer,
+    /// used for witness -> non-witness conversion.
```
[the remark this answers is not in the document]
ethercflow: Fixed, thanks!

components/raftstore/src/store/peer.rs
```diff
@@ -910,6 +912,13 @@ where
     ///   target peer.
     /// - all read requests must be rejected.
     pub pending_remove: bool,
+    /// Currently it's used to indicate whether the witness -> non-witess
+    /// convertion operation is complete. The meaning of completion is that
+    /// this peer must contain the applied data, then PD can consider that
+    /// the conversion operation is complete, and can continue to schedule
+    /// other operators to prevent the existence of multiple witnesses in
+    /// the same time period.
+    pub unavailable: bool,
```
Connor1996, line 921: How about renaming it to `initialized` and check by `is_initialized`, then normal peers can utilize it as well

Connor1996 (commented):
Please add a test if possible

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,7 +2631,34 @@ where
...
+            self.fsm
+                .peer
+                .send_extra_message(resp, &mut self.ctx.trans, from);
+            info!(
+                "non-witness response availability info to leader";
```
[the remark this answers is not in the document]
ethercflow: changed to generic

components/raftstore/src/store/peer.rs
ethercflow: sure

ethercflow:
> Please add a test if possible

I found it's hard to add tests without witness feature, so I implied tests in the next pr.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2487,7 +2488,10 @@ where
         }
 
         if msg.has_extra_msg() {
-            self.on_extra_message(msg);
+            self.on_extra_message(&mut msg);
+        }
+
+        if !msg.has_message() {
```
BusyJay, line 2494: Why the change?

components/raftstore/src/store/peer.rs
```diff
@@ -2580,6 +2593,30 @@ where
         true
     }
 
+    fn notify_leader_the_peer_is_available<T: Transport>(
+        &mut self,
+        ctx: &mut PollContext<EK, ER, T>,
+    ) {
+        if !self.is_initialized() {
```
BusyJay, line 2600: What if it's still applying snapshot?

components/raftstore/src/store/peer.rs
```diff
@@ -1536,7 +1547,7 @@ where
 
     #[inline]
     pub fn is_initialized(&self) -> bool {
-        self.get_store().is_initialized()
+        self.get_store().is_initialized() && self.initialized
```
BusyJay, line 1550: This is confusing. Hard to know whether use the wrapped version or peer storage version. Better change the name or add a new method.

components/raftstore/src/store/peer.rs
ethercflow: This function will only be called in `CheckApplyingSnapStaus::Success` branch. So there will be no applying snapshots?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2487,7 +2488,7 @@ where
         }
 
         if msg.has_extra_msg() {
-            self.on_extra_message(msg);
+            self.on_extra_message(&mut msg);
```
BusyJay, line 2491: Why `&mut`?

components/raftstore/src/store/peer.rs
```diff
@@ -910,6 +912,13 @@ where
     ///   target peer.
     /// - all read requests must be rejected.
     pub pending_remove: bool,
+    /// Currently it's used to indicate whether the witness -> non-witess
+    /// convertion operation is complete. The meaning of completion is that
+    /// this peer must contain the applied data, then PD can consider that
+    /// the conversion operation is complete, and can continue to schedule
+    /// other operators to prevent the existence of multiple witnesses in
+    /// the same time period.
+    pub initialized: bool,
```
BusyJay, line 921: Why not name it `miss_data`?

components/raftstore/src/store/fsm/peer.rs
ethercflow: Sorry, forgot to revert `on_extra_message`'s signature.

components/raftstore/src/store/peer.rs
```diff
@@ -1539,6 +1550,11 @@ where
         self.get_store().is_initialized()
     }
 
+    #[inline]
+    pub fn is_peer_miss_data(&self) -> bool {
```
BusyJay, line 1554:
```suggestion
    pub fn miss_data(&self) -> bool {
```

components/raftstore/src/store/peer.rs
ethercflow: Well, I wanted to be more general, but agree that miss data is more precise.

components/raftstore/src/store/peer.rs
```diff
@@ -2580,6 +2598,30 @@ where
         true
     }
 
+    fn notify_leader_the_peer_is_available<T: Transport>(
+        &mut self,
+        ctx: &mut PollContext<EK, ER, T>,
+    ) {
+        if self.is_peer_miss_data() {
```
BusyJay, line 2605: Why not just `self.miss_data`? Why add a new method?

components/raftstore/src/store/peer.rs
```diff
@@ -2580,6 +2598,30 @@ where
...
+            self.miss_data = false;
+            fail_point!("ignore notify leader the peer is available", |_| {});
+            let leader_id = self.leader_id();
+            let leader = self.get_peer_from_cache(leader_id);
+            if let Some(leader) = leader {
+                let mut msg = ExtraMessage::default();
+                msg.set_type(ExtraMessageType::MsgTracePeerAvailabilityInfo);
+                msg.miss_data = false;
+                msg.safe_ts = self.read_progress.safe_ts();
```
BusyJay, line 2614: Is this still necessary?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,6 +2628,33 @@ where
...
+            info!(
+                "peer responses availability info to leader";
+                "region_id" => self.region().get_id(),
+                "peer_id" => self.fsm.peer.peer.get_id(),
+                "leader_id" => from.id,
+            );
+            return;
+        }
+        if !msg.miss_data {
+            self.fsm.peer.peers_miss_data.remove(&from.get_id());
```
BusyJay, line 2648: I'm confused. What's the difference between "peer_id not in peers_miss_data` and `peers_miss_data[peer_id] is false`?

ethercflow: Currently, in `on_ready_change_peer`, only `witness` will be put in `peers_miss_data`: 


```rust
if self.fsm.peer.is_leader() {
        if peer.is_witness {
                self.fsm.peer.peers_miss_data.insert(peer.id, false);
        } else peer_prev_is_witness {
                self.fsm.peer.peers_miss_data.insert(peer.id, true);
                self.register_track_non_witnesses_availability_tick();
        }
}
```

So "peer_id not in peers_miss_data" means it's not a witness.

components/raftstore/src/store/peer.rs
ethercflow: No need anymore.

components/raftstore/src/store/fsm/peer.rs
BusyJay: That's even more confusing. I think for `is_witness`, it must miss data. But the code says it doesn't.

ethercflow: > That's even more confusing. I think for is_witness, it must miss data. But the code says it doesn't.
Indeed, how about

``` rust
if self.fsm.peer.is_leader() {
        if peer.is_witness {
                self.fsm.peer.peers_miss_data.insert(peer.id, true);
        } else peer_prev_is_witness {
                self.fsm.peer.peers_miss_data.insert(peer.id, false);
                self.register_track_non_witnesses_availability_tick();
        }
}
```

then collect pending peers, if it's in peers_miss_data & it's non-witness & miss_data is true, add it into pending_peers.

So "peer_id not in peers_miss_data" means a peer never miss data after inited.

BusyJay: Why it's set to false if it's witness in the past? If peers_miss_data[peer_id] is false, one should interpret it as peer_id doesn't miss any data. Then why bother checking the availability?

ethercflow: Yes it shouldn't, sorry when I describe

> then collect pending peers, if it's in peers_miss_data & it's non-witness & miss_data is true, then it into pending_peers.

I thought miss_data was still true, didn't realize it was changed to false in the else branch

BusyJay: I think what you want to do is check if all voters have data. I suggest to use a `wait_data_peer: Vec<u64>`. If a peer is non-witness, they should be added to the vector. If they have data, then it's removed from the vector.

ethercflow: Fixed, thanks!

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3674,13 +3677,6 @@ where
                     // Add this peer to peer_heartbeats.
                     self.fsm.peer.peer_heartbeats.insert(peer_id, now);
                     if self.fsm.peer.is_leader() {
-                        if peer.is_witness {
```
ethercflow: Move this to the next pr

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,6 +2628,36 @@ where
...
+            resp.wait_data = self.fsm.peer.wait_data;
+            self.fsm
+                .peer
+                .send_extra_message(resp, &mut self.ctx.trans, from);
+            info!(
+                "peer responses availability info to leader";
+                "region_id" => self.region().get_id(),
+                "peer_id" => self.fsm.peer.peer.get_id(),
+                "leader_id" => from.id,
+            );
```
tonyxuqqi, line 2644: I'm wondering if info! is too much logs. Can it be debug?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,6 +2628,36 @@ where
...
+            return;
+        }
+        if !msg.wait_data {
+            self.fsm
+                .peer
+                .wait_data_peers
+                .retain(|id| *id != from.get_id());
+            info!(
```
tonyxuqqi, line 2652: same here.

components/raftstore/src/store/peer.rs
```diff
@@ -1112,6 +1121,7 @@ where
             long_uncommitted_threshold: cfg.long_uncommitted_base_threshold.0,
             peer_cache: RefCell::new(HashMap::default()),
             peer_heartbeats: HashMap::default(),
+            wait_data_peers: Vec::default(),
```
tonyxuqqi, line 1124: Question: when is this wait_data_peers be updated? I did not see it in this PR.

ethercflow: Because the current update logic is related to the witness, it is placed in this PR：https://github.com/tikv/tikv/pull/13491/files#diff-7a4718b9cddd27a4ef7e563024d5283cf14b4ba3abf26edc33b974b87680a5fbR3684

components/raftstore/src/store/fsm/peer.rs
ethercflow: Large numbers changed to non-witness is possible, will change to debug.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5858,6 +5894,35 @@ where
...
+                    .peer
+                    .send_extra_message(msg, &mut self.ctx.trans, &peer);
+                info!(
+                    "check peer availability";
+                    "target peer id" => *peer_id,
+                );
+                continue;
+            }
+            // TODO: make sure if the path is reasonable
+            warn!(
```
BusyJay, line 5917: Is it necessary?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,6 +2628,36 @@ where
         self.fsm.hibernate_state.count_vote(from.get_id());
     }
 
+    fn on_trace_peer_availability_info(&mut self, from: &metapb::Peer, msg: &ExtraMessage) {
+        if !self.fsm.peer.is_leader() {
```
BusyJay, line 2632: What if it sent out the message when it was still leader, but got response after became follower? Then it will step into dead loop.

ethercflow: How about split `MsgTracePeerAvailabilityInfo` into a request type and a response type? When a pre-leader peer got a response type, it drops the msg.

BusyJay: It makes more sense.

components/raftstore/src/store/peer.rs
```diff
@@ -1547,7 +1547,12 @@ where
 
     #[inline]
     pub fn is_initialized(&self) -> bool {
-        self.get_store().is_initialized() && self.initialized
+        self.get_store().is_initialized()
+    }
+
+    #[inline]
+    pub fn is_peer_initialized(&self) -> bool {
```
Connor1996: still confusing

Connor1996 (approved):
LGTM

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5858,6 +5894,35 @@ where
...
+
+    fn on_check_peers_availability(&mut self) {
+        for peer_id in self.fsm.peer.wait_data_peers.iter() {
+            if let Some(peer) = self.fsm.peer.get_peer_from_cache(*peer_id) {
+                let mut msg = ExtraMessage::default();
+                msg.set_type(ExtraMessageType::MsgTracePeerAvailabilityInfo);
+                self.fsm
+                    .peer
+                    .send_extra_message(msg, &mut self.ctx.trans, &peer);
+                info!(
```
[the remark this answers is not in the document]
ethercflow: changed to debug

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,6 +2628,42 @@ where
...
+            debug!(
+                "receive peer ready info";
+                "peer_id" => self.fsm.peer.peer.get_id(),
+            );
+            return;
+        }
+        self.register_check_peers_availability_tick();
+    }
+
+    fn on_availability_response(&mut self, from: &metapb::Peer) {
```
BusyJay, line 2649: You may mix the function name.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2627,6 +2628,42 @@ where
...
+                .peer
+                .wait_data_peers
+                .retain(|id| *id != from.get_id());
+            debug!(
+                "receive peer ready info";
+                "peer_id" => self.fsm.peer.peer.get_id(),
+            );
+            return;
+        }
+        self.register_check_peers_availability_tick();
```
BusyJay, line 2646: If `wait_data_peers` is empty, no need to register it again.

BusyJay (commented):
Any test case?

components/raftstore/src/store/fsm/peer.rs
ethercflow: I think the previous if judgment guarantees this：
- If `msg.wait_data` is true, the peer should still in `wait_data_peers`, so it's not empty.
- If `msg.wait_data` is false and already remove from `wait_data_peers` through `push`, it'll return directly, so won't register it again.

components/raftstore/src/store/fsm/peer.rs
ethercflow: Oops :(

ethercflow:
> Any test case?

I've added integration tests [here](https://github.com/tikv/tikv/pull/13491) (not update yet) because  currently  there is no place for these functions to be called in this PR.

BusyJay (approved):

PR #13212, opened 2022-08-03
raftstore: reuse v2 conf change logic for v1

Connor1996:

Issue Number: Ref #12876 

What's Changed: It's a precedent code clean for #12972 

```commit-message
Reuse conf change logic for both v1 and v2
```

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -87,8 +87,8 @@ use crate::{
         peer_storage::{write_initial_apply_state, write_peer_state},
         util,
         util::{
-            admin_cmd_epoch_lookup, check_region_epoch, compare_region_epoch, is_learner,
-            ChangePeerI, ConfChangeKind, KeysInfoFormatter, LatencyInspector,
+            admin_cmd_epoch_lookup, check_region_epoch, compare_region_epoch, ChangePeerI,
+            ConfChangeKind, KeysInfoFormatter, LatencyInspector,
         },
         Config, RegionSnapshot, RegionTask,
     },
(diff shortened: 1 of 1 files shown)


ethercflow (approved):
LGTM

BusyJay:
I think making v1 use its own implementation is on purpose. So when we change the implementation of v2, we don't need to consider compatibility issue.

Connor1996:
> I think making v1 use its own implementation is on purpose. So when we change the implementation of v2, we don't need to consider compatibility issue.

PD still adds learners and delete learners by `ChangePeerV1`, and if don't reuse, I need to implement witness logic for both v1 and v2 which is verbose.

BusyJay:
> PD still adds learners and removes learners by `ChangePeerV1`

Why not use `ChangePeerV2`?

Connor1996:
> > PD still adds learners and removes learners by `ChangePeerV1`
> 
> Why not use `ChangePeerV2`?

I don't know. /cc @HunDunDM

HunDunDM:
According to [the RFC of joint-consensus](https://github.com/tikv/rfcs/blob/master/text/0054-joint-consensus.md).

> It's recommended to use ChangePeerV2 only for replacing peers for now.

It is very simple to change the logic of Add Peer and Remove Peer from using ChangePeer first to using ChangePeerV2 first in PD. In effect, extra code is added to prioritize the use of ChangePeer.

After modification, the steps of `RemoveVoter` will be changed from `RemovePeer` to `DemoteVoter + RemoveLearner`, like [pd#4444](https://github.com/tikv/pd/issues/4444).

BusyJay:
After these years of practice, I think we can use `ConfChangeV2` for most of operations. On the other hand, we can always use V2 to do witness related operations, so we don't need to touch the v1 code. There have been cases that modifying existing conf change implementation breaks compatibility, so let's be cautious.

Connor1996:
> After these years of practice, I think we can use ConfChangeV2 for most of operations. On the other hand, we can always use V2 to do witness related operations, so we don't need to touch the v1 code. There have been cases that modifying existing conf change implementation breaks compatibility, so let's be cautious.

If there is no issue for V2, let's all operations use V2, and only use V1 for cluster that is not support joint consensus

Connor1996:
use V2 for all operators https://github.com/tikv/pd/pull/5408, and close this PR to keep the v1 logic unchanged for compatibility.

PR #13215, opened 2022-08-04
*: all in camel case

### LINKED ISSUE #13216: Use camel case for struct naming

tabokie:

Issue Number: Close #13216
What's Changed:

Apparently the default clippy setting isn't doing much to detect the wrong naming. Update it and fix all issues.

```commit-message
```

### Check List 

Tests
- No code

### Release note 

```release-note
None
```


components/error_code/src/engine.rs
```diff
@@ -7,7 +7,7 @@ define_error_codes!(
     NOT_IN_RANGE => ("NotInRange", "", ""),
     PROTOBUF => ("Protobuf", "", ""),
     IO => ("IO", "", ""),
```
Connor1996, line 9: what about `IO`

components/raftstore/src/store/util.rs
```diff
@@ -692,7 +692,7 @@ fn timespec_to_u64(ts: Timespec) -> u64 {
 ///
 /// # Panics
 ///
-/// If nsec is negative or GE than 1_000_000_000(nano seconds pre second).
+/// If nsec is negative or Ge than 1_000_000_000(nano seconds pre second).
```
Connor1996, line 695: greater than or equal to

Connor1996 (commented):
How about enabling https://rust-lang.github.io/rust-clippy/master/index.html#use_self as well

components/error_code/src/engine.rs
tabokie: No, it's a constant not a type.

tabokie:
> How about enabling https://rust-lang.github.io/rust-clippy/master/index.html#use_self as well

I tried it, there're too many false positives, and a few regression IMO.
E.g.

```rust
     #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
64 | pub enum LogFormat {
   |          ^^^^^^^^^ help: use the applicable keyword: `Self`

          #[derive(Clone, Copy, Debug, PartialEq, Serialize)]
300 |     pub enum TestEnum {
    |              ^^^^^^^^ help: use the applicable keyword: `Self`

         fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
             match self {
31 |             ConfigValue::Duration(v) => write!(f, "{}ms", v),
   |             ^^^^^^^^^^^ help: use the applicable keyword: `Self`
```

Connor1996 (approved):
LGTM

sticnarf:
It may be still controversial to choose `TiDb` or `Tidb`. I find `Tikv` and `TiKv` are both used in the repo. I think we should unify them.

Personally, I prefer `Tikv` rather than `TiKV` because we usually use "tikv" as a word and if we camelize it, we will get `Tikv`.

Connor1996:
> It may be still controversial to choose `TiDb` or `Tidb`. I find `Tikv` and `TiKv` are both used in the repo. I think we should unify them.
> 
> Personally, I prefer `Tikv` rather than `TiKV` because we usually use "tikv" as a word and if we camelize it, we will get `Tikv`.

vote for `Tikv`

tabokie:
I'll change it. But it can't be checked with clippy though.

Connor1996:
> I'll change it. But it can't be checked with clippy though.

use `disallowed_names`

tabokie:
> use disallowed_names

I don't think so, it checks for exact match.

sticnarf:
> Also, kvproto still uses TiDb (`ChangeDataRequestKvApi::TiDb`).

Go has a different naming convention. It's inevitable that we'll break the rule in some languages.

components/engine_traits/src/lib.rs
```diff
@@ -191,7 +191,7 @@
 //!
 //! At the end of this phase the `engine` crate will be deleted.
 //!
-//! ## 3) "Pulling up" the generic abstractions through TiKv
+//! ## 3) "Pulling up" the generic abstractions through Tikv
```
sticnarf, line 194: It seems not a type name. `TiKV` should be used here, I think.

components/error_code/src/engine.rs
```diff
@@ -7,7 +7,7 @@ define_error_codes!(
...
-    CF_NAME => ("CFName", "", ""),
+    CF_NAME => ("CfName", "", ""),
```
sticnarf, line 10: Will the change affect external components?

tabokie: I don't know. @5kbpers

sticnarf (approved):

PR #13227, opened 2022-08-04
raftstore-v2: add ready processing

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
This PR adds the basic ready processing for v2. Note test case can't be
run for now as there is still API missing from raft engine.

Compared to v1, v2 always use async raft.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/src/batch/store.rs
```diff
@@ -143,6 +171,10 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport + 'static> PollHandler<PeerFsm<E
         };
         let mut delegate = PeerFsmDelegate::new(peer, &mut self.poll_ctx);
         delegate.handle_msgs(&mut self.peer_msg_buf);
+        delegate
+            .fsm
+            .peer_mut()
+            .handle_raft_ready(delegate.store_ctx);
```
tabokie, line 177: Why not do it inside `handle_msgs`?

components/raftstore-v2/src/batch/store.rs
```diff
@@ -182,6 +214,10 @@ struct StorePollerBuilder<EK: KvEngine, ER: RaftEngine, T> {
     engine: ER,
     tablet_factory: Arc<dyn TabletFactory<EK>>,
     trans: T,
+    router: StoreRouter<EK, ER>,
+    log_fetch_scheduler: Scheduler<RaftlogFetchTask>,
+    write_senders: Vec<Sender<WriteMsg<EK, ER>>>,
+    io_reschedule_concurrent_count: Arc<AtomicUsize>,
```
tabokie, line 220: Could we be more consistent in variable ordering? E.g. how about move cfg in the same place (`cfg`, `io_reschedule_concurrent_count`)

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -95,18 +113,103 @@ impl<EK: KvEngine, ER: RaftEngine> Fsm for PeerFsm<EK, ER> {
...
 
-impl<'a, EK: KvEngine, ER: RaftEngine, T> PeerFsmDelegate<'a, EK, ER, T> {
-    pub fn new(fsm: &'a mut PeerFsm<EK, ER>, store_ctx: &'a mut StoreContext<T>) -> Self {
+impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER, T> {
+    pub fn new(fsm: &'a mut PeerFsm<EK, ER>, store_ctx: &'a mut StoreContext<EK, ER, T>) -> Self {
         Self { fsm, store_ctx }
     }
 
-    pub fn handle_msgs(&self, peer_msgs_buf: &mut Vec<PeerMsg<EK>>) {
+    pub fn schedule_tick(&mut self, tick: PeerTick) {
```
tabokie, line 125: The names are not intuitive enough IMO (or need more documentations), to handle a tick we go through four functions:
```
- handle_tick(tick)
	- on_raft_tick
		- register_raft_tick
			- schedule_tick(tick)
```

But from the names of these steps, I can't tell the difference or their relationships.

BusyJay: handle_tick - handle any type of ticks
on_raft_tick - only handle raft tick
register_raft_tick - only register raft tick
schedule_tick - schedule any type of tick

There are many types of tick, raft tick is just one of them.

components/raftstore-v2/src/batch/store.rs
BusyJay: Because handle raft ready is not handling messages.

components/raftstore-v2/src/fsm/peer.rs
tabokie: I know, that's why I added `(tick)` in my pseudo code. The problem is the "verb", in your explanation, it seems they should all be called handle except for `schedule_tick`.

components/raftstore-v2/src/batch/store.rs
BusyJay: I don't get it. Why they should be moved together? `io_reschedule_concurrent_count` and `write_senders` are put together because they are all related to async write.

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -95,18 +113,103 @@ impl<EK: KvEngine, ER: RaftEngine> Fsm for PeerFsm<EK, ER> {
...
+            PeerTick::ReactivateMemoryLock => unimplemented!(),
+            PeerTick::ReportBuckets => unimplemented!(),
+            PeerTick::CheckLongUncommitted => unimplemented!(),
+        }
+    }
+
+    pub fn handle_msgs(&mut self, peer_msgs_buf: &mut Vec<PeerMsg<EK>>) {
         for msg in peer_msgs_buf.drain(..) {
-            // TODO: handle the messages.
+            match msg {
```
tabokie, line 195: Also, I think it makes the code cleaner to unify how we handle different tick/msg. For example, in v1, handling a message usually goes to `on_xx_msg`, but there are exceptions (like propose cmd). How about we stick to `on_xx` this time, and group their definitions together as well.

components/raftstore-v2/src/batch/store.rs
tabokie: Then these should be grouped with newlines around them, otherwise how would you expect a newcomer to read these and grasp the subtext you are trying to imply here?

components/raftstore-v2/src/batch/store.rs
tabokie: `poller::handle_normal` is defined as handling normal messages, Handling ready should be counted as part of handling a message. I think it would be better to avoid leaking details like "ready" into a message handling module `batch`.

components/raft_log_engine/src/engine.rs
```diff
@@ -563,12 +563,17 @@ impl RaftEngine for RaftLogEngine {
...
 
-    fn for_each_raft_group<E, F>(&self, _f: &mut F) -> std::result::Result<(), E>
+    fn for_each_raft_group<E, F>(&self, f: &mut F) -> std::result::Result<(), E>
     where
         F: FnMut(u64) -> std::result::Result<(), E>,
         E: From<engine_traits::Error>,
     {
-        unimplemented!()
+        for id in self.0.raft_groups() {
+            if id != STORE_REGION_ID {
```
Connor1996, line 572: name `STORE_REGION_ID` is confusing

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -95,18 +113,103 @@ impl<EK: KvEngine, ER: RaftEngine> Fsm for PeerFsm<EK, ER> {
...
+            PeerTick::CheckPeerStaleState => unimplemented!(),
+            PeerTick::EntryCacheEvict => unimplemented!(),
+            PeerTick::CheckLeaderLease => unimplemented!(),
+            PeerTick::ReactivateMemoryLock => unimplemented!(),
+            PeerTick::ReportBuckets => unimplemented!(),
+            PeerTick::CheckLongUncommitted => unimplemented!(),
+        }
+    }
+
+    pub fn handle_msgs(&mut self, peer_msgs_buf: &mut Vec<PeerMsg<EK>>) {
```
Connor1996, line 193: Does it have something different from v1, can we just refactor v1 and reuse it? Otherwise witness need to just reimplemented again in v2.

BusyJay: Yes, a lot of different. I suggest you to implement witness with good abstraction so that v2 can just call structs/functions to use it.

components/raft_log_engine/src/engine.rs
BusyJay: What's your suggestion?

components/raftstore-v2/src/batch/store.rs
BusyJay: No, handling ready is a separate operation. It can be executed after handling a batch of messages, it can also be executed after handling several batch of messages. For example, in old versions, it's called in `end` instead of `handle_normal`. The timing of processing ready is subject to change for performance evaluation.

components/raft_log_engine/src/engine.rs
Connor1996: FAKE_REGION_ID?

components/raftstore-v2/src/router/message.rs
```diff
@@ -244,6 +245,7 @@ pub enum PeerMsg<EK: KvEngine> {
     ApplyRes {
         res: ApplyTaskRes<EK::Snapshot>,
     },
+    FetchedLogs(FetchedLogs),
```
Connor1996, line 248: SignificantMsg and CasualMsg are obsolete in v2?

components/raft_log_engine/src/engine.rs
BusyJay: It doesn't represent that it will be counted as global state.

components/raftstore-v2/src/router/message.rs
BusyJay: No, they will be added back by demand.

components/raftstore-v2/tests/integrations/test_election.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+use raftstore_v2::PeerMsg;
+
+// TODO: finish test case when callback is
```
Connor1996, line 5: incomplete comment

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -42,32 +42,39 @@ impl fmt::Display for Task {
...
 
-pub struct Runner<EK, ER, R>
+#[derive(Debug)]
+pub struct FetchedLogs {
+    pub context: GetEntriesContext,
+    pub logs: Box<RaftlogFetchResult>,
+}
+
+/// A router for receiving fetched result.
+pub trait LogFetchedNotifier: Send {
```
Connor1996, line 52: Can we unify `Log` with `Entry` by the way, they are mix-used throughout the code

BusyJay: `Log` seems more straightforward. And better do it in other thread.

Connor1996 (approved):
LGTM

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -1,37 +1,50 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
+//! This module contains the peer implementation for batch system.
+//!
+//! You may get confused about the peer in fsm and peer in raft module. The
```
tabokie, line 5: Another thing, could you comment when should we put implementations in `operation/xx`?

components/raftstore-v2/src/operation/ready/async_writer.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Debug)]
+struct UnpersistedReady {
+    /// Number of ready.
+    number: u64,
+    /// Max number of following ready whose data to be persisted is empty.
+    max_empty_number: u64,
+    raft_msgs: Vec<Vec<RaftMessage>>,
+}
+
+/// A writer that handles asynchrouns writes.
```
tabokie, line 31:
```suggestion
/// A writer that handles asynchronous writes.
```

components/raft_log_engine/src/engine.rs
tonyxuqqi: how about GLOBAL_STATE_FAKE_REGION_ID

components/raftstore-v2/src/batch/store.rs
```diff
@@ -27,44 +38,58 @@ use crate::{
...
+        trans: T,
+        router: StoreRouter<EK, ER>,
+        write_senders: WriteSenders<EK, ER>,
+        logger: Logger,
+    ) -> Self {
         Self {
             logger,
             trans,
-            cfg,
+            router,
```
tonyxuqqi, line 70: should the order of fields be same as their declaration?

components/raftstore-v2/src/batch/store.rs
```diff
@@ -249,24 +289,44 @@ where
...
             self.logger.clone(),
         );
         let cfg_tracker = self.cfg.clone().tracker("raftstore".to_string());
         StorePoller::new(poll_ctx, cfg_tracker)
     }
 }
 
+/// A set of background threads that will processing offloaded work from
+/// raftstore.
+struct Workers<EK: KvEngine, ER: RaftEngine> {
```
tonyxuqqi, line 309: In the future, shall we consider to use thread pool for these background workers, instead of dedicated thread for each worker?

components/raftstore-v2/src/batch/store.rs
```diff
@@ -142,7 +167,11 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport + 'static> PollHandler<PeerFsm<E
             HandleResult::stop_at(0, false)
         };
         let mut delegate = PeerFsmDelegate::new(peer, &mut self.poll_ctx);
-        delegate.handle_msgs(&mut self.peer_msg_buf);
+        delegate.on_msgs(&mut self.peer_msg_buf);
+        delegate
+            .fsm
+            .peer_mut()
+            .handle_raft_ready(delegate.store_ctx);
```
tonyxuqqi, line 174: poll_ctx, store_ctx are essentially same object. Can we use the poll_ctx as it's widely used in the code.

components/raftstore-v2/src/operation/ready/async_writer.rs
```diff
@@ -0,0 +1,199 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
tonyxuqqi, line 1: for async_write, what's the key difference between v1 and v2? (except the difference due to new structure names)

components/raftstore-v2/src/operation/ready/async_writer.rs
```diff
@@ -0,0 +1,199 @@
...
+struct UnpersistedReady {
+    /// Number of ready.
+    number: u64,
+    /// Max number of following ready whose data to be persisted is empty.
+    max_empty_number: u64,
+    raft_msgs: Vec<Vec<RaftMessage>>,
+}
+
+/// A writer that handles asynchronous writes.
+pub struct AsyncWriter<EK: KvEngine, ER: RaftEngine> {
```
tonyxuqqi, line 32: I cannot find its counterpart of v1. Which structure in v1 has similar function?

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// Raft relies on periodic ticks to keep the state machine sync with other
+    /// peers.
+    pub fn on_raft_tick(&mut self) {
+        if self.fsm.peer_mut().tick() {
+            self.fsm.peer_mut().set_has_ready();
+        }
+        self.schedule_raft_tick();
+    }
+
+    pub fn schedule_raft_tick(&mut self) {
```
tonyxuqqi, line 48: i think this function is a bit redundant.  Just call self.schedule_tick(PeerTick::Raft) is straightforward enough.

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    #[inline]
+    fn tick(&mut self) -> bool {
+        self.raft_group_mut().tick()
+    }
+
+    /// Callback for fetching logs asynchronously.
+    pub fn on_fetched_logs(&mut self, fetched_logs: FetchedLogs) {
```
tonyxuqqi, line 60: This code essentially same as on_raft_log_fetched (except structure name). Is there anyway to reuse it?

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    /// Processing the ready of raft. A detail description of how it's handled
+    /// can be found at https://docs.rs/raft/latest/raft/#processing-the-ready-state.
+    ///
+    /// It's should be called at the end of every round of processing. Any
+    /// writes will be handled asynchronously, and be notified once writes
+    /// are persisted.
+    #[inline]
+    pub fn handle_raft_ready<T: Transport>(&mut self, ctx: &mut StoreContext<EK, ER, T>) {
```
tonyxuqqi, line 181: is this similar to handle_raft_ready_append in v1?

components/raftstore-v2/src/batch/store.rs
BusyJay: Thread pool is already used. But there are still some background work that can take too much time and starving all other jobs. So dedicated thread is used for those tasks.

components/raftstore-v2/src/batch/store.rs
BusyJay: No, they are not. There maybe some common fields, those parts can be extracted.

components/raftstore-v2/src/operation/ready/async_writer.rs
BusyJay: There will be no writes to kv engine in v2.

components/raftstore-v2/src/operation/ready/async_writer.rs
BusyJay: There is no such struct in v1. It's a reorganized of code. In v1, async write is highly clustered with the ready processing.

components/raftstore-v2/src/operation/ready/mod.rs
BusyJay: The function is small and simple, I don't think we need to do complex extraction to reuse them.

components/raftstore-v2/src/operation/ready/mod.rs
BusyJay: Yes. There is no snapshot processing for now.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -323,12 +401,47 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
...
+
+        workers.store_writers.shutdown();
+        workers.log_fetch_worker.stop();
+    }
+}
+
+#[derive(Clone)]
+pub struct StoreRouter<EK: KvEngine, ER: RaftEngine> {
+    router: BatchRouter<PeerFsm<EK, ER>, StoreFsm>,
+    logger: Logger,
```
5kbpers, line 420: Can we use a global logger instead of passing a logger everywhere?

BusyJay: Using a logger explicitly is better for following two reasons:
- Custom KVs can be added, for example, no need to write "peer_id" => peer_id, "region_id" => region_id everywhere.
- Slightly faster access, because we don't need to check for global loggers.

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+            self.entry_storage_mut()
+                .update_async_fetch_res(low, Some(logs));
+        }
+        self.raft_group_mut().on_entries_fetched(context);
+        // clean the async fetch result immediately if not used to free memory
+        self.entry_storage_mut().update_async_fetch_res(low, None);
+        self.set_has_ready();
+    }
+
+    /// Partially filled a raft message that will be sent to otheer peer.
```
SpadeA-Tang, line 79: typo: otheer -> other

components/raft_log_engine/src/engine.rs
BusyJay: I rename it to `STORE_STATE_ID` as it is used to persist store level states.

components/raftstore-v2/src/batch/store.rs
tonyxuqqi: I mean just in this function, we have two names for the same thing:
Here it's called poll_ctx.
let mut delegate = PeerFsmDelegate::new(peer, &mut self.poll_ctx);
But in PeerFsmDelegate, it's called store_ctx.

components/raftstore-v2/src/operation/ready/mod.rs
tonyxuqqi: I think this function is quite important and very likely be updated in the future. It will be better to reuse the same logic, given the fact that all the core logic is exactly same.

BusyJay: > very likely be updated in the future

Why?

tonyxuqqi (approved):

PR #13228, opened 2022-08-04
gc: support multi-rocksdb in compaction filter

### LINKED ISSUE #13131: GC with compaction filter support multi tablets
## Development Task
Currently, GC with compaction filter works globally against the global rocksdb engine. We need to make it support multi tablets.
The basic idea is to make the global gc context and compaction_filter from global to per tablet.

BornChanger:
Signed-off-by: BornChanger <dawn_catcher@126.com>

Issue Number: Close #13131 

What's Changed:

```
Associate proper region id, suffix id and optional engine factory to WriteCompactionFilterFactory and RawCompactionFilterFactory at cfs option creation time. At create_compaction_filter time, retrieve the proper rocksdb engine instead of db in the global GC_CONTEXT.
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


src/config.rs
```diff
@@ -1214,6 +1208,21 @@ impl DbConfig {
         ]
     }
 
+    pub fn set_write_compaction_filter_factory(
```
tonyxuqqi, line 1211: there's other RawCompactionFilterFactory may need similar change.

src/server/engine_factory.rs
```diff
@@ -167,6 +168,9 @@ impl KvEngineFactory {
         };
         let shared_block_cache = self.inner.block_cache.is_some();
         kv_engine.set_shared_block_cache(shared_block_cache);
+        // RocksEngine is there now, we can set its write compaction filter factory
+        DbConfig::set_write_compaction_filter_factory(kv_engine.clone(), kv_cfs_opts);
```
tonyxuqqi, line 172: I'm not sure it will work by this way. The engine is created already, and maybe the set_write_compaction_filter_factory cannot not take effect.

BornChanger:
cc @tabokie

src/server/engine_factory.rs
```diff
@@ -134,6 +134,7 @@ impl KvEngineFactory {
         tablet_path: &Path,
         region_id: u64,
         suffix: u64,
+        tablet_factory: Option<KvEngineFactoryV2>,
```
tonyxuqqi, line 137: 在v1的EngineFactory直接引用一个KvEngineFactoryV2, 感觉有点奇怪。实际上KvEngineFactoryV2是依赖v1的KvEngineFactory的。

src/config.rs
```diff
@@ -1197,17 +1204,32 @@ impl DbConfig {
         cache: &Option<Cache>,
         region_info_accessor: Option<&RegionInfoAccessor>,
         api_version: ApiVersion,
+        region_id: u64,
+        suffix_id: u64,
+        tablet_factory: Option<KvEngineFactoryV2>,
```
tonyxuqqi, line 1209: 不应该直接用v2的结构，而是用一个TabletFactory的Trait。v1的代码不应该依赖v2的代码。

src/server/gc_worker/compaction_filter.rs
```diff
@@ -197,13 +201,48 @@ impl CompactionFilterInitializer<RocksEngine> for RocksEngine {
     }
 }
 
-pub struct WriteCompactionFilterFactory;
+pub struct WriteCompactionFilterFactory {
+    region_id: u64,
+    suffix: u64,
+    tablet_factory: Option<KvEngineFactoryV2>,
```
tonyxuqqi, line 207: 这里应该是用TabletFactory这个trait。

src/server/engine_factory.rs
tonyxuqqi: 而且不应该在factory里传另外一个factory，这个struct本身就能create tablet的。

src/server/engine_factory.rs
```diff
@@ -185,11 +191,11 @@ impl KvEngineFactory {
...
-        let _kv_cfs_opts = self.inner.rocksdb_config.build_cf_opts(
-            &self.inner.block_cache,
-            self.inner.region_info_accessor.as_ref(),
-            self.inner.api_version,
-        );
+        // let _kv_cfs_opts = self.inner.rocksdb_config.build_cf_opts(
+        //    &self.inner.block_cache,
+        //    self.inner.region_info_accessor.as_ref(),
+        //    self.inner.api_version,
+        //);
```
tonyxuqqi, line 198: why delete？

src/server/gc_worker/compaction_filter.rs
tonyxuqqi: 还有一个办法，我觉得可能更加清晰。把tablet_factory放GC_CONTEXT里，如果有global db就走v1，有tablet_factory就走v2。

src/config.rs
BornChanger: > 不应该直接用v2的结构，而是用一个TabletFactory的Trait。v1的代码不应该依赖v2的代码。

嗯,重构了

src/server/engine_factory.rs
BornChanger: image

src/server/engine_factory.rs
BornChanger: 重构了

BornChanger:
cc @SpadeA-Tang

tonyxuqqi (approved):

src/server/engine_factory.rs
```diff
@@ -185,11 +189,11 @@ impl KvEngineFactory {
         if let Some(filter) = self.create_raftstore_compaction_listener() {
             kv_db_opts.add_event_listener(filter);
         }
-        let _kv_cfs_opts = self.inner.rocksdb_config.build_cf_opts(
-            &self.inner.block_cache,
-            self.inner.region_info_accessor.as_ref(),
-            self.inner.api_version,
-        );
+        // let _kv_cfs_opts = self.inner.rocksdb_config.build_cf_opts(
```
tabokie, line 192: Don't comment it out, just update the `build_cf_opts` call.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -193,17 +199,47 @@ impl CompactionFilterInitializer<RocksEngine> for RocksEngine {
...
             #[cfg(any(test, feature = "failpoints"))]
             callbacks_on_drop: vec![],
+            tablet_factory,
         });
     }
 }
 
-pub struct WriteCompactionFilterFactory;
+pub trait RocksdbCompactionFilterFactory: CompactionFilterFactory {
+    fn get_rocksdb(&self, region_id: u64, suffix: u64) -> engine_traits::Result<RocksEngine> {
```
tabokie, line 208: Just move these code into `create_compaction_filter`. A trait seems unnecessary.

src/server/gc_worker/compaction_filter.rs
```diff
@@ -60,6 +62,7 @@ pub struct GcContext {
     pub(crate) region_info_provider: Arc<dyn RegionInfoProvider + 'static>,
     #[cfg(any(test, feature = "failpoints"))]
     callbacks_on_drop: Vec<Arc<dyn Fn(&WriteCompactionFilter) + Send + Sync>>,
+    pub(crate) tablet_factory: Option<Arc<dyn TabletFactory<RocksEngine> + Send + Sync>>,
```
tabokie, line 65: I presume there's no need to keep the `db` field anymore? The single-rocksdb case can also be handled with `KvEngineFactory`.

src/server/gc_worker/gc_manager.rs
```diff
@@ -51,18 +52,27 @@ pub struct AutoGcConfig<S: GcSafePointProvider, R: RegionInfoProvider> {
     /// This will be called when a round of GC has finished and goes back to
     /// idle state. This field is for test purpose.
     pub post_a_round_of_gc: Option<Box<dyn Fn() + Send>>,
+
+    /// Used to save tablet factory of TiKV instance
+    pub tablet_factory: Option<Arc<dyn TabletFactory<RocksEngine> + Send + Sync>>,
```
tabokie, line 57: Drop the `Option`, there should always be an available table factory.

src/server/gc_worker/compaction_filter.rs
BornChanger: The only purpose is to eliminate dup code for create_compaction_filter for both raw and write compaction filter factories.

tabokie: Then a function is enough, because you didn't use `self` here.

src/server/gc_worker/gc_manager.rs
BornChanger: I have to keep it as `Option` to support legacy unit test cases, where it's hard to construct a factory.

src/server/gc_worker/compaction_filter.rs
BornChanger: The `db` field represents the global one. And it's needed to support legacy unit cases.

tabokie: You can use `TestTabletFactory`.

src/server/gc_worker/compaction_filter.rs
BornChanger: Good catch~ Rust compiler doesn't warm me.

src/server/gc_worker/gc_manager.rs
SpadeA-Tang: Actually, there are many similar cases where using Option can ease the adaptation of legacy unit tests when making modifications for v2, and I am also interested in whether this is a good method for this.

src/server/engine_factory.rs
BornChanger: fixed.

tabokie (commented):
`GcWorker` needs to be updated as well, and you need to introduce tablet_id to `GcTask`. It would be better if you can add a multi-rocksdb test case to verify all this.

src/server/gc_worker/compaction_filter.rs
BornChanger: fixed.

src/server/gc_worker/gc_manager.rs
BornChanger: Since tablet_factory is definitely not None at runtime, I removed the Option.

src/storage/kv/test_engine_builder.rs
```diff
@@ -67,23 +67,30 @@ impl TestEngineBuilder {
     }
 
     /// Build a `RocksEngine`.
-    pub fn build(self) -> Result<RocksEngine> {
+    pub fn build(self, region_id: u64, suffix_id: u64) -> Result<RocksEngine> {
```
tabokie, line 70: For builder style, try not to append args in `build`. Instead, you can add this:
```rust
struct TestEngineBuilder {
  ..
  region_id: u64,
  suffix_id: u64,
}

impl TestEngineBuilder {
  pub fn tablet_id(mut self, region_id: u64, suffix_id: u64) -> Self {
    self.region_id = region_id;
    self.suffix_id = suffix_id;
  }
}
```

This way you don't need to change all the legacy tests and make this PR too hard to review.

BornChanger: Sorry, I don't get the point here.  Builder can build multiple engines with different region_id and suffxi_id, which shares the same root path, default limiter, api_version, etc.  It seems natural especially in multi-rocksdb scenario. And how can your suggested approach save change to legacy cases ?

tabokie: All the `build(0, 0)` can be reverted.

BornChanger: Do you mean give region_id, suffix_id default value 0 for structure TestEngineBuilder?

tabokie: Yes.

src/storage/kv/test_engine_builder.rs
```diff
@@ -66,6 +70,13 @@ impl TestEngineBuilder {
         self
     }
 
+    /// Called at multi-rocksdb scenarios
+    pub fn set_tablet_ids(mut self, region_id: u64, suffix_id: u64) -> Self {
```
tabokie, line 74:
```suggestion
    pub fn tablet_ids(mut self, region_id: u64, suffix_id: u64) -> Self {
```

tabokie (commented):
> GcWorker needs to be updated as well, and you need to introduce tablet_id to GcTask. It would be better if you can add a multi-rocksdb test case to verify all this.

This comment is not addressed, right now the compaction filter still doesn't support multi-rocks yet. Or do you plan to do it in another PR?

BornChanger:
> > GcWorker needs to be updated as well, and you need to introduce tablet_id to GcTask. It would be better if you can add a multi-rocksdb test case to verify all this.
> 
> This comment is not addressed, right now the compaction filter still doesn't support multi-rocks yet. Or do you plan to do it in another PR?

Newly added case test_compaction_filter_for_multi_rocksdb is to test the compaction filter support for multi-rocks.

components/engine_test/src/lib.rs
```diff
@@ -115,31 +116,33 @@ pub mod kv {
...
         root_db: Arc<Mutex<Option<KvTestEngine>>>,
     }
 
     impl TestTabletFactory {
-        pub fn new(
-            root_path: &Path,
-            db_opt: DbOptions,
-            cf_opts: Vec<(&'static str, KvTestCfOptions)>,
-        ) -> Self {
+        pub fn new(root_path: &Path, db_opt: DbOptions) -> Self {
```
tabokie, line 123: Why change?

components/engine_test/src/lib.rs
```diff
@@ -232,16 +235,17 @@ pub mod kv {
...
             Self {
-                inner: TestTabletFactory::new(root_path, db_opt, cf_opts),
+                inner: TestTabletFactory::new(root_path, db_opt),
                 registry: Arc::default(),
             }
         }
+
+        pub fn register_tablet(&self, region_id: u64, suffix: u64, tablet: KvTestEngine) {
```
tabokie, line 245: You should make `create_tablet` accept tablet ids, and use `open_tablet` in the test.

src/server/gc_worker/gc_worker.rs
```diff
@@ -2684,6 +2730,28 @@ mod tests {
...
+            cfg.ratio_threshold = 0.5;
+            cfg.enable_compaction_filter = true;
+            GcWorkerConfigManager(Arc::new(VersionTrack::new(cfg)))
+        };
+
+        let worker_builder = WorkerBuilder::new("gc-worker").pending_capacity(GC_MAX_PENDING_TASKS);
+        let worker = worker_builder.create().lazy_build("gc-worker");
+        let gc_scheduler = worker.scheduler();
+
+        new_gc_context(
```
tabokie, line 2745: Why not use `init_compaction_filter`?

src/server/gc_worker/gc_worker.rs
```diff
@@ -2684,6 +2730,28 @@ mod tests {
...
+        let cfg_tracker = {
+            let mut cfg = GcConfig::default();
+            // let ratio_threshold less than 1
+            cfg.ratio_threshold = 0.5;
+            cfg.enable_compaction_filter = true;
+            GcWorkerConfigManager(Arc::new(VersionTrack::new(cfg)))
+        };
+
+        let worker_builder = WorkerBuilder::new("gc-worker").pending_capacity(GC_MAX_PENDING_TASKS);
+        let worker = worker_builder.create().lazy_build("gc-worker");
```
tabokie, line 2742: I'm having a hard time understand why this test works at all. The `worker` here will be destroyed at the end of this function.

src/server/gc_worker/gc_worker.rs
```diff
@@ -2614,16 +2637,13 @@ mod tests {
         assert!(ks.is_empty());
     }
 
-    // setup engine and prepare some data:
-    //  three regions:
-    //  region 1: includes ("k00", "value-00") to ("k09", "value-09")
-    //  region 2: includes ("k10", "value-10") to ("k19", "value-19")
-    //  region 3: includes ("k20", "value-20") to ("k29", "value-29")
+    #[allow(non_snake_case)]
```
tabokie, line 2640: Which part violates this rule?

src/config.rs
```diff
@@ -675,6 +675,8 @@ impl DefaultCfConfig {
         cache: &Option<Cache>,
         region_info_accessor: Option<&RegionInfoAccessor>,
         api_version: ApiVersion,
+        region_id: u64,
```
tabokie, line 678: Looks a bit ugly, but I can't think of a better way either. cc @BusyJay

components/engine_test/src/lib.rs
BornChanger: When it comes to multi-rocksdb, cf_opts is bound to individual rocksdb, instead of tablet factory.

src/server/gc_worker/gc_worker.rs
BornChanger: image

components/engine_test/src/lib.rs
BornChanger: This is just a helper function for multi_gc_engine_setup(), where engine_rocks::util::new_engine_opt is called to create tablet instead of open_tablet. All those new functions are for test mocking purpose only.

BornChanger: I tried but only found TestTabletFactory::create_tablet() can't access structures like BlockCacheConfig, DbConfig etc due to the dependency issue.

src/server/gc_worker/gc_worker.rs
tabokie: Which toolchain are you using? This should be fixed in https://github.com/rust-lang/rust/pull/89473 last year.

components/engine_test/src/lib.rs
tabokie: You need to manually create the compaction filter here (instead of reusing the `build_opt` in tikv crate).

src/server/gc_worker/gc_worker.rs
BornChanger: image

tabokie: Oh, I see it now. You need to add an extra `_ => (*cf, RocksCfOptions::default()),`, right now rust mistakenly treats `CF_RAFT` as a new identifier instead of the constant.

BornChanger: image
i remember I tried your suggestion already, and above the other warning. So, It should be the rustc issue anyway.

components/engine_test/src/lib.rs
BornChanger: I don't see your proposal can make code more readable...

src/server/gc_worker/gc_worker.rs
tabokie: Need to `use engine_traits::CF_RAFT`.

components/engine_test/src/lib.rs
tabokie: It's not a problem of readability, it's not correct. The code should be able to build against different engine implementations, e.g. `cargo test --package tikv --features test-engines-panic --no-default-features`. To accomplish that you cannot associate a fixed type of engine with `TestTabletFactory`.

@SpadeA-Tang I run this locally and it appears you introduced a lot of changes that broke this build.

tabokie:
@BornChanger Please don't resolve discussions that are in fact still open. PR author can only resolve a review comment when the suggestion has been taken in its exact form. Most of the time we just wait for the reviewer to resolve them.

components/engine_test/src/lib.rs
BornChanger: @tabokie the problem should origins from @tonyxuqqi commit 03b44b5e219fb795b20ebb9367b66b2adf3800ac

tabokie: Fixing it in #13585.

src/server/gc_worker/gc_worker.rs
BornChanger: > Why not use `init_compaction_filter`?

fixed.

src/server/gc_worker/gc_worker.rs
BornChanger: > I'm having a hard time understand why this test works at all. The `worker` here will be destroyed at the end of this function.

reworked.

src/config.rs
BusyJay: I suggest to move compaction filter generation to the `KvEngineFactory` instead of configuration.

BornChanger: ok. I took your suggestion.

BornChanger:
/idc-jenkins-ci-tikv/clippy-linux-arm64

BornChanger:
/run-idc-jenkins-ci-tikv/clippy-linux-arm64

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -766,6 +766,29 @@ impl Clone for MockRegionInfoProvider {
 }
 
 impl RegionInfoProvider for MockRegionInfoProvider {
+    fn seek_region(&self, from: &[u8], callback: SeekRegionCallback) -> Result<()> {
```
tabokie, line 769: Let's not change function ordering. This makes it hard to track code change history.

components/server/src/server.rs
```diff
@@ -1070,7 +1070,13 @@ where
                 self.concurrency_manager.clone(),
             )
             .unwrap_or_else(|e| fatal!("gc worker failed to observe lock apply: {}", e));
-        if let Err(e) = gc_worker.start_auto_gc(auto_gc_config, safe_point) {
+        // table_factory of TiKvServer won't be None
```
tabokie, line 1073: If you are unwrapping the `Option` below, there's no need to assert it again.

components/engine_test/src/lib.rs
```diff
@@ -184,6 +178,19 @@ pub mod kv {
...
+        fn create_shared_db(&self) -> Result<KvTestEngine> {
+            let tablet_path = self.tablet_path(0, 0);
+            let tablet = self.create_tablet(&tablet_path)?;
+            let mut root_db = self.root_db.lock().unwrap();
+            root_db.replace(tablet.clone());
+            Ok(tablet)
+        }
+
+        #[inline]
+        fn destroy_tablet(&self, _id: u64, _suffix: u64) -> engine_traits::Result<()> {
```
tabokie, line 190: ditto, restore the function order.

tests/failpoints/cases/test_gc_worker.rs
```diff
@@ -405,12 +410,17 @@ fn test_orphan_versions_from_compaction_filter() {
...
+    cluster: &Cluster<ServerCluster>,
+    store_id: u64,
+    safe_point: Arc<AtomicU64>,
+) {
     #[derive(Clone)]
     struct MockSafePointProvider;
     impl GcSafePointProvider for MockSafePointProvider {
         fn get_safe_point(&self) -> GcWorkerResult<TimeStamp> {
-            Ok(TimeStamp::from(0))
+            Ok(TimeStamp::from(100))
```
tabokie, line 423: Why change?

tests/failpoints/cases/test_gc_worker.rs
```diff
@@ -438,10 +448,35 @@ fn init_compaction_filter(cluster: &Cluster<ServerCluster>, store_id: u64) {
 
     let sim = cluster.sim.rl();
     let gc_worker = sim.get_gc_worker(store_id);
+    gc_worker.set_feature_gate_verion("5.0.0").unwrap();
+    let kv_engine = cluster.get_engine(store_id);
+    // Building a tablet factory
+    let ops = DbOptions::default();
+    let path = Builder::new()
+        .prefix("test_gc_keys_with_region_info_provider")
+        .tempdir()
```
tabokie, line 457: All the usage of tempdir is incorrect. Because when the `Tempdir` object goes out of scope, the directory will be automatically deleted, and the engine data will be lost too.

tests/failpoints/cases/test_table_properties.rs
```diff
@@ -167,6 +168,7 @@ fn do_gc(
 }
 
 #[test]
+#[cfg(not(feature = "test-engines-panic"))]
```
tabokie, line 171: Why not `test-engine-kv-rocksdb`. The problem with `test-engines-panic` is, the build will still fail if we uses `--features test-engine-kv-panic,test-engine-raft-rocksdb` or `--features test-engines-another-kind-of-engine`.

components/server/src/server.rs
BornChanger: removed the redundant check.

tests/failpoints/cases/test_gc_worker.rs
BornChanger: Not necessary now. removed.

tests/failpoints/cases/test_table_properties.rs
BornChanger: It seems I can't express that way.  When ENABLE_FEATURES is not set by default, config predicate feature=test-engine-kv-rocksdb won't be satisfied either.

tests/failpoints/cases/test_gc_worker.rs
BornChanger: What you said is right.  I have corrected other occurrence for path construction except for prepare_gc(). At prepare_gc() invocation time, engine is ready (passed as one parameter), so the temp dir dropping is ok. It's just a skeleton for factory construction, while the factory won't really create any rocksdb later and is only for storing the registry of <<region_id, suffix_id>, rocksdb> in GC_Context.

tonyxuqqi (approved):

tabokie:
Superseded by https://github.com/tikv/tikv/pull/13982.

PR #13230, opened 2022-08-05
raftstore: simplify v1 propose path

tabokie:

Issue Number: Ref #12842 

What's Changed:

- Add comments for batch system code
- Simplify `propose_batch_raft_command(bool)`
  - It's considered a bad practice to control behavior with a boolean unless necessary
  - There's only one code path where `propose_batch_raft_command(false)` is called, refactor it into `on_loop_finished` for better readability
- Refactor `check_proposal_normal_with_disk_usage`
  - Keep the related logic clustered in one function, avoid passing data with references
- Split `schedule(.., bool)` into two functions for the same reason as `propose_batch_raft_command`

```commit-message
```

### Check List 

Tests
- Unit test

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/peer.rs
```diff
@@ -631,31 +631,32 @@ where
...
-                    let req_size = cmd.request.compute_size();
                     if self.ctx.cfg.cmd_batch
-                        && self.fsm.batch_req_builder.can_batch(&self.ctx.cfg, &cmd.request, req_size)
                         // Avoid to merge requests with different `DiskFullOpt`s into one,
                         // so that normal writes can be rejected when proposing if the
                         // store's disk is full.
                         && ((self.ctx.self_disk_usage == DiskUsage::Normal
                             && !self.fsm.peer.disk_full_peers.majority())
                             || cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull)
+                        && let size = cmd.request.compute_size()
```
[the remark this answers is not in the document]
tabokie: This is just a minor optimization to short-circuit the `compute_size()` if disk_full_opt doesn't allow batching.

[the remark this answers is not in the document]
Connor1996: I think it should be optimized by compiler already

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -688,53 +689,52 @@ where
...
-        {
-            return;
+    #[inline]
+    #[allow(unused_mut, clippy::redundant_closure_call)]
+    fn on_loop_finished(&mut self) {
+        let mut ready_concurrency = self.ctx.cfg.cmd_batch_concurrent_ready_max_count;
+        let mut should_propose = self.ctx.sync_write_worker.is_some()
+            || ready_concurrency == 0
+            || self.fsm.peer.unpersisted_ready_len() < ready_concurrency;
+        #[cfg(feature = "failpoints")]
```
Connor1996, line 702: why? `fail_point!` would check it innerly.

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
tabokie: Maybe so, but I think your discomfort is a habit-thing and it's actually idiomatic to write it like this. `let Some(x) =` and `let x =` both are pattern matching (and both aren't boolean expr), they shouldn't be treated differently.

components/raftstore/src/store/fsm/peer.rs
tabokie: It's just a habit of mine. Even though the empty function call will be optimized away, but I feel uncomfortable to leave a trail in production code.

components/batch-system/benches/batch-system.rs
```diff
@@ -48,9 +48,9 @@ fn bench_spawn_many(c: &mut Criterion) {
     system.shutdown();
 }
 
-/// Bench how it performs if two hot FSMs are shown up at the same time.
+/// Bench how it performs if two hot Fsms are shown up at the same time.
```
BusyJay, line 51: Why change?

components/batch-system/src/batch.rs
```diff
@@ -479,15 +482,15 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
...
             }
             to_skip_end.clear();
             self.handler.end(&mut batch.normals);
 
             // Because release use `swap_remove` internally, so using pop here
-            // to remove the correct FSM.
+            // to remove the correct Fsm.
             while let Some(r) = reschedule_fsms.pop() {
-                batch.schedule(&self.router, r, false);
+                batch.schedule(&self.router, r, false /* inplace */);
```
BusyJay, line 493: In fact, no matter using vscode or intellij, the arg name will be shown alongside with the value, the comment is a bit verbose.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -631,31 +631,32 @@ where
...
-                    let req_size = cmd.request.compute_size();
                     if self.ctx.cfg.cmd_batch
-                        && self.fsm.batch_req_builder.can_batch(&self.ctx.cfg, &cmd.request, req_size)
                         // Avoid to merge requests with different `DiskFullOpt`s into one,
                         // so that normal writes can be rejected when proposing if the
                         // store's disk is full.
                         && ((self.ctx.self_disk_usage == DiskUsage::Normal
                             && !self.fsm.peer.disk_full_peers.majority())
                             || cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull)
+                        && let req_size = cmd.request.compute_size()
```
BusyJay, line 647: Weird style.

components/batch-system/src/batch.rs
tabokie: I don't think it's a good reason at all. GitHub review for one doesn't have plugins. And it doesn't make sense to intentionally "punish" those contributors (I for instance don't use them) without the necessary plugins.

components/raftstore/src/store/fsm/peer.rs
tabokie: I searched around and found several similar cases in rust-clippy:
https://github.com/rust-lang/rust-clippy/blob/f7e2cb4470f5288e6657166fb0fcdf4b7043517f/clippy_lints/src/dereference.rs#L755
https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/manual_non_exhaustive.rs#L130

BusyJay: It doesn't justify it's a weird style. Putting pattern matching that actually just does assignment to bool field doesn't make sense to me.

tabokie: As I explained in https://github.com/tikv/tikv/pull/13230#discussion_r940047751, I presume you treat this particular `let` differently because it can never fail. But pattern matching is never a boolean expression. Thinking it is is a misperception.

components/batch-system/src/batch.rs
BusyJay: How about removing the bool parameter?

components/batch-system/benches/batch-system.rs
tabokie: Just to unify the old "FSM" and "fsm". I think Fsm is slightly more good-looking than FSM. But I can switch it back if you have a stronger preference.

components/raftstore/src/store/fsm/peer.rs
BusyJay: > you treat this particular let differently because it can never fail.

I treat it differently because it "just does assignment".

components/raftstore/src/store/fsm/peer.rs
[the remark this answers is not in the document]
Connor1996: TIL

components/raftstore/src/store/fsm/peer.rs
Connor1996: it's not neat and of no use.

components/batch-system/benches/batch-system.rs
BusyJay: FSM itself is an abbreviation. `Fsm` is a name in code. `FSM` explains the meaning itself while `Fsm` does not and needs extra explanation on what it is.

components/raftstore/src/store/fsm/peer.rs
tabokie: I don't think it's no use though. My brain can filter out `#[cfg(feature = "failpoints")]` and `fail_point!(..)` for me. But it has to pause a second for the nested closure call.

BUT, considering it has become a common pattern in TiKV's codebase, I'll keep it consistent.

components/raftstore/src/store/fsm/peer.rs
tabokie: Fine, I'll keep it simple to reduce friction. But for the record I didn't expect so much push-back from this line..

components/batch-system/src/batch.rs
```diff
@@ -478,17 +483,19 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
...
             to_skip_end.clear();
             self.handler.end(&mut batch.normals);
 
-            // Because release use `swap_remove` internally, so using pop here
-            // to remove the correct FSM.
-            while let Some(r) = reschedule_fsms.pop() {
-                batch.schedule(&self.router, r, false);
+            // Iterate larger index first, so that `unsafe_reclaim` won't affect other FSMs
+            // in the list.
+            for index in reschedule_fsms.iter().rev() {
```
BusyJay, line 494: Why the change?

components/batch-system/src/batch.rs
```diff
@@ -232,12 +226,19 @@ impl<N: Fsm, C: Fsm> Batch<N, C> {
...
             f.policy.take();
             self.normals[index] = res;
-        } else if !inplace {
+        }
+    }
+
+    /// Reclaims the slot storage if there is no FSM located at `index`. It will
+    /// alter the positions of other FSMs with index larger than `index`.
+    #[inline]
+    pub fn unsafe_reclaim(&mut self, index: usize) {
```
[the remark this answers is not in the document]
tabokie: It is unsafe in the sense that it can change other FSM's location index, and the index stability is a primary assumption of this container. To mark it unsafe urges caller taking responsibility to protect the invariants.

Even though it's not "Rust unsafe", but we used this word in other cases too (`unsafe_destroy_range` or `unsafe_recover`).

components/batch-system/src/batch.rs
```diff
@@ -478,17 +483,19 @@ impl<N: Fsm, C: Fsm, Handler: PollHandler<N, C>> Poller<N, C, Handler> {
...
+                batch.schedule(&self.router, *index);
             }
             to_skip_end.clear();
             self.handler.end(&mut batch.normals);
 
-            // Because release use `swap_remove` internally, so using pop here
-            // to remove the correct FSM.
-            while let Some(r) = reschedule_fsms.pop() {
-                batch.schedule(&self.router, r, false);
+            // Iterate larger index first, so that `unsafe_reclaim` won't affect other FSMs
```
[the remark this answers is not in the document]
tabokie: In my version, reader only need to understand the semantics of `unsafe_reclaim`, which is commented. In the older version, readers are forced to read the implementation of `schedule` to understand the usage of `swap_remove` and how that makes it "incorrect" to remove FSM.

components/batch-system/src/batch.rs
tabokie: To be consistent with the handling of `to_skip_end`. And `rev` conveys the message (which is iterating order) better than `pop`. I personally don't keep a record of whether `Vec::pop` pops from the beginning or the end.

components/batch-system/src/batch.rs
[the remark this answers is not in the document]
BusyJay: unsafe_destroy_range breaks consistency and unsafe_recover break durability, so they are prefixed with `unsafe_`. For this function, we can name it `swap_reclaim` to indicate it will change order.

components/batch-system/src/mailbox.rs
```diff
@@ -13,12 +13,23 @@ use crate::fsm::{Fsm, FsmScheduler, FsmState};
...
-/// Every mailbox should have one and only one owner, who will receive all
-/// messages sent to this mailbox.
+/// A mailbox holds an FSM owner, and the sending end of a channel to send
+/// messages to that owner.
 ///
-/// When a message is sent to a mailbox, its owner will be checked whether it's
-/// idle. An idle owner will be scheduled via `FsmScheduler` immediately, which
-/// will drive the fsm to poll for messages.
+/// Mailbox is designed to be shared. Multiple producers can send messages to
+/// the same mailbox concurrently.
```
BusyJay, line 20: Actually it should be obvious in Rust as it's `Sync`.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -117,32 +125,29 @@ impl<EK: KvEngine, ER: RaftEngine, T: Transport + 'static> PollHandler<PeerFsm<E
...
-        let received_cnt = store.recv(&mut self.store_msg_buf);
+        let received_cnt = fsm.recv(&mut self.store_msg_buf);
         let expected_msg_count = if received_cnt == self.messages_per_tick() {
             None
         } else {
             Some(0)
         };
-        let mut delegate = StoreFsmDelegate::new(store, &mut self.poll_ctx);
-        delegate.handle_msgs(&mut self.store_msg_buf);
+        let mut fsm = StoreFsmDelegate::new(fsm, &mut self.poll_ctx);
```
BusyJay, line 136: Why name it `fsm`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -631,31 +631,32 @@ where
...
                     let req_size = cmd.request.compute_size();
                     if self.ctx.cfg.cmd_batch
-                        && self.fsm.batch_req_builder.can_batch(&self.ctx.cfg, &cmd.request, req_size)
                         // Avoid to merge requests with different `DiskFullOpt`s into one,
                         // so that normal writes can be rejected when proposing if the
                         // store's disk is full.
                         && ((self.ctx.self_disk_usage == DiskUsage::Normal
                             && !self.fsm.peer.disk_full_peers.majority())
                             || cmd.extra_opts.disk_full_opt == DiskFullOpt::NotAllowedOnFull)
+                        && self.fsm.batch_req_builder.can_batch(&self.ctx.cfg, &cmd.request, req_size)
```
BusyJay, line 648: Why move the order?

components/raftstore/src/store/peer.rs
```diff
@@ -3421,53 +3421,11 @@ where
...
-                        }
-                    }
-                    let errmsg = format!(
-                        "propose failed: tikv disk full, cmd diskFullOpt={:?}, leader diskUsage={:?}",
-                        disk_full_opt, ctx.self_disk_usage
-                    );
-                    Err(Error::DiskFull(stores, errmsg))
+                    disk_full_opt = DiskFullOpt::AllowedOnAlmostFull;
                 }
+                self.check_normal_proposal_with_disk_full_opt(ctx, disk_full_opt)
```
BusyJay, line 3427: Why not:
```
self.check_normal_proposal_with_disk_full_opt(ctx, disk_full_opt)?;
self.propose_normal(ctx, req)
```

tests/failpoints/cases/test_cmd_epoch_checker.rs
```diff
@@ -111,12 +111,12 @@ fn test_reject_proposal_during_region_split() {
         .unwrap_err();
 
     // Try to put a key.
-    let propose_batch_raft_command_fp = "propose_batch_raft_command";
+    let force_delay_propose_batch_raft_command_fp = "force_delay_propose_batch_raft_command";
```
BusyJay, line 114: Where is the failpoint defined?

components/raftstore/src/store/peer.rs
```diff
@@ -4837,56 +4795,74 @@ where
...
-        if self.disk_full_peers.is_empty() {
-            return true;
-        }
-
-        for peer in self.get_store().region().get_peers() {
-            let (peer_id, store_id) = (peer.get_id(), peer.get_store_id());
-            if self.disk_full_peers.peers.get(&peer_id).is_some() {
-                disk_full_stores.push(store_id);
+            // Try to transfer leader to a node with disk usage normal to keep write
+            // availability not downback. If majority node is disk full, to transfer leader
```
BusyJay, line 4812: What is "downback"

components/raftstore/src/store/peer.rs
```diff
@@ -4837,56 +4795,74 @@ where
...
+                    })
+                    .cloned();
+                if let Some(p) = target_peer {
+                    debug!(
+                        "try to transfer leader because of current leader disk full";
+                        "region_id" => self.region_id,
+                        "peer_id" => self.peer.get_id(),
+                        "target_peer_id" => p.get_id(),
+                    );
+                    self.pre_transfer_leader(&p);
```
BusyJay, line 4834: Maybe out of scope, this should be checked by PD.

components/raftstore/src/store/fsm/peer.rs
tabokie: The original intention is to short-circuit the `compute_size` if disk usage check fails (with `&& let`). But I guess it doesn't matter now if we just trust compiler to do all the reorders.

components/raftstore/src/store/peer.rs
tabokie: It shouldn't escape the function, the error is handled below.

tests/failpoints/cases/test_cmd_epoch_checker.rs
tabokie: Hmm, apparently a typo, but it didn't break the test.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -688,53 +689,51 @@ where
...
+            || self.fsm.peer.unpersisted_ready_len() < ready_concurrency;
+        (|| {
+            fail_point!("force_delay_propose_batch_raft_command", |_| {
+                should_propose = false
+            })
+        })();
+        // Propose batch request which may be still waiting for more raft-command
+        if should_propose {
+            self.propose_pending_batch_raft_command();
+        } else if let Some(cmd) = self.fsm.batch_req_builder.request.take() {
```
BusyJay, line 710: This should be executed when it has not been executed and there is proposed callback.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -688,53 +689,51 @@ where
...
+    #[inline]
+    #[allow(unused_mut, clippy::redundant_closure_call)]
+    fn on_loop_finished(&mut self) {
+        let mut ready_concurrency = self.ctx.cfg.cmd_batch_concurrent_ready_max_count;
+        let mut should_propose = self.ctx.sync_write_worker.is_some()
+            || ready_concurrency == 0
+            || self.fsm.peer.unpersisted_ready_len() < ready_concurrency;
+        (|| {
+            fail_point!("force_delay_propose_batch_raft_command", |_| {
+                should_propose = false
```
BusyJay, line 704: I think it should be `self.ctx.sync_write_worker.is_some()`.

components/raftstore/src/store/fsm/peer.rs
tabokie: I don't know what you mean. But I fixed some inconsistency with old behavior.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -688,53 +689,57 @@ where
                 }
             }
         }
+        self.on_loop_finished();
+    }
+
+    #[inline]
+    #[allow(unused_mut, clippy::redundant_closure_call)]
```
BusyJay, line 696: Why `unused_mut`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -688,53 +689,57 @@ where
...
+    fn on_loop_finished(&mut self) {
+        let mut ready_concurrency = self.ctx.cfg.cmd_batch_concurrent_ready_max_count;
+        let mut should_propose = self.ctx.sync_write_worker.is_some()
+            || ready_concurrency == 0
+            || self.fsm.peer.unpersisted_ready_len() < ready_concurrency;
+        (|| {
+            fail_point!(
+                "force_delay_propose_batch_raft_command",
+                self.ctx.sync_write_worker.is_some(),
```
BusyJay, line 705: In the past, it's `!force`. So here should be `is_none`.

BusyJay (approved):
Rest LGTM

components/raftstore/src/store/fsm/peer.rs
tabokie: The `mut should_propose` is only used in failpoint.

BusyJay: You can get around the problem by putting failpoint to or chain.
```
let should_propose = self.ctx.sync_write_worker.is_some()
            || (ready_concurrency != 0
                && self.fsm.peer.unpersisted_ready_len() < ready_concurrency
                && (|| {
                       fail_point!(
                "force_delay_propose_batch_raft_command",
                |_| false
            );
            true
        })();
```

tabokie: Seems more difficult to read. So many brackets.

BusyJay: Or
```
let delay_by_fp = (|| {
    fail_point!(
            "force_delay_propose_batch_raft_command",
           self.ctx.sync_write_worker.is_none()
            |_| true
    );
    false
})();
should_propose &= !delay_by_fp;
```

Connor1996 (approved):
LGTM

PR #13234, opened 2022-08-05
server: support customized addr/status_addr

CalvinNeo:

Issue Number: ref #12849 

What's Changed:

```commit-message
Support self-defined addr/status_addr
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


src/server/node.rs
```diff
@@ -47,19 +48,19 @@ const CHECK_CLUSTER_BOOTSTRAPPED_RETRY_INTERVAL: Duration = Duration::from_secs(
 
 /// Creates a new storage engine which is backed by the Raft consensus
 /// protocol.
-pub fn create_raft_storage<S, EK, R: FlowStatsReporter, F: KvFormat>(
+pub fn create_raft_storage<S, EK, R: FlowStatsReporter, F: KvFormat, LM: LockManagerTrait>(
```
BusyJay, line 51: Bad import. Instead of using `xxxTrait`, `module_path::xxx` is preferred.

src/server/status_server/mod.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
-mod profile;
+/// Provides profilers for TiKV.
+pub mod profile;
```
BusyJay, line 4: Only export used type instead.

BusyJay (commented):
Rest LGTM

src/server/node.rs
```diff
@@ -119,17 +120,25 @@ where
             Some(s) => s,
         };
         store.set_id(INVALID_ID);
-        if cfg.advertise_addr.is_empty() {
-            store.set_address(cfg.addr.clone());
-            store.set_peer_address(cfg.addr.clone());
-        } else {
-            store.set_address(cfg.advertise_addr.clone());
-            store.set_peer_address(cfg.advertise_addr.clone());
+        if store.get_address() == "" {
```
BusyJay, line 123: Should use `is_empty` instead.

CalvinNeo: fixed

src/server/status_server/mod.rs
CalvinNeo: fixed

src/server/node.rs
CalvinNeo: fixed

src/server/status_server/mod.rs
```diff
@@ -39,6 +40,10 @@ use openssl::{
     x509::X509,
 };
 use pin_project::pin_project;
+pub use profile::{
+    activate_heap_profile, deactivate_heap_profile, jeprof_heap_profile, list_heap_profiles,
+    read_file, start_one_cpu_profile, start_one_heap_profile,
```
BusyJay, line 45: Be cautious with `read_file`. Better not use it.

BusyJay (approved):

src/server/status_server/mod.rs
CalvinNeo: TiFlash implements its own status_server by reusing code from TiKV and adding some specific path to TiKV's status_server.
We expose `read_file` because it is used in TiKV's status_server

breezewish (approved):
I feel something may be wrong when TiFlash proxy is going to use the status_server directly. It may not be an ideal abstraction for TiFlash, since status_server contains too many TiKV logic. We can keep an eye on this and get it right in future.

CalvinNeo:
> I feel something may be wrong when TiFlash proxy is going to use the status_server directly. It may not be an ideal abstraction for TiFlash, since status_server contains too many TiKV logic. We can keep an eye on this and get it right in future.

TiFlash Proxy has its own status server aka https://github.com/pingcap/tidb-engine-ext/blob/raftstore-proxy-6.2/components/proxy_server/src/status_server.rs

We just reuse mod profile

PR #13244, opened 2022-08-08
raftstore: Use thread-local metrics for local read 

Connor1996:

Issue Number: Ref #12876 

What's Changed: Use thread-local metrics for local read to simplify code, and theoretically it reduces memory usage and improves cache locality. 

```commit-message
Use tls local read metrics
```

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/worker/metrics.rs
+++ b/components/raftstore/src/store/worker/metrics.rs
@@ -1,8 +1,11 @@
 // Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
 
+use std::{cell::RefCell, time::Duration};
+
 use lazy_static::lazy_static;
-use prometheus::*;
+use prometheus::{local::LocalIntCounter, *};
 use prometheus_static_metric::*;
+use tikv_util::time::Instant;
 
 make_auto_flush_static_metric! {
     pub label_enum SnapType {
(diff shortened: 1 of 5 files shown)


BusyJay:
> it reduces memory usage and improves cache locality

I don't think there is a difference. And in fact, accessing local metrics is way faster than thread local metrics. Though simplicity is preferred if there is no notice-able performance difference.

components/raftstore/src/store/worker/metrics.rs
```diff
@@ -44,14 +47,54 @@ make_static_metric! {
...
     }
 
-    pub struct ReadRejectCounter : IntCounter {
-       "reason" => RejectReason
+    pub struct LocalReadRejectCounter : LocalIntCounter {
+       "reason" => RejectReason,
     }
 }
 
+pub struct LocalReadMetrics {
```
SpadeA-Tang, line 59: Why some fields of `ReadMetrics` are not in `LocalReadMetrics`?

Connor1996: they are sub-fields in `reject_reason`

Connor1996:
```rust 
 use test::Bencher;
    #[bench]
    fn bench_tls_inc(b: &mut Bencher) {
        b.iter(|| {
            for _ in 0..1000 {
                TLS_LOCAL_READ_METRICS.with(|m| { m.borrow_mut().local_executed_requests.inc() });
            }
        });
    }

    #[bench]
    fn bench_local_inc(b: &mut Bencher) {
        b.iter(|| {
            let mut count = 0;
            for _ in 0..1000 {
                count += 1;
                test::black_box(count);
            }
        });
    }
```

The micro bench above shows tls does have overhead: 
```
test store::worker::read::tests::bench_local_inc                              ... bench:         456 ns/iter (+/- 78)
test store::worker::read::tests::bench_tls_inc                                  ... bench:       3,089 ns/iter (+/- 199)
```

Though the metrics account for only a small proportion, when it comes to the extent of `get_snapshot`, the difference is little.
```
test store::worker::read::tests::bench_get_snapshot(master)                ... bench:          25 ns/iter (+/- 3)
test store::worker::read::tests::bench_get_snapshot(tls)                        ... bench:          28 ns/iter (+/- 2)
```

Furthermore, considering the whole read/write process, the difference is not notable.
Here is a sysbench read_only workload comparison(ignore the jitter, it was caused by other task) 
6e0f1893-604a-422d-816c-7bc6f8225316

@BusyJay

BusyJay:
Do you have a full benchmark? Note recently I spend a lot of time in finding and reverting PRs that can cause regression in 6.2 #13243. Finding the regression ahead of time is better than hunting before releasing new version.

Connor1996:
> Do you have a full benchmark that CPU load is high? Note recently I spend a lot of time in finding and reverting PRs that can cause regression in 6.2 #13243. Finding the regression ahead of time is better than hunting before releasing new version.

It is OPS of getting the snapshot that does matter. So here I choose the read-only workload to get a higher OPS of getting the snapshot. What do you mean by `a full benchmark`? Is a read-only workload that exhausts 8c CPU okay with you?

BusyJay:
> What do you mean by `a full benchmark`? Is a read-only workload that exhausts 8c CPU okay with you?

Yes, it's OK. Just to make sure whether CPU is the bottleneck.

Connor1996:
Run two rounds of oltp_read_only for tls and master version, the result is not stable under heave CPU usage. Anyway, tls version never shows regression comparing to the master version

image

components/raftstore/src/store/worker/read.rs
```diff
@@ -718,21 +714,20 @@ where
...
         // Check region epoch.
         if util::check_region_epoch(req, &delegate.region, false).is_err() {
-            self.metrics.rejected_by_epoch += 1;
+            TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().reject_reason.epoch.inc());
             // Stale epoch, redirect it to raftstore to get the latest region.
             debug!("rejected by epoch not match"; "tag" => &delegate.tag);
             return Ok(None);
         }
 
         let mut inspector = Inspector {
```
BusyJay, line 729: Seems can be simplified.

BusyJay (approved):

components/raftstore/src/store/worker/read.rs
Connor1996: Can't. Trait `Inspector` needs `DerefMut`, but `CachedReadDelegate` only implements `Deref` as there is an `Arc` in it. To remove the `Arc`, a lot of changes are involved. Won't do it in this PR.

ethercflow (approved):
LGTM

tabokie (approved):

PR #13245, opened 2022-08-08
raftstore-v2: add read write channel

BusyJay:
Issue Number: Ref #12842 

What's Changed:
```commit-message
v2 uses channel instead of callbacks for proposals, so async/await has first class
support. We can further reduce allocations by introducing channel pool.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore/src/store/msg.rs
```diff
@@ -88,19 +87,20 @@ pub enum Callback<S: Snapshot> {
...
+    Read(Box<dyn FnOnce(ReadResponse<S>) + Send>),
     /// Write callback.
     Write {
-        cb: WriteCallback,
+        cb: Box<dyn FnOnce(WriteResponse) + Send>,
         /// `proposed_cb` is called after a request is proposed to the raft
-        /// group successfully. It's used to notify the caller to move on early
-        /// because it's very likely the request will be applied to the
-        /// raftstore.
+        /// group successfully. It's used to notify the caller to move
```
Connor1996, line 95: the comment shouldn't be changed

tonyxuqqi (commented):
some comments. Not finished yet.

components/raftstore-v2/src/router/message.rs
```diff
@@ -207,64 +92,88 @@ impl StoreTick {
     }
 }
 
-/// Raft command is the command that is expected to be proposed by the
-/// leader of the target raft group.
-#[derive(Debug)]
-pub struct RaftCommand<S: Snapshot> {
+/// Command that can be handled by raftstore.
+pub struct RaftCommand {
```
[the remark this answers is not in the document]
BusyJay: There are only 3 fields, we can always extract a new struct to unify the entry if necessary.

components/raftstore/src/store/msg.rs
```diff
@@ -73,9 +73,8 @@ where
     }
 }
 
-pub type ReadCallback<S> = Box<dyn FnOnce(ReadResponse<S>) + Send>;
```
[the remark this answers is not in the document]
BusyJay: Because `ReadCallback` is a trait now. `WriteCallbackExt` is a bad name, it's not an extension.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+    event: AtomicU64,
+    res: UnsafeCell<Option<Res>>,
+    // Waker can be changed, need to use `AtomicWaker` to guarantee no data race.
+    waker: AtomicWaker,
+}
+
+unsafe impl<Res: Send> Send for EventCore<Res> {}
+
+const PAYLOAD_EVENT: u64 = 0;
+const CANCEL_EVENT: u64 = 31;
```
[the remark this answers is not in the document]
BusyJay: The value will not be passed by user. It's an implementation details.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+}
+
+#[inline]
+const fn set_bit_of(event: u64) -> u64 {
+    1 << (event * 2 + 1)
+}
+
+impl<Res> EventCore<Res> {
+    #[inline]
+    pub fn notify_event(&self, event: u64) {
```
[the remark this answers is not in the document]
BusyJay: User will never call `notify_event` directly. And they should not use `EventCore` directly. Instead, `EventCore` should be wrapped with meaningful event definition just like what `ReadChannel` and `WriteChannel` do.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+//! Variants of channels for `Msg`.
+//!  - `Read`: a channel for read only requests including `StatusRequest`,
+//!    `GetRequest` and `SnapRequest`
+//!  - `Write`: a channel for write only requests including `AdminRequest`
+//!    `PutRequest`, `DeleteRequest` and `DeleteRangeRequest`.
+//!
+//! Prefer channel over callback because:
```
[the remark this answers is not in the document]
BusyJay: Only in tests.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+            match event.compare_exchange_weak(
+                e,
+                e | subscribed_bit,
+                Ordering::AcqRel,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => return Poll::Pending,
+                Err(v) => e = v,
+            };
+            if let Some(b) = check_bit(e, set_bit) {
```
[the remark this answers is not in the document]
BusyJay: This line can be reached only when `compare_exchange_weak` fails. It fails means some other thread is updating `event` successfully. Unless there is some thread keep updating `event` successfully, can this become infinite loop.

components/raftstore-v2/src/router/response_channel.rs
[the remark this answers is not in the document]
tonyxuqqi: > User will never call `notify_event` directly. And they should not use `EventCore` directly. Instead, `EventCore` should be wrapped with meaningful event definition just like what `ReadChannel` and `WriteChannel` do.

IC. If that, maybe the "pub" can be changed to pub(self) or pub(super)?

components/raftstore/src/store/msg.rs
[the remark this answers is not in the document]
tonyxuqqi: Then rename the old ReadCallback with some other name. It's inconsistent that ExtCallback exists, but read/write callback does not have the alias, just because you use them somewhere else.

components/raftstore-v2/src/router/message.rs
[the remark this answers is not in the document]
tonyxuqqi: > There are only 3 fields, we can always extract a new struct to unify the entry if necessary.

For example, LocalReader is the new code Chenjie refactored to make it sharable. But now with this RaftCommand, that part needs additional work. 
" fn redirect(&mut self, mut cmd: RaftCommand<E::Snapshot>) {"
This is just one example, there're many other places. The accumulated extra work is noticeable, even if it looks like a  trivial change. 
I think without reusing these structures, it's simply too much work to reuse v1's code.

components/raftstore-v2/src/router/message.rs
```diff
@@ -207,64 +92,88 @@ impl StoreTick {
...
+impl RaftCommand {
+    pub fn new(request: RaftCmdRequest) -> Self {
         RaftCommand {
+            send_time: Instant::now(),
             request,
+        }
+    }
+}
+
+pub struct ReadCommand {
```
tonyxuqqi, line 110: So the original code is to use the same RaftCommand for both read and write. 
Now the RaftCommand does not have the response channel, and instead response channel is move to the upper layer ReadCommand/WriteCommand. 

From the PR, I don't see there's need to differentiate Read/Write Command. What's the change in the future that would need to separate read and write command? To me the real source of truth is always in RaftCmdRequest. So what if WriteCommand has the read RaftCmdRequest?  I think the original code is fine.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+
+/// A struct allows to watch and notify specific events.
+///
+/// There are two different events: state and payload. Obviously, state events
+/// have no payload. At most 30 states can be defined. There can be only one
+/// type of payload.
+struct EventCore<Res> {
+    /// Every event will have two bits.
+    /// - 0b00 means the event is not fired and not watched.
+    /// - 0b01 means the event is fired and not watched.
```
tonyxuqqi, line 45: define these 0bxx as more readable const?

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+
+unsafe impl Send for WriteSubscriber {}
+unsafe impl Sync for WriteSubscriber {}
+
+pub struct WriteChannel {
+    core: ManuallyDrop<Arc<EventCore<RaftCmdResponse>>>,
+}
+
+impl WriteChannel {
+    const PROPOSED_EVENT: u64 = 2;
```
tonyxuqqi, line 235: all the events should be defined in the same place, to avoid duplicated value.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    // TODO: support executing hooks inside setting result.
+    #[inline]
+    fn set_result(mut self, res: RaftCmdResponse) {
+        self.core.set_result(res);
+        unsafe {
+            ManuallyDrop::drop(&mut self.core);
+        }
+        mem::forget(self);
```
tonyxuqqi, line 297: though self.core is handled by ManuallyDrop, will the self be leaked forever?

components/raftstore-v2/src/router/response_channel.rs
tonyxuqqi: I think it will be better to define the event enum so that we can index all kinds of events easily in one place and it's safer. 
Also the logic of bit operation can be wrapped into that enum. Instead of using subscribed_bit_of and set_bit_of in notify_event, we can use more user friendly name such as fire(), is_subscribed(). Even though the notify_event is not called directly. It's still useful to make code more readable.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,448 @@
...
+    type Output = bool;
+
+    #[inline]
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let event = &self.core.event;
+        let mut e = event.load(Ordering::Relaxed);
+        let set_bit = set_bit_of(self.event);
+        if let Some(b) = check_bit(e, set_bit) {
+            return Poll::Ready(b);
+        }
```
tonyxuqqi, line 143: if the u64 event is enum,  set_bit_of can be simply renamed as check_fired(&self).
now the usage of this is weird---first i need to atomic get the event value e, then get its set bit, then call check_bit(). It can be wrapped into one func with very clear semantic. 
Same logic repeats in WaitResult

components/raftstore-v2/src/router/message.rs
BusyJay: Read and write are always separated inside raftstore. If they are mixed in one command, the command will be rejected. It's the same as V1. Mixing read and write in the public interface adds more unnecessary encapsulation.

components/raftstore-v2/src/router/response_channel.rs
BusyJay: How do you apply atomic operation on enum?

components/raftstore-v2/src/router/response_channel.rs
BusyJay: Different channel has different event types. `ReadChannel` doesn't use the event of `WriteChannel`.

components/raftstore-v2/src/router/response_channel.rs
BusyJay: All allocations and drop methods are called. There is no leak.

components/raftstore-v2/src/router/response_channel.rs
tonyxuqqi: So can the event be defined as 0, 31 or  > 32 here?  By only looking at WriteChannel itself, we have no idea about that.

BusyJay: I will add docs describing the expected range. They are different events for different structs, and will never be mixed used together. It's meaningless to put them together.

components/raftstore-v2/src/router/message.rs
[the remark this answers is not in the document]
BusyJay: Reusing v1's code is not a goal. It's a nice to have. Having a correct, easy to maintain and performant new version is the goal.

components/raftstore-v2/src/router/response_channel.rs
BusyJay: `check_bit` is also used in other places. Wrapping them into a function just to avoid repeating bit operations, which is very easy to make mistake.

components/raftstore-v2/src/router/response_channel.rs
tonyxuqqi: Probably need some validation check in public APIS for event value.

components/raftstore-v2/src/router/message.rs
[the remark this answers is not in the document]
tonyxuqqi: > Maintaining two version of the code are exactly opposite to the easy maintenance goal.  And here I don't see the change has anything to do with performance. 
To me, this change itself is OK, but in our current situation, they're not necessary and will add up overall implementation and maintenance cost. 

What's more, even though right now the V2's RaftCommand is simpler than V1. For example, it does not have RaftCmdExtraOpts field. But RaftCmdExtraOpts exists for the reason--and disk full feature needs it. We will finally need it in v2 too. So I think eventually v2 will likely be very similar than v1.

components/raftstore-v2/src/router/response_channel.rs
BusyJay: Event value is never public.

components/raftstore-v2/src/router/message.rs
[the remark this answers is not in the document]
BusyJay: Disk full implementation is a bit verbose to me. In v2, it can be implemented by flow control instead of hijacking every messages.

tonyxuqqi (approved):

components/raftstore-v2/src/router/message.rs
```diff
@@ -208,65 +94,92 @@ impl StoreTick {
...
+        Self {
+            cmd: RaftCommand::new(request),
+            ch,
+        }
+    }
+}
+
+pub struct WriteCommand {
+    pub cmd: RaftCommand,
+    pub ch: WriteChannel,
```
5kbpers, line 129: Prefer to name it `WriteResponseChannel` or some name to represent it used for sending responses, not requests.

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -0,0 +1,449 @@
...
+/// There are two different events: state and payload. Obviously, state events
+/// have no payload. At most 30 states can be defined. There can be only one
+/// type of payload.
+struct EventCore<Res> {
+    /// Every event will have two bits.
+    /// - 0b00 means the event is not fired and not subscribed.
+    /// - 0b01 means the event is fired and not subscribed.
+    /// - 0b10 means the event is not fired and subscribed.
+    /// - 0b11 means the event is fired and subscribed.
+    event: AtomicU64,
```
5kbpers, line 48: So the number of user-defined events should be in [1, 30]? Better to add a comment for this.

5kbpers (approved):
LGTM

PR #13253, opened 2022-08-10
engine: toggle purge worker with `need_manual_purge`

tabokie:

Issue Number: Ref #11119

What's Changed:

Do not create a useless worker if the engine doesn't require it.
```commit-message
None
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


Connor1996 (approved):
LGTM

components/engine_traits/src/raft_engine.rs
```diff
@@ -115,9 +115,16 @@ pub trait RaftEngine: RaftEngineReadOnly + PerfContextExt + Clone + Sync + Send
...
 
+    fn need_manual_purge(&self) -> bool {
+        false
+    }
+
     /// Purge expired logs files and return a set of Raft group ids
     /// which needs to be compacted ASAP.
-    fn purge_expired_files(&self) -> Result<Vec<u64>>;
+    fn manual_purge(&self) -> Result<Vec<u64>> {
+        debug_assert!(false, "missing `manual_purge` implementation");
```
BusyJay, line 125: Bad style. Better leave it unimplemented.

tabokie: The intention is to only panic in debug build, because this issue (call `manual_purge` when `need_manual_purge` returns false, or haven't implemented `manual_purge` when it returns true) is not a critical one.

BusyJay: I understand the intention, but it's not a big deal to force the implementer to think whether it needs to do purge.

BusyJay (approved):

PR #13254, opened 2022-08-10
storage: precheck whether the peer is leader when acquiring latches failed

### LINKED ISSUE #12966: one az is isolated from other az networks, and the service is unavailable for 4 minutes
./tikv-server -V
 TiKV 
Release Version:   6.1.0
Edition:           Community
Git Commit Hash:   080d086832ae5ce2495352dccaf8df5d40f30687
Git Commit Branch: heads/refs/tags/v6.1.0
UTC Build Time:    2022-06-10 11:22:39
Enable Features:   jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure

8core、32g

### Steps to reproduce
inject one az is isolated from other az networks

### What did you expect?
the service is available

### What did happened?
the service is unavailable for 4 minutes
image

more info see https://pingcap.feishu.cn/wiki/wikcnzK33Ck6q1BQldXOnMUMvPc

cosven:
Issue Number: Close #12966

What's Changed:

```commit-message
When a tikv is isolated from other tikv instances, some requests will be 
blocked in raftstore and the corresponding latches are not released. 
Following requests which require the latches will receive ServerIsBusy error
and keep retrying. However, In such case, peers on the tikv are not leader
anymore. The client is supposed to receive NotLeader error immediately.

This commit introduces fail fast mode to scheduler. When a request 
fails to acquire any latch, scheduler checks if the peer is still leader.
If it still the leader, schedule the request as usual, fail fast otherwise.
```

### Related changes

### Check List 

Tests 

- [x] Unit test
  - [x] latch 
  - [x] storage
- [x] Integration test
  - [x] raftkv
  - Check issue #12966

Now
image

### Release note 

```release-note
Fix the problem that QPS may drop to zero for several mintues when a tikv is partitioned.
```


components/raftstore/src/store/fsm/store.rs
```diff
@@ -120,6 +120,8 @@ pub struct StoreMeta {
     pub region_ranges: BTreeMap<Vec<u8>, u64>,
     /// region_id -> region
     pub regions: HashMap<u64, Region>,
+    /// region_id -> LocalLeaderInfo
+    pub region_leaders: HashMap<u64, LocalLeaderInfo>,
```
tonyxuqqi, line 124: what's LocalLeaderInfo for? Is the bool not enough?

cosven: > what's LocalLeaderInfo for? Is the bool not enough?

A bool is enough for the issue we met, but it would be better if we can put the leader info is in NotLeader error. 
```rust
#[error("peer is not leader for region {0}, leader may {1:?}")]
    NotLeader(u64, Option<metapb::Peer>),
```

> Also this StoreMeta is shared by many code

I've struggled with this before, the StoreMeta is shared by many code currently. 

However, after reading the related code, I think some fields are suitable for storing in StoreMeta and some are not. Actually, I think the following fields are suitable, and so does the `region_leaders` info.
https://github.com/tikv/tikv/blob/6b3ca45c8e828d5ef8b36492791237ea9d62cfca/components/raftstore/src/store/fsm/store.rs#L118-L123
And maybe the following are not.
https://github.com/tikv/tikv/blob/6b3ca45c8e828d5ef8b36492791237ea9d62cfca/components/raftstore/src/store/fsm/store.rs#L151

-------------------

I think a better way is to remove these unsuitable fields instead of move suitable fields like `region_leaders` into another struct. Otherwise, we will have another struct just like StoreMeta, and each time StoreMeta is updated, we need to update the new struct.

cosven: After some investigation, I do think that we just need a hashset to store these region leader ids. And the implementation can be simplified. StoreMeta is not needed :)

cosven: After some investigation, I do think that we just need a hashset to store these region leader ids. And the implementation can be simplified. StoreMeta is not needed :)

cosven: done

BusyJay:
Can `RegionInfoAccessor` be used instead?

components/tikv_kv/src/rocksdb_engine.rs
```diff
@@ -209,6 +209,18 @@ impl Engine for RocksEngine {
         write_modifies(&self.engines.kv, modifies)
     }
 
+    fn precheck_write_with_ctx(&self, _ctx: &Context) -> Result<()> {
+        if self.not_leader.load(Ordering::SeqCst) {
+            let not_leader = {
+                let mut header = kvproto::errorpb::Error::default();
+                header.mut_not_leader().set_region_id(100);
```
tonyxuqqi, line 216: Why it's 100?

components/raftstore/src/store/peer.rs
```diff
@@ -2202,6 +2207,16 @@ where
...
+
+        // Update region leader info stored in poll context.
+        {
+            let mut region_leaders = ctx.region_leaders.write().unwrap();
+            if self.peer_id() == leader_id {
+                region_leaders.insert(self.region_id);
+            } else {
+                region_leaders.remove(&self.region_id);
+            }
+        }
```
tonyxuqqi, line 2219: is there any test code for this part of the change?

src/storage/txn/scheduler.rs
```diff
@@ -463,6 +440,106 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+            .inc();
+
+        let cmd_ctx = cmd.ctx().clone();
+        let mut task_slot = self.inner.get_task_slot(cid);
+        let tctx = task_slot.entry(cid).or_insert_with(|| {
+            self.inner
+                .new_task_context(Task::new(cid, tracker, cmd), callback)
+        });
+
+        if fail_fast {
```
sticnarf, line 503: I prefer to acquire latches like before, but when it fails to get all latches, spawn a background job to run the precheck (like the deadline one).

When the precheck fails, `try_own` the task, take the callback, and do the early return. In that way, I think the code will be simpler and the performance might also be better.

src/storage/txn/scheduler.rs
```diff
@@ -426,6 +439,7 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
             .get(priority_tag)
             .inc();
 
+        let cmd_ctx = cmd.ctx().clone();
```
sticnarf, line 442: I think it's better to clone it after it fails to acquire latches to avoid unnecessary clones.

sticnarf (approved):

components/raftstore/src/router.rs
```diff
@@ -125,6 +129,12 @@ where
     fn release_snapshot_cache(&self);
 }
 
+/// A checker do some checking before routing raft write requests.
+pub trait WritePreChecker: Send + Clone {
```
BusyJay, line 133: Why not make `RegionInfoAccessor` a field of scheduler directly? `Coprocessor` is supposed to cross abstraction layers.

cosven: Currently, scheduler only communicates with `Engine`. The engine can be RaftKV and it can also be RocksEngine, BTreeEngine or others. However, RegionInfoAccessor only works with the RaftKV engine. So, I think it would be better to bind RegionInfoAccessor with RaftKV.

cosven:
Manual test results
Red (using the latest commit): https://github.com/cosven/tikv/tree/758743d364e507726370358a151dafcc27371c63
Blue: https://github.com/cosven/tikv/tree/8feb1962e70a7d8205a487f3fa81bb9a1da9e606

We can see the system can recover immediately.

image

tonyxuqqi (approved):

components/raftstore/src/router.rs
BusyJay: As I said, `RegionInfoAccessor` is part of coprocessor, it can be used in scheduler.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -585,6 +592,11 @@ pub struct RegionInfoAccessor {
     // https://github.com/tikv/tikv/issues/9044
     worker: Worker,
     scheduler: Scheduler<RegionInfoQuery>,
+
+    /// Region leader ids set on the store.
+    ///
+    /// Others can access this info directly, such as ServerRaftStoreRouter.
+    pub region_leaders: Arc<RwLock<HashSet<u64>>>,
```
BusyJay, line 599: Bad access level.

BusyJay (commented):
Rest LGTM

components/raftstore/src/coprocessor/region_info_accessor.rs
cosven: fixed

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -620,6 +620,11 @@ impl RegionInfoAccessor {
         }
     }
 
+    /// Get a set of region leader ids.
+    pub fn get_region_leaders(&self) -> Arc<RwLock<HashSet<u64>>> {
```
BusyJay, line 624: Getter should not have `get_` prefix.

cosven: fixed

PR #13257, opened 2022-08-11
make read quota limiter work for coprocessor as well

### LINKED ISSUE #13256: QuotaLimiter: coprocessor read traffic should be controlled by foreground quota limiter's read bytes
6.0 or later

### Steps to reproduce
set tikv config: quota.foreground_read_bytes: 5MB
run sysbench read-only workload

### What did you expect?
The coprocessor's read response size is close to 5MB/s per tikv
### What did happened?
The coprocessor's read response size is much larger than 5MB/s

tonyxuqqi:

Issue Number: Close #13256

What's Changed:

```commit-message
The coprocessor's read bytes are not calculated by foreground quota limiter. 
```

### Related changes

### Check List 

Tests 

- Unit test
- Manual test 

Side effects
- None

### Release note 

```release-note
None
```


components/tidb_query_executors/src/runner.rs
```diff
@@ -471,6 +471,9 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
                     &mut ctx,
                 )?
             };
+            if chunk.has_rows_data() {
```
BornChanger, line 474: Though I prefer to use following approach.  ForwardScanner maintains the statistics (via the Cursor interface), but it's lost before returning to internal_handle_request() in the test.  This patch is ok to me at present. 
image

tests/integrations/coprocessor/test_select.rs
```diff
@@ -76,6 +77,7 @@ fn test_select() {
         let result_encoded = datum::encode_value(&mut EvalContext::default(), &row).unwrap();
         assert_eq!(result_encoded, &*expected_encoded);
     }
+    assert!(limiter.total_read_bytes_consumed(true) > 0); // the consume_sample is called due to read bytes quota
```
BornChanger, line 80: why not give a solid number?

components/tidb_query_executors/src/runner.rs
tonyxuqqi: note here the limiter is foreground limiter---that is to limits actual response size back to TiDB, not the IO limiter.

tests/integrations/coprocessor/test_select.rs
tonyxuqqi: Here the logic is to verify if the read bytes based limiter is called. We don't know the actual response size here and so no way to get accurate wait time. 
For actual accurate wait time, it's covered by unit test.

BornChanger (approved):

ethercflow (approved):
LGTM

5kbpers (approved):

Issue #13269, opened 2022-08-11
raftstore: follower replication for write-flow


LintianShi:
## Develop Task
1. A new mechanism in Raft algorithm which allows a follower to replicate log entries or snapshots to other followers and learners. This feature is mainly aimed to reduce cross-AZ (short for Availability Zones) traffic on cloud. RFC: https://github.com/tikv/rfcs/pull/98
1.1. Definition of new message type `MsgGroupBroadcast` (in raft-rs).
1.2. Handler for new message type  (in raft-rs).
1.3. Merge `MsgAppend` and replace it with `MsgGroupBroadcast` when peer sends raft messages.
3. Maintaining AZ information in peer from PD.

PR #13270, opened 2022-08-11
[WIP]raft store: follower replication for write-flow

LintianShi:

Issue Number: ref #13269

What's Changed:

```commit-message
* Record and maintain availability zone information in Peer.
* Definition new PeerMsg that updates AZ information in Peer, when TiKV server resolves status from PD.
* Merge MsgAppend into MsgGroupBroadcast in the function that builds raft messages.
```

### Related changes

### Check List 

Tests 

- No code

Side effects

- Performance regression
    - The peers receiving log entries from the agent follower have more latency than leader election.
    - If the agent follower is down or partitioned, the leader may not replicate data to other peers in this AZ in a short. This situation will recover quickly. 

### Release note 

```release-note
This feature introduces a new mechanism in Raft algorithm which allows a follower to replicate log entries or snapshots to other followers and learners. This feature is mainly aimed to reduce cross-AZ (short for Availability Zones) traffic in data synchronization between TiKV instances and TiFlash instances via Raft consensus algorithm.

In the classical Raft protocol, a follower or learner can only receive new raft log entries and snapshots from the leader. It inevitably leads to a large amount of cross-AZ traffic, because the leader has to replicate data to every follower and learner even in different AZ.
With follower replication, the leader only needs to replicate data one follower in each AZ, and this follower will replicate data to other followers and learners in the same AZ, which avoids a lot of cross-AZ traffic.
```

### DIFF
--- a/components/pd_client/src/util.rs
+++ b/components/pd_client/src/util.rs
@@ -816,7 +816,6 @@ pub fn check_resp_header(header: &ResponseHeader) -> Result<()> {
             Err(Error::GlobalConfigNotFound(err.get_message().to_owned()))
         }
         ErrorType::Ok => Ok(()),
-        ErrorType::DuplicatedEntry | ErrorType::EntryNotFound => Err(box_err!(err.get_message())),
         ErrorType::Unknown => Err(box_err!(err.get_message())),
     }
 }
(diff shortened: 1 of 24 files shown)


LintianShi:
Replaced by PR #13533.

PR #13272, opened 2022-08-12
raft_engine: set recover_state

5kbpers:

Issue Number: Ref #12901

What's Changed:

```commit-message
Add `put_recover_from_raft_db` and `recover_from_raft_db` to raft engine for checking if store should be recovered from states in raftdb.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/engine_traits/src/raft_engine.rs
```diff
@@ -138,6 +144,10 @@ pub trait RaftEngine: RaftEngineReadOnly + PerfContextExt + Clone + Sync + Send
     where
         F: FnMut(u64) -> std::result::Result<(), E>,
         E: From<Error>;
+
+    fn store_version(&self) -> Result<Option<StoreVersion>>;
```
tabokie, line 148: Should implement for log batch?

5kbpers: Why?

BusyJay:
Why use bitflags instead of a number?

5kbpers:
> Why use bitflags instead of a number?

For extensibility, the name `version` may not accurate 😹

components/engine_traits/src/raft_engine.rs
tabokie: I mean `put_store_version`.

BusyJay:
> For extensibility...

In what cases a bitflags is better than number?

5kbpers:
> > For extensibility...
> 
> In what cases a bitflags is better than number?

With bitflags we can represent different options with different bits. I don't know if numbers can do this.

BusyJay:
> With bitflags we can represent different options with different bits.

Do we need that? Though the design makes it possible to control the behavior by configuration, but in the long term, we expect disabling WAL is the only behavior. With bitflags, there can be a lot of combination of features. We don't have the resource to test them through. Better stick to one and make it work the best.

components/engine_traits/src/raft_engine.rs
5kbpers: Not necessary, it will be called only once after restart.

components/engine_panic/src/raft_engine.rs
```diff
@@ -145,6 +145,14 @@ impl RaftEngine for PanicEngine {
     {
         panic!()
     }
+
+    fn recover_from_raftdb(&self) -> Result<bool> {
```
BusyJay, line 149: Should be `recover_from_raft_db`

components/engine_traits/Cargo.toml
```diff
@@ -8,6 +8,7 @@ publish = false
 failpoints = ["fail/failpoints"]
 
 [dependencies]
+bitflags = "1.0.1"
```
BusyJay, line 11: Not necessary any more.

components/engine_traits/src/raft_engine.rs
```diff
@@ -10,6 +10,8 @@ use crate::*;
 
 pub const RAFT_LOG_MULTI_GET_CNT: u64 = 8;
 
+pub const STORE_VERSION_RECOVER_FROM_RAFTDB: u64 = 0b00000001;
```
BusyJay, line 13: Should be 1.

components/engine_traits/src/raft_engine.rs
```diff
@@ -138,6 +140,10 @@ pub trait RaftEngine: RaftEngineReadOnly + PerfContextExt + Clone + Sync + Send
     where
         F: FnMut(u64) -> std::result::Result<(), E>,
         E: From<Error>;
+
+    fn recover_from_raftdb(&self) -> Result<bool>;
+
+    fn put_recover_from_raftdb(&self, recover_from_raftdb: bool) -> Result<()>;
```
BusyJay, line 146: Add some comment to describe what's the purpose of the function. And it seems wrong to me as the key is named `store_version_key`, but the method name is `put_recover_from_raftdb`. What if `store_version` is assigned to another value? It can easily be overwritten by mistake.

components/engine_traits/src/raft_engine.rs
```diff
@@ -138,6 +138,14 @@ pub trait RaftEngine: RaftEngineReadOnly + PerfContextExt + Clone + Sync + Send
     where
         F: FnMut(u64) -> std::result::Result<(), E>,
         E: From<Error>;
+
+    /// Indicate whether region states should be recovered from raftdb and
+    /// replay raft logs.
+    /// Generally it would be true when write-ahead-log of kvdb was disabled
+    /// before.
+    fn recover_from_raft_db(&self) -> Result<bool>;
```
tabokie, line 146: Could you just implement it here? They are using trait methods I presume?

components/raft_log_engine/src/engine.rs
```diff
@@ -344,6 +345,7 @@ const STORE_IDENT_KEY: &[u8] = &[0x01];
 const PREPARE_BOOTSTRAP_REGION_KEY: &[u8] = &[0x02];
 const REGION_STATE_KEY: &[u8] = &[0x03];
 const APPLY_STATE_KEY: &[u8] = &[0x04];
+const RECOVER_FROM_RAFT_DB_KEY: &[u8] = &[0x05];
```
tabokie, line 348: This is confusing. Why use a different key?

5kbpers: Just follow the style that all special keys written to raft engine were defined in the `raft_log_engine` crate.

tabokie: We should either move the key in "keys" to engine_rocks, or unify the key in different engines.

5kbpers: Those keys can not be simply moved to engine_rocks, for example the`LOCAL_PREFIX` was also used in raftstore. The right way is abstracting all operations related to keys to methods, which needs a lot of extra work and should not be included in this PR.

components/raft_log_engine/src/engine.rs
```diff
@@ -617,6 +619,27 @@ impl RaftEngine for RaftLogEngine {
         }
         Ok(())
     }
+
+    fn recover_from_raft_db(&self) -> Result<bool> {
```
tabokie, line 623: `get_recover_state`? and put it to `RaftEngineReadOnly`.

components/engine_traits/src/raft_engine.rs
```diff
@@ -138,6 +139,12 @@ pub trait RaftEngine: RaftEngineReadOnly + PerfContextExt + Clone + Sync + Send
     where
         F: FnMut(u64) -> std::result::Result<(), E>,
         E: From<Error>;
+
+    /// Indicate whether region states should be recovered from raftdb and
+    /// replay raft logs.
+    /// When kvdb's write-ahead-log is disabled, the sequence number of the last
+    /// boot time is saved.
+    fn put_recover_state(&self, seqno: u64) -> Result<()>;
```
BusyJay, line 147: I think this is ambiguous. It can be interpreted as "put a state, which is used for recover", but it actually means "put a number that will make start recover states from raft db".

tabokie: I think there's too much information to fit in one name (i.e. recovery mode, seqno). And it can change in the future. Also I think it is still essentially "a state used for recover". How we interpret it is application-level.

tabokie (approved):

components/engine_traits/src/raft_engine.rs
BusyJay: Then there should be a struct `RecoverState` instead of just a number.

BusyJay (commented):
Rest LGTM

components/engine_rocks/src/raft_engine.rs
```diff
@@ -151,6 +151,13 @@ impl RaftEngineReadOnly for RocksEngine {
         let key = keys::apply_state_key(raft_group_id);
         self.get_msg_cf(CF_DEFAULT, &key)
     }
+
+    fn get_recover_state(&self) -> Result<Option<StoreRecoverState>> {
+        let state = self
+            .get_value(keys::RECOVER_STATE_KEY)?
+            .map(|v| serde_json::from_slice(&v).unwrap());
```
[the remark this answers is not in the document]
5kbpers: https://github.com/pingcap/kvproto/pull/984 OK, submitted a PR for kvproto here.

5kbpers:
@BusyJay PTAL again

PR #13273, opened 2022-08-12
server: make tablet cache a local variable

### LINKED ISSUE #13214: EnginesResourcesInfo should be compatible with Multi-Rocks DB version.

SpadeA-Tang:

Issue Number: Ref #13214

What's Changed:

Remove field `cached_latest_tablets` of EnginesResourceInfo and  pass it in `update` as a parameter instead which is used to avoid repetitive memory allocation in the thread calling `update` periodically. 

Make EnginesResourceInfo be generic over tablet_factory which enable it to use KvEngineFactoryV2.

Add unit test for update.

### Check List

unit test

### Release note 

```release-note
None
```


components/server/src/server.rs
```diff
@@ -1843,45 +1853,52 @@ impl<EK: KvEngine, R: RaftEngine> EngineMetricsManager<EK, R> {
     }
 }
 
-pub struct EnginesResourceInfo {
-    tablet_factory: Arc<KvEngineFactory>,
+pub struct EnginesResourceInfo<T>
+where
+    T: TabletFactory<RocksEngine> + TabletAccessor<RocksEngine> + Sync + Send,
+{
+    tablet_factory: Arc<T>,
```
BusyJay, line 1860: Why not `Arc<dyn trait>`?

components/server/src/server.rs
```diff
@@ -1955,3 +1973,96 @@ impl IoBudgetAdjustor for EnginesResourceInfo {
...
+        let mut old_pending_compaction_bytes = 0;
+        let mut new_pending_compaction_bytes = 0;
+        for i in 1..6 {
+            let tablet = factory
+                .open_tablet(i, Some(10), OpenOptions::default().set_create_new(true))
+                .unwrap();
+
+            // Prepare some data for two tablets of the same region. So we can test whether
+            // we fetch the bytes from the latest one.
+            if i == 1 {
```
BusyJay, line 2016: You don't have to put it in loop.

components/server/src/server.rs
```diff
@@ -1955,3 +1973,96 @@ impl IoBudgetAdjustor for EnginesResourceInfo {
...
+
+        let engines_info = Arc::new(EnginesResourceInfo::new(Arc::new(factory), None, 10));
+
+        let mut cached_latest_tablets: HashMap<u64, (u64, RocksEngine)> = HashMap::new();
+        engines_info.update(Instant::now(), &mut cached_latest_tablets);
+
+        // The memory allocation should be reserved
+        assert!(cached_latest_tablets.capacity() >= 5);
+        // The tablet cache should be cleared
+        for i in 1..6 {
```
BusyJay, line 2055: assert is_empty should be enough.

components/server/src/server.rs
SpadeA-Tang: What's the benefit of Arc<dyn trait> over this generic?

components/server/src/server.rs
SpadeA-Tang: Good catch.

components/server/src/server.rs
BusyJay: `Arc` is used in both versions. I guess your question is why use trait object instead of generic. Just to make code consistent as `Arc<dyn TabletFactory>` is commonly used in the codebase. And use trait object should also make code simpler.

tonyxuqqi (approved):

BusyJay (approved):

Issue #13276, opened 2022-08-12
Improve the ResetToVersionWorker


JmPotato:
[`ResetToVersionWorker`](https://github.com/tikv/tikv/blob/38655bff985289560d6a6095bff54665d9cfd254/src/server/reset_to_version.rs#L46) is used to delete all keys whose version is later than the given timestamp. It could work as a flashback feature for the user if they want. We can improve its performance a litter bit according to these two mehods.

- [ ] Introduce [`TsFilter`](https://github.com/tikv/tikv/blob/38655bff985289560d6a6095bff54665d9cfd254/components/engine_rocks/src/options.rs#L97) to filter the SST whose latest version is smaller than the given one. This could help if the user has a lot of cold data.
- [ ] Introduce the multi-thread. Currently, `ResetToVersionWorker` only has a single thread to scan all three CFs of RocksDB (write/default/lock) and delete 256 keys for a batch with the default `BATCH_SIZE`. We can make this worker multi-thread to improve the deletion speed.

you06:
> Reject any scheduling, read or write request during the resetting process.

Is there a design of it? I have a simple idea, limiting the upper bound of timestamp from PD.

- Start resetting to a version `reset_ts`
- Let PD stop allocating timestamp, and responsing with the latest timestamp of the system `latest_ts`
- Clean the versions between `(reset_ts, latest_ts]`, and TiKV should reject requests with timestamp less than `latest_ts`
- Let PD continue allocating timestamps, and the cluster can continue serving requests

JmPotato:
> Is there a design of it? I have a simple idea, limiting the upper bound of timestamp from PD.

You can check https://github.com/tikv/tikv/pull/13348 to see how this works.

Limiting the allocation of TSO is reasonable but I'm afraid the difficulty comes from the current implementation that once PD stops allocating the TSO, many components such as TiDB and TiKV may panic or not work well. This should be considered first.

PR #13278, opened 2022-08-12
tikv: add TsFilter for ResetToVersionWorker

JmPotato:

Issue Number: Ref #13276.

What's Changed:

```commit-message
Add `TsFilter` for `ResetToVersionWorker` to filter out the SST that does not have a newer version than the TS to reset in `CF_WRITE`.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/src/server/reset_to_version.rs
+++ b/src/server/reset_to_version.rs
@@ -4,7 +4,7 @@ use std::{
     cell::RefCell,
     ops::Bound,
     sync::{Arc, Mutex},
-    thread::JoinHandle,
+    thread::{self, JoinHandle},
 };
 
 use engine_rocks::{RocksEngine, RocksEngineIterator, RocksWriteBatchVec};
@@ -234,33 +234,39 @@ impl ResetToVersionManager {
             warn!("A reset-to-version process is already in progress! Wait until it finish first.");
             self.wait();
         }
-        *self.worker_handle.borrow_mut() = Some(std::thread::Builder::new()
-            .name("reset_to_version".to_string())
-            .spawn_wrapper(move || {
-                tikv_util::thread_group::set_properties(props);
-                tikv_alloc::add_thread_memory_accessor();
+        *self.worker_handle.borrow_mut() = Some(
+            thread::Builder::new()
+                .name("reset_to_version".to_string())
+                .spawn_wrapper(move || {
+                    tikv_util::thread_group::set_properties(props);
+                    tikv_alloc::add_thread_memory_accessor();
 
-                let mut total_scanned = 0;
-                while worker.process_next_batch(BATCH_SIZE, &mut wb).expect("reset_to_version failed when removing invalid writes") {
-                }
-                {
-                    let mut state = worker.state.lock()
-                            .expect("failed to lock `ResetToVersionWorker::state` in `ResetToVersionWorker::process_next_batch`");
-                    total_scanned += *state.scanned();
-                    *state = ResetToVersionState::RemovingLock { scanned: 0 };
-                }
-                while worker.process_next_batch_lock(BATCH_SIZE, &mut wb).expect("reset_to_version failed when removing invalid locks") {
-                }
-                {
-                    let mut state = worker.state.lock()
-                            .expect("failed to lock `ResetToVersionWorker::state` in `ResetToVersionWorker::process_next_batch_lock`");
-                    total_scanned += *state.scanned();
-                    *state = ResetToVersionState::Done { total_scanned };
-                }
-                info!("Reset to version done!");
-                tikv_alloc::remove_thread_memory_accessor();
-            })
-            .expect("failed to spawn reset_to_version thread"));
+                    let mut total_scanned = 0;
+                    while worker
+                        .process_next_batch(BATCH_SIZE, &mut wb)
+                        .expect("reset_to_version failed when removing invalid writes")
+                    {
+                    }
+                    {
+                        let mut state = worker.state.lock().unwrap();
+                        total_scanned += *state.scanned();
+                        *state = ResetToVersionState::RemovingLock { scanned: 0 };
+                    }
+                    while worker
+                        .process_next_batch_lock(BATCH_SIZE, &mut wb)
+                        .expect("reset_to_version failed when removing invalid locks")
+                    {
+                    }
+                    {
+                        let mut state = worker.state.lock().unwrap();
+                        total_scanned += *state.scanned();
+                        *state = ResetToVersionState::Done { total_scanned };
+                    }
+                    info!("Reset to version done!");
+                    tikv_alloc::remove_thread_memory_accessor();
+                })
+                .expect("failed to spawn reset_to_version thread"),
+        );
     }
 
     /// Current process state.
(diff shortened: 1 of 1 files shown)


longfangsong (approved):

src/server/reset_to_version.rs
```diff
@@ -215,8 +216,16 @@ impl ResetToVersionManager {
         let readopts = IterOptions::new(None, None, false);
         let write_iter = self
             .engine
-            .iterator_opt(CF_WRITE, readopts.clone())
+            .iterator_opt(CF_WRITE, {
+                // Filter out the SST that does not have a newer version than `ts` in
+                // `CF_WRITE`.
+                let mut cf_write_readopts = readopts.clone();
+                cf_write_readopts.set_hint_min_ts(Bound::Excluded(ts.into_inner()));
```
Connor1996, line 223: seems should use `hint_max_ts`?

JmPotato: What we need to do here is to filter out the SST whose max TS is smaller than the TS we want to reset, in this case, we don't need to delete any key from it.

Connor1996: oh, my bad

src/server/reset_to_version.rs
```diff
@@ -231,16 +240,23 @@ impl ResetToVersionManager {
...
 
+                let mut total_scanned = 0;
                 while worker.process_next_batch(BATCH_SIZE, &mut wb).expect("reset_to_version failed when removing invalid writes") {
                 }
-                *worker.state.lock()
-                        .expect("failed to lock `ResetToVersionWorker::state` in `ResetToVersionWorker::process_next_batch`")
-                    = ResetToVersionState::RemovingLock { scanned: 0 };
+                {
+                    let mut state = worker.state.lock()
+                            .expect("failed to lock `ResetToVersionWorker::state` in `ResetToVersionWorker::process_next_batch`");
```
Connor1996, line 248: `unwrap` is adequate, it's a little verbose here

Connor1996 (approved):
rest LGTM

src/server/reset_to_version.rs
```diff
@@ -234,33 +234,39 @@ impl ResetToVersionManager {
...
+            thread::Builder::new()
+                .name("reset_to_version".to_string())
+                .spawn_wrapper(move || {
+                    tikv_util::thread_group::set_properties(props);
+                    tikv_alloc::add_thread_memory_accessor();
+
+                    let mut total_scanned = 0;
+                    while worker
+                        .process_next_batch(BATCH_SIZE, &mut wb)
+                        .expect("reset_to_version failed when removing invalid writes")
```
Connor1996, line 247: print out the error?

src/server/reset_to_version.rs
```diff
@@ -225,26 +234,39 @@ impl ResetToVersionManager {
...
+                    {
+                    }
+                    {
+                        let mut state = worker.state.lock().unwrap();
+                        total_scanned += *state.scanned();
+                        *state = ResetToVersionState::RemovingLock { scanned: 0 };
```
Connor1996, line 253: update the state for each batch of the loop.

src/server/reset_to_version.rs
```diff
@@ -85,10 +83,7 @@ impl ResetToVersionWorker {
 
     fn next_write(&mut self) -> Result<Option<(Vec<u8>, Write)>> {
         if self.write_iter.valid().unwrap() {
```
Connor1996, line 88:
```suggestion
        if self.write_iter.valid()? {
```

Connor1996: same for other places

src/server/reset_to_version.rs
```diff
@@ -225,34 +239,43 @@ impl ResetToVersionManager {
...
-            .expect("failed to spawn reset_to_version thread"));
+                    while worker
+                        .process_next_batch(BATCH_SIZE, &mut wb)
+                        .map_err(|err| {
+                            format!(
+                                "reset_to_version failed when removing invalid writes: {}",
+                                err,
+                            )
+                        })
+                        .unwrap()
```
Connor1996, line 257: should print the error but avoid panic

Connor1996 (commented):
the short value is inlined in write cf, check the commit info first and delete default cf only when needed `box_try!(wb.delete_cf(CF_DEFAULT, default_key.as_encoded()));`

src/server/reset_to_version.rs
```diff
@@ -171,6 +172,11 @@ impl ResetToVersionWorker {
             wb.write().unwrap();
             wb.clear();
         }
+        if !has_more {
```
Connor1996, line 175: no need for `has_more` as `wb.is_empty()` indicate it.

src/server/reset_to_version.rs
```diff
@@ -149,15 +152,13 @@ impl ResetToVersionWorker {
...
+                let mut state = self.state.lock().unwrap();
                 debug_assert!(matches!(
                     *state,
                     ResetToVersionState::RemovingLock { scanned: _ }
                 ));
                 *state.scanned() += 1;
+                self.total_scanned += 1;
                 drop(state);
 
                 box_try!(wb.delete_cf(CF_LOCK, self.lock_iter.key()));
```
Connor1996, line 164: related default value is not cleaned

src/server/reset_to_version.rs
```diff
@@ -225,34 +231,50 @@ impl ResetToVersionManager {
...
+                            }
+                            Err(e) => {
+                                error!("reset_to_version failed when removing from write/default cfs"; "error" => ?e);
+                                break;
+                            }
+                        }
+                    }
+                    // Delete the lock data in batch then.
+                    loop {
+                        match worker.process_next_batch_lock(&mut wb) {
```
Connor1996, line 256: I don't know the purpose of dividing it into two stages. @longfangsong If an error happened in the midst of `process_next_batch`, would it leave an incorrect state?

src/server/reset_to_version.rs
```diff
@@ -225,34 +231,50 @@ impl ResetToVersionManager {
...
+                                    break;
+                                }
+                            }
+                            Err(e) => {
+                                error!("reset_to_version failed when removing from lock cf"; "error" => ?e);
+                                break;
+                            }
+                        }
+                    }
+                    info!("Reset to version done!");
```
Connor1996, line 268: print `total_scanned` here

src/server/reset_to_version.rs
```diff
@@ -120,57 +120,55 @@ impl ResetToVersionWorker {
...
+                let mut state = self.state.lock().unwrap();
                 debug_assert!(matches!(
                     *state,
                     ResetToVersionState::RemovingLock { scanned: _ }
                 ));
                 *state.scanned() += 1;
+                self.total_scanned += 1;
                 drop(state);
 
                 box_try!(wb.delete_cf(CF_LOCK, self.lock_iter.key()));
```
Connor1996, line 156: User may not pass a `ts` which is larger than the `min-resolved-ts`. What if the primary key is committed, but the secondary lock is left due to error. Better add a paranoid check to make sure the lock's start ts is smaller than provided `ts`.

Connor1996: /cc @MyonKeminta

JmPotato:
Transaction correctness-related comments will be addressed in https://github.com/tikv/tikv/pull/13289.

Issue #13284, opened 2022-08-15
Encode causal timestamp in storage module for RawKV APIV2


haojinming:
## Development Task
Encode ts in storage module can benifit from following aspects:
1. Compatible with solution of issue: https://github.com/tikv/tikv/issues/12020
2. Resolved the rawkv cdc issue, that when propose failed, there may be dead lock in `resolved_ts`, which is currently resolved by dead lock detection in https://github.com/tikv/tikv/blob/7824699f4e05feb3eb37b12265ecc1e26511a4d7/components/cdc/src/endpoint.rs#L913
3. Helpful to implement stale read in rawkv, ref : https://github.com/tikv/rfcs/pull/96.

To implement this, we need three PRs to achieve it:

- [x] Let raw modify interface in storage use scheduler instead of grpc thread, including `raw_put`, `raw_delete`, `raw_batch_put`, `raw_batch_delete`, `raw_batch_put_atomic`, `raw_batch_delete_atomic`, `raw_compare_and_swap_atomic`. https://github.com/tikv/tikv/pull/13286
- [x] Get causal ts from CausalTsProvider and append ts to key in storage, remove [pre_propose observer](https://github.com/tikv/tikv/blob/7824699f4e05feb3eb37b12265ecc1e26511a4d7/components/causal_ts/src/observer.rs#L84-L114) in `CausalObserver`. Use `ConcurrencyManager` to lock on-road key instead of cdc [raw lock heap](https://github.com/tikv/tikv/blob/7824699f4e05feb3eb37b12265ecc1e26511a4d7/components/resolved_ts/src/resolver.rs#L48).  https://github.com/tikv/tikv/pull/13326
- [x] Remove [dead lock detection](https://github.com/tikv/tikv/blob/7824699f4e05feb3eb37b12265ecc1e26511a4d7/components/cdc/src/endpoint.rs#L913) and move to [tikv-cdc](https://github.com/tikv/migration/tree/main/cdc) and raw `resolved_ts` related code. https://github.com/tikv/tikv/pull/13496

haojinming:
All PR is merged in master branch. But there is still one issue unfixed: https://github.com/tikv/tikv/issues/13502.

PR #13286, opened 2022-08-15
rawkv: Reuse scheduler worker pool for raw modify command

haojinming:

Issue Number: Ref #13284 

What's Changed:

Let raw modify command reuse the txn schedduler instead of grpc thread.  

### Check List 

Tests 

- Unit test
- Integration test

Following performance test result shows that performance has slight increase compared with baseline. It should come from the balance usage of CPU. Both grpc and scheduler are used. 

  | ops | avg(ms) | p99(ms) | grpc poll CPU | scheduler cpu
-- | -- | -- | -- | -- | --
baeline-v6.2.0 | 17068.1 | 35 | 130 | 26%-34% | 0
PR | 17580.7 | 34 | 122 | 28%-34% | 20%-30%
Diff | +3.0% | -2.8% | -6.1% |   |  

  
### Release note 

```release-note
None
```


src/storage/mod.rs
```diff
@@ -1446,6 +1446,24 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
         Ok(())
     }
 
+    // Schedule raw modify commands, which reuse the txn scheduler.
+    // TODO: seperate the txn and raw commands if needed in the future.
+    fn sched_raw_command<T>(&self, future: T) -> Result<()>
```
pingyu, line 1451: 1. Suggest to check deadline exceeded in the begining of future, to avoid processing requests already timeout in client side. Refer to [here](https://github.com/tikv/tikv/blob/73c5d13455e0f07e5b38dd5de2d346afa2f1ddef/src/storage/txn/scheduler.rs#L511).

2. Suggest to add some scheduler related metrics, such as `SCHED_STAGE_COUNTER_VEC`, `SCHED_HISTOGRAM_VEC_STATIC`, to help trouble shootings (e.g., slow writes).

src/storage/mod.rs
```diff
@@ -1817,44 +1835,51 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
-        )?;
-        KV_COMMAND_COUNTER_VEC_STATIC.raw_put.inc();
-        Ok(())
+            let mut batch = WriteData::from_modifies(vec![m]);
+            batch.set_allowed_on_disk_almost_full();
+            let async_ret =
+                engine.async_write(&ctx, batch, Box::new(|res| cb(res.map_err(Error::from))));
+            let v: Result<()> = match async_ret {
+                Err(e) => Err(Error::from(e)),
+                Ok(_) => f.await.unwrap(),
```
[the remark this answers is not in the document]
haojinming: `f.wait` return  `Result<Result<(), Error>, Canceled>`, the outer `Result` is future result and the inner `Result` is `res.map_err(Error::from)` , the inner `Result` is handled by `cb`. I', not sure how to handle the outer Result, because spawn method ignoe the returned error here: https://github.com/tikv/tikv/blob/0591e684686183c3f7dbc051dea292f09cdc1457/components/tikv_util/src/yatp_pool/future_pool.rs#L156.  And I think future error should not occurs.

src/storage/mod.rs
haojinming: Added deadline check and metric. PTAL~

src/storage/mod.rs
```diff
@@ -1446,6 +1447,34 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
+    where
+        T: Future + Send + 'static,
+    {
+        SCHED_STAGE_COUNTER_VEC.get(tag).new.inc();
+        if self
+            .sched
+            .get_sched_pool(CommandPri::Normal)
+            .pool
+            .spawn(future)
+            .is_err()
```
pingyu, line 1462: Nit: try `map_err(Error::from(ErrorInner::SchedTooBusy))?`.

pingyu (commented):
Rest LGTM~

pingyu (approved):
LGTM~

haojinming:
@Connor1996 PTAL~  thanks.

src/storage/mod.rs
```diff
@@ -1446,6 +1447,29 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
         Ok(())
     }
 
+    // Schedule raw modify commands, which reuse the txn scheduler.
+    // TODO: seperate the txn and raw commands if needed in the future.
```
Connor1996, line 1451: separate

src/storage/mod.rs
```diff
@@ -1446,6 +1447,29 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
         Ok(())
     }
 
+    // Schedule raw modify commands, which reuse the txn scheduler.
```
Connor1996, line 1450: not txn scheduler, just scheduler worker pool

src/storage/mod.rs
```diff
@@ -1896,18 +1936,32 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
+            batch.set_allowed_on_disk_almost_full();
+            let (cb, f) = tikv_util::future::paired_future_callback();
+            let async_ret =
+                engine.async_write(&ctx, batch, Box::new(|res| cb(res.map_err(Error::from))));
+            let v: Result<()> = match async_ret {
+                Err(e) => Err(Error::from(e)),
+                Ok(_) => f.await.unwrap(),
+            };
+            callback(v);
+            KV_COMMAND_COUNTER_VEC_STATIC.raw_batch_put.inc();
```
Connor1996, line 1959: KV_COMMAND_COUNTER_VEC_STATIC.get(CMD).inc()

Connor1996: same for other places

src/storage/mod.rs
```diff
@@ -1817,44 +1841,60 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
-            F::encode_raw_value_owned(raw_value),
-        );
-
-        let mut batch = WriteData::from_modifies(vec![m]);
-        batch.set_allowed_on_disk_almost_full();
-
-        self.engine.async_write(
-            &ctx,
-            batch,
-            Box::new(|res| callback(res.map_err(Error::from))),
```
Connor1996, line 1838: What's the purpose for moving it to scheduler worker? Does the callback do heavy work?

src/storage/mod.rs
```diff
@@ -1962,31 +2025,45 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
+            let end_key = F::encode_raw_key_owned(end_key, None);
+
+            let mut batch =
+                WriteData::from_modifies(vec![Modify::DeleteRange(cf, start_key, end_key, false)]);
+            batch.set_allowed_on_disk_almost_full();
+
+            // TODO: special notification channel for API V2.
+            let (cb, f) = tikv_util::future::paired_future_callback();
+            let async_ret =
+                engine.async_write(&ctx, batch, Box::new(|res| cb(res.map_err(Error::from))));
```
Connor1996, line 2055:
```rust
let res = engine.async_write();
callback(res.map(f.await.unwrap()));
```
not sure if that is possible

src/storage/mod.rs
haojinming: The main reason is that raw modify interface will `get_ts` from `CausalTsProvider` and encode ts to apiv2 key in storage, which will be implemented in next pr. In grpc thread, [`get_ts`](https://github.com/tikv/tikv/blob/58fa80e0de0d43d473dc456081a4d2b08939e0aa/components/causal_ts/src/tso.rs#L566-L609) cause issue because it block on the thread and send tso request. So move the process to scheduler thread pool is the first step.

src/storage/mod.rs
haojinming: emmm, I tried this, ownership of `callback` is moved to box in `engine.async_write(&ctx, batch, Box::new(|res| callback(res.map_err(Error::from))));`, so it cannot be called after `async_write`.

src/storage/mod.rs
haojinming: Addressed, PTAL~

src/storage/mod.rs
Connor1996: `callback` is not `cb`

haojinming: oh, sorry for misunderstanding the point. I'm afraid `callback(|_| res.map(f.await.unwrap()));`  does not work as `await` does not work in closure. Compile error is like the following:
```
`await` is only allowed inside `async` functions and blocks
only allowed inside `async` functions and blocks
```

Issue #13203, opened 2022-08-02
v6.2.0 dr auto sync: data is not consistent after switch to backup dc and reset version to min resolved ts


mayjiang0203:
### What version of TiKV are you using?

TiKV
Release Version:   6.2.0-alpha
Edition:           Community
Git Commit Hash:   8a4c8b3828fe689e977c671796381d2e9b16bc07
Git Commit Branch: heads/refs/tags/v6.2.0
UTC Build Time:    2022-08-01 09:10:03
Rust Version:      rustc 1.62.0-nightly (7c4b47696 2022-04-30)
Enable Features:   pprof-fp jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
Profile:           dist_release

### Steps to reproduce

1、Deploy a cluster in dr auto sync mode;
2、Down primary dc when cluster in sync_recover mode;
3、Startup backup dc and use online recover to recover the cluster on backup db;
4、reset-to-version to min resolved ts; 

### What did you expect?
data should meet ACID.

### What did happened?

![image](https://user-images.githubusercontent.com/9443637/182285138-e2aed199-f96f-490c-aac7-7bf44f754b13.png)
![image](https://user-images.githubusercontent.com/9443637/182285164-58862f0c-6417-47cc-97b2-bd12e12e88ef.png)

w/a is waiting some more time, the bank count total will be correct.

PR #13289, opened 2022-08-15
tikv: fix the transaction correctness of ResetToVersionWorker

JmPotato:

Issue Number: ref #13276.

What's Changed:

```commit-message
This PR tries to fix the transaction correctness of `ResetToVersionWorker` without considering the case
that the TS to reset may be greater than the resolved TS and adds more test cases.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/src/server/reset_to_version.rs
+++ b/src/server/reset_to_version.rs
@@ -339,7 +339,11 @@ impl ResetToVersionManager {
                             }
                         }
                     }
-                    info!("Reset to version done!"; "total_scanned" => worker.total_scanned);
+                    info!("Reset to version done!";
+                        "ts_to_reset" => ?ts,
+                        "total_scanned" => worker.total_scanned,
+                        "total_deleted" => worker.total_deleted,
+                    );
                     tikv_alloc::remove_thread_memory_accessor();
                 }).unwrap(),
         );
--- a/src/server/reset_to_version.rs
+++ b/src/server/reset_to_version.rs
@@ -363,7 +363,7 @@ mod tests {
 
     use super::*;
 
-    struct CFData<'a> {
+    struct CfData<'a> {
         // key, start_ts, commit_ts, write_type, is_short_value
         write: Vec<(&'a [u8], u64, u64, WriteType, bool)>,
         // key, start_ts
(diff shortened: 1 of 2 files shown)


src/server/reset_to_version.rs
```diff
@@ -64,51 +100,57 @@ struct Batch {
...
         Ok(None)
     }
 
-    fn scan_next_batch(&mut self, batch_size: usize) -> Result<Batch> {
-        let mut writes = Vec::with_capacity(batch_size);
+    fn scan_next_batch(&mut self) -> Result<Batch> {
+        let mut writes = Vec::with_capacity(BATCH_SIZE);
         let mut has_more = true;
-        for _ in 0..batch_size {
+        for _ in 0..BATCH_SIZE {
```
Connor1996, line 151: while writes.len() < BATCH_SIZE

src/server/reset_to_version.rs
```diff
@@ -120,56 +162,89 @@ impl ResetToVersionWorker {
...
+                            deleted: _
+                        }
+                    ));
+                    *state.scanned() += 1;
+                    self.total_scanned += 1;
+                }
+                let key = self.lock_iter.key();
+                self.write_batch.delete_cf(CF_LOCK, key)?;
+                deleted += 1;
+                let lock = box_try!(Lock::parse(self.lock_iter.value()));
```
Connor1996, line 222:
```suggestion
                let lock = Lock::parse(self.lock_iter.value())?;
```

nolouch:
Does it fix the https://github.com/tikv/tikv/issues/13203?

JmPotato:
> Does it fix the #13203?

I am not quite sure what bug causes it, but the current implementation of `ResetToVersionWorker` has several correctness problems. Let's fix the basic implementation first and re-test it fully again later.

src/server/reset_to_version.rs
```diff
@@ -218,41 +294,60 @@ impl ResetToVersionManager {
...
+                                    break;
+                                }
+                            }
+                            Err(e) => {
+                                error!("Reset to version failed when resetting lock cf"; "error" => ?e);
+                                break;
+                            }
+                        }
+                    }
+                    info!("Reset to version done!"; "total_scanned" => worker.total_scanned);
```
Connor1996, line 342: print out the ts

Connor1996: print out how many keys are deleted for default/write/lock

Connor1996 (approved):

longfangsong (approved):

PR #13290, opened 2022-08-16
configs: add switch for tracking requests

sticnarf:

Issue Number: ref #12362

What's Changed:

```commit-message
A dynamic config server.track-request is added to control whether
the metrics of each request should be tracked and returned to the
client.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


src/server/service/kv.rs
```diff
@@ -1567,12 +1581,17 @@ fn future_scan<E: Engine, L: LockManager, F: KvFormat>(
...
     storage: &Storage<E, L, F>,
     mut req: BatchGetRequest,
+    configs: &AtomicConfigs,
 ) -> impl Future<Output = ServerResult<BatchGetResponse>> {
-    let tracker = GLOBAL_TRACKERS.insert(Tracker::new(RequestInfo::new(
-        req.get_context(),
-        RequestType::KvBatchGet,
-        req.get_version(),
-    )));
+    let tracker = if configs.get_track_request() {
```
TonsnakeLin, line 1586: Could you wrapper the code block `let tracker = if {} else {}` to a function? I found several duplicated code.

sticnarf: The types are different, so it's sadly impossible to extract the whole pattern as a function. And even the function names are not totally consistent, a macro cannot solve the problems either.

sticnarf: Ah, if you only want to eliminate the if/else but still extract the fields manually, it's possible.

PR #13294, opened 2022-08-16
raftstore: provide a common WaitApply interface

### LINKED ISSUE #13292: Provide a common WaitApply interface
Wait for a certain region or all regions to apply the latest or specified target index log is a common need in TiKV. However, the current implementation of `WaitApply` is coupled with the progress of Online Unsafe Recovery, so if some other services want to re-use this part of the code, it will be inconvenient. It would be nice if we can provide a common interface.

JmPotato:

Issue Number: Close #13292.

What's Changed:

```commit-message
Decouple WaitApply from the Unsafe Recovery code as much as possible.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -900,7 +900,7 @@ where
         });
         self.fsm
             .peer
-            .unsafe_recovery_maybe_finish_wait_apply(/* force= */ self.fsm.stopped);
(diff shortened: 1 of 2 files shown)


fengou1 (approved):
it looks good to me

Connor1996 (commented):
So how do other jobs trigger wait apply, I don't find the interface

PR #13299, opened 2022-08-17
copr: fix compatiblity of `json_extract` function

### LINKED ISSUE #13297: The `json_extract` function is not compatible with MySQL
There are two major differences between TiKV implementation and MySQL:

1. The result should be an array, if the path expression contains any glob.
2. The result in a single path expression should be unique (with the meaning of reference).

For example:

```
SELECT JSON_EXTRACT('[{"a": "a1", "b": 20.08, "c": false}, true]', '$**.c')
```

should be `[false]`, because it should be any array, as the path contains glob

And 

```
SELECT JSON_EXTRACT('[[0,1],[2,3],[4,[5,6]]]', '$**[0]');
```

should be `[[0, 1], 0, 1, 2, 3, 4, 5, 6]`, because the more `0`, `2`, `4`, `5` is duplicated, and will not be appended to the result.

YangKeao:

Issue Number: Close #13297

What's Changed:

1. The result will be an array, iff the path contains glob and matched at least one item.
2. The result from a single path expression will be unique (with the meaning of reference).

### Check List 

Tests 

- [x] Unit test
- Integration test
- No code

Side effects

- Performance regression
    - [x] Consumes more CPU
    - Consumes more MEM

The performance can be improved through constructing map to do the deduplication. However, the `PartialEq` is implemented not for reference, which makes it little more complicated. As the MySQL is also using a brute force compare-and-append algorithm, I think it's fine to leave it slow here.

### Release note 

```release-note
Fix the issue of incompatibility of `json_extract` function with MySQL
```


YangKeao:
PTAL @xiongjiwei

components/tidb_query_datatype/src/codec/mysql/json/json_extract.rs
```diff
@@ -14,21 +14,37 @@ impl<'a> JsonRef<'a> {
...
+        for elem in elem_list.iter() {
+            if elem.ref_eq(new_elem) {
+                continue 'outer;
+            }
         }
-        Ok(Some(Json::from_array(
-            elem_list.drain(..).map(|j| j.to_owned()).collect(),
-        )?))
+
+        elem_list.push(*new_elem)
```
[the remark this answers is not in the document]
YangKeao: Good! Let me remember the initial length, and only compare with item in `elem_list[..initial_length]`

xiongjiwei (approved):

wshwsh12 (approved):
LGTM

components/tidb_query_datatype/src/codec/mysql/json/json_extract.rs
```diff
@@ -14,21 +14,45 @@ impl<'a> JsonRef<'a> {
...
+}
+
+// append the elem_list vector, if the referenced json object doesn't exist
+// unlike the append in std, this function **doesn't** set the `other` length to
+// 0
+//
+// To use this function, you have to ensure both `elem_list` and `other` are
+// unique, with this assumption, we optimize the performance from O((n+m)*m) to
+// O(n*m).
+fn append_if_ref_unique<'a>(elem_list: &mut Vec<JsonRef<'a>>, other: &Vec<JsonRef<'a>>) {
```
wshwsh12, line 43: Could we use a hashMap to optimize the performance to O(n+m)?

YangKeao: Technically yes.

Let me add an extra struct to implement `PartialEq` with `ref_eq`, and insert it in the hashMap :thinking: .

YangKeao: Fixed.

gengliqi (approved):

PR #13300, opened 2022-08-17
raftstore-v2: support status query

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
And as an example to show how to setup test case.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/tests/integrations/mod.rs
```diff
@@ -165,10 +175,44 @@ impl Transport for TestTransport {
...
+    cfg.report_region_buckets_tick_interval = ReadableDuration::ZERO;
+    cfg.check_long_uncommitted_interval = ReadableDuration::ZERO;
+}
+
 fn setup_default_cluster() -> (TestNode, Receiver<RaftMessage>, TestRouter) {
     let mut node = TestNode::new();
-    let cfg = Default::default();
+    let mut cfg = v2_default_config();
+    disable_all_auto_ticks(&mut cfg);
+    cfg.store_io_pool_size = 1;
```
tabokie, line 214: rm

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+                return Err(box_err!("{:?} invalid status command!", self.logger.list()));
+            }
+        }
+
+        // Bind peer current term here.
+        cmd_resp::bind_term(resp, self.term());
+        Ok(())
+    }
+
+    fn fill_region_details(
```
tabokie, line 72: This short function could be inlined. I see the old `execute_region_detail` isn't reused.

BusyJay: It's not short, I prefer keep the match body small.

tabokie: It only has 10 lines when inlined. It should be considered short,

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+                if let Some(leader) = self.leader() {
+                    status_resp.mut_region_leader().set_leader(leader);
+                }
+            }
+            StatusCmdType::RegionDetail => {
+                if !self.storage().is_initialized() {
+                    let region_id = req.get_header().get_region_id();
+                    return Err(Error::RegionNotInitialized(region_id));
+                }
+                let status_resp = resp.mut_status_response();
```
tabokie, line 66: rm

tabokie (approved):

components/raftstore-v2/src/router/message.rs
```diff
@@ -95,49 +95,17 @@ impl StoreTick {
 }
 
 /// Command that can be handled by raftstore.
-pub struct RaftRequest {
+pub struct RaftRequest<C> {
```
tonyxuqqi, line 98: Why removing the RaftQuery and RaftCommand ?  And why not renaming it to RaftCommand (so that it's consistent with v1) ?

tonyxuqqi (approved):

components/raftstore-v2/src/router/message.rs
BusyJay: Because the definition of `RaftQuery` and `RaftCommand` are the same except callback. So I use generic to simplify the code. Note `RaftQuery` and `RaftCommand` are still used in `PeerMsg`, so renaming `RaftRequest` to `RaftCommand` can be confusing. We are rewriting raftstore, so the consistency with v1 is not important, but writing more readable code is. As commented, `RaftQuery` is for those query that doesn't need to go through log commit process, `RaftCommand` is for those commands that need to be executed in the same order on all replicas. Both `RaftQuery` and `RaftCommand` are request sent from client, hence name the shared struct as `RaftRequest`.

PR #13307, opened 2022-08-18
raftstore: Simplify raft local metrics

Connor1996:

Issue Number: Ref #12876
What's Changed:

```commit-message
Simplify raft local metrics by using local counter provided by rust-prometheus
```

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/local_metrics.rs
+++ b/components/raftstore/src/store/local_metrics.rs
@@ -25,8 +25,8 @@ macro_rules! set_send_status {
 
 pub struct RaftSendMessageMetrics(RaftSentMessageCounterVec);
 
-impl RaftSendMessageMetrics {
-    pub fn new() -> Self {
+impl Default for RaftSendMessageMetrics {
+    fn default() -> Self {
         Self(RaftSentMessageCounterVec::from(
             &STORE_RAFT_SENT_MESSAGE_COUNTER_VEC,
         ))
@@ -103,7 +103,7 @@ impl RaftMetrics {
     pub fn new(waterfall_metrics: bool) -> Self {
         Self {
             ready: RaftReadyCounterVec::from(&STORE_RAFT_READY_COUNTER_VEC),
-            send_message: RaftSendMessageMetrics::new(),
+            send_message: RaftSendMessageMetrics::default(),
             message_dropped: RaftDroppedMessageCounterVec::from(
                 &STORE_RAFT_DROPPED_MESSAGE_COUNTER_VEC,
             ),
--- a/components/raftstore/src/store/local_metrics.rs
+++ b/components/raftstore/src/store/local_metrics.rs
@@ -6,112 +6,51 @@ use std::sync::{Arc, Mutex};
 use collections::HashSet;
 use prometheus::local::LocalHistogram;
 use raft::eraftpb::MessageType;
+use tikv_util::time::{Duration, Instant};
 use tracker::{Tracker, TrackerToken, GLOBAL_TRACKERS};
 
 use super::metrics::*;
 
-/// The buffered metrics counters for raft ready handling.
-#[derive(Debug, Default, Clone)]
-pub struct RaftReadyMetrics {
-    pub message: u64,
-    pub commit: u64,
-    pub append: u64,
-    pub snapshot: u64,
-    pub pending_region: u64,
-    pub has_ready_region: u64,
-}
+const METRICS_FLUSH_INTERVAL: u64 = 10_000; // 10s
 
-impl RaftReadyMetrics {
-    /// Flushes all metrics
-    fn flush(&mut self) {
-        // reset all buffered metrics once they have been added
-        if self.message > 0 {
-            STORE_RAFT_READY_COUNTER.message.inc_by(self.message);
-            self.message = 0;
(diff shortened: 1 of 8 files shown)


BusyJay:
What's the performance impact?

Connor1996:
> What's the performance impact?

No difference, it doesn't involve the thread local storage. It's still the local metrics stored in `PollContext`, but we don't need to define the struct manually. It's already generated by `make_static_metircs!` with types of `LocalIntCounter`.

components/raftstore/src/store/local_metrics.rs
```diff
@@ -123,317 +62,62 @@ impl RaftSendMessageMetrics {
...
+
+    pub leader_missing: Arc<Mutex<HashSet<u64>>>,
+
+    last_flush_time: Instant,
 }
 
 impl RaftMetrics {
     pub fn new(waterfall_metrics: bool) -> Self {
         Self {
+            ready: RaftReadyCounterVec::from(&STORE_RAFT_READY_COUNTER_VEC),
```
Connor1996, line 105: @BusyJay here is the key change. Before we define `RaftReadyMetrics` manually by repeating the fields in the static metrics definition. Now we use the local of the static metrics directly.

Connor1996: it's quite like calling `STORE_RAFT_READY_COUNTER_VEC.local()` but a static version

Connor1996: see https://github.com/tikv/rust-prometheus/blob/master/static-metric/examples/simple.rs#L24

BusyJay: Yes, my question is what's the difference between using provided APIs and defining it manually? Are they the exact same struct?

Connor1996: Yes, it's just a wrapper for `LocalIntCounter` inherently, see generated code https://github.com/tikv/rust-prometheus/blob/master/static-metric/examples/make_auto_flush_static_counter.rs#L270

components/raftstore/src/store/peer.rs
```diff
@@ -1569,19 +1569,28 @@ where
         self.raft_group.snap()
     }
 
-    fn add_ready_metric(&self, ready: &Ready, metrics: &mut RaftReadyMetrics) {
-        metrics.message += ready.messages().len() as u64;
-        metrics.commit += ready.committed_entries().len() as u64;
-        metrics.append += ready.entries().len() as u64;
+    fn add_ready_metric(&self, ready: &Ready, metrics: &mut RaftMetrics) {
```
BusyJay, line 1572: Why the change?

components/raftstore/src/store/worker/metrics.rs
```diff
@@ -78,7 +78,7 @@ thread_local! {
     );
 }
 
-const METRICS_FLUSH_INTERVAL: u64 = 15_000; // 15s
+const METRICS_FLUSH_INTERVAL: u64 = 10_000; // 10s
```
BusyJay, line 81: Why?

Connor1996: All other places use 10s

components/raftstore/src/store/peer.rs
Connor1996: Don't want to introduce import from metrics as `RaftReadyMetrics` is deleted now.

tabokie (approved):
I hope we can run some microbenchmarks (on TiKV) for these changes. I see there're some in tests/benches, but there's no guide on which one we should run.

Connor1996:
> I hope we can run some microbenchmarks (on TiKV) for these changes. I see there're some in tests/benches, but there's no guide on which one we should run.

No need cause there is no difference. It doesn't like other PRs which introduce additional overhead for collecting metrics. This PR just writes the code in a neater way.

tabokie:
I'm not saying that targeting this specific PR. But as a general idea to avoid repeating the regression disaster in 6.2.

But for this PR, even if you are sure of it in theory, it is certainly better to have actual results as proof. Otherwise we wouldn't need CI because I'm sure we are all pretty confident of our own changes until proven wrong.

ethercflow (approved):
LGTM

Connor1996:
> I'm not saying that targeting this specific PR. But as a general idea to avoid repeating the regression disaster in 6.2.
> 
> But for this PR, even if you are sure of it in theory, it is certainly better to have actual results as proof. Otherwise we wouldn't need CI because I'm sure we are all pretty confident of our own changes until proven wrong.

I agree it's better to take care of performance overhead to avoid repeating the regression. But not every PR worth to do the bench, it's a waste of the time(I cherish the memory of `/bench` bot).

tabokie:
@Connor1996 That's why I'm talking about microbenchmarks. There're some many benches already but nobody seems to use them.

Connor1996:
> @Connor1996 That's why I'm talking about microbenchmarks. There're some many benches already but nobody seems to use them.

I don't think the micro bench can find the previous regressions. It needs under heavy load. The daily bench with fine sensitivity would be a good practice.

Issue #13303, opened 2022-08-18
Tracking Issue for Flashback


JmPotato:
# Background

Since we store all MVCC versions of every key in TiKV since the last GC point, we have the ability to restore any key in any given range to any of the existing specified MVCC versions. Providing such an interface whose name is `kv_flashback_to_version` allows for data recovery capabilities like the Flashback feature.

# Task

- [x] Update https://github.com/pingcap/kvproto for the new gRPC interface. (https://github.com/pingcap/kvproto/pull/971)
- [x] Implement the `kv_flashback_to_version` gRPC interface. (https://github.com/tikv/tikv/pull/13378)
- [x] Implement the `flashback_to_version` KV transaction command and action. (https://github.com/tikv/tikv/pull/13345)
- [x] Reject any scheduling, reading, or writing request during the flashback. (https://github.com/tikv/tikv/pull/13348)
- [x] Wait for the latest last index to be committed and applied before the flashback. (https://github.com/tikv/tikv/pull/13348)
- [x] Check the ResolvedTS during the preparation to make sure the flashback version is not greater than it. (https://github.com/tikv/tikv/pull/13412)
- [x] https://github.com/tikv/tikv/issues/13397
- [x] Re-implement the `flashback_to_version` KV transaction command to use MVCC overwriting. (https://github.com/tikv/tikv/pull/13450)
- [x] https://github.com/tikv/tikv/issues/13493
- [x] https://github.com/tikv/tikv/issues/13519

JmPotato:
Close since the basic Flashback ability has finished development.

PR #13308, opened 2022-08-18
*: add kv_flashback_to_version interface

JmPotato:

Issue Number: ref #13303.

What's Changed:

```commit-message
Add `kv_flashback_to_version` interface.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -11,7 +11,7 @@ use std::{
     },
     iter::{FromIterator, Iterator},
     mem,
-    sync::{Arc, Mutex, mpsc::SyncSender},
+    sync::{mpsc::SyncSender, Arc, Mutex},
     time::{Duration, Instant},
     u64,
 };
@@ -78,10 +78,9 @@ use crate::{
         metrics::*,
         msg::{Callback, ExtCallback, InspectedRaftMessage},
         peer::{
-            ConsistencyState, ForceLeaderState, Peer, PersistSnapshotResult, StaleState,
-            UnsafeRecoveryExecutePlanSyncer, UnsafeRecoveryFillOutReportSyncer,
+            ConsistencyState, FlashbackState, ForceLeaderState, Peer, PersistSnapshotResult,
+            StaleState, UnsafeRecoveryExecutePlanSyncer, UnsafeRecoveryFillOutReportSyncer,
             UnsafeRecoveryForceLeaderSyncer, UnsafeRecoveryState, UnsafeRecoveryWaitApplySyncer,
-            FlashbackState, FlashbackWaitApplySyncer,
             TRANSFER_LEADER_COMMAND_REPLY_CTX,
         },
         transport::Transport,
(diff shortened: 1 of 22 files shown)


src/storage/mod.rs
```diff
@@ -1515,6 +1519,76 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
...
+            self.api_version,
+            ctx.api_version,
+            CommandKind::flashback_to_version,
+            [(Some(req.get_start_key()), Some(req.get_end_key()))],
+        )?;
+
+        let region_id = ctx.get_region_id();
+        // Reject any following scheduling, read or writing request, and wait for the
+        // applied index to reach the latest committed index.
+        let (tx, rx) = sync_channel(1);
```
Connor1996, line 1539: the future is spawned on grpc thread, you must't do sync wait here, otherwise it blocks the grpc thread.

src/storage/txn/commands/flashback_to_version.rs
```diff
@@ -0,0 +1,75 @@
...
+            self.version,
+            // Clear the whole `CF_LOCK`, so we use `TimeStamp::zero()`.
+            TimeStamp::zero(),
+            Some(self.start_key),
+            Some(self.end_key),
+        );
+        flashback_to_version(&mut txn, &mut reader)?;
+        let mut txn_extra = TxnExtra::default();
+        txn_extra.flashback = true;
+        let mut write_data = WriteData::new(txn.into_modifies(), txn_extra);
```
Connor1996, line 63: it may be too large to fit raft entry size limit

Connor1996: you can refer to RESOLVE_LOCK_BATCH_SIZE of ResolveLock command to do it in multiple rounds

src/storage/mvcc/reader/reader.rs
```diff
@@ -626,6 +626,121 @@ impl<S: EngineSnapshot> MvccReader<S> {
...
+            let _ok = cursor.seek_to_first(&mut self.statistics.lock);
+            // TODO: handle the result of `seek_to_first` properly.
+            self.lock_cursor = Some(cursor);
+        }
+        Ok(())
+    }
+
+    /// Scan all the MVCC versions of keys within the given key range in
+    /// `CF_WRITE` that are committed after the `min_write_version`.
+    pub fn scan_next_batch_write(&mut self) -> Result<Vec<(Key, Write)>> {
```
Connor1996, line 700: can we add it to `MvccReader`?

src/storage/mvcc/reader/reader.rs
```diff
@@ -626,6 +626,121 @@ impl<S: EngineSnapshot> MvccReader<S> {
...
+                ));
+            }
+            ok = cursor.next(&mut self.statistics.write);
+        }
+        Ok(writes)
+    }
+
+    /// Scan all the MVCC versions of keys within the given key range in
+    /// `CF_LOCK` that are written after the `min_lock_version`.
+    pub fn scan_next_batch_lock(&mut self) -> Result<Vec<(Key, Lock)>> {
```
Connor1996, line 724: seems it can reuse `scan_locks` of `MvccReader`

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2080,6 +2105,24 @@ where
         }
     }
 
+    fn maybe_finish_flashback_wait_apply(&mut self) {
+        if let Some(flashback_state) = &self.fsm.peer.flashback_state {
+            // TODO: do we need to handle the just-proposed but uncommitted entry?
```
Connor1996, line 2110: need to consider just-proposed but uncommitted entry

components/raftstore/src/store/peer.rs
```diff
@@ -3405,6 +3424,19 @@ where
         if self.pending_remove {
             return false;
         }
+        // When in the flashback state, we should not allow any other request to be
```
Connor1996, line 3427: do we need to reject reads?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2080,6 +2105,24 @@ where
         }
     }
 
+    fn maybe_finish_flashback_wait_apply(&mut self) {
```
Connor1996, line 2108: better reuse the wait apply logic

components/raftstore/src/store/peer.rs
Connor1996: should check it in `pre_propose_raft_command`, returning false directly here would cause callback uninvoked.

PR #13309, opened 2022-08-18
storage: record and return pessimistic_lock_wait time

OneSizeFitsQuorum:
Signed-off-by: OneSizeFitQuorum <tanxinyu@apache.org>

Issue Number: ref #12362

Depends on [pingcap/kvproto#965](https://github.com/pingcap/kvproto/pull/965). There is still some controversy about how to organize the returned metrics in the protobuf.

What's Changed:

```commit-message
This commit record the pessimistic_lock_wait time for pessimistic transactions in the
waitManager.
```
### Check List 

Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note
None
```

### DIFF
--- a/components/pd_client/src/util.rs
+++ b/components/pd_client/src/util.rs
@@ -848,11 +848,12 @@ pub fn check_resp_header(header: &ResponseHeader) -> Result<()> {
         ErrorType::IncompatibleVersion => Err(Error::Incompatible),
         ErrorType::StoreTombstone => Err(Error::StoreTombstone(err.get_message().to_owned())),
         ErrorType::RegionNotFound => Err(Error::RegionNotFound(vec![])),
-        ErrorType::Unknown => Err(box_err!(err.get_message())),
         ErrorType::GlobalConfigNotFound => {
             Err(Error::GlobalConfigNotFound(err.get_message().to_owned()))
         }
         ErrorType::Ok => Ok(()),
+        ErrorType::DuplicatedEntry | ErrorType::EntryNotFound => Err(box_err!(err.get_message())),
+        ErrorType::Unknown => Err(box_err!(err.get_message())),
     }
 }
 
--- a/src/server/lock_manager/waiter_manager.rs
+++ b/src/server/lock_manager/waiter_manager.rs
@@ -111,7 +111,7 @@ pub enum Task {
         lock: Lock,
         timeout: WaitTimeout,
         diag_ctx: DiagnosticContext,
-        wait_timer: Instant,
+        start_waiting_time: Instant,
     },
     WakeUp {
         // lock info
@@ -183,7 +183,7 @@ pub(crate) struct Waiter {
     pub(crate) lock: Lock,
     pub diag_ctx: DiagnosticContext,
     delay: Delay,
-    wait_timer: Instant,
+    start_waiting_time: Instant,
 }
 
 impl Waiter {
@@ -194,7 +194,7 @@ impl Waiter {
         lock: Lock,
         deadline: Instant,
         diag_ctx: DiagnosticContext,
-        wait_timer: Instant,
+        start_waiting_time: Instant,
     ) -> Self {
         Self {
             start_ts,
@@ -203,7 +203,7 @@ impl Waiter {
             lock,
             delay: Delay::new(deadline),
             diag_ctx,
-            wait_timer,
+            start_waiting_time,
         }
     }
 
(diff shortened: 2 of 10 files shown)


src/server/lock_manager/waiter_manager.rs
```diff
@@ -201,6 +206,7 @@ impl Waiter {
             delay: Delay::new(deadline),
             diag_ctx,
             _lifetime_timer: WAITER_LIFETIME_HISTOGRAM.start_coarse_timer(),
+            wait_timer,
```
sticnarf, line 209: I think `wait_timer` duplicates with `_life_timer`. They may be combined into one.

It seems the place you create `wait_timer` records the time more accurately.

OneSizeFitsQuorum: > I think wait_timer duplicates with _life_timer. They may be combined into one.

Good idea~I have combined them together. 

> It seems the place you create wait_timer records the time more accurately.

I'm conservative in metric, not wanting to ignore the Runtime overhead.

components/pd_client/src/util.rs
```diff
@@ -848,11 +848,12 @@ pub fn check_resp_header(header: &ResponseHeader) -> Result<()> {
         ErrorType::IncompatibleVersion => Err(Error::Incompatible),
         ErrorType::StoreTombstone => Err(Error::StoreTombstone(err.get_message().to_owned())),
         ErrorType::RegionNotFound => Err(Error::RegionNotFound(vec![])),
-        ErrorType::Unknown => Err(box_err!(err.get_message())),
         ErrorType::GlobalConfigNotFound => {
             Err(Error::GlobalConfigNotFound(err.get_message().to_owned()))
         }
         ErrorType::Ok => Ok(()),
+        ErrorType::DuplicatedEntry | ErrorType::EntryNotFound => Err(box_err!(err.get_message())),
```
TonsnakeLin, line 855: ErrorType::DuplicatedEntry and ErrorType::EntryNotFound in kvproto are added by you?

sticnarf: No, it's added by others but not yet used.

src/server/service/kv.rs
```diff
@@ -2091,7 +2091,12 @@ txn_command_future!(future_acquire_pessimistic_lock, PessimisticLockRequest, Pes
                 tracker.write_write_detail(resp.mut_exec_details_v2().mut_write_detail());
             });
         },
-        Err(e) | Ok(Err(e)) => resp.set_errors(vec![extract_key_error(&e)].into()),
+        Err(e) | Ok(Err(e)) =>  {
+            GLOBAL_TRACKERS.with_tracker(tracker, |tracker| {
```
sticnarf, line 2095: What about just moving L2089 out of match? There can be `scan_detail` even if it fails to lock.

components/pd_client/src/util.rs
OneSizeFitsQuorum: No, but this commit was earlier than we wanted it to be, so I had to improvise an easy solution or the build wouldn't pass. @sticnarf did the same thing in this [pr](https://github.com/tikv/tikv/pull/13159)
image

src/server/service/kv.rs
OneSizeFitsQuorum: done

src/server/lock_manager/waiter_manager.rs
```diff
@@ -110,6 +111,7 @@ pub enum Task {
         lock: Lock,
         timeout: WaitTimeout,
         diag_ctx: DiagnosticContext,
+        wait_timer: Instant,
```
sticnarf, line 114: If you call this `xx_timer`, what about still using `WAITER_LIFETIME_HISTOGRAM.start_timer()`?

OneSizeFitsQuorum: According to the existing API definition, it is not convenient to obtain the specific time in the HistogramTimer, so I directly used `WAITER_LIFETIME_HISTOGRAM.Observe (duration_to_sec(elapsed))` to record the metric. I think we could change the name just to make it simple? What about `start_waiting_time` or `enqueue_time` ? What's your opinion?

sticnarf: Umm, right.

`start_waiting_time` sounds ok.

OneSizeFitsQuorum: Fixed~

PR #13312, opened 2022-08-19
resolved_ts: fix check_leader to tiflash proxy

### LINKED ISSUE #13310: The address of the check leader message sent to the TiFlash proxy is wrong
https://github.com/tikv/tikv/blob/1ea26a2ac8761af356cc5c0825eb89a0b8fc9749/components/resolved_ts/src/advance.rs#L400
For TiFlash proxy, should use `store.peer_address` instead of `store.address`.

hehechen:
Signed-off-by: hehechen <awd123456sss@gmail.com>

Issue Number: Close #13310  Ref #12092

What's Changed:
```commit-message
For TiFlash proxy, should use store.peer_address instead of store.address.
For TiKV, always use the same peer_address and address, so the change won't affect.
```

### Check List 

Tests 

- Unit test
- Integration test
Add some log in TiFlash Proxy when receive check_leader RPC.
- No code

### Release note 

```release-note
None
```


ethercflow (approved):
LGTM

Issue #13313, opened 2022-08-19
Read Flow Resource Usage Control


glorv:
## Development Task

In the current implement, big queries may consume too much resources that can significantly impact the throughput and latency of other queries, in some cases, the impact is too high that it's unacceptable in the production environment. We should optimize the schedule and execution mechanism in order the solve this problem.

There are some are that we would like to optimize:
- Better task schedule in the unified-thread-pool/yatp.
- Fairer reschedule of coprocessor task
- Thread pool auto scale
- ...

PR #13314, opened 2022-08-19
coprocessor: support auto scale read-pool

glorv:

Issue Number: ref #13313

What's Changed:

```
Support automatically scale read pool thread count based on the CPU usage.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Support automatically scale read pool thread count based on the CPU usage.
```


BornChanger:
please fix the compilation issue.

src/read_pool.rs
```diff
@@ -291,15 +310,182 @@ impl From<Vec<FuturePool>> for ReadPool {
...
+    fn adjust_pool_size(&mut self) {
+        if !self.auto_adjust
+            || (self.cur_thread_count == self.max_thread_count
+                && self.core_thread_count == self.max_thread_count)
+        {
+            return;
+        }
+        let cpu_quota = SysQuota::cpu_cores_quota();
+        self.thread_stats.record();
+        let thread_stats = self.thread_stats.get_cpu_usages();
```
sticnarf, line 377: A problem here is that if there is a lot of IO done in the thread, the CPU usage of the threads will be far less than how much the thread pool is occupied actually.

Maybe it's better to use the portion of time used in the pool over the total thread number.

glorv: Thanks for the reminding, I'll try that.

glorv:
@sticnarf PTAL again, thanks~

src/read_pool.rs
```diff
@@ -292,14 +305,216 @@ impl From<Vec<FuturePool>> for ReadPool {
...
+            && running_tasks < self.cur_thread_count as i64 * RUNNING_TASKS_PER_THREAD_THRESHOLD
+        {
+            self.cur_thread_count - 1
+        } else {
+            self.cur_thread_count
+        };
+
+        if new_thread_count != self.cur_thread_count {
+            self.handle.scale_pool_size(new_thread_count);
+            info!("auto adjust read pool thread"; "read_pool_cpu" => read_pool_cpu,
```
sticnarf, line 441: Could this log be too noisy for real production systems? Is a gauge metric enough for this?

src/read_pool.rs
```diff
@@ -292,14 +305,216 @@ impl From<Vec<FuturePool>> for ReadPool {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Task::PoolSize(s) => write!(f, "PoolSize({})", *s),
+            Task::AutoAdjust(s) => write!(f, "AutoAdjust({})", *s),
+        }
+    }
+}
+
+pub struct ReadPoolConfigManager {
+    scheduler: Scheduler<Task>,
```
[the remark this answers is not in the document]
glorv: This reuses the background task thread, so there should be no much overhead.

glorv:
@sticnarf PTAL again, thanks~

components/tidb_query_executors/src/runner.rs
```diff
@@ -439,15 +433,15 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
+        let mut batch_size = batch_initial_size();
         let mut warnings = self.config.new_eval_warnings();
         let mut ctx = EvalContext::new(self.config.clone());
         let mut record_all = 0;
 
         loop {
             let mut chunk = Chunk::default();
             let mut sample = self.quota_limiter.new_sample(true);
-            let (drained, record_len) = {
+            let (exec_duration, (drained, record_len)) = {
```
sticnarf, line 444: Well, it seems more tricky to use the cpu time from `observe_cpu_async`.

First, it may be better to use wall clock time instead of thread cpu time, due to the same reason in https://github.com/tikv/tikv/pull/13314#discussion_r949900731.

And also, the sampler doesn't always record the thread time because it is expensive to get thread CPU time. It depends on whether there is a quota limit. So, you may get zero with the default config.

glorv: Oh, yes, if the cpu limit is not enabled, the cpu time is not actually observed by the quota sample. I will switch back to use wall clock though it's a bit weird to use it in an async context.

glorv: Seems there is no good way the track the cpu time used in the async context, and since we already moved the reschedule check to a lower level, the batch size adjust here should be less effective. I will rollback it to the original logic.

src/read_pool.rs
```diff
@@ -292,14 +313,215 @@ impl From<Vec<FuturePool>> for ReadPool {
...
+}
+
+impl Runnable for ReadPoolConfigRunner {
+    type Task = Task;
+    fn run(&mut self, task: Self::Task) {
+        match task {
+            Task::PoolSize(s) => {
+                if s != self.core_thread_count {
+                    self.handle.scale_pool_size(s);
+                    self.core_thread_count = s;
```
sticnarf, line 379: Sorry, I don't get it. If we always change both `core_thread_count` and `cur_thread_count` to the same value, how to understand the meaning of `core_thread_count`?

components/tidb_query_executors/src/runner.rs
sticnarf: Let's remove `The "Adjust coprocessor batch-size based on the cost time of each round." ` from the PR description. We can add it back in later PRs if we find some good way.

src/read_pool.rs
glorv: `core_thread_count` is the config value of active thread count, `current_thread_count` is the current active thread count. Here, because use changes the config value, so the core and current thread count is the same after we scales the pool size.

components/tidb_query_executors/src/runner.rs
glorv: done.

src/read_pool.rs
```diff
@@ -12,18 +13,30 @@ use online_config::{ConfigChange, ConfigManager, ConfigValue, Result as CfgResul
...
 use crate::{
     config::{UnifiedReadPoolConfig, UNIFIED_READPOOL_MIN_CONCURRENCY},
     storage::kv::{destroy_tls_engine, set_tls_engine, Engine, FlowStatsReporter},
 };
 
+// the duration to check auto-scale unified-thread-pool's thread
+const READ_POOL_THREAD_CHECK_DURATION: Duration = Duration::from_secs(10);
+// consider scale up read pool size if the average thread cpu usage is higher
+// than this threahold.
+const READ_POOL_THREAD_HIGH_THREAHOLD: f64 = 0.8;
```
sticnarf, line 36: typo `THRESHOLD`

src/read_pool.rs
sticnarf: Ah, I see.

sticnarf (approved):
LGTM

glorv:
@Connor1996 @BusyJay PTAL, thanks~

glorv:
@BusyJay PTAL

src/read_pool.rs
```diff
@@ -292,14 +313,215 @@ impl From<Vec<FuturePool>> for ReadPool {
...
+    // the time when the threads are off-cpu, so it might be much higher than the actual cpu time.
+    prev_total_task_handling_time_us: u64,
+    prev_check_time: Instant,
+    prev_cpu_per_sec: f64,
+}
+
+impl ReadPoolCpuTimeTracker {
+    fn new(pool_name: String) -> Self {
+        let prev_check_time = Instant::now_coarse();
+        let prev_total_task_handling_time_us = MULTILEVEL_LEVEL_ELAPSED
```
Connor1996, line 328: can we use static metrics

glorv: Yes, It's better to use a static metric.  @Connor1996  PTAL again, thank you

src/read_pool.rs
```diff
@@ -292,14 +313,214 @@ impl From<Vec<FuturePool>> for ReadPool {
...
+        let process_cpu = match self.process_stats.cpu_usage() {
+            Ok(p) => p,
+            Err(e) => {
+                warn!("fetch process cpu usage failed"; "err" => ?e);
+                return;
+            }
+        };
+        let cpu_quota = SysQuota::cpu_cores_quota();
+
+        let new_thread_count = if self.cur_thread_count < self.max_thread_count
```
Connor1996, line 432: add comment to illustrate the policy

Connor1996 (approved):
rest LGTM

src/read_pool.rs
glorv: done

PR #13317, opened 2022-08-19
copr: use manually written json path parser

### LINKED ISSUE #13316: JSON path is is incompatible with MySQL
The json path parser is incompatible with MySQL. It doesn't validate the ECMAScript identifier, fails to return the position of error and is extremely hard to read and extend.

For example, it fails to parse: `$ .  你好`, `$  . $`, through they are all valid identifiers.

And the expected return value should be in form: 

```
ERROR 3143 (42000): Invalid JSON path expression. The error is around character position X.
```

But not a simple error `Invalid JSON path: XXX`

YangKeao:

Issue Number: Close #13316

What's Changed:

Use a manually written json path parser to parse the path.

### Check List

Tests

- [x] Unit test
- Integration test
- No code

### Release note

```release-note
Fix the issue that the json path format is incompatible with MySQL
```

### DIFF
--- a/components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
+++ b/components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
@@ -25,20 +25,12 @@
 //     select json_extract('{"a": "b", "c": [1, "2"]}', '$.*') -> ["b", [1, "2"]]
 // ```
 
-use std::ops::Index;
-
-use regex::Regex;
+use std::{iter::Peekable, str::CharIndices};
 
 use super::json_unquote::unquote_string;
 use crate::codec::Result;
 
 pub const PATH_EXPR_ASTERISK: &str = "*";
-
-// [a-zA-Z_][a-zA-Z0-9_]* matches any identifier;
-// "[^"\\]*(\\.[^"\\]*)*" matches any string literal which can carry escaped
-// quotes.
-const PATH_EXPR_LEG_RE_STR: &str =
-    r#"(\.\s*([a-zA-Z_][a-zA-Z0-9_]*|\*|"[^"\\]*(\\.[^"\\]*)*")|(\[\s*([0-9]+|\*)\s*\])|\*\*)"#;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum PathLeg {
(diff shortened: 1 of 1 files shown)


YangKeao:
PTAL @xiongjiwei

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,325 @@ impl PathExpression {
...
+
+                Ok(PathExpressionToken::Leg((
+                    PathLeg::Key(PATH_EXPR_ASTERISK.to_string()),
+                    Position {
+                        start,
+                        end: self.current_index(),
+                    },
+                )))
+            }
+            (mut key_start, '"') => {
```
xiongjiwei, line 177: can you please add a test like `'$."a\t"'`. MySQL will report an invalid path since `"a\t"` is not a valid json string. 

it is a corner case, I think it is ok to add the check in the future. ignore this if the test case is passed

xiongjiwei (approved):
LGTM

xiongjiwei:
lint failed

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
YangKeao: Oops! I didn't realize it. I'll add a test case and fix.

YangKeao: Fixed. I checked the ascii control character. However, after unquote, I cannot tell the concrete position of the original input, so I only return the beginning position of the string literal.

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,325 @@ impl PathExpression {
...
+    Start(Position),
+}
+
+impl<'a> Iterator for PathExpressionTokenizer<'a> {
+    type Item = Result<PathExpressionToken>;
+
+    // next will try to parse the next path leg and return
+    // if it returns None, it means the input is over.
+    // if it returns Some(Err(..)), it means the format is error.
+    // if it returns Some(Ok(..)), it means the
```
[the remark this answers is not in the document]
YangKeao: Fixed.

YangKeao:
PTAL @wshwsh12 @gengliqi

YangKeao:
@xhebox PTAL

xhebox (approved):

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,331 @@ impl PathExpression {
...
+                    Some((pos, _)) => {
+                        return Err(box_json_path_err!(pos));
+                    }
+                    None => {
+                        return Err(box_json_path_err!(self.current_index()));
+                    }
+                }
+                self.char_iterator.next().unwrap();
+
+                let index = self.input[number_start..number_end].parse::<i32>().unwrap();
```
sticnarf, line 306: If it overflows, it will make TiKV panic?

sticnarf: And I'm not sure what the max acceptable index should be. I try MySQL 5.7.39 and 8.0.30 in docker, the max allowed index is `2^63-1` and `2^32-1` respectively...

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -175,76 +416,164 @@ mod tests {
...
+            (
+                "$**",
+                Some("Invalid JSON path expression. The error is around character position 3."),
+                None,
+            ),
+            (
+                "$.\"a\\t\"",
+                Some("Invalid JSON path expression. The error is around character position 4."),
+                None,
+            ),
```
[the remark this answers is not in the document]
YangKeao: OK! Add two tests: `$ .a $` and `$ .a . $`, the second one is valid :smiley:

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
YangKeao: Nice catch! But I didn't modify the data type, so the upper limit of current implementation is still `2^31-1`. We are using signed type here because we use the `-1` as a special digit represent `*`. It should be refractored to be an `enum {Asterisk, Index(u64)}`. I'll add an issue for it.

sticnarf (approved):
LGTM

YangKeao:
@wshwsh12 @gengliqi PTAL

gengliqi (commented):
One comment just for coding style: Use `///` instead of `//` to document symbols(like functions or variables in struct/enum) and capitalize the first letter in comments. (Most codes I have seen obey this rule)
FYI: https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments

YangKeao:
> One comment just for coding style: Use /// instead of // to document symbols(like functions or variables in struct/enum) and capitalize the first letter in comments. (Most codes I have seen obey this rule)

@gengliqi Good! Fixed :heart:

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,333 @@ impl PathExpression {
...
-        Some(i) => i,
-        None => return Err(box_err!("Invalid JSON path: {}", path_expr)),
-    };
-    if path_expr
-        .index(0..dollar_index)
-        .char_indices()
-        .any(|(_, c)| !c.is_ascii_whitespace())
-    {
-        return Err(box_err!("Invalid JSON path: {}", path_expr));
+// the position is counted from 1
```
gengliqi, line 68: should also be changed here.

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,333 @@ impl PathExpression {
...
+
+    char_iterator: Peekable<CharIndices<'a>>,
+}
+
+struct Position {
+    start: usize,
+    end: usize,
+}
+
+// PathExpressionToken represents a section in path expression and its position
```
gengliqi, line 89: ditto.

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,333 @@ impl PathExpression {
...
+enum PathExpressionToken {
+    Leg((PathLeg, Position)),
+    // represents the beginning "$" in the expression
```
gengliqi, line 92: ditto.

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -73,82 +65,333 @@ impl PathExpression {
...
+    Start(Position),
+}
+
+impl<'a> Iterator for PathExpressionTokenizer<'a> {
+    type Item = Result<PathExpressionToken>;
+
+    /// Next will try to parse the next path leg and return
+    /// if it returns None, it means the input is over.
+    /// if it returns Some(Err(..)), it means the format is error.
+    /// if it returns Some(Ok(..)), it represents the next token.
```
gengliqi, line 102:
```suggestion
    /// Next will try to parse the next path leg and return.
    /// If it returns None, it means the input is over.
    /// If it returns Some(Err(..)), it means the format is error.
    /// If it returns Some(Ok(..)), it represents the next token.
```

gengliqi (approved):

YangKeao:
@gengliqi Thanks for pointing out :heart: . Fixed :man_facepalming:

PR #13320, opened 2022-08-19
raftstore: move snapshot raft state to raftdb

5kbpers:

Issue Number: Ref https://github.com/tikv/tikv/issues/12901

What's Changed:

```commit-message
Move snapshot raft state to raftdb
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
Move snapshot raft state from kvdb to raftdb
```

### DIFF
--- a/components/engine_panic/src/raft_engine.rs
+++ b/components/engine_panic/src/raft_engine.rs
@@ -52,6 +52,10 @@ impl RaftEngineReadOnly for PanicEngine {
     fn get_apply_state(&self, raft_group_id: u64) -> Result<Option<RaftApplyState>> {
         panic!()
     }
+
+    fn get_snapshot_raft_state(&self, raft_group_id: u64) -> Result<Option<RaftLocalState>> {
+        panic!()
+    }
 }
 
 impl RaftEngineDebug for PanicEngine {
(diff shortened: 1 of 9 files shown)


BusyJay:
Why is snapshot raft state still necessary?

5kbpers:
> Why is snapshot raft state still necessary?

Do you mean merge it to raft state?  I don't want to introduce other changes like changes related to recovering in this PR, maybe could do it later.

BusyJay:
> Do you mean merge it to raft state? I don't want to introduce other changes like changes related to recovering in this PR, maybe could do it later.

What I mean is now that all states are in raftdb, applying snapshot doesn't need to be two steps. So the reason of snapshot raft state doesn't exist anymore, and there is no need to recover raft state from snapshot raft state at all.

5kbpers:
> > Do you mean merge it to raft state? I don't want to introduce other changes like changes related to recovering in this PR, maybe could do it later.
> 
> What I mean is now that all states are in raftdb, applying snapshot doesn't need to be two steps. So the reason of snapshot raft state doesn't exist anymore, and there is no need to recover raft state from snapshot raft state at all.

OK I got it, so can just remove snapshot raft state after changing the process of recovering.

PR #11922, opened 2022-01-27
gc_worker: Limit the key range to scan for GcKeys tasks

MyonKeminta:
Signed-off-by: MyonKeminta <MyonKeminta@users.noreply.github.com>

### What is changed and how it works?

Close #11902
Close #11752
What's Changed: Limits the range to scan when processing `GcKeys` tasks in GC Worker. Hopefully this may reduce unnecessary tombstone scan in GcWorker that slows down GC tasks.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- PR to update `pingcap/tidb-ansible`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test

Side effects

-

### Release note 

```release-note
None
```

PR #13322, opened 2022-08-22
gc_worker: use async_snapshot instead of raw API in GC

### LINKED ISSUE #13319: Use stale read to acquire snapshot instead of raw APIs
## Development Task

In the past, we introduce `snapshot_on_kv_engine` and `modify_on_kv_engine` to avoid breaking hibernate regions when doing gc. But now we support stale read, if we always use ts 0 to acquire snapshot, then it will never wake up any regions. So `snapshot_on_kv_engine` can be removed. If we can get the snapshot, then there must be a way to get its underlying storage engine, hence `modify_on_kv_engine` is also unnecessary. There two methods can be removed and make code cleaner.

SpadeA-Tang:

Issue Number: Close #13319

What's Changed:

In the past, we have some raw APIs in `Engine` trait to avoid breaking hinerrnate regions when doing GC works: `snapshot_on_kv_engine` and `modify_on_kv_engine`.

Now, we have supported stale read and so we can use stale read + `start_ts` 0 to acquire the snapshot which is implemented in this PR. In addition, this PR provider a way to fetch the inner kv engine from the snapshot.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,128 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
             }
         }
     }
 }
 
+// If `region_or_provider` is Region, it means all keys are located in the same
+// region. If `region_or_provider` is RegionInfoProvider, then we should find
+// all regions of thoes keys.
+// We return an iterator which yeild item of `Key` and the region taht the key
+// is located. We also return the reion of the first key.
```
MyonKeminta, line 275:
```suggestion
// We return an iterator which yields items of `Key` and the region that the key
// is located. We also return the region of the first key.
```

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,128 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+                        ));
+                    }
+                };
 
-            let keys = keys.into_iter().peekable();
-            return Ok(Box::new(KeysInRegions { keys, regions }));
+                let keys = keys.into_iter().peekable();
+                Ok((Box::new(KeysInRegions { keys, regions }), region))
+            } else {
+                // We only have on key.
```
MyonKeminta, line 314:
```suggestion
                // We only have one key.
```

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,128 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+                    .peekable();
+
+                let region = match regions.peek() {
+                    Some(region) => region.clone(),
+                    None => {
+                        return Err(box_err!(
+                            "No region has been found for range [{:?}, {:?})",
+                            start,
+                            end
+                        ));
```
MyonKeminta, line 307: I think this case actually means nothing needs to be processed, which is not an error.

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,128 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+                    }
+                };
 
-            let keys = keys.into_iter().peekable();
-            return Ok(Box::new(KeysInRegions { keys, regions }));
+                let keys = keys.into_iter().peekable();
+                Ok((Box::new(KeysInRegions { keys, regions }), region))
```
MyonKeminta, line 312: Did you consider the case that the first key isn't in any region currently in this TiKV instance, in which case we expect to skip the key?

components/tikv_kv/src/lib.rs
```diff
@@ -252,6 +252,8 @@ pub struct SnapContext<'a> {
     pub pb_ctx: &'a Context,
     pub read_id: Option<ThreadReadId>,
     pub start_ts: TimeStamp,
+    // `gc` is used by gc oeprations to bypass the start_ts check.
+    pub gc: bool,
```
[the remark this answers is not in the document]
SpadeA-Tang: Good suggestion.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -106,6 +107,13 @@ impl RangeKey {
             RangeKey::Finite(key)
         }
     }
+
+    pub fn into_key(self) -> Vec<u8> {
```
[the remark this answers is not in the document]
SpadeA-Tang: I forget to delete it.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -712,23 +720,55 @@ impl RegionInfoProvider for RegionInfoAccessor {
...
+            .map(|region_info| region_info.region)
+            .collect_vec())
+    }
+
+    fn seek_region(&self, from: &[u8], callback: SeekRegionCallback) -> Result<()> {
+        let region_infos = self.0.lock().unwrap();
+        let from = RangeKey::from_start_key(from.to_vec());
+        let mut iter = region_infos.iter().filter(|&region_info| {
+            RangeKey::from_start_key(region_info.region.get_start_key().to_vec()) <= from
+                && from < RangeKey::from_end_key(region_info.region.get_end_key().to_vec())
```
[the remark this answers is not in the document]
SpadeA-Tang: Yes, good catch. I will modify it.

tests/integrations/storage/test_storage.rs
```diff
@@ -714,12 +729,13 @@ pub fn test_txn_store_gc_multiple_keys_cluster_storage(n: usize, prefix: String)
...
     let mut last_region = cluster.get_region(b"");
-    store.gc_ok_for_cluster(&mut cluster, b"", 30);
+    store.gc_ok_for_cluster(&mut cluster, b"", last_region.clone(), 30);
     for k in &keys {
         // clear data whose commit_ts < 30
         let region = cluster.get_region(k.as_bytes());
         if last_region != region {
-            store.gc_ok_for_cluster(&mut cluster, k.as_bytes(), 30);
+            store.gc_ok_for_cluster(&mut cluster, k.as_bytes(), region.clone(), 30);
+            println!("Last region {:?} region {:?}", last_region, region);
```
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

src/server/gc_worker/gc_manager.rs
```diff
@@ -545,21 +546,20 @@ impl<S: GcSafePointProvider, R: RegionInfoProvider + 'static, E: KvEngine> GcMan
...
+        if region.is_none() {
+            return Ok(None);
         };
+        let region = region.unwrap();
 
-        let hex_start = format!("{:?}", log_wrappers::Value::key(&start));
-        let hex_end = format!("{:?}", log_wrappers::Value::key(&end));
+        let hex_start = format!("{:?}", log_wrappers::Value::key(region.get_start_key()));
+        let hex_end = format!("{:?}", log_wrappers::Value::key(region.get_end_key()));
         debug!("trying gc"; "start_key" => &hex_start, "end_key" => &hex_end);
```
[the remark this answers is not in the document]
SpadeA-Tang: okay

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: So we can return an Result<Option<...>>, and return directly when when the option is None?

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: I have made some modifications for this. PTAL.

tonyxuqqi:
> ### What is changed and how it works?
> * Performance regression
>   
>   * Consumes more CPU
>   * Consumes more MEM
> * Breaking backward compatibility
> 
How much is the performance regression?

src/server/raftkv.rs
```diff
@@ -201,7 +200,7 @@ where
         cb: Callback<CmdRes<E::Snapshot>>,
     ) -> Result<()> {
         let mut header = self.new_request_header(ctx.pb_ctx);
-        if ctx.pb_ctx.get_stale_read() && !ctx.start_ts.is_zero() {
+        if ctx.pb_ctx.get_stale_read() && (!ctx.start_ts.is_zero() || ctx.is_snap_for_gc) {
```
tonyxuqqi, line 203: Do you know originally why ctx.start_ts.is_zero() is not allowed?  I'm not sure if there's other changes needed to accommodate with this change.

SpadeA-Tang: The original checks for ctx.start_ts.is_zero() is to avoid the casee: when TiKV start/restart, the safe_ts will be 0, and TiDB sends stale read with 0 start_ts.

src/server/gc_worker/gc_worker.rs
```diff
@@ -640,7 +778,11 @@ where
         self.flow_info_sender
             .send(FlowInfo::BeforeUnsafeDestroyRange(ctx.region_id))
             .unwrap();
-        let local_storage = self.engine.kv_engine();
+
+        let region = seek_region(store_id, start_key.as_encoded(), regions_provider)?;
```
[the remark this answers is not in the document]
SpadeA-Tang: It has not considered multi-rocks now. The intention here is just to fetch the kv_engine.

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,107 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+                    .peekable();
+
+                let keys = keys.into_iter().peekable();
+                Ok(Box::new(KeysInRegions { keys, regions }))
+            } else {
+                // We only have one key.
+                let region = seek_region(
+                    store_id,
+                    keys.first().unwrap().as_encoded(),
+                    region_provider,
```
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,107 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
-            let end = keys.last().unwrap().as_encoded();
-            let regions = box_try!(region_info_provider.get_regions_in_range(start, end))
-                .into_iter()
-                .filter(move |r| find_peer(r, store_id).is_some())
-                .peekable();
-
-            let keys = keys.into_iter().peekable();
-            return Ok(Box::new(KeysInRegions { keys, regions }));
+    region_or_provider: Either<Region, Arc<dyn RegionInfoProvider>>,
+) -> Result<Box<dyn Iterator<Item = (Key, Region)>>> {
```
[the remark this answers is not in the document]
SpadeA-Tang: Good suggestion. I have changed it. PTAL again.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -450,6 +462,19 @@ impl RegionCollector {
...
     }
 
+    pub fn handle_find_region_by_key(&self, key: Vec<u8>, callback: Callback<Option<Region>>) {
+        let key = RangeKey::from_start_key(key);
+        let mut region = None;
+        if let Some((_, region_id)) = self.region_ranges.range((Excluded(key), Unbounded)).next() {
+            region = self
+                .regions
+                .get(region_id)
+                .map(|region_info| region_info.region.clone());
```
MyonKeminta, line 472: I think you need to check if the region contains the specified key. If the region that contains the key is not present in the collection, it seems that the first region that has keys greater than the given key will be found, which I think is not what you expect.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -709,26 +741,80 @@ impl RegionInfoProvider for RegionInfoAccessor {
...
 impl RegionInfoProvider for MockRegionInfoProvider {
     fn get_regions_in_range(&self, _start_key: &[u8], _end_key: &[u8]) -> Result<Vec<Region>> {
-        Ok(self.0.lock().unwrap().clone())
+        Ok(self
+            .0
+            .lock()
+            .unwrap()
+            .clone()
+            .into_iter()
+            .map(|region_info| region_info.region)
```
MyonKeminta, line 806:
```suggestion
            .iter()
            .map(|region_info| region_info.region.clone())
```

Does this work? I think it doesn't need to allocate new Vec twice (though it's not important for test code).

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -709,26 +741,80 @@ impl RegionInfoProvider for RegionInfoAccessor {
...
+            .collect_vec())
+    }
+
+    fn seek_region(&self, from: &[u8], callback: SeekRegionCallback) -> Result<()> {
+        let region_infos = self.0.lock().unwrap();
+        let mut iter = region_infos.iter().filter(|&region_info| {
+            RangeKey::from_start_key(region_info.region.get_start_key().to_vec())
+                <= RangeKey::from_start_key(from.to_vec())
```
MyonKeminta, line 814: It seems that you returned the region that contains `from` and all regions *before* `from`, however we expects the region that contains `from` and all regions *after* `from`. I think what you need is:

```suggestion
            RangeKey::from_end_key(region_info.region.get_end_key().to_vec())
                > RangeKey::from_start_key(from.to_vec())
```

MyonKeminta: Actually I think it might be a good idea to add some simple tests for it, though it's a mock.

components/raftstore/src/coprocessor/region_info_accessor.rs
SpadeA-Tang: Good catch.

components/raftstore/src/coprocessor/region_info_accessor.rs
SpadeA-Tang: Yes, good suggestion.

components/raftstore/src/coprocessor/region_info_accessor.rs
SpadeA-Tang: Good suggestion.

Cargo.toml
```diff
@@ -281,24 +281,24 @@ members = [
 default-members = ["cmd/tikv-server", "cmd/tikv-ctl"]
 
 [profile.dev.package.grpcio-sys]
-debug = false
-opt-level = 1
+debug = true
```
BusyJay, line 284: This settings are added for a reason. They are used to make target directory small.

Cargo.toml
```diff
@@ -281,24 +281,24 @@ members = [
...
 [profile.dev.package.tests]
-debug = 1
-opt-level = 1
+debug = true
+opt-level = 0
 
 [profile.dev]
 opt-level = 0
-debug = 0
+debug = true
```
BusyJay, line 301: I think the CONTRIBUTING.md has a clear guide how to enable debug info.

Cargo.toml
SpadeA-Tang: My fault. I change it to debug and forget to change it back.

Cargo.toml
SpadeA-Tang: ditto

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -450,6 +462,32 @@ impl RegionCollector {
...
+            .range((Excluded(key.clone()), Unbounded))
+            .next()
+        {
+            region = self
+                .regions
+                .get(region_id)
+                .map(|region_info| region_info.region.clone())
+                .and_then(|region| {
+                    if RangeKey::from_start_key(region.get_start_key().to_vec()) <= key
+                        && key < RangeKey::from_end_key(region.get_end_key().to_vec())
```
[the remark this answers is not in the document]
SpadeA-Tang: Yes, good catch.

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,127 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+            } else {
+                // We only have one key.
+                let key = keys.first().unwrap().as_encoded();
+                let region = box_try!(region_provider.find_region_by_key(key));
+
+                if region.is_none() {
+                    return Err(box_err!(
+                        "The region that contains {:?} has not been found",
+                        key
+                    ));
```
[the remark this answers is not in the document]
SpadeA-Tang: Done.

src/server/gc_worker/gc_worker.rs
```diff
@@ -349,29 +441,35 @@ where
...
+        kv_engine: EK,
+    ) -> Result<()> {
         let write_size = txn.write_size();
-        let modifies = txn.into_modifies();
+        let mut modifies = txn.into_modifies();
         if !modifies.is_empty() {
             limiter.blocking_consume(write_size);
-            engine.modify_on_kv_engine(modifies)?;
+            engine.adjust_modify(&mut modifies);
+            write_modifies(&kv_engine, modifies)?;
```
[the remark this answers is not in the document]
SpadeA-Tang: Done.

MyonKeminta (approved):

components/raftstore/src/store/worker/read.rs
```diff
@@ -782,7 +782,6 @@ where
                     // Replica can serve stale read if and only if its `safe_ts` >= `read_ts`
                     RequestPolicy::StaleRead => {
                         let read_ts = decode_u64(&mut req.get_header().get_flag_data()).unwrap();
-                        assert!(read_ts > 0);
```
BusyJay, line 785: We need to make sure it's OK to remove the assert.

BusyJay (changes requested):

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Without setting is_snap_for_gc in SnapContext, stale read behaves the same with before.

tonyxuqqi (approved):

components/raftstore/src/store/worker/read.rs
tonyxuqqi: I would suggest to somehow pass is_snap_for_gc to here and keep the assert when is_snap_for_gc is false/

SpadeA-Tang: > I would suggest to somehow pass is_snap_for_gc to here and keep the assert when is_snap_for_gc is false/

Good suggestion.

components/tikv_kv/src/lib.rs
```diff
@@ -251,7 +251,8 @@ impl WriteData {
 pub struct SnapContext<'a> {
     pub pb_ctx: &'a Context,
     pub read_id: Option<ThreadReadId>,
-    pub start_ts: TimeStamp,
+    // When start_ts is None and `stale_read` is true, we can make sure to get the snapshot.
```
BusyJay, line 254:
```suggestion
    // When start_ts is None and `stale_read` is true, it means acquire a snapshot without any consistency guarantee.
```

components/tikv_kv/src/lib.rs
```diff
@@ -363,19 +364,27 @@ pub trait Snapshot: Sync + Send + Clone {
...
     // The maximum key can be fetched from the snapshot should less than the upper
     // bound.
     #[inline]
     fn upper_bound(&self) -> Option<&[u8]> {
         None
     }
 
+    #[inline]
+    fn inner_engine(&self) -> Self::E {
+        unimplemented!()
```
BusyJay, line 385: Why?

src/server/raftkv.rs
```diff
@@ -201,10 +200,12 @@ where
         cb: Callback<CmdRes<E::Snapshot>>,
     ) -> Result<()> {
         let mut header = self.new_request_header(ctx.pb_ctx);
-        if ctx.pb_ctx.get_stale_read() && !ctx.start_ts.is_zero() {
+        if ctx.pb_ctx.get_stale_read()
+            && (ctx.start_ts.is_none() || !ctx.start_ts.as_ref().unwrap().is_zero())
```
BusyJay, line 204: You mean `ctx.start_ts.map_or(true, |ts| !ts.is_zero())`?

src/server/raftkv.rs
```diff
@@ -427,8 +415,12 @@ where
 
         let mut req = Request::default();
         req.set_cmd_type(CmdType::Snap);
-        if !ctx.key_ranges.is_empty() && !ctx.start_ts.is_zero() {
-            req.mut_read_index().set_start_ts(ctx.start_ts.into_inner());
+        if !ctx.key_ranges.is_empty()
+            && ctx.start_ts.is_some()
```
BusyJay, line 419: You mean `ctx.start_ts.map_or(false, |ts| !ts.is_zero())`?

src/storage/mvcc/reader/reader.rs
```diff
@@ -124,6 +124,9 @@ pub struct MvccReader<S: EngineSnapshot> {
     lock_cursor: Option<Cursor<S::Iter>>,
     write_cursor: Option<Cursor<S::Iter>>,
 
+    lower_bound: Option<Key>,
```
BusyJay, line 127: Why not use `ForwardScanner`? It can use customized boundaries.

components/tikv_kv/src/lib.rs
```diff
@@ -265,10 +266,9 @@ pub trait Engine: Send + Clone + 'static {
     /// Local storage engine.
     fn kv_engine(&self) -> Self::Local;
 
-    fn snapshot_on_kv_engine(&self, start_key: &[u8], end_key: &[u8]) -> Result<Self::Snap>;
-
-    /// Write modifications into internal local engine directly.
-    fn modify_on_kv_engine(&self, modifies: Vec<Modify>) -> Result<()>;
+    /// Do some specialization encoding works on `Modify`s according to the
+    /// different implementation of `Engine`
+    fn adjust_modify(&self, _modifies: &mut Vec<Modify>) {}
```
BusyJay, line 271: Bad name.

components/tikv_kv/src/lib.rs
```diff
@@ -265,10 +266,9 @@ pub trait Engine: Send + Clone + 'static {
     /// Local storage engine.
     fn kv_engine(&self) -> Self::Local;
 
-    fn snapshot_on_kv_engine(&self, start_key: &[u8], end_key: &[u8]) -> Result<Self::Snap>;
-
-    /// Write modifications into internal local engine directly.
-    fn modify_on_kv_engine(&self, modifies: Vec<Modify>) -> Result<()>;
+    /// Do some specialization encoding works on `Modify`s according to the
```
BusyJay, line 269:
```suggestion
    /// Do some specialization encoding works on `modifies`s according to the
```

BusyJay: Should mention when this method should be used.

components/tikv_kv/src/lib.rs
```diff
@@ -363,19 +364,27 @@ pub trait Snapshot: Sync + Send + Clone {
...
+
     // The maximum key can be fetched from the snapshot should less than the upper
     // bound.
     #[inline]
     fn upper_bound(&self) -> Option<&[u8]> {
         None
     }
 
+    #[inline]
+    fn inner_engine(&self) -> Self::E {
```
BusyJay, line 384: Missing docs.

components/tikv_kv/src/lib.rs
SpadeA-Tang: Any sugguestion?

src/storage/mvcc/reader/reader.rs
SpadeA-Tang: Why we can use it? It cannot use snapshot.

src/server/raftkv.rs
SpadeA-Tang: yes

src/storage/mvcc/reader/reader.rs
BusyJay: It can, otherwise it won't be defined as `struct ForwardScanner<S: Snapshot, P: ScanPolicy<S>>`.

SpadeA-Tang: I don't think we should use ForwardScanner here, GC is highly combined with MvccReader.

BusyJay: Then I suggest to remove these boundaries. It's limited by the region ranges already.

SpadeA-Tang: The boundaries is used to avoid scanning tombstones unnecessarily. See https://github.com/tikv/tikv/pull/11922.

components/engine_panic/src/snapshot.rs
```diff
@@ -9,14 +9,21 @@ use crate::{db_vector::PanicDbVector, engine::PanicEngine};
 #[derive(Clone, Debug)]
 pub struct PanicSnapshot;
 
-impl Snapshot for PanicSnapshot {}
+impl Snapshot for PanicSnapshot {
+    type E = PanicEngine;
+
+    fn inner_engine(&self) -> Self::E {
+        PanicEngine {}
```
BusyJay, line 16: `{}` is unnecessary.

components/engine_rocks/src/snapshot.rs
```diff
@@ -30,9 +30,20 @@ impl RocksSnapshot {
...
+        self.db.clone()
+    }
 }
 
-impl Snapshot for RocksSnapshot {}
+impl Snapshot for RocksSnapshot {
+    type E = RocksEngine;
+
+    #[inline]
+    fn inner_engine(&self) -> Self::E {
```
BusyJay, line 43: People may assume the returned value is reliable. But actually it's not. All fields except `db` are initialized with the default value.

components/raftstore/src/coprocessor/region_info_accessor.rs
```diff
@@ -709,31 +752,101 @@ impl RegionInfoProvider for RegionInfoAccessor {
                 })
             })
     }
+
+    fn find_region_by_key(&self, key: &[u8]) -> Result<Option<Region>> {
+        let (tx, rx) = mpsc::channel();
+        let msg = RegionInfoQuery::GetRegionByKey {
```
BusyJay, line 758: Why not use `RegionInfoQuery::SeekRegion`?

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,128 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+    ctx.stale_read = true;
+
+    let peers: Vec<_> = region
+        .peers
+        .iter()
+        .filter(|peer| peer.store_id == store_id)
+        .collect();
+
+    if peers.len() == 1 {
+        ctx.peer = SingularPtrField::some(peers[0].clone());
```
BusyJay, line 373: You mean `ctx.set_peer(peers[0].clone())`?

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +252,128 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+fn init_snap_ctx(store_id: u64, region: &Region) -> Context {
+    let mut ctx = Context::default();
+    ctx.region_id = region.id;
+    ctx.region_epoch = region.region_epoch.clone();
+    ctx.stale_read = true;
+
+    let peers: Vec<_> = region
+        .peers
+        .iter()
+        .filter(|peer| peer.store_id == store_id)
```
BusyJay, line 369: Why not use `.find`?

components/raftstore/src/coprocessor/region_info_accessor.rs
SpadeA-Tang: SeekRegion finds all regions whose end key is larger than `key` where `key` may not be located in any of them. It's not convenient to use it in this case.

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: yes

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: Good suggestion.

components/raftstore/src/coprocessor/region_info_accessor.rs
BusyJay: Isn't it just
```
self.seek_region(key, Box::new(|iter| {
   if let Some(info) = iter.next() && info.region.get_start_key() < key {
        tx.send(info.region.clone());
    }
}))
```

SpadeA-Tang: But, after encapsulating it in `RegionInfoProvider`, it just needs one line.

BusyJay: So you use 23 lines + 2 lines + 21 lines to encapsulate 4 lines?

SpadeA-Tang: find_region_by_key is a common pattern that may be used in many places.

src/server/gc_worker/gc_worker.rs
```diff
@@ -349,29 +435,39 @@ where
...
-            engine.modify_on_kv_engine(modifies)?;
+            engine.adjust_modify(&mut modifies);
+            write_modifies(&kv_engine, modifies)?;
         }
         Ok(())
     }
 
-    fn gc(&mut self, start_key: &[u8], end_key: &[u8], safe_point: TimeStamp) -> Result<()> {
-        if !self.need_gc(start_key, end_key, safe_point) {
+    fn gc(&mut self, store_id: u64, region: Region, safe_point: TimeStamp) -> Result<()> {
```
BusyJay, line 458: Why not make `store_id` a field of gc runner or gc worker? So you don't need to pass the value around as it never changes once initialized.

src/server/gc_worker/gc_worker.rs
```diff
@@ -383,166 +479,236 @@ where
...
-                reader = MvccReader::new(snapshot, Some(ScanMode::Forward), false);
             }
         }
-        Self::flush_txn(txn, &self.limiter, &self.engine)?;
+
+        Self::flush_txn(txn, &self.limiter, &self.engine, kv_engine)?;
         Ok((handled_keys, wasted_keys))
     }
 
+    fn get_reader_and_kv_engine(
```
BusyJay, line 595:
```suggestion
    fn create_reader(
```

components/engine_traits/src/snapshot.rs
```diff
@@ -12,4 +12,9 @@ pub trait Snapshot
 where
     Self: 'static + Peekable + Iterable + Send + Sync + Sized + Debug,
 {
+    type E: KvEngine;
+
+    /// The returned engine is not reliable. It can only be used by write
```
BusyJay, line 17:
```suggestion
    /// The returned engine is not reliable. It can only be used for write
```

components/raftstore/src/coprocessor/region_info_accessor.rs
BusyJay: If it's a common pattern, then it won't be added 3 years after `SeekRegion` is introduced.

components/tikv_kv/src/lib.rs
```diff
@@ -265,10 +267,10 @@ pub trait Engine: Send + Clone + 'static {
     /// Local storage engine.
     fn kv_engine(&self) -> Self::Local;
 
-    fn snapshot_on_kv_engine(&self, start_key: &[u8], end_key: &[u8]) -> Result<Self::Snap>;
-
-    /// Write modifications into internal local engine directly.
-    fn modify_on_kv_engine(&self, modifies: Vec<Modify>) -> Result<()>;
+    /// Do some specialization encoding works on `modifies`s according to the
+    /// different implementation of `Engine` before flushing them to the
+    /// underlying kv engine.
```
BusyJay, line 272:
```suggestion
    /// Change the encoding of `Modify` so that they can be written to local
    /// engine directly. After the method, `modifies` should not be passed to
    /// any write methods of `Engine` trait any more.
```

components/tikv_kv/src/lib.rs
BusyJay:
```suggestion
    fn encode_in_place(&self, _modifies: &mut Vec<Modify>) {}
```

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +247,114 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
+        Some((Box::new(keys.into_iter()), region))
     }
 }
 
+// If `region_or_provider` is Region, it means all keys are located in the same
+// region. If `region_or_provider` is RegionInfoProvider, then we should find
+// all regions of thoes keys.
+// We return an iterator which yields items of `Key` and the region taht the key
+// is located.
 fn get_keys_in_regions(
```
BusyJay, line 285: This seems a bit over complicated. Why not just use 2 for loops?
```
for region in regions {
    let keys = get_keys_in_region(&keys, region);
    // process keys
}
```

src/server/gc_worker/gc_worker.rs
```diff
@@ -164,14 +163,14 @@ where
...
-                start_key,
-                end_key,
-                safe_point,
-                ..
+                region, safe_point, ..
             } => f
                 .debug_struct("Gc")
-                .field("start_key", &log_wrappers::Value::key(start_key))
-                .field("end_key", &log_wrappers::Value::key(end_key))
+                .field(
```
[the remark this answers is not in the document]
SpadeA-Tang: To be consistent with before. Other infomation in Region may not be needed.

tests/failpoints/cases/test_stale_read.rs
```diff
@@ -526,3 +529,47 @@ fn test_stale_read_during_merging_2() {
 
     fail::remove(leader_commit_prepare_merge_fp);
 }
+
+#[test]
+fn test_stale_read_with_ts0() {
```
BusyJay, line 534: If no failpoint is used, then it should be put in integration directory.

tests/integrations/storage/test_storage.rs
```diff
@@ -698,7 +704,14 @@ pub fn test_txn_store_gc_multiple_keys_single_storage(n: usize, prefix: String)
         store.put_ok(k.as_bytes(), b"v1", 5, 10);
         store.put_ok(k.as_bytes(), b"v2", 15, 20);
     }
-    store.gc_ok(30);
+
+    let store_id = 1;
+    let mut region = metapb::Region::default();
+    region.set_peers(RepeatedField::from_vec(vec![metapb::Peer {
```
BusyJay, line 710: You mean `region.mut_peers().push(new_peer(store_id, 0))`?

There are still many places using `RepeatedField`.

components/test_storage/Cargo.toml
```diff
@@ -26,6 +26,7 @@ collections = { path = "../collections" }
 futures = "0.3"
 kvproto = { git = "https://github.com/pingcap/kvproto.git" }
 pd_client = { path = "../pd_client", default-features = false }
+protobuf = { version = "2.8", features = ["bytes"] }
```
BusyJay, line 29: This dependency should not be added.

tests/failpoints/cases/test_stale_read.rs
SpadeA-Tang: Done

tests/integrations/storage/test_storage.rs
SpadeA-Tang: Done

components/test_storage/src/assert_storage.rs
```diff
@@ -197,16 +208,17 @@ impl<F: KvFormat> AssertionStorage<SimulateEngine, F> {
         &mut self,
         cluster: &mut Cluster<ServerCluster>,
         region_key: &[u8],
+        mut region: metapb::Region,
         safe_point: impl Into<TimeStamp>,
     ) {
         let safe_point = safe_point.into();
         for _ in 0..3 {
-            let ret = self.store.gc(self.ctx.clone(), safe_point);
+            let ret = self.store.gc(region, self.ctx.clone(), safe_point);
```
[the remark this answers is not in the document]
SpadeA-Tang: The region may be changed.

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +247,129 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
 }
 
 impl<R: Iterator<Item = Region>> Iterator for KeysInRegions<R> {
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

src/server/gc_worker/gc_worker.rs
```diff
@@ -247,40 +247,129 @@ struct KeysInRegions<R: Iterator<Item = Region>> {
...
-                .filter(move |r| find_peer(r, store_id).is_some())
-                .peekable();
+// Return regions that keys are related to.
+// If `region_or_provider` is Region, it means all keys are located in the same
+// region. If `region_or_provider` is RegionInfoProvider, then we should find
+// all regions of thoes keys.
+fn get_regions_for_gc(
+    store_id: u64,
+    keys: &Vec<Key>,
+    region_or_provider: Either<Region, Arc<dyn RegionInfoProvider>>,
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

src/server/gc_worker/gc_worker.rs
```diff
@@ -400,149 +456,224 @@ where
...
+            Either::Left(region) => vec![region].into_iter().peekable(),
+            Either::Right(region_provider) => get_regions_for_gc(store_id, &keys, region_provider)?
+                .into_iter()
+                .peekable(),
         };
 
-        let (mut handled_keys, mut wasted_keys) = (0, 0);
+        // First item is fetched to initialize the reader and kv_engine
+        let region = regions.peek();
+        if region.is_none() {
```
BusyJay, line 484: `regions.is_empty()` is more straightforward.

src/server/gc_worker/gc_worker.rs
```diff
@@ -400,149 +456,224 @@ where
...
-        let mut next_gc_key = keys.next();
-        while let Some(ref key) = next_gc_key {
-            if let Err(e) = self.gc_key(safe_point, key, &mut gc_info, &mut txn, &mut reader) {
-                GC_KEY_FAILURES.inc();
-                error!(?e; "GC meets failure"; "key" => %key,);
-                // Switch to the next key if meets failure.
-                gc_info.is_completed = true;
+        let mut keys = keys.into_iter().peekable();
+        for region in regions {
+            if !first_iteration {
```
BusyJay, line 499: Why not
```
for region in regions {
    let (reader, kv_engine) = self.create_reader()?;
    let txn = Self::new_txn();
    xxxx;
    Self::flush_txn();
}
```

When one tablet per region, you can't flush transaction across tablet.

SpadeA-Tang: But it can also have single-rocksdb setting.

components/engine_rocks/src/snapshot.rs
```diff
@@ -30,6 +30,10 @@ impl RocksSnapshot {
             }
         }
     }
+
+    pub fn db(&self) -> Arc<DB> {
```
BusyJay, line 34: Not necessary anymore, people can use `as_inner`.

src/server/gc_worker/gc_worker.rs
```diff
@@ -727,15 +800,31 @@ where
...
     ) -> Result<Vec<LockInfo>> {
-        let snap = self
-            .engine
-            .snapshot_on_kv_engine(start_key.as_encoded(), &[])
-            .unwrap();
-        let mut reader = MvccReader::new(snap, Some(ScanMode::Forward), false);
-        let (locks, _) = reader
-            .scan_locks(Some(start_key), None, |l| l.ts <= max_ts, limit)
-            .map_err(TxnError::from_mvcc)?;
+        let regions = box_try!(regions_provider.get_regions_in_range(start_key.as_encoded(), &[]))
```
BusyJay, line 805: We need to evaluate the performance impact of changing from one seek to multiple seek.

src/server/gc_worker/gc_worker.rs
```diff
@@ -727,15 +800,31 @@ where
...
+                    first_round = false;
+                    start_key.clone()
+                } else {
+                    Key::from_raw(region.get_start_key())
+                }
+            };
+            let snap = self.get_snapshot(self.store_id, &region)?;
+            let mut reader = MvccReader::new(snap, Some(ScanMode::Forward), false);
+            let (locks_this_region, _) = reader
+                .scan_locks(Some(&start_key), None, |l| l.ts <= max_ts, limit)
```
BusyJay, line 823: /cc @lhy1024 Will it impact the statistics? As there will be many scan ops and end key is always None.

BusyJay (commented):
Rest LGTM

components/server/src/server.rs
```diff
@@ -591,22 +591,14 @@ where
...
-        let mut gc_worker = GcWorker::new(
+        let gc_worker = GcWorker::new(
             engines.engine.clone(),
             self.router.clone(),
             self.flow_info_sender.take().unwrap(),
             self.config.gc.clone(),
             self.pd_client.feature_gate().clone(),
+            Arc::new(self.region_info_accessor.clone()),
         );
-        gc_worker
```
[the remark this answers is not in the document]
MyonKeminta: Starting of the gc_worker is delayed to `init_servers` function, just before `start_auto_gc`. The purpose seems to be getting the store id when starting the gc_worker. I think delaying the starting of gc_worker doesn't cause any severe problem if test passes, but I would suggest wrapping the three "start" functions into a single `start_gc_worker` function.

components/engine_rocks/src/snapshot.rs
SpadeA-Tang: Done

src/server/gc_worker/gc_worker.rs
```diff
@@ -495,7 +490,7 @@ where
         let mut txn = Self::new_txn();
         let mut gc_info = GcInfo::default();
         let mut keys = keys.into_iter().peekable();
-        for region in regions {
+        for region in regions.into_iter() {
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: Scan is also limited by region boundries of snapshot.

src/server/gc_worker/gc_worker.rs
MyonKeminta: As Green GC is going to be deprecated, I think we don't need to concern about it too much.
Maybe it is the time to deprecate it now, since only scanning regions found in region_info_provider may possibly affect the scanning's correctness.

src/server/gc_worker/gc_worker.rs
lhy1024: If this pr only affects gc, it has no effect on hotspot statistics, because pd does not process gc-related statistics for the time being

src/server/gc_worker/gc_worker.rs
BusyJay: Indeed.

BusyJay (approved):

MyonKeminta (approved):

src/server/gc_worker/gc_worker.rs
```diff
@@ -1084,11 +1182,13 @@ where
             applied_lock_collector: None,
             gc_manager_handle: Arc::new(Mutex::new(None)),
             feature_gate,
+            region_info_provider,
         }
     }
 
     pub fn start_auto_gc<S: GcSafePointProvider, R: RegionInfoProvider + Clone + 'static>(
         &self,
+        kv_engine: &E::Local,
```
tabokie, line 1191: @SpadeA-Tang Could you help explain a bit why this is necessary?

SpadeA-Tang: Okay... I have remembered why this change exists. I tried to remove the kv_engine() from Engine trait in some point of time but I gave up this way later without changeing it back. I will propose another PR for reverting it.

Issue #11720, opened 2021-12-24
get the a rather accurate keys of each region than the current approximate keys


winoros:
### Is your feature request related to a problem? Please describe:

To speed up the analyze, we need to control the calculation of the analyze.
From the paper https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.53.1734&rep=rep1&type=pdf, we know that to control the error rate of the histogram, the relation between the sample size and the table size is logarithmic, i.e. When the table size is 1 billion, 0.1 million samples can fit our needs. When the table size is 10 billion, maybe the 0.15 million samples can fit our needs. So the sample rate is not a fixed value, we need to calculate it with the table size.

So we need to know the table size to calculate the sample rate to help a much big table finish its analyze as soon as possible. But the current approximate keys seems to have very bad estimation sometimes, see https://github.com/tikv/tikv/issues/11669.

### Describe the feature you'd like:

We get an accurate or approximate result of the number of keys with a bounded error rate.(`The maximum approximate result / The real key count * 100% = 200% is also acceptable. We need a maximum and minimum error rate to make decision).

### Describe alternatives you've considered:

And i think this would help TiDB to know whether a region has changed a lot to have a better analyze strategy and better cost estimation/count estimation.

PR #13328, opened 2022-08-23
cmd: output key count in ctl region size

tabokie:

Issue Number: Ref #11720

What's Changed:

Approximate key can often be zero for wide table where the write-cf region range doesn't overlap multiple RocksDB table properties. It would be assuring to have ctl also output the accurate key count.

TODO: update kvproto

```commit-message
Output key count in tikv-ctl size
```

### Related changes

### Release note 

```release-note
None
```

### DIFF
--- a/cmd/tikv-ctl/src/cmd.rs
+++ b/cmd/tikv-ctl/src/cmd.rs
@@ -107,7 +107,7 @@ pub enum Cmd {
         #[structopt(subcommand)]
         cmd: RaftCmd,
     },
-    /// Print region size
+    /// Print region size and key count
     Size {
         #[structopt(short = "r")]
         /// Set the region id, if not specified, print all regions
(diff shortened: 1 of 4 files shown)


cmd/tikv-ctl/src/cmd.rs
```diff
@@ -107,7 +107,7 @@ pub enum Cmd {
         #[structopt(subcommand)]
         cmd: RaftCmd,
     },
-    /// Print region size
+    /// Print region size and key count
     Size {
```
Connor1996, line 111: the command is called `size`, but it actually prints size and key count

tabokie: How about adding a new command? I don't want to rename an existing command.

Connor1996: +1

PR #13335, opened 2022-08-24
*: move RegionMeta to raftstore

BusyJay:

Issue Number: Ref #12842 

This PR is extracted from #13334.
What's Changed:
```commit-message
This PR moves `RegionMeta` to raftstore module, so that both v1 and v2
can use it for debugging. Several fields are re-arranged to be more
concise and more informations.
```

### Check List 

Tests 

- Unit test
- Integration test

Side effects

  The response of HTTP API /region/id is changed to include more fields. Since the API is used mainly for debugging, it may breaks some OP scripts.

### Release note 

```release-note
Change the response of peer query to include more information.
```


BusyJay:
PTAL

components/raftstore/src/store/region_meta.rs
```diff
@@ -0,0 +1,260 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```
Connor1996, line 1: 2020 -> 2022

Connor1996 (approved):
the title should be `Introduce RegionMeta for debugging`?

BusyJay:
No, `RegionMeta` is defined already in server module. I just move it to raftstore.

Connor1996:
> No, `RegionMeta` is defined already in server module. I just move it to raftstore.

Seems you didn't remove the origin

5kbpers (approved):

components/raftstore-v2/src/router/response_channel.rs
```diff
@@ -201,10 +213,66 @@ impl<'a, Res> Future for WaitResult<'a, Res> {
...
+    pub async fn result(mut self) -> Option<Res> {
+        WaitResult { core: &self.core }.await
+    }
+}
+
+unsafe impl<Res: Send> Send for BaseSubscriber<Res> {}
+unsafe impl<Res: Send> Sync for BaseSubscriber<Res> {}
+
+/// A base channel that contains most common implementation of channels.
+pub struct BaseChannel<Res> {
```
Connor1996, line 233: miss the review of the previous PR, why use `ManuallyDrop`

BusyJay:
/hold

It was added to control the drop order. But now there is only one field, it's not necessary anymore.

PR #13337, opened 2022-08-24
coprocessor: move task reschedule from runner to scanner

glorv:

Issue Number: ref #13313

What's Changed:

```
```
This is another attempt to alleviate the performance impact of slow big queries. By move task reschedule from `runner` to `scanner` we can control the length of one time slice more precisely. Thus, we can avoid the huge impact caused by single task eaten too mach cpu time.

In my benchmark, after this change, the time slice used by a single poll is much even and the overall performance is much better. The only drawback is that big task will  costs more time to finish because it need more round of scheduling. I think this is expected behavior.

### Check List 

Tests 

- Unit test
- No code

Side effects

- Performance regression
    - long tail slow tasking is likely to be longer than before if the cpu load is high.

### Release note 

```release-note
None
```

### DIFF
--- a/components/tidb_query_common/src/storage/scanner.rs
+++ b/components/tidb_query_common/src/storage/scanner.rs
@@ -9,6 +9,11 @@ use super::{range::*, ranges_iter::*, OwnedKvPair, Storage};
 use crate::error::StorageError;
 
 const KEY_BUFFER_CAPACITY: usize = 64;
+/// Batch executors are run in coroutines. `MAX_TIME_SLICE` is the maximum time
+/// a coroutine can run without being yielded.
+const MAX_TIME_SLICE: Duration = Duration::from_millis(1);
+/// the number of scanned keys that should trigger a reschedule.
+const CHECK_KEYS: usize = 32;
 
 /// A scanner that scans over multiple ranges. Each range can be a point range
 /// containing only one row, or an interval range containing multiple rows.
@@ -31,13 +36,12 @@ pub struct RangesScanner<T> {
     rescheduler: RescheduleChecker,
 }
 
+// TODO: maybe it's better to make it generic to avoid directly depending
+// on yatp's rescheduler.
 struct RescheduleChecker {
     prev_start: Instant,
     prev_key_count: usize,
 }
-
-const MAX_TIME_SLICE: Duration = Duration::from_millis(1);
-const CHECK_KEYS: usize = 32;
 
 impl RescheduleChecker {
     fn new() -> Self {
--- a/components/tidb_query_common/src/storage/scanner.rs
+++ b/components/tidb_query_common/src/storage/scanner.rs
@@ -51,6 +51,7 @@ impl RescheduleChecker {
         }
     }
 
+    #[inline(always)]
     async fn check_reschedule(&mut self, force_check: bool) {
         self.prev_key_count += 1;
         if (force_check || self.prev_key_count % CHECK_KEYS == 0)
--- a/components/tidb_query_common/src/storage/scanner.rs
+++ b/components/tidb_query_common/src/storage/scanner.rs
@@ -1,5 +1,6 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
+use async_trait::async_trait;
 use super::{range::*, ranges_iter::*, OwnedKvPair, Storage};
 use crate::error::StorageError;
 
@@ -24,6 +25,12 @@ pub struct RangesScanner<T> {
     working_range_begin_key: Vec<u8>,
     working_range_end_key: Vec<u8>,
 }
+
+#[async_trait]
+pub trait ReScheduleChecker {
+    async fn check_pull(&mut self);
+}
+
 
 pub struct RangesScannerOptions<T> {
     pub storage: T,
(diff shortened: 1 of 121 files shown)


breezewish (commented):
Thanks for this big change! I'm uncertain how the improvement comes from, as from my understanding previously the runner will always **try to break** every 1024 rows? Does this improvement comes from that now we may try to break every 32 rows instead?

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -23,6 +33,34 @@ pub struct RangesScanner<T> {
...
+
+impl RescheduleChecker {
+    fn new() -> Self {
+        Self {
+            prev_start: Instant::now(),
+            prev_key_count: 0,
+        }
+    }
+
+    async fn check_reschedule(&mut self, force_check: bool) {
```
breezewish, line 54: This function deserves careful compiler hints like `likely` and `inline(always)`, as it will be invoked for each row and ensuring its performance is critical.

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -23,6 +33,34 @@ pub struct RangesScanner<T> {
...
+        self.prev_key_count += 1;
+        if (force_check || self.prev_key_count % CHECK_KEYS == 0)
+            && self.prev_start.saturating_elapsed() > MAX_TIME_SLICE
+        {
+            reschedule().await;
```
breezewish, line 59: I'm not sure, but if we indeed want a reschedule check every 32 rows (considering that every 1024 rows are too rough), how about simply reducing the batch size?

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -23,6 +33,34 @@ pub struct RangesScanner<T> {
...
+            self.prev_start = Instant::now();
```
breezewish, line 60: Is this instant using Coarse timer? We may need to make sure of it, as now it is called pretty frequently, once every 32 rows.

breezewish:
> The only drawback is that big task will costs more time to finish

How bad is the performance drop for large queries?

components/tidb_query_executors/src/runner.rs
```diff
@@ -450,26 +444,13 @@ impl<SS: 'static> BatchExecutorsRunner<SS> {
...
             let (drained, record_len) = {
-                let _guard = sample.observe_cpu();
-                self.internal_handle_request(
-                    false,
-                    batch_size,
-                    &mut chunk,
-                    &mut warnings,
-                    &mut ctx,
-                )?
+                // let _guard = sample.observe_cpu();
```
kevin-xianliu, line 451: It seems observe_cpu_async will be called later, we would better to remove this line.

components/tidb_query_common/src/storage/scanner.rs
glorv: I have tested adjusting the batch-size based on the duration used by each poll, the performance is better in some case but the duration of one poll can still be very large(> 10ms) and the performance impact is still very large. I think one of the root cause is that if the iterator contains a lot of small ranges, it need to rebuild the scanner/iterator after each small range,  the is much more costly. 
BTW, in the future we may push down the reschedule to a lower level, or we may eventually let rocksdb support async io and eliminate this kind of manually reschedule, either way, this should be a pre-step to do so.

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -1,9 +1,19 @@
...
+use tikv_util::time::Instant;
+use yatp::task::future::reschedule;
+
 use super::{range::*, ranges_iter::*, OwnedKvPair, Storage};
 use crate::error::StorageError;
 
 const KEY_BUFFER_CAPACITY: usize = 64;
+/// Batch executors are run in coroutines. `MAX_TIME_SLICE` is the maximum time
+/// a coroutine can run without being yielded.
+const MAX_TIME_SLICE: Duration = Duration::from_millis(1);
```
kevin-xianliu, line 14: MAX_TIME_SLICE better to be configurable.

components/tidb_query_common/src/storage/scanner.rs
glorv: > > The only drawback is that big task will costs more time to finish
> 
> How bad is the performance drop for large queries?

I my benchmark, the P999 latency increase 3 times from 100ms to 300ms because after this change it needs about 50 times of reschedule to finish while before is less than 5 times.  But the benefit is that the throughput increased by 2 times from about 1000qps to 3000+qps.

glorv:
> > The only drawback is that big task will costs more time to finish
> 
> How bad is the performance drop for large queries?

I my benchmark, the P999 latency increase 3 times from 100ms to 300ms because after this change it needs about 50 times of reschedule to finish while before is less than 5 times.  But the benefit is that the throughput increased by 2 times from about 1000qps to 3000+qps.

components/tidb_query_executors/src/runner.rs
glorv: fixed

components/tidb_query_common/src/storage/scanner.rs
glorv: In my benchmark (in the internal idc linux environment),  one call of `now` cost 28ns while `now_coarse` is 7ns. Not sure there are big difference in other env(like arm linux?). As the precision of `_coarse` is about 1ms, it a bit too large because the time slice is also 1ms. But in my test, there is no observable performance different between this two method.

glorv:
@breezewish  PTAL again, thank you~

breezewish (approved):

components/tidb_query_common/src/storage/scanner.rs
glorv: Maybe we can change `MAX_TIME_SLICE` to be configurable in a separate pr if need. 😁

glorv:
@sticnarf PTAL, thanks~

components/tidb_query_common/src/storage/scanner.rs
```diff
@@ -58,24 +97,26 @@ impl<T: Storage> RangesScanner<T> {
...
     /// Fetches next row.
     /// Note: `update_scanned_range` can control whether update the scanned
     /// range when `is_scanned_range_aware` is true.
-    pub fn next_opt(
+    pub async fn next_opt(
         &mut self,
         update_scanned_range: bool,
     ) -> Result<Option<OwnedKvPair>, StorageError> {
         loop {
+            let mut force_check = true;
```
sticnarf, line 119: Nice finding. Didn't notice that iterator renew/re-seek is the cause of the long duration.

sticnarf (approved):
LGTM

PR #13342, opened 2022-08-25
copr: add json opaque value and implement conversion, comparison...

### LINKED ISSUE #13340: Support JSON Opaque value and related conversion
## Development Task

The TiDB has added the JSON Opaque value, so the TiKV should also support it in the cop to handle the push-down format.

YangKeao:

Issue Number: Close #13340

What's Changed:

Add a JSON Opaque value to implement the conversion from binary string.

### Check List 

Tests 

- [x] Unit test
- Integration test
- No code

```
drop table if exists test;
create table test (a JSON);
insert into test select json_objectagg('a', b'01010101');
select * from test where cast(json_extract(a, '$.a') as signed) = 0;
```

**NOTE**: it has a difference with mysql. The `select * from test where json_extract(a, '$.a') = 0;` will produce a result in this PR, as it's automatically casted to `double BINARY`. But it will give no result in MySQL. I didn't investigate this problem further. I'll create an issue after this PR is merged :cry: . (Though, MySQL source code tells me any opaque value will be casted to 0. Is there a way to find whether it does cast or not?).

### Release note 

```release-note
Fix the issue that the conversion from binary to json is not compatible with MySQL
```


YangKeao:
PTAL @xiongjiwei  @wshwsh12 @gengliqi

xiongjiwei:
@xhebox

xiongjiwei (approved):

xhebox (approved):

wshwsh12 (approved):

gengliqi (approved):

PR #13345, opened 2022-08-25
storage: implement the FlashbackToVersion txn command

JmPotato:

Issue Number: ref #13303.

What's Changed:

```commit-message
Implement the `FlashbackToVersion` txn command, which contains two phases: 
`command::FlashbackToVersionReadPhase` and `command::FlashbackToVersion`.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```


src/storage/mvcc/reader/reader.rs
```diff
@@ -166,6 +169,28 @@ impl<S: EngineSnapshot> MvccReader<S> {
...
             scan_mode,
             current_key: None,
+            hint_min_version: None,
+            fill_cache: !ctx.get_not_fill_cache(),
+            term: ctx.get_term(),
+            version: ctx.get_region_epoch().get_version(),
+        }
+    }
+
+    pub fn new_min_version_hint_with_ctx(
```
Connor1996, line 179: how about passing `hint_min_version` in `scan_writes`, you can add a parameter for `create_write_cursor`.

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -0,0 +1,114 @@
...
+    request_type!(KvFlashbackToVersion);
+    property!(readonly);
+    gen_lock!(empty);
+
+    fn write_bytes(&self) -> usize {
+        0
+    }
+}
+
+impl<S: Snapshot> ReadCommand<S> for FlashbackToVersionReadPhase {
```
Connor1996, line 43: why need read phase

JmPotato: I took the design of `ResolveLock`/`ResolveLockReadPhase` and divide this operation, which requires reading before writing, into two stages. I think the advantage of this is that it is more in line with the design model of `txn::Command`, so that we don't have to write some special logic because we can't determine the keys involved in `FlashbackToVersion` to bypass the latch lock fetching.

Connor1996 (approved):
LGTM

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn process_read(self, snapshot: S, statistics: &mut Statistics) -> Result<ProcessResult> {
+        let mut reader = MvccReader::new_with_ctx(snapshot, Some(ScanMode::Forward), &self.ctx);
+        // Scan the locks.
+        let mut key_locks = Vec::with_capacity(0);
+        let mut has_remain_locks = false;
+        if self.next_lock_key.is_some() {
+            let key_locks_result = reader.scan_locks(
+                self.next_lock_key.as_ref(),
+                self.end_key.as_ref(),
+                // To flashback `CF_LOCK`, we need to delete all locks.
```
you06, line 53: If a lock is secondary and its primary lock has been committed, and its `commit_ts < self.version`, will the atomicity be violated?

src/storage/mod.rs
```diff
@@ -4399,6 +4401,219 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_flashback_to_version() {
+        let storage = TestStorageBuilderApiV1::new(DummyLockManager)
+            .build()
+            .unwrap();
+        let writes = vec![
```
you06, line 4409: What about adding a test case for non-short value here?

src/storage/mod.rs
```diff
@@ -4399,6 +4401,219 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_flashback_to_version() {
```
you06, line 4405: Maybe there is a test about chain scheduling `FlashbackToVersionReadPhase` and `FlashbackToVersion` when the flashback keys exceed `FLASHBACK_BATCH_SIZE`.

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -0,0 +1,110 @@
...
+                |_| true,
+                FLASHBACK_BATCH_SIZE,
+            );
+            statistics.add(&reader.statistics);
+            (key_locks, has_remain_locks) = key_locks_result?;
+        }
+        // Scan the writes.
+        let mut key_writes = Vec::with_capacity(0);
+        let mut has_remain_writes = false;
```
you06, line 62: `has_remain_writes` will be false if `key_locks.len() >= FLASHBACK_BATCH_SIZE`, though there are still writes to be cleaned.

src/storage/txn/commands/flashback_to_version_read_phase.rs
JmPotato: We will make sure (in the later PR) that the `self.version` given here does not exceed the `store_min_resolved_ts` to avoid this problem.

src/storage/txn/commands/flashback_to_version_read_phase.rs
you06: `next_write_key` is set to none and the command is finished after `FlashbackToVersion`.

JmPotato: Nice catch! I will fix it and add a test case to cover.

you06 (approved):

PR #13347, opened 2022-08-25
raftstore: prevent removing voters directly in 2 replica raft group

### LINKED ISSUE #10595: Remove voter from a 2 replica raft group may lead to region unavailable

LintianShi:
Signed-off-by: LintianShi <lintian.shi@pingcap.com>

Issue Number: Close #10595

Problem Summary:

It may lead to region unavailable when removing a voter from a 2 replica raft group directly.

What's Changed:

```commit
Prevent tikv from removing voters directly in 2 replica raft group. The work on PD side is done in https://github.com/tikv/pd/pull/4564. 
```

### Related changes

### Check List 

Tests 

- Integration test

Side effects

### Release note 

```release-note
When a raft group contains less than three replicas, voters cannot be removed directly or it may lead to region unavailable. Before removing a voter, it should be demoted first.
```


tests/integrations/raftstore/test_multi.rs
```diff
@@ -644,13 +654,13 @@ fn test_remove_leader_with_uncommitted_log<T: Simulator>(cluster: &mut Cluster<T
 
 #[test]
 fn test_node_remove_leader_with_uncommitted_log() {
-    let mut cluster = new_node_cluster(0, 2);
+    let mut cluster = new_node_cluster(0, 3);
```
BusyJay, line 657: How can leader be demoted when there are 3 nodes?

components/test_pd_client/src/pd.rs
```diff
@@ -1185,7 +1188,31 @@ impl TestPdClient {
     }
 
     pub fn must_remove_peer(&self, region_id: u64, peer: metapb::Peer) {
-        self.remove_peer(region_id, peer.clone());
+        let mut region = None;
```
BusyJay, line 1191: Why not update the case instead? I'm afraid it accidentally breaks existing test cases.

components/raftstore/src/store/util.rs
```diff
@@ -934,6 +934,15 @@ pub fn check_conf_change(
             (ConfChangeType::RemoveNode, PeerRole::Voter) if kind != ConfChangeKind::Simple => {
                 return Err(box_err!("{:?}: can not remove voter directly", cp));
             }
+            (ConfChangeType::RemoveNode, PeerRole::Voter)
+                if kind == ConfChangeKind::Simple && current_voter.len() <= 2 =>
+            {
+                // Details are in Issue 10595 (https://github.com/tikv/tikv/issues/10595).
```
BusyJay, line 940: Better give a brief explanation in comment.

tests/integrations/raftstore/test_conf_change.rs
```diff
@@ -931,3 +940,49 @@ fn test_conf_change_fast() {
...
+
+    cluster.must_put(b"k1", b"v");
+
+    let resp =
+        call_conf_change(&mut cluster, r1, ConfChangeType::RemoveNode, new_peer(1, 1)).unwrap();
+    let exec_res = resp
+        .get_header()
+        .get_error()
+        .get_message()
+        .contains("can not remove voter directly from a raft group that contains less than 3 voters, please demote voter firstly.");
```
BusyJay, line 969: Too precise. It should be OK to just match "demote".

tests/integrations/raftstore/test_region_info_accessor.rs
```diff
@@ -152,7 +152,23 @@ fn test_region_info_accessor_impl(cluster: &mut Cluster<NodeCluster>, c: &Region
         &[(&b""[..], &b"k1"[..]), (b"k1", b"k4"), (b"k4", b"")],
     );
 
-    pd_client.must_remove_peer(region3.get_id(), find_peer(&region3, 1).unwrap().clone());
+    pd_client.must_joint_confchange(
+        region3.get_id(),
+        vec![(
+            ConfChangeType::AddLearnerNode,
+            new_learner_peer(
```
BusyJay, line 159: How about cache it, so you don't need to write the same code again.

BusyJay (approved):
Rest LGTM

components/raftstore/src/store/util.rs
LintianShi: A brief explanation is added.

PR #13375, opened 2022-08-30
raftstore-v2: implement local read for raftstore-v2

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

This PR involves:
- Refactor `LocalReader` of raftstore-v1 to make some code sharable with raftstore-v2
- Implement `LocalReader` of raftstore-v2 and add relevant tests
- Replace `callback` with `ReadCallback` trait

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/test_raftstore/src/server.rs
```diff
@@ -284,7 +284,7 @@ impl ServerCluster {
             }
         }
 
-        let local_reader = LocalReader::new(
+        let local_reader = LocalReaderCore::new(
```
BusyJay, line 287: Why use `Core`?

components/raftstore-v2/src/batch/store.rs
```diff
@@ -490,3 +496,20 @@ where
     };
     (StoreRouter { router, logger }, system)
 }
+
+pub trait MsgRouter: Send {
```
BusyJay, line 500: It seems only used in tests. I suggest to put it to `local.rs`.

SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -467,6 +467,12 @@ impl<EK: KvEngine, ER: RaftEngine> DerefMut for StoreRouter<EK, ER> {
     }
 }
 
+impl<EK: KvEngine, ER: RaftEngine> CasualRouter<EK> for StoreRouter<EK, ER> {
```
BusyJay, line 470: Why is this necessary? I think it's the same as send.

components/raftstore/src/store/worker/read.rs
```diff
@@ -673,10 +601,7 @@ where
         rd.filter(|r| !r.pending_remove)
     }
 
-    pub fn pre_propose_raft_command(
-        &mut self,
-        req: &RaftCmdRequest,
-    ) -> Result<Option<(D, RequestPolicy)>> {
+    pub fn request_check(&mut self, req: &RaftCmdRequest) -> Result<Option<D>> {
```
BusyJay, line 604: What does `request_check` mean?

SpadeA-Tang: Validate request.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -29,20 +30,206 @@ use raftstore::{
...
+        }
+
+        // try to renew the lease by sending read query where the reading process may
+        // renew the lease
+        let (msg, mut sub) = PeerMsg::raft_query(req.clone());
+        MsgRouter::send(&self.router, region_id, msg);
+        if let Some(query_res) = sub.result().await {
+            if query_res.read().is_some() {
+                // Query successful, so try again.
+
```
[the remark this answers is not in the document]
SpadeA-Tang: indicate this comment is for this `if branch` rather than the follower line of code.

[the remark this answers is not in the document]
BusyJay: Then it should be commented on top of `if`.

components/raftstore/src/store/worker/read.rs
BusyJay: Bad style, we use `verb + noun` instead of `noun + verb`.

SpadeA-Tang: Fine.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -145,27 +331,107 @@ where
...
+    }
+
+    fn inspect_lease(&mut self) -> LeaseState {
+        // TODO: disable localreader if we did not enable raft's check_quorum.
+        if self.delegate.leader_lease.is_some() {
+            // We skip lease check, because it is postponed until `handle_read`.
+            LeaseState::Valid
+        } else {
+            debug!(self.logger, "rejected by leader lease"; "tag" => &self.delegate.tag);
+            // TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().reject_reason.no_lease.inc());
```
5kbpers, line 401: Why remove this?

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -29,20 +30,205 @@ use raftstore::{
...
+
+                    let region = Arc::clone(&delegate.region);
+                    let snap = RegionSnapshot::from_snapshot(
+                        delegate.get_snapshot(None, &mut None),
+                        region,
+                    );
+
+                    delegate.check_stale_read_safe::<E>(read_ts)?;
+
+                    // TLS_LOCAL_READ_METRICS.with(|m|
```
5kbpers, line 150: (ditto)

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: I don't know... It's strange why I remove it.
I have added it back.

components/raftstore/src/store/worker/read.rs
```diff
@@ -538,109 +537,36 @@ impl Progress {
 /// #[RaftstoreCommon]: LocalReader is an entry point where local read requests are dipatch to the
 /// relevant regions by LocalReader so that these requests can be handled by the
 /// relevant ReadDelegate respectively.
-pub struct LocalReader<C, E, D, S>
+pub struct LocalReaderCore<E, D, S>
```
BusyJay, line 540: `E` is not needed anymore.

SpadeA-Tang: Good catch.

components/raftstore/src/store/worker/read.rs
```diff
@@ -437,10 +439,10 @@ impl ReadDelegate {
         false
     }
 
-    pub fn check_stale_read_safe<S: Snapshot>(
+    pub fn check_stale_read_safe<EK: KvEngine>(
```
BusyJay, line 442: Wny need a generic?

components/raftstore/src/store/worker/read.rs
```diff
@@ -728,16 +645,116 @@ where
...
+    fn clone(&self) -> Self {
+        LocalReaderCore {
+            store_meta: self.store_meta.clone(),
+            store_id: self.store_id.clone(),
+            delegates: LruCache::with_capacity_and_sample(0, 7),
+        }
+    }
+}
+
+pub struct LocalReader<C, E, D, S>
```
BusyJay, line 666: Now that we have two different `LocalReader`, it's meaningless to use generic `D` and `S` anymore.

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: It may be used before, but now it is not needed.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -41,7 +41,7 @@ use crate::{
 };
 
 mod lease;
-mod local;
+pub mod local;
```
BusyJay, line 44: Should not be public.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,72 @@
...
 }
+
+/// A router that routes messages to the raftstore
+pub struct ServerRaftStoreRouter<EK, ER>
+where
+    EK: KvEngine,
+    ER: RaftEngine,
+{
+    router: StoreRouter<EK, ER>,
+    local_reader: RefCell<LocalReader<EK, StoreRouter<EK, ER>>>,
```
BusyJay, line 32: Why `RefCell`?

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,72 @@
...
+            router,
+            local_reader,
+        }
+    }
+
+    pub fn router(&self) -> &StoreRouter<EK, ER> {
+        &self.router
+    }
+
+    pub fn router_mut(&mut self) -> &mut StoreRouter<EK, ER> {
```
BusyJay, line 69: Why need this?

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,72 @@
...
+        logger: Logger,
+    ) -> Self {
+        let local_reader = RefCell::new(LocalReader::new(
+            StoreMetaDelegate::new(store_meta),
+            router.clone(),
+            logger,
+        ));
+        ServerRaftStoreRouter {
+            router,
+            local_reader,
```
BusyJay, line 61: It doesn't seem to be used.

components/raftstore-v2/src/operation/mod.rs
```diff
@@ -7,4 +7,5 @@ mod ready;
 
 pub use command::{SimpleWriteDecoder, SimpleWriteEncoder};
 pub use life::DestroyProgress;
+pub use query::local::{CachedReadDelegate, LocalReader, MsgRouter, StoreMetaDelegate};
```
BusyJay, line 10: Why export `CachedReadDelegate`, `MsgRouter` and `StoreMetaDelegate`?

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: `impl DerefMut for TestRouter` needs this.

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: So get_snapshot can be called on & rather than &mut. Is this not needed?

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: It is used now.

components/raftstore-v2/src/operation/mod.rs
SpadeA-Tang: StoreMetaDelegate is needed while the others are not now.

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -2,19 +2,20 @@
...
 
-use std::borrow::Cow;
+use std::{borrow::Cow, sync::Arc};
 
 use batch_system::{BasicMailbox, Fsm};
 use crossbeam::channel::TryRecvError;
 use engine_traits::{KvEngine, RaftEngine, TabletFactory};
 use kvproto::metapb;
-use raftstore::store::{Config, Transport};
+use raftstore::store::{Config, ReadDelegate, TrackVer, Transport};
```
BusyJay, line 11: Why import these?

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -96,7 +271,7 @@ where
 }
 
 #[derive(Clone)]
-struct StoreMetaDelegate<E>
+pub struct StoreMetaDelegate<E>
```
BusyJay, line 274: It's only used in this file, doesn't need to be public.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -41,7 +41,7 @@ use crate::{
 };
 
 mod lease;
-mod local;
+pub(crate) mod local;
```
BusyJay, line 44: Should re-export internal type instead of making the whole module public.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -61,6 +64,15 @@ pub struct Storage<ER> {
     logger: Logger,
 }
 
+impl<ER: RaftEngine> Deref for Storage<ER> {
```
BusyJay, line 66: It's not implemented on purpose to avoid confusion.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,87 @@
...
+    pub fn router_mut(&mut self) -> &mut StoreRouter<EK, ER> {
+        &mut self.router
+    }
+
+    #[allow(clippy::await_holding_refcell_ref)]
+    pub fn get_snapshot(
+        &self,
+        req: RaftCmdRequest,
+    ) -> std::result::Result<RegionSnapshot<EK::Snapshot>, RaftCmdResponse> {
+        block_on({
```
BusyJay, line 80: Should be async.

components/raftstore-v2/tests/integrations/mod.rs
```diff
@@ -50,19 +52,19 @@ mod test_read;
...
     fn deref(&self) -> &Self::Target {
-        &self.0
+        self.0.router()
     }
 }
 
 impl DerefMut for TestRouter {
     fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
+        self.0.router_mut()
```
BusyJay, line 67: Should make target as `ServerRaftStoreRouter` instead of `StoreRouter`. raftstore should only export one type of router.

components/raftstore-v2/src/fsm/peer.rs
SpadeA-Tang: Residual code, I will remove them.

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: It's also used in router/imp.rs

components/raftstore-v2/src/operation/query/mod.rs
SpadeA-Tang: Godd suggestion.

components/raftstore-v2/src/raft/storage.rs
SpadeA-Tang: Fine.

components/raftstore-v2/tests/integrations/mod.rs
SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/router/imp.rs
BusyJay: What could be the problem if using `&mut`?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::fmt::{self, Debug, Formatter};
+use std::{
```
BusyJay, line 3: Useless imports.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -2,44 +2,234 @@
...
+                // It can not handle other policies.
+                Ok(_) => Ok(None),
+                Err(e) => Err(e),
+            }
+        } else {
+            Err(Error::RegionNotFound(req.get_header().get_region_id()))
+        }
+    }
+
+    fn try_get_snapshot(
```
tonyxuqqi, line 101: Do we only support snapshot here? It seems CmdType::Get is used in some test. Get is just a simple wrapper of snapshot through.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -2,44 +2,234 @@
...
+        &mut self,
+        req: RaftCmdRequest,
+    ) -> std::result::Result<Option<RegionSnapshot<E::Snapshot>>, RaftCmdResponse> {
+        match self.pre_propose_raft_command(&req) {
+            Ok(Some((mut delegate, policy))) => match policy {
+                RequestPolicy::ReadLocal => {
```
tonyxuqqi, line 107: TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().local_executed_requests.inc()); is missed?

components/raftstore/src/store/worker/read.rs
```diff
@@ -538,109 +533,31 @@ impl Progress {
...
-    D: ReadExecutor<E> + Deref<Target = ReadDelegate>,
-    S: ReadExecutorProvider<E, Executor = D>,
+    D: ReadExecutor + Deref<Target = ReadDelegate>,
+    S: ReadExecutorProvider<Executor = D>,
 {
     pub store_id: Cell<Option<u64>>,
     store_meta: S,
-    kv_engine: E,
     // region id -> ReadDelegate
     // The use of `Arc` here is a workaround, see the comment at `get_delegate`
```
tonyxuqqi, line 544: // The use of `Arc` here is a workaround, see the comment at `get_delegate`
This comment seems obsolete already. please remove it.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,81 @@
...
+            router,
+            local_reader,
+        }
+    }
+
+    pub fn send(&self, addr: u64, msg: PeerMsg) -> Result<(), TrySendError<PeerMsg>> {
+        self.router.send(addr, msg)
+    }
+
+    pub fn send_raft_message(
```
tonyxuqqi, line 66: what's the caller code of this function, same as send()

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,81 @@
...
+
+impl<EK, ER> Clone for ServerRaftStoreRouter<EK, ER>
+where
+    EK: KvEngine,
+    ER: RaftEngine,
+{
+    fn clone(&self) -> Self {
+        ServerRaftStoreRouter {
+            router: self.router.clone(),
+            local_reader: self.local_reader.clone(),
```
tonyxuqqi, line 44: if local_reader is being called with borrow_mut(), clone will panic.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,81 @@
...
+    ) -> std::result::Result<(), TrySendError<Box<RaftMessage>>> {
+        self.router.send_raft_message(msg)
+    }
+
+    #[allow(clippy::await_holding_refcell_ref)]
+    pub async fn get_snapshot(
+        &self,
+        req: RaftCmdRequest,
+    ) -> std::result::Result<RegionSnapshot<EK::Snapshot>, RaftCmdResponse> {
+        let mut local_reader = self.local_reader.borrow_mut();
```
tonyxuqqi, line 78: self.local_reader.borrow_mut() will panic if it's called concurrently.

components/raftstore-v2/src/router/imp.rs
tonyxuqqi: > So get_snapshot can be called on & rather than &mut. Is this not needed?

I think either make get_snapshot use &mut or make the underlying get_snapshot accept &. To me both way is fine, the question is which one needs less code change. Using RefCell here is quite dangerous  as it's easily mis-used and cause panic.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -2,44 +2,234 @@
...
+        }
+
+        let mut resp = RaftCmdResponse::default();
+        resp.mut_header().set_error(err);
+        Err(resp)
+    }
+
+    // If the remote lease will be expired in near future send message
+    // to `raftstore` to renew it
+    fn maybe_renew_lease_in_advance(
```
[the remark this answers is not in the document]
SpadeA-Tang: You mean even after the filter of `need_renew_lease` ?

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: Good catch.

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: It's already used in some tests.

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: So, why v1 uses `RefCell`?

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -2,44 +2,234 @@
...
+    ) {
+        if !delegate.need_renew_lease(ts) {
+            return;
+        }
+
+        let region_id = req.header.get_ref().region_id;
+        TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().renew_lease_advance.inc());
+        // Send a read query which may renew the lease
+        let (msg, sub) = PeerMsg::raft_query(req.clone());
+        let _ = MsgRouter::send(&self.router, region_id, msg);
```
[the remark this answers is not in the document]
SpadeA-Tang: Okay.

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: It may be the same issue in v1 as well?

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: Yes, we may remove CmdType::Get later in some time.

components/raftstore-v2/src/router/imp.rs
BusyJay: You may also try to get rid of `RefCell` for v1.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,78 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
+use std::{
+    cell::RefCell,
```
BusyJay, line 4: Useless import.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,78 @@
...
+use crate::{batch::StoreRouter, operation::LocalReader, StoreMeta};
 
 impl<EK: KvEngine, ER: RaftEngine> LogFetchedNotifier for StoreRouter<EK, ER> {
     fn notify(&self, region_id: u64, fetched: FetchedLogs) {
         let _ = self.force_send(region_id, PeerMsg::FetchedLogs(fetched));
     }
 }
+
+/// A router that routes messages to the raftstore
+pub struct ServerRaftStoreRouter<EK, ER>
```
BusyJay, line 27:
```suggestion
pub struct RaftRouter<EK, ER>
```

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,78 @@
...
+            local_reader: self.local_reader.clone(),
+        }
+    }
+}
+
+impl<EK: KvEngine, ER: RaftEngine> ServerRaftStoreRouter<EK, ER> {
+    pub fn new(
+        store_meta: Arc<Mutex<StoreMeta<EK>>>,
+        router: StoreRouter<EK, ER>,
+        logger: Logger,
```
BusyJay, line 53: logger should be taken from `StoreRouter` instead.

components/raftstore-v2/src/router/imp.rs
```diff
@@ -1,13 +1,78 @@
...
+        ServerRaftStoreRouter {
+            router: self.router.clone(),
+            local_reader: self.local_reader.clone(),
+        }
+    }
+}
+
+impl<EK: KvEngine, ER: RaftEngine> ServerRaftStoreRouter<EK, ER> {
+    pub fn new(
+        store_meta: Arc<Mutex<StoreMeta<EK>>>,
```
BusyJay, line 51: It's strange to pass the parameter from outside. `store_meta` is expected to be created and maintained by raftstore.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -382,4 +374,25 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
     pub fn proposals_mut(&mut self) -> &mut ProposalQueue<Vec<CmdResChannel>> {
         &mut self.proposals
     }
+
+    pub fn generate_read_delegate(&self) -> ReadDelegate {
+        let region = self.region().clone();
+        let region_id = region.get_id();
+        let peer_id = self.peer().get_id();
+        ReadDelegate {
```
BusyJay, line 382: Better create a constructor for `ReadDelegate`.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -142,39 +345,339 @@ where
     }
 }
 
+struct SnapRequestInspector<'r> {
+    delegate: &'r ReadDelegate,
+    logger: &'r Logger,
+}
+
+impl<'r> RequestInspector for SnapRequestInspector<'r> {
+    fn inspect(&mut self, req: &RaftCmdRequest) -> Result<RequestPolicy> {
```
BusyJay, line 354: What's the point of implementing `RequestInspector`?

SpadeA-Tang: Remove unnecessary checks, and assert for the snap request.

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/query/local.rs
BusyJay: What's the difference between implementing as methods instead of trait?

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: Done.

components/raftstore-v2/src/router/imp.rs
SpadeA-Tang: Done.

components/raftstore-v2/src/operation/mod.rs
```diff
@@ -8,3 +8,5 @@ mod ready;
 pub use command::{SimpleWriteDecoder, SimpleWriteEncoder};
 pub use life::DestroyProgress;
 pub use ready::AsyncWriter;
+
+pub(crate) use self::query::LocalReader;
```
5kbpers, line 12: If it could be used by other mods, using `pub` would be better.

SpadeA-Tang: It's supposed not to be used by other module.

5kbpers (approved):

components/raftstore-v2/src/operation/query/local.rs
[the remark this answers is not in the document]
tonyxuqqi: > You mean even after the filter of `need_renew_lease` ?
Yes. So There're two scenarios:
1) The region is active and 1K read requests hit read local and the lease is about expire. need_renew_lease will return true for all these requests. And they're all sent to Raftstore to renew the lease. 
2) The region is inactive and 1K read requests hit read index and they're all sent to Raftstore to renew the lease. 
The ideal result is that for 1) only one pre-renew lease is sent per region. And for 2), there's also only one renew lease needed, other requests just need to wait for its finish.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -2,44 +2,251 @@
...
+    // renew the lease
+    async fn try_to_renew_lease(
+        &self,
+        region_id: u64,
+        req: &RaftCmdRequest,
+    ) -> std::result::Result<Option<QueryResult>, RaftCmdResponse> {
+        let (msg, sub) = PeerMsg::raft_query(req.clone());
+        let mut err = errorpb::Error::default();
+        match MsgRouter::send(&self.router, region_id, msg) {
+            Ok(()) => return Ok(sub.result().await),
```
tonyxuqqi, line 205: I'm wondering if we can clone and cache sub here. 
So when sub exists,  we don't need to issue the MsgRouter::send again for later requests that are before sub.result().await return . Just wait on sub.result().await. And once await returns, clear the cache. 
This is to dedup unnecessary renew lease requests to raftstore. 
With this, it can even remove the read index amend logic in raftstore v2. Because redundant read index messages will be greatly reduced or in fact completely removed. @BusyJay What do you think?

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -2,44 +2,251 @@
...
+        ts: Timespec,
+    ) {
+        if !delegate.need_renew_lease(ts) {
+            return;
+        }
+
+        let region_id = req.header.get_ref().region_id;
+        TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().renew_lease_advance.inc());
+        // Send a read query which may renew the lease
+        let (msg, sub) = PeerMsg::raft_query(req.clone());
```
tonyxuqqi, line 239: Same as above. If we cache the sub after send successfully, it can reduce the unnecessary renew lease requests.

components/raftstore-v2/src/operation/mod.rs
tonyxuqqi: In v1's code, it's called in components/server as well.

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: If it is possible, I think we can propose another PR to implement it and see the improvements.

tonyxuqqi (approved):

BusyJay:
Why merge this? I think it depends on #13568

components/raftstore-v2/src/operation/query/local.rs
BusyJay: ReadIndex amend works for requests across threads.

SpadeA-Tang:
> Why merge this? I think it depends on #13568

The circumstance here is much easier. It only has snap oepration and does not have cache.

BusyJay:
> The circumstance here is much easier. It only has snap oepration and does not have cache.

It doesn't have cross snapshot cache, but it may has cache within the same region. And this PR also lacks the check of last_valid_ts.

SpadeA-Tang:
> 

Checking last_valid_ts means we have to acquire the system time twice? One before getting the snapshot and one after it.

SpadeA-Tang:
> Checking last_valid_ts means we have to acquire the system time twice? One before getting the snapshot and one after it.

Checking last_valid_ts means we have to acquire the system time twice? One before getting the snapshot and one after it.

SpadeA-Tang:
> > The circumstance here is much easier. It only has snap oepration and does not have cache.
> 
> It doesn't have cross snapshot cache, but it may has cache within the same region. And this PR also lacks the check of last_valid_ts.

Checking last_valid_ts means we have to acquire the system time twice? One before getting the snapshot and one after it.

BusyJay:
> Checking last_valid_ts means we have to acquire the system time twice? One before getting the snapshot and one after it.

Yes, for V1. For v2, only check if latest tablet is `None` is sufficient.

PR #13377, opened 2022-08-30
raftstore: prettify snapshot build flow

Connor1996:

Issue Number: Ref #12876 

What's Changed:

```commit-message
prettify snapshot build flow
```

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```


components/raftstore/src/store/peer_storage.rs
```diff
@@ -1003,72 +1000,57 @@ where
...
     );
 
-    let msg = kv_snap
+    let apply_state: RaftApplyState = kv_snap
         .get_msg_cf(CF_RAFT, &keys::apply_state_key(region_id))
-        .map_err(into_other::<_, raft::Error>)?;
-    let apply_state: RaftApplyState = match msg {
-        None => {
-            return Err(storage_error(format!(
+        .and_then(|res| match res {
```
BusyJay, line 1005: You mean `res.ok_or_else`?

components/raftstore/src/store/peer_storage.rs
```diff
@@ -1003,72 +1000,57 @@ where
...
+            None => Err(box_err!(
```
BusyJay, line 1006: This is too implicit. What will the exact error be returned?

Connor1996: I don't know, probably `Error::Other`. L1026 does the same way.

Connor1996: refine it

BusyJay (approved):

ethercflow (approved):
LGTM

PR #13384, opened 2022-08-31
raftstore: optimize region destroy

### LINKED ISSUE #12421: Take more than 5 hours to merge empty region after drop 6T data

[2022/04/23 23:53:15.874 +08:00] [INFO] [client.go:376] ["Cluster version information"] [type=tikv] [version=6.0.0-alpha] [git_hash=ed7e1aa945d9ada9dc3ad3bd6420a9879d49ed44]

### What did you expect?
Since there are only less than 50G data during region merge, should not cost so much time.

### What did happened?
image
image
image

tabokie:

Issue Number: Close #12421

What's Changed:

Optimizations:
- Use more accurate bounds to create iterator during `delete_all_in_range_cf_by_ingest`, testing shows it can speed up merge to 10x. (previously we use the max/min of a batch of ranges)
- Only call `self.delete_all_in_range` once when cleaning overlap ranges. (previously we delete keys in all overlap ranges, then delete keys in to-apply range)
- When inserting a pending range, only delete files of overlap pending ranges. (previously we also delete keys)

Refactors:
- Change `SnapContext` to a data-only struct `SnapGenContext`. It can be shared by cloning, but fields like `pending_delete_ranges` cannot be shared this way.

```commit-message
Optimize the performance of merging empty regions
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Optimize the performance of merging empty regions
```


BusyJay:
Any stats to show the improvement?

BusyJay:
If it needs to be picked, then release note should be filled.

components/raftstore/src/store/worker/region.rs
```diff
@@ -243,28 +248,79 @@ impl PendingDeleteRanges {
...
 
-#[derive(Clone)]
-struct SnapContext<EK, R>
+struct SnapGenContext<EK, R> {
+    engine: EK,
+    mgr: SnapManager,
+    router: R,
+}
+
+pub struct Runner<EK, R, T>
```
Connor1996, line 257: let's not reorder it

tabokie:
@BusyJay "testing shows it can speed up merge to 5x". I didn't verify other improvements though, they are not as important IMO.

components/raftstore/src/store/worker/region.rs
tabokie: There's no reorder, the old `SnapContext` is removed.

BusyJay:
Which optimization contributes the most performance?

tabokie:
I don't think there should be an "enough". The second one, though seems less impressive compared to the first one, can still reduce 50% seek overhead when applying a snapshot. The third one optimizes an edge case using the new util function used in the second optimization.

components/raftstore/src/store/worker/region.rs
```diff
@@ -806,9 +798,15 @@ where
...
 
+                let ctx = SnapGenContext {
+                    engine: self.engine.clone(),
+                    mgr: self.mgr.clone(),
+                    router: self.router.clone(),
+                };
                 self.pool.spawn(async move {
                     tikv_alloc::add_thread_memory_accessor();
-                    ctx.handle_gen(
+                    Self::handle_gen(
```
Connor1996, line 808: why not make it as the member function os `SnapGenContext`

components/raftstore/src/store/worker/region.rs
```diff
@@ -575,33 +612,43 @@ where
...
             })
             .collect();
-        if let Err(e) = self.cleanup_range(&ranges) {
+
+        self.engine
+            .delete_all_in_range(DeleteStrategy::DeleteFiles, &ranges)
+            .unwrap_or_else(|e| {
+                error!("failed to delete files in range"; "err" => %e);
+            });
+        if let Err(e) = self.delete_all_in_range(&ranges) {
```
Connor1996, line 641: It's confusing here, two `delete_all_in_range` but does different things

components/raftstore/src/store/worker/region.rs
```diff
@@ -485,80 +529,73 @@ where
...
-    fn insert_pending_delete_range(&mut self, region_id: u64, start_key: &[u8], end_key: &[u8]) {
+    /// Tries to clean up files in pending ranges overlapping with the given
+    /// bounds. These pending ranges will be removed, and the given bounds will
+    /// be updated to include them. Caller must ensure the remaining keys in
+    /// these ranges will be deleted properly.
+    fn clean_overlap_ranges_fast(&mut self, start_key: &mut Vec<u8>, end_key: &mut Vec<u8>) {
         let overlap_ranges = self
             .pending_delete_ranges
             .drain_overlap_ranges(start_key, end_key);
         if !overlap_ranges.is_empty() {
```
Connor1996, line 540: if overlap_ranges.is_empty() {
    return;
}

components/raftstore/src/store/worker/region.rs
```diff
@@ -575,33 +612,43 @@ where
...
-                  "end_key" => log_wrappers::Value::key(end));
+                    "start_key" => log_wrappers::Value::key(start),
+                    "end_key" => log_wrappers::Value::key(end));
                 Range::new(start, end)
             })
             .collect();
-        if let Err(e) = self.cleanup_range(&ranges) {
+
+        self.engine
+            .delete_all_in_range(DeleteStrategy::DeleteFiles, &ranges)
```
Connor1996, line 637: How about renaming it to `delete_ranges_cfs`

components/raftstore/src/store/worker/region.rs
tabokie: How about I do it after you all review the changes, because it will require reordering a few functions.

components/raftstore/src/store/worker/region.rs
```diff
@@ -485,80 +529,73 @@ where
...
-        self.delete_all_in_range(&ranges)
-    }
-
-    /// Inserts a new pending range, and it will be cleaned up with some delay.
-    fn insert_pending_delete_range(&mut self, region_id: u64, start_key: &[u8], end_key: &[u8]) {
+    /// Tries to clean up files in pending ranges overlapping with the given
+    /// bounds. These pending ranges will be removed, and the given bounds will
+    /// be updated to include them. Caller must ensure the remaining keys in
+    /// these ranges will be deleted properly.
+    fn clean_overlap_ranges_fast(&mut self, start_key: &mut Vec<u8>, end_key: &mut Vec<u8>) {
```
Connor1996, line 536: how about returnning them instead of inplace update

components/raftstore/src/store/worker/region.rs
Connor1996: And remove `roughly_cleanup_ranges` of engine trait BTW, it's unnecessary

components/raftstore/src/store/worker/region.rs
```diff
@@ -389,18 +444,7 @@ where
...
-            .pending_delete_ranges
-            .drain_overlap_ranges(&start_key, &end_key);
-        if !overlap_ranges.is_empty() {
-            CLEAN_COUNTER_VEC
-                .with_label_values(&["overlap-with-apply"])
-                .inc();
-            self.cleanup_overlap_regions(overlap_ranges)?;
-        }
-        self.delete_all_in_range(&[Range::new(&start_key, &end_key)])?;
+        self.clean_overlap_ranges(start_key, end_key)?;
```
Connor1996, line 447: blobs of Titan are not cleaned here

tabokie: I think it's difficult to do that. We want to avoid deleting blobs when the rocksdb key still exists. But the ranges that are safe to perform delete_files on have already been removed when we finished `self.delete_all_in_range`.

components/raftstore/src/store/worker/region.rs
Connor1996: ok

components/engine_traits/src/misc.rs
```diff
@@ -33,7 +42,7 @@ pub trait MiscExt: CfNamesExt + FlowControlFactorsExt {
 
     fn flush_cf(&self, cf: &str, wait: bool) -> Result<()>;
 
-    fn delete_all_in_range(&self, strategy: DeleteStrategy, ranges: &[Range<'_>]) -> Result<()> {
+    fn delete_ranges(&self, strategy: DeleteStrategy, ranges: &[Range<'_>]) -> Result<()> {
```
Connor1996, line 45: following the naming of `flush_cfs`, it should be `delete_ranges_cfs`

BusyJay (approved):

Issue #11276, opened 2021-11-07
Upgrade outdated dependencies


kennytm:
## Development Task

Major version changes

- [ ] async-stream 0.2 → 0.3
- [ ] crc64fast 0.1 → 1.0
- [ ] fail 0.4 → 0.5
- [ ] gag 0.1 → 1.0
- [ ] nom 5 → 7
- [ ] pnet_datalink 0.23 → 0.28
- [ ] prometheus 0.12 → 0.13
- [ ] rand 0.7 → 0.8
- [ ] semver 0.11 → 1.0
- [ ] strum 0.20 → 0.22
- [ ] sysinfo 0.16 → 0.20
- [ ] time 0.1 → 0.3
- [ ] zipf 6 → 7

Minor version changes (simply `cargo update` is sufficient)

- anyhow 1.0.26 → 1.0.45
- backtrace 0.3.61 → 0.3.63
- byteorder 1.3 → 1.4
- chrono 0.4.11 → 0.4.19
- crc32fast 1.2.0 → 1.2.1
- crossbeam 0.8.0 → 0.8.1
- futures 0.3.15 → 0.3.17
- futures-executor 0.3.15 → 0.3.17
- futures-util 0.3.15 → 0.3.17
- hex 0.4.2 → 0.4.3
- http 0.2.4 → 0.2.5
- hyper 0.14.11 → 0.14.14
- itertools 0.10.0 → 0.10.1
- libc 0.2.106 → 0.2.107
- libloading 0.7.0 → 0.7.1
- mime 0.3.14 → 0.3.16
- notify 4.0.16 → 4.0.17
- openssl 0.10.32 → 0.10.38
- parking_lot 0.11.1 → 0.11.2
- paste 1.0.4 → 1.0.6
- pin-project 1.0.1 → 1.0.8
- serde 1.0.106 → 1.0.130
- serde_derive 1.0.106 → 1.0.130
- serde_json 1.0.64 → 1.0.69
- slog 2.5 → 2.7
- thiserror 1.0.22 → 1.0.30
- tokio 1.12 → 1.13
- tokio-openssl 0.6.1 → 0.6.3
- toml 0.5.7 → 0.5.8
- walkdir 2.3.1 → 2.3.2

PR #13385, opened 2022-08-31
*: Update sysinfo dependency

dveeden:

Issue Number: ref #11276

What's Changed:

Update [sysinfo](https://crates.io/crates/sysinfo) dependency from 0.16.4 to 0.26.5

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1444,12 +1444,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "doc-comment"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


BusyJay:
I can see there are several PRs that tries to update dependencies. How about also include those changes in this PR?

https://github.com/tikv/tikv/pulls/app%2Fdependabot

dveeden:
@BusyJay I prefer to handle these in separate PRs to limit the scope and the number of changes per PR

dveeden:
PTAL @BusyJay

BusyJay (approved):

hicqu (approved):

PR #13388, opened 2022-09-01
raftstore-v2: support write proposal

BusyJay:
Issue Number: Ref #12842

Ref #12020.

What's Changed:
```commit-message
This PR implements write proposal and proposal batch.

Compared to v1, a custom codec is introduced to make batch more efficient.

Epoch checker will be added in next PR.
```

### Related changes

Base on #13334.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+            None => return,
+        };
+        let current_term = self.term();
+        let mut entry_and_proposals = vec![];
+        let queue = self.proposals_mut();
+        if !queue.is_empty() {
+            for e in committed_entries {
+                let mut proposal = queue.find_proposal(e.term, e.index, current_term);
+                if let Some(p) = &mut proposal && p.must_pass_epoch_check {
+                    // In this case the apply can be guaranteed to be successful. Invoke the
```
tonyxuqqi, line 146: why it can be guaranteed? Say what if right after notify_committed() the node is down?

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+            }
+        }
+        if let Err(e) = self.validate_command(&req, &mut ctx.raft_metrics) {
+            let resp = cmd_resp::new_error(e);
+            ch.report_error(resp);
+            return;
+        }
+        // To maintain propose order, we need to make pending proposal first.
+        self.propose_pending_command(ctx);
+        match RawWriteEncoder::new(req, (ctx.cfg.raft_entry_max_size.0 as f64 * 0.4) as usize) {
```
tonyxuqqi, line 51: I think it will be more clear to separate the follow logic that could make it return Err(). 
if !Self::allow_request(&req) {
            return Err(req);
        }
  And in fact allow_request can be Peer's private fn.

components/raftstore-v2/src/operation/command/write/raw_write.rs
```diff
@@ -0,0 +1,484 @@
...
+#[derive(Debug)]
+pub struct RawWriteEncoder {
+    header: SingularPtrField<RaftRequestHeader>,
+    buf: Vec<u8>,
+    channels: Vec<CmdResChannel>,
+    size_limit: usize,
+}
+
+impl RawWriteEncoder {
+    pub fn new(
```
tonyxuqqi, line 28: As mentioned above, it's much clear if the following block can be moved to the caller side and new() will return RawWriteEncoder object directly.  The test code then does not need to call .unwrap() each time. 

        if !Self::allow_request(&req) {
            return Err(req);
        }

components/raftstore-v2/src/operation/command/write/raw_write.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    #[inline]
+    pub fn amend(&mut self, req: RaftCmdRequest) -> Result<(), RaftCmdRequest> {
+        if Self::allow_request(&req) && req.header == self.header {
+            let last_length = self.buf.len();
+            for r in req.get_requests() {
+                encode(r, &mut self.buf);
+            }
+            // The default size limit is 8 * 0.4 = 3.2MiB.
+            if self.buf.len() < self.size_limit {
```
tonyxuqqi, line 78: why not checking the size inside the for block to avoid unnecessary encode.

components/raftstore-v2/src/operation/command/write/raw_write.rs
```diff
@@ -0,0 +1,484 @@
...
+        }
+    }
+
+    #[inline]
+    pub fn data_size(&self) -> usize {
+        self.buf.len()
+    }
+
+    #[inline]
+    pub fn encode(self) -> (Vec<u8>, Vec<CmdResChannel>) {
```
tonyxuqqi, line 95: encode() is verb and here it actually means the encoded buf and its channels. 
I would suggest to separate encode into buf() and channels() two functions for better readability.

components/raftstore-v2/src/operation/command/write/raw_write.rs
```diff
@@ -0,0 +1,484 @@
...
+
+// MAGIC number to hint raw write is used.
+// TODO: use protobuf blob request seems better.
+const MAGIC_PREFIX: u8 = 0x00;
+
+/// We usually use `RaftCmdRequest` for read write request. But the codec is
+/// not efficient enough for simple request. `RawWrite` is introduce to make
+/// codec alloc less and fast.
+#[derive(Debug)]
+pub struct RawWriteEncoder {
```
tonyxuqqi, line 20: Is there any micro-benchmark of RawWriteEncoder vs RaftCmdRequest?

components/raftstore-v2/src/operation/command/write/raw_write.rs
```diff
@@ -0,0 +1,484 @@
...
+
+const DEFAULT_CF_TAG: u8 = 0;
+const WRITE_CF_TAG: u8 = 1;
+const LOCK_CF_TAG: u8 = 2;
+const ARBITRARY_CF_TAG: u8 = 3;
+
+// Generally the length of most key is within 128. The length of value is
+// within 2GiB.
+#[inline]
+fn encode_len(len: u32, buf: &mut Vec<u8>) {
```
tonyxuqqi, line 186: Add comment to clarify the encoding is variable length integer encoding?
Or maybe can just use https://docs.rs/vu64/latest/vu64/

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: Then it will be applied after being restarted.

components/raftstore-v2/src/operation/command/write/mod.rs
BusyJay: I think they are the same. It's quite common for `new` to return a result.

components/raftstore-v2/src/operation/command/write/raw_write.rs
BusyJay: No, but you can check the benchmark I did https://github.com/tikv/tikv/issues/12020.

components/raftstore-v2/src/operation/command/write/raw_write.rs
BusyJay: In most case, encode will succeed.

components/raftstore-v2/src/operation/command/write/raw_write.rs
BusyJay: buf and channel are always paired. So when buf is taken, its channel must also be taken.

components/raftstore-v2/src/operation/command/write/raw_write.rs
BusyJay: I change the codec to sqlite4's version, which encodes small numbers more efficiently. I also add a case to test all the boundaries. It's only tens line of code, better not to introduce a new dependency.

components/raftstore-v2/src/operation/command/write/raw_write.rs
tonyxuqqi: ok, then maybe take_encode would be more accurate.

BusyJay: It's not `take_xx`. `take_xx` means take some field out of the struct. Both the struct and field are still valid. Here the struct is moved.

tonyxuqqi (approved):

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+        ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+        ch: CmdResChannel,
+    ) {
+        if !self.serving() {
+            apply::notify_req_region_removed(self.region_id(), ch);
+            return;
+        }
+        if let Some(encoder) = self.raw_write_encoder_mut() {
+            match encoder.amend(req) {
```
tabokie, line 35: This relies on a strong assumption that amended requests is always valid. It requires `validate_command` only checks header and cmd type, and the state of this fsm is unchanged.

BusyJay: After being proposed, there can be only three types of failure: epoch not matched, stale command due to leadership change, key not in range. Because headers of all batched commands are the same, there will be no extra commands failure due to first reason. For the second reason, it can happen no matter batch or not. For the third reason, it's possible an invalid command can make correct command fail. But generally, a correctly implemented client should always check epoch first and then check range, so it's impossible to trigger such error.

So the assumption here is all the clients are implemented correctly.

tabokie (approved):

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+        req: RaftCmdRequest,
+    ) -> Result<u64> {
+        let data = req.write_to_bytes().unwrap();
+        self.propose(ctx, data)
+    }
+
+    #[inline]
+    fn propose<T>(&mut self, ctx: &mut StoreContext<EK, ER, T>, data: Vec<u8>) -> Result<u64> {
+        ctx.raft_metrics.propose.normal.inc();
+        PEER_PROPOSE_LOG_SIZE_HISTOGRAM.observe(data.len() as f64);
```
Connor1996, line 97: use local metrics?

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -212,8 +212,7 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER,
                 }
                 PeerMsg::RaftCommand(cmd) => {
                     self.on_receive_command(cmd.send_time);
-                    // self.on_command(cmd.cmd.request, cmd.ch)
-                    unimplemented!()
+                    self.on_command(cmd.request, cmd.ch)
```
Connor1996, line 215: how about add duration metrics for the `on_msgs`, we are lack of finer granularity within append log duration

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -24,6 +27,8 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
     /// peer list, for example, an isolated peer may need to send/receive
     /// messages with unknown peers after recovery.
     peer_cache: Vec<metapb::Peer>,
+    raw_write_encoder: Option<RawWriteEncoder>,
```
Connor1996, line 30: please add comments for some necessary fields

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -24,6 +27,8 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
...
+    proposals: ProposalQueue<Vec<CmdResChannel>>,
     pub(crate) async_writer: AsyncWriter<EK, ER>,
```
Connor1996, line 32: how about moving public fields together

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: v1 use global metrics. Perhaps we can do that all together in other PR.

components/raftstore-v2/src/fsm/peer.rs
BusyJay: We will have a final PR to check missing metrics when everything is about finished.

components/raftstore-v2/src/raft/peer.rs
BusyJay: They are group by functionality instead of accessibility.

Connor1996: seems they aren't group now

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+                Ok(()) => {
+                    encoder.add_response_channel(ch);
+                    self.set_has_ready();
+                    return;
+                }
+                Err(r) => req = r,
+            }
+        }
+        if let Err(e) = self.validate_command(&req, &mut ctx.raft_metrics) {
```
Connor1996, line 44: admin commands need validation as well. call it `on_command`?

components/raftstore-v2/src/operation/command/write/mod.rs
Connor1996, line 51: magic number

components/raftstore-v2/src/operation/command/write/raw_write.rs
```diff
@@ -0,0 +1,497 @@
...
+            encode_bytes(delete_range_req.get_start_key(), buf);
+            encode_bytes(delete_range_req.get_end_key(), buf);
+            buf.push(delete_range_req.get_notify_only() as u8);
+        }
+        CmdType::Invalid
+        | CmdType::Get
+        | CmdType::Snap
+        | CmdType::Prewrite
+        | CmdType::IngestSst
+        | CmdType::ReadIndex => unreachable!("denifed request should be filtered already"),
```
Connor1996, line 312: denifed -> defined

components/raftstore-v2/src/operation/command/write/raw_write.rs
Connor1996: +1

components/raftstore-v2/src/operation/command/write/mod.rs
BusyJay: It will be called when admin commands are implemented. Because encoder can batch all the writes. They will share the same header and only needs to be checked once.

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+        ch: CmdResChannel,
+    ) {
+        if !self.serving() {
+            apply::notify_req_region_removed(self.region_id(), ch);
+            return;
+        }
+        if let Some(encoder) = self.raw_write_encoder_mut() {
+            match encoder.amend(req) {
+                Ok(()) => {
+                    encoder.add_response_channel(ch);
```
Connor1996, line 37: why not pass the `ch` with `req` when calling `amend`

BusyJay: To make test code easier to write.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -24,9 +27,19 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
     /// peer list, for example, an isolated peer may need to send/receive
     /// messages with unknown peers after recovery.
     peer_cache: Vec<metapb::Peer>,
+
+    /// Encoder for batching proposals and encoding them in a more efficient way
+    /// than protobuf.
+    raw_write_encoder: Option<RawWriteEncoder>,
```
Connor1996, line 33: How about `batch_proposal_encoder`? Better avoid using `raw_write` that could be regarded as something related to rawkv.

Connor1996 (commented):
rest LGTM

components/raftstore-v2/src/raft/peer.rs
BusyJay: `Raw` here means the codec is not PB, which is the significant reason for better performance. And the encoder can only merge write commands instead of arbitrary proposals. Perhaps `WriteEncoder` can ease confusions.

Connor1996 (approved):
LGTM

PR #13398, opened 2022-09-05
*: introduce the non-retryable error FlashbackInProgress

### LINKED ISSUE #13397: Introduce the non-retryable error FlashbackInProgress

For now, if a read/write request is sent to a region in the flashback progress, it will reuse the `RecoveryInProgress` error to indicate that this region is temporarily inaccessible. However, this error is [set to retryable](https://github.com/tikv/client-go/pull/485) in the client-go, which is somewhat incompatible with the flashback's feature. Since we expect that the workload should be stopped during the flashback progress, if a request meets the `FlashbackInProgress` error, it should stop retrying immediately to avoid unnecessary backoff and potential unexpected data status to the user.

Return the non-retryable error `FlashbackInProgress` rather than `RecoveryInProgress` during the flashback progress.

JmPotato:

Issue Number: Close #13397. Should merge https://github.com/pingcap/kvproto/pull/982 before this PR.

What's Changed:

```commit-message
Introduce the non-retryable error `FlashbackInProgress` for the region in the flashback progress to reject any read or write.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/etc/error_code.toml
+++ b/etc/error_code.toml
@@ -1,11 +1,11 @@
-["KV:Cloud:IO"]
+["KV:Cloud:Io"]
 error = '''
-KV:Cloud:IO
+KV:Cloud:Io
 '''
 
-["KV:Cloud:SSL"]
+["KV:Cloud:Ssl"]
 error = '''
-KV:Cloud:SSL
+KV:Cloud:Ssl
 '''
 
 ["KV:Cloud:Proto"]
@@ -143,9 +143,9 @@ error = '''
 KV:Encryption:Rocks
 '''
 
-["KV:Encryption:IO"]
+["KV:Encryption:Io"]
 error = '''
-KV:Encryption:IO
+KV:Encryption:Io
 '''
 
 ["KV:Encryption:Crypter"]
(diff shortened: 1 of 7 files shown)


HuSharp (approved):
LGTM

etc/error_code.toml
```diff
@@ -1,11 +1,11 @@
-["KV:Cloud:IO"]
+["KV:Cloud:Io"]
```
Connor1996, line 1: why change

Connor1996 (commented):
should update the client side first

etc/error_code.toml
JmPotato: https://github.com/tikv/tikv/pull/13215 changed all namings to use the camel case but without calling `make error-code`, which I did.

Connor1996: That's intentional. Not sure if the error code is of case sensitivity. /cc @5kbpers

5kbpers: It's OK because error codes are used for printing only currently.

5kbpers (approved):

Connor1996 (approved):
LGTM

PR #13407, opened 2022-09-06
tikv_util: introduce future channel

### LINKED ISSUE #12842: Support physical isolation between regions

BusyJay:

Issue Number: Ref #12842 

What's Changed:

```commit-message
It's a super set of batch channel. It can be used as a batch channel
or just a future channel.

This is the first PR to use unified thread pool in apply system for v2.
```

I bench it with the old implementation, the result is highly unstable as if the receiver goes into sleep state, it will take more time to be woken up. Anyway, I post the result as below. There is no big difference between implementations.

```
running 7 tests
test bench_channel::bench_receiver_stream                                       ... bench:     820,119 ns/iter (+/- 2,278,412)
test bench_channel::bench_receiver_stream_batch                                 ... bench:     770,672 ns/iter (+/- 509,937)
test bench_channel::bench_receiver_stream_unbounded_batch_future                ... bench:     845,217 ns/iter (+/- 608,816)
test bench_channel::bench_receiver_stream_unbounded_nobatch_future              ... bench:     968,996 ns/iter (+/- 1,435,816)
test bench_channel::bench_receiver_stream_unbounded_batch_old                   ... bench:     774,375 ns/iter (+/- 1,822,096)
test bench_channel::bench_receiver_stream_unbounded_nobatch_old                 ... bench:     761,047 ns/iter (+/- 3,013,100)
test bench_channel::bench_receiver_stream_unbounded_nobatch_async_std_channel ... bench:   2,253,294 ns/iter (+/- 2,903,571)
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/tikv_util/src/mpsc/future.rs
```diff
@@ -0,0 +1,470 @@
...
+    type Item = T;
+
+    #[inline]
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let queue = unsafe { &*self.queue };
+        if let Some(t) = queue.queue.pop() {
+            return Poll::Ready(Some(t));
+        }
+        queue.waker.register(cx.waker());
+        if let Some(t) = queue.queue.pop() {
```
Connor1996, line 116: add a comment for why try pop twice

components/tikv_util/src/mpsc/future.rs
```diff
@@ -0,0 +1,470 @@
...
+impl<T: Send> Stream for Receiver<T> {
+    type Item = T;
+
+    #[inline]
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let queue = unsafe { &*self.queue };
+        if let Some(t) = queue.queue.pop() {
+            return Poll::Ready(Some(t));
+        }
+        queue.waker.register(cx.waker());
```
Connor1996, line 115: What if the below pop consume the element, then the next wake would be necessary.

Connor1996: I mean, should it be 
```
        if let Some(t) = queue.queue.pop() {
            return Poll::Ready(Some(t));
        }
        if let Some(t) = queue.queue.pop() {
            return Poll::Ready(Some(t));
        }
        queue.waker.register(cx.waker());
``

Comment (BusyJay) [reply components/tikv_util/src/mpsc/future.rs:115]: It's for edge cases only, which should hardly happen. If that happens, clear the waker may cause unnecessary contention.

**Comment by BusyJay** [comment]
Benched with sysbench read_write, insert, update index, point select, the differences in througput and latency are within 3%.

PTAL

**Review by Connor1996 — COMMENTED** [review]
rest LGTM

File: components/tikv_util/src/mpsc/future.rs [hunk 3]
```diff
@@ -0,0 +1,471 @@
...
+            // The last sender is dropped, we need to wake up the receiver.
+            queue.waker.wake();
+        } else if previous == SENDER_COUNT_BASE {
+            atomic::fence(Ordering::Acquire);
+            drop(unsafe { Box::from_raw(self.queue) });
+        }
+    }
+}
+
+unsafe impl<T: Send> Send for Sender<T> {}
```
sticnarf, line 100: What about also `Sync`? Sometimes non-`Sync` types may cause difficulties when `Sync` bounds are required.

components/tikv_util/src/mpsc/future.rs
```diff
@@ -0,0 +1,471 @@
...
+            // Our implementation is safe to poll twice.
+            match ctx.rx.poll_next_unpin(cx) {
+                Poll::Ready(Some(m)) => {
+                    let collection = ctx.elem.get_or_insert_with(&ctx.initializer);
+                    if let Some(m) = ctx.collector.collect(collection, m) {
+                        received = Some(m);
+                        break false;
+                    }
+                    count += 1;
+                    if count >= ctx.max_batch_size {
```
sticnarf, line 238: If `max_batch_size` belongs to `BatchReceiver`, in which case we expect `BatchCollector::collect` returns `Some` and split?

sticnarf, line 238: If `max_batch_size` belongs to `BatchReceiver`, in which case do we expect `BatchCollector::collect` returns `Some` and split?

components/tikv_util/src/mpsc/future.rs
```diff
@@ -0,0 +1,471 @@
...
+
+        if ctx.elem.is_none() && finished {
+            return Poll::Ready(None);
+        } else if ctx.elem.is_none() {
+            return Poll::Pending;
+        }
+        let elem = ctx.elem.take();
+        if let Some(m) = received {
+            let collection = ctx.elem.get_or_insert_with(&ctx.initializer);
+            let _received = ctx.collector.collect(collection, m);
```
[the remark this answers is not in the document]
BusyJay: Otherwise `received` will be dropped and message is lost.

components/tikv_util/src/mpsc/future.rs
BusyJay: `max_batch_size` is for the number of message, `BatchCollector` can add other limits like data size.

BusyJay: The limit is for raft client in the past. Now raft client uses its own batch algorithm, so the API is unnecessary. I remove it for simplicity.

components/tikv_util/src/mpsc/future.rs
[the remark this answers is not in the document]
BusyJay: The logic is simplified, now it won't collect. again.

sticnarf (approved):
LGTM

components/tikv_util/src/mpsc/future.rs
```diff
@@ -0,0 +1,431 @@
...
+    initializer: I,
+    collector: C,
+}
+
+impl<T, I, C> BatchReceiver<T, I, C> {
+    /// Creates a new `BatchReceiver` with given `initializer` and `collector`.
+    /// `initializer` is used to generate a initial value, and `collector`
+    /// will collect every (at most `max_batch_size`) raw items into the
+    /// batched value.
+    pub fn new(rx: Receiver<T>, max_batch_size: usize, initializer: I, collector: C) -> Self {
```
Connor1996, line 182: What's the use case for the collector? Isn't `vec` adequate?

Connor1996 (approved):
LGTM anyway

components/tikv_util/src/mpsc/future.rs
BusyJay: Check

https://github.com/tikv/tikv/blob/8e311cda5369e2e5231ac40e4a5feb931f21d674/src/server/service/kv.rs#L2357-L2369

PR #13410, opened 2022-09-06
server: Record the duration of executing the snapshot and feedback to PD Server

### LINKED ISSUE #13409: Record the duration of executing the snapshot and feedback to PD Server
## Development Task
ref issue: https://github.com/tikv/pd/issues/5467
In this issue, PD will use the snapshot duration to control the current size of the `move-peer`

bufferflies:

Issue Number: Close #13409 

What's Changed:

```commit-message
1. collect the duration of generating and sending snapshot 
2. records the the total duration between receiving the snapshot task and finish to sending all snapshot .
3. report the metrics to the pd server. 
```

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/store.rs
```diff
@@ -2367,6 +2367,8 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
         let snap_stats = self.ctx.snap_mgr.stats();
         stats.set_sending_snap_count(snap_stats.sending_count as u32);
         stats.set_receiving_snap_count(snap_stats.receiving_count as u32);
+        stats.set_snapshot_stats(::protobuf::RepeatedField::from_vec(snap_stats.stats));
```
lhy1024, line 2370: Will this pr make the msg of the heartbeat very large?

bufferflies: yes, I will ignore the record that  duration of snapshot is less than 1s, so the total should be less than store heartbeat interval (default 10).

lhy1024 (approved):

SpadeA-Tang (approved):

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2464,6 +2464,8 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
...
```
pingyu, line 2467: Nit:
```suggestion
        stats.set_snapshot_stats(snap_stats.stats.into());
```

components/raftstore/src/store/worker/region.rs
```diff
@@ -258,6 +258,7 @@ where
         notifier: SyncSender<RaftSnapshot>,
         for_balance: bool,
         allow_multi_files_snapshot: bool,
+        start: UnixSecs,
```
pingyu, line 261: How about put `start` as member of `SnapGenContext` ? Seems it's reasonable as a kind of "context" of snapshot generation. And then we don't need to pass the variable all the way along.

Besides, why not use `Instant` type ? `Instant` is monotonic, and we use it in most conditions in TiKV.

components/raftstore/src/store/snap.rs
```diff
@@ -123,6 +124,8 @@ pub struct SnapKey {
     pub region_id: u64,
     pub term: u64,
     pub idx: u64,
+    pub start: u64,
```
pingyu, line 127: It seems that we should not add these two member here. Then we can not get the expected snapshot by `(region_id, term, idx)` from registry.

components/raftstore/src/store/snap.rs
```diff
@@ -1363,6 +1371,7 @@ pub enum SnapEntry {
 pub struct SnapStats {
     pub sending_count: usize,
     pub receiving_count: usize,
+    pub stats: Vec<SnapshotStat>,
```
pingyu, line 1374: Nit:
```suggestion
    pub snap_stats: Vec<SnapshotStat>,
```

components/raftstore/src/store/snap.rs
```diff
@@ -1727,9 +1749,11 @@ impl SnapManager {
             }
         }
 
+        let stats = self.core.stats.wl().drain(..).collect();
```
pingyu, line 1752: Try `let stats = mem::take(&mut self.core.stats.wl())`.

components/raftstore/src/store/snap.rs
```diff
@@ -1376,6 +1385,7 @@ struct SnapManagerCore {
     encryption_key_manager: Option<Arc<DataKeyManager>>,
     max_per_file_size: Arc<AtomicU64>,
     enable_multi_snapshot_files: Arc<AtomicBool>,
+    stats: Arc<RwLock<Vec<SnapshotStat>>>,
```
pingyu, line 1388: How about utilize histogram to store the distribution of durations ? If it's precise enough. Then we can reduce the usage of memory and network.

bufferflies: The stats will report pd, pd will depence on this to adjust store limit.

components/raftstore/src/store/snap.rs
bufferflies: fixed

components/raftstore/src/store/fsm/store.rs
bufferflies: fixed

components/raftstore/src/store/worker/region.rs
bufferflies: > How about put start as member of `SnapGenContext` ?
yes，It is better to put `start` into the context .
 >why not use Instant type?
 The start only record the snapshot's beginning timestamp, and it will be sent into another thread to record the total duration of snapshot.

src/server/snap.rs
```diff
@@ -185,6 +188,20 @@ pub fn send_snap(
             Ok(_) => {
                 fail_point!("snapshot_delete_after_send");
                 mgr.delete_snapshot(&key, &chunks.snap, true);
+                let cost = UnixSecs::now().into_inner() - snapshot_tracker.start;
+                // it should ignore if the duration of snapshot is less than 1s to decrease the
```
nolouch, line 192: why not do some aggregation in tikv side?

nolouch (commented):
rest lgtm

components/raftstore/src/store/snap.rs
```diff
@@ -1657,6 +1695,18 @@ impl SnapManager {
         self.core.limiter.speed_limit()
     }
 
+    pub fn collect_stat(&self, snap: SnapshotStat) {
+        debug!(
+            "collect snapshot stat";
+            "region_id" => snap.region_id,
+            "total_size" => snap.get_transport_size(),
+            "total_duration_sec" => snap.get_total_duration_sec(),
```
nolouch, line 1703: Does the total duration only contain the `generate +send`?

src/server/snap.rs
```diff
@@ -185,6 +188,20 @@ pub fn send_snap(
...
+                // grpc data size.
+                if cost >= 1 {
+                    let mut stat = SnapshotStat::default();
+                    stat.set_region_id(snapshot_tracker.region_id);
```
nolouch, line 196: it's better to unify use `snapshot_tracker` to trace the all stage, and then `mgr.collect_stat(snapshot_tracker)`

components/raftstore/src/store/snap.rs
```diff
@@ -1051,7 +1084,10 @@ impl Snapshot {
         // set snapshot meta data
         snap_data.set_file_size(total_size);
         snap_data.set_version(SNAPSHOT_VERSION);
-        snap_data.set_meta(self.meta_file.meta.as_ref().unwrap().clone());
+        let meta = self.meta_file.meta.as_mut().unwrap();
+        meta.set_start(start.into_inner());
+        meta.set_generate_duration_sec(t.saturating_elapsed().as_secs());
```
nolouch, line 1089: can we not use `meta` to pass the context?

src/server/snap.rs
bufferflies: PD will decide the count of operator depences on these duration, it can adjust by some configuration.

components/raftstore/src/store/snap.rs
bufferflies: I think it's too waster and complex  become more if we pass this stats by snapshot manager.

src/server/snap.rs
bufferflies: yes，I will remove `snapshot_tracker` and only use `SnapshotStats`

components/raftstore/src/store/worker/region.rs
```diff
@@ -309,7 +311,7 @@ where
             return;
         }
 
-        let start = Instant::now();
+        let t = Instant::now();
```
nolouch, line 314: why change this?

nolouch (approved):
lgtm

src/server/snap.rs
```diff
@@ -193,6 +201,22 @@ pub fn send_snap(
             Ok(_) => {
                 fail_point!("snapshot_delete_after_send");
                 mgr.delete_snapshot(&key, &chunks.snap, true);
+                let now = UnixSecs::now().into_inner();
+                let mut cost = 0;
```
pingyu, line 205: Nit:

```suggestion
                let cost = now.saturating_sub(snap_start);
```

pingyu (approved):
LGTM~

src/server/snap.rs
bufferflies: fixed

PR #13411, opened 2022-09-06
components: introduce tirocks module

BusyJay:
Issue Number: Ref #13058 

What's Changed:
```commit-message
Only make it in the codebase, will not compile it. It will replace
engine_rocks once it's finished.
```

### Check List 

Tests 

- Manual test
   Uncomment the member list and `make dev`.

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_tirocks/src/lib.rs
+++ b/components/engine_tirocks/src/lib.rs
@@ -1,0 +1,12 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+//! A new implementation of engine_traits using tirocks.
+//!
+//! When all features of engine_rocks are implemented in this module,
+//! engine_rocks will be removed and TiKV will switch to tirocks.
+
+extern crate tikv_alloc as _;
+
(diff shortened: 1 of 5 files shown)


components/engine_tirocks/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+//! A new implementation of engine_traits using tirocks.
+//!
+//! When all features of engine_rocks are implemented in this module,
+//! engine_rocks will be removed and TiKV will switch to tirocks.
+
+extern crate tikv_alloc as _;
```
tabokie, line 8: Why is the `as _`?

tabokie (approved):

components/engine_tirocks/src/lib.rs
BusyJay: It's only used for link purpose.

Connor1996 (approved):

PR #13426, opened 2022-09-07
storage: skip Rollback when checking newer version for non-pessimisitc keys

### LINKED ISSUE #13425: False positive PessimisticLockNotFound when a non-pessimistic key has newer rollbacks
A concrete case is from https://github.com/pingcap/tidb/issues/35525#issuecomment-1171846720

> It's caused by a rollback record on non-unique index keys whose ts is greater than the for_update_ts of the pessimistic transaction throwing the error.
> 
> The sequence of events can be like:
> 
> Pessimistic txn-A locks row key using for_update_ts = 1.
> Optimistic txn-B with start_ts = 2 prewrites the index key.
> txn-A finds the locks on index keys and rolls back txn-B, leaving rollback records of ts=2 on index keys.
> txn-A prewrite index keys, and find a write conflict.

Because a Rollback record can be written at any timestamp. It may cause pessimistic transactions to throw `PessimisticLockNotFound` on non-pessimistic keys even if the corresponding pessimistic lock still exists.

sticnarf:

Issue Number: Close #13425

Ref: https://github.com/pingcap/tidb/issues/35525

What's Changed:

```commit-message
Don't treat newer Rollback records as write conflicts for non-pessimistic
keys in pessimistic transactions. They can cause false positive errors
because they can be written even if the pessimistic lock of the
corresponding row key exists.

Rollback records are only used to prevent retried prewrite from
succeeding. Even if the Rollback record of the current transaction is
collapsed by a newer record, it is safe to prewrite this non-pessimistic
key because either the primary key is rolled back or it's protected
because it's written by CheckSecondaryLocks.
```

#### Safety

The only risk of newer rollback records is that we may succeed in prewriting even if the lock was rolled back and collapsed.

- If it's not an async-commit transaction, rollback records are written in two cases:
  - [x] The transaction aborts itself during 2PC. The primary lock must not be committed. In this case, the wrongly prewritten lock will be finally rolled back by others.
  - [x] A lock is rolled back by ResolveLock. Before that, the primary lock must have been rolled back. So, the new lock will also be rolled back by others.
- If it's an async-commit transaction:
  - [x]  The transaction aborts itself during 2PC. This means some other key must fail to prewrite. In this case, the wrongly prewritten lock will be finally rolled back by others.
  - [x] A lock is rolled back by ResolveLock. Before that, some other key must have been rolled back. So, the new lock will also be rolled back by other readers.
  - [x] The rollback is written by CheckSecondaryLocks. But CheckSecondaryLocks always writes a protected rollback. The rollback must be detected by our retried prewrite.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Reduce false-positive PessimisticLockNotFound errors in conflicting auto-commit workloads.
```


cfzjywxk:
Is there internal regression sysbench test cases results to verify the false-positive fix?

ekexium:
> Is there internal regression sysbench test case result to verify the false-positive fix?

https://github.com/pingcap/tidb/issues/35525. I remember that it's easy to reproduce the problem.

sticnarf:
> > Is there internal regression sysbench test case result to verify the false-positive fix?
> 
> [pingcap/tidb#35525](https://github.com/pingcap/tidb/issues/35525). I remember that it's easy to reproduce the problem. The key is small data (=> conflict) and region error.

Yes. It's easy to reproduce the original issue when running sysbench with zipfian distribution. With this PR, the PessimisticLockNotFound error no longer exists.

ekexium:
> A lock is rolled back by ResolveLock. Before that, some other key must have been rolled back. So, the new lock will also be rolled back by other readers.

Is it possible that the rollback record of _some other key_ is collapsed?

sticnarf:
> > A lock is rolled back by ResolveLock. Before that, some other key must have been rolled back. So, the new lock will also be rolled back by other readers.
> 
> Is it possible that the rollback record of _some other key_ is collapsed?

In this case, the _some other key_ should be rolled back by either `CheckTxnStatus` or `CheckSecondaryLocks`.

If it is the primary key, the rollback of the primary key of a pessimistic transaction must be protected.

If it is a secondary key, `CheckSecondaryLocks` also writes protected rollbacks.

ekexium (approved):

MyonKeminta:
Should we consider the possibility that the current transaction ignores the rollback, but committed at a calculated commit ts that's the same as the rollback's start_ts, so that the rollback is overwritten?

MyonKeminta (approved):

PR #13434, opened 2022-09-08
*: move the codec function together

rleungx:

Issue Number: Close #xxx

What's Changed:

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```


breezewish:
v1 and v2 should produce same results. Do you have further interest to reduce the usage of v1? :)

components/codec/src/stream_event.rs
```diff
@@ -71,6 +71,28 @@ impl Iterator for EventIterator {
     }
 }
 
+// TODO: It's added to solve the dependency problem. Remove it in the future
+// since there is a same implementation in tikv_util.
+#[derive(Debug, Clone, Copy)]
+pub enum Either<L, R> {
```
BusyJay, line 77: I don't see the point of using `Either` here. The returned type of `encode_event` is `[impl AsRef<[u8]>; 4]`, which hide the concrete type. So technically, it's the same as just using `[[u8; 4]; 4]`.

rleungx: Removed, PTAL

BusyJay:
> v1 and v2 should produce same results. Do you have further interest to reduce the usage of v1? :)

I have the same confusion. For example, the only difference of number_v1::decode_u64 and number_v2::decode_u64 is whether to panic or return result when length is too small. We can unify them instead of defining two versions.

rleungx:
> > v1 and v2 should produce same results. Do you have further interest to reduce the usage of v1? :)
> 
> I have the same confusion. For example, the only difference of number_v1::decode_u64 and number_v2::decode_u64 is whether to panic or return result when length is too small. We can unify them instead of defining two versions.

I agree, but maybe we can do it later? Right now, it's not clear to us which codec function should be used.

PR #13441, opened 2022-09-09
raftstore, resolved_ts: clarify the use of term safe_ts and resolved_ts

### LINKED ISSUE #13437: Clarify the use of term `safe_ts` and `resolved_ts`
Internally, `safe_ts` is calculated from the `resolved_ts` and is used to indicate the upper safe read bound for a stale read request, so they are basically the same. However, [the mix of the two within TKV](https://github.com/tikv/tikv/blob/76f4a4e7ca6b15a1cc8e65c54049b816c2a1c45d/components/raftstore/src/store/worker/pd.rs#L678) can confuse the developers who do not know the intrinsic connection. So I propose the following changes to clarify the use of terms `safe_ts` and `resolved_ts`:

- Only use `safe_ts` in the stale read path to keep in line with the "safe to read with the ts" meaning.
- Use `resolved_ts` in other places to indicate the original meaning of "the biggest ts upper bound that there is no any potential uncommitted transaction before it".
- Use a different function name or comment to distinguish between the above two uses.

JmPotato:

Issue Number: Close #13437.

What's Changed:

```commit-message
Clarify the use of term `safe_ts` and `resolved_ts`.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/raftstore/src/store/util.rs
+++ b/components/raftstore/src/store/util.rs
@@ -939,6 +939,29 @@ impl RegionReadProgressRegistry {
             .map(|rp| rp.core.lock().unwrap().applied_index)
     }
 
+    // NOTICE: this function is an alias of `get_safe_ts` to distinguish the
+    // semantics.
+    pub fn get_resolved_ts(&self, region_id: &u64) -> Option<u64> {
+        self.registry
+            .lock()
+            .unwrap()
+            .get(region_id)
+            .map(|rp| rp.resolved_ts())
+    }
+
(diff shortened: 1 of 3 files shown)


JmPotato:
/cc @5kbpers @lhy1024 

PTAL, thx!

lhy1024 (approved):

5kbpers (approved):

PR #13444, opened 2022-09-09
storage: skip Rollback when checking newer version for non-pessimisitc keys (#13426)

### LINKED ISSUE #13425: False positive PessimisticLockNotFound when a non-pessimistic key has newer rollbacks

ti-srebot:
cherry-pick #13426 to release-6.1
```bash
# In tikv repo:
git pr https://github.com/tikv/tikv/pull/13444
```

```bash
git push git@github.com:ti-srebot/tikv.git pr/13444:release-6.1-3f5acade42d6
```

---

Issue Number: Close #13425

Ref: https://github.com/pingcap/tidb/issues/35525

What's Changed:

```commit-message
Don't treat newer Rollback records as write conflicts for non-pessimistic
keys in pessimistic transactions. They can cause false positive errors
because they can be written even if the pessimistic lock of the
corresponding row key exists.

Rollback records are only used to prevent retried prewrite from
succeeding. Even if the Rollback record of the current transaction is
collapsed by a newer record, it is safe to prewrite this non-pessimistic
key because either the primary key is rolled back or it's protected
because it's written by CheckSecondaryLocks.
```

#### Safety

The only risk of newer rollback records is that we may succeed in prewriting even if the lock was rolled back and collapsed.

- If it's not an async-commit transaction, rollback records are written in two cases:
  - [x] The transaction aborts itself during 2PC. The primary lock must not be committed. In this case, the wrongly prewritten lock will be finally rolled back by others.
  - [x] A lock is rolled back by ResolveLock. Before that, the primary lock must have been rolled back. So, the new lock will also be rolled back by others.
- If it's an async-commit transaction:
  - [x]  The transaction aborts itself during 2PC. This means some other key must fail to prewrite. In this case, the wrongly prewritten lock will be finally rolled back by others.
  - [x] A lock is rolled back by ResolveLock. Before that, some other key must have been rolled back. So, the new lock will also be rolled back by other readers.
  - [x] The rollback is written by CheckSecondaryLocks. But CheckSecondaryLocks always writes a protected rollback. The rollback must be detected by our retried prewrite.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Reduce false-positive PessimisticLockNotFound errors in conflicting auto-commit workloads.
```


ekexium (approved):

MyonKeminta (approved):

MyonKeminta:
Ah, build failed.

MyonKeminta (approved):

Issue #13298, opened 2022-08-17
Enhanced queueing for conflicting pessimistic transactions


MyonKeminta:
## Background

Currently, we are adopting a simple but rough implementation for pessimistic lock waiting and waking up. It works just fine in most normal cases, however, there will be problems under workloads that has frequent lock conflicts. When there are many requests trying to acquire the lock on a single key in TiKV:

* The order that the transactions successfully get the lock has too much randomness
* When a pessimistic lock request is woken up after waiting a while in TiKV, it always returns `WriteConflict` to TiDB so that TiDB will retry the statement.
* When many pessimistic lock requests are waiting on the same key in TiKV and the current lock is released, one of the requests will be woken up, and the others will be woken up too after delaying for `wake-up-delay-duration` (which is a configurable value that defaults to 10ms). If the first awakened transaction's retrying request arrives late, the key might already have been locked by another transaction.

This may lead to some problems:

* The latency to acquire the lock might be very unstable, causing high tail latency.
* There might be too many statement-retries that are useless but wastes resources.
* When the problem is severe, it's also possible to cause errors like lock wait timeout and statement retry limit exceeded.

By carefully adjusting `wake-up-delay-duration`, we are actually able to try to make it work well in most scenarios. However, it's very hard to adjust it since it's difficult to predict the effect of changing it. It's even harder if a user who don't understand the internal implementation of TiKV want to tune it. Besides, its a store level config, which means, if a user have more than one workloads that has different characteristics in a single cluster, it's impossible to use different `wake-up-delay-duration` for different workloads.

We expects it would be better to make transactions in such kind of scenarios to execute sequentially, avoiding too much useless statement retries and high tail latency. However, this will be very complicated to implement in our current architecture, especially considering that we needs to guarantee the backward compatibility. We have a complicated design for this problem and it took us much efforts to trying to demonstrate it. We've confirmed that the complicated design is helpful for solving the problem stated above in those kinds of scenarios. However, the stronger queueing behavior in lock waiting introduces higher average latency, or to say, lower QPS. We do have ideas to optimize it further, however, it would make the design much more complicated.

Now we are going to productize the idea. As the first step, we will first implement a subset of the original design: we will enhance the queueing behavior for pessimistic lock requests that involves only one key. This will be helpful for many cases that has frequent lock conflicts on some single rows. And since there's possibly QPS regression, the new behavior will be optional and disabled by default. A user will need to choose to get lower tail latency at the expense of higher average latency if the user needs.

[...]

PR #13447, opened 2022-09-09
storage, lock_manager: Use the new lock waiting queue instead of WaiterManager to handle pessimistic lock waking up

MyonKeminta:

Issue Number: ref #13298

What's Changed:

This PR refactors the implementation of lock waiting and waking up by introducing a new lock waiting queue, **without changing the current lock-waiting behavior**. This will be part of the work of introducing the new lock waiting model (#13298)

Note that this PR doesn't introduce any optimization to the lock-waiting model. It's a refectory which is the basis of the optimization.

Requires:

* https://github.com/tikv/tikv/pull/13486 : Introduces the new lock waiting facilities, and this PR updates the write path of acquiring lock and releasing lock to make use of them. Some important points are:
* https://github.com/tikv/tikv/pull/13560 : Adds metrics to the new lock waiting queue.
* https://github.com/tikv/tikv/pull/13584 : Avoids the stale entry issue by using another implementation of priority queue that supports randomly removing instead of the std BinaryHeap.

1. `WriteResultLockInfo` (returned by `AcquirePessimisticLock::process_write`) carries parameters, which can be used for resuming the request in the future.
2. `WriteResultLockInfo` will be converted into `LockWaitContext` and `LockWaitEntry`, and then send to both `LockManager` and the new `LockWaitQueues`.
3. When a storage command releases some locks, will return the released locks to `Scheduler::process_write`, which will then call `on_release_locks` to pop lock waiting entries from the queues and wake up them asynchronously (to avoid increasing too much latency of the current command).
4. The `LockManager` (and its inner module `WaiterManager`) no longer has the responsibility for waking up waiters, but keeps its functionality of handling timeout and performing deadlock detection. Instead, it has a new `remove_lock_wait` method to remove a waiter from it.
5. Waiters in `WaiterManager` can now be uniquely identified by a `LockWaitToken`, and the data structure in `WaiterManager` is therefore changed. Accessing by lock hash and transaction ts is still necessary to handle the result of deadlock detection.

```commit-message
Updates the write path of acquiring lock and releasing lock to make use of the new `LockWaitQueue`. Some important points are:

1. `WriteResultLockInfo` (returned by `AcquirePessimisticLock::process_write`) carries parameters, which can be used for resuming the request in the future.
2. `WriteResultLockInfo` will be converted into `LockWaitContext` and `LockWaitEntry`, and then send to both `LockManager` and the new `LockWaitQueues`.
3. When a storage command releases some locks, will return the released locks to `Scheduler::process_write`, which will then call `on_release_locks` to pop lock waiting entries from the queues and wake up them asynchronously (to avoid increasing too much latency of the current command).
4. The `LockManager` (and its inner module `WaiterManager`) no longer has the responsibility for waking up waiters, but keeps its functionality of handling timeout and performing deadlock detection. Instead, it has a new `remove_lock_wait` method to remove a waiter from it.
5. Waiters in `WaiterManager` can now be uniquely identified by a `LockWaitToken`, and the data structure in `WaiterManager` is therefore changed. Accessing by lock hash and transaction ts is still necessary to handle the result of deadlock detection.
```

### Related changes

### Check List 

Tests (WIP) 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


MyonKeminta:
Some test results:

## Hot update

branch | threads | qps | min | avg | 95 | max | -
-|-|-|-|-|-|-|-
this|1|410.96|5.00|7.30|8.58|213.08
this|16|427.75|4.96|112.19|484.44|1258.74
this|64|424.94|5.22|451.48|1561.52|2231.76
this|128|426.06|4.83|899.94|3511.19|4320.65
this|512|410.84|4.94|3715.17|16819.24|20016.79
master|1|402.02|5.45|7.46|8.74|37.34
master|16|401.51|5.51|119.52|475.79|919.10
master|64|405.23|4.80|473.39|1506.29|2981.45
master|128|414.36|5.05|925.11|3040.14|4031.46
master|512|412.20|4.62|3702.80|12163.09|19075.38

## Sysbench common



branch|scenario|threads|qps|avg|99|999
-|-|-|-|-|-|-
this|oltp_read_write|200|82390|2.17|14.6|28.8
master|oltp_read_write|200|83619|2.12|14.2|27.7
this|oltp_read_write|400|86140|4.09|27.8|48.8
master|oltp_read_write|400|86207|4.03|27.3|47.4
this|oltp_read_write|800|84894|8.22|54.3|88.6
master|oltp_read_write|800|84330|8.23|52.6|73.5
this|oltp_insert|200|46839|4.16|13.9|22.3
master|oltp_insert|200|45583|4.28|14.2|21.8
this|oltp_insert|400|49513|7.91|25.6|38.2
master|oltp_insert|400|47189|8.34|40.4|47.0
this|oltp_insert|800|53247|14.8|47.1|57.2
master|oltp_insert|800|49555|15.9|57.1|80.3
this|oltp_update_non_index|200|52848|3.67|12.3|26.1
master|oltp_update_non_index|200|52431|3.71|11.3|24.7
this|oltp_update_non_index|400|62729|6.25|25.0|32.9
master|oltp_update_non_index|400|60274|6.51|26.3|34.6
this|oltp_update_non_index|800|64708|12.2|42.9|64.4
master|oltp_update_non_index|800|65010|12.1|46.2|69.4

## TPCC

WIP

zhangjinpeng1987:
> 

@MyonKeminta seems no obvious improvement?

MyonKeminta:
> > 
> 
> @MyonKeminta seems no obvious improvement?

This PR is a refactory and is the base of the next step of the optimization. We only expect this PR doesn't introduce any obvious performance regression.

MyonKeminta:
About half of the PR is tests. I think this PR is hard to split further into small ones. It would be good if you have any idea about that.

src/storage/lock_manager/mod.rs
```diff
@@ -124,31 +180,68 @@ pub trait LockManager: Clone + Send + 'static {
 
 // For test
 #[derive(Clone)]
-pub struct DummyLockManager;
+pub struct DummyLockManager {
```
sticnarf, line 183: Since it has some functionality now, do you think it more appropriate to call it a `MockLockManager` instead of `Dummy`?

src/storage/mod.rs
```diff
@@ -3055,6 +3057,7 @@ impl<E: Engine, L: LockManager, F: KvFormat> TestStorageBuilder<E, L, F> {
             config,
             pipelined_pessimistic_lock: Arc::new(AtomicBool::new(false)),
             in_memory_pessimistic_lock: Arc::new(AtomicBool::new(false)),
+            wake_up_delay_duration_ms: Arc::new(AtomicU64::new(0)),
```
sticnarf, line 3060: Is `0` always a good value in tests?

src/storage/txn/commands/mod.rs
```diff
@@ -398,62 +390,50 @@ pub struct WriteResult {
...
-            is_first_lock,
-            wait_timeout,
+            lock_info_pb,
+            parameters,
         }
     }
 }
 
+#[derive(Default)]
+pub struct ReleasedLocks(pub Vec<ReleasedLock>);
```
sticnarf, line 422: I think it better to wrap necessary methods instead of using `pub` directly.

src/storage/txn/commands/commit.rs
```diff
@@ -59,11 +59,10 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for Commit {
 
         let rows = self.keys.len();
         // Pessimistic txn needs key_hashes to wake up waiters
-        let mut released_locks = ReleasedLocks::new(self.lock_ts, self.commit_ts);
+        let mut released_locks = ReleasedLocks::new();
         for k in self.keys {
             released_locks.push(commit(&mut txn, &mut reader, k, self.commit_ts)?);
         }
-        released_locks.wake_up(context.lock_mgr);
```
sticnarf, line 66: It seems we no longer need `lock_mgr` in `WriteContext`.

src/storage/mvcc/txn.rs
```diff
@@ -114,8 +126,13 @@ impl MvccTxn {
         self.modifies.push(Modify::PessimisticLock(key, lock))
     }
 
-    pub(crate) fn unlock_key(&mut self, key: Key, pessimistic: bool) -> Option<ReleasedLock> {
-        let released = ReleasedLock::new(&key, pessimistic);
+    pub(crate) fn unlock_key(
+        &mut self,
+        key: Key,
+        pessimistic: bool,
+        commit_ts: Option<TimeStamp>,
```
sticnarf, line 133: Maybe we can just make it simple and consistent.

If we treat `commit_ts = 0` as rollback, can we just accept `0` instead of `None` here so as to avoid the additional `unwrap_or(0)` in `ReleasedLock::new`?

src/storage/mvcc/txn.rs
```diff
@@ -37,16 +37,28 @@ impl GcInfo {
 /// waiting for locks.
 #[derive(Debug, PartialEq)]
 pub struct ReleasedLock {
+    pub start_ts: TimeStamp,
+    pub commit_ts: TimeStamp,
+    pub key: Key,
     /// The hash value of the lock.
     pub hash: u64,
```
sticnarf, line 44: Is it still useful now?

src/storage/txn/commands/check_secondary_locks.rs
```diff
@@ -142,10 +142,10 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for CheckSecondaryLocks {
 
         let mut rows = 0;
         if let SecondaryLocksStatus::RolledBack = &result {
-            // Lock is only released when result is `RolledBack`.
-            released_locks.wake_up(context.lock_mgr);
             // One row is mutated only when a secondary lock is rolled back.
             rows = 1;
+        } else {
+            released_locks.0.clear();
```
sticnarf, line 148: This seems unnecessary?

src/server/lock_manager/waiter_manager.rs
```diff
@@ -568,84 +592,92 @@ impl WaiterManager {
...
                 TASK_COUNTER_METRICS.wait_for.inc();
             }
-            Task::WakeUp {
-                lock_ts,
-                hashes,
-                commit_ts,
-            } => {
-                self.handle_wake_up(lock_ts, hashes, commit_ts);
+
+            Task::RemoveLockWait { token } => {
```
cfzjywxk, line 655: Do we have tests to cover the removing triggered by different reasons or it's planned in the next PRs?

src/server/lock_manager/mod.rs
```diff
@@ -206,7 +199,8 @@ impl LockManager {
         &self,
         host: &mut CoprocessorHost<impl KvEngine>,
     ) {
-        let role_change_notifier = RoleChangeNotifier::new(self.detector_scheduler.clone());
+        let role_change_notifier =
```
sticnarf, line 202: Looks like an unnecessary change?

src/server/lock_manager/mod.rs
```diff
@@ -225,78 +219,79 @@ impl LockManager {
...
-
-    fn remove_from_detected(&self, txn_ts: TimeStamp) -> bool {
-        let mut detected = self.detected[detected_slot_idx(txn_ts)].lock();
-        detected.remove(&txn_ts)
-    }
 }
 
 impl LockManagerTrait for LockManager {
+    fn allocate_token(&self) -> LockWaitToken {
+        LockWaitToken(Some(self.token_allocator.fetch_add(1, Ordering::SeqCst)))
```
sticnarf, line 237: `Ordering::Relaxed` is enough for an id allocator because we only need atomicity.

src/server/lock_manager/mod.rs
```diff
@@ -225,78 +219,79 @@ impl LockManager {
...
-        is_pessimistic_txn: bool,
-    ) {
+    fn update_wait_for(&self, updated_items: Vec<UpdateWaitForEvent>) {
+        self.waiter_mgr_scheduler.update_wait_for(updated_items);
+    }
+
+    fn on_keys_wakeup(&self, _wake_up_events: Vec<KeyWakeUpEvent>) {}
+
+    fn remove_lock_wait(&self, token: LockWaitToken) {
         // If `hashes` is some, there may be some waiters waiting for these locks.
```
sticnarf, line 291: The comment needs updating

src/server/lock_manager/waiter_manager.rs
```diff
@@ -512,51 +532,55 @@ impl WaiterManager {
...
             return;
         }
-        let duration: Duration = self.wake_up_delay_duration.into();
-        let new_timeout = Instant::now() + duration;
-        for hash in hashes {
-            let lock = Lock { ts: lock_ts, hash };
-            if let Some((mut oldest, others)) = wait_table.remove_oldest_waiter(lock) {
-                // Notify the oldest one immediately.
+        // let duration: Duration = self.wake_up_delay_duration.into();
+        // let _new_timeout = Instant::now() + duration;
```
sticnarf, line 558: Should remove them?

src/storage/txn/commands/commit.rs
MyonKeminta: Yes. Let's do these kind of fixups in other PRs.

src/storage/txn/scheduler.rs
```diff
@@ -722,27 +737,126 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+                        MvccErrorInner::WriteConflict {
+                            start_ts: lock_info.parameters.start_ts,
+                            conflict_start_ts: released_lock.start_ts,
+                            conflict_commit_ts: released_lock.commit_ts,
+                            key: released_lock.key.into_raw().unwrap(),
+                            primary: lock_info.parameters.primary,
+                            reason: kvrpcpb::WriteConflictReason::PessimisticRetry,
+                        },
+                    )));
+                    cb(Err(e.into()));
```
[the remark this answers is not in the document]
MyonKeminta: They are currently still able to be tracked in `Waiter.cancel` in `WatierManager`. I think we can consider move them here later.

src/server/lock_manager/waiter_manager.rs
```diff
@@ -484,11 +502,11 @@ pub struct WaiterManager {
     detector_scheduler: DetectorScheduler,
     /// It is the default and maximum timeout of waiter.
     default_wait_for_lock_timeout: ReadableDuration,
-    /// If more than one waiters are waiting for the same lock, only the
-    /// oldest one will be waked up immediately when the lock is released.
-    /// Others will be waked up after `wake_up_delay_duration` to reduce
-    /// contention and make the oldest one more likely acquires the lock.
-    wake_up_delay_duration: ReadableDuration,
+    // /// If more than one waiters are waiting for the same lock, only the
```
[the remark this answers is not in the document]
MyonKeminta: Ah, sorry, I forgot to remove it.

src/server/lock_manager/waiter_manager.rs
MyonKeminta: Currently a waiter in `WaiterManager` may only be removed in following cases:

* Receiving an `RemoveLockWait` message outside (because the lock-wait-entry is popped out from the queue and woken up, thus `LockWaitContext.finish_request` is called)
* Canceled due to timeout.
* Canceled due to deadlock

Canceling by region changes is not supported in this PR.
But I just noticed that `RemoveLockWait` is not covered in tests. I'll add it later.

src/storage/lock_manager/mod.rs
```diff
@@ -67,9 +91,14 @@ impl From<u64> for WaitTimeout {
...
+    pub key: Key,
+    pub lock_digest: LockDigest,
+    pub lock_info: LockInfo,
+}
+
 /// Uniquely identifies a lock-waiting request in a `LockManager`.
-///
-/// Not used yet, but necessary for implementing `LockWaitQueues`.
 #[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
 pub struct LockWaitToken(pub Option<u64>);
```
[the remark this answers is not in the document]
MyonKeminta: When there is not a valid token allocated. None represents invalid.
Actually it can be a simple u64 and use `Option<LockWaitToken>` when it's possibly invalid. I'll consider update this in further PRs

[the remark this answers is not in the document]
sticnarf: Is it possible that a token fails to be allocated?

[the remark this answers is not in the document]
MyonKeminta: No, but it's possible to be not-allocated-yet.

src/server/lock_manager/mod.rs
MyonKeminta: I'm not sure, can it guarantee the uniqueness of the result?

sticnarf: Yes. Here are some discussions:
https://stackoverflow.com/questions/54639439/will-fetch-add-with-relaxed-memory-order-return-unique-values
https://www.reddit.com/r/rust/comments/aeii1i/is_memory_orderingrelaxed_sufficient_to_generate/

src/server/lock_manager/waiter_manager.rs
```diff
@@ -220,179 +247,155 @@ impl Waiter {
...
+    fn take_waiter(&mut self, token: LockWaitToken) -> Option<Waiter> {
+        let waiter = self.waiter_pool.remove(&token)?;
         self.waiter_count.fetch_sub(1, Ordering::SeqCst);
-        WAIT_TABLE_STATUS_GAUGE.txns.dec();
-        if waiters.is_empty() {
-            self.remove(lock);
-        }
+        self.wait_table
+            .remove(&(waiter.wait_info.lock_digest.hash, waiter.start_ts));
+        // WAIT_TABLE_STATUS_GAUGE.txns.dec();
```
sticnarf, line 359: Is the gauge still valid?

src/storage/txn/scheduler.rs
```diff
@@ -722,27 +737,126 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+                    &released_lock.key,
+                    released_lock.start_ts,
+                    released_lock.commit_ts,
+                    wake_up_delay_duration_ms,
+                ) {
+                    Some(e) => e,
+                    None => return,
+                };
+
+            legacy_wake_up_list.push((lock_wait_entry, released_lock));
```
sticnarf, line 812: Although `allow_lock_with_conflict` is not implemented now, I have a question that should the new kind of lock entries be added to this `legacy_wake_up_list`?

src/server/lock_manager/waiter_manager.rs
```diff
@@ -220,179 +247,155 @@ impl Waiter {
...
-        } else {
-            WAIT_TABLE_STATUS_GAUGE.txns.inc();
-            None
-        }
-        // Here we don't increase waiter_count because it's already updated in
-        // LockManager::wait_for()
-    }
+    fn add_waiter(&mut self, token: LockWaitToken, waiter: Waiter) -> bool {
+        self.wait_table
+            .insert((waiter.wait_info.lock_digest.hash, waiter.start_ts), token);
```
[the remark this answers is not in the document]
MyonKeminta: Indexing by (hash, start_ts) is here to provide compatibility since the deadlock detector tries to wake up by (hash, start_ts) when it founds deadlock. I only supposed to keep the token unique, and just replace and ignore it when there are duplicated (hash, start_ts) pair. May there be any problem?

src/server/lock_manager/waiter_manager.rs
MyonKeminta: I forgot to remove this line. I want to use replace it with `LOCK_WAIT_QUEUE_ENTRIES_GAUGE_VEC`.

src/storage/txn/scheduler.rs
MyonKeminta: No. Those entries will be returned and kept temporarily, and when the current cmd finished and releases its latches, we keep necessary latches in it (do not release it) and then start a new command to execute these woken up entries.

src/server/lock_manager/waiter_manager.rs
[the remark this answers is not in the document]
sticnarf: OK. Now I see the duplicated waiter token will finally be cleared when it times out.

It seems not a big problem that the old duplicated waiter is changed to get notified at timeout instead of immediately.

sticnarf (commented):
The rest looks okay to me.

As it's so large a pull request, I'm not confident that code review could find most of the problems. We probably need to rely on further integration tests in this aspect.

sticnarf (approved):

MyonKeminta:
@ekexium @you06 Please help me review this PR if you have time, thanks! Ask me if you have any question.

cfzjywxk (approved):
LGTM

src/server/lock_manager/mod.rs
```diff
@@ -225,78 +220,75 @@ impl LockManager {
...
-        // Try to wake up them.
-        if !hashes.is_empty() && self.has_waiter() {
-            self.waiter_mgr_scheduler
-                .wake_up(lock_ts, hashes, commit_ts);
-        }
-        // If a pessimistic transaction is committed or rolled back and it once sent
-        // requests to detect deadlock, clean up its wait-for entries in the
-        // deadlock detector.
-        if is_pessimistic_txn && self.remove_from_detected(lock_ts) {
-            self.detector_scheduler.clean_up(lock_ts);
```
ekexium, line 299: Dumb question: is there an equivalent part of the clean up logic now?

MyonKeminta: Here the old logic wakes up all waiters waiting for the lock with `lock_ts` on the specified key, and cleans all edges that waits for the `lock_ts` from the detector. In new logic, waking-up happens in the new lock waiting queue, and when a lock-waiting request is finished (either canceled or resumed and successfully acquired, which is not yet supported), the `remove_lock_wait` function in LockManager will be invoked, which then leads to a call to `clean_up_wait_for` ([here](https://github.com/tikv/tikv/blob/4785619d18027001c16af878fad567c02ae0dd42/src/server/lock_manager/waiter_manager.rs#L558))

src/server/lock_manager/metrics.rs
```diff
@@ -12,6 +12,7 @@ make_auto_flush_static_metric! {
             detect,
             clean_up_wait_for,
             clean_up,
+            update_wait_for,
```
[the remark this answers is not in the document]
MyonKeminta: Yes, it's unused currently, so does the `update_wait_for` method of `LockManager`.

src/server/lock_manager/config.rs
```diff
@@ -88,29 +89,35 @@ impl LockManagerConfigManager {
...
-        };
+        if let Some(p) = change.remove("wait_for_lock_timeout").map(Into::into) {
+            self.waiter_mgr_scheduler.change_config(Some(p));
+            self.detector_scheduler.change_ttl(p.into());
+        }
+        if let Some(p) = change
+            .remove("wake_up_delay_duration")
+            .map(ReadableDuration::from)
+        {
+            info!(
```
TonsnakeLin, line 114: Why only the config `wake_up_delay_duration` need to print log?

MyonKeminta: The logs about the two config items mentioned here were previously printed in `waiter_manager.rs` when handling `Task::ChangeConfig` message. Now the `wake_up_delay_duration` is changed to be handled somewhere else, so I printed it here. When changing `wait_for_lock_timeout`, the log will still be printed at the old place.

src/server/lock_manager/mod.rs
```diff
@@ -225,78 +220,75 @@ impl LockManager {
...
 
         // If it is the first lock the transaction tries to lock, it won't cause
         // deadlock.
         if !is_first_lock {
-            self.add_to_detected(start_ts);
-            self.detector_scheduler.detect(start_ts, lock, diag_ctx);
+            self.detector_scheduler
+                .detect(start_ts, wait_info.clone(), diag_ctx.clone()); // TODO: Try to avoid cloning.
         }
+        self.waiter_mgr_scheduler.wait_for(
```
TonsnakeLin, line 272: Is there any special meanings that you move `self.waiter_mgr_scheduler.wait_for` behind `self.detector_scheduler.detect`?

src/server/lock_manager/waiter_manager.rs
```diff
@@ -220,179 +247,152 @@ impl Waiter {
...
 
+    #[allow(dead_code)]
     fn reset_timeout(&self, deadline: Instant) {
         self.delay.reset(deadline);
     }
 
-    /// `Notify` consumes the `Waiter` to notify the corresponding transaction
-    /// going on.
-    fn notify(self) {
+    /// Consumes the `Waiter` to notify the corresponding transaction `going on.
```
TonsnakeLin, line 255:
```suggestion
    /// Consumes the `Waiter` to notify the corresponding transaction going on.
```

src/server/lock_manager/waiter_manager.rs
```diff
@@ -220,179 +247,152 @@ impl Waiter {
...
+    fn cancel(self, error: Option<StorageError>) -> KeyLockWaitInfo {
         let elapsed = self.start_waiting_time.saturating_elapsed();
         GLOBAL_TRACKERS.with_tracker(self.diag_ctx.tracker, |tracker| {
             tracker.metrics.pessimistic_lock_wait_nanos = elapsed.as_nanos() as u64;
         });
         WAITER_LIFETIME_HISTOGRAM.observe(duration_to_sec(elapsed));
         // Cancel the delay timer to prevent removing the same `Waiter` earlier.
         self.delay.cancel();
-        self.cb.execute(self.pr);
+        if let Some(error) = error {
```
TonsnakeLin, line 264: Using `(self.cancel_callback)(error.unwrap())` maybe simpler, or `error` maybe `none` ?

src/server/lock_manager/waiter_manager.rs
```diff
@@ -220,179 +247,152 @@ impl Waiter {
...
-        });
-        self.pr = ProcessResult::Failed {
-            err: StorageError::from(TxnError::from(mvcc_err)),
-        };
+    fn cancel_for_finished(self) -> KeyLockWaitInfo {
+        self.cancel(None)
+    }
+
+    fn cancel_for_timeout(self, _skip_resolving_lock: bool) -> KeyLockWaitInfo {
+        let lock_info = self.wait_info.lock_info.clone();
```
TonsnakeLin, line 275: Why do we need to clone `lock_info`?

ekexium (approved):
I don't see other problems

src/server/lock_manager/mod.rs
MyonKeminta: No. It seems to be an accident. I'll change it back.

src/server/lock_manager/waiter_manager.rs
MyonKeminta: If we move it, we will not be able to call another method `cancel` on `self` since `self` is partially moved. And also, the function needs to return a complete `KeyLockWaitInfo` to be used by the caller.

src/server/lock_manager/waiter_manager.rs
```diff
@@ -499,11 +511,13 @@ impl WaiterManager {
...
     ) -> Self {
+        let wait_table = WaitTable::new(waiter_count);
+
         Self {
-            wait_table: Rc::new(RefCell::new(WaitTable::new(waiter_count))),
+            wait_table: Rc::new(RefCell::new(wait_table)),
             detector_scheduler,
             default_wait_for_lock_timeout: cfg.wait_for_lock_timeout,
-            wake_up_delay_duration: cfg.wake_up_delay_duration,
+            // wake_up_delay_duration: cfg.wake_up_delay_duration,
```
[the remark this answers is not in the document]
MyonKeminta: Yes. I forgot it.

PR #13464, opened 2022-09-14
engine: update raft-engine

tabokie:

Issue Number: Ref #11119

What's Changed:

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


tabokie:
It's just an update. Some trait is changed and some config is deprecated. You can read the release note for more details: https://github.com/tikv/raft-engine/releases/tag/0.3.0

etc/config-template.toml
```diff
@@ -1047,12 +1047,6 @@
 ## If you set this configuration item to `0`, compression is disabled.
 # batch-compression-threshold = "8KB"
 
-## Specifies the maximum accumulative size of buffered writes. When this
```
BusyJay, line 1050: What's the new sync behavior?

tabokie: There's no limit of buffered writes now.

BusyJay: Will it cause too many large IO?

tabokie: It will if you write a lot with sync=false, but it shouldn't be a problem with tikv. Note that the IO of one write group will not be split in both the old and new version.

BusyJay (approved):

Connor1996 (approved):

PR #13484, opened 2022-09-16
engine_tirocks: port iterator and snapshot

BusyJay:

Issue Number: Ref #13058 

What's Changed:

```commit-message
Everything is almost the same as engine_rocks except engine_tirocks
put snapshot into an `Arc`. The old implementation is not safe as the
snapshot can be released while the iterator is still being used.
```

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_tirocks/src/engine.rs
+++ b/components/engine_tirocks/src/engine.rs
@@ -3,7 +3,15 @@
 use std::{fs, path::Path, sync::Arc};
 
 use engine_traits::{Code, Error, Result, Status};
-use tirocks::{db::RawCfHandle, Db};
+use tirocks::{
+    db::RawCfHandle,
+    option::{ReadOptions, WriteOptions},
+    Db, Iterator,
+};
+
+use crate::{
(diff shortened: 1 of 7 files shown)


Connor1996 (approved):
LGTM

components/engine_tirocks/src/engine.rs
```diff
@@ -43,7 +51,250 @@ impl RocksEngine {
...
+        RocksSnapshot::new(self.0.clone())
+    }
+}
+
+impl engine_traits::Iterable for RocksEngine {
+    type Iterator = RocksEngineIterator;
+
+    fn iterator_opt(&self, cf: &str, opts: engine_traits::IterOptions) -> Result<Self::Iterator> {
+        let opt = engine_iterator::to_tirocks_opt(opts);
+        let handle = util::cf_handle(&self.0, cf)?;
```
tabokie, line 87: Why use this instead of self.cf()?

tabokie (approved):

PR #13485, opened 2022-09-19
gc_worker: make unsafe detroy range compatible with multi-rocks db

### LINKED ISSUE #13406: Make GC compatible with Multi-Rocks DB
## Development Task

In Multi-Rocks DB, every region has it's own rocks db, so GC should be made some adaptations for it.

SpadeA-Tang:

Issue Number: Close #13406

What's Changed:

This PR makes unsafe-destroy-range compatible with multi-rocks db.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


src/server/gc_worker/gc_worker.rs
```diff
@@ -817,6 +818,45 @@ where
...
                 warn!("unsafe destroy range: failed sending ClearRegionSizeInRange"; "err" => ?e);
             });
+        } else {
+            let cfs = &[CF_LOCK, CF_DEFAULT, CF_WRITE];
+            let keys = vec![start_key.clone(), end_key.clone()];
+            let regions = get_regions_for_gc(self.store_id, &keys, regions_provider)?;
+            let count = regions.len();
+
+            let mut region_modifies = HashMap::default();
+            for i in 0..count {
```
MyonKeminta, line 828: consider: `for (i, region) in regions.into_iter().enumerate()`

src/server/gc_worker/gc_worker.rs
```diff
@@ -817,6 +818,45 @@ where
...
+                };
+                let range_end = if i == count - 1 {
+                    end_key.clone()
+                } else {
+                    Key::from_raw(regions[i].get_end_key())
+                };
+
+                let mut modifies = Vec::new();
+                for cf in cfs {
+                    modifies.push(Modify::DeleteRange(
```
MyonKeminta, line 842: cc @BusyJay 
Is it possible to delete large ranges in a safe way (instead of bypassing raft) after supporting multi-rocks?

src/server/gc_worker/gc_worker.rs
```diff
@@ -817,6 +818,45 @@ where
                 // Warn and ignore it.
                 warn!("unsafe destroy range: failed sending ClearRegionSizeInRange"; "err" => ?e);
             });
+        } else {
+            let cfs = &[CF_LOCK, CF_DEFAULT, CF_WRITE];
+            let keys = vec![start_key.clone(), end_key.clone()];
+            let regions = get_regions_for_gc(self.store_id, &keys, regions_provider)?;
```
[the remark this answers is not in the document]
SpadeA-Tang: Generally, it is. But sometimes keys may only contain one key, so in this case, it is not appropriate to call range?

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: Good suggestion.

src/server/gc_worker/gc_worker.rs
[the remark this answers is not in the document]
MyonKeminta: It can be considered as a range [`key`, `key.append(0)`) where key is in raw format, thus there is at most one key in the range.

src/server/gc_worker/gc_worker.rs
BusyJay: Yes, just send them like normal commands.

src/server/gc_worker/gc_worker.rs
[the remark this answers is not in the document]
SpadeA-Tang: So，just rename the method? Or I should change the keys to `Range` type?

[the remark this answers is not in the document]
MyonKeminta: Maybe you can rename it to something like `get_regions_for_range_of_keys` without changing the parameter it accepts. And explain in comments that the function is used to find *all* regions in the range inferred by a list of sorted keys.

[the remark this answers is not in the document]
SpadeA-Tang: Thanks for the suggestion.

src/server/gc_worker/gc_worker.rs
```diff
@@ -740,20 +746,20 @@ where
...
             .send(FlowInfo::BeforeUnsafeDestroyRange(ctx.region_id))
             .unwrap();
-        let local_storage = self.engine.kv_engine().unwrap();
 
-        // Convert keys to RocksDB layer form
-        // TODO: Logic coupled with raftstore's implementation. Maybe better design is
-        // to do it in somewhere of the same layer with apply_worker.
-        let start_data_key = keys::data_key(start_key.as_encoded());
-        let end_data_key = keys::data_end_key(end_key.as_encoded());
+        if let Some(local_storage) = self.engine.kv_engine() {
```
tonyxuqqi, line 750: please add comment this branch is for single rocksdb case.

tonyxuqqi (approved):

src/server/gc_worker/gc_worker.rs
SpadeA-Tang: Fine.

MyonKeminta (approved):

PR #13486, opened 2022-09-19
storage: Add new implementation of lock waiting queue

MyonKeminta:

Issue Number: Ref #13298

What's Changed:

```commit-message
Implements a new version of lock waiting queue, which is important for supporting resumable acquire_pessimistic_lock requests. (ref #13298) 

* Make `storage::lock_manager` an directory
* Add new implementation of the lock waiting queue and other related stuff.
```

\* Please refer to the *Lock Waiting Queue* section of the [RFC](https://github.com/tikv/rfcs/blob/b597d74fd2e1f5a1eec8562ef74e87656e54ca5a/text/0100-enhanced-pessimistic-lock-queueing.md#lock-waiting-queue) for why this change is needed.

Some unresolved problems:

* Stale entries in the queue: It's possible that the request is cancelled outside the queue (may be cancelled due to timeout in `WaiterManager`, which will be implemented in https://github.com/tikv/tikv/pull/13447 ). Possible solutions:
    * Use an atomic counter to count valid entries in the queue of each key. When it becomes 0 but the queue is not empty, remove it from the map.
    * Periodically iterate the map and remove the stale entries.
    * Introduce a priority queue that supports efficiently removing (by either introducing third-party library or implementing by ourselves), and remove the entry when it's cancelled outside.
* Potential inefficient deadlock detection in some rare cases:
    * An example: let transaction 10 hold the lock, and the waiting queue be:  [20, 30, 50]
    * 10 Release the lock, 20 is woken up, 30 & 50 is going to be notified after a delay. The queue: [30*, 50*]
    * Transaction 15 acquired the lock, then a request of transaction 40 arrives and waits for 15, then the queue is: [30*, 40, 50*]
    * Delayed wake up happened, pops 30 and stops at 40 since it arrives later than the previous waking up, then the queue becomes: [40, 50*]
    * Here 50 waits for 15, but no additional deadlock detection is performed. If 15 is also waiting for 50, the deadlock won't be found until the 50's lock-waiting request exceeds its timeout.
    * Possible solutions:
        * Ignore this case since it won't be frequent, just like we didn't handle the case that waiter won't be woken up if the leader transfers during waiting.
        * Iterate and rebuild the whole priority queue to get all entries to be woken up.
        * Use double queues to manage entries that have / haven't been scheduled for delayed wake up respectively.
* Lacking of metrics

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,997 @@
...
+            lock_mgr,
+        }
+    }
+
+    /// Enqueues a lock wait entry. The key is indicated by the `key` field of
+    /// the `lock_wait_entry`. The caller also need to provide the
+    /// information of the current-holding lock.
+    pub fn push_lock_wait(
+        &self,
+        mut lock_wait_entry: Box<LockWaitEntry>,
```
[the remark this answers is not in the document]
MyonKeminta: The struct contains too many fields. I've seen warnings complaining about large variant when I tried to put it in enums. Also, it may be frequently moved when it's put in `BinaryHeap`s. So in order to reduce the cost of moving, I think always putting it in Box should be better.

src/storage/types.rs
```diff
@@ -121,6 +122,22 @@ pub struct PrewriteResult {
     pub one_pc_commit_ts: TimeStamp,
 }
 
+#[derive(Clone)]
```
TonsnakeLin, line 125: why not add `Clone` to `cfg_attr(test...)` ?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,997 @@
...
+    pub req_states: Option<Arc<LockWaitContextSharedState>>,
+    pub current_legacy_wake_up_cnt: Option<usize>,
+    pub key_cb: Option<SyncWrapper<PessimisticLockKeyCallback>>,
+}
+
+/// A wrapper to [`LockWaitEntry`] to define its ordering.
+/// A [`LockWaitEntry`] takes precedence over another if it has smaller
+/// `start_ts`.
+#[repr(transparent)]
+struct LockWaitEntryComparableWrapper(pub Box<LockWaitEntry>);
```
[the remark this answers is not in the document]
MyonKeminta: Yes, if you think it would be better. I was trying to hide the comparison definition, since I think it doesn't make sense to compare them outside the module.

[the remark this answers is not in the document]
sticnarf: Ah, I see. You want it only comparable in this module. I think it makes sense although the name looks a bit too long. `ComparableLockWaitEntry` may be a bit shorter. And `into_inner` is the more common name for than `unwrap` for extracting the inner object.

src/storage/types.rs
MyonKeminta: It's not supposed only to be used in tests.
But it's actually totally unused yet. I'll remove it.

src/storage/lock_manager/lock_wait_context.rs
```diff
@@ -0,0 +1,254 @@
...
+use crate::storage::{
+    errors::SharedError,
+    lock_manager::{lock_waiting_queue::PessimisticLockKeyCallback, LockManager, LockWaitToken},
+    Error as StorageError, PessimisticLockRes, ProcessResult, StorageCallback,
+};
+
+pub struct LockWaitContextInner {
+    #[allow(dead_code)]
+    pr: ProcessResult,
+    cb: StorageCallback,
```
sticnarf, line 38: Could you add comments _here_ about what `pr` and `cb` actually is and when they will be called or used. It will be easier for readers to learn the code.

src/storage/lock_manager/lock_wait_context.rs
```diff
@@ -0,0 +1,254 @@
...
+    #[allow(dead_code)]
+    lock_wait_token: LockWaitToken,
+}
+
+pub struct LockWaitContextSharedState {
```
sticnarf, line 43: Could you also comment about who will hold the shared state? It's especially useful when this PR is extracted without knowing how other code will integrate with it.

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,942 @@
...
+    /// Dequeues the head of the lock waiting queue of the specified key,
+    /// assuming the popped entry will be woken up.
+    ///
+    /// If it's waking up a legacy request and the queue is not empty, a future
+    /// will be returned and the caller will be responsible for executing it.
+    /// The future waits until `wake_up_delay_duration` is elapsed since the
+    /// most recent waking-up, and then wakes up all lock waiting entries that
+    /// exists at the time when the latest waking-up happens. The future
+    /// will return a `LockWaitEntry` if a resumable entry is popped out
+    /// from the queue while executing, and in this case the caller will be
```
cfzjywxk, line 257: Is the returned future expected to be executed by the scheduler?

src/storage/lock_manager/lock_wait_context.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Get the callback that should be invoked when finishes executing the
+    /// scheduler command that issued the lock-waiting.
+    ///
+    /// When we support partially finishing a pessimistic lock request (i.e.
+    /// when acquiring lock multiple keys in one single request, allowing
+    /// some keys to be locked successfully while the others are blocked or
+    /// failed), this will be useful for handling the result of the first
+    /// write batch. But currently, the first write batch of a lock-waiting
+    /// request is always empty, so the callback is just noop.
```
TonsnakeLin, line 105: Why is the first write batch of a lock-waiting request always empty?

src/storage/lock_manager/lock_wait_context.rs
```diff
@@ -0,0 +1,254 @@
...
+            return;
+        };
+
+        self.shared_states.finished.store(true, Ordering::Release);
+
+        // TODO: Uncomment this after the corresponding change of `LockManager` is done.
+        // self.lock_manager
+        //     .remove_lock_wait(ctx_inner.lock_wait_token);
+
+        if !self.allow_lock_with_conflict {
```
TonsnakeLin, line 149: What does it stand for when self.allow_lock_with_conflict is false ?

src/storage/lock_manager/lock_wait_context.rs
sticnarf: I think it's a placeholder for more complex optimizations in the future. Well, it may be a question whether we should introduce them now.

src/storage/lock_manager/lock_wait_context.rs
cfzjywxk: For the new mode, the pessimistic lock request would still be processed, while in the legacy mode the write conflict error would be returned to the client directly.

src/storage/lock_manager/lock_wait_context.rs
```diff
@@ -0,0 +1,254 @@
...
+        rx.recv_timeout(Duration::from_millis(20)).unwrap_err();
+        (ctx.get_callback_for_blocked_key())(Err(SharedError::from(write_conflict())));
+        let res = rx.recv().unwrap().unwrap_err();
+        assert!(matches!(
+            &res,
+            StorageError(box StorageErrorInner::Txn(TxnError(
+                box TxnErrorInner::Mvcc(MvccError(box MvccErrorInner::WriteConflict { .. }))
+            )))
+        ));
+        rx.recv().unwrap_err();
```
TonsnakeLin, line 237: The receiver `rx` has received the `result` at line 230, it called `recv()` again here , is it blocked here?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,942 @@
...
+    /// Then `wake_up_delay_duration` is elapsed since the previous waking up.
+    /// Here, we expect that the lock wait entries 30 and 40 can be woken
+    /// up, since they exists when the previous waking up occurs. But we
+    /// don't want to wake up later-arrived entries (50 and 60) since it
+    /// introduces useless pessimistic retries to transaction 50 and 60 when
+    /// they don't need to. The solution is, only wake up the entries that
+    /// has `entry.current_legacy_wake_up_cnt <
+    /// key_lock_wait_state.legacy_wake_up_cnt`. Therefore, we only wakes up
+    /// entries 30 and 40 who has `current_legacy_wake_up_cnt < 1`, while 50 and
+    /// 60 will be left untouched.
```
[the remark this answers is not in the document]
MyonKeminta: I'll try to emphasize that it doesn't take effect for resumable requests.
I think it doesn't seem to be an optimization.

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,942 @@
...
+            let current_delay_ms = delay_ms.load(Ordering::Acquire);
+            if current_delay_ms == 0 {
+                // Cancelled.
+                return None;
+            }
+
+            if current_delay_ms <= prev_delay_ms
+                || (start_time.saturating_elapsed().as_millis() as u64) >= current_delay_ms
+            {
+                // Timed out.
```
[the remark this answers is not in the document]
MyonKeminta: Here it checks whether `wake-up-delay-duration` has been past since the most recent waking up on the key, so that it can perform a delayed_notify_all. Is `wait-for-lock-timeout` useful here?

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: Yes. We just say the caller is responsible to execute it, and in fact the caller is always Scheduler (except unit tests)

src/storage/lock_manager/lock_wait_context.rs
MyonKeminta: I'll add more comments to the field in `PessimisticLockParameters`.

src/storage/lock_manager/lock_wait_context.rs
MyonKeminta: It's expected the sender is dropped and the receiver will get a error indicating the channel is disconnected. I'll add some comments to explain this.

src/storage/lock_manager/lock_waiting_queue.rs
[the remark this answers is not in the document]
cfzjywxk: If the `delay_ms` is updated continuously, and the wait time has exceeded the `wait-for-lock-timeout` which is the maximum wait time, would it be possible that this loop is still being executed though the waiter has been canceled?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+            inner: Arc::new(LockWaitQueueInner {
+                queue_map: dashmap::DashMap::new(),
+                id_allocated: AtomicU64::new(1),
+            }),
+            lock_mgr,
+        }
+    }
+
+    /// Enqueues a lock wait entry. The key is indicated by the `key` field of
+    /// the `lock_wait_entry`. The caller also need to provide the
```
TonsnakeLin, line 237:
```suggestion
    /// the `lock_wait_entry`. The caller also needs to provide the
```

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+    /// information of the current-holding lock.
+    pub fn push_lock_wait(
+        &self,
+        mut lock_wait_entry: Box<LockWaitEntry>,
+        current_lock: kvrpcpb::LockInfo,
+    ) {
+        let mut entry = self
```
TonsnakeLin, line 244: `entry` is a common name of the elements in a Map, and it's easy to confuse with `LockWaitEntry`. 
um...,  it maybe ok ?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+    fn pop_for_waking_up_impl(
+        &self,
+        key: &Key,
+        conflicting_start_ts: TimeStamp,
+        conflicting_commit_ts: TimeStamp,
+        wake_up_delay_duration_ms: Option<u64>,
+    ) -> Option<(Box<LockWaitEntry>, Option<DelayedNotifyAllFuture>)> {
+        let mut result = None;
+
+        // We don't want other thread insert insert any more entries between finding the
```
TonsnakeLin, line 292:
```suggestion
        // We don't want other threads insert any more entries between finding the
```

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+            delayed_notify_all_state: None,
+        }
+    }
+}
+
+pub type DelayedNotifyAllFuture = Pin<Box<dyn Future<Output = Option<Box<LockWaitEntry>>> + Send>>;
+
+pub struct LockWaitQueueInner {
+    queue_map: dashmap::DashMap<Key, KeyLockWaitState>,
+    id_allocated: AtomicU64,
```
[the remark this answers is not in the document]
MyonKeminta: It's used to allocate ID for `KeyLockWaitState`. The ID is used in `delayed_notify_all` to prevent the following case:

1. Waking up occurs on key A, and there are remaining lock-waiting requests so schedules a call to `delayed_notify_all`
2. Waking up continues occurring on key A and the queue is cleared, so that the key is removed from the queue map.
3. Lock wait occurs on key A, and a new KeyLockWaitState (with initial legacy_wake_up_cnt = 0) is added to the map.
4. `delayed_notify_all` scheduled at step 1 is called, but nothing in the current queue should be notified.

With this ID, the previously-removed `KeyLockWaitState` and the later inserted one can be distinguished. So in this case the `delayed_notify_all` invocation at step 4 does nothing.

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: Maybe there can be some better name for  `LockWaitEntry?
cc @sticnarf , Is there any suggested name?

TonsnakeLin: My intention is to rename the temporary variable `entry` to a name relative to its meanings, like `lock_wait_state` or else.

sticnarf: Yep, I also agree the local variable `entry` can be changed to `state because `dashmap::Entry` can be deref as `KeyLockWaitState`. I don't have any good name suggestion for `LockWaitEntry`...

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+        // value of `delay_ms`. So we loop until we find that the elapsed
+        // duration is larger than `delay_ms`.
+        loop {
+            let current_delay_ms = delay_ms.load(Ordering::Acquire);
+            if current_delay_ms == 0 {
+                // Cancelled.
+                return None;
+            }
+
+            if current_delay_ms <= prev_delay_ms
```
TonsnakeLin, line 393: What happened if `current_delay_ms < prev_delay_ms` ?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+                    conflicting_start_ts.into(),
+                    conflicting_commit_ts.into(),
+                    Some(50),
+                )
+                .unwrap();
+            assert!(f.is_none());
+            res
+        }
+
+        fn must_not_contains_key(&self, key: &[u8]) {
```
TonsnakeLin, line 694:
```suggestion
        fn must_not_contain_key(&self, key: &[u8]) {
```

src/storage/lock_manager/lock_waiting_queue.rs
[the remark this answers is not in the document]
MyonKeminta: Yes. But it's still possible that there are some new waiters added later and need to be woken up after delay. So in this case we still continue. For example, consider the queue:
[10, 20, 30]
Wake up: [20*, 30*] and schedule `delayed_notify_all` (item marked with * means need to be woken up after delay)
Key is locked by other transaction and some new waiters are added: [20*, 30*, 40, 50]
Wake up: [30*, 40*, 50*]
Wait for 30 timeout: [(30 stale), 40*, 50*]
And then call `delayed_notify_all`.

In this case, though 30 (which exists when we schedule `delayed_notify_all`) is expired, but it still need to cancel 40 and 50 which is added later.

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+    fn pop_for_waking_up_impl(
+        &self,
+        key: &Key,
+        conflicting_start_ts: TimeStamp,
+        conflicting_commit_ts: TimeStamp,
+        wake_up_delay_duration_ms: Option<u64>,
+    ) -> Option<(Box<LockWaitEntry>, Option<DelayedNotifyAllFuture>)> {
+        let mut result = None;
+
+        // We don't want other threads insert any more entries between finding the
```
sticnarf, line 292: I don't have comment on `remove_if_mut` here. But I'm curious about when there could be concurrent insert here. The latch should prevent concurrent locking requests?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+
+/// Represents an `AcquirePessimisticLock` request that's waiting for a lock,
+/// and contains the request's parameters.
+pub struct LockWaitEntry {
+    pub key: Key,
+    pub lock_hash: u64,
+    // TODO: Use term to filter out stale entries in the queue.
+    // pub term: Option<NonZeroU64>,
+    pub parameters: PessimisticLockParameters,
+    pub lock_wait_token: LockWaitToken,
```
sticnarf, line 97: When will `lock_wait_token` in `LockWaitEntry` be used? I find it also in `LockWaitContextSharedState`, and find `lock_wait_token` here never used...

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+    }
+
+    async fn async_delayed_notify_all(
+        self,
+        key: Key,
+        start_time: Instant,
+        delay_ms: Arc<AtomicU64>,
+        notify_id: u64,
+    ) -> Option<Box<LockWaitEntry>> {
+        // tokio::task::yield_now().await;
```
sticnarf, line 381: Debug code?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+            .for_each(|h| expect_write_conflict(&h.wait_for_result().unwrap_err().0, 5, 6));
+        // 14 is not woken up.
+        assert!(
+            handles2[0]
+                .wait_for_result_timeout(Duration::from_millis(100))
+                .is_none()
+        );
+
+        queues.mock_lock_wait(b"k1", 9, 5, false);
+        // 9 will be woken up and 14 to 16 will be scheduled for delayed wake up.
```
TonsnakeLin, line 863: You've defined `15` as a resumable lock-waiting request, so is the description `14 to 16 will be scheduled for delayed wake up` right ?

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+            }
+
+            // Clear the state which indicates the scheduled `delayed_notify_all` has
+            // finished.
+            v.delayed_notify_all_state = None;
+
+            conflicting_start_ts = v.last_conflict_start_ts;
+            conflicting_commit_ts = v.last_conflict_commit_ts;
+
+            let legacy_wake_up_index = v.legacy_wake_up_cnt;
```
sticnarf, line 438: `index` sounds more appropriate. So, why not also use this name in `KeyLockWaitState`?

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: It should not be possible now. I covered the case for completeness.

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+
+        // Don't need to create new future if there already exists one for the key.
+        let entry = queues.must_pop_with_no_delayed_notify(b"k1", 9, 10);
+        entry.check_key(b"k1").check_start_ts(17);
+        queues.must_have_next_entry(b"k1", 18);
+
+        queues.mock_lock_wait(b"k1", 19, 5, false);
+        assert!(delayed_wake_up_future.await.is_none());
+        // 18 will be cancelled with ts of the latest wake-up event.
+        expect_write_conflict(&handle18.wait_for_result().unwrap_err().0, 9, 10);
```
TonsnakeLin, line 933: I think `18` is never cancelled ?

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: Yes, for `pop_for_waking_up` function in current implementation.  I think doing this makes it less coupling with Scheduler's logic that makes use of the lock waiting queue.

By the way I think there might be some more possible usages of it in the future. For example, when a new pessimistic lock request finds that the key is locked by checking the lock waiting queue, it can be put to the queue without entering scheduler and reading the data. In this case, an entry might be inserted without holding the latch. (This optimization is currently just an incomplete idea).

By the way*2, `delayed_notify_all` also uses `remove_if_mut` and that function won't be executed while holding latch.

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: The field in `LockWaitEntry` will be used in the future for updating lock waiting info to the `WaiterManager`, to provide up-to-time diagnostic info and redo deadlock detection when the current-holding-lock is changed.
The field in `LockWaitContextSharedState` will be used in function `finish_request` to remove the waiter from the `WaiterManager`, but commented out since `WaiterManager` is not yet adapted.
The `LockWaitToken` type is used to uniquely identify a lock-waiting request in `WaiterManager`, which is not changed in this PR. The related changes are in #13447 .

src/storage/lock_manager/lock_waiting_queue.rs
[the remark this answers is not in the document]
MyonKeminta: Sorry, I explained it wrong. The id is not attached to the `KeyLockWaitState`, but will be allocated when scheduling a new `delayed_notify_all`, and is stored in the `delayed_notify_all_state` field in the `KeyLockWaitState`. It may change when a `delayed_notify_all` is done and a new one is scheduled. But it performs the same functionality as I described above.

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: 18 will be cancelled when awaiting for the `delayed_wake_up_future`. At that time, the queue contains [18, 19] and 18 has an older legacy_wake_up_cnt, therefore it will be cancelled by `delayed_notify_all`.

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -0,0 +1,950 @@
...
+    /// becomes:
+    ///
+    /// ```text
+    /// legacy_wake_up_cnt: 1, queue: [(30, 0), (40, 0), (50, 1), (60, 1)]
+    /// ```
+    ///
+    /// Then `wake_up_delay_duration` is elapsed since the previous waking up.
+    /// Here, we expect that the lock wait entries 30 and 40 can be woken
+    /// up, since they exists when the previous waking up occurs. But we
+    /// don't want to wake up later-arrived entries (50 and 60) since it
```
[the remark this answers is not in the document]
MyonKeminta: Since 50 and 60 is pushed to the queue, they must have seen some lock on the key. They are added *after* the previous waking up, therefore there must be someone else who acquired the lock after it, otherwise one of 50 and 60 should have successfully acquired the lock instead of started waiting.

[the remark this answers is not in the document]
ekexium: Gotcha

TonsnakeLin:
I always feel that there are several comments in the test case that do not correspond to the code, but it doesn't matter, it will be decided by the actual execution in the end, and there are no other problems.
LGTM.

TonsnakeLin (approved):

cfzjywxk:
@MyonKeminta 
Are there any benchbot regression test results or local test results to verify the performance impact?

sticnarf:
> @MyonKeminta Are there any benchbot test results or local test results to verify the performance impact like improvements or regressions for this pr?

I think this PR itself should have no performance impact because the added functions are not used yet. I see https://github.com/tikv/tikv/pull/13447#issuecomment-1250682427 has some benchmark result after using these new functions.

cfzjywxk (approved):
LGTM

sticnarf (approved):

PR #13487, opened 2022-09-19
*: Refine interfaces for TiFlash when using TiKV as a component

### LINKED ISSUE #12849: Support TiKV using different KvEngines

CalvinNeo:

Issue Number: Close #12849

What's Changed:

```commit-message
Export necessary functions for TiFlash when using TiKV as a component
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


breezewish (approved):

src/storage/config_manager.rs
```diff
@@ -19,20 +19,26 @@ use crate::{
     storage::{lock_manager::LockManager, txn::flow_controller::FlowController, TxnScheduler},
 };
 
-pub struct StorageConfigManger<E: Engine, L: LockManager> {
-    tablet_factory: Arc<dyn TabletFactory<E::Local> + Send + Sync>,
+pub struct StorageConfigManger<E: Engine, EL: engine_traits::KvEngine, L: LockManager> {
```
BusyJay, line 22: Why?

CalvinNeo: There are two arguments in this structure:

1. `tablet_factory` accepts an instance of `KvEngineFactory` which impls `TabletFactory<engine_rocks::Rocks>,` so `E::Local` should be `engine_rocks::RocksEngine`.
1. `scheduler` is in `Storage` which has type parameter `EK: KvEngine` which is `engine_tiflash::RocksEngine` in our case. So `E::Local` should be `engine_tiflash::RocksEngine`.

So there is a conflict here, so I want to decouple these constraints.

src/storage/config_manager.rs
```diff
@@ -19,20 +19,26 @@ use crate::{
...
+    for StorageConfigManger<E, EL, L>
+{
+}
+unsafe impl<E: Engine, EL: engine_traits::KvEngine, L: LockManager> Sync
+    for StorageConfigManger<E, EL, L>
+{
+}
 
-impl<E: Engine, L: LockManager> StorageConfigManger<E, L> {
+impl<E: Engine, EL: engine_traits::KvEngine, L: LockManager> StorageConfigManger<E, EL, L> {
```
5kbpers, line 39: Better to import `KvEngine` instead of using full path.

BusyJay (approved):

PR #13491, opened 2022-09-19
raftstore: support switch witness

ethercflow:
Issue Number: ref tikv/tikv#12876

Issue Number: Close #xxx

What's Changed:

```
support switch witness
```

### Related changes

- No

### Check List 

Tests 

- integrations

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
support for converting witness to non-witness
```


ethercflow:
This PR is based on https://github.com/tikv/tikv/pull/12972.  Please help me review whether the execution process of the code is correct, whether there are any omissions, etc. Thank you. @Connor1996 @BusyJay

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -909,6 +909,9 @@ where
     /// in same Ready should be applied failed.
     pending_remove: bool,
 
+    /// Indicates whether the peer is unavailable. See more in `Peer`.
+    unavailable: Arc<AtomicBool>,
```
Connor1996, line 913: return the state by ExecRes instead of a shared value

ethercflow: Here I have a question, is `return the state by ExecRes` safe enough? For example, if a learner read occurs. I set `unavailable` to `true` before set `exist_peer's witness` to `false`. So it can Forbid reads.

Connor1996: The PeerFsm is held by one raftstore thread at one time. So it changes `unavailable = true` and `is_witness=false` atomically. Reads are forbidden both before and after.

ethercflow: Got it, thanks!

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -4913,6 +4955,12 @@ where
         let leader_id = self.fsm.peer.leader_id();
         let request = msg.get_requests();
 
+        // Forbid reads and writes when it's just became non-witness but not done apply
+        // snapshot.
+        if self.fsm.peer.wait_data {
```
Connor1996, line 4960: Shouldn't do it here. You should keep it from polling raft group like other peers applying snapshot. See `check_snap_status`

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5314,6 +5362,20 @@ where
...
 
+    fn on_request_snapshot_tick(&mut self) {
+        fail_point!("ignore request snapshot", |_| {});
+        assert!(!self.fsm.peer.is_leader());
+        if self
+            .fsm
+            .peer
+            .raft_group
+            .request_snapshot(self.fsm.peer.get_store().first_index())
+            .is_err()
```
Connor1996, line 5373: If it's the first time to call it, it should be not err. Then how do we keep ticking the tick?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2111,6 +2118,8 @@ where
 
         let state = if self.pending_remove {
             PeerState::Tombstone
+        } else if self.wait_data {
+            PeerState::Unavailable
```
Connor1996, line 2122: Does it reset to `Normal` after applying the snapshot?

components/raftstore/src/store/fsm/peer.rs
Connor1996: I thought it wrong, forget about it.

Connor1996: Please move it to L5010

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1029,15 +1033,18 @@ where
...
-            if expect_index != entry.get_index() {
-                panic!(
-                    "{} expect index {}, but got {}, ctx {}",
-                    self.tag,
-                    expect_index,
-                    entry.get_index(),
-                    apply_ctx.tag,
-                );
+            let peer = util::find_peer_by_id(&self.region, self.id).unwrap();
+            if !peer.is_witness && !self.wait_data {
```
Connor1996, line 1037: we should stop applying following commands when `wait_data` is set, until the snapshot is applied

Connor1996: Should also consider `ready_to_handle_pending_snap`

components/raftstore/src/store/fsm/peer.rs
ethercflow: Changed no longer judges its return value, but registers a tick again (the current interval setting is 1min), the next time it is triggered, it will first check wait_data, if it is still true, repeat the above process, otherwise return directly.

components/raftstore/src/store/fsm/apply.rs
ethercflow: Yes, the apply snap worker will change this state.

components/raftstore/src/store/fsm/apply.rs
ethercflow: > we should stop applying following commands when `wait_data` is set, until the snapshot is applied

It is currently detected that `wait_data` is `true`, return `yeild` until the apply snapshot is completed.

> Should also consider ready_to_handle_pending_snap

Sure, if `wait_data` is detected, the index judgment condition is ignored, as apply fsm has been in yield state.

tests/integrations/raftstore/test_witness.rs
```diff
@@ -5,30 +5,21 @@ use std::{iter::FromIterator, sync::Arc, time::Duration};
 use futures::executor::block_on;
 use kvproto::{metapb, raft_cmdpb::ChangePeerRequest, raft_serverpb::PeerState};
 use pd_client::PdClient;
+// use protobuf::Message;
 use raft::eraftpb::ConfChangeType;
 use test_raftstore::*;
 use tikv_util::store::find_peer;
 
 fn become_witness(cluster: &Cluster<ServerCluster>, region_id: u64, peer: &mut metapb::Peer) {
```
Connor1996, line 13: we can remove it now

components/raftstore/src/store/worker/read.rs
```diff
@@ -418,6 +418,7 @@ pub struct ReadDelegate {
     pub txn_ext: Arc<TxnExt>,
     pub read_progress: Arc<RegionReadProgress>,
     pub pending_remove: bool,
+    pub wait_data: bool,
```
Connor1996, line 421: Add a comment

components/raftstore/src/store/worker/metrics.rs
```diff
@@ -59,6 +59,7 @@ make_static_metric! {
         witness,
         flashback_not_prepared,
         flashback_in_progress,
+        non_witness,
```
Connor1996, line 62: how about `wait_data`?

components/raftstore/src/store/worker/read.rs
```diff
@@ -785,6 +789,12 @@ where
             return Err(e);
         }
 
+        // Check non-witness hasn't finish applying snapshot yet.
+        if delegate.wait_data {
```
Connor1996, line 793: let's move it near `check witness`

components/raftstore/src/store/util.rs
```diff
@@ -226,7 +226,7 @@ pub fn admin_cmd_epoch_lookup(admin_cmp_type: AdminCmdType) -> AdminCmdEpochStat
         AdminCmdType::PrepareFlashback | AdminCmdType::FinishFlashback => {
             AdminCmdEpochState::new(true, true, false, false)
         }
-        AdminCmdType::BatchSwitchWitness => unimplemented!(),
+        AdminCmdType::BatchSwitchWitness => AdminCmdEpochState::new(false, false, false, false),
```
Connor1996, line 229: switch witness should change the epoch

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5056,6 +5089,29 @@ where
...
+                .get_switch_witnesses()
+                .iter()
+                .any(|s| s.get_peer_id() == self.fsm.peer.peer.get_id() && s.get_is_witness())
+        {
+            self.ctx.raft_metrics.invalid_proposal.witness.inc();
+            // TODO: use a dedicated error type
+            return Err(Error::RecoveryInProgress(self.region_id()));
+        }
+
+        // Forbid requets when it becomes to non-witness but not finish applying
```
Connor1996, line 5108: requets -> requests

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5056,6 +5089,29 @@ where
...
+        // snapshot.
+        if self.fsm.peer.wait_data {
+            self.ctx.raft_metrics.invalid_proposal.non_witness.inc();
```
Connor1996, line 5111: How about `wait_data`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3053,6 +3090,85 @@ where
...
+            .get_switch_witnesses()
+            .to_vec();
+
+        info!(
+            "exec BatchSwitchWitness";
+            "region_id" => self.region_id(),
+            "epoch" => ?self.region.get_region_epoch(),
+        );
+
+        let region = self.apply_batch_switch_witness(switches.as_slice())?;
```
Connor1996, line 3110: It's simple enough, no need to extract a function

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3809,8 +3932,9 @@ where
         if self.delegate.pending_remove || self.delegate.stopped {
             return;
         }
-        if self.delegate.peer.is_witness {
-            // witness shouldn't generate snapshot.
+        if self.delegate.peer.is_witness || self.delegate.wait_data {
+            // witness or non-witess hasn't finish applying snapshot shouldn't generate
```
Connor1996, line 3936: witess -> witness

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3053,6 +3090,85 @@ where
...
+
+        if let Err(e) = write_peer_state(ctx.kv_wb_mut(), &region, state, None) {
+            panic!("{} failed to update region state: {:?}", self.tag, e);
+        }
+
+        let resp = AdminResponse::default();
+        Ok((
+            resp,
+            ApplyResult::Res(ExecResult::BatchSwitchWitness(SwitchWitness {
+                succeeded: true,
```
Connor1996, line 3128: What's this used for?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1166,6 +1181,14 @@ where
 
         if !data.is_empty() {
             if !self.peer.is_witness || !can_witness_skip(entry) {
+                // If it's witness before, but a command changes it to non-witness,
+                // and starts applying all following commands. It can cause problem
+                // if the command depends on previous side effect. So stop applying
+                // following commands when `wait_data` is set, until the snapshot is
+                // applied.
+                if self.wait_data {
```
Connor1996, line 1189: should check it in `resume_pending`

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1232,6 +1255,14 @@ where
         fail_point!("yield_apply_conf_change_3", self.id() == 3, |_| {
             ApplyResult::Yield
         });
+        // If it's witness before, but a command changes it to non-witness,
+        // and starts applying all following commands. It can case problem
+        // if the command depends on previous side effect. So stop applying
+        // following commands when `wait_data` is set, until the snapshot is
+        // applied.
+        if self.wait_data {
```
Connor1996, line 1263: Can we set yield_state directly after `SwitchWitness` is applied?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3156,6 +3272,7 @@ pub struct Apply<C> {
     pub entries_size: usize,
     pub cbs: Vec<Proposal<C>>,
     pub bucket_meta: Option<Arc<BucketMeta>>,
+    pub wait_data: bool,
```
Connor1996, line 3275: Notify the change by a separate msg, see `Msg::LogsUpToDate` for reference

components/raftstore/src/store/fsm/apply.rs
ethercflow: I thought wrong before, it is used to mark apply failure, read the error handling part, no need

components/raftstore/src/store/fsm/apply.rs
Connor1996: Why? The `ApplyFsm` is never manipulated by multiple threads at the same time.

tests/integrations/raftstore/test_witness.rs
```diff
@@ -522,7 +514,7 @@ fn test_witness_leader_down() {
     // forbid reads
     let get = new_get_cmd(b"k1");
     must_get_error_recovery_in_progress(&mut cluster, &region, get);
-    // forbid read index
+    //  forbid read index
```
Connor1996, line 517:
```suggestion
    // forbid read index
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1109,7 +1109,22 @@ where
...
             match res {
                 ApplyResult::None => {}
-                ApplyResult::Res(res) => results.push_back(res),
+                ApplyResult::Res(res) => {
+                    results.push_back(res);
+                    if self.wait_data {
+                        apply_ctx.committed_count -= committed_entries_drainer.len();
+                        let mut pending_entries =
+                            Vec::with_capacity(committed_entries_drainer.len());
+                        pending_entries.extend(committed_entries_drainer);
```
Connor1996, line 1117: The snapshot would be older than the current commit index, only applying these pending entries is not enough. How about skipping all entries when `wait_data = true` until receiving a message to notify the snapshot is applied. The msg should include entries ranging from (snapshot_index, commit_index)/

Connor1996: And please add a test case to cover "The snapshot would be older than the current commit index"

components/raftstore/src/store/fsm/apply.rs
ethercflow: I thought to share `waita_data` bewteen store thread (`PeerFsm`) and apply thread (`ApplyFsm`).

Because after the `wait_data` of `ApplyFsm` becomes `true`, the current PR implementation is always pending, so it cannot read the message will be sent by the `PeerFsm` thread after the apply snapshot is successful. 

The comment you mentioned below seems to have solved the problem, I'll think about it first.

components/raftstore/src/store/fsm/apply.rs
Connor1996: If the leader always send the latest snapshot(snapshot after switch witness), I think "The snapshot would be older than the current commit index" won't happen.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -623,6 +631,9 @@ where
         delegate: &mut ApplyDelegate<EK>,
         results: VecDeque<ExecResult<EK::Snapshot>>,
     ) {
+        if delegate.wait_data && results.is_empty() {
```
Connor1996, line 634: is it necessary?

components/raftstore/src/store/peer.rs
```diff
@@ -2550,7 +2567,43 @@ where
...
+                                    self.region_id,
+                                    self.term(),
+                                    self.last_applying_idx + 1,
+                                    commit_term,
+                                    entries,
+                                    vec![],
+                                    self.region_buckets.as_ref().map(|b| b.meta.clone()),
+                                );
+                                ctx.apply_router
+                                    .schedule_task(self.region_id, ApplyTask::apply_gap(apply));
```
Connor1996, line 2593: `ApplyGap` can be merged with `Recover`

components/raftstore/src/store/peer_storage.rs
```diff
@@ -457,6 +457,13 @@ where
         }
 
         if find_peer_by_id(&self.region, to).map_or(false, |p| p.is_witness) {
+            if self.applied_index() < request_index {
```
Connor1996, line 460: check `last_applying_idx` instead of `applied_index`, gen snapshot request goes through apply worker first, so it can make sure getting the latest snapshot.

ethercflow: @Connor1996  I don't get your point: as `self.last_applying_idx = last_entry.get_index();`  `request_index` is equal to `switch witness idx` which must be <= `self.last_applying_idx`. So how to use it to check?

BTW, should non-witness set `request_index` to its `last_applying_idx` instead of `switch witness idx`? as we dropped logs between [`switch witness idx`, `last_applying_idx`], if we set `request_index` to its `last_applying_idx`, we don't need `ApplyGap` anymore?

Connor1996: `request_index is equal to switch witness idx which must be <= self.last_applying_idx` is not true, leader may get stuck somehow, so it would be slower than the new non-witness.

Connor1996: ` should non-witness set request_index to its last_applying_idx instead of switch witness idx` no need, as `self.last_applying_idx` is always changing, `ApplyGap` is still needed

components/raftstore/src/store/fsm/apply.rs
ethercflow: No need anymore, removed

components/raftstore/src/store/peer.rs
ethercflow: Merged, thanks!

components/raftstore/src/store/peer.rs
```diff
@@ -2550,7 +2567,40 @@ where
                 // Update apply index to `last_applying_idx`
                 self.read_progress
                     .update_applied(self.last_applying_idx, &ctx.coprocessor_host);
-                self.notify_leader_the_peer_is_available(ctx);
+                if self.wait_data {
```
tonyxuqqi, line 2570: Question: Why do we need to call apply_gap after the snapshot is applied for wait_data == true? I mean why not catch up the gaps just like normal cases (apply snapshot and then apply new logs since the last applied index)

components/raftstore/src/store/peer.rs
```diff
@@ -2550,7 +2567,40 @@ where
...
+                    self.notify_leader_the_peer_is_available(ctx);
```
tonyxuqqi, line 2571: what will happen if notify_leader_the_peer_is_available failed?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1094,7 +1106,13 @@ where
 
             match res {
                 ApplyResult::None => {}
-                ApplyResult::Res(res) => results.push_back(res),
+                ApplyResult::Res(res) => {
+                    results.push_back(res);
+                    if self.wait_data {
+                        apply_ctx.committed_count -= committed_entries_drainer.len();
+                        break;
+                    }
```
tonyxuqqi, line 1114: please add some comments on this block.
Question: why if it's wait_data, it breaks the loop. 
One of my confusion is when witness is converted to non-witness, does it have any difference with the scenario a new peer is created and needs to catch up the data by snapshot.

components/raftstore/src/store/peer.rs
ethercflow: After the leader has applied the non-witness switch, it will register a tick. If it has not received the non-witness push information for a long time, it will take the initiative to obtain the information.

components/raftstore/src/store/peer.rs
ethercflow: For normal cases, `ready.committed_messages` have all been processed before `apply snapshot`. But for this case, in order to avoid the side effects of continuing to apply other cmds after applying the `non-witness switch`, the logs behind `non-witness switch` will be skipped directly. So we need to make sure the skipped logs will be all applied.

components/raftstore/src/store/fsm/apply.rs
ethercflow: > please add some comments on this block.

Sure, will add after all your questions have been confirmed.

>  Question: why if it's wait_data, it breaks the loop.

If it's witness before, but a command changes it to non-witness, and starts applying all following command. It can cause problem if the command depends on previous side effect. 

> One of my confusion is when witness is converted to non-witness, does it have any difference with the scenario a new peer is created and needs to catch up the data by snapshot.

The difference is that the new peer is `uninitialized` before `apply snapshot`, and applied `non-witness switch` makes `apply thread` thinks it is already `non-witness`, but it is not. So it should not continue to apply subsequent logs until the `apply snapshot` is complete.

components/raftstore/src/store/peer.rs
```diff
@@ -2550,7 +2567,40 @@ where
...
+                    let mut apply = None;
+                    if self.last_applying_idx < self.get_store().commit_index() {
+                        match self.get_store().entries(
+                            self.last_applying_idx + 1,
+                            self.get_store().commit_index() + 1,
+                            NO_LIMIT,
+                            GetEntriesContext::empty(true),
```
Connor1996, line 2578: should be false

components/raftstore/src/store/peer.rs
```diff
@@ -894,6 +894,8 @@ where
     /// the same time period.
     pub wait_data: bool,
 
+    pub request_index: u64,
```
Connor1996, line 897: add comment

components/raftstore/src/store/worker/read.rs
```diff
@@ -418,6 +418,8 @@ pub struct ReadDelegate {
     pub txn_ext: Arc<TxnExt>,
     pub read_progress: Arc<RegionReadProgress>,
     pub pending_remove: bool,
+    /// Indicates whether the peer is waiting data. See more in `Peer`.
```
Connor1996, line 421: should update it in `fn update`

Connor1996: and add a case to cover it

components/raftstore/src/store/fsm/store.rs
```diff
@@ -561,6 +561,9 @@ where
             self.cfg.check_long_uncommitted_interval.0;
         self.tick_batch[PeerTick::CheckPeersAvailability as usize].wait_duration =
             self.cfg.check_peers_availability_interval.0;
+        // TODO: Is it reasonable to use pd_heartbeat_tick_interval?
```
Connor1996, line 564: let's use a dedicated config

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -6258,6 +6341,42 @@ where
...
+                &mut self.fsm.peer,
+                RegionChangeReason::SwitchWitness,
+            );
+        }
+        for s in sw.switches {
+            let (peer_id, is_witness) = (s.get_peer_id(), s.get_is_witness());
+            if self.fsm.peer_id() == peer_id {
+                if is_witness && !self.fsm.peer.is_leader() {
+                    let _ = self.fsm.peer.get_store().clear_data();
+                    self.fsm.peer.raft_group.set_priority(-1);
```
Connor1996, line 6359: when will it set back to normal?

Connor1996: Should do that for snapshot as well, please check commentted code in `on_persist_snapshot`

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3196,6 +3220,80 @@ where
...
+    ) -> Result<(AdminResponse, ApplyResult<EK::Snapshot>)> {
+        assert!(request.has_switch_witnesses());
+        let switches = request
+            .get_switch_witnesses()
+            .get_switch_witnesses()
+            .to_vec();
+
+        info!(
+            "exec BatchSwitchWitness";
+            "region_id" => self.region_id(),
```
BusyJay, line 3236: Should include peer id.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3196,6 +3220,80 @@ where
...
+
+        let mut region = self.region.clone();
+        for s in switches.as_slice() {
+            PEER_ADMIN_CMD_COUNTER.batch_switch_witness.all.inc();
+            let (peer_id, is_witness) = (s.get_peer_id(), s.get_is_witness());
+            for p in region.mut_peers().iter_mut() {
+                if p.id == peer_id {
+                    if p.is_witness == is_witness {
+                        return Err(box_err!(
+                            "can't duplicated switch peer {:?} on region {:?}",
```
BusyJay, line 3248:
```suggestion
                            "switch peer {:?} on region {:?} is no-op",
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3196,6 +3220,80 @@ where
...
+                            find_peer_by_id(&self.region, peer_id),
```
BusyJay, line 3249: You mean `p`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3196,6 +3220,80 @@ where
...
+                            region
+                        ));
+                    }
+                    p.is_witness = is_witness;
+                    PEER_ADMIN_CMD_COUNTER.batch_switch_witness.success.inc();
```
BusyJay, line 3254: You can inc the counter outside the loop.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3196,6 +3220,80 @@ where
...
+            "exec BatchSwitchWitness";
+            "region_id" => self.region_id(),
+            "epoch" => ?self.region.get_region_epoch(),
+        );
+
+        let mut region = self.region.clone();
+        for s in switches.as_slice() {
+            PEER_ADMIN_CMD_COUNTER.batch_switch_witness.all.inc();
+            let (peer_id, is_witness) = (s.get_peer_id(), s.get_is_witness());
+            for p in region.mut_peers().iter_mut() {
```
BusyJay, line 3244: What if the peer is not found in `region`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3589,6 +3687,11 @@ where
     #[cfg(any(test, feature = "testexport"))]
     #[allow(clippy::type_complexity)]
     Validate(u64, Box<dyn FnOnce(*const u8) + Send>),
+    ApplyGap {
```
BusyJay, line 3690: What is `ApplyGap`?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2170,6 +2178,18 @@ where
             return;
         }
 
+        // Keep ticking if there are non-witness peers waiting for snapshot.
+        if !self.fsm.peer.wait_data_peers.is_empty() {
```
BusyJay, line 2182: These constraint should be put into `check_after_tick`.

components/raftstore/src/store/fsm/apply.rs
ethercflow: Eg:

T0 Loop 1: The witness handles `ready.committed_entries()`, gets `entries [6, 20]`, and the index of `switch to non-witness` is `9`. After the `witness` peer applies `9`, it will skip apply `[10, 20]`.

T1 The leader will send a snapshot, its index may between `[10, 20]`, eg `12`.

T2 Loop 2: The applied snapshot `non-witness` handle next  `ready.committed_entries()`, gets `entries [21, 40]`.

So there is a log gap `[13, 20]` without `apply`. The `ApplyGap` is for `apply` this `gap`.

components/raftstore/src/store/peer.rs
```diff
@@ -2587,22 +2641,19 @@ where
...
-                msg.set_type(ExtraMessageType::MsgAvailabilityResponse);
-                msg.wait_data = false;
-                self.send_extra_message(msg, &mut ctx.trans, &leader);
-                info!(
-                    "notify leader the leader is available";
-                    "region id" => self.region().get_id(),
-                    "peer id" => self.peer.id
-                );
-            }
+        fail_point!("ignore notify leader the peer is available", |_| {});
```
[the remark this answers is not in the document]
ethercflow: > What if it was not witness at all?

At present, the witness check is done before calling this function.

before is 
``` rust
fn notify_leader() {
   if self.wait_data { // witness check
    ... 
   }
}
```

current is 

``` rust
fn notify_leader() {
    ... 
}

if self.wait_data { // witness check
    self.notify_leader();
    ...  // other logics which also need witness check
}
```

components/raftstore/src/store/fsm/apply.rs
ethercflow: Oh, right. no need to find.

components/raftstore/src/store/fsm/apply.rs
BusyJay: I think it will panic at T1 as shown in:

https://github.com/tikv/raft-rs/blob/5ce52b480065ff31ecef16b9b77c7c3b7c57c8c7/src/raw_node.rs#L520-L529

components/raftstore/src/store/fsm/apply.rs
ethercflow: Added a check to return an error if the peer does not exist.

components/raftstore/src/store/fsm/apply.rs
ethercflow: I see,  so I may need to make non-witness `request_index` to T0's `the lastest committed_index`,  and forbid persistence of new raft log entries (maybe they were synchronized before the leader received the snapshot request), so even leader updated and sync `new committed index` to follower, the follower's raft `persisted` still equal to  `the prev_lastest committed_index` to make https://github.com/tikv/raft-rs/blob/5ce52b480065ff31ecef16b9b77c7c3b7c57c8c7/src/raw_node.rs#L520-L529 it pass?

ethercflow: > I see, so I may need to make non-witness `request_index` to `the lastest committed_index`, and forbid persistence of new raft log entries (maybe they were synchronized before the leader received the snapshot request), so even leader updated and sync `new committed index` to non-witness, the non-witness's raft `persisted` still equal to `the prev_lastest committed_index` to make https://github.com/tikv/raft-rs/blob/5ce52b480065ff31ecef16b9b77c7c3b7c57c8c7/src/raw_node.rs#L520-L529 pass?

I'll think more about it.

BusyJay: Maybe `last_applying_idx` should be decreased if a snapshot is received and accepted.

Connor1996: Make non-witness `request_index` to its committed_index, and reject raft append msg when in `wait_data`. And on leader side, because gen-snapshot tasks go through apply worker, the snapshot's index must be equal to or larger than non-witness's committed_index. Then `ApplyGap` is not needed and won't encounter the assertion.

BusyJay: > because gen-snapshot tasks go through apply worker, the snapshot's index must be equal to or larger than non-witness's committed_index

It's not true when paging is enabled for committed entries.

ethercflow: > > because gen-snapshot tasks go through apply worker, the snapshot's index must be equal to or larger than non-witness's committed_index
> 
> It's not true when paging is enabled for committed entries.

When gen snapshot, it will check snapshot's idx and `request_index`: https://github.com/tikv/tikv/blob/master/components/raftstore/src/store/peer_storage.rs#L407
So if we set `request_index` to non-witness `latest comit index` (after `wait_data` set to `true`, reject `MsgAppend`), it should be ok?

Connor1996: @BusyJay What's paging?

Connor1996: Can we just remove the assertion?

BusyJay: I think there are two problems:

1. Requesting latest committed logs are not enough because leader may just keep sending committed entries to witness and increase its `commit_since_idx`. Reject `MsgAppend` will make a raft group with one follower isolated unavailable immediately. The assert can help find potential flaws in `Storage` implementations. I would suggest to make the assert more strict that allows explicit requested snapshots. Note `commit_since_idx` is just one of the case, I think raft_log.persisted will also be affected and maintained carefully.
2. Skipping applying logs may make the follower unable to compaign when there are pending conf changes. So if the wait data non-witness has the most up to date logs when leader is isolated, the group will also be unavailable.

The simplest solution is to request the snapshot at last_index and then reject any logs afterwards. The term should be also attached to the snapshot requests, leader generates snapshot if `local_term > request_term || local_index >= request_index`. This solution will not solve the unavailability corner cases described above, but is safest.

In opinion, switch witness is no different from remove the node and then add a new node except it can still campaign if leader is isolated during the time. So it has a little higher availability then remove + add.

tonyxuqqi (approved):

Connor1996 (approved):
LGTM

PR #13495, opened 2022-09-20
raftstore-v2: add basic apply

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
This PR adds the implementation of apply for v2. In v2, we don't need to
batch writes across regions, so there is no need to use batch system.
Instead, future pool is used to simplify implementations.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -147,19 +206,160 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
-            self.raft_group().raft.raft_log.committed,
-            self.raft_group().raft.raft_log.persisted,
-        );
-        let commit_term = self.raft_group().raft.raft_log.term(commit_index).unwrap();
-        // TODO: schedule apply task
+            entry_and_proposals = committed_entries.into_iter().map(|e| (e, vec![])).collect();
+        }
+        // Unlike v1, v2 doesn't need to persist commit index and commit term. The
+        // point of persist commit index/term of raft apply state is to recover commit
+        // index when the writes to raft engine is lost but writes to kv engine is
```
tonyxuqqi, line 216: Question: In v1, can entries in ready.committed_entries() be unpersisted?  Otherwise, what does it mean "writes to raft engine is lost?"

tonyxuqqi: In raft_log.rs next_entries_since, it seems it should return all committed and persisted entries.

components/raftstore-v2/src/fsm/apply.rs
```diff
@@ -1,74 +1,101 @@
...
+                // TODO: yield after some time.
+
+                // Perhaps spin sometime?
+                match self.receiver.try_recv() {
+                    Ok(t) => task = t,
+                    Err(TryRecvError::Empty) => break,
+                    Err(TryRecvError::Disconnected) => return,
+                }
+            }
+            self.apply.flush();
```
tonyxuqqi, line 98: should it call flush in case there're too many entries?

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -147,19 +206,160 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+                Ok(new_response(req.get_header()))
+            }
+        }
+    }
+
+    #[inline]
+    pub fn flush(&mut self) {
+        if let Some(wb) = self.write_batch_mut() && !wb.is_empty() {
+            let mut write_opt = WriteOptions::default();
+            write_opt.set_disable_wal(true);
```
tonyxuqqi, line 339: We don't have remove wal for multi-rocks yet, right?

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -147,19 +206,160 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+            .advance_apply_to(apply_res.applied_index);
+        let is_leader = self.is_leader();
+        let entry_storage = self.entry_storage_mut();
+        entry_storage
+            .apply_state_mut()
+            .set_applied_index(apply_res.applied_index);
+        entry_storage.set_applied_term(apply_res.applied_term);
+        if !is_leader {
+            entry_storage.compact_entry_cache(apply_res.applied_index + 1);
+            // TODO: handle read.
```
tonyxuqqi, line 245: Question: why need to handle read here?  Do you mean the empty entry in read index when transferring leader?

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -83,10 +90,67 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
             let (data, chs) = encoder.encode();
             let res = self.propose(ctx, data);
             self.post_propose_write(ctx, res, chs);
         }
     }
 }
+
+impl<EK: KvEngine, R> Apply<EK, R> {
+    #[inline]
+    pub fn apply_put(&mut self, cf: &str, key: &[u8], value: &[u8]) -> Result<()> {
```
tonyxuqqi, line 103: It seems missing relevant metrics such as lock_cf_written_bytes, size_diff_hint, etc.

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -1,23 +1,108 @@
...
+    pub fn reset_state_changed(&mut self) -> bool {
+        std::mem::take(&mut self.state_changed)
+    }
+
+    /// Publish the tablet so that it can be used by read worker.
+    ///
+    /// Note, during split/merge, lease is expired explicitly and read is
+    /// forbidden. So publishing it immediately is OK.
+    #[inline]
+    pub fn publish_tablet(&mut self, tablet: EK) {
```
tonyxuqqi, line 104: It seems not called in the code. When should it be called?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -96,10 +96,10 @@ use crate::{
     Error, Result,
 };
 
-const DEFAULT_APPLY_WB_SIZE: usize = 4 * 1024;
-const APPLY_WB_SHRINK_SIZE: usize = 1024 * 1024;
-const SHRINK_PENDING_CMD_QUEUE_CAP: usize = 64;
-const MAX_APPLY_BATCH_SIZE: usize = 64 * 1024 * 1024;
+pub const DEFAULT_APPLY_WB_SIZE: usize = 4 * 1024;
```
tonyxuqqi, line 99: please add comment to indicate it's shared in v2.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: > can entries in ready.committed_entries() be unpersisted

Yes. As long as it's committed (persisted in quorum), it will be applied without waiting for being persisted.

components/raftstore-v2/src/fsm/apply.rs
BusyJay: Yes, it should. It's just not implemented.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: Yes, but it will. So implemented with the new design in mind.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: There may be reads are paused due to split/merge/new leaders.

components/raftstore-v2/src/operation/command/write/mod.rs
BusyJay: There may be too many TODOs. Metrics will be added all together in a separate PR.

components/raftstore-v2/src/raft/apply.rs
BusyJay: After split/merge.

components/raftstore-v2/src/operation/command/mod.rs
tonyxuqqi: OK. I would suggest to add a TODO here but still keep it false. Until remove wal is implemented for multi-rocks then change it to true.

components/raftstore-v2/src/fsm/apply.rs
tonyxuqqi: Then please add a TODO here.

tonyxuqqi: Also I think maybe we should flush not by fixed batch (entry) size, but by actual bytes size, it should be more accurate for time to spent.

components/raftstore-v2/src/operation/command/mod.rs
tonyxuqqi: > "After read index, it may wait for application to respond."
What's the exact respond need to be done here? Is v1's behavior same?

components/raftstore-v2/src/operation/command/mod.rs
tonyxuqqi: From the code, it seems even v1 will need the entries be persisted before apply.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: That will not satisfy the performance requirement. Unless correctness test is involved, we don't need to set it to false.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: The same. Just check if log has applied to the corresponding index, if it does, respond to read.

components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -83,12 +85,23 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
     pub fn on_query_debug_info(&self, ch: DebugInfoChannel) {
         let entry_storage = self.storage().entry_storage();
-        let meta = RegionMeta::new(
+        let mut meta = RegionMeta::new(
             self.storage().region_state(),
             entry_storage.apply_state(),
             GroupState::Ordered,
             self.raft_group().status(),
         );
+        // V2 doesn't persist commit term, so fill it on demand.
```
tabokie, line 94: I don't understand this comment.

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -83,10 +90,67 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+    }
+
+    #[inline]
+    pub fn apply_delete(&mut self, cf: &str, key: &[u8]) -> Result<()> {
+        util::check_key_in_region(key, self.region_state().get_region())?;
+        let res = if cf.is_empty() || cf == CF_DEFAULT {
+            // TODO: use write_vector
+            self.write_batch_or_default().delete(key)
+        } else {
+            self.write_batch_or_default().delete_cf(cf, key)
```
tabokie, line 131: Could `cf == CF_DEFAULT` use this path too?

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -1,23 +1,108 @@
...
+    tablet::CachedTablet,
+};
 
 /// Apply applies all the committed commands to kv db.
-pub struct Apply<EK: KvEngine> {
-    tablet: CachedTablet<EK>,
-    logger: Logger,
+pub struct Apply<EK: KvEngine, R> {
+    remote_tablet: CachedTablet<EK>,
+    tablet: EK,
```
tabokie, line 18: Why two fields? Isn't `remote_tablet.cached() == tablet`?

BusyJay: `remote_tablet` is only used for updating local tablet. It's not safe to query it directly as peer may change the tablet when applying a snapshot.

components/raftstore-v2/src/operation/query/mod.rs
BusyJay: In v2, `commit_index` and `commit_term` are always 0. So here needs to initialized the value for debug purpuse.

components/raftstore-v2/src/operation/command/write/mod.rs
BusyJay: Yes, but rocksdb has optimization on default cf, which can make the write batch smaller.

BusyJay:
PTAL

components/raftstore-v2/src/operation/query/mod.rs
tabokie:
```suggestion
        // V2 doesn't persist commit index and term, fill them with in-memory values.
```

components/raftstore-v2/src/fsm/apply.rs
```diff
@@ -1,74 +1,102 @@
...
-    pub fn new(fsm: &'a mut ApplyFsm<EK>, apply_ctx: &'a mut ApplyContext) -> Self {
-        Self { fsm, apply_ctx }
+impl<EK: KvEngine, R> ApplyFsm<EK, R> {
+    pub fn new(
+        region_state: RegionLocalState,
+        res_reporter: R,
+        remote_tablet: CachedTablet<EK>,
+        logger: Logger,
+    ) -> (ApplyScheduler, Self) {
+        let (tx, rx) = future::unbounded(WakePolicy::Immediately);
```
tabokie, line 65: Is the future pool perf regression fixed?

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -23,15 +45,33 @@ use tikv_util::{box_err, time::monotonic_raw_now};
...
 
 mod write;
 
 pub use write::{SimpleWriteDecoder, SimpleWriteEncoder};
 
+use self::write::SimpleWrite;
+
+#[derive(Debug)]
+pub struct CommittedEntries {
+    /// Entries need to be applied. Note some entries may not be included for
```
tabokie, line 61: I don't understand again... Are these entries delayed or? Why is it necessary to comment this detail (i.e. where could rely on the "wholeness" of CommittedEntries?)

components/raftstore-v2/src/fsm/apply.rs
BusyJay: Not yet, but will.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: Yes, those entries are delayed. There is an applying entries window.

> where could rely on the "wholeness" of CommittedEntries

It's wrong to assume the last index of `entry_and_proposals` is the latest commit index. raftstore-v2 doesn't need latest committed index during apply anymore. But better keep the comment in case we want committed index again.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -147,19 +206,160 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+                Err(e) => {
+                    if let Some(wb) = self.write_batch_mut() {
+                        wb.rollback_to_save_point().unwrap();
+                    }
+                    cmd_resp::new_error(e)
+                }
+            };
+            self.callbacks_mut().push((ch, resp));
+            // Flush may be triggerred in the middle, so always update the index and term.
+            self.set_apply_progress(e.index, e.term);
```
tabokie, line 270: Why the progress is updated even when `apply_entry` failed?

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -147,19 +206,160 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+    pub async fn apply_committed_entries(&mut self, ce: CommittedEntries) {
+        for (e, ch) in ce.entry_and_proposals {
+            if let Some(wb) = self.write_batch_mut() {
+                wb.set_save_point();
+            }
+            let resp = match self.apply_entry(&e).await {
+                Ok(resp) => resp,
+                Err(e) => {
+                    if let Some(wb) = self.write_batch_mut() {
+                        wb.rollback_to_save_point().unwrap();
```
tabokie, line 263: What if wb is none at L256, and is some at L262? There will be no save point to rollback.

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: Because it's applied. We only care about consistency instead of success.

tabokie (approved):

tonyxuqqi (approved):

PR #13516, opened 2022-09-21
util: add customize waker

BusyJay:
Issue Number: Close #xxx

What's Changed:

```commit-message
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


Connor1996 (commented):
Where does the performance improvement come from? I don't see anywhere using `Box`

BusyJay:
> Where does the performance improvement come from? I don't see anywhere using `Box`

The framegraph shows the difference are in `AtomicWaker::register`. It can be caused by both the frequency of waking up and the cost of `register` itself. For the former, Period policy wakes up the thread on a steady rate, but `TillReach` can wake up the waker one more time. For example, if there are 10 elements, and after 8 elements are sent, both policies will wake up once. But `TillReach` will keep calling wake after that. If waker is registered in race cases, it will also be called.

For the latter, box waker use `AtomicPtr`, which is only a swap operation during register. But `AtomicWaker` requires multiple atomic load and stores. Using `AtomicPtr` requires allocation, which may seem worse, but when using jemalloc, the cost of allocation is a lot smaller. Though unit tests show that `AtomicWaker::register` performs similar with (actually slightly better than) `AtomicPtr`, but bank benchmark shows that the regression goes away only when two changes are introduced.

sticnarf:
Is the issue in the PR description (#13516) mislinked?

BusyJay:
> Is the issue in the PR description (#13516) mislinked?

Yes and fixed.

Connor1996:
> > Where does the performance improvement come from? I don't see anywhere using `Box`
> 
> The framegraph shows the difference are in `AtomicWaker::register`. It can be caused by both the frequency of waking up and the cost of `register` itself. For the former, Period policy wakes up the thread on a steady rate, but `TillReach` can wake up the waker one more time. For example, if there are 10 elements, and after 8 elements are sent, both policies will wake up once. But `TillReach` will keep calling wake after that. If waker is registered in race cases, it will also be called.
> 
> For the latter, box waker use `AtomicPtr`, which is only a swap operation during register. But `AtomicWaker` requires multiple atomic load and stores. Using `AtomicPtr` requires allocation, which may seem worse, but when using jemalloc, the cost of allocation is a lot smaller. Though unit tests show that `AtomicWaker::register` performs similar with (actually slightly better than) `AtomicPtr`, but bank benchmark shows that the regression goes away only when two changes are introduced.

Where is the framegraph? Seems the issue is still mislinked.

sticnarf (approved):
LGTM.

Does the new channel implementation bring no regression now?

BusyJay:
Yes, and it actually performs slightly better in bank test.

| Hash | Test 0 | Test 1 | Test 2 |
|-|-|-|-|
|1e5773cc439e5035970c1481993bc494278dfd1f (This PR) | 74051 | 69281 | 66032|
|b219a823b994f68b5a54e7ef8a10d82b047b68ef (Revert future channel) | 67405 | 66515 | 64979|

Connor1996 (approved):
LGTM

PR #13518, opened 2022-09-22
raftstore: Do not print error log for "region not exist but not tombstone"

### LINKED ISSUE #13002: Do not print log in error level for "region not exist but not tombstone"
## Development Task

Sometimes when add-peer runs slowly occasionally after a region split, a message from leader may arrive before finishing adding peer. In this case, an error complaining about "region not exist but not tombstone" will be produced:

https://github.com/tikv/tikv/blob/5c941586f5163ca9dbb82a0d14e7f02b09732181/components/raftstore/src/store/fsm/store.rs#L1743-L1747

And then it will cause an error log here:

https://github.com/tikv/tikv/blob/5c941586f5163ca9dbb82a0d14e7f02b09732181/components/raftstore/src/store/fsm/store.rs#L698-L701

However, this case is usually considered normal, and the logs in error level looks like something went wrong and may confuse the user (especially considering that some user may want to put the error log to their alert system).

It would be better if we can avoid printing error log for this kind of error. Looks like we'd better avoid using `box_err` here and define a specific kind of error for it.

MyonKeminta:

Issue Number: Close #13002 

What's Changed:

```commit-message
Avoid printing the log about "region not exist but not tombstone" in error level since it may happen in normal cases and confuse users.
```

### Related changes

### Check List 

Tests 

\-

Side effects

\-

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore/src/store/fsm/store.rs
+++ b/components/raftstore/src/store/fsm/store.rs
@@ -45,7 +45,7 @@ use resource_metering::CollectorRegHandle;
 use sst_importer::SstImporter;
 use tikv_alloc::trace::TraceEvent;
 use tikv_util::{
-    box_err, box_try,
+    box_try,
     config::{Tracker, VersionTrack},
     debug, defer, error,
     future::poll_future_notify,
@@ -101,7 +101,7 @@ use crate::{
         Callback, CasualMessage, GlobalReplicationState, InspectedRaftMessage, MergeResultKind,
         PdTask, PeerMsg, PeerTick, RaftCommand, SignificantMsg, SnapManager, StoreMsg, StoreTick,
     },
-    Result,
+    Error, Result,
 };
 
 type Key = Vec<u8>;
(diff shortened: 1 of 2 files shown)


components/raftstore/src/store/fsm/store.rs
```diff
@@ -708,10 +708,20 @@ impl<'a, EK: KvEngine + 'static, ER: RaftEngine + 'static, T: Transport>
...
                     if let Err(e) = self.on_raft_message(msg) {
-                        error!(?e;
-                            "handle raft message failed";
-                            "store_id" => self.fsm.store.id,
-                        );
+                        if matches!(&e, Error::RegionNotRegistered { .. }) {
+                            // This may happen in normal cases when add-peer runs slowly
+                            // occasionally after a region split. Avoid printing error
+                            // log here, which may confuse users.
+                            warn!("handle raft message failed";
```
BusyJay, line 715: How about changing it to `info`?

MyonKeminta: ok

gengliqi (approved):

BusyJay (approved):

PR #13521, opened 2022-09-22
storage: Rollback all locks to prevent from losing the pessimistic lock in flashback

### LINKED ISSUE #13493: Rollback all locks to prevent from losing the pessimistic lock in flashback
## Feature Request

Ref #13303.

Currently, we will delete all locks unconditionally when performing the flashback. Though we will check the minimal resolved TS before a flashback multiple times (https://github.com/pingcap/tidb/pull/37607, https://github.com/tikv/tikv/pull/13412) to prevent breaking any transaction consistency, there is still a case that we should handle when it comes to a pessimistic transaction.

Assume that we have two clients {c1, c2} and a key k1:

1. c1 begins a pessimistic transaction with ts1.
2. c1 creates and writes the k1, which will put a lock record in `lock_cf`.
3. c2 starts to flashback to version ts0 that is smaller than t1, which will delete the pessimistic lock written in step 2.
4. c1 commits the transaction successfully.

We should better fail this kind of transaction to prevent breaking the flashback result, a possible method is to roll back the lock rather than deleting it directly, so that the commit will find a conflict and fail eventually.

HuSharp:

Issue Number: Close #13493 

What's Changed:

```commit-message
Rollback all locks to prevent from losing the pessimistic lock in flashback
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/actions/flashback_to_version.rs
+++ b/src/storage/txn/actions/flashback_to_version.rs
@@ -1,6 +1,6 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
-use txn_types::{Key, Lock, LockType, TimeStamp, Write, WriteType};
+use txn_types::{Key, Lock, TimeStamp, Write, WriteType};
 
 use crate::storage::{
     mvcc::{MvccReader, MvccTxn, SnapshotReader, MAX_TXN_WRITE_SIZE},
--- a/src/storage/txn/actions/flashback_to_version.rs
+++ b/src/storage/txn/actions/flashback_to_version.rs
@@ -2,6 +2,7 @@
 
 use txn_types::{Key, Lock, LockType, TimeStamp, Write, WriteType};
 
+use super::check_txn_status::rollback_lock;
 use crate::storage::{
     mvcc::{MvccReader, MvccTxn, SnapshotReader, MAX_TXN_WRITE_SIZE},
     txn::{Error, ErrorInner, Result as TxnResult},
@@ -96,7 +97,14 @@ pub fn flashback_to_version<S: Snapshot>(
             *next_lock_key = Some(key);
             break;
         }
-        txn.unlock_key(key.clone(), lock.is_pessimistic_txn());
+        rollback_lock(
+            txn,
+            reader,
+            key.clone(),
+            &lock,
+            lock.is_pessimistic_txn(),
+            true,
+        )?;
         rows += 1;
         // If the short value is none and it's a `LockType::Put`, we should delete the
         // corresponding key from `CF_DEFAULT` as well.
@@ -156,9 +164,12 @@ pub mod tests {
     use super::*;
     use crate::storage::{
         mvcc::tests::{must_get, must_get_none, write},
-        txn::actions::{
-            commit::tests::must_succeed as must_commit,
-            tests::{must_prewrite_delete, must_prewrite_put, must_rollback},
+        txn::{
+            actions::{
+                commit::tests::must_succeed as must_commit,
+                tests::{must_prewrite_delete, must_prewrite_put, must_rollback},
+            },
+            tests::{must_acquire_pessimistic_lock, must_pessimistic_prewrite_put_err},
         },
         Engine, TestEngineBuilder,
     };
@@ -188,16 +199,20 @@ pub mod tests {
         )
         .unwrap();
         assert!(!has_remain_writes);
+        let (key_locks, _) =
+            flashback_to_version_read_lock(&mut reader, &Some(key.clone()), &None, &mut statistics)
+                .unwrap();
+
         let cm = ConcurrencyManager::new(TimeStamp::zero());
         let mut txn = MvccTxn::new(start_ts, cm);
         let snapshot = engine.snapshot(Default::default()).unwrap();
-        let mut reader = SnapshotReader::new_with_ctx(version, snapshot, &ctx);
+        let mut reader = SnapshotReader::new_with_ctx(start_ts, snapshot, &ctx);
         let rows = flashback_to_version(
             &mut txn,
             &mut reader,
             &mut None,
             &mut Some(key),
-            vec![],
+            key_locks,
             key_old_writes,
             start_ts,
             commit_ts,
@@ -295,5 +310,27 @@ pub mod tests {
             0
         );
         must_get_none(&engine, k, ts);
+    }
+
+    #[test]
+    fn test_flashback_to_version_pessimistic() {
+        use kvproto::kvrpcpb::PrewriteRequestPessimisticAction::*;
+
+        let engine = TestEngineBuilder::new().build().unwrap();
+        let (k, v) = (b"k1", b"v");
+        // Prewrite and commit Put(k -> v1) with stat_ts = 10, commit_ts = 20.
+        must_prewrite_put(&engine, k, v, k, 10);
+        must_commit(&engine, k, 10, 20);
+
+        let v1 = b"v1";
+        must_acquire_pessimistic_lock(&engine, k, k, 30, 30);
+
+        // Flashback to version 25 with start_ts = 30, commit_ts = 40.
+        assert_eq!(must_flashback_write(&engine, k, 25, 30, 40), 1);
+
+        // Pessimistic Prewrite Put(k -> v1) with stat_ts = 30 will be error with
+        // Rollback.
+        must_pessimistic_prewrite_put_err(&engine, k, v1, k, 30, 30, DoPessimisticCheck);
+        must_get(&engine, k, 45, v);
     }
 }
--- a/src/storage/txn/actions/flashback_to_version.rs
+++ b/src/storage/txn/actions/flashback_to_version.rs
@@ -96,7 +96,6 @@ pub fn flashback_to_version<S: Snapshot>(
             *next_lock_key = Some(key);
             break;
         }
-        
         rollback_lock(
             txn,
             reader,
@@ -105,7 +104,6 @@ pub fn flashback_to_version<S: Snapshot>(
             lock.is_pessimistic_txn(),
             true,
         )?;
-
         rows += 1;
         // If the short value is none and it's a `LockType::Put`, we should delete the
         // corresponding key from `CF_DEFAULT` as well.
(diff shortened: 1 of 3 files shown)


JmPotato:
PTAL /cc @Connor1996 @sticnarf

JmPotato:
/cc @Connor1996 @sticnarf 

PTAL, thx!

sticnarf (approved):
LGTM

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -2,6 +2,7 @@
 
 use txn_types::{Key, Lock, LockType, TimeStamp, Write, WriteType};
 
+use super::check_txn_status::rollback_lock;
```
JmPotato, line 5: Maybe can use the `crate::storage::xxx` rather than `super` like below?

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -188,16 +199,20 @@ pub mod tests {
...
         assert!(!has_remain_writes);
+        let (key_locks, _) =
+            flashback_to_version_read_lock(&mut reader, &Some(key.clone()), &None, &mut statistics)
+                .unwrap();
+
         let cm = ConcurrencyManager::new(TimeStamp::zero());
         let mut txn = MvccTxn::new(start_ts, cm);
         let snapshot = engine.snapshot(Default::default()).unwrap();
-        let mut reader = SnapshotReader::new_with_ctx(version, snapshot, &ctx);
+        let mut reader = SnapshotReader::new_with_ctx(start_ts, snapshot, &ctx);
```
JmPotato, line 209: Why change to `start_ts`?

HuSharp: It's my fault. This is the result of not modifying back after testing...

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -96,7 +96,12 @@ pub fn flashback_to_version<S: Snapshot>(
             *next_lock_key = Some(key);
             break;
         }
+        // We need to rollback with lock.ts rather than using version timestamp by
+        // invoking rollback funtion
+        let write = Write::new_rollback(lock.ts, false);
```
JmPotato, line 101: I think we should use `rollback_lock` function in the cleanup action as much as possible.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -291,9 +306,32 @@ pub mod tests {
...
+        must_acquire_pessimistic_lock(&engine, k, k, 30, 30);
+        must_pessimistic_locked(&engine, k, 30, 30);
+
+        // Flashback to version 25 with start_ts = 30, commit_ts = 40.
+        assert_eq!(must_flashback_to_version(&engine, k, 25, 30, 40), 1);
+
+        // Pessimistic Prewrite Put(k -> v1) with stat_ts = 30 will be error with
+        // Rollback.
+        must_pessimistic_prewrite_put_err(&engine, k, v1, k, 30, 30, DoPessimisticCheck);
+        must_get(&engine, k, 45, v);
```
JmPotato, line 335: Better write two MVCC values before the pessimistic lock test and flashback to the first one, so that we can test whether the flashback works well.

HuSharp: Done.

src/storage/txn/actions/flashback_to_version.rs
HuSharp: Okay, also need to change the initialization of the reader to start_ts

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -96,7 +96,16 @@ pub fn flashback_to_version<S: Snapshot>(
             *next_lock_key = Some(key);
             break;
         }
-        txn.unlock_key(key.clone(), lock.is_pessimistic_txn());
+
```
JmPotato, line 99:
```suggestion
```

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -96,7 +96,16 @@ pub fn flashback_to_version<S: Snapshot>(
...
+        
+        rollback_lock(
+            txn,
+            reader,
+            key.clone(),
+            &lock,
+            lock.is_pessimistic_txn(),
+            true,
+        )?;
+
```
JmPotato, line 108:
```suggestion
```

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -176,6 +189,11 @@ pub mod tests {
         let snapshot = engine.snapshot(Default::default()).unwrap();
         let mut reader = MvccReader::new_with_ctx(snapshot, Some(ScanMode::Forward), &ctx);
         let mut statistics = Statistics::default();
+        let (key_locks, has_remain_locks) =
+            flashback_to_version_read_lock(&mut reader, &Some(key.clone()), &None, &mut statistics)
+                .unwrap();
+        assert!(!has_remain_locks);
+
```
JmPotato, line 196:
```suggestion
```

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -188,16 +206,17 @@ pub mod tests {
         )
         .unwrap();
         assert!(!has_remain_writes);
+
```
JmPotato, line 209:
```suggestion
```

JmPotato (approved):
The rest LGTM.

src/storage/txn/actions/flashback_to_version.rs
HuSharp: Because the rollback_lock function uses the start_ts of the reader in flashback_to_version.If the ts filled in is version timestamp, this will cause the reader after version to not be read by the start_ts when the transaction is prewritten before the version timestamp which flashback wants to go.

HuSharp:
cc @Connor1996 PTAL, thx!

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -191,13 +206,13 @@ pub mod tests {
         let cm = ConcurrencyManager::new(TimeStamp::zero());
         let mut txn = MvccTxn::new(start_ts, cm);
         let snapshot = engine.snapshot(Default::default()).unwrap();
-        let mut reader = SnapshotReader::new_with_ctx(version, snapshot, &ctx);
+        let mut reader = SnapshotReader::new_with_ctx(start_ts, snapshot, &ctx);
```
sticnarf, line 209: Emm... The `start_ts` here is just the ts of the flashback command? Do we guarantee that it equals to the start ts of the locks? If not, then that doesn't sound correct...

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -291,9 +306,35 @@ pub mod tests {
...
+        must_commit(&engine, k, 10, 15);
+        // Prewrite and commit Put(k -> v2) with stat_ts = 20, commit_ts = 25.
+        must_prewrite_put(&engine, k, v2, k, 20);
+        must_commit(&engine, k, 20, 25);
+
+        must_acquire_pessimistic_lock(&engine, k, k, 30, 30);
+        must_pessimistic_locked(&engine, k, 30, 30);
+
+        // Flashback to version 17 with start_ts = 30, commit_ts = 40.
+        assert_eq!(must_flashback_to_version(&engine, k, 17, 30, 40), 2);
```
sticnarf, line 333: In other words, I wonder if the `start_ts = lock_ts = 30` here is guaranteed by the client.

src/storage/txn/actions/flashback_to_version.rs
JmPotato: It's not guaranteed, so it looks like we'll have to rollback the lock one by one according to its own `lock.ts` rather than the flashback `start_ts`. @HuSharp

sticnarf: Yes...

JmPotato: Or we can set `reader.start_ts` to `lock.ts` each time before we call `rollback_lock`.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -96,7 +96,14 @@ pub fn flashback_to_version<S: Snapshot>(
             *next_lock_key = Some(key);
             break;
         }
-        txn.unlock_key(key.clone(), lock.is_pessimistic_txn());
+        rollback_lock(
+            txn,
+            reader,
```
JmPotato, line 101: We should set `reader.start_ts` to `lock.ts` each time before calling `rollback_lock` as discussed previously.

JmPotato (changes requested):

src/storage/txn/actions/flashback_to_version.rs
HuSharp: okay. I will do that.

JmPotato (approved):

src/storage/txn/actions/flashback_to_version.rs
sticnarf: Let's use a start_ts other than 30 here

sticnarf (commented):
The rest lgtm.

sticnarf (approved):

src/storage/txn/actions/flashback_to_version.rs
HuSharp: Thx for the reminder! It has been revised.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -96,7 +96,16 @@ pub fn flashback_to_version<S: Snapshot>(
...
+            key.clone(),
+            &lock,
+            lock.is_pessimistic_txn(),
+            true,
+        )?;
         rows += 1;
         // If the short value is none and it's a `LockType::Put`, we should delete the
         // corresponding key from `CF_DEFAULT` as well.
```
JmPotato, line 111: This check has been done by the `rollback_lock`, we can remove it.

HuSharp: Okay, so the judgment here is just to add the record of deleted row.

JmPotato: Maybe we can add an argument to the `rollback_lock` like `&mut rows` to add it inside to avoid writing the duplicated check code.

HuSharp: resolved it by using txn.modifies.len().

Connor1996 (approved):
LGTM

PR #13522, opened 2022-09-23
raftstore, engine: add flush listener and seqno worker

5kbpers:

Issue Number: Ref https://github.com/tikv/tikv/issues/12901

What's Changed:

Add flush listener and seqno worker

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/engine_rocks/src/misc.rs
```diff
@@ -130,7 +130,12 @@ impl MiscExt for RocksEngine {
         for cf in self.cf_names() {
             handles.push(util::get_cf_handle(self.as_inner(), cf)?);
         }
-        self.as_inner().flush_cfs(&handles, wait).map_err(r2e)
+        let latest_seqno = self.get_latest_sequence_number();
```
tabokie, line 133: Why not call this inside `on_flush_completed`?

components/engine_rocks/src/flush_listener.rs
```diff
@@ -0,0 +1,148 @@
...
+use parking_lot_core::SpinWait;
+use rocksdb::{EventListener, FlushJobInfo, MemTableInfo};
+use tikv_util::debug;
+
+use crate::RocksEngine;
+
+#[derive(Clone)]
+pub struct FlushListener {
+    notifier: Arc<Mutex<Box<dyn MemtableEventNotifier>>>,
+    engine: Arc<Mutex<Option<RocksEngine>>>,
```
tabokie, line 20: I think the locks can be removed in some way, you only modify them once.

components/engine_rocks/src/misc.rs
5kbpers: Some CFs may not have any data then `on_flush_completed ` will never be called.

components/engine_rocks/src/flush_listener.rs
5kbpers: Do you mean using `once_cell`?

tabokie: Create the listener after engine is created.

components/engine_rocks/src/misc.rs
tabokie: Maybe you only need one cf to call `on_flush_completed`?

components/engine_rocks/src/flush_listener.rs
tabokie: Forget it, the ordering is tricky.

components/engine_rocks/src/misc.rs
5kbpers: How to distinguish flush from rocksdb and manual flush?

components/engine_rocks/src/misc.rs
```diff
@@ -130,7 +130,12 @@ impl MiscExt for RocksEngine {
...
+        self.as_inner().flush_cfs(&handles, wait).map_err(r2e)?;
+        if let Some(listener) = self.flush_listener.as_ref() {
+            listener.notify_flush_cfs(latest_seqno);
```
BusyJay, line 136: Why need to call it explicitly?

5kbpers: Some CFs may have no data so we cannot get any notification from the listener.

BusyJay: But if wait is false, the function is returned immediately. I think you need to check the seqno from manifest somehow.

5kbpers: This function will return immediately if wait is false, `notify_flush_cfs ` cannot block. I don't think we could check manifest when CFs like default CF do not have any data, not any info will be recorded in manifest in this case.

5kbpers: oh, so the notification should only be triggered when wait is `true`.

And for WAL, as I understood rocksdb actually traces WALs with corresponding memtables, everytime when `SwitchMemtable`, [a new WAL would be created](https://github.com/tikv/rocksdb/blob/ce0ea61aecd53ac02a0e7da2b166961c7d98650a/db/db_impl/db_impl_write.cc#L2181), and stale WALs would be created later when `FindObsoleteFiles` 

And the way rocksdb checks if a memtable was stale is through [log number](https://github.com/tikv/rocksdb/blob/ce0ea61aecd53ac02a0e7da2b166961c7d98650a/db/column_family.h#L587) recorded in manifest. This log number in manifest would be updated when [picking memtable for flushing](https://github.com/tikv/rocksdb/blob/ce0ea61aecd53ac02a0e7da2b166961c7d98650a/db/flush_job.cc#L189) through `VersionEdit`.

5kbpers: Actually the reason we introduce the notification here is raft log GC, maybe we can just call `notify` when flushing memtables for raft log GC, instead of calling `notify` when every flushing?

5kbpers: Removed this notification here, and this notification would be sent when we need flushing memtable to advance raft log GC. (will be included in next PRs)

PR #13527, opened 2022-09-23
storage: remove histogram about locking key in prewrite

### LINKED ISSUE #13526: Remove useless `CONCURRENCY_MANAGER_LOCK_DURATION_HISTOGRAM`
## Development Task

It is impossible to get blocked when acquiring a lock from the concurrency manager when generating async-commit timestamp because  the latch already ensures there can't be race on the same key.

We don't need to know how long it takes to call `lock_keys` because it's always fast.

sticnarf:

Issue Number: Close #13526

What's Changed:

```commit-message
It is impossible to get blocked when acquiring a lock from the concurrency
manager when generating async-commit timestamp because the latch
already ensures there can't be race on the same key.

So, the histogram recording the duration is useless and is removed in
this commit.
```

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/mvcc/metrics.rs
+++ b/src/storage/mvcc/metrics.rs
@@ -68,12 +68,6 @@ lazy_static! {
         exponential_buckets(1.0, 2.0, 30).unwrap()
     )
     .unwrap();
-    pub static ref CONCURRENCY_MANAGER_LOCK_DURATION_HISTOGRAM: Histogram = register_histogram!(
-        "tikv_concurrency_manager_lock_duration",
-        "Histogram of the duration of lock key in the concurrency manager",
-        exponential_buckets(1e-7, 2.0, 20).unwrap() // 100ns ~ 100ms
-    )
-    .unwrap();
     pub static ref MVCC_CONFLICT_COUNTER: MvccConflictCounterVec = {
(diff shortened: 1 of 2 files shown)


ekexium (approved):

MyonKeminta (approved):
It's ok to me to remove it if it's confirmed to be a fact that its indeed always fast. But it seems that when reading requests checks the memory lock, it also need to lock the mutex of the key, which is not in latch. 😕

sticnarf:
> It's ok to me to remove it if it's confirmed to be a fact that its indeed always fast. But it seems that when reading requests checks the memory lock, it also need to lock the mutex of the key, which is not in latch. 😕

It's different.

`lock_key` locks the `AsyncMutex` in `KeyHandle`. It never has mutex conflicts.

What may have conflicts is the `Mutex` on `lock_store`. We lock the `Mutex` of `lock_store` when checking or modifying the memory lock.

sticnarf:
@MyonKeminta Will the new locking mechanism you're working on use the concurrency manager and lock the key mutex?

MyonKeminta:
> We lock the `Mutex` of `lock_store` when checking or modifying

No. I didn't touch the concurrency manager.

Issue #13528, opened 2022-09-23
raftstore-v2: integration test need supports multiple nodes and transport


nolouch:
## Development Task
Some tests like: snapshot, config change need it.

PR #13439, opened 2022-09-08
raftstore-v2: support generate snapshot 

nolouch:
### What is changed and how it works?

Issue Number: Ref https://github.com/tikv/tikv/issues/12842

What's Changed:

```commit-message
raftstore-v2: support generate snapshot
This PR implementation generate snapshot relative changes.
- add snap worker for multi-rocksdb
- implement raft::Storage snapshot interface for raftstorev2::storage 
- add tests and let the test framework supports multi-nodes.
Next part will implement apply snapshot
```

### Related changes
Base on https://github.com/tikv/tikv/pull/13388

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 

```release-note
None
```

PR #13529, opened 2022-09-23
raftstore-v2/integrations: Support for multiple nodes and transport

nolouch:

Issue Number: Close https://github.com/tikv/tikv/issues/13528
What's Changed:

```commit-message
Enhance the test infra of raftstore v2
- support multiple node `with_count`, and initial the `count` replicas for the fist region
- add transport, peers can communicate with each others
```
part of https://github.com/tikv/tikv/pull/13439
### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


BusyJay:
It's on purpose not to make raftstore-v2 depend on test-raftstore, so it won't need to depend on unnecessary dependencies like tikv crate etc. And it's also on purpose not to add transport. The goal is to make tests in raftstore-v2 more deterministic by triggering ticks and message one by one explicitly. The tests should be implemented more like raft-rs style.

nolouch:
no need.

PR #13541, opened 2022-09-27
*: Add a Raft admin command to put the region into a locking flashback state.

### LINKED ISSUE #13519: Make FlashbackToVersion become a two-phase request
Ref https://github.com/tikv/tikv/issues/13303.

## Background

The current flashback process is described below:

1. Client determines which key range to perform the flashback request, e.g, TiDB uses the table key prefix `t_` as the key range to flashback the whole cluster.
2. Client sends multiple requests to different regions on different stores with the latest `start_ts` and `commit_ts`.
3. Each region handles its own flashback progress independently.
	1. Lock the Raft proposing and lease read to block all requests including reading, writing, and scheduling.
	2. Read the old MVCC data and write it again with the given `start_ts` and `commit_ts` to pretend it's a new transaction commit.
	3. Release the Raft proposing lock and resume the lease read.
4. Client checks whether all the requests returned successfully, and retries those that failed with **new** `start_ts` and `commit_ts` until the whole flashback is done.

Since the current implementation of the flashback is not atomic, there might be some transaction problems during the process. For example, if another client tries to read the data during the flashback, it might see an incomplete transaction because some regions may have completed the flashback, while others have not.

Another problem is that since regions that have not yet performed flashback will have the opportunity to have new data written by other clients, we introduce a retry mechanism, which makes different regions inside one flashback progress may use different `start_ts` and `commit_ts` to write data that overwrite the MVCC version. This makes the flashback not atomic.

## Solution

The solution is to make `FlashbackToVersion` become a two-phase request. The new flashback process should look like this:

1. Client determines which key range to perform the flashback request, e.g, TiDB uses the table key prefix `t_` as the key range to flashback the whole cluster.
2. Client sends multiple requests to different regions on different stores to lock the Raft proposing and lease read to block all requests including reading, writing, and scheduling.
3. Client checks whether all the requests returned successfully, and retries those that failed until the whole lock phase is done.
4. Client sends multiple requests to different regions with the latest `start_ts` and `commit_ts`.
5. Each region handles its own flashback progress independently.
	1. Read the old MVCC data and write it again with the given `start_ts` and `commit_ts` to pretend it's a new transaction commit.
	2. Release the Raft proposing lock and resume the lease read.
6. Client checks whether all the requests returned successfully, and retries those that failed with **same** `start_ts` and `commit_ts` until the whole flashback is done.

Since we will make sure all regions won't have any newer writing by locking it in the first phase, all requests in the same flashback will write the overwriting MVCC data with the same `start_ts` and `commit_ts`. The lock phase also guarantees that there is no other transaction that could commit successfully during and after the flashback except the optimistic inserting transaction (see https://github.com/pingcap/tidb/issues/37961).

## Task

- [ ] Add a Raft admin command to put the region into a lock state to prevent any reading, writing, and scheduling and persist the state in the `RegionLocalState`. (https://github.com/tikv/tikv/pull/13541)
- [ ] Make `kv_flashback_to_version` become a two-phase request as described above. (https://github.com/tikv/tikv/pull/13557)
- [ ] Support only blocking the read and write operations within the key range rather than the whole region.
- [ ] Fix the case that a region that needs to apply a snapshot will lose the flashback local state due to the snapshot doesn't contain the data in `raft_cf`.

HuSharp:

Issue Number: ref #13519  ref #13303 

What's Changed:

```commit-message
Add a Raft admin command to put the region into a lock state to prevent any reading, writing, and scheduling and persist the state in the RegionLocalState.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1547,6 +1553,18 @@ where
             AdminCmdType::PrepareMerge => self.exec_prepare_merge(ctx, request),
             AdminCmdType::CommitMerge => self.exec_commit_merge(ctx, request),
             AdminCmdType::RollbackMerge => self.exec_rollback_merge(ctx, request),
+            AdminCmdType::PrepareFlashback => self.exec_flashback_apply(
+                ctx,
+                request,
+                FlashbackState::Processing,
+                ExecResult::PrepareFlashback {},
```
JmPotato, line 1560: Why do we need to pass the `ExecResult` as a parameter?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1547,6 +1553,18 @@ where
...
+            ),
+            AdminCmdType::FinishFlashback => self.exec_flashback_apply(
+                ctx,
+                request,
+                FlashbackState::Finish,
+                ExecResult::FinishFlashback {},
```
JmPotato, line 1566: Ditto.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -994,29 +994,84 @@ where
...
+        let state_key = keys::region_state_key(region_id);
+        if let Some(target_state) = self
+            .ctx
+            .engines
+            .kv
+            .get_msg_cf::<RegionLocalState>(CF_RAFT, &state_key)
+            .unwrap()
+        {
+            info!(
+                "Prepare Flashback, already has a flashback local state for {:?} ",
```
JmPotato, line 1014: What about printing out the current state?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -994,29 +994,84 @@ where
...
+                self.region(),
+            );
+            if target_state.get_flashback_state() == FlashbackState::Processing {
+                info!("region is already in flashback state");
+                // check memory state
+                if self.fsm.peer.flashback_state.is_none() {
```
JmPotato, line 1020:
```suggestion
                if self.fsm.peer.flashback_state.is_some() {
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -994,29 +994,84 @@ where
...
+                    info!("region is already in flashback state");
```
JmPotato, line 1021: The same log as line 1018?

src/server/service/kv.rs
```diff
@@ -1705,14 +1708,46 @@ fn future_flashback_to_version<
         // we start the flashback command.
         let region_id = req.get_context().get_region_id();
         let (result_tx, result_rx) = oneshot::channel();
+        let failed = Arc::new(Mutex::new(false));
```
JmPotato, line 1711: Why not just use a channel to send the result?

components/raftstore/src/store/fsm/peer.rs
HuSharp: This means that if the memory state does not exist, a callback will be called to move to the next stage: wait_apply.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2813,31 @@ where
...
+        _req: &AdminRequest,
+        state: FlashbackState,
+        result: ExecResult<EK::Snapshot>,
+    ) -> Result<(AdminResponse, ApplyResult<EK::Snapshot>)> {
+        let region = self.region.clone();
+        write_peer_state(
+            ctx.kv_wb_mut(),
+            &region,
+            PeerState::Normal,
+            None,
```
JmPotato, line 2828: Will these values overwrite the original states? Shouldn't we try to read the old values first and then just write to the flashback state?

JmPotato (changes requested):

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1547,6 +1553,18 @@ where
             AdminCmdType::PrepareMerge => self.exec_prepare_merge(ctx, request),
             AdminCmdType::CommitMerge => self.exec_commit_merge(ctx, request),
             AdminCmdType::RollbackMerge => self.exec_rollback_merge(ctx, request),
+            AdminCmdType::PrepareFlashback => self.exec_flashback_apply(
+                ctx,
+                request,
+                FlashbackState::Processing,
+                ExecResult::PrepareFlashback {},
+            ),
+            AdminCmdType::FinishFlashback => self.exec_flashback_apply(
```
JmPotato, line 1562: I think we should merge these two commands into one.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2807,40 @@ where
         ))
     }
 
+    fn set_flashback_state(
+        &self,
+        ctx: &mut ApplyContext<EK>,
+        _req: &AdminRequest,
```
JmPotato, line 2813: Why not remove `_req` since we don't need it?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2807,40 @@ where
...
+        state: FlashbackState,
+    ) -> Result<(AdminResponse, ApplyResult<EK::Snapshot>)> {
+        let region_id = self.region_id();
+        let region_state_key = keys::region_state_key(region_id);
+        let mut old_state: RegionLocalState =
+            match ctx.engine.get_msg_cf(CF_RAFT, &region_state_key) {
+                Ok(Some(s)) => s,
+                e => panic!(
```
JmPotato, line 2821: When will this panic happen?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +987,95 @@ where
...
-    // finish. We place a flag in the request, which is checked when the
-    // pre_propose_raft_command is called. Stopping tasks is done by applying
-    // the flashback-only command in this way, But for RW local reads which need
-    // to be considered, we let the leader lease to None to ensure that local reads
-    // are not executed.
-    fn on_prepare_flashback(&mut self, ch: Sender<bool>) {
+    // Call msg PrepareFlashback is to check if the persistent state is present in
+    // RegionLocalState. Once called, the persistent state and memory state will
+    // first be checked and if neither is found, then a Raft Admin Command will be
+    // proposed to persist the state in the RegionLocalState to lock the region.
```
JmPotato, line 993: Better explain the behavior or purpose of "lock".

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +987,95 @@ where
...
+    // Once this command is submitted, a callback will be invoked to call msg
+    // WaitApplyFlashback to complete the stop read stop write stop scheduling.
```
JmPotato, line 995:
```suggestion
    // WaitApplyFlashback is to wait for the new change to be applied completely. So we can perform the flashback safely later.
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +987,95 @@ where
...
+        {
+            info!(
+                "Prepare Flashback, already has a flashback local state:{:?} for {:?} ",
+                target_state.get_flashback_state(),
+                self.region(),
+            );
+            if target_state.get_flashback_state() == FlashbackState::Processing {
+                // check memory state
+                if self.fsm.peer.flashback_state.is_none() {
+                    info!("region is already has a flashback state in memory");
```
JmPotato, line 1020: If it's none, shouldn't we set it and say that "the region is not set a flashback state in memory"?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +987,95 @@ where
...
+            let mut request = new_admin_request(region_id, self.fsm.peer.peer.clone());
+            let mut admin = AdminRequest::default();
+            admin.set_cmd_type(AdminCmdType::PrepareFlashback);
+            request.set_admin_request(admin);
+            request
+        };
+        self.propose_raft_command(req, cb, DiskFullOpt::AllowedOnAlmostFull);
+    }
+
+    // Call msg WaitApplyFlashback to stop the scheduling and RW tasks, and write
```
JmPotato, line 1038: Rephrase these according to the comment before.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +987,95 @@ where
...
+    fn on_wait_apply_flashback(&mut self, cb: Callback<EK::Snapshot>) {
+        let region_id = self.region().get_id();
+        info!(
+            "wait apply flashback";
+            "region_id" => region_id,
             "peer_id" => self.fsm.peer.peer_id(),
         );
         if self.fsm.peer.flashback_state.is_some() {
-            ch.send(false).unwrap();
+            cb.invoke_with_response(new_error(Error::FlashbackInProgress(region_id)));
```
JmPotato, line 1053: I don't get why we need to invoke an error if it's in a flashback state. To Wait for the application after the flashback state has been set should be expected.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -6128,6 +6191,24 @@ where
         self.fsm.has_ready = true;
     }
 
+    fn on_ready_check_flashback_state(&mut self, state: FlashbackState) {
```
JmPotato, line 6194: I think setting the memory flashback state here is more reasonable.

src/server/service/kv.rs
```diff
@@ -1704,15 +1703,37 @@ fn future_flashback_to_version<
         // operations and then wait for the latest Raft log to be applied before
         // we start the flashback command.
         let region_id = req.get_context().get_region_id();
-        let (result_tx, result_rx) = oneshot::channel();
+        let (result_tx, result_rx) = crossbeam::channel::bounded(1);
+        let raft_router_clone_wait = raft_router_clone.clone();
+        let callback = Callback::write(Box::new(move |resp| {
```
JmPotato, line 1708: Do we need to check whether the `resp` has an error inside first?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1547,6 +1553,10 @@ where
             AdminCmdType::PrepareMerge => self.exec_prepare_merge(ctx, request),
             AdminCmdType::CommitMerge => self.exec_commit_merge(ctx, request),
             AdminCmdType::RollbackMerge => self.exec_rollback_merge(ctx, request),
+            AdminCmdType::PrepareFlashback => {
+                self.set_flashback_state(ctx, FlashbackState::Processing)
+            }
+            AdminCmdType::FinishFlashback => self.set_flashback_state(ctx, FlashbackState::Finish),
```
JmPotato, line 1559: Why not merge these two into one `AdminCmdType::SetFlashbackState`?

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2805,38 @@ where
...
+            Ok(Some(s)) => s,
+            _ => {
+                return Err(box_err!("failed to get region state of {}", region_id));
+            }
+        };
+        old_state.set_flashback_state(state);
+        ctx.kv_wb_mut()
+            .put_msg_cf(CF_RAFT, &keys::region_state_key(region_id), &old_state)
+            .unwrap_or_else(|e| {
+                panic!(
```
JmPotato, line 2828: Better avoid panic.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,95 @@ where
...
+            .kv
+            .get_msg_cf::<RegionLocalState>(CF_RAFT, &state_key)
+            .unwrap()
+        {
+            info!(
+                "Prepare Flashback, already has a flashback local state:{:?} for {:?} ",
+                target_state.get_flashback_state(),
+                self.region(),
+            );
+            if target_state.get_flashback_state() == FlashbackState::Prepare {
```
JmPotato, line 1019: We could make `flashback_state` optional, so it's more clear to tell that the state is `None` rather than `Prepare`.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,95 @@ where
...
-    fn on_finish_flashback(&mut self) {
+    fn on_finish_flashback(&mut self, ch: Sender<bool>) {
+        let region_id = self.region().get_id();
         info!(
             "finish flashback";
-            "region_id" => self.region().get_id(),
+            "region_id" => region_id,
             "peer_id" => self.fsm.peer.peer_id(),
         );
+        let cb = Callback::write(Box::new(move |resp| {
```
JmPotato, line 1063: Better let the kv thread know that the finish is committed or applied before returning a success.

components/raftstore/src/store/fsm/apply.rs
HuSharp: Do u mean that we can set the FlashbackState by setting a field in the admin request that serves to synchronize the current Flashback state?

JmPotato: Oh, I get it we use different admin cmd types to write different states.

components/raftstore/src/store/fsm/peer.rs
HuSharp: Put it in the committed callback to check it.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2803,41 @@ where
...
     }
 
+    fn set_flashback_state_persist(
+        &self,
+        ctx: &mut ApplyContext<EK>,
+        req: &AdminRequest,
+    ) -> Result<(AdminResponse, ApplyResult<EK::Snapshot>)> {
+        let region_id = self.region_id();
+        let region_state_key = keys::region_state_key(region_id);
+        let mut old_state: RegionLocalState = match ctx
```
JmPotato, line 2813: Type annotation here is unnecessary.

```suggestion
        let mut old_state = match ctx
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2803,41 @@ where
...
+            .engine
+            .get_msg_cf::<RegionLocalState>(CF_RAFT, &region_state_key)
+        {
+            Ok(Some(s)) => s,
+            _ => {
+                return Err(box_err!("failed to get region state of {}", region_id));
+            }
+        };
+        let shouild_in_flashback = req.get_cmd_type() == AdminCmdType::PrepareFlashback;
```
JmPotato, line 2822:
```suggestion
        let should_in_flashback = req.get_cmd_type() == AdminCmdType::PrepareFlashback;
```

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,101 @@ where
...
+                let raft_router_clone = self.ctx.router.clone();
+                let ch_clone = ch.clone();
+                let cb = Callback::write(Box::new(move |resp| {
+                    raft_router_clone
+                        .force_send(
+                            region_id,
+                            PeerMsg::SignificantMsg(SignificantMsg::PrepareFlashback(ch_clone)),
+                        )
+                        .unwrap();
+                    if resp.response.get_header().has_error() {
```
JmPotato, line 1030: Why not check the response first before sending the `SignificantMsg`?

components/raftstore/src/store/peer.rs
```diff
@@ -844,12 +846,11 @@ pub enum UnsafeRecoveryState {
 // it is checked every time this peer applies a new entry or a snapshot,
 // if the latest committed index is met, the syncer will be called to notify the
 // result.
-#[derive(Debug)]
-pub struct FlashbackState(Option<Sender<bool>>);
+pub struct FlashbackMemoryState(Option<Sender<bool>>);
```
JmPotato, line 849: There is no need to emphasize that this is a memory state since it will only be set in memory, let's keep the original name.

JmPotato (commented):
The rest LGTM.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +989,102 @@ where
...
+                self.region(),
+            );
+            if !state.get_is_in_flashback() {
+                // callback itself to wait for the new change to be applied completely.
+                let raft_router_clone = self.ctx.router.clone();
+                let ch_clone = ch.clone();
+                let cb = Callback::write(Box::new(move |resp| {
+                    if resp.response.get_header().has_error() {
+                        ch.send(false).unwrap();
+                        error!("send flashback prepare msg failed"; "region_id" => region_id);
```
JmPotato, line 1028: We should just return here.

src/server/service/kv.rs
```diff
@@ -1726,7 +1727,17 @@ fn future_flashback_to_version<
...
+        raft_router_clone
+            .significant_send(region_id, SignificantMsg::FinishFlashback(result_tx))
+            .unwrap();
+        if !result_rx.recv().unwrap() {
+            return Err(Error::Other(box_err!(
+                "failed to finish the region {} for flashback",
+                region_id
+            )));
+        }
+
```
JmPotato, line 1740:
```suggestion
```

JmPotato (approved):

JmPotato:
/cc @Connor1996 

PTAL, thx!

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,98 @@ where
...
             "finish flashback";
-            "region_id" => self.region().get_id(),
+            "region_id" => region_id,
             "peer_id" => self.fsm.peer.peer_id(),
         );
+        let ch_clone = ch.clone();
+        let cb = Callback::write(Box::new(move |resp| {
+            if resp.response.get_header().has_error() {
+                ch_clone.send(false).unwrap();
+                error!("send flashback prepare msg failed"; "region_id" => region_id);
```
JmPotato, line 1069: Should return here. Otherwise, the channel will be notified twice.

HuSharp: U r right...I will add it.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,98 @@ where
...
+                    let mut admin = AdminRequest::default();
+                    admin.set_cmd_type(AdminCmdType::PrepareFlashback);
+                    request.set_admin_request(admin);
+                    request
+                };
+                self.propose_raft_command(req, cb, DiskFullOpt::AllowedOnAlmostFull);
+                return;
+            }
+            // check memory state
+            if self.fsm.peer.flashback_state.is_none() {
```
JmPotato, line 1049: Better cover the case that `self.fsm.peer.flashback_state` is not `None` but the inside `ch` is.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2791,6 +2807,42 @@ where
...
+                error!(
+                    "{} failed to change flashback state to {:?} for region {}: {:?}",
+                    self.tag, req, region_id, e
+                )
+            });
+
+        Ok((
+            AdminResponse::default(),
+            ApplyResult::Res(ExecResult::SetFlashbackState {
+                region: self.region.clone(),
```
JmPotato, line 2841: This is still the old region, we should return the one that has been set `is_in_flashback`.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -5109,6 +5161,18 @@ mod tests {
...
+                        return region.get_is_in_flashback();
+                    }
+                    false
+                }
+                AdminCmdType::FinishFlashback => {
+                    if let Some(region) = region_state.modified_region.as_ref() {
+                        return !region.get_is_in_flashback();
+                    }
+                    false
+                }
```
JmPotato, line 5175: We don't need to check the actual `is_in_flashback` flag here, as long as this cmd is a `PrepareFlashback` or `FinishFlashback`, the `should_write` always needs to be true to flush the change.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,104 @@ where
...
+                let ch_clone = ch.clone();
+                let cb = Callback::write(Box::new(move |resp| {
+                    if resp.response.get_header().has_error() {
+                        ch.send(false).unwrap();
+                        error!("send flashback prepare msg failed"; "region_id" => region_id);
+                        return;
+                    }
+                    // callback itself to wait for the new change to be applied completely.
+                    raft_router_clone
+                        .force_send(
```
JmPotato, line 1031: Maybe using `significant_send`.

HuSharp: kv inside is RaftStoreRouter, fsm inside is RaftRouter.For RaftRouter because there is no SignificantRouter trait is no significant_send function, and significant_send will eventually go to force_send.

src/server/service/kv.rs
```diff
@@ -1698,21 +1697,24 @@ fn future_flashback_to_version<
...
         // we start the flashback command.
         let region_id = req.get_context().get_region_id();
-        let (result_tx, result_rx) = oneshot::channel();
+        let (result_tx, result_rx) = crossbeam::channel::bounded(1);
         raft_router_clone
-            .significant_send(region_id, SignificantMsg::PrepareFlashback(result_tx))?;
-        if !result_rx.await? {
+            .significant_send(region_id, SignificantMsg::PrepareFlashback(result_tx))
+            .unwrap();
+        if !result_rx.recv().unwrap() {
```
Connor1996, line 1711: why change it, it blocks the thread

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,104 @@ where
...
+        let state_key = keys::region_state_key(region_id);
+        if let Some(state) = self
+            .ctx
+            .engines
+            .kv
+            .get_msg_cf::<RegionLocalState>(CF_RAFT, &state_key)
+            .unwrap()
+        {
+            info!(
+                "Prepare Flashback, at now flashback persist state is: {} for {:?} ",
```
Connor1996, line 1016: please follow the format with other logs

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,104 @@ where
...
-        if self.fsm.peer.flashback_state.is_some() {
-            ch.send(false).unwrap();
-            return;
+        // check persist state
+        let state_key = keys::region_state_key(region_id);
+        if let Some(state) = self
+            .ctx
+            .engines
+            .kv
+            .get_msg_cf::<RegionLocalState>(CF_RAFT, &state_key)
```
Connor1996, line 1012: Why check persist state here? Persistent state is used only on start to recover the memory state.

components/raftstore/src/store/util.rs
```diff
@@ -277,6 +280,25 @@ pub fn compare_region_epoch(
...
+        }
+        if req.has_admin_request()
+            && (req.get_admin_request().get_cmd_type() == AdminCmdType::PrepareFlashback
+                || req.get_admin_request().get_cmd_type() == AdminCmdType::FinishFlashback)
+        {
+            return Ok(());
+        }
+        return Err(Error::FlashbackInProgress(region.get_id()));
+    }
+    Ok(())
```
JmPotato, line 299:
```suggestion
    Err(Error::FlashbackInProgress(region.get_id()))
```

components/raftstore/src/store/fsm/peer.rs
JmPotato: I think it's because we can not know whether the current peer is in a flashback state in `PeerFSM` before we read once from the persistent state?

Connor1996: you can just check the memory state. Memory state is filled on start-up.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,104 @@ where
...
+                    if state.channel_is_none() {
+                        self.fsm.peer.flashback_state = Some(FlashbackState::new(Some(ch)));
+                    }
+                }
+            }
         }
-        self.fsm.peer.flashback_state = Some(FlashbackState::new(ch));
         // Let the leader lease to None to ensure that local reads are not executed.
         self.fsm.peer.leader_lease_mut().expire_remote_lease();
         self.fsm.peer.maybe_finish_flashback_wait_apply();
```
Connor1996, line 1060: It is not needed anymore as the command is going thourgh the Raft now. When it's applied, all the previous writes must have been applied.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -987,36 +988,104 @@ where
...
+            if !state.get_region().get_is_in_flashback() {
+                let raft_router_clone = self.ctx.router.clone();
+                let ch_clone = ch.clone();
+                let cb = Callback::write(Box::new(move |resp| {
+                    if resp.response.get_header().has_error() {
+                        ch.send(false).unwrap();
+                        error!("send flashback prepare msg failed"; "region_id" => region_id);
+                        return;
+                    }
+                    // callback itself to wait for the new change to be applied completely.
```
Connor1996, line 1029: Why? the callback is exactly called when it's applied.

src/server/service/kv.rs
```diff
@@ -1698,21 +1697,24 @@ fn future_flashback_to_version<
...
         // later flashback. This will first block all scheduling, read and write
         // operations and then wait for the latest Raft log to be applied before
         // we start the flashback command.
         let region_id = req.get_context().get_region_id();
-        let (result_tx, result_rx) = oneshot::channel();
+        let (result_tx, result_rx) = crossbeam::channel::bounded(1);
         raft_router_clone
-            .significant_send(region_id, SignificantMsg::PrepareFlashback(result_tx))?;
-        if !result_rx.await? {
+            .significant_send(region_id, SignificantMsg::PrepareFlashback(result_tx))
```
Connor1996, line 1709: As it's a admin command now, no need to trigger it by significant message. We can just followe the normal way to propose admin command. Check `fn send_admin_request` for reference.

components/raftstore/src/store/fsm/peer.rs
HuSharp: I think since the `notify` method will be called after `apply` to send the result of apply to peer fsm.The memory state will be set in the `on_flashback_memory_set` function. So here we need to set the meta information of cf and current region.

HuSharp:
Finished addressing all the comments, PTAL! @Connor1996 @JmPotato

components/raftstore/src/store/peer.rs
```diff
@@ -1047,7 +1020,7 @@ where
     /// lead_transferee if the peer is in a leadership transferring.
     pub lead_transferee: u64,
     pub unsafe_recovery_state: Option<UnsafeRecoveryState>,
-    pub flashback_state: Option<FlashbackState>,
+    pub is_in_flashback: bool,
```
Connor1996, line 1023: add comment here

src/server/service/kv.rs
```diff
@@ -2400,6 +2415,57 @@ fn needs_reject_raft_append(reject_messages_on_memory_ratio: f64) -> bool {
...
+    let mut req = RaftCmdRequest::default();
+    req.mut_header().set_region_id(ctx.get_region_id());
+    req.mut_header()
+        .set_region_epoch(ctx.get_region_epoch().clone());
+    req.mut_header().set_peer(ctx.get_peer().clone());
+    req.mut_header()
+        .set_flags(WriteBatchFlags::FLASHBACK.bits());
+    req.set_admin_request(admin);
+    req.mut_header()
+        .set_flags(WriteBatchFlags::FLASHBACK.bits());
```
Connor1996, line 2443: duplicated

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1548,6 +1561,9 @@ where
             AdminCmdType::PrepareMerge => self.exec_prepare_merge(ctx, request),
             AdminCmdType::CommitMerge => self.exec_commit_merge(ctx, request),
             AdminCmdType::RollbackMerge => self.exec_rollback_merge(ctx, request),
+            AdminCmdType::PrepareFlashback | AdminCmdType::FinishFlashback => {
+                self.set_flashback_state_persist(ctx, request)
```
Connor1996, line 1565: let's follow the naming, `exec_flashback`

Connor1996 (approved):
LGTM

JmPotato (approved):

AndreMouche (approved):
LGTM

PR #13555, opened 2022-09-28
raftstore-v2: add snapshot basic logic in peer storage

nolouch:

What's Changed:

Part of https://github.com/tikv/tikv/pull/13439

```commit-message
this commit is part of support snapshot in raftstore v2
- add snapshot basic implementation in peer storage
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore-v2/src/raft/storage.rs
```diff
@@ -58,9 +72,15 @@ pub struct Storage<ER> {
...
+
+    /// Snapshot state
+    snap_state: RefCell<SnapState>,
+    gen_snap_task: RefCell<Option<GenSnapTask>>,
+    snap_tried_cnt: RefCell<usize>,
+    tablet_suffix: Option<u64>,
 }
 
-impl<ER> Debug for Storage<ER> {
+impl<ER: RaftEngine> Debug for Storage<ER> {
```
BusyJay, line 83: Why the change?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -58,9 +72,15 @@ pub struct Storage<ER> {
     /// at least once dispite whether the state changes since create.
     ever_persisted: bool,
     logger: Logger,
+
+    /// Snapshot state
+    snap_state: RefCell<SnapState>,
+    gen_snap_task: RefCell<Option<GenSnapTask>>,
+    snap_tried_cnt: RefCell<usize>,
+    tablet_suffix: Option<u64>,
```
BusyJay, line 80: Isn't it recorded in region_state?

BusyJay (commented):
Now generating and applying snapshot become very lightweight, I suggest to handling applying snapshot in async writer and generating snapshot in raft log fetch worker.

nolouch:
>  I suggest to handling applying snapshot in async writer and generating snapshot in raft log fetch worker.

I think putting the tasks in a work more readable currently and there is a `destroy task` need to maintain. maybe we can consider unifying some works later.

BusyJay:
> I think putting the tasks in a work more readable currently and there is a `destroy task` need to maintain. maybe we can consider unifying some works later.

Destroying a peer is just a simple write in async writer. Applying snapshot is also a simple write in async writer, what else needs to be done is renaming the directory.

nolouch:
Moved generate to log fetch worker

components/raftstore/src/store/worker/raftlog_fetch.rs
```diff
@@ -19,7 +22,15 @@ pub enum Task {
         tried_cnt: usize,
         term: u64,
     },
-    // More to support, suck as fetch entries ayschronously when apply and schedule merge
+
+    // GenTabletSnapshot is used to generate tablet snapshot.
+    GenTabletSnapshot {
```
BusyJay, line 27: There will be no states in tablet, so the metadata needs to be either stored in `Storage` or in the task.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +262,170 @@ impl<ER: RaftEngine> Storage<ER> {
...
+            STORE_SNAPSHOT_VALIDATION_FAILURE_COUNTER.epoch.inc();
+            return false;
+        }
+
+        true
+    }
+
+    /// Gets a snapshot. Returns `SnapshotTemporarilyUnavailable` if there is no
+    /// unavailable snapshot.
+    pub fn snapshot(&self, request_index: u64, to: u64) -> raft::Result<Snapshot> {
```
BusyJay, line 326: How about making it retry indefinitely, and report error as a metrics? This should simplify a lot.

components/raftstore/src/store/worker/raftlog_fetch.rs
nolouch: Do you mean `apply state` or else? I will add it to the next PR.

BusyJay: Apply state and region state.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +258,159 @@ impl<ER: RaftEngine> Storage<ER> {
...
+        }
+
+        if !tried || !last_canceled {
+            STORE_SNAPSHOT_RETIRES_COUNTER.inc();
+        }
+
+        info!(
+            self.logger(),
+            "requesting snapshot";
+            "region_id" => self.region().get_id(),
```
BusyJay, line 375: All IDs are included in the logger.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +258,159 @@ impl<ER: RaftEngine> Storage<ER> {
...
+            "peer_id" => self.peer().get_id(),
+            "tablet_index" => self.get_tablet_index(),
+            "request_index" => request_index,
+            "request_peer" => to,
+        );
+
+        let (sender, receiver) = mpsc::sync_channel(1);
+        // TODO(nolouch): remove cancel?
```
BusyJay, line 383: Cancel is still necessary.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +258,159 @@ impl<ER: RaftEngine> Storage<ER> {
...
+        let canceled = Arc::new(AtomicBool::new(false));
+        let index = Arc::new(AtomicU64::new(0));
+        *snap_state = SnapState::Generating {
+            canceled: canceled.clone(),
+            index,
+            receiver,
+        };
+
+        let store_id = self
```
BusyJay, line 392: What's the purpose?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +258,159 @@ impl<ER: RaftEngine> Storage<ER> {
...
+        let mut tried = false;
+        let mut last_canceled = false;
+        if let SnapState::Generating {
+            ref canceled,
+            ref receiver,
+            ..
+        } = *snap_state
+        {
+            tried = true;
+            last_canceled = canceled.load(Ordering::SeqCst);
```
BusyJay, line 334: If the snapshot is cancel, the state should be reset. I don't see why still need to check `last_canceled`.

components/raftstore-v2/src/raft/storage.rs
nolouch: Do you mean the logger will init with the head like `[region xxx][1,2,3]`?  I don't see that.

components/raftstore-v2/src/raft/storage.rs
nolouch: `canceled` is a shared value between storage and the runner work. If the snapshot is canceled, only change the value of `canceled`, then the runner work will drop the task, then the state will be reset.

components/raftstore-v2/src/raft/storage.rs
nolouch: I mistake here, that was used for tiflash  with `multi_sst_sanpshot` in v1.

components/raftstore-v2/src/raft/storage.rs
BusyJay: > Do you mean the logger will init with the head like [region xxx][1,2,3]?

Yes, check

https://github.com/tikv/tikv/blob/94c8aa5c3c5335ea99281bea7c6c79f61cb201de/components/raftstore-v2/src/raft/storage.rs#L202

components/raftstore-v2/src/raft/storage.rs
BusyJay: Why not just change the value and drop the state?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +258,172 @@ impl<ER: RaftEngine> Storage<ER> {
...
+            _ => self.region_state.get_tablet_index(),
+        }
+    }
+
+    /// Cancel generating snapshot.
+    pub fn cancel_generating_snap(&mut self, compact_to: Option<u64>) {
+        let snap_state = self.snap_state.borrow();
+        if let SnapState::Generating {
+            ref canceled,
+            ref index,
```
BusyJay, line 278: How about redefine the struct for v2? For example, index doesn't need to be atomic.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,6 +258,172 @@ impl<ER: RaftEngine> Storage<ER> {
...
+            ..
+        } = *snap_state
+        {
+            if !canceled.load(Ordering::SeqCst) {
+                if let Some(idx) = compact_to {
+                    let snap_index = index.load(Ordering::SeqCst);
+                    if snap_index == 0 || idx <= snap_index + 1 {
```
BusyJay, line 285: When will it be 0?

nolouch: Once the `generate_and_schedule_snapshot` not be called. such as wait apply sync the write before  schedule the snapshot task?

components/raftstore-v2/src/raft/storage.rs
nolouch: added apply task relative log, the index will be used to store the `last_applied_index`, then storage can use it to judge whether if the snapshot task should be canceled after compact.

nolouch:
PTAL @BusyJay @tonyxuqqi

components/raftstore/src/store/mod.rs
```diff
@@ -15,7 +15,7 @@ pub mod transport;
 #[macro_use]
 pub mod util;
 
-mod async_io;
+pub mod async_io;
```
BusyJay, line 18: Should not be public.

components/raftstore/src/store/metrics.rs
```diff
@@ -489,6 +489,12 @@ lazy_static! {
     pub static ref STORE_SNAPSHOT_VALIDATION_FAILURE_COUNTER: SnapValidVec =
         auto_flush_from!(STORE_SNAPSHOT_VALIDATION_FAILURE_COUNTER_VEC, SnapValidVec);
 
+    pub static ref STORE_SNAPSHOT_RETIRES_COUNTER: IntCounter =
```
BusyJay, line 492: Why not reuse existing metrics?

components/raftstore/src/store/async_io/read.rs
```diff
@@ -19,7 +23,18 @@ pub enum Task {
...
+
+    // GenTabletSnapshot is used to generate tablet snapshot.
+    GenTabletSnapshot {
+        region_id: u64,
+        tablet_index: u64,
+        region_state: RegionLocalState,
+        last_applied_term: u64,
+        last_applied_index: u64,
+        canceled: Arc<AtomicBool>,
+        notifier: SyncSender<RaftSnapshot>,
```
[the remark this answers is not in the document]
nolouch: the notify is a channel created at `Storage::snapshot`, which will receive the result.

components/raftstore/src/store/metrics.rs
nolouch: cannot find a name suitable existing metrics

components/raftstore/src/store/async_io/read.rs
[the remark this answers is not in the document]
BusyJay: Yes, but why not? Using router can speed up snapshot by manually ping after the snapshot is generated.

components/raftstore/src/store/peer.rs
```diff
@@ -1043,7 +1043,7 @@ where
         store_id: u64,
         cfg: &Config,
         region_scheduler: Scheduler<RegionTask<EK::Snapshot>>,
-        raftlog_fetch_scheduler: Scheduler<RaftlogFetchTask>,
+        raftlog_fetch_scheduler: Scheduler<ReadTask<EK>>,
```
BusyJay, line 1046: Field name not updated.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -1081,7 +1082,7 @@ pub struct RaftPollerBuilder<EK: KvEngine, ER: RaftEngine, T> {
     split_check_scheduler: Scheduler<SplitCheckTask>,
     cleanup_scheduler: Scheduler<CleanupTask>,
     raftlog_gc_scheduler: Scheduler<RaftlogGcTask>,
-    raftlog_fetch_scheduler: Scheduler<RaftlogFetchTask>,
+    raftlog_fetch_scheduler: Scheduler<ReadTask<EK>>,
```
BusyJay, line 1085: Field name not updated.

components/raftstore/src/store/async_io/read.rs
```diff
@@ -119,6 +141,11 @@ where
                     },
                 );
             }
+            ReadTask::GenTabletSnapshot { region_id, .. } => {
+                // TODO(FIXME): implement generate tablet snapshot for raftstore v2
```
BusyJay, line 145: If there is TODO, then there should be no FIXME.

components/raftstore/src/store/transport.rs
```diff
@@ -173,10 +174,15 @@ where
...
+impl<EK: KvEngine, ER: RaftEngine> AsyncReadNotifier for RaftRouter<EK, ER> {
     #[inline]
-    fn notify(&self, region_id: u64, fetched: FetchedLogs) {
+    fn notify_fetched_logs(&self, region_id: u64, fetched: FetchedLogs) {
         // Ignore region not found as it may be removed.
         let _ = self.significant_send(region_id, SignificantMsg::RaftlogFetched(fetched));
     }
+
+    #[inline]
+    fn notify_snapshot_generated(&self, _region_id: u64, _snapshot: Box<RaftSnapshot>) {
```
BusyJay, line 185: Should be in the same format as `notify_fetched_logs`.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -343,6 +345,27 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
         }
     }
 
+    pub fn handle_snapshot(&mut self, snap_task: GenSnapTask) {
```
BusyJay, line 348: It can be moved to a new file in operation/ready/snapshot.rs.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -373,9 +373,9 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
         workers
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
-        let log_fetch_scheduler = workers.log_fetch_worker.start(
-            "raftlog-fetch-worker",
-            RaftlogFetchRunner::new(router.clone(), raft_engine.clone()),
+        let read_scheduler = workers.async_read_worker.start(
+            "async-read-worker",
```
BusyJay, line 377: We should check if any metrics/alerts depends on the name.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,9 +262,157 @@ impl<ER: RaftEngine> Storage<ER> {
...
+
+    fn get_tablet_index(&self) -> u64 {
+        match self.region_state.get_state() {
+            PeerState::Tombstone | PeerState::Applying => 0,
+            _ => self.region_state.get_tablet_index(),
+        }
+    }
+
+    /// Cancel generating snapshot.
+    pub fn cancel_generating_snap(&mut self, compact_to: Option<u64>) {
```
BusyJay, line 278: Better move this function to the new file.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,9 +262,157 @@ impl<ER: RaftEngine> Storage<ER> {
...
+    /// unavailable snapshot.
+    pub fn snapshot(&self, request_index: u64, to: u64) -> raft::Result<Snapshot> {
+        let mut snap_state = self.snap_state.borrow_mut();
+        match *snap_state {
+            SnapState::Generating { .. } => {
+                return Err(raft::Error::Store(
+                    raft::StorageError::SnapshotTemporarilyUnavailable,
+                ));
+            }
+            SnapState::Generated(ref s) => {
```
BusyJay, line 370: When will the field be cleared?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -240,9 +262,157 @@ impl<ER: RaftEngine> Storage<ER> {
...
+            "request_index" => request_index,
+            "request_peer" => to,
+        );
+        let (_, receiver) = mpsc::sync_channel(1);
+        let canceled = Arc::new(AtomicBool::new(false));
+        let index = Arc::new(AtomicU64::new(0));
+        *snap_state = SnapState::Generating {
+            canceled: canceled.clone(),
+            index: index.clone(),
+            receiver,
```
BusyJay, line 401: Better define a new struct, so we don't need to create a meaningless receiver.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -49,18 +60,22 @@ pub fn write_initial_states(wb: &mut impl RaftLogBatch, region: Region) -> Resul
...
     region_state: RegionLocalState,
     /// Whether states has been persisted before. If a peer is just created by
     /// by messages, it has not persisted any states, we need to persist them
     /// at least once dispite whether the state changes since create.
     ever_persisted: bool,
     logger: Logger,
+
+    /// Snapshot state
+    snap_state: RefCell<SnapState>,
+    gen_snap_task: RefCell<Option<GenSnapTask>>,
```
BusyJay, line 75: I suggest to use `Box` to reduce the struct size.

components/raftstore/src/store/peer.rs
nolouch: I didn't change it in v1.

components/raftstore-v2/src/batch/store.rs
nolouch: Only change it for v2.

BusyJay: V2 also reuse the grafana dashboard.

nolouch: Only one metric about raftlog's pending fetch task. I added it in the grafana tmpl.

components/raftstore/src/store/fsm/store.rs
nolouch: I will keep it as `raftlog_fetch_scheduler` in v1 first. I can change it in a separated PR if needed.

components/raftstore/src/store/transport.rs
nolouch: changed `notify_fetched_logs` to `notify_logs_fetched`.

nolouch:
PTAL @BusyJay @Connor1996

metrics/grafana/tikv_details.json
```diff
@@ -17417,6 +17417,14 @@
               "interval": "",
               "legendFormat": "pending-task",
               "refId": "D"
+            },
+            {
+              "exemplar": true,
+              "expr": "sum(tikv_worker_pending_task_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=~\"$tidb_cluster.*\", instance=~\"$instance\", name=~\"async-read-worker\"})",
+              "hide": true,
+              "interval": "",
+              "legendFormat": "pending-task-v2",
```
BusyJay, line 17426: What is v2?

components/raftstore-v2/src/router/internal_message.rs
```diff
@@ -1,13 +1,85 @@
...
+            canceled,
+            for_balance: false,
+        }
+    }
+
+    pub fn set_for_balance(&mut self) {
+        self.for_balance = true;
+    }
+
+    pub fn generate_and_schedule_snapshot<EK: KvEngine>(
```
BusyJay, line 48: Should be implemented for `Apply` and defined in "ready/snapshot.rs".

components/raftstore-v2/src/router/internal_message.rs
```diff
@@ -1,13 +1,85 @@
...
+use raft::eraftpb::Snapshot as RaftSnapshot;
+use raftstore::store::{fsm::ChangePeer, ReadTask};
+use tikv_util::{box_try, worker::Scheduler};
+
+use crate::{
+    operation::{AdminCmdResult, CommittedEntries},
+    Result,
+};
+
+pub struct GenSnapTask {
```
BusyJay, line 23: Should be defined in "ready/snapshot.rs".

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -117,6 +126,9 @@ impl<EK: KvEngine, R> Apply<EK, R> {
     }
 
     #[inline]
+    pub fn tablet(&self) -> &EK {
+        &self.tablet
+    }
```
BusyJay, line 131: Missing blank line and `inline`.

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -298,6 +299,14 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
             self.handle_raft_committed_entries(ctx, ready.take_committed_entries());
         }
 
+        // Check whether there is a pending generate snapshot task, the task
+        // needs to be sent to the apply system.
+        // Always sending snapshot task behind apply task, so it gets latest
```
BusyJay, line 304:
```suggestion
        // Always sending snapshot task after apply task, so it gets latest
```

metrics/grafana/tikv_details.json
nolouch: changed to `async-read-pending-task`

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,276 @@
...
+            canceled,
+            for_balance: false,
+        }
+    }
+
+    pub fn set_for_balance(&mut self) {
+        self.for_balance = true;
+    }
+
+    pub fn generate_and_schedule_snapshot<EK: KvEngine>(
```
BusyJay, line 75: How about implement the function for `Apply` instead?

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,276 @@
...
+            .field("region_id", &self.region_id)
+            .finish()
+    }
+}
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    pub fn on_snapshot_generated(&mut self, snapshot: Box<Snapshot>) {
+        self.storage_mut().on_snapshot_generated(snapshot);
+        self.raft_group_mut().ping();
+        self.set_has_ready();
```
BusyJay, line 110: Snapshot state should be cleared. Otherwise it can be leaked.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub fn snapshot(&self, request_index: u64, to: u64) -> raft::Result<Snapshot> {
+        let mut snap_state = self.snap_state_mut();
+        match *snap_state {
+            SnapState::Generating { .. } => {
+                return Err(raft::Error::Store(
+                    raft::StorageError::SnapshotTemporarilyUnavailable,
+                ));
+            }
+            SnapState::Generated(ref s) => {
+                let snap = *s.clone();
```
BusyJay, line 153: How about:
```
let SnapState::Generated(snap) = mem::replace(&mut snap_state, SnapState::Relax) else { unreachable!() };
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,276 @@
...
+                if self.validate_snap(&snap, request_index) {
+                    return Ok(snap);
+                }
+            }
+            _ => {}
+        }
+
+        if SnapState::Relax != *snap_state {
+            panic!(
+                "[region {}] unexpected state: {:?}",
```
BusyJay, line 164: Should use `logger.list()`.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,276 @@
...
+            return false;
+        }
+
+        true
+    }
+
+    /// Cancel generating snapshot.
+    pub fn cancel_generating_snap(&mut self, compact_to: Option<u64>) {
+        let mut snap_state = self.snap_state_mut();
+        if let SnapState::Generating {
```
BusyJay, line 238: How about:
```
let SnapState::Generating { canceled, index } = snap_state else { return };
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,276 @@
...
+            );
+            STORE_SNAPSHOT_VALIDATION_FAILURE_COUNTER.cancel.inc();
+        }
+    }
+
+    pub fn on_snapshot_generated(&mut self, snap: Box<Snapshot>) {
+        let mut snap_state = self.snap_state_mut();
+        if let SnapState::Generating {
+            ref canceled,
+            ref index,
```
BusyJay, line 265: Maybe we should check if index is matched.

metrics/grafana/tikv_details.json
nolouch: legend-format is the name for this metrics, like `{{instance}}-{{type}}`.
image

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,275 @@
...
+            (&SnapState::Generated(ref snap1), &SnapState::Generated(ref snap2)) => {
+                *snap1 == *snap2
+            }
+            _ => false,
+        }
+    }
+}
+
+pub struct GenSnapTask {
+    pub(crate) region_id: u64,
```
BusyJay, line 52: Why it needs to be public?

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,275 @@
...
+    // Fill it when you are going to generate the snapshot.
+    // index used to check if the gen task should be canceled.
+    pub index: Arc<AtomicU64>,
+    // Fetch it to cancel the task if necessary.
```
BusyJay, line 56:
```suggestion
    // Set it to true to cancel the task if necessary.
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
+    /// Handle snapshot.
+    ///
+    /// Will schedule a task to read worker and then generate a snapshot
+    /// asynchronously.
+    pub fn handle_snapshot(&mut self, snap_task: GenSnapTask) {
+        // Flush before do snapshot.
+        self.flush();
```
BusyJay, line 104: Perhaps should check if `cancel` is set first.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,275 @@
...
+            "snapshot is canceled";
+            "compact_to" => compact_to,
+        );
+        STORE_SNAPSHOT_VALIDATION_FAILURE_COUNTER.cancel.inc();
+    }
+
+    /// Try to switch snap state to generated. only `Generating` can switch to
+    /// `Generated`.
+    ///  TODO: make the snap state more clearer, the snapshot must be consumed.
+    pub fn try_switch_snap_state_to_generated(&self, snap: Box<Snapshot>) -> bool {
```
BusyJay, line 253:
```suggestion
    pub fn on_snapshot_generated(&self, snap: Box<Snapshot>) -> bool {
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,275 @@
...
+        let mut snap_state = self.snap_state_mut();
+        let SnapState::Generating {
+            ref canceled,
+            ref index,
+         } = *snap_state else { return false };
+
+        if snap.get_metadata().get_index() < index.load(Ordering::SeqCst) {
+            drop(snap_state);
+            self.cancel_generating_snap(None);
```
BusyJay, line 262: Why cancel?

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,275 @@
...
+        let canceled = Arc::new(AtomicBool::new(false));
+        let index = Arc::new(AtomicU64::new(0));
+        *snap_state = SnapState::Generating {
+            canceled: canceled.clone(),
+            index: index.clone(),
+        };
+
+        let task = GenSnapTask::new(self.region().get_id(), index, canceled);
+        let mut gen_snap_task = self.gen_snap_task_mut();
+        assert!(gen_snap_task.is_none());
```
BusyJay, line 177: If:

0. raft calls snapshot()
1. Somehow peer cancel snapshot
2. raft calls snapshot() again
3. handle ready

This line may panic.

components/raftstore-v2/src/operation/ready/snapshot.rs
nolouch: the snap index is invalid here. but I think it impossible to happen. I see you commented to check the index before, what do you mean?

BusyJay: It can happen in following sequence:

0. raft calls snapshot
1. snapshot is scheduled and generated
2. snapshot is canceled
3. raft calls snapshot again
4. snapshot is scheduled and generated
5. peer storage on snapshot generated with old snapshot generated at 1.

I don't think it needs to be canceled as it's still waiting for snapshot generated at 4.

nolouch: got, I removed cancel here.

components/raftstore-v2/src/operation/ready/snapshot.rs
nolouch: added logic `takethe task` in cancel.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,290 @@
...
+//! the curren snapshot state.
+//!
+//! # Porcess Overview
+//!
+//! generate snapshot:
+//! - Raft call `snapshot` interface to acquire a snapshot, then storage setup
+//!   the gen_snap_task.
+//! - handle ready will send the gen_snap_task to the apply work
+//! - apply worker flush the data, then schedule a gen tablet snapshot task to
+//!   async read work with region state and apply state.
```
BusyJay:
```suggestion
//!   async read worker with region state and apply state.
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,290 @@
...
+//! - async read worker generates the tablet snapshot and sends the result to
+//!   store fsm.
```
BusyJay:
```suggestion
//!   peer fsm.
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,290 @@
...
+         } = *snap_state else { return false };
+
+        if snap.get_metadata().get_index() < index.load(Ordering::SeqCst) {
+            return false;
+        }
+        // Should changed `SnapState::Generated` to `SnapState::Relax` when the
+        // snap is consumed or canceled. Such as leader changed, the state of generated
+        // should be reset.
+        *snap_state = SnapState::Generated(snap);
+        info!(
```
BusyJay, line 284: Useless log.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -0,0 +1,290 @@
...
+
+impl<EK: KvEngine, ER: RaftEngine> Storage<EK, ER> {
+    /// Gets a snapshot. Returns `SnapshotTemporarilyUnavailable` if there is no
+    /// unavailable snapshot.
+    pub fn snapshot(&self, request_index: u64, to: u64) -> raft::Result<Snapshot> {
+        let mut snap_state = self.snap_state_mut();
+        match *snap_state {
+            SnapState::Generating { ref canceled, .. } => {
+                if canceled.load(Ordering::SeqCst) {
+                    self.cancel_generating_snap(None);
```
BusyJay, line 154: Seems can be simplified as `*snap_state = SnapState::Relax; metrics.inc()`

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -358,4 +446,78 @@ mod tests {
...
+        ));
+        // create tablet with region_id 1
+        let tablet = factory
+            .open_tablet(1, Some(10), OpenOptions::default().set_create_new(true))
+            .unwrap();
+        // setup read runner worker and peer storage
+        let mut worker = Worker::new("test-read-worker").lazy_build("test-read-worker");
+        let sched = worker.scheduler();
+        let logger = slog_global::borrow_global().new(o!());
+        let mut s = Storage::uninit(
```
BusyJay, line 477: Technically, it's wrong to take a snapshot from uninitialized storage.

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -358,4 +446,78 @@ mod tests {
...
+
+        // test cancel snapshot
+        let snap = s.snapshot(0, 0);
+        let unavailable = RaftError::Store(StorageError::SnapshotTemporarilyUnavailable);
+        assert_eq!(snap.unwrap_err(), unavailable);
+        let gen_task = s.gen_snap_task.borrow_mut().take().unwrap();
+        apply.schedule_gen_snapshot(gen_task);
+        let res = rx.recv_timeout(Duration::from_secs(1)).unwrap();
+        s.cancel_generating_snap(None);
+        assert_eq!(*s.snap_state.borrow(), SnapState::Relax);
```
BusyJay, line 521: How about also adding cases mentioned in this PR?

BusyJay (commented):
Rest LGTM

components/raftstore-v2/src/operation/ready/snapshot.rs
nolouch: I think it's better unify use cancel function here, it's not in key path not hurt performance.

components/raftstore-v2/src/raft/storage.rs
nolouch: I added TODO here. will add the test after implement async task works, otherwise the snapshot metadata is a mock data.

BusyJay (approved):

components/raftstore/src/store/async_io/read.rs
```diff
@@ -1,15 +1,20 @@
...
-use raft::GetEntriesContext;
+use kvproto::raft_serverpb::RegionLocalState;
+use raft::{eraftpb::Snapshot as RaftSnapshot, GetEntriesContext};
 use tikv_util::worker::Runnable;
 
 use crate::store::{RaftlogFetchResult, MAX_INIT_ENTRY_COUNT};
 
-pub enum Task {
+pub enum ReadTask<EK> {
     PeerStorage {
```
Connor1996, line 18: How about `FetchLogs`?

components/raftstore-v2/src/raft/storage.rs
```diff
@@ -49,18 +54,22 @@ pub fn write_initial_states(wb: &mut impl RaftLogBatch, region: Region) -> Resul
 /// A storage for raft.
 ///
 /// It's similar to `PeerStorage` in v1.
-pub struct Storage<ER> {
-    entry_storage: EntryStorage<ER>,
+pub struct Storage<EK: KvEngine, ER> {
```
Connor1996, line 57: Is `:KvEngine` necessary?

Connor1996:
> Now generating and applying snapshot become very lightweight, I suggest to handling applying snapshot in async writer and generating snapshot in raft log fetch worker.

@BusyJay  @nolouch raft log fetch worker may accumulate lots of tasks, it's unwise to generate snapshot there.

components/raftstore-v2/src/raft/storage.rs
nolouch: yes, otherwise it reports as unsafe to send.

components/raftstore/src/store/async_io/read.rs
nolouch: maybe a separated PR later.

BusyJay:
> @BusyJay @nolouch raft log fetch worker may accumulate lots of tasks, it's unwise to generate snapshot there.

Then we should use a thread pool for the read tasks.

Connor1996:
> > @BusyJay @nolouch raft log fetch worker may accumulate lots of tasks, it's unwise to generate snapshot there.
> 
> Then we should use a thread pool for the read tasks.

Fetch log tasks can still exhaust the threads, and affect snapshot generation. Better not to mix it with heavy io tasks.

BusyJay:
> Fetch log tasks can still exhaust the threads, and affect snapshot generation. Better not to mix it with heavy io tasks.

It's not an ideal way to create a thread(pool) for every type of tasks. Fairness between tasks should be solved by the thread pool itself.

BusyJay:
> But for the blocking IO, it must suffer from blocking by fetch log tasks.

This is also doable. For example, using yatp for read tasks and change the fetch log tasks to execution slices. And v2 snapshot can be assigned with high priority. It's just the same as coprocessor.

> Why bother to move it out from the region worker?

To reduce threads and unify concepts.

Connor1996:
> > But for the blocking IO, it must suffer from blocking by fetch log tasks.
> 
> This is also doable. For example, using yatp for read tasks and change the fetch log tasks to execution slices. And v2 snapshot can be assigned with high priority. It's just the same as coprocessor.
> 
> > Why bother to move it out from the region worker?
> 
> To reduce threads and unify concepts.

It may cost several hundred milliseconds, even seconds for one fetch log task. You can't keep the execution slice small.

BusyJay:
> It may cost several hundred milliseconds, even seconds for one fetch log task. You can't keep the execution slice small based on current raft engine implementation.

It's an IO operation for both snapshot and log fetches, seconds delay is OK. In practice, log fetch is unlikely to happen and seconds delay is even unlikely to happen.

Connor1996:
> > It may cost several hundred milliseconds, even seconds for one fetch log task. You can't keep the execution slice small based on current raft engine implementation.
> 
> It's an IO operation for both snapshot and log fetches, seconds delay is OK. In practice, log fetch is unlikely to happen and seconds delay is even unlikely to happen.

But once it happens, a lot of tasks are triggered from different tasks. The number of pending tasks may keep at a high level for a while. And we currently have the mechanism to prefill cache before transfer-leader, it's likely to happen often.

BusyJay:
> But once it happens, a lot of tasks are triggered from different tasks. The number of pending tasks may keep at a high level for a while. And we currently have the mechanism to prefill cache before transfer-leader, it's likely to happen often.

Let's find out whether seconds delay really matters in real world tests after all are implemented. It's always an easy job to offload the snapshot to a dedicate thread if necessary.

Connor1996:
> > But once it happens, a lot of tasks are triggered from different tasks. The number of pending tasks may keep at a high level for a while. And we currently have the mechanism to prefill cache before transfer-leader, it's likely to happen often.
> 
> Let's find out whether seconds delay really matters in real world tests after all are implemented. It's always an easy job to offload the snapshot to a dedicate thread if necessary.

Then use yatp and assign the task high priority, otherwise the delay is minutes rather seconds.

nolouch:
@Connor1996  I will try to replace it with `yatp` later. ptal current part.

Connor1996 (approved):
LGTM

Issue #13060, opened 2022-07-20
Enhance fetch raft log policy to aviod QPS spike


Connor1996:
## Development Task

When there is a restart or raft entry cache is dropped for some reason, TiKV may suffer from QPS spike because the raft entry needs to fetch from disk which slows down the duration of replicating raft log and commit log. After enabling raft engine by default, it's worse since raft engine's read performance is not as good as rocksdb. 

To alleviate these kinds of issues, here are some optimizations can do:
- [ ] better entry cache GC policy(need consider all members including learner)
- [ ] make async fetch worker multi-thread
- [ ] remove the high bound of async fetch to reduce the possibility of invalid fetch
- [ ] fetch raft entry to fill the cache before accepting transferring leader
- [ ] enhance raft engine read performance

PR #13556, opened 2022-09-28
raftstore: warm up entry cache before leadership transfer

cosven:
Signed-off-by: cosven <yinshaowen241@gmail.com>

Issue Number: ref #13060

What's Changed:

```commit-message
Warm up the entry cache before becoming leader to avoid QPS spike.
```

### Related changes

### Check List 

Tests 

- [x] Unit test
- [x] Integration test
- No code

The left result (in green) is tested with patched tiup([pr](https://github.com/pingcap/tiup/pull/2051)) + patched tikv(this pr)
image

Side effects

### Release note 

```release-note
Warm up the entry cache before becoming leader to avoid QPS spike.
```


components/raftstore/src/store/peer.rs
```diff
@@ -4513,33 +4570,133 @@ where
...
+        // 1. During rolling upgrade, old instances may not support warmup.
+        // 2. The leader's entry cache is empty.
+        if warmup_range_start == 0 {
+            WARM_UP_ENTRY_CACHE_COUNTER.no_need.inc();
+            should_ack_now = true;
+        } else if warmup_range_start > last_index {
+            // warmup range start should not be large than the last index.
+            // Since the index is leader's entry_cache_first_index, so this is
+            // unlikely to happen.
+            WARM_UP_ENTRY_CACHE_COUNTER.invalid_index.inc();
```
tonyxuqqi, line 4634: I think we should not use metrics for the potential bug. If it's unlikely to happen, just add an error log

components/raftstore/src/store/peer.rs
```diff
@@ -4513,33 +4570,133 @@ where
...
+            );
+            should_ack_now = true;
+        } else {
+            if warmup_range_start < self.last_compacted_idx {
+                low = self.last_compacted_idx
+            };
+            // Check if the entry cache is already warmed up.
+            if let Some(first_index) = self.get_store().entry_cache_first_index() {
+                if low >= first_index {
+                    WARM_UP_ENTRY_CACHE_COUNTER.already.inc();
```
tonyxuqqi, line 4650: I think the metrics is too verbose..what's the purpose of "already" counter?

components/raftstore/src/store/peer.rs
```diff
@@ -4513,33 +4570,133 @@ where
...
+        let last_index = self.get_store().last_index();
+        let mut should_ack_now = false;
+        let mut low = warmup_range_start;
+
+        // Need not to warm up when the index is 0.
+        // There are two cases where index can be 0:
+        // 1. During rolling upgrade, old instances may not support warmup.
+        // 2. The leader's entry cache is empty.
+        if warmup_range_start == 0 {
+            WARM_UP_ENTRY_CACHE_COUNTER.no_need.inc();
```
tonyxuqqi, line 4628: counter no_need seems not necessary

components/raftstore/src/store/config.rs
```diff
@@ -299,6 +299,15 @@ pub struct Config {
...
     pub long_uncommitted_base_threshold: ReadableDuration,
 
+    /// Interval to check pre become leader.
+    #[doc(hidden)]
+    pub pre_become_leader_state_tick_interval: ReadableDuration,
+    #[doc(hidden)]
+    pub exit_pre_become_leader_state_ticks: usize,
+    /// Max tick count for the entry cache to be warmed up
+    /// before a follower becomes leader. Set it to 0 to disable warmup.
+    pub warm_up_raft_entry_cache_ticks: usize,
```
tonyxuqqi, line 309: what's the difference between exit_pre_become_leader_state_ticks and warm_up_raft_entry_cache_ticks?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1836,25 +1837,90 @@ where
...
+            || self.fsm.peer.pending_remove
+        {
             self.fsm.peer.mut_store().clean_async_fetch_res(low);
             return;
         }
 
         if self.fsm.peer.term() != res.term {
             // term has changed, the result may be not correct.
             self.fsm.peer.mut_store().clean_async_fetch_res(low);
+        } else if !self.fsm.peer.is_leader() {
```
tonyxuqqi, line 1852: I think checking self.fsm.peer.pre_become_leader_state.is_some() is better. 
And also the following block should be wrapped in a func.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -1836,25 +1837,90 @@ where
...
+                    .update_async_fetch_res(low, Some(res));
+                self.fsm
+                    .peer
+                    .mut_store()
+                    .on_warmup_range_fetched(entry_cache_warmup_range);
+                // Clean the async fetch result immediately if it is not used.
+                self.fsm.peer.mut_store().update_async_fetch_res(low, None);
+            }
+            // Need not to ack here if warmup is timeout.
+            if !is_entry_cache_warmup_timeout {
```
tonyxuqqi, line 1911: I think if it's timeout, we should ack transfer leader anyway. Otherwise the transfer leader is greatly delayed?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5306,6 +5394,28 @@ where
...
+            if state.ticks >= self.ctx.cfg.exit_pre_become_leader_state_ticks {
+                self.fsm.peer.pre_become_leader_state = None;
+                return;
+            }
+
+            if state.ticks >= self.ctx.cfg.warm_up_raft_entry_cache_ticks
+                && !state.is_entry_cache_warmup_timeout()
+            {
+                WARM_UP_ENTRY_CACHE_COUNTER.timeout.inc();
+                self.fsm.peer.ack_transfer_leader_msg(false);
```
tonyxuqqi, line 5413: I think this tick's purpose is to ensure ack_transfer_leader_msg can be called within a timeout window.  Once ack_transfer_leader_msg is called, this tick should not be needed anymore.  And I think exit_pre_become_leader_state_ticks and warm_up_raft_entry_cache_ticks should be same thing.

tonyxuqqi:
The code seems a bit more complicated than its necessary. For example, some metrics are not necessary. And also configuration may be simplified.

components/raftstore/src/store/config.rs
cosven: 1. `warm_up_raft_entry_cache_ticks` refers to the timeout of the log fetching operation.
2. `exit_pre_become_leader_state_ticks` refers to the timeout of the leadership transfer operation. When entries are fetched into cache, we should ensure they are not compacted before it becomes leader.

TonsnakeLin: 1. what's the unit of `warm_up_raft_entry_cache_ticks` and `exit_pre_become_leader_state_ticks`, counts or time? 
2. Could you add the comments for `exit_pre_become_leader_state_ticks`?

cosven: I'm working on code refactoring to simplify these configurations.

components/raftstore/src/store/entry_storage.rs
```diff
@@ -147,6 +147,48 @@ impl EntryCache {
...
+        if let Some(entry_last_index) = entries.last().map(|e| e.get_index()) {
+            // Check if the entries are valid to append.
+            let mut is_valid = true;
+            if let Some(first_index) = self.first_index() {
+                if first_index != entry_last_index + 1 {
+                    is_valid = false;
+                }
+            } else {
+                // The entry_last_index should be equal to the last index,
+                // which also means it must be large than the persisted index.
```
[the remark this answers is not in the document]
cosven: fixed

components/raftstore/src/store/config.rs
cosven: Currently, there is only one configuration item.

components/raftstore/src/store/fsm/peer.rs
cosven: fixed. Currently, there is only one configuraion item and it should be more clear.

components/raftstore/src/store/peer.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
```diff
@@ -147,6 +147,48 @@ impl EntryCache {
         }
     }
 
+    /// Append entries to the left.
+    ///
+    /// Index of the last entry in entries should exactly equal to
+    ///    `start index of the cached entries - 1` when cache is not empty.
+    ///
+    /// Return true when entries are appended successfully.
+    fn appendleft(&mut self, entries: &[Entry]) -> bool {
```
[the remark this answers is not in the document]
cosven: fixed

components/raftstore/src/store/entry_storage.rs
```diff
@@ -147,6 +147,48 @@ impl EntryCache {
         }
     }
 
+    /// Append entries to the left.
```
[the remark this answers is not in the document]
cosven: fixed

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,6 +1022,69 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
+
+    pub fn on_warmup_range_fetched(&mut self, range: (u64, u64)) {
+        WARM_UP_ENTRY_CACHE_COUNTER.finished.inc();
+        let (low, high) = range;
+        match self.entries(low, high, u64::MAX, GetEntriesContext::empty(true)) {
+            Ok(entries) => {
+                let ok = self.cache.appendleft(&entries);
+                debug_assert!(ok);
+                if !ok {
+                    WARM_UP_ENTRY_CACHE_COUNTER.unexpected_err.inc();
```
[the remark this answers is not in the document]
cosven: fixed

components/raftstore/src/store/peer.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
```diff
@@ -147,6 +147,48 @@ impl EntryCache {
...
+    ///    `start index of the cached entries - 1` when cache is not empty.
+    ///
+    /// Return true when entries are appended successfully.
+    fn appendleft(&mut self, entries: &[Entry]) -> bool {
+        if let Some(entry_last_index) = entries.last().map(|e| e.get_index()) {
+            // Check if the entries are valid to append.
+            let mut is_valid = true;
+            if let Some(first_index) = self.first_index() {
+                if first_index != entry_last_index + 1 {
+                    is_valid = false;
```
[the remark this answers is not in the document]
cosven: fixed. 

The first index should always be equal to `entry_last_index + 1`. Add an assertion here.

components/raftstore/src/store/config.rs
```diff
@@ -299,6 +299,10 @@ pub struct Config {
     #[doc(hidden)]
     pub long_uncommitted_base_threshold: ReadableDuration,
 
+    /// Max time for the entry cache to be warmed up.
```
tonyxuqqi, line 302: minor: Max duration  for warming up entry cache. The original comment can be misunderstood as the max time (count) a specific entry cache can be warmed up.

cosven:
One test case fails and it is not related to this PR. I have submitted another PR to fix it https://github.com/tikv/tikv/pull/13602 .

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,8 +1060,127 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
+        match &res.ents {
+            Ok(entries) => {
+                let last_entry_index = entries.last().map(|e| e.index);
+                if let Some(index) = last_entry_index {
+                    // Generally speaking, when the res.low is the same as the warmup
+                    // range start, the fetch result is exactly used for warmup.
+                    // As the low index of each async_fetch task is different.
+                    // There should exist only one exception. A async fetch task
+                    // with same low index is triggered before the warmup task.
+                    if index + 1 == range.1 {
```
Connor1996, line 1139: I think we don't need to match the high index exactly. It's okay for the entries to be part of warmup state's range, as the warmup state's range may be changed during async fetching.

components/raftstore/src/store/entry_storage.rs
```diff
@@ -147,6 +149,25 @@ impl EntryCache {
         }
     }
 
+    /// Push entries to the left of the cache.
+    ///
+    /// When cache is not empty, the index of the last entry in entries
+    /// should be equal to `cache first index - 1`. When cache is
+    /// empty, it should be equal to the store's last index. Otherwise,
+    /// append new entries may fail due to unexpected hole.
+    fn push_front(&mut self, entries: &[Entry]) {
```
Connor1996, line 158: how about calling it `prepend`

components/raftstore/src/store/entry_storage.rs
```diff
@@ -530,6 +551,63 @@ pub fn init_applied_term<ER: RaftEngine>(
...
+
+    pub fn mark_task_timeout(&mut self) {
+        WARM_UP_ENTRY_CACHE_COUNTER.timeout.inc();
+        self.is_task_timeout = true;
+    }
+
+    /// Whether the state is stale.
+    ///
+    /// Exit the state when the state is stale.
+    pub fn is_stale(&self) -> bool {
```
Connor1996, line 599: unify it with `is_task_timeout` like
```rust
pub fn check_timeout(&mut self, duration: Duration) -> bool {
    if self.started_at.elapsed() > duration {
        self.timeout = true;
    }
    self.timeout
}
```

components/raftstore/src/store/peer.rs
```diff
@@ -3850,10 +3852,12 @@ where
         // Broadcast heartbeat to make sure followers commit the entries immediately.
         // It's only necessary to ping the target peer, but ping all for simplicity.
         self.raft_group.ping();
+
         let mut msg = eraftpb::Message::new();
         msg.set_to(peer.get_id());
         msg.set_msg_type(eraftpb::MessageType::MsgTransferLeader);
         msg.set_from(self.peer_id());
+        msg.set_index(self.get_store().entry_cache_first_index().unwrap_or(0));
```
Connor1996, line 3860: If there is a lagging peer which keeps the cache from being dropped. Then transfer leader would be slow to fetch so many entries. How about the index that has been replicted to quorum?

components/raftstore/src/store/entry_storage.rs
tonyxuqqi: +1
push_front seems more used on single entry's push.

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,8 +1060,127 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
 
+    pub fn entry_cache_warmup_state(&self) -> &Option<CacheWarmupState> {
+        &self.cache_warmup_state
+    }
+
+    pub fn entry_cache_warmup_state_mut(&mut self) -> &mut Option<CacheWarmupState> {
+        &mut self.cache_warmup_state
+    }
+
+    pub fn exit_entry_cache_warmup_state(&mut self) {
```
tonyxuqqi, line 1071: maybe call clear_entry_cache_warmup_state is better?

components/raftstore/src/store/peer.rs
```diff
@@ -4513,33 +4517,96 @@ where
...
+
+        if should_ack_now {
+            return true;
+        }
+
+        // Check if the warmup operation is timeout if warmup is already started.
+        if let Some(state) = self.mut_store().entry_cache_warmup_state_mut() {
+            // If it is timeout, this peer should ack the message so that
+            // the leadership transfer process can continue.
+            let is_task_timeout = state.elapsed() >= ctx.cfg.max_raft_entry_cache_warmup_time.0;
```
tonyxuqqi, line 4596: As bokang mentioned above, here the three lines can be merged into one fn check_task_timeout().

components/raftstore/src/store/peer.rs
tonyxuqqi: > If there is a lagging peer which keeps the cache from being dropped. Then transfer leader would be slow to fetch so many entries. How about the index that has been replicted to quorum?

It has the timeout mechanism so that the transfer leader delay will be capped. Using the quorum's replicated index may defeat the purpose because the leader may still try to replicate logs to the lagging one if it's alive.

components/raftstore/src/store/peer.rs
```diff
@@ -4513,33 +4517,96 @@ where
...
+    /// This return whether the msg should be acked. When cache is warmed up
+    /// or the warmup operation is timeout, it is true.
+    pub fn pre_ack_transfer_leader_msg<T>(
+        &mut self,
+        ctx: &mut PollContext<EK, ER, T>,
+        msg: &eraftpb::Message,
+    ) -> bool {
+        // The start index of warmup range is leader's entry_cache_start_index,
+        // and it is equal to the lowest matched index in general.
+        let warmup_range_start = msg.get_index();
```
tonyxuqqi, line 4560: it seems to me no need to have warmup_range_start and mut low. They can be one mut low

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3212,10 +3234,17 @@ where
...
-            self.fsm
-                .peer
-                .execute_transfer_leader(self.ctx, msg.get_from(), peer_disk_usage, false);
+        } else if !self
+            .fsm
+            .peer
+            .maybe_reject_transfer_leader_msg(self.ctx, msg, peer_disk_usage)
+        {
+            // If max warmup time is 0, then skip the warmup process and ack directly.
+            if self.ctx.cfg.max_raft_entry_cache_warmup_time.0 == Duration::from_secs(0)
```
tonyxuqqi, line 3243: it can be wrapped into a inline "fn pre_warmup_entry_cache_enabled() -> bool" for better readability.

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,8 +1060,127 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
+    /// Trigger a task to warm up the entry cache.
+    ///
+    /// This will ensure the range [low..high..last_index] are loaded into
+    /// cache. Return the high index of the warmup range if a task is
+    /// successfully triggered.
+    pub fn async_warm_up_entry_cache(&mut self, low: u64) -> Option<u64> {
+        let high = if let Some(first_index) = self.entry_cache_first_index() {
+            if low >= first_index {
+                // Already warmed up.
+                self.cache_warmup_state = Some(CacheWarmupState::new());
```
tonyxuqqi, line 1084: why cache_warmup_state needs to be set as it will be cleared anyway after the function returns.

tonyxuqqi (commented):
overall it looks good to me. There're a few minor comments.

components/raftstore/src/store/peer.rs
Connor1996: Yes, but the leader can replicate logs to the lagging one by the previous async fetch mechanism. No need to fetch all at once when transferring leader.

cosven: > How about the index that has been replicted to quorum?

I think this is not enough. Assume there are 3 instances A, B, C. A and B has latest logs, during rolling upgrade, when A or B is restarted, it needs to wait for C to catch up.

Connor1996: reasonable

components/raftstore/src/store/config.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
cosven: fixed

components/raftstore/src/store/peer.rs
cosven: fixed

components/raftstore/src/store/fsm/peer.rs
cosven: fixed

components/raftstore/src/store/peer.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
```diff
@@ -583,14 +583,18 @@ impl CacheWarmupState {
...
     pub fn is_task_timeout(&self) -> bool {
         self.is_task_timeout
     }
 
-    pub fn mark_task_timeout(&mut self) {
-        WARM_UP_ENTRY_CACHE_COUNTER.timeout.inc();
-        self.is_task_timeout = true;
+    /// Check if the task is timeout.
+    pub fn check_task_timeout(&mut self, duration: Duration) -> bool {
+        if self.elapsed() > duration {
```
Connor1996, line 593: does early return
```
if self.is_task_timeout { return true; }
```

components/raftstore/src/store/entry_storage.rs
cosven: The state is not cleared after the function returns. It is only cleared in two cases
1. the peer becomes leader
2. the state becomes stale

components/raftstore/src/store/entry_storage.rs
```diff
@@ -583,14 +583,18 @@ impl CacheWarmupState {
...
+            WARM_UP_ENTRY_CACHE_COUNTER.timeout.inc();
+            self.is_task_timeout = true;
+        }
+        self.is_task_timeout
     }
 
     /// Whether the state is stale.
```
Connor1996, line 600: use `check_task_timeout` to replace `is_stale` as well

components/raftstore/src/store/entry_storage.rs
cosven: I changed the `mark_task_timeout` to `check_task_timeout` and I think it is reasonable.For `is_stale`, I'm not sure if it should be also changed to `check_stale`, and I think it brings little benefit 🤔.

PTAL again ❤️

components/raftstore/src/store/entry_storage.rs
cosven: `check_stale` is added

components/raftstore/src/store/entry_storage.rs
cosven: fixed

Connor1996 (commented):
rest LGTM

components/raftstore/src/store/entry_storage.rs
```diff
@@ -155,7 +155,7 @@ impl EntryCache {
     /// should be equal to `cache first index - 1`. When cache is
     /// empty, it should be equal to the store's last index. Otherwise,
     /// append new entries may fail due to unexpected hole.
-    fn prepend(&mut self, entries: &[Entry]) {
+    fn prepend(&mut self, entries: Vec<Entry>) {
         let mut mem_size_change = 0;
         let old_capacity = self.cache.capacity();
         for e in entries.iter().rev() {
```
Connor1996, line 161:
```suggestion
        for e in entries.into_iter().rev() {
```

Connor1996 (approved):
LGTM

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,8 +1073,128 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
+            first_index
+        } else {
+            self.last_index() + 1
+        };
+
+        // Fetch entries [low, high) to trigger an async fetch task in background.
+        self.cache_warmup_state = Some(CacheWarmupState::new_with_range(low, high));
+        match self.entries(low, high, u64::MAX, GetEntriesContext::empty(true)) {
+            Ok(_) => {
+                // This should not happen, but it's OK :)
```
tabokie, line 1110: debug_assert?

tests/integrations/config/test-custom.toml
```diff
@@ -668,4 +669,4 @@ sample-num = 20
 sample-threshold = 100
 byte-threshold = 31457280
 split.split-balance-score = 0.25
-split.split-contained-score = 0.5
+split.split-contained-score = 0.5
```
tabokie, line 672: add a newline at the end.

components/raftstore/src/store/peer.rs
```diff
@@ -4560,10 +4622,23 @@ where
...
-    /// 3. ready_to_transfer_leader on leader:
+    /// 2. pre_ack_transfer_leader_msg on follower:
+    ///     If follower passes all necessary checks, it will try to warmup
+    ///     the entry cache.
+    /// 3. ack_transfer_leader_msg on follower:
+    ///     When the entry cache has been warmed up or the operator is timeout,
+    ///     the follower reply an ACK with type MsgTransferLeader and
+    ///     its promised persistent index.
+    ///
+    /// (Condition: when there are remaining pessimistic locks to propose)
```
tabokie, line 4633: A little confusing,
```suggestion
    /// Special notes for behavior when there are remaining pessimistic locks to propose
    ///    /*explain where and how do we detect this condition*/
```
And isn't `ready_to_transfer_leader` called after `ack_transfer_leader_msg`? Why explain it first?

components/raftstore/src/store/entry_storage.rs
```diff
@@ -530,6 +551,76 @@ pub fn init_applied_term<ER: RaftEngine>(
...
+        self.started_at.saturating_elapsed()
+    }
+
+    /// Whether the warmup task is already timeout.
+    pub fn is_task_timeout(&self) -> bool {
+        self.is_task_timeout
+    }
+
+    /// Check whether the task is timeout.
+    pub fn check_task_timeout(&mut self, duration: Duration) -> bool {
```
tabokie, line 594: We should hardcode the duration here in this function. Otherwise you cannot defend code patterns that use different durations and cause inconsistency.

components/raftstore/src/store/peer.rs
cosven: `ready_to_transfer_leader` behaves differently in different context (conditions).

components/raftstore/src/store/entry_storage.rs
cosven: Do you mean to store the duration of timeout in the struct like the following?

```rust
pub struct CacheWarmupState {
    ...
    task_deadline: Duration,
    is_task_timeout: bool,
}

impl CacheWarmupState {
    fn check_task_timeout(&mut self) {
        if self.is_task_timeout {
            return true;
        }
        if self.elapsed() > self.task_deadline {
            WARM_UP_ENTRY_CACHE_COUNTER.timeout.inc();
            self.is_task_timeout = true;
        }
        self.is_task_timeout
    }
}
```

tabokie: Oh, I was mainly thinking about `check_stale`. How about this:
```
struct State {
  cached_elapsed_duration: Duration
}

impl State {
  fn check_timeout(&mut self, timeout: Duration) {
    if self.cached_elapsed_duration > timeout { return true; }
    self.cached_elapsed_duration = self.start.elapsed();
    self.cached_elapsed_duration > timeout
  }
}
```

cosven: >  I was mainly thinking about `check_stale`

For `check_stale`, I personally prefer the following solution to check if the state is stale or not  
```rust
impl State {
    pub fn is_stale(&self) -> bool {
        self.elapsed() >= MAX_WARMED_UP_CACHE_KEEP_TIME
    }
}
```

cosven: > Oh, I was mainly thinking about check_stale.

>We should hardcode the duration here in this function. Otherwise you cannot defend code patterns that use different durations and cause inconsistency.

Considering the two comments above, I have another idea
```rust
struct State {
    is_stale: bool,
}
impl State {
  fn check_stale(&mut self) {
        if self.is_stale {
            return true;
        }
        if self.elapsed() > MAX_WARMED_UP_CACHE_KEEP_TIME {
            self.is_stale = true;
        }
        self.is_stale
  }
}
```

tabokie: I'm fine with it, but `check_task_timeout` still bothers me. My previous comment seems to fix both functions at the same time.

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,8 +1073,128 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
+        &mut self.cache_warmup_state
+    }
+
+    pub fn clear_entry_cache_warmup_state(&mut self) {
+        self.cache_warmup_state = None;
+    }
+
+    /// Trigger a task to warm up the entry cache.
+    ///
+    /// This will ensure the range [low..high..last_index] are loaded into
```
tabokie, line 1090:
```suggestion
    /// This will ensure the range [low..=last_index] are loaded into
```

components/raftstore/src/store/entry_storage.rs
```diff
@@ -980,8 +1073,128 @@ impl<ER: RaftEngine> EntryStorage<ER> {
...
+    }
+
     pub fn compact_entry_cache(&mut self, idx: u64) {
-        self.cache.compact_to(idx);
+        let mut can_compact = true;
+        if let Some(state) = self.entry_cache_warmup_state_mut() {
+            if state.check_stale(MAX_WARMED_UP_CACHE_KEEP_TIME) {
+                self.clear_entry_cache_warmup_state();
+            } else {
+                can_compact = false;
```
tabokie, line 1192: This looks suspicious (in triggering OOM). Is it really needed?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -3212,10 +3235,14 @@ where
...
-        } else {
-            self.fsm
-                .peer
-                .execute_transfer_leader(self.ctx, msg.get_from(), peer_disk_usage, false);
+        } else if !self
+            .fsm
+            .peer
+            .maybe_reject_transfer_leader_msg(self.ctx, msg, peer_disk_usage)
+            && (!self.ctx.cfg.warmup_entry_cache_enabled()
+                || self.fsm.peer.pre_ack_transfer_leader_msg(self.ctx, msg))
```
tabokie, line 3243: Maybe check for `warmup_entry_cache_enabled` inside `pre_ack_transfer_leader_msg`. As in the comment, `pre_ack_transfer_leader_msg` is a must-do item before transfer leader.

tabokie (commented):
Rest LG

components/raftstore/src/store/entry_storage.rs
cosven: Before, a peer loads entries into cache after it becomes leader. Now, entires are pre-fetched for a peer which is about to become leader. So in theory, the memory consumption should be similar.

If a peer has entry_cache_warmup_state, it means this peer is ready to become leader and it has a great possibility to become leader.

components/raftstore/src/store/fsm/peer.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
cosven: fixed

components/raftstore/src/store/entry_storage.rs
cosven: fixed

tests/integrations/config/test-custom.toml
cosven: fixed

components/raftstore/src/store/entry_storage.rs
tabokie: > loads entries into cache after it becomes leader

Could you point me to the code? I presume it's async? i.e. can it still become leader if entries are not all in cache.

cosven: > > loads entries into cache after it becomes leader
> 
> Could you point me to the code? I presume it's async? i.e. can it still become leader if entries are not all in cache.

When a leader replicates it's entries to other followers which has lagging logs, it fetch entries from raftdb (by `async_fetch`) and load them into entry_cache. 
https://github.com/tikv/tikv/blob/7fbbcdcee6f7381a9c048030c8f95fac055c42bb/components/raftstore/src/store/entry_storage.rs#L843-L844

tabokie: Ignore me, as long as `evict_entry_cache` is not changed, the OOM situation should be fine.

components/raftstore/src/store/peer.rs
```diff
@@ -4560,10 +4626,23 @@ where
...
-    /// 3. ready_to_transfer_leader on leader:
+    /// 2. pre_ack_transfer_leader_msg on follower:
+    ///     If follower passes all necessary checks, it will try to warmup
+    ///     the entry cache.
+    /// 3. ack_transfer_leader_msg on follower:
+    ///     When the entry cache has been warmed up or the operator is timeout,
+    ///     the follower reply an ACK with type MsgTransferLeader and
+    ///     its promised persistent index.
+    ///
+    /// Additional steps for behavior when there are remaining pessimistic
```
tabokie, line 4637:
```suggestion
    /// Additional steps when there are remaining pessimistic
```

tabokie (approved):

PR #13557, opened 2022-09-29
*: introduce the two-phase kv_flashback_to_version

### LINKED ISSUE #13519: Make FlashbackToVersion become a two-phase request

JmPotato:

Issue Number: Close #13519.

What's Changed:

```commit-message
Make `FlashbackToVersion` become a two-phase request as described in #13519.
```

### Related changes

- Need to merge #13541 before this PR first.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -21,7 +21,7 @@ use collections::{HashMap, HashSet};
 use engine_traits::{Engines, KvEngine, RaftEngine, SstMetaInfo, WriteBatchExt, CF_LOCK, CF_RAFT};
 use error_code::ErrorCodeExt;
 use fail::fail_point;
-use futures::channel::mpsc::UnboundedSender;
+use futures::channel::{mpsc::UnboundedSender, oneshot::Sender};
 use keys::{self, enc_end_key, enc_start_key};
 use kvproto::{
     brpb::CheckAdminResponse,
@@ -986,6 +986,12 @@ where
         syncer.report_for_self(self_report);
     }
 
+    // Check if the region is in the flashback state.
+    fn on_check_flashback(&mut self, ch: Sender<bool>) {
+        // TODO: maybe should check the persistent state also.
+        ch.send(self.fsm.peer.is_in_flashback).unwrap();
+    }
+
     fn on_check_pending_admin(&mut self, ch: UnboundedSender<CheckAdminResponse>) {
         if !self.fsm.peer.is_leader() {
             // no need to check non-leader pending conf change.
@@ -1436,6 +1442,7 @@ where
                 self.on_snapshot_recovery_wait_apply(syncer)
             }
             SignificantMsg::CheckPendingAdmin(ch) => self.on_check_pending_admin(ch),
+            SignificantMsg::CheckFlashback(ch) => self.on_check_flashback(ch),
         }
     }
 
--- a/components/raftstore/src/store/fsm/peer.rs
+++ b/components/raftstore/src/store/fsm/peer.rs
@@ -988,8 +988,22 @@ where
 
     // Check if the region is in the flashback state.
     fn on_check_flashback(&mut self, ch: Sender<bool>) {
-        // TODO: maybe should check the persistent state also.
-        ch.send(self.fsm.peer.is_in_flashback).unwrap();
+        let mut is_in_flashback = self.fsm.peer.is_in_flashback;
+        // Read the region local state to check whether the region is in the flashback
+        // state when the memory flag is false.
+        if !is_in_flashback {
+            let state_key = keys::region_state_key(self.region_id());
+            if let Ok(region_local_state) = self
+                .ctx
+                .engines
+                .kv
+                .get_msg_cf::<RegionLocalState>(CF_RAFT, &state_key)
+            {
+                is_in_flashback = region_local_state
+                    .map_or(false, |state| state.get_region().get_is_in_flashback());
+            }
+        }
+        ch.send(is_in_flashback).unwrap();
     }
 
     fn on_check_pending_admin(&mut self, ch: UnboundedSender<CheckAdminResponse>) {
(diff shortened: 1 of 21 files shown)


tests/integrations/server/kv_service.rs
```diff
@@ -682,16 +672,7 @@ fn test_mvcc_flashback_block_rw() {
...
-    flashback_to_version_req.set_context(ctx.clone());
-    flashback_to_version_req.set_start_ts(1);
-    flashback_to_version_req.set_commit_ts(2);
-    flashback_to_version_req.version = 0;
-    flashback_to_version_req.start_key = b"a".to_vec();
-    flashback_to_version_req.end_key = b"z".to_vec();
-    let flashback_resp = client
-        .kv_flashback_to_version(&flashback_to_version_req)
-        .unwrap();
+    let flashback_resp = must_flashback_to_version(&client, ctx.clone(), 0, 1, 2);
```
HuSharp, line 675: I think it is better to be unified with the above ts in `test_mvcc_flashback`
```suggestion
    let flashback_resp = must_flashback_to_version(&client, ctx.clone(), 0, ts + 1, ts + 2);
```

JmPotato:
PTAL, thx!

/cc @HuSharp @Connor1996

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -986,6 +986,26 @@ where
         syncer.report_for_self(self_report);
     }
 
+    // Check if the region is in the flashback state.
+    fn on_check_flashback_state(&mut self, ch: Sender<bool>) {
+        let mut is_in_flashback = self.fsm.peer.is_in_flashback;
+        // Read the region local state to check whether the region is in the flashback
```
Connor1996, line 992: don't need, memory state should be always the latest state

JmPotato: Since the memory state is set by `on_ready_result`, it's possible that there is an interval between when apply completes and when the memory state is set 🤔.

Connor1996: The ready result must be handled before your flashback proposal is handled. Messages to one region's mailbox are serialized

JmPotato: Oh, I get it.

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -988,22 +988,7 @@ where
...
-                .engines
-                .kv
-                .get_msg_cf::<RegionLocalState>(CF_RAFT, &state_key)
-            {
-                is_in_flashback = region_local_state
-                    .map_or(false, |state| state.get_region().get_is_in_flashback());
-            }
-        }
-        ch.send(is_in_flashback).unwrap();
+        ch.send(self.fsm.peer.is_in_flashback).unwrap();
```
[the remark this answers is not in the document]
Connor1996: it's done by batches, I think it won't be too wasteful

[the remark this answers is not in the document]
JmPotato: Ok, let's keep it simple.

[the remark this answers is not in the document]
Connor1996: Furthermore in flashback state, read is also forbidden. So it would return an error when getting the snapshot before modification.

AndreMouche (approved):
LGTM

AndreMouche (approved):

PR #13558, opened 2022-09-29
engine_tirocks: add properties

BusyJay:
Issue Number: Ref #13058 

What's Changed:

```commit-message
Significant differences are:
- This PR uses codec components instead of tikv_util::codec.
- Files are re-arranged to make properties related logic better organized.
- An extra allocation is reduced by passing `UserCollectedProperties` directly instead of creating a new HashMap.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/engine_tirocks/src/properties/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+
+pub use self::{
+    mvcc::MvccPropertiesCollectorFactory,
+    range::{RangeProperties, RangePropertiesCollectorFactory},
+    table::{RocksTablePropertiesCollection, RocksUserCollectedProperties},
+    ttl::TtlPropertiesCollectorFactory,
+};
+
+#[derive(Clone, Debug, Default)]
+pub struct IndexHandle {
```
tabokie, line 30: This struct is really hard to understand. From the code I don't see how it interacts with "stored block".

BusyJay: The struct is copied from engine_rocks. It's named in the old days. Now I change it to `PropIndex`.

components/engine_tirocks/src/properties/range.rs
```diff
@@ -182,7 +182,7 @@ impl RangeProperties {
 impl From<SizeProperties> for RangeProperties {
     fn from(p: SizeProperties) -> RangeProperties {
         let mut res = RangeProperties::default();
-        for (key, size_handle) in p.index_handles.into_map() {
+        for (key, size_handle) in p.prop_indexes.into_map() {
             let range = RangeOffsets {
                 size: size_handle.offset,
```
tabokie, line 187: Why the offset is size?

components/engine_tirocks/src/properties/mvcc.rs
```diff
@@ -151,14 +151,14 @@ impl TablePropertiesCollector for MvccPropertiesCollector {
 
         // Add new row.
         if self.row_versions == 1 {
-            self.cur_index_handle.size += 1;
-            self.cur_index_handle.offset += 1;
-            if self.cur_index_handle.offset == 1
-                || self.cur_index_handle.size >= PROP_ROWS_INDEX_DISTANCE
+            self.cur_prop_index.prop += 1;
+            self.cur_prop_index.offset += 1;
```
tabokie, line 155: Does this mean the `prop` covers a range ends with `offset` (instead of starting from `offset`)

components/engine_tirocks/src/properties/range.rs
BusyJay: For `SizeProperties`, the `offset` is accumulation of the `size`. For `RangeOffsets`, `size` is the total bytes of raw KVs from the file start.

components/engine_tirocks/src/properties/mvcc.rs
BusyJay: Yes.

tabokie (approved):

Connor1996 (approved):
LGTM

PR #13565, opened 2022-09-29
raftstore-v2: add read handling after apply

tonyxuqqi:
And add renew lease in handle_raft_committed_entries.

Fix test case test_generate_bucket_mvcc_by_approximate

Issue Number: Ref #12842

What's Changed:

```commit-message
Renew lease if possible after write proposal committed.
Respond to read index if any pending reads after apply.
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore-v2/src/operation/query/mod.rs
```diff
@@ -377,4 +380,41 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+
+    // the v1's post_apply
+    // As the logic is mostly for read, rename it to handle_read_after_apply
+    pub fn handle_read_on_apply<T>(
+        &mut self,
+        ctx: &mut StoreContext<EK, ER, T>,
+        apply_res: ApplyRes,
+        progress_to_be_updated: bool,
+    ) -> bool {
+        let mut has_ready = false;
```
BusyJay, line 392: We don't need that. If it needs to be marked has ready, just call `set_has_ready` is OK. It's needed in the past because `has_ready` is part of `PeerFsm`.

BusyJay (approved):

ethercflow (approved):
LGTM

PR #13568, opened 2022-09-30
local_reader: solve the race condition when acquiring the snapshot

### LINKED ISSUE #13553: Batch snapshot is unsound
Nightly

Doesn't matter

### Steps to reproduce
Concurrent read and conf change may read incomplete data.

### What did happened?
https://github.com/tikv/tikv/pull/7878 introduces batch snapshot across regions. It reorders the snapshot operation and lease check operation, so snapshot is taken after checking lease. It's possible that after checking the lease and before snapshot is taken, the region is removed by conf change. So the snapshot may not contain the complete data.

To solve the problems, we need to make sure snapshot is taken before lease check.

/cc @hicqu

SpadeA-Tang:

Issue Number: Close #13553

What's Changed:

The current code encounters the case that after passing the lease check and before the acquiring the snapshot, the lease expires.
This PR fixes it by implementing `SnapCacheContext` which ensures the timestamp used for the lease check is generated before the snapshot acquisition.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


tests/failpoints/cases/test_local_read.rs
```diff
@@ -0,0 +1,59 @@
...
+    let mut req = new_request(
+        region_id,
+        region.get_region_epoch().clone(),
+        vec![new_snap_cmd()],
+        false,
+    );
+    req.mut_header().set_peer(leader);
+
+    let (cb, rx) = make_cb(&req);
+    let handle = thread::spawn(move || {
```
BusyJay, line 46: If using `new_server_cluster`, you don't need to spawn a new thread.

tests/failpoints/cases/test_local_read.rs
```diff
@@ -0,0 +1,59 @@
...
+        thread::sleep(Duration::from_millis(200));
+        // This is equivalent to expiring the lease
+        fail::cfg("local_reader_fail_lease_check", "return").unwrap();
+        fail::cfg("before_execute_get_snapshot", "off").unwrap();
+    });
+    cluster.sim.wl().async_read(node_id, None, req, cb);
+
+    // Request should be redirected to the raftstore, but the failpoint above makes
+    // it just return. So cb will not be called.
+    let _ = rx.recv_timeout(Duration::from_secs(5)).unwrap_err();
```
BusyJay, line 56: Should test on the result instead of code path. Code path can be changed, optimized, but the result should stay the same.

components/raftstore/src/store/worker/read.rs
```diff
@@ -763,20 +867,22 @@ where
...
                         };
-                        if !delegate.is_in_leader_lease(snapshot_ts) {
+
+                        let mut snap_cache_ctx =
+                            SnapCacheContext::new(&mut self.snap_cache, read_id);
+                        snap_cache_ctx.maybe_update_snapshot(delegate.get_tablet());
+
+                        let region = Arc::clone(&delegate.region);
+                        let response =
+                            delegate.execute(&req, &region, None, Some(&mut snap_cache_ctx));
```
BusyJay, line 877: Should check lease before executing requests.

tests/failpoints/cases/test_local_read.rs
SpadeA-Tang: Good suggestion.

tests/failpoints/cases/test_local_read.rs
SpadeA-Tang: Good catch.

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Good catch.

components/raftstore/src/store/worker/read.rs
```diff
@@ -208,12 +214,80 @@ where
...
 
 /// #[RaftstoreCommon]: LocalReadContext combines some LocalReader's fields for temporary usage.
-pub struct LocalReadContext<'a, E>
+pub struct SnapCacheContext<'a, E>
 where
     E: KvEngine,
 {
-    read_id: &'a mut ThreadReadId,
-    snap_cache: &'a mut Box<Option<Arc<E::Snapshot>>>,
+    // read_id: &'a mut ThreadReadId,
```
TonsnakeLin, line 221: The comments and code are not corresponded

SpadeA-Tang: Good catch. I forget to update it.

components/raftstore/src/store/worker/read.rs
```diff
@@ -207,13 +213,80 @@ where
...
+pub struct SnapshotContext<'a, E>
 where
     E: KvEngine,
 {
-    read_id: &'a mut ThreadReadId,
-    snap_cache: &'a mut Box<Option<Arc<E::Snapshot>>>,
+    read_id: Option<ThreadReadId>,
+    snap_cache: &'a mut SnapCache<E>,
+
+    snapshot: Option<Arc<E::Snapshot>>,
```
BusyJay, line 226: Why not always update `snap_cache`? What's the point of duplicated definition?

SpadeA-Tang: Good suggestion.

components/raftstore/src/store/worker/mod.rs
```diff
@@ -1,5 +1,4 @@
 // Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
-
```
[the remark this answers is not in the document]
SpadeA-Tang: By accident.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -88,8 +88,7 @@ where
 
     fn get_snapshot(
         &mut self,
-        _: Option<ThreadReadId>,
-        _: &mut Option<raftstore::store::LocalReadContext<'_, E>>,
+        _: &mut Option<&mut raftstore::store::LocalReadContext<'_, E>>,
```
BusyJay, line 91: Bad style. Should import it.

tests/failpoints/cases/test_local_read.rs
```diff
@@ -0,0 +1,81 @@
...
+            .async_remove_peer(region_id, leader)
+            .unwrap()
+            .recv()
+            .unwrap()
+            .get_header()
+            .has_error()
+    );
+
+    // Wait for data to be cleaned
+    thread::sleep(Duration::from_millis(2000));
```
BusyJay, line 77: How do you know data is cleaned?

components/raftstore/src/store/worker/read.rs
```diff
@@ -551,8 +646,10 @@ where
     // region id -> ReadDelegate
     // The use of `Arc` here is a workaround, see the comment at `get_delegate`
     pub delegates: LruCache<u64, D>,
-    snap_cache: Box<Option<Arc<E::Snapshot>>>,
-    cache_read_id: ThreadReadId,
+    // snap_cache: Box<Option<Arc<E::Snapshot>>>,
```
BusyJay, line 649: Why keeping these?

components/raftstore/src/store/worker/read.rs
```diff
@@ -207,13 +213,61 @@ where
...
+        Self {
+            snap_cache,
+            read_id,
+        }
+    }
+
+    // Update the snapshot in the `snap_cache` if the read_id is None or does not
+    // match. When the read_id is None, it ok to just update the snapshot without
+    // touching the `cached_read_id` because only a consecutive requests will
+    // have the same read_id, which means this `cached_read_id` will not be matched
```
BusyJay, line 238: > which means this `cached_read_id` will not be matched anymore

If that's the case, then clear `cached_read_id` seems also OK.

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: Done.

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Fine

components/raftstore/src/store/worker/read.rs
```diff
@@ -207,13 +213,66 @@ where
...
+    // match.
+    // When the read_id is None, it means the `snap_cache` has been cleared
+    // before and the `cached_read_id` of it is None because only a consecutive
+    // requests will have the same cache and the cache will be cleared after the
+    // last request of the batch.
+    fn maybe_update_snapshot(&mut self, engine: &E, delegate_last_valid_ts: Timespec) -> bool {
+        if let Some(read_id) = self.read_id.as_ref() {
+            if self.snap_cache.cached_read_id.is_some()
+                && *self.snap_cache.cached_read_id.as_ref().unwrap() == *read_id
+                && self.snap_cache.cached_snapshot_ts >= delegate_last_valid_ts
```
BusyJay, line 245: Should compare read_id's create time instead of snapshot time. For example, if the snapshot is taken before the peer is created, the check can pass, but no data can be read.

components/raftstore/src/store/worker/read.rs
```diff
@@ -207,13 +213,66 @@ where
...
+    }
+
+    // Update the snapshot in the `snap_cache` if the read_id is None or does not
+    // match.
+    // When the read_id is None, it means the `snap_cache` has been cleared
+    // before and the `cached_read_id` of it is None because only a consecutive
+    // requests will have the same cache and the cache will be cleared after the
+    // last request of the batch.
+    fn maybe_update_snapshot(&mut self, engine: &E, delegate_last_valid_ts: Timespec) -> bool {
+        if let Some(read_id) = self.read_id.as_ref() {
```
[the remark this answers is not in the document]
SpadeA-Tang: What if they are both None?

components/raftstore/src/store/worker/read.rs
```diff
@@ -93,8 +100,7 @@ pub trait ReadExecutor<E: KvEngine> {
         msg: &RaftCmdRequest,
         region: &Arc<metapb::Region>,
         read_index: Option<u64>,
-        mut ts: Option<ThreadReadId>,
-        mut read_context: Option<LocalReadContext<'_, E>>,
+        mut snap_cache_ctx: Option<&mut LocalReadContext<'_, E>>,
```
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

components/raftstore/src/store/worker/read.rs
```diff
@@ -207,13 +213,66 @@ where
...
+    E: KvEngine,
+{
+    fn new(snap_cache: &'a mut SnapCache<E>, read_id: Option<ThreadReadId>) -> Self {
+        Self {
+            snap_cache,
+            read_id,
+        }
+    }
+
+    // Update the snapshot in the `snap_cache` if the read_id is None or does not
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -86,11 +86,7 @@ where
         self.cached_tablet.latest().unwrap()
     }
 
-    fn get_snapshot(
-        &mut self,
-        _: Option<ThreadReadId>,
-        _: &mut Option<raftstore::store::LocalReadContext<'_, E>>,
-    ) -> Arc<E::Snapshot> {
+    fn get_snapshot(&mut self, _: &mut Option<&mut LocalReadContext<'_, E>>) -> Arc<E::Snapshot> {
```
TonsnakeLin, line 89: Why changed `&mut Option<LocalReadContext` to ` &mut Option<&mut LocalReadContext` ?

SpadeA-Tang: Good catch. It does not need to be mut now.

components/raftstore/src/store/worker/read.rs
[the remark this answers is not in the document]
tonyxuqqi: > What if they are both None?

The self.read_id.is_some()

components/raftstore/src/store/worker/read.rs
```diff
@@ -811,6 +879,13 @@ where
                     }
                     _ => unreachable!(),
                 };
+
+                TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().local_executed_requests.inc());
```
tonyxuqqi, line 883: I think these metrics update can be moved to maybe_update_snapshot.
Then no need to handle snap_updated or not

SpadeA-Tang: What if lease does not pass?

components/raftstore/src/store/worker/read.rs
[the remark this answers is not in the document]
SpadeA-Tang: My bad.

components/raftstore/src/store/worker/read.rs
tonyxuqqi: > What if lease does not pass?

OK.

tonyxuqqi (approved):

components/raftstore/src/store/worker/read.rs
```diff
@@ -6,6 +6,7 @@ use std::{
     fmt::{self, Display, Formatter},
     ops::Deref,
     sync::{
+        atomic,
```
BusyJay, line 9: Should be `atomic::{self, AtomicU64, Ordering}`.

components/raftstore/src/store/worker/read.rs
```diff
@@ -47,22 +48,32 @@ pub trait ReadExecutor {
     type Tablet: KvEngine;
 
     fn get_tablet(&mut self) -> &Self::Tablet;
+
+    /// Get the snapshot fo the tablet.
+    ///
+    /// If the tablet has not been ready, `None` is returned.
```
BusyJay, line 54:
```suggestion
    /// If the tablet is not ready, `None` is returned.
```

components/raftstore/src/store/worker/read.rs
```diff
@@ -187,13 +199,66 @@ where
...
+        // before and the `cached_read_id` of it is None because only a consecutive
+        // requests will have the same cache and the cache will be cleared after the
+        // last request of the batch.
+
+        if self.read_id.is_some() {
+            if self.snap_cache.cached_read_id == self.read_id
+                && self.read_id.as_ref().unwrap().create_time >= delegate_last_valid_ts
+            {
+                // Cache hit
+                assert!(self.snap_cache.snapshot.as_ref().is_some());
```
BusyJay, line 234: Useless assert.

components/raftstore/src/store/worker/read.rs
```diff
@@ -187,13 +199,66 @@ where
...
+                return false;
+            }
+
+            self.snap_cache.cached_read_id = self.read_id.clone();
+        } else {
+            // snap_cache should already have been cleared
+            assert!(self.snap_cache.cached_read_id.is_none());
```
BusyJay, line 241: It's OK if it's not cleared.

components/raftstore/src/store/worker/read.rs
```diff
@@ -566,6 +632,46 @@ impl Progress {
     }
 }
 
+struct SnapCache<E>
+where
+    E: KvEngine,
+{
+    cached_read_id: Option<ThreadReadId>,
+    snapshot: Box<Option<Arc<E::Snapshot>>>,
```
BusyJay, line 640: Why put it in `Box`?

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Good catch.

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: It is insider Box before. I think I can remove it.

components/raftstore/src/store/worker/read.rs
```diff
@@ -837,12 +923,14 @@ where
                             return;
                         }
 
-                        let read_ctx = self.local_read_context();
+                        let mut local_read_ctx =
+                            LocalReadContext::new(&mut self.snap_cache, read_id);
+                        snap_updated = local_read_ctx
+                            .maybe_update_snapshot(delegate.get_tablet(), last_valid_ts);
```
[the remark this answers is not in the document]
SpadeA-Tang: It is used before.
And does batch get command support stale read?

components/raftstore/src/store/worker/read.rs
```diff
@@ -1570,111 +1647,100 @@ mod tests {
         let mut delegate = delegate.unwrap();
         let tablet = delegate.get_tablet();
         assert_eq!(kv_engine.as_inner().path(), tablet.as_inner().path());
-        let snapshot = delegate.get_snapshot(read_id_copy, &mut read_context);
+        let snapshot = delegate.get_snapshot(&read_context).unwrap();
```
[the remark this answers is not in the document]
SpadeA-Tang: I think I can simplify this test to test read executor provider only.

components/raftstore/src/store/worker/read.rs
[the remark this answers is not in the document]
BusyJay: Not sure. Can you find out the answer?

components/raftstore/src/store/worker/read.rs
```diff
@@ -1546,135 +1627,274 @@ mod tests {
...
-        let snapshot = delegate.get_snapshot(read_id_copy, &mut read_context);
+    }
+
+    #[test]
+    fn test_snap_across_regions() {
+        let store_id = 2;
+        let store_meta = Arc::new(Mutex::new(StoreMeta::new(0)));
+        let (_tmp, mut reader, rx) = new_reader("test-local-reader", store_id, store_meta.clone());
+
+        fn prepare_read_delegate(
```
BusyJay, line 1651: Bad style, should avoid nested function.

tests/failpoints/cases/test_local_read.rs
```diff
@@ -0,0 +1,81 @@
...
+    get_req.set_context(ctx);
+    get_req.key = b"key1".to_vec();
+    let mut receiver = client.raw_get_async(&get_req).unwrap();
+
+    thread::sleep(Duration::from_millis(200));
+
+    let mut peer = leader.clone();
+    cluster.must_transfer_leader(1, new_peer(2, 2));
+    pd_client.must_remove_peer(region_id, leader);
+    peer.id += 1000;
```
BusyJay, line 67: Very confusing. If you just want to assign a new ID, then assign it directly instead of adding arbitrary number.

components/raftstore/src/store/worker/read.rs
```diff
@@ -1546,135 +1627,274 @@ mod tests {
...
+            ep.set_conf_ver(1);
+            ep.set_version(3);
+            ep
+        };
+        let term6 = 6;
+
+        // Register region1
+        let pr_ids1 = vec![2, 3, 4];
+        let prs1 = new_peers(store_id, pr_ids1.clone());
+        prepare_read_delegate(
```
BusyJay, line 1705: Why not just use `new_reader`?

components/raftstore/src/store/worker/read.rs
```diff
@@ -1546,135 +1627,274 @@ mod tests {
...
+                assert_eq!(snap.get_region().id, 2);
+                assert_eq!(
+                    snap.get_snapshot().get_value(b"a1").unwrap().unwrap(),
+                    b"val1"
+                );
+            })),
+        );
+        must_not_redirect_with_read_id(&mut reader, &rx, task, read_id);
+        assert_eq!(
+            TLS_LOCAL_READ_METRICS.with(|m| m.borrow_mut().local_executed_snapshot_cache_hit.get()),
```
BusyJay, line 1780: Or you can just check if `Arc` is the same by `Arc::ptr_eq`.

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Done

tests/failpoints/cases/test_local_read.rs
SpadeA-Tang: Done

components/raftstore/src/store/worker/read.rs
SpadeA-Tang: Done

BusyJay (approved):
Rest LGTM

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -386,8 +371,8 @@ impl<'r> SnapRequestInspector<'r> {
             return Ok(RequestPolicy::ReadIndex);
         }
 
-        // If applied index's term is different from current raft's term, leader
-        // transfer must happened, if read locally, we may read old value.
+        // If applied index's term is differ from current raft's term, leader transfer
```
tonyxuqqi, line 374: minor: is differ --> differs

components/raftstore/src/store/worker/read.rs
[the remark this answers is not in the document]
SpadeA-Tang: `batch_get_command` will set read_id be None for stale read so it will not use cache.

PR #13575, opened 2022-10-08
raftstore: run pre_persist when yield

CalvinNeo:

Issue Number: ref #12849

What's Changed:

In #12849, we introduced a observer `pre_persist` which is called before every `write_to_db` which suggests whether should we actually do a persistence or not.
However, we find this observer do not cover when a delegate yields when it is run in another priority. So we need to add this hook.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```

### DIFF
--- a/components/raftstore/src/store/fsm/apply.rs
+++ b/components/raftstore/src/store/fsm/apply.rs
@@ -617,7 +617,7 @@ where
         delegate: &mut ApplyDelegate<EK>,
         results: VecDeque<ExecResult<EK::Snapshot>>,
     ) {
-        if self.host.pre_persist(&delegate.region, true, None) {
+        if self.host.pre_persist(&delegate.region, true, false, None) {
             if !delegate.pending_remove {
                 delegate.write_apply_state(self.kv_wb_mut());
             }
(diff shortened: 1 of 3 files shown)


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1135,7 +1137,11 @@ where
                 has_unflushed_data = false;
             }
             if self.priority != apply_ctx.priority {
-                if has_unflushed_data {
+                if has_unflushed_data
+                    && apply_ctx
```
BusyJay, line 1141: When will it return false?

CalvinNeo: Always in TiFlash

BusyJay: TiKV share writebatch across peers. Not flushing the writebatch can reorder the apply. For example if a peer put(k1, v1) to writebatch 1, and put(k1, v2) to writebatch 2. The final value of k1 depends on which writebatch is written first.

CalvinNeo: In TiFlash‘s scenario, we guarantee there is no real KV data in WriteBatch. We hereby prevent a persist here to persist advanced apply state.

PR #13576, opened 2022-10-08
Makefile: Default enable SSE instructions in x86 arch

### LINKED ISSUE #13574: use uname -m instead of uname -p when detecting arch
## Development Task

We were using `uname -p` to check the arch:

https://github.com/tikv/tikv/blob/956610725039835557e7516828b069a44073c36d/Makefile#L85

However, `uname -p` is not portable and could be even `unknown` in some popular distributions, for example, debian 11.

Instead, we could use `uname -m`.

Possible `uname -m` values:

https://stackoverflow.com/questions/45125516/possible-values-for-uname-m
https://en.wikipedia.org/wiki/Uname

breezewish:
Signed-off-by: Wish <breezewish@outlook.com>

Issue Number: Close #13574

What's Changed:

In Debian 11 ARM, TiKV cannot compile, because in a variety of Linux distributions, `uname -p` is assigned to `unknown` ([ref](https://unix.stackexchange.com/questions/307955/uname-p-i-are-unknown)).

This PR changes to use `uname -m`, which should always work.

As there will be never unknowns, the enabling rule is more strict now: SSE is only enabled when arch **is** x86-like. In other arch, it is not enabled.

Additionally, for MacOS in Macbook, `uname -m` outputs:

- Macbook M1: arm64
- Macbook Intel: x86_64

### Check List 

Tests 

- No code

### Release note 

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,15 +81,21 @@ ROCKSDB_SYS_PORTABLE=0
 RUST_TEST_THREADS ?= 2
 endif
 
-# Disable SSE on ARM
-ifeq ($(shell uname -p),aarch64)
-ROCKSDB_SYS_SSE=0
+# Enable sse4.2 by default on x86 and x64
+ifeq ($(shell uname -m),x86_64)
+ROCKSDB_SYS_SSE=1
 endif
-ifeq ($(shell uname -p),arm)
-ROCKSDB_SYS_SSE=0
+ifeq ($(shell uname -m),x64)
+ROCKSDB_SYS_SSE=1
 endif
-ifeq ($(shell uname -p),arm64)
-ROCKSDB_SYS_SSE=0
+ifeq ($(shell uname -m),x86)
+ROCKSDB_SYS_SSE=1
+endif
+ifeq ($(shell uname -m),i386)
+ROCKSDB_SYS_SSE=1
+endif
+ifeq ($(shell uname -m),i686)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -81,25 +81,30 @@ ROCKSDB_SYS_PORTABLE=0
 RUST_TEST_THREADS ?= 2
 endif
 
-# Disable SSE on ARM
-ifeq ($(shell uname -p),aarch64)
-ROCKSDB_SYS_SSE=0
+# Enable sse4.2 by default on x86 and x64
+ifeq ($(shell uname -m),x86_64)
```
BusyJay, line 85: How about test by `cc -msse3 -dM -E - </dev/null|grep SSE`?

breezewish: Looks like this will be very similar to how RocksDB enables SSE intelligently in https://github.com/tikv/rocksdb/blob/ce0ea61aecd53ac02a0e7da2b166961c7d98650a/CMakeLists.txt#L333 (by looking at the instruction set availability instead of the platform). There is a check SSE script doing similar things, I'm not sure whether this is fine.

BusyJay: This is too ugly. How about utilizing `rust -vV`? If there is 86 in the host line, we can force enabling SSE.

breezewish: Good idea!

PR #13583, opened 2022-10-11
tikv_util: Add a priority queue that supports efficiently removing elements by key

MyonKeminta:

Issue Number: Ref #13298

What's Changed:

```commit-message
Add a priority queue based on binary heap, and additionally supports efficiently removing elements by key. 

This is useful for implementing the new lock waiting queue, where the requests waits in a queue and may be canceled (due to timeout or deadlock) at any time.
```

### Related changes

-

### Check List 

Tests 

- Unit test

Side effects

-

### Release note 

```release-note
None
```


MyonKeminta:
> Do we have to implement our own priority queue? Have we checked existing crates?

I've found several. Some of them looks heavy, and some of them have licenses that looks doubtful about if we can use. Thinking that writing a binary heap is not so hard, I wrote one in my spare time.

sticnarf:
I don't think about it carefully, but if we don't have equal priority (start_ts), is a `HashMap` (key -> start_ts) plus a `BTreeMap` (start_ts => entry) enough for our needs?

MyonKeminta:
For simplicity (to write correct code more easily), this implementation didn't try to get better performance by making use of unsafe code to process holes in the heap. I think the performance is enough. With 2000 elements in the heap, the performance of pushing and popping is about 1/3 of the std BinaryHeap.

```
test priority_queue::tests::bench_push_pop            ... bench:         122 ns/iter (+/- 23)
test priority_queue::tests::bench_std_push_pop        ... bench:          40 ns/iter (+/- 2)
```

MyonKeminta:
> I don't think about it carefully, but if we don't have equal priority (start_ts), is a `HashMap` (key -> start_ts) plus a `BTreeMap` (start_ts => entry) enough for our needs?

I'm afraid there isn't such a guarantee that there will never be duplicated start_ts on the same key (for example, in case there are network issues). Also, I don't think we will use start_ts as the priority forever.  
I think it's possible to provide such a uniqueness by other approach, for example, appending a unique id to the key. I chooses priority queue just because I think it looks reasonable here. How do you think?

MyonKeminta:
Thanks. After some research and reconsideration, I decide to use https://crates.io/crates/keyed_priority_queue instead.

PR #13584, opened 2022-10-11
storage/lock_manager: Avoid stale entries in the new lock waiting queue

MyonKeminta:

Issue Number: Ref #13298 
~~Requires: https://github.com/tikv/tikv/pull/13583~~

What's Changed:

```commit-message
Avoid stale entries in the new lock waiting queue.

This is done by making use of another implementation of the priority queue (instead of the std BinaryHeap) that supports efficiently removing element by key. So that when a lock-waiting request is canceled, the entry can be removed from the queue immediately, instead of waiting for the lazy-cleaning-up.
```

### Related changes

\-

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more MEM

### Release note 

```release-note
None
```


src/storage/lock_manager/lock_wait_context.rs
```diff
@@ -128,37 +124,48 @@ impl<L: LockManager> LockWaitContext<L> {
...
-            debug!("double invoking of finish_request of LockWaitContext";
-                "start_ts" => self.start_ts,
-                "for_update_ts" => self.for_update_ts
-            );
-            return;
-        };
+            // TODO: Uncomment this after the corresponding change of
+            // `LockManager` is done. self.lock_wait_queues.
+            // get_lock_mgr()     .remove_lock_wait(ctx_inner.
+            // lock_wait_token);
```
sticnarf, line 160: Will we remove the entry from the queue in this branch? In other words, I'm wondering whether the `finished` field in `LockWaitContextSharedState` is still necessary since there will be no stale entried in the queue.

MyonKeminta: This branch is only used when the entry is popped out from the queue and finished. Therefore the entry doesn't exist in the queue but may exist in the `WaiterManager`.
The `finished` field is not necessary now and can be removed. I'll remove it.

ekexium (approved):

sticnarf (approved):

PR #13585, opened 2022-10-11
*: fix build with panic engines

tabokie:

Issue Number: Ref #13131

`cargo test --package tikv --features test-engines-panic --no-default-features` fails.

What's Changed:

- Fix cargo manifest so that we can use non-default engines at workspace level
  - Use `ENABLE_FEATURES=test-engines-panic NO_DEFAULT_TEST_ENGINES=1 make clippy` to check panic engines
- Fix the code to build with panic engines
- Add an additional clippy check in "clippy-linux-arm64" pipeline

```commit-message
None
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


BornChanger (commented):
Where is the related pr for CI enhancement for case "cargo test --package tikv --features test-engines-panic --no-default-features" ?

SpadeA-Tang:
Would you mind telling me what's the purpose or benefit of `build with panic engines` here?

tabokie:
@SpadeA-Tang We don't actually use panic engines, but building against it makes sure the code can work with arbitrary engines that adhere to engine traits. For the purpose of engine traits, please read the documentation in "components/engine_traits".

tabokie:
@BornChanger I'll add a new clippy pipeline later.

Cargo.toml
```diff
@@ -70,13 +70,13 @@ async-trait = "0.1"
 backtrace = "0.3"
 batch-system = { path = "components/batch-system", default-features = false }
 byteorder = "1.2"
-case_macros = { path = "components/case_macros" }
-causal_ts = { path = "components/causal_ts" }
+case_macros = { path = "components/case_macros", default-features = false }
```
BusyJay, line 73: Why the change?

tabokie: Some path dependencies use default features which eventually enables `test-engine-kv-rocksdb` and `test-engine-raft-raft-engine`. This makes it impossible to disable them and use other engines.

I didn't bother checking exactly which package is the root cause, instead I added an automated pass to force disable default features for all path dependencies.

BusyJay: I think this is too verbose and confusing. For example, there is no features at all in `case_macros`. You can check what features are enabled by `cargo tree -e features`.

tabokie: We don't have CI pipelines to exhaustively check all custom feature combinations. Forcing them to disable at least avoids feature conflicts for arbitrary feature builds. As for you example, the `case_macros` might have features in the future, but we have no way to properly guard it except unconditionally disallowing using default features.

BusyJay: > the case_macros might have features in the future,

`case_macros` has been created for more than 1 year without any changes.

tabokie: The point is the user package cannot deduce whether a dependency package have default features. Unless you have a way to assert a package have no default features, the only way to programmatically make sure the feature doesn't conflict is to disallowing the possibility entirely.

components/test_coprocessor_plugin/example_plugin/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "example_plugin"
+name = "test_coprocessor_plugin"
```
BusyJay, line 2: I think it's meant to be an example instead of a test.

Makefile
```diff
@@ -334,7 +334,7 @@ pre-format: unset-override
 
 format: pre-format
 	@cargo fmt
-	@cargo sort -w ./Cargo.toml ./*/Cargo.toml components/*/Cargo.toml cmd/*/Cargo.toml >/dev/null 
+	@cargo sort -w >/dev/null
```
BusyJay, line 337: If some packages are not in workspace members, will they also be sorted?

Cargo.toml
```diff
@@ -295,6 +296,74 @@ members = [
 ]
 default-members = ["cmd/tikv-server", "cmd/tikv-ctl"]
 
+[workspace.dependencies]
```
BusyJay, line 299: Only following needs to be configured with `default-features = false`:
```
> grep 'default =' **/*.toml | grep engine
Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "cloud-aws", "cloud-gcp", "cloud-azure"]
cmd/tikv-ctl/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "cloud-aws", "cloud-gcp", "cloud-azure"]
cmd/tikv-server/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "cloud-aws", "cloud-gcp", "cloud-azure"]
components/backup-stream/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/backup/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "cloud-aws", "cloud-gcp", "cloud-azure"]
components/cdc/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/engine_test/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/engine_traits_tests/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/raftstore-v2/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/raftstore/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "engine_rocks"]
components/test_coprocessor/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/test_raftstore/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "cloud-aws", "cloud-gcp", "cloud-azure"]
components/test_storage/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
components/tikv_kv/Cargo.toml:default = ["test-engine-kv-rocksdb", "test-engine-raft-raft-engine"]
tests/Cargo.toml:default = ["failpoints", "testexport", "test-engine-kv-rocksdb", "test-engine-raft-raft-engine", "cloud-aws", "cloud-gcp", "cloud-azure"]
```

components/cloud/aws/Cargo.toml
```diff
@@ -9,36 +9,36 @@ failpoints = ["fail/failpoints"]
...
-cloud = { path = "../", default-features = false }
+cloud = { workspace = true }
 fail = "0.5"
 futures = "0.3"
 futures-util = { version = "0.3", default-features = false, features = ["io"] }
 # This is only a dependency to vendor openssl for rusoto. It's not clear exactly
 # how openssl is built for tikv, but it seems to be controlled by grpcio. This
 # makes `cargo test -p aws` link correctly.
-grpcio = { version = "0.10",  default-features = false, features = ["openssl-vendored", "protobuf-codec"] }
+grpcio = { version = "0.10", default-features = false, features = ["openssl-vendored", "protobuf-codec"] }
```
BusyJay, line 21: This can also be workspace dependency.

components/cloud/azure/Cargo.toml
```diff
@@ -6,18 +6,18 @@ publish = false
 
 [dependencies]
 async-trait = "0.1"
-azure_core = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust"}
+azure_core = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust" }
 azure_identity = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust" }
-azure_storage = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust", default-features = false, features = ["account", "blob"] }
+azure_storage = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust", features = ["account", "blob"] }
```
BusyJay, line 11: Why remove `default-features = false`?

scripts/clippy
```diff
@@ -48,4 +48,4 @@ CLIPPY_LINTS=(
 
 cargo clippy --workspace \
     --exclude fuzz-targets --exclude fuzzer-honggfuzz --exclude fuzzer-afl --exclude fuzzer-libfuzzer \
-    --features "${TIKV_ENABLE_FEATURES}" "$@" -- "${CLIPPY_LINTS[@]}"
+    --no-default-features --features "${TIKV_ENABLE_FEATURES}" "$@" -- "${CLIPPY_LINTS[@]}"
```
BusyJay, line 51: So cloud storage like "cloud-azure" will not be tested?

tabokie: They will be enabled in Makefile.

components/cloud/azure/Cargo.toml
tabokie: Must be a mismatch when using global replace.

Cargo.toml
tabokie: It's really the same discussion again. We don't know which package will introduce new default features. Disabling them all doesn't seem very annoying compared to debugging the feature conflict manually.

BusyJay: You have added the new CI pipeline, new mis-configured default features should be detected before being merged.

Makefile
tabokie: I don't know, I only know some tomls are not sorted with the old command, e.g. "components/cloud/gcp/Cargo.toml".

Cargo.toml
tabokie: For people aware of the internal reasoning to disable default features, my approach reduces mental load (the writer doesn't need to check if the package indeed has features).
For people unaware of it, my approach saves him from needing to understand how to differentiate the two conditions.
Of course we can always add some comments to explain the reasons behind if you think the "unnecessary" annotations is misrepresenting the fact.

components/test_coprocessor_plugin/example_plugin/Cargo.toml
tabokie: Will rename to "example_coprocessor_plugin".

Cargo.toml
BusyJay: Disabling default features is not consistent and causing confusions. We will put more dependencies into workspace dependencies including kvproto, grpcio, etc. Those also have default features, but will not be disabled by default. 

We should keep code clear and understood.

components/external_storage/export/examples/scli.rs
```diff
@@ -164,7 +173,10 @@ fn create_gcs_storage(opt: &Opt) -> Result<StorageBackend> {
     if let Some(prefix) = &opt.prefix {
         config.prefix = prefix.to_string();
     }
-    Ok(make_gcs_backend(config))
+    #[cfg(feature = "cloud-gcp")]
+    return Ok(make_gcs_backend(config));
+    #[cfg(not(feature = "cloud-gcp"))]
+    return Err(Error::new(ErrorKind::Other, "missing feature"));
```
BusyJay, line 179: Why need `return`?

scripts/check-bins.py
```diff
@@ -13,7 +13,7 @@
 WHITE_LIST = {
     "online_config", "online_config_derive", "tidb_query_codegen",
     "panic_hook", "fuzz", "fuzzer_afl", "fuzzer_honggfuzz", "fuzzer_libfuzzer",
-    "coprocessor_plugin_api", "example_plugin", "memory_trace_macros", "case_macros",
+    "coprocessor_plugin_api", "test_coprocessor_plugin", "memory_trace_macros", "case_macros",
```
BusyJay, line 16: Wrong name.

components/external_storage/export/examples/scli.rs
tabokie: Something to do with the cfg gate?
```
expected `;`, found `#`
   --> components/external_storage/export/examples/scli.rs:216:36
    |
216 |     Ok(make_azblob_backend(config))
    |                                    ^ help: add `;` here
217 |     #[cfg(not(feature = "cloud-azure"))]
    |     - unexpected token
```

BusyJay (approved):

Connor1996 (approved):
LGTM

PR #13626, opened 2022-10-18
raftstore-v2: test_read is not stable

### LINKED ISSUE #13625: test_read of raftstore-v2 is not stable and often fails in CI

SpadeA-Tang:

Issue Number: Close #13625

What's Changed:

It fails in CI occasionally.
The main reason for it I think is the lease expires unexpectedly due to the slow execution.
This  PR change the lease from 1 sec to 10 sec and refactor the test to make less thread creation.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore-v2/src/operation/query/local.rs
+++ b/components/raftstore-v2/src/operation/query/local.rs
@@ -564,7 +564,7 @@ mod tests {
         let leader2 = prs[0].clone();
         region1.set_region_epoch(epoch13.clone());
         let term6 = 6;
-        let mut lease = Lease::new(Duration::seconds(10), Duration::milliseconds(250));
+        let mut lease = Lease::new(Duration::seconds(10), Duration::milliseconds(2500));
         let read_progress = Arc::new(RegionReadProgress::new(&region1, 1, 1, "".to_owned()));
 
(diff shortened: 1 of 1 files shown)


components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -552,7 +564,7 @@ mod tests {
         let leader2 = prs[0].clone();
         region1.set_region_epoch(epoch13.clone());
         let term6 = 6;
-        let mut lease = Lease::new(Duration::seconds(1), Duration::milliseconds(250));
+        let mut lease = Lease::new(Duration::seconds(10), Duration::milliseconds(250));
```
sticnarf, line 567: It would look more reasonable that `renew_leader_lease_advance_duration` is changed to 2.5s according to the new max lease?

sticnarf (approved):

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: Good suggestio.

tonyxuqqi (approved):

PR #13631, opened 2022-10-18
raftstore-v2: support conf change

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
Most workflow is the same as v1 except if apply failed, v2 will not return apply result
back to raftstore. This behavior is a legacy behavior that only works for conf change
before joint consensus and can be removed now.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore-v2/src/operation/command/admin/conf_change.rs
+++ b/components/raftstore-v2/src/operation/command/admin/conf_change.rs
@@ -37,7 +37,7 @@ use crate::{
 
 /// The apply result of conf change.
 #[derive(Default, Debug)]
-pub struct ConfChange {
+pub struct ConfChangeResult {
     pub index: u64,
     // The proposed ConfChangeV2 or (legacy) ConfChange
     // ConfChange (if it is) will convert to ConfChangeV2
@@ -126,7 +126,7 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
         Ok(proposal_index)
     }
 
-    pub fn on_apply_res_conf_change(&mut self, conf_change: ConfChange) {
+    pub fn on_apply_res_conf_change(&mut self, conf_change: ConfChangeResult) {
         // TODO: cancel generating snapshot.
 
         // Snapshot is applied in memory without waiting for all entries being
@@ -265,7 +265,7 @@ impl<EK: KvEngine, R> Apply<EK, R> {
         }
         let mut resp = AdminResponse::default();
         resp.mut_change_peer().set_region(new_region);
-        let mut conf_change = ConfChange {
+        let mut conf_change = ConfChangeResult {
             index,
             conf_change: cc,
             changes: changes.to_vec(),
--- a/components/raftstore-v2/src/operation/command/admin/conf_change.rs
+++ b/components/raftstore-v2/src/operation/command/admin/conf_change.rs
@@ -1,0 +1,509 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+//! This module implements the configuration change command.
+//!
+//! The command will go through the following steps:
+//! - Propose conf change
+//! - Apply after conf change is committed
+//! - Update raft state using the result of conf change
+
+use collections::HashSet;
+use engine_traits::{KvEngine, RaftEngine};
+use kvproto::{
+    metapb::{self, PeerRole},
+    raft_cmdpb::{AdminRequest, AdminResponse, ChangePeerRequest, RaftCmdRequest},
+    raft_serverpb::{PeerState, RegionLocalState},
+};
+use protobuf::Message;
+use raft::prelude::*;
+use raft_proto::ConfChangeI;
+use raftstore::{
+    store::{
+        metrics::{PEER_ADMIN_CMD_COUNTER_VEC, PEER_PROPOSE_LOG_SIZE_HISTOGRAM},
+        util::{self, ChangePeerI, ConfChangeKind},
+        ProposalContext,
+    },
+    Error, Result,
+};
+use slog::{error, info, warn};
+use tikv_util::box_err;
+
+use super::AdminCmdResult;
+use crate::{
+    batch::StoreContext,
+    raft::{Apply, Peer},
+    router::ApplyRes,
+};
+
+/// The apply result of conf change.
+#[derive(Default, Debug)]
+pub struct ConfChange {
+    pub index: u64,
+    // The proposed ConfChangeV2 or (legacy) ConfChange
+    // ConfChange (if it is) will convert to ConfChangeV2
+    pub conf_change: ConfChangeV2,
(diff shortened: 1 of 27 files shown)


components/raftstore-v2/src/operation/command/admin/conf_change.rs
```diff
@@ -0,0 +1,509 @@
...
+        }
+        if remove_self {
+            self.mark_for_destroy(None);
+        }
+    }
+}
+
+impl<EK: KvEngine, R> Apply<EK, R> {
+    #[inline]
+    pub fn apply_conf_change(
```
Connor1996, line 178: let's remove the support of `ChangePeerv1` in raftstore v2

BusyJay: I'm afraid it's impossible if we need to support upgrading from old version.

tonyxuqqi: > I'm afraid it's impossible if we need to support upgrading from old version.

What's the default version in 6.1 TLS? We can require the minimal version to upgrade from is 6.1. If it can help remove v1 support.

components/raftstore-v2/src/operation/command/admin/conf_change.rs
```diff
@@ -0,0 +1,509 @@
...
+use super::AdminCmdResult;
+use crate::{
+    batch::StoreContext,
+    raft::{Apply, Peer},
+    router::ApplyRes,
+};
+
+/// The apply result of conf change.
+#[derive(Default, Debug)]
+pub struct ConfChange {
```
tonyxuqqi, line 40: It's easy to confuse with raft::eraftpb::ConfChange. 
How about renaming it to ConfChangeRes

components/raftstore-v2/src/operation/command/admin/conf_change.rs
```diff
@@ -0,0 +1,509 @@
...
+    pub fn propose_conf_change<T>(
+        &mut self,
+        ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+    ) -> Result<u64> {
+        if self.raft_group().raft.has_pending_conf() {
+            info!(
+                self.logger,
+                "there is a pending conf change, try later";
+            );
```
tonyxuqqi, line 63: I guess the log is likely not needed as the caller may print the error too.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -368,9 +470,7 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
         let (index, term) = self.apply_progress();
         apply_res.applied_index = index;
         apply_res.applied_term = term;
-        if self.reset_state_changed() {
```
tonyxuqqi, line 371: Question: Why is it removed?

components/raftstore/src/store/util.rs
```diff
@@ -847,6 +848,118 @@ impl<'a> ChangePeerI for &'a ChangePeerV2Request {
...
+///
+/// Make sure the peer can serve read and write when ignore safety, otherwise
+/// it may produce stale result or cause unavailability.
+pub fn check_conf_change(
+    cfg: &Config,
+    node: &RawNode<impl raft::Storage>,
+    leader: &metapb::Peer,
+    change_peers: &[ChangePeerRequest],
+    cc: &impl ConfChangeI,
+    ignore_safety: bool,
```
tonyxuqqi, line 865: is ignore_safety only applicable to test scenario?

components/raftstore-v2/src/operation/command/admin/conf_change.rs
BusyJay: Both v1 and v2 are used.

components/raftstore/src/store/util.rs
BusyJay: No, for example, when doing unsafe online recovery, safety is guaranteed by other mechanism.

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -1,30 +1,35 @@
...
 /// Apply applies all the committed commands to kv db.
 pub struct Apply<EK: KvEngine, R> {
+    peer: metapb::Peer,
     remote_tablet: CachedTablet<EK>,
     tablet: EK,
     write_batch: Option<EK::WriteBatch>,
 
     callbacks: Vec<(Vec<CmdResChannel>, RaftCmdResponse)>,
 
+    tombstone: bool,
```
Connor1996, line 27: keep the name of `pending_remove`? or add a comment here

components/raftstore-v2/src/operation/command/mod.rs
BusyJay: Because the state is updated in place, no need to detect changes.

components/raftstore-v2/src/operation/command/admin/conf_change.rs
Connor1996: Sadly, we only enforce all operators go through v2 in v6.3 https://github.com/tikv/pd/pull/5408

Connor1996 (commented):
rest LGTM

components/raftstore-v2/src/operation/command/admin/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+            apply::notify_req_region_removed(self.region_id(), ch);
+            return;
+        }
+        if let Err(e) = self.validate_command(&req, &mut ctx.raft_metrics) {
+            let resp = cmd_resp::new_error(e);
+            ch.report_error(resp);
+            return;
+        }
+
+        if !self.applied_to_current_term() {
```
tonyxuqqi, line 54: minor: it would be better to add the comment why applied_to_current_term is needed here.

components/raftstore-v2/src/operation/command/admin/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+        self.propose_pending_command(ctx);
+        let cmd_type = req.get_admin_request().get_cmd_type();
+        let res = if apply::is_conf_change_cmd(&req) {
+            self.propose_conf_change(ctx, req)
+        } else {
+            // propose other admin command.
+            unimplemented!()
+        };
+        if let Err(e) = &res {
+            info!(
```
tonyxuqqi, line 75: info --> error ?

tonyxuqqi (approved):
overall LGTM except some minor issues.

components/raftstore-v2/src/operation/command/admin/mod.rs
BusyJay: It's OK to fail. It does no harm as client/pd should retry. Unless there is bugs.

Connor1996 (approved):

PR #13633, opened 2022-10-18
Protect Raft Engine Disk Usage

### LINKED ISSUE #13642: Stop write flow if raftdb/raft engine is full
## Development Task
Stop write flow if raftdb/raft engine is full.

jiayang-zheng:
Signed-off-by: Jarvis Zheng <jiayang@hust.edu.cn>

Issue Number: Close #13642 

What's Changed:  

```
Stop write flow if raft disk is full.
```

### Related changes
https://github.com/tikv/raft-engine/pull/276

### Check List 

Tests 

- Integration test
https://pingcap.feishu.cn/docx/TGf0d1fcJozC2HxRcJDch1sQn4e

Side effects
None.

### Release note 

```release-note
Stop write flow if raft disk is full.
```


tonyxuqqi (commented):
LGTM.
It will be better to add a unit test case for path_in_diff_mount_point and an integration test for raft log full scenario.

etc/config-template.toml
```diff
@@ -413,6 +413,11 @@
 ## old log files to recycle disk space ASAP.
 # raft-engine-purge-interval = "10s"
 
+## The percentage of used raft engine capacity to trigger the disk usage stat to almost full.
+## It should be set between 1 and 100. If availble capactiy of raftdb is less than
+## (100 - `raft_engine_almost_full_percent`)/2, the disk usage stat will be marked as already full.
+# raft-engine-almost-full-percent = 90
```
tabokie, line 419: Could we not add another parameter? The old disk full threshold is hardcoded.

jiayang-zheng: My understanding is that the old disk full threshold can be controlled by `reserve_space`, which is configurable. `raft-engine-almost-full-percent` could be used to set the almost full threshold if raft disk is deployed separately.

jiayang-zheng:
It's should be a bug for github: when I clicked re-request review for tabokie, the system will remove review request for BusyJay. 😥

components/raftstore/src/store/config.rs
```diff
@@ -92,6 +92,9 @@ pub struct Config {
     #[online_config(skip)]
     pub raft_reject_transfer_leader_duration: ReadableDuration,
 
+    // the percentage used capacity of raft engine to reach `almost full` disk stat
+    pub raft_engine_almost_full_percent: u64,
```
tonyxuqqi, line 96: As you mentioned today we use reserve-space for kv disk protection. I think it would be consistant to mirror the name: reserve-raft-disk-space for raft disk.  And same as kv disk if it's zero, then raft disk protection is disabled.

components/server/src/server.rs
```diff
@@ -1472,14 +1479,47 @@ where
                     .get_engine_size()
                     .expect("get raft engine size");
 
+                let mut raft_disk_status = disk::DiskUsage::Normal;
```
tonyxuqqi, line 1482: I think we can add the failpoint here so that we can set the raft_disk_status to whatever we like, and then can follow the test code in tests/failpoints/cases/test_disk_full.rs. We just need one integration test to ensure the functionality works.

jiayang-zheng: Integration test could not run into `fn init_storage_stats_task` in which you marked here, so it's may not easy to have an integration case for the case. Fortunately,  the manual test should cover the code path.

components/raftstore/src/store/config.rs
jiayang-zheng: This may introduce some works, I will take a look after my vacation.

components/server/src/server.rs
```diff
@@ -1472,14 +1479,47 @@ where
...
+                    let raft_disk_cap = raft_disk_stats.total_space();
+                    let mut raft_disk_available =
+                        raft_disk_cap.checked_sub(raft_size).unwrap_or_default();
+                    raft_disk_available = cmp::min(raft_disk_available, raft_disk_stats.available_space());
+                    raft_disk_status = if raft_disk_available * 100
+                        <= raft_disk_cap * ((100 - raft_engine_almost_full_percent) / 2)
+                    {
+                        disk::DiskUsage::AlreadyFull
+                    } else if raft_disk_available * 100
+                        <= raft_disk_cap * (100 - raft_engine_almost_full_percent)
```
JmPotato, line 1504: Maybe we can just calculate these values just once before the check.

components/server/src/server.rs
```diff
@@ -1443,6 +1446,10 @@ where
             info!("disk space checker not enabled");
             return;
         }
+        let raft_path = engines.raft.get_engine_path().to_string();
+        let separate_raft_path = path_in_diff_mount_point(raft_path.as_str(), engines.kv.path());
```
JmPotato, line 1450: I think `separated_raft_disk_path` or `separated_raft_mount_path` is more proper.

JmPotato (commented):
Will we have runtime real-time disk usage detection? The code for this PR looks like it only checks to take effect at startup.

components/server/src/server.rs
jiayang-zheng: These values are meaningful only when `sepearte_raft_path` is true, so calculate them here should be easier to understanding. It may cost very little for very `DEFAULT_STORAGE_STATS_INTERVAL` check.

jiayang-zheng:
> Will we have runtime real-time disk usage detection? The code for this PR looks like it only checks to take effect at startup.

It checks in a background worker for every `DEFAULT_STORAGE_STATS_INTERVAL`.

components/server/src/server.rs
jiayang-zheng: Change `Full Percentage` to `Reserved Space`, no need to calculate these value now. Thanks.

components/server/src/server.rs
jiayang-zheng: Fixed.

components/server/src/server.rs
```diff
@@ -1475,14 +1517,45 @@ where
                     .get_engine_size()
                     .expect("get raft engine size");
 
+                let mut raft_disk_status = disk::DiskUsage::Normal;
+                if separated_raft_mount_path {
```
tonyxuqqi, line 1521: minor: I think it might be better to use reserve_raft_space != 0 instead of "separated_raft_mount_path" so that when it's configured to 0, the following block can be skipped completely.

tonyxuqqi (approved):

components/server/src/server.rs
jiayang-zheng: Thanks for your suggestion - will check separated_raft_mount_path and reserve_raft_space here.

etc/config-template.toml
jiayang-zheng: Change `raft-engine-almost-full-percent` to `reserve-raft-space` to set the reserved space for raft disk if it is deployed separately.

etc/config-template.toml
```diff
@@ -253,6 +253,12 @@
 ## Set it to 0 will cause no space is reserved at all. It's generally used for tests.
 # reserve-space = "5GB"
 
+## Reserve some space for raft disk if raft disk is separated deployed with kv disk.
+## `max(reserve-raft-space, raft disk capacity * 5%)` will be reserved exactly.
+##
+## Set it to 0 will cause no space is reserved at all. It's generally used for tests.
+#reserve-raft-space = "1GB"
```
tabokie, line 260:
```suggestion
# reserve-raft-space = "1GB"
```

components/engine_panic/src/raft_engine.rs
```diff
@@ -144,6 +144,10 @@ impl RaftEngine for PanicEngine {
         panic!()
     }
 
+    fn get_engine_path(&self) -> &str {
```
tabokie, line 147:
```suggestion
    fn path(&self) -> &str {
```

components/server/src/server.rs
```diff
@@ -1442,10 +1478,16 @@ where
...
         let snap_mgr = self.snap_mgr.clone().unwrap();
         let reserve_space = disk::get_disk_reserved_space();
-        if reserve_space == 0 {
+        let reserve_raft_space = disk::get_raft_disk_reserved_space();
+        if reserve_space == 0 && reserve_raft_space == 0 {
             info!("disk space checker not enabled");
             return;
         }
+        let raft_path = engines.raft.get_engine_path().to_string();
+        let separated_raft_mount_path =
```
tabokie, line 1487: Shouldn't `set_raft_disk_reserved_space` if mount is the same.

components/engine_panic/src/raft_engine.rs
jiayang-zheng: KV engine also have an interface `path` in misc trait - that may introduce some ambiguous name problems

components/server/src/server.rs
jiayang-zheng: use the default value zero. raft_almost_full_threshold/raft_already_full_threshold are not used if separated_raft_mount_path is false.

tabokie: Then there's no need to check the mountpoint again here.

jiayang-zheng:
``` Rust
                let used_size = if !separated_raft_mount_path {
                    snap_size + kv_size + raft_size + placeholder_size
                } else {
                    snap_size + kv_size + placeholder_size
                };
```
Calculate the used size may need this value.

components/server/src/server.rs
```diff
@@ -1493,18 +1566,28 @@ where
...
                 let prev_disk_status = disk::get_disk_status(0); //0 no need care about failpoint.
-                let cur_disk_status = if available <= already_full_threshold {
+                let cur_kv_disk_status = if available <= already_full_threshold {
                     disk::DiskUsage::AlreadyFull
                 } else if available <= almost_full_threshold {
                     disk::DiskUsage::AlmostFull
                 } else {
                     disk::DiskUsage::Normal
                 };
+                let cur_disk_status = match (raft_disk_status, cur_kv_disk_status) {
```
tabokie, line 1576: Move this into a method of `DiskUsage`, e.g. impl std::ops::And<DiskUsage> for DiskUsage

components/server/src/server.rs
tabokie: The point is these two things are equivalent. There's no need to use separated_raft_mount_path either. In fact, a more readable way to write is `let separated_raft_mount_path = reserve_raft_space != 0`.

components/server/src/server.rs
```diff
@@ -527,36 +530,69 @@ where
         // enough space to do compaction and region migration when TiKV recover.
         // This file is created in data_dir rather than db_path, because we must not
         // increase store size of db_path.
```
tabokie, line 532: Something like this is cleaner:
```rust
fn calculate_logical_reserved_space(dir: &Path, logical_capacity: u64) -> u64 { .. }
fn reserve_physical_space(dir: &Path, logical_reserved_space) { ... }
let size = calculate_logical_reserved_space(kv_dir, self.config.raft_store.capacity.0);
disk::set_disk_reserved_space(size);
reserve_physical_space(kv_dir, size);
/* raft dir */
```

components/server/src/server.rs
jiayang-zheng: separated_raft_mount_path is the result of whether raft path and kv path are in same mount point; reserve_raft_space is get from the config toml, it can be set to zero by manual. If reserve_raft_space is set to zero, it means user do not like disk usage effected by raft disk(raft disk status alway `Normal`).

jiayang-zheng: reserve_raft_space works only when separated_raft_mount_path is true.

components/server/src/server.rs
jiayang-zheng: Fixed.

components/server/src/server.rs
jiayang-zheng: Since DiskUsage is generated by proto, just fix it by adding a `calculate_disk_usage` function.

components/engine_panic/src/raft_engine.rs
jiayang-zheng: TODO is added.

tonyxuqqi (approved):

PR #13639, opened 2022-10-20
raftstore-v2: support tablet split

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

Enable TiKV to split region in raftstore v2.

* There some changes compared to the split operation in v1:
Each region has a separate rocksdb, namely `tablet`, in v2, so the `tablet` will be split too.
Now, for simplicity, we use the checkpoint API of the RocksDb to clone the parent peer's tablet for split peers. Then, DeleteFilesInRange can be used to free some sst files quickly, and we use compaction filter + manual compaction (it is not ready until #13228 is merged) to clean up.

* Split in v2 does not create split regions' peer directly. Instead, it sends `StoreMsg::Creation` to the store to create split peers if it is not existed now. Then, the store will send initialization information to the relevant peer to initialize itself. When the parent peer receives all the initialization responses of the split peers, the split is done.

todo:
* We use checkpoint here in the way that the memtable of the source rocksdb will be flushed which could cause jitter. So, consider freezing the memtable.
* Only when parent receives all the responses can the apply index be updated.
* Consider more when the split peers is already initialized.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


BusyJay:
Please split the PR, it's too large to be reviewed.

components/engine_test/src/lib.rs
```diff
@@ -244,9 +244,19 @@ pub mod kv {
 
     // Extract tablet id and tablet suffix from the path.
     fn get_id_and_suffix_from_path(path: &Path) -> (u64, u64) {
+        let split_path = path
+            .as_os_str()
+            .to_str()
+            .map_or(false, |s| s.contains("split"));
```
tonyxuqqi, line 250: > contains
Should it be strictly prefix? contains may match the string token in the middle of the path.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,884 @@
...
+    }
+}
+
+fn get_range_not_in_region(region: &metapb::Region) -> Vec<Range<'_>> {
+    let mut ranges = Vec::new();
+    if !region.get_start_key().is_empty() {
+        ranges.push(Range::new(b"", region.get_start_key()));
+    }
+    if !region.get_end_key().is_empty() {
+        ranges.push(Range::new(region.get_end_key(), b""));
```
tonyxuqqi, line 201: are you sure here the end key can be "" and the caller will handle it properly?
In v1 it's converted to  keys::DATA_MAX_KEY

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,884 @@
...
+        }
+
+        drop(meta);
+
+        let ranges_to_delete = get_range_not_in_region(&region);
+        tablet
+            .delete_ranges_cfs(DeleteStrategy::DeleteFiles, &ranges_to_delete)
+            .unwrap_or_else(|e| {
+                error!(self.logger,"failed to delete files in range"; "err" => %e);
+            });
```
tonyxuqqi, line 383: > let ranges_to_delete = get_range_not_in_region(&region);
>         tablet
>             .delete_ranges_cfs(DeleteStrategy::DeleteFiles, &ranges_to_delete)
>             .unwrap_or_else(|e| {
>                 error!(self.logger,"failed to delete files in range"; "err" => %e);
>             });

There's no point to call it here which blocks the peer's serving requests.
Because it's still the best effort style and DeleteStrategy::DeleteFiles will not delete entries of the ranges_to_delete if they live in a file shared with other ranges.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,884 @@
...
+                .tablet_factory
+                .split_tablet_path(region_id, tablet_index);
+            let tablet = store_ctx
+                .tablet_factory
+                .load_tablet(&split_tablet_path, region_id, tablet_index)
+                .unwrap();
+
+            let ranges_to_delete = get_range_not_in_region(self.region());
+            // todo: async version
+            tablet
```
tonyxuqqi, line 456: >  let ranges_to_delete = get_range_not_in_region(self.region());
>             // todo: async version
>             tablet
>                 .delete_ranges_cfs(DeleteStrategy::DeleteFiles, &ranges_to_delete)
>                 .unwrap_or_else(|e| {
>                     error!(self.logger,"failed to delete files in range"; "err" => %e);
>                 });

I think there's no point to delete it here as it does not delete all the unwanted data and it blocks the peer's serving requests. Instead, using compaction filter is enough.

src/server/engine_factory_v2.rs
```diff
@@ -482,4 +503,16 @@ mod tests {
...
+    #[test]
+    fn test_extract_id_and_suffix() {
+        let path = Path::new("xxx/1_10");
+        assert_eq!(get_id_and_suffix_from_path(path), (1, 10));
+        let path = Path::new("xxxx/split_1_10");
+        assert_eq!(get_id_and_suffix_from_path(path), (1, 10));
+        let path = Path::new("xxxasdafge");
+        assert_eq!(get_id_and_suffix_from_path(path), (0, 1));
+        let path = Path::new("xxxx/split_1");
+        assert_eq!(get_id_and_suffix_from_path(path), (1, 1));
```
tonyxuqqi, line 516: what about xxxsplit_1

PR #13641, opened 2022-10-20
raftstore: send heartbeat to pd as soon as it is started

### LINKED ISSUE #13627: Send heartbeat to PD as soon as the service is ready
**Problem**
After a tikv instance starts, it will wait for 10s (pd_heartbeat_tick_interval) before sending the first heartbeat to pd. PD treats the store as disconnected before receiving its heartbeat, and PD will not do scheduling work for the store.

For example, in the rolling restart scenario , the service is ready after a tikv restart, but no leader will be migrated within 10s. But in fact, at this time, it can already accept leaders.

**Solution**
Sending a heartbeat to the PD immediately after the service is ready,  which is more reasonable.

/assign @cosven

cosven:
Signed-off-by: cosven <yinshaowen241@gmail.com>

Issue Number: Close #13627

What's Changed:

```commit-message
Send first heartbeat to pd as soon as the tikv is started.
```

### Check List 

Tests 

  - case1: tikv can be deployed successfully

Side effects

### Release note 

```release-note
Send heartbeat to PD as soon as a instance is started instead of waiting another 10s
```

### DIFF
--- a/components/raftstore/src/store/fsm/store.rs
+++ b/components/raftstore/src/store/fsm/store.rs
@@ -773,6 +773,10 @@ impl<'a, EK: KvEngine + 'static, ER: RaftEngine + 'static, T: Transport>
         self.register_compact_lock_cf_tick();
         self.register_snap_mgr_gc_tick();
         self.register_consistency_check_tick();
+
+        // Send store heartbeat to PD, so that PD knows this store is started.
+        // PD will do some scheduling for this store when it receives this heartbeat.
+        self.store_heartbeat_pd(None);
     }
 }
 


cosven:
/cc @JmPotato @Connor1996
Please help review this PR ❤️

cosven:
/cc @BusyJay 
Please help review this PR ❤️

components/raftstore/src/store/fsm/store.rs
```diff
@@ -773,6 +773,10 @@ impl<'a, EK: KvEngine + 'static, ER: RaftEngine + 'static, T: Transport>
         self.register_compact_lock_cf_tick();
         self.register_snap_mgr_gc_tick();
         self.register_consistency_check_tick();
+
+        // Send store heartbeat to PD, so that PD knows this store is started.
+        // PD will do some scheduling for this store when it receives this heartbeat.
+        self.store_heartbeat_pd(None);
```
JmPotato, line 779: What about move this into `self.register_pd_store_heartbeat_tick()`?

JmPotato (approved):

Connor1996 (commented):
Please add test

components/raftstore/src/store/fsm/store.rs
cosven: Generally, `on_pd_store_heartbeat_tick` will invoke `store_heartbeat_pd` and register the tick again. So I think we should not call `store_heartbeat_pd` in register_pd_store_heartbeat_tick.

tests/integrations/raftstore/test_start.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2016 TiKV Project Authors. Licensed under Apache-2.0.
```
Connor1996, line 1: 2022

tests/integrations/raftstore/test_start.rs
```diff
@@ -0,0 +1,39 @@
...
+/// Store should send heartbeats to PD just after it starts.
+#[test]
+fn test_store_heartbeat_after_start() {
+    let mut cluster = new_node_cluster(0, 1);
+    cluster.cfg.raft_store.pd_store_heartbeat_tick_interval = ReadableDuration::millis(2000);
+    cluster.run();
+    let pd_client = Arc::clone(&cluster.pd_client);
+    // The store needs not to wait for a round of tick, and it should report
+    // heartbeat just after it starts.
+    must_wait_until(1000, 50, || -> bool {
```
BusyJay, line 36: Meaningless to write another function.

components/raftstore/src/store/fsm/store.rs
```diff
@@ -782,6 +782,10 @@ impl<'a, EK: KvEngine + 'static, ER: RaftEngine + 'static, T: Transport>
...
```
BusyJay, line 788: Will it be conflict with the protection that was introduced when implementing async log fetch?

tests/integrations/raftstore/test_start.rs
cosven: Do you mean to remove the function `must_wait_until` and just put the logic in the test?
```rust
fn test_store_heartbeat_after_start() {
  ...
  let timer = Instant::now();
      loop {
          if timer.saturating_elapsed().as_millis() > 1000 as u128 {
              panic!("wait timeout");
          }
          if f() {
              return;
          }
          sleep_ms(50);
      }
}
```

BusyJay: Yes.

cosven:
After some discussion with @BusyJay , we decide to hold this PR until the problem mentioned in  https://github.com/tikv/tikv/pull/11921 is resolved.

PR #13645, opened 2022-10-21
copr, json: support json path range selection

### LINKED ISSUE #13644: Support json array range selection
## Development Task

Support json path with following syntax:

1. `[1 to 2]`, range selection
2. From last index: `[last]`, `[ 1 to last]`, `[last - 10]`, `[last - 10 to last - 5]`

YangKeao:

Issue Number: Close #13644

What's Changed:

Support json array range selection, and use `nom` to refractor the json path parser (it looks more convincing now).

Really regret that I didn't use `nom` last time :cry: :facepalm: 

### Check List 

Tests 

- [x] Unit test
- Integration test
- No code

### Release note 

```release-note
Support array range selection and array index from last
```


YangKeao:
/cc @sticnarf 
/cc @AndreMouche 

PTAL

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -536,18 +515,17 @@ mod tests {
...
                 "$.\"\\u33\"",
-                // TODO: pass the position in the unquote unicode error
-                Some("Invalid unicode, byte len too short"),
+                Some("Invalid JSON path expression. The error is around character position 3."),
                 None,
             ),
             (
                 "$**",
-                Some("Invalid JSON path expression. The error is around character position 3."),
+                Some("Invalid JSON path expression. The error is around character position 4."),
```
sticnarf, line 523: The error position is a bit confusing...?

YangKeao: Yep! Fixed.

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
```diff
@@ -556,8 +534,23 @@ mod tests {
...
             (
-                "$ [ 2147483648 ]",
-                Some("Invalid JSON path expression. The error is around character position 15."),
+                "$ [ 4294967296 ]",
+                Some("Invalid JSON path expression. The error is around character position 5."),
+                None,
+            ),
+            (
+                "$ [ 1to2 ]",
+                Some("Invalid JSON path expression. The error is around character position 2."),
```
sticnarf, line 543: The position is also a bit strange. Can we lift error to failure after we already match `[`?

sticnarf (commented):
Mostly lgtm

components/tidb_query_datatype/src/codec/mysql/json/path_expr.rs
YangKeao: Yes! I forgot to lift to failure for the ending ']'. It successfully parses an index `1`, but failed to get the `]` (which is actually a `t`).

Now, after lifting, the error position is 6 (the index of `t`). PTAL

sticnarf (approved):
LGTM

components/tidb_query_datatype/src/codec/mysql/json/binary.rs
```diff
@@ -5,9 +5,26 @@ use std::convert::TryInto;
...
 use super::{constants::*, JsonRef, JsonType, ERR_CONVERT_FAILED};
-use crate::codec::Result;
+use crate::codec::{mysql::json::path_expr::ArrayIndex, Result};
 
 impl<'a> JsonRef<'a> {
+    /// Gets the index from the ArrayIndex
+    ///
+    /// If the idx is from end, and is greater than the count, it will return
+    /// None
+    pub fn array_get_index(&self, idx: ArrayIndex) -> Option<usize> {
```
AndreMouche, line 15: Could you please record the ref to `tidb` as L30 did? I have no idea about the logic with no reference==

YangKeao: The parser of json path in TiDB locates in https://github.com/pingcap/tidb/blob/a9373ea526e987396d286878a136501bbcc94d1d/types/json_path_expr.go#L304, and the function `array_get_index` is similar with https://github.com/pingcap/tidb/pull/38584/files#diff-c122a0d90d56f1003723e643f1e0fe9a4f986b964dfa8de985722a4a990ea0eaR58.

There are two difference between TiDB implementation and TiKV:
1. TiDB doesn't have good lightweight parser libraries like `nom`, so it's manually implemented.
2. Go doesn't have tagged union, so we use negative number to represent `[last - xxx]`, use positive number to represent `[xxx]`.

The related PR is https://github.com/pingcap/tidb/pull/38584

AndreMouche (approved):
LGTM

Issue #13524, opened 2022-09-23
qps recovery after 15min when one tikv io hang


Lily2025:
### What version of TiKV are you using?
./tikv-server -V
 TiKV 
Release Version:   6.2.0-alpha
Edition:           Community
Git Commit Hash:   b1854926ad92eeedb2cb38dff2dc4bb6b30d49f0
Git Commit Branch: heads/refs/tags/v6.3.0-alpha
UTC Build Time:    2022-09-22 11:03:33
Rust Version:      rustc 1.64.0-nightly (0f4bcadb4 2022-07-30)
Enable Features:   pprof-fp jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
Profile:           dist_release
2022-09-23T14:03:46.213+0800	INFO	k8s/client.go:131	it should be noted that a long-running command will not be interrupted even the use case has ended. For more information, please refer to https://github.com/pingcap/test-infra/discussions/129
./pd-server -V
 Release Version: v6.3.0-alpha
Edition: Community
Git Commit Hash: 3cc82517395deba65bf90fac664927954e8907f6
Git Branch: heads/refs/tags/v6.3.0-alpha
UTC Build Time:  2022-09-20 11:03:03

### What operating system and CPU are you using?
8c、16G

### Steps to reproduce
inject one tikv io hang
open async io and evict-slow-store-scheduler

    -   name: "one_tikv_io_delay"
        faultType: io_delay(volumeName:tikv,delay:60m)
        selector: "tikv(random)"
        warmUpTime: 10m
        faultDuration: 120m
        recoveryTime: 10m

### What did you expect?
qps can recovery in 2min

### What did happened?
qps recovery after 15min
![BFpsi8VuVc](https://user-images.githubusercontent.com/84712107/191911259-8db288aa-2f8b-4272-bfe1-c0e8605e3103.jpg)
![img_v2_d81eb44c-c0cc-41e7-a398-3a82398225fg](https://user-images.githubusercontent.com/84712107/191911273-7cf33016-ba0f-437b-926c-c147344ec3e9.jpg)

tonyxuqqi:
it's fixed by polished slow score/slow trend

PR #13654, opened 2022-10-24
Optimize slow-score mechanism to reduce the time-cost on recovery when IO hung. 

### LINKED ISSUE #13648: Optimize slow-score mechanism for reducing the time-cost on failure recovery.
## Development Task
We need to optimize the slow-score mechanism to make the report of slow-score more timely, and make some improvements to it for reducing the time-cost on failure recovery when one TiKV node is in abnormal state, i.e. IO delay or IO hang.

Here, the related development tasks includes:
- [x] Step 1: Make the reporting of Slow-Score more timely.
- [x] Step 2: Awaken hibernated regions on healthy nodes timely.

And the related issue can be reviewed from followings:
- ISSUE: #13524

LykxSassinator:

Issue Number: Close bug #13524, Close #13648
What's Changed:

```commit-message
1. Make the reporting of Slow-Score more timely.
Optimize the reporting strategy of slow-score. Making it more timely, even if store-heartbeat is delayed because of IO delay in one TiKV node.

2. Awaken hibernated regions on healthy nodes timely when IO hang.
If one TiKV node is hung by abnormal IO, we can awaken related hibernated regions in time, to trigger self-revoting in these raft-groups for recovery.
```
### Related changes
### Check List 
Tests 
- Unit test
- Integration test
- No code
Side effects
- Performance regression
    - Consumes more CPU
    - Consumes more MEM
### Release note 
```release-note
1. Make slow-score reporting more timely to reduce the recovery time when one TiKV is delayed because of IO delay.

2. Awaken hibernated regions to trigger raft-group self-revoting in time when one TiKV node is hung by abnormal IO.
```


LykxSassinator:
The update of `kvproto` needs ref: https://github.com/pingcap/kvproto/pull/994 is closed in advance.

components/tikv_util/src/store/mod.rs
```diff
@@ -9,6 +9,78 @@ pub use self::{
...
+            for (store_id, peer_id) in s4.into_iter().zip(10..) {
+                r2.mut_peers().push(new_learner_peer(store_id, peer_id));
+            }
+            let res = region_on_same_stores(&r1, &r2);
+            assert_eq!(res, exp, "{:?} vs {:?}", r1, r2);
+        }
+    }
+
+    #[test]
+    fn test_check_region_on_store() {
```
5kbpers, line 57: Better to put this test to `region.rs` and not move the above test.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1207,13 +1214,17 @@ where
...
 
         stats = collect_report_read_peer_stats(HOTSPOT_REPORT_CAPACITY, report_peers, stats);
-        let (capacity, used_size, available) = match collect_engine_size(
-            &self.coprocessor_host,
-            Some(&store_info),
-            self.snap_mgr.get_total_snap_size().unwrap(),
-        ) {
-            Some((capacity, used_size, available)) => (capacity, used_size, available),
-            None => return,
+        let (capacity, used_size, available) = if store_info.is_some() {
```
5kbpers, line 1217:
```suggestion
        let (capacity, used_size, available) = if let Some(store_info) = store_info {
```

components/tikv_util/src/store/mod.rs
LykxSassinator: The reason why I check and modify the uts from `regions.rs` into `mod.rs`is that, the original ut in `peer.rs` includes function callings from `region.rs`.
https://github.com/tikv/tikv/blob/92aaf95e0dc83f855cf6d9bfd046c2ea0dfc9f3d/components/tikv_util/src/store/peer.rs#L96

IMO, it's more approapriate to assemble these special uts into `mod.rs`.

components/raftstore/src/store/worker/pd.rs
LykxSassinator: IMO, `is_some` is better?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -2525,6 +2528,44 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> StoreFsmDelegate<'a, EK, ER
         self.register_compact_lock_cf_tick();
     }
 
+    fn on_wake_up_regions(&self, abnormal_stores: Vec<u64>) {
+        info!("try to wake up all hibernated regions in this store";
+            "to_all" => abnormal_stores.is_empty());
+        let meta = self.ctx.store_meta.lock().unwrap();
+        for region_id in meta.regions.keys() {
+            let region = &meta.regions[region_id];
+            if !region_on_stores(region, &abnormal_stores) {
```
innerr, line 2537: Is it a N * N loop? (assume we have N regions in this store)
If so, will it be too slow?

components/raftstore/src/store/worker/pd.rs
```diff
@@ -2065,6 +2126,15 @@ where
         }
         if !self.slow_score.last_tick_finished {
             self.slow_score.record_timeout();
+            // If the last slow_score already reached abnormal state and was delayed for
+            // reporting by `store-heartbeat` to PD, we should report it here manually.
+            let now = UnixSecs::now();
+            let interval_second = now.into_inner() - self.store_stat.last_report_ts.into_inner();
+            if self.slow_score.should_force_report_slow_store()
+                && interval_second >= self.store_heartbeat_interval.as_secs()
```
innerr, line 2134: Should we use a new util fn `store_stat.store_heartbeat_delayed()` instead of:
`interval_second >= self.store_heartbeat_interval.as_secs()`?

innerr: Another thinking:
Normally we take `2 * expected-time-duration` as delayed, is this `interval_second >= self.store_heartbeat_interval.as_secs()` too sensitive?

components/tikv_util/src/store/region.rs
```diff
@@ -38,6 +38,18 @@ pub fn region_on_same_stores(lhs: &Region, rhs: &Region) -> bool {
     })
 }
 
+/// Check if the given region exists on stores.
+pub fn region_on_stores(region: &Region, store_ids: &Vec<u64>) -> bool {
```
innerr, line 42: This is a bit weird for me, could we just check the `self store id` in `store_ids`?

innerr: Then we dont need this fn

components/raftstore/src/store/msg.rs
```diff
@@ -847,6 +847,10 @@ where
     },
 
     GcSnapshotFinish,
+
+    AwakenRegions {
+        abnormal_stores: Vec<u64>,
```
innerr, line 852: Could we skip all `store ids` which are not equals to `self store id`?

components/raftstore/src/store/fsm/store.rs
LykxSassinator: On the one hand, more specifically, it's a `N * K` loop (assume we have N regions in this store, and we have K abnormal stores, K << N), so it's might be acceptable.

And on the other hand, I wanna give less pressure to PD, so the selection strategy of abnormal regions by `region_on_stores`, is pushed down to TiKV side. In this way, we just pass `abnormal stores` rather than `regions` to other TiKV nodes and let TiKV do the related computation in a `N * K` loop.

components/tikv_util/src/store/region.rs
LykxSassinator: Precisely, this function is used to check and filter out regions neither exists on `self store_id` nor `abnormal store_ids`. That is, we just wanna wake up hibernated regions both existing in `abnormal stores` and held by `self store_id`.

components/raftstore/src/store/msg.rs
LykxSassinator: Nope, as mentioned above, we wanna push down the filtration from PD to TiKV side by `stores`. And without `abnormal stores`, we won't know which store is abnormal.

components/tikv_util/src/store/region.rs
LykxSassinator: And I've update and polish the related annotations to reduce your confusions.

components/raftstore/src/store/worker/pd.rs
LykxSassinator: > Another thinking: Normally we take `2 * expected-time-duration` as delayed, is this `interval_second >= self.store_heartbeat_interval.as_secs()` too sensitive?

Here, precisely, the above `2 * expected-time-duration` u mentioned is `2 * self.pd_store_heartbeat_tick_interval` ?
IMO, it's acceptable in the scenario where IO is hung. `self.store_heartbeat_interval` is kept same with `self.pd_store_heartbeat_tick_interval == secs(10)`, that is, the timeout threshold is `10s` in this case.

components/raftstore/src/store/msg.rs
innerr: Yes you'r right

components/raftstore/src/store/fsm/store.rs
innerr: It's K * N * 3, fairly good, close

innerr: It's K * N * 3, fairly good

components/raftstore/src/store/worker/pd.rs
innerr: I think minghua means from his way, the var `store_info` will be unboxed for the following usage, which could be more convenient

LykxSassinator: Got it.
But we need the original wrapped `store_info` to make a further judgement whether the given task is a fake `StoreHeartbeat` generated by `force_report_slow_score` function, I revert the modification here to the original one => still use `is_some()`.

This can be reviewed in:
https://github.com/LykxSassinator/tikv/blob/41204d9748236f8a2d7d3d53d381b5fb1d68a9a6/components/raftstore/src/store/worker/pd.rs#L1281

LykxSassinator:
Basic performance testing has been finished, and it proved that this PR did not introduce any regression.
ping @innerr.

LykxSassinator:
@innerr PTAL, errors incurred by merging have been fixed.

components/raftstore/src/store/worker/pd.rs
```diff
@@ -771,7 +781,7 @@ impl SlowScore {
 
             inspect_interval,
             ratio_thresh: OrderedFloat(0.1),
-            min_ttr: Duration::from_secs(5 * 60),
+            min_ttr: Duration::from_secs(STORE_HEARTBEAT_DELAY_LIMIT),
```
innerr, line 784: I dont think min_ttr related to STORE_HEARTBEAT_DELAY_LIMIT, it's the min decreasing duration setting value

components/raftstore/src/store/worker/pd.rs
```diff
@@ -1786,6 +1831,57 @@ where
             health_service.set_serving_status("", status);
         }
     }
+
+    fn is_store_heartbeat_delayed(&self) -> bool {
+        let now = UnixSecs::now();
+        let interval_second = now.into_inner() - self.store_stat.last_report_ts.into_inner();
+        // To avoid reporting too many fake heartbeats to PD, we should also
```
innerr, line 1838: Too many comments in code for the logic `fake heartbeats`, any chance we could make it formal code?

LykxSassinator:
Ping @innerr , thx

innerr (approved):
LGTM

tonyxuqqi (approved):

PR #13659, opened 2022-10-24
mvcc: log the mvcc key if default not found error happens

### LINKED ISSUE #13655: use the mvcc key in the error information for default not found
## Development Task

If the default value not found error happens, the raw key will be printed in the error log, like
```
tikv aborts txn: Error(Txn(Error(Mvcc(Error(DefaultNotFound
{ key: [116, 128, 0, 0, 0, 0, 0, 1, 171, 95, 114, 128, 0, 0, 13, 234, 3, 59, 57] }
```
image

To locate the actual region and do the unsafe recovery the mvcc key is needed, so it's more convenient to print the mvcc key in the error log.

cfzjywxk:

Issue Number: Close #13655 

What's Changed:

```commit-message
Log the mvcc key if the default not error happens. Usually, the next step is to locate the key region and 
do an unsafe recovery, so mvcc key format logging is more convenient.
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
NONE
```

### DIFF
--- a/src/storage/mvcc/reader/point_getter.rs
+++ b/src/storage/mvcc/reader/point_getter.rs
@@ -347,7 +347,11 @@ impl<S: Snapshot> PointGetter<S> {
             Ok(value)
         } else {
             Err(default_not_found_error(
-                user_key.to_raw()?,
+                user_key
+                    .clone()
+                    .append_ts(write_start_ts)
+                    .as_encoded()
+                    .to_vec(),
                 "load_data_from_default_cf",
             ))
         }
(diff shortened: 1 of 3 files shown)


src/storage/mvcc/reader/point_getter.rs
```diff
@@ -347,7 +347,11 @@ impl<S: Snapshot> PointGetter<S> {
             Ok(value)
         } else {
             Err(default_not_found_error(
-                user_key.to_raw()?,
+                user_key
+                    .clone()
+                    .append_ts(write_start_ts)
+                    .as_encoded()
+                    .to_vec(),
```
MyonKeminta, line 354:
```suggestion
                    .into_encoded(),
```

MyonKeminta (commented):
`.as_encoded().to_vec()` can be simplified to `into_encoded()`, which consumes the key and avoids an extra cloning. Rest LGTM.

MyonKeminta (approved):

ekexium (approved):

Issue #13668, opened 2022-10-25
tikv panic due to corrupted raft message


glorv:
### What version of TiKV are you using?

v5.4.1

### Steps to reproduce

![image](https://user-images.githubusercontent.com/5196885/197696618-5b36b871-5c47-4fac-9826-b7236ca0978c.png)

After the first panic, tikv panic consistantly and can't recovery by itself.

A similar issue in asktug:

https://asktug.com/t/topic/842878

glorv:
downgrade to minor because it is not recur. And it's more likely to be a hardware issue.

glorv:
After meeting this issue several times, we found the root cause were all corrupted data caused by memory bit flip. So this should be a hardware issue instead of a bug. So I'm going to close this issue and feel free to reopen it if needed.

PR #13669, opened 2022-10-25
tikv-ctl,raftstore: add a log to output corrupted raft msg

glorv:

Issue Number: ref #13668

What's Changed:

```
Add a log to output corrupted raft log and let tikv-ctl to output raft log in binary format.
```
Because issue like #13668 is hard to reproduce and investigate the root cause, add a log of the corrupted data to help investigate why raft log is corrupted.

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/store/util.rs
```diff
@@ -685,6 +686,7 @@ pub(crate) fn u64_to_timespec(u: u64) -> Timespec {
 pub fn parse_data_at<T: Message + Default>(data: &[u8], index: u64, tag: &str) -> T {
     let mut result = T::default();
     result.merge_from_bytes(data).unwrap_or_else(|e| {
+        error!("meet corrupted raft message: {}", log_wrappers::Value::value(data));
```
BusyJay, line 689: How about also making it part of panic message?

glorv: I'm not sure if it's a good idea, because the raft log body can be very big which may make it hard to see the panic info.

BusyJay: There is no difference. Either panic message hard to read or the log is hard to read. Not putting it into panic message will require additional operation to search logs.

glorv:
@Connor1996  PTAL

Connor1996 (approved):
LGTM

components/raftstore/src/store/util.rs
```diff
@@ -912,7 +912,7 @@ pub fn get_entry_header(entry: &Entry) -> RaftRequestHeader {
 pub fn parse_data_at<T: Message + Default>(data: &[u8], index: u64) -> T {
     let mut result = T::default();
     result.merge_from_bytes(data).unwrap_or_else(|e| {
-        panic!("{} data is corrupted : {:?}", index, e);
+        panic!("{} data is corrupted : {:?}. hex value: {}", index, e, log_wrappers::Value::value(data));
```
BusyJay, line 915: region id is important.

glorv: The tag is removed by #14335, so shall we add it back?

BusyJay: Yes. And you can add region_id and peer_id as parameters and log them like `[region_id={}]`.

PR #13678, opened 2022-10-27
Raftstore-v2: tablet factory supports temporary split path

### LINKED ISSUE #13679: TabletFactory need to support handling split path
## Development Task
During split execution, split regions will fork tablet from the parent tablet with the "split_" prefix in the tablet path.

SpadeA-Tang:

Issue Number: Close #13679

What's Changed:

During split execution in raftstore-v2, we clone the parent tablet for split regions. In this phase, the tablet path should have "split_" prefix in it, and if the store crashes and restarts, we can remove these tablets. 

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


tonyxuqqi:
> Close #xxx

Please add the issue #

SpadeA-Tang:
PTAL

SpadeA-Tang:
> > Close #xxx
> 
> Please add the issue #

Yes, I will add one.

components/engine_test/src/lib.rs
```diff
@@ -349,6 +362,13 @@ pub mod kv {
                 .join(format!("tablets/{}_{}", id, suffix))
         }
 
+        #[inline]
+        fn split_tablet_path(&self, id: u64, suffix: u64) -> PathBuf {
```
tonyxuqqi, line 366: split_tablet_path => split_temp_tablet_path?

tonyxuqqi (approved):

components/engine_test/src/lib.rs
SpadeA-Tang: Good suggestion.

BusyJay:
I don't get it. Can you show how these APIs are used during split?

SpadeA-Tang:
> 
For example:
```
for region in split_regions {
    let path = factory.split_temp_tablet_path(region_id, RAFT_INIT_LOG_INDEX);
    parent_tablet.create_checkpoint(&path);
}
```

components/engine_traits/src/engine.rs
```diff
@@ -263,6 +274,14 @@ pub trait TabletFactory<EK>: TabletAccessor<EK> + Send + Sync {
     /// Get the tablet path by id and suffix
     fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf;
 
+    /// Get the tablet path by id and suffix with "split_" prefix in it.
+    ///
+    /// Only used during split execution.
+    fn split_temp_tablet_path(&self, _id: u64, _suffix: u64) -> PathBuf {
```
BusyJay, line 280: Instead of using an explict method, I suggest to change existing `tablet_path` and make it accept a custom prefix.

components/engine_traits/src/engine.rs
```diff
@@ -282,6 +301,11 @@ pub trait TabletFactory<EK>: TabletAccessor<EK> + Send + Sync {
         unimplemented!();
     }
 
+    /// Unregister the tablet if it is not used
+    fn unregister_tablet(&self, _region_id: u64, _suffix: u64) {
```
BusyJay, line 305: When will it be used?

components/engine_traits/src/engine.rs
SpadeA-Tang: Except split, is there any situations that we need customized tablet path?

components/engine_traits/src/engine.rs
SpadeA-Tang: After the derived peer load the new cloned tablet in factory, it is used to unregister the old one.

components/engine_traits/src/engine.rs
BusyJay: Merge will also create a temporary tablet.

components/engine_traits/src/engine.rs
BusyJay: Why not only keeping one tablet for one region?

SpadeA-Tang: This is exactly the reason that I need it.

BusyJay: Why not index the tablet by region ID only?

SpadeA-Tang: This is used to clear cache only.
After creating the checkpoint of the tablet, source peer opens the new cloned tablet (ex: old tablet -> 1_5, and new cloned one -> split_1_10) .  The source region opens the split_1_10, so there are two tablets reserved in the cache (1_5, split_1_10).  `unregister_tablet` is used to remove 1_5 from cache.

SpadeA-Tang: "What I mean is when register 1_10, 1_5 should be removed automatically."
How this can be done, by traversing?

BusyJay: By "index the tablet by region ID only".

SpadeA-Tang: You mean change the signature of `fn open_tablet(id, suffix, options)` to `fn open_tablet(id, options)` ?

BusyJay: No, by changing the type of registry from `Arc<Mutex<HashMap<(u64, u64), RocksEngine>>>` to `Arc<Mutex<HashMap<u64, RocksEngine>>>`

SpadeA-Tang: Okay, I see.

components/engine_traits/src/engine.rs
```diff
@@ -162,14 +162,16 @@ impl Drop for TabletErrorCollector {
...
-#[derive(Default, Clone)]
+#[derive(Default, Clone, Debug)]
 pub struct OpenOptions {
     // create tablet if non-exist
     create: bool,
     create_new: bool,
     read_only: bool,
     cache_only: bool,
     skip_cache: bool,
+    split_use: bool,
```
BusyJay, line 173: This is too strange. We have `skip_cache` and `open_tablet_raw`, the combination is for the exact purpose.

SpadeA-Tang: Ok，done.

components/engine_test/src/lib.rs
```diff
@@ -366,7 +356,7 @@ pub mod kv {
         #[inline]
         fn destroy_tablet(&self, id: u64, suffix: u64) -> engine_traits::Result<()> {
```
BusyJay, line 357: What if the suffix doesn't match?

components/engine_test/src/lib.rs
```diff
@@ -267,27 +257,22 @@ pub mod kv {
...
                     }
                     return Ok(tablet);
                 }
-            } else if options.cache_only() {
-                // This branch reads an arbitrary tablet with region id `id`
-
-                if let Some(k) = reg.keys().find(|k| k.0 == id) {
-                    return Ok(reg.get(k).unwrap().clone());
-                }
+            } else if let Some((tablet, _)) = reg.get(&id) {
```
BusyJay, line 274: What if `create_new` is set to true?

components/engine_traits/src/engine.rs
```diff
@@ -261,8 +264,19 @@ pub trait TabletFactory<EK>: TabletAccessor<EK> + Send + Sync {
...
+    ///
+    /// Prefix is used in special situations.
+    /// Ex: split/merge.
     fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf;
 
+    /// Get the tablet path by id and suffix
+    ///
+    /// Used in special situations
+    /// Ex: split/merge.
+    fn tablet_path_with_prefix(&self, _id: u64, _suffix: u64, _prefix: &str) -> PathBuf {
```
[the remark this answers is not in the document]
SpadeA-Tang: In the v2's implementation, `tablet_path` calls `tablet_path_with_prefix`. But it is not suitable for v1. So I do not write it in the trait directly.

components/engine_traits/src/engine.rs
```diff
@@ -261,8 +264,19 @@ pub trait TabletFactory<EK>: TabletAccessor<EK> + Send + Sync {
...
+        unimplemented!();
```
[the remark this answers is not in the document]
SpadeA-Tang: v1 should not use it at all.

components/engine_test/src/lib.rs
SpadeA-Tang: If create_new is set be true, then suffix should not be None.

components/engine_test/src/lib.rs
SpadeA-Tang: Good catch.

components/engine_traits/src/engine.rs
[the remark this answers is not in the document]
BusyJay: Why not suitable? Aren't v1 just ignore everything?

components/engine_test/src/lib.rs
BusyJay: How to enforce that?

SpadeA-Tang: I can add a check and return an err in that case.

components/engine_test/src/lib.rs
```diff
@@ -343,17 +334,22 @@ pub mod kv {
         }
 
         #[inline]
-        fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf {
+        fn tablet_path_with_prefix(&self, id: u64, suffix: u64, prefix: &str) -> PathBuf {
             self.inner
                 .root_path
-                .join(format!("tablets/{}_{}", id, suffix))
+                .join(format!("tablets/{}{}_{}", prefix, id, suffix))
```
BusyJay, line 340: Better in format `split_id_suffix`.

components/engine_test/src/lib.rs
```diff
@@ -343,17 +334,22 @@ pub mod kv {
...
         }
 
         #[inline]
         fn mark_tombstone(&self, region_id: u64, suffix: u64) {
             let path = self.tablet_path(region_id, suffix).join(TOMBSTONE_MARK);
             std::fs::File::create(&path).unwrap();
```
BusyJay, line 346: If the path doesn't exist, the function will panic.

components/engine_traits/src/engine.rs
```diff
@@ -261,7 +264,18 @@ pub trait TabletFactory<EK>: TabletAccessor<EK> + Send + Sync {
...
+    /// Ex: split/merge.
+    fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf {
+        self.tablet_path_with_prefix(id, suffix, "")
+    }
+
+    /// Get the tablet path by id and suffix
+    ///
+    /// Used in special situations
+    /// Ex: split/merge.
+    fn tablet_path_with_prefix(&self, _id: u64, _suffix: u64, _prefix: &str) -> PathBuf;
```
BusyJay, line 278: The order should be `prefix`, `id` and `suffix`. And there should be no `_`.

src/server/engine_factory_v2.rs
```diff
@@ -154,18 +145,23 @@ impl TabletFactory<RocksEngine> for KvEngineFactoryV2 {
     }
 
     #[inline]
-    fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf {
+    fn tablet_path_with_prefix(&self, id: u64, suffix: u64, prefix: &str) -> PathBuf {
```
BusyJay, line 148: Any test case?

components/engine_test/src/lib.rs
SpadeA-Tang: If {}_{}_{} is used, and if the prefix is "", it will be _x_y.

BusyJay: Isn't solved by a simple if/else? The point is `split_id_suffix` is a better format, not to do it with one line.

Though you can also do it by one line like
```
format!("{prefix}{:_>width$}{id}_{suffix}", "", width = !prefix.is_empty() as usize);
```

SpadeA-Tang: As this method is used in special situations, I predefined the prefix

`pub const SPLIT_PREFIX: &str = "split_";`
`pub const MERGE_PREFIX: &str = "merge_";`

and it is called by using these predefined const str rather than passing a arbitrary str.

BusyJay: This works too.

src/server/engine_factory_v2.rs
```diff
@@ -400,14 +404,39 @@ mod tests {
...
+        factory.destroy_tablet(1, 30).unwrap();
+
+        let result = factory.open_tablet(1, Some(30), OpenOptions::default());
         result.unwrap_err();
 
         assert!(!factory.is_single_engine());
+
+        assert!(
+            factory
+                .tablet_path_with_prefix("split_", 1, 10)
```
BusyJay, line 437: Should use the predefined const instead.

BusyJay (approved):
Rest LGTM

components/engine_traits/src/engine.rs
```diff
@@ -261,7 +264,18 @@ pub trait TabletFactory<EK>: TabletAccessor<EK> + Send + Sync {
     fn exists_raw(&self, path: &Path) -> bool;
 
     /// Get the tablet path by id and suffix
-    fn tablet_path(&self, id: u64, suffix: u64) -> PathBuf;
+    ///
+    /// Prefix is used in special situations.
+    /// Ex: split/merge.
```
tabokie, line 269:
```suggestion
```

src/server/engine_factory_v2.rs
```diff
@@ -154,18 +145,23 @@ impl TabletFactory<RocksEngine> for KvEngineFactoryV2 {
...
             .store_path()
-            .join(format!("tablets/{}_{}", id, suffix))
+            .join(format!("tablets/{}{}_{}", prefix, id, suffix))
     }
 
     #[inline]
     fn mark_tombstone(&self, region_id: u64, suffix: u64) {
         let path = self.tablet_path(region_id, suffix).join(TOMBSTONE_MARK);
-        std::fs::File::create(&path).unwrap();
+        let _ = std::fs::File::create(&path);
```
tabokie, line 157: Comment why we could ignore the error?

tabokie (approved):

src/server/engine_factory_v2.rs
SpadeA-Tang: yes, I have added.

PR #13683, opened 2022-10-27
gc_worker/compaction_filter:tiny refactor for do_filter

AndreMouche:
Signed-off-by: shirly <AndreMouche@126.com>

Issue Number: Close #xxx
tiny refactor the function do_filter

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None.
```

### DIFF
--- a/src/server/gc_worker/compaction_filter.rs
+++ b/src/server/gc_worker/compaction_filter.rs
@@ -294,7 +294,6 @@ struct WriteCompactionFilter {
     regions_provider: (u64, Arc<dyn RegionInfoProvider>),
 
     mvcc_key_prefix: Vec<u8>,
-    remove_older: bool,
 
     // Some metrics about implementation detail.
     versions: usize,
@@ -336,7 +335,6 @@ impl WriteCompactionFilter {
             regions_provider,
 
             mvcc_key_prefix: vec![],
-            remove_older: false,
 
             versions: 0,
             filtered: 0,
@@ -398,6 +396,39 @@ impl WriteCompactionFilter {
         }
     }
 
+    fn do_filter_new_key(
+        &mut self,
+        mvcc_key: &[u8],
+        write: WriteRef<'_>,
+    ) -> Result<CompactionFilterDecision, String> {
+        if self.mvcc_deletion_overlaps.take() == Some(0) {
+            self.handle_bottommost_delete();
+            if self.mvcc_deletions.len() >= DEFAULT_DELETE_BATCH_COUNT {
+                self.gc_mvcc_deletions();
+            }
+        }
+        self.switch_key_metrics();
+        self.mvcc_key_prefix.clear();
+        match write.write_type {
+            WriteType::Rollback | WriteType::Lock => {
+                self.mvcc_rollback_and_locks += 1;
+                self.filtered += 1;
+                return Ok(CompactionFilterDecision::Remove);
+            }
+            WriteType::Put => {}
+            WriteType::Delete => {
+                if self.is_bottommost_level {
+                    self.mvcc_deletion_overlaps = Some(0);
+                    GC_COMPACTION_FILTER_MVCC_DELETION_MET
+                        .with_label_values(&[STAT_TXN_KEYMODE])
+                        .inc();
+                }
+            }
+        }
+        self.mvcc_key_prefix.extend_from_slice(mvcc_key);
+        return Ok(CompactionFilterDecision::Keep);
+    }
+
     fn do_filter(
         &mut self,
         _start_level: usize,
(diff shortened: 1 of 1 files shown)


src/server/gc_worker/compaction_filter.rs
```diff
@@ -287,14 +287,12 @@ struct WriteCompactionFilter {
     gc_scheduler: Scheduler<GcTask<RocksEngine>>,
     // A key batch which is going to be sent to the GC worker.
     mvcc_deletions: Vec<Key>,
-    // The count of records covered the current mvcc-deletion mark. The mvcc-deletion
-    // mark will be sent to the GC worker only if `mvcc_deletion_overlaps` is 0. It's
-    // a little optimization to reduce modifications on write CF.
-    mvcc_deletion_overlaps: Option<usize>,
+    // The mvcc-deletion mark will be sent to the GC worker only if `mvcc_del_gc_needed` is
+    // not none. It's a little optimization to reduce modifications on write CF.
+    mvcc_del_gc_needed: Option<bool>,
```
sticnarf, line 292: It's impossible to be `Some(false)` now? Then, is it necessary to make it an `Option`?

src/server/gc_worker/compaction_filter.rs
```diff
@@ -398,6 +401,34 @@ impl WriteCompactionFilter {
...
+        write: WriteRef<'_>,
+    ) -> Result<CompactionFilterDecision, String> {
+        self.handle_bottommost_delete();
+        self.switch_key_metrics();
+        self.mvcc_key_prefix.clear();
+        match write.write_type {
+            WriteType::Rollback | WriteType::Lock => {
+                self.mvcc_rollback_and_locks += 1;
+                self.filtered += 1;
+                return Ok(CompactionFilterDecision::Remove);
```
MyonKeminta, line 416: So in this case the current key (`self.mvcc_key_prefix`) is actually not switched (but set to empty), and when we found the next version of the same user key, this function will be called again, and `handle_bottommost_delete` and `switch_key_metrics` are also called again?

I'm afraid this doesn't seem to make the code clearer and easier to read...  The `remove_older` flag still looks necessary I think

src/server/gc_worker/compaction_filter.rs
```diff
@@ -412,66 +443,31 @@ impl WriteCompactionFilter {
...
-                }
-            }
+        if self.mvcc_key_prefix != mvcc_key_prefix {
+            return self.do_filter_new_key(mvcc_key_prefix, write);
         }
 
-        if !filtered {
-            return Ok(CompactionFilterDecision::Keep);
-        }
+        // if there is any version after WriteType::Delete in bottomost level,
```
MyonKeminta, line 451:
```suggestion
        // If there is any version after WriteType::Delete in bottommost level,
```

src/server/gc_worker/compaction_filter.rs
```diff
@@ -380,10 +377,16 @@ impl WriteCompactionFilter {
     }
 
     fn handle_bottommost_delete(&mut self) {
```
Connor1996, line 379: maybe_handle_bottommost_delete

src/server/gc_worker/compaction_filter.rs
```diff
@@ -287,14 +287,12 @@ struct WriteCompactionFilter {
     gc_scheduler: Scheduler<GcTask<RocksEngine>>,
     // A key batch which is going to be sent to the GC worker.
     mvcc_deletions: Vec<Key>,
-    // The count of records covered the current mvcc-deletion mark. The mvcc-deletion
-    // mark will be sent to the GC worker only if `mvcc_deletion_overlaps` is 0. It's
-    // a little optimization to reduce modifications on write CF.
-    mvcc_deletion_overlaps: Option<usize>,
+    // The mvcc-deletion mark will be sent to the GC worker only if `mvcc_del_gc_needed` is
+    // not none. It's a little optimization to reduce modifications on write CF.
+    mvcc_del_gc_needed: Option<()>,
```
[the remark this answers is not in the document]
AndreMouche: no, I just want to use the method `take() `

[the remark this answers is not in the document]
MyonKeminta: I think you don't need to do so. A bool makes more sense here... to me

src/server/gc_worker/compaction_filter.rs
```diff
@@ -287,14 +287,15 @@ struct WriteCompactionFilter {
     gc_scheduler: Scheduler<GcTask<RocksEngine>>,
     // A key batch which is going to be sent to the GC worker.
     mvcc_deletions: Vec<Key>,
-    // The count of records covered the current mvcc-deletion mark. The mvcc-deletion
-    // mark will be sent to the GC worker only if `mvcc_deletion_overlaps` is 0. It's
-    // a little optimization to reduce modifications on write CF.
-    mvcc_deletion_overlaps: Option<usize>,
+    // The mvcc-deletion mark will be sent to the GC worker only if `mvcc_del_gc_needed` is
+    // true. It's a little optimization to reduce modifications on write CF.
```
MyonKeminta, line 291: The *little optimization* still exist here? Previously, `GcKeys` task is send only when: 1) the latest version before safepoint is `DELETE` and 2) we don't find any older version than the `DELETE` of the key. In case there are multiple versions before safepoint and the latest one is `DELETE`, the `mvc_deletion_overlaps` will be `Some(x)` where `x > 0`, so `handle_bottommost_delete` is skipped in this case. This optimization doesn't seem to be kept after your change. 😕

AndreMouche: Oh，I kept this logic:  If any older version should be removed in compaction filter, we shouldn't gc_keys for this key

PR #13680, opened 2022-10-27
txn: Add lock-with-conflict support to acquire_pessimistic_lock

MyonKeminta:
Signed-off-by: MyonKeminta <MyonKeminta@users.noreply.github.com>

### What is changed and how it works?

Issue Number: Ref #13298 

What's Changed:

```commit-message
Add `lock_with_conflict` support to  `acquire_pessimistic_lock`, and it's currently always disabled.

Side changes:
Updated the type for holding the results of `acquire_pessimistic_lock` (removed `PessimisticLockRes` and added new `PessimisticLockResults` and `PessimisticLockKeyResult`), and other necessary changes to adapt the new type.
```

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
None
```

PR #13687, opened 2022-10-28
txn: Add batch-resumed mode for acquire_pessimistic_lock storage command 

### LINKED ISSUE #13298: Enhanced queueing for conflicting pessimistic transactions

MyonKeminta:

Issue Number: Ref #13298 

Requires: https://github.com/tikv/tikv/pull/13680

What's Changed:

```commit-message
Add batch-resumed mode for acquire_pessimistic_lock storage command.

Now the storage command `AcquirePessimisticLock` contains an enum to determine whether it's executing a normal request or it's a batch of requests resumed after waiting for another lock.
```

……but I doubt that whether it would be better if I add a new command instead of making `AcquirePessimisticLock` an enum inside

### Related changes

### Check List 

Tests 

- It will be covered by tests after another PR that makes use of it.

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
None
```


sticnarf:
> ……but I doubt that whether it would be better if I add a new command instead of making `AcquirePessimisticLock` an enum inside

I think it looks better to add a new command. Is there any difficulty doing this way?

MyonKeminta:
> > ……but I doubt that whether it would be better if I add a new command instead of making `AcquirePessimisticLock` an enum inside
> 
> I think it looks better to add a new command. Is there any difficulty doing this way?

No. At first I was trying to reuse most of the code, but finally it looks totally different. I'll split it into two commands.

src/storage/txn/commands/acquire_pessimistic_lock.rs
```diff
@@ -46,9 +46,9 @@ command! {
             /// later read in the same transaction.
             return_values: bool,
             min_commit_ts: TimeStamp,
-            old_values: OldValues,
```
TonsnakeLin, line 49: Why removed the field `old_values`? Do we still support this feature of `old_value`?
um..., I know little about `old_value`.

src/storage/txn/commands/acquire_pessimistic_lock.rs
```diff
@@ -69,27 +69,28 @@ impl CommandExt for AcquirePessimisticLock {
...
-        )))))) => info,
-        _ => panic!("unexpected mvcc error"),
-    }
-}
-
 impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for AcquirePessimisticLock {
-    fn process_write(mut self, snapshot: S, context: WriteContext<'_, L>) -> Result<WriteResult> {
+    fn process_write(self, snapshot: S, context: WriteContext<'_, L>) -> Result<WriteResult> {
+        if self.allow_lock_with_conflict && self.keys.len() > 1 {
+            // Currently multiple keys with `allow_lock_with_conflict` set is not supported.
```
TonsnakeLin, line 75: Do we support a single request which has multiple keys and `allow_lock_with_conflict` set finally ?

MyonKeminta: Maybe we'll supported it in the future. It depends.

src/storage/txn/commands/acquire_pessimistic_lock.rs
MyonKeminta: It seems that the old_value is never assigned as any valid value outside the command, and its only purpose is to hold the results produced during the pessimistic lock command and return it after finishing the command, which seems can be done by a local variable.

src/storage/txn/commands/acquire_pessimistic_lock.rs
```diff
@@ -104,101 +105,96 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for AcquirePessimisticLock
...
+                        start_ts,
+                        lock_ttl: self.lock_ttl,
+                        for_update_ts: self.for_update_ts,
+                        wait_timeout: self.wait_timeout,
+                        return_values: self.return_values,
+                        min_commit_ts: self.min_commit_ts,
+                        check_existence: self.check_existence,
+                        is_first_lock: self.is_first_lock,
+                        lock_only_if_exists: self.lock_only_if_exists,
+                        allow_lock_with_conflict: false,
```
TonsnakeLin, line 133: Always Set `allow_lock_with_conflict` to false instead of self.allow_lock_with_conflict?

MyonKeminta: Sorry, it's a mistake.

src/storage/txn/commands/mod.rs
```diff
@@ -389,7 +392,7 @@ pub struct WriteResult {
     pub to_be_write: WriteData,
     pub rows: usize,
     pub pr: ProcessResult,
-    pub lock_info: Option<WriteResultLockInfo>,
+    pub lock_info: Vec<WriteResultLockInfo>,
```
sticnarf, line 395: What will be the behavior when we encounter multiple locks in the future?

src/storage/txn/commands/macros.rs
```diff
@@ -183,6 +183,13 @@ macro_rules! gen_lock {
             crate::storage::txn::latch::Lock::new(keys)
         }
     };
+    ($field: ident: enum_match { $( $arm: pat => $expr: expr ), * }) => {
+        fn gen_lock(&self) -> crate::storage::txn::latch::Lock {
+            match &self.$field {
+                $( $arm => crate::storage::txn::latch::Lock::new($expr) ),*
+            }
+        }
+    }
```
sticnarf, line 192: Where is this rule used?

src/storage/txn/commands/acquire_pessimistic_lock_resumed.rs
```diff
@@ -0,0 +1,211 @@
...
+            {
+                if let Some(txn) = txn {
+                    if !txn.is_empty() {
+                        modifies.extend(txn.into_modifies());
+                    }
+                }
+                txn = Some(MvccTxn::new(
+                    params.start_ts,
+                    context.concurrency_manager.clone(),
+                ));
```
sticnarf, line 97:
```suggestion
                if let Some(txn) = txn.replace(MvccTxn::new(...)) {
                    if !txn.is_empty() {
                        modifies.extend(txn.into_modifies());
                    }
                }
```

sticnarf: `replace` is a bit simpler. Same for the `reader` below.

sticnarf: Is `!txn.is_empty()` necessary? IMO the code will look a bit simpler if removing the `if`.

src/storage/txn/commands/acquire_pessimistic_lock_resumed.rs
```diff
@@ -0,0 +1,211 @@
...
+            ) {
+                Ok((key_res, old_value)) => {
+                    res.push(key_res);
+                    new_locked_keys.push((params.start_ts, key.clone()));
+                    if old_value.resolved() {
+                        let key = key.append_ts(txn.start_ts);
+                        // MutationType is unknown in AcquirePessimisticLock stage.
+                        let mutation_type = None;
+                        old_values.insert(key, (old_value, mutation_type));
+                    }
```
sticnarf, line 135: Seems we can extract this common code from acquire_pessimistic_lock, acquire_pessimistic_lock_resumed and prewrite.

src/storage/txn/commands/macros.rs
MyonKeminta: Ah sorry. This is not used now and I forgot to delete it.

sticnarf:
Please fix the conflict.

MyonKeminta:
I'll fix conflicts after https://github.com/tikv/tikv/pull/13777 is merged.

sticnarf (approved):

TonsnakeLin (approved):

src/storage/mod.rs
```diff
@@ -1452,6 +1452,18 @@ impl<E: Engine, L: LockManager, F: KvFormat> Storage<E, L, F> {
                 )?;
                 check_key_size!(keys, self.max_key_size, callback);
             }
+            Command::AcquirePessimisticLockResumed(AcquirePessimisticLockResumed {
```
cfzjywxk, line 1455: Should we make the command name consistent with the "wakeup mode" in kv proto in the later pr?

src/storage/txn/commands/acquire_pessimistic_lock_resumed.rs
```diff
@@ -0,0 +1,414 @@
...
+                ResumedPessimisticLockItem {
+                    key: item.key,
+                    should_not_exist: item.should_not_exist,
+                    params: item.parameters,
+                    lock_wait_token: item.lock_wait_token,
+                }
+            })
+            .collect();
+
+        assert!(!items.is_empty());
```
cfzjywxk, line 192: By now should we use `assert_eq(items.len(), 1)`?

cfzjywxk (approved):
LGTM

src/storage/mod.rs
MyonKeminta: Sorry but I didn't quite understand. What's your suggested name of this command?

cfzjywxk: Since it seems we're not going to use the "resume" here(https://github.com/pingcap/kvproto/pull/872/files), should we change the name here to "Force" too?

MyonKeminta: This command is not used at first. When a request arrives TiKV, it's handled in `AcquirePessimisticLock` firstly; if it's blocked by lock, after waiting and woken up, this new command will be used. Therefore I think `resumed` is a proper word for it, but the behavior described by the word `Force` (which I think means the same thing as allow_lock_with_conflict) exists in both `AcquirePessimisticLock` and this new command.

PR #13689, opened 2022-10-28
raftstore-v2: support tablet split - part 1

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

Enable TiKV to split region in raftstore v2.

This PR implements the propose and apply part of the tablet split:

**There are some changes compared to the split in v1:**
Each region has a separate rocksdb, namely tablet, in v2, so the tablet will be split too.
Now, for simplicity, we use the checkpoint API of the RocksDb to clone the parent peer's tablet for split peers with temporary split path. The state information of new regions will not be persisted in this phase, whereas v1 will persist them. So, if split fails, just delete these split paths and redo the split.

The apply of the split result will be done in the following PRs.

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore-v2/src/raft/apply.rs
```diff
@@ -17,11 +22,15 @@ use crate::{
 
 /// Apply applies all the committed commands to kv db.
 pub struct Apply<EK: KvEngine, R> {
+    store_id: u64,
+
     peer: metapb::Peer,
     remote_tablet: CachedTablet<EK>,
     tablet: EK,
```
tabokie, line 29: Comment here when can `tablet != remote_tablet.cache()`.

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -106,6 +139,16 @@ impl<EK: KvEngine, R> Apply<EK, R> {
         &mut self.region_state
     }
 
+    #[inline]
+    pub fn set_region_state(&mut self, region_state: RegionLocalState) {
+        self.region_state = region_state;
+    }
+
+    #[inline]
+    pub fn tablet(&mut self) -> Option<EK> {
```
tabokie, line 148: The dynamics between `self.tablet` and `remote_tablet` is confusing. From the code I have no idea how to choose between `self.tablet`, `remote_tablet.cache()`, `remote_tablet.latest()`.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,533 @@
...
+
+use crate::{
+    batch::StoreContext,
+    fsm::{PeerFsm, PeerFsmDelegate},
+    operation::AdminCmdResult,
+    raft::{write_initial_states, Apply, Peer, Storage},
+    router::{ApplyRes, PeerMsg, StoreMsg},
+};
+
+fn create_checkpoint(
```
tabokie, line 42: rm

components/raftstore-v2/src/raft/apply.rs
SpadeA-Tang: I don't know why we need filed `tablet`. The code before uses it to generate write batch if self.write_batch is none in apply thread. But raftstore can update the remote_tablet (ex: when parent peer change from split_x_y to _x_y) so in apply thread, it should always check whether the tablet has a newer version. @BusyJay

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/raft/apply.rs
BusyJay: No, `remote_tablet` is only used for publishing new tablets. Check `publish_tablet`. You should always use `self.tablet` instead of querying remote.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -451,18 +481,7 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
-            if let Err(e) = wb.write_opt(&write_opt) {
-                panic!("failed to write data: {:?}", self.logger.list());
-            }
-            if wb.data_size() <= APPLY_WB_SHRINK_SIZE {
-                wb.clear();
-            } else {
-                self.write_batch_mut().take();
-            }
-        }
+        self.flush_write();
```
BusyJay, line 484: Why the change?

components/raftstore-v2/src/operation/command/admin/mod.rs
```diff
@@ -72,7 +75,7 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
             self.propose_conf_change(ctx, req)
         } else {
             // propose other admin command.
-            unimplemented!()
+            self.propose_command(ctx, req)
```
BusyJay, line 78: Split should have its own customization.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,527 @@
...
+impl<EK: KvEngine, R> Apply<EK, R> {
+    pub fn exec_split(
+        &mut self,
+        req: &AdminRequest,
+        log_index: u64,
+    ) -> Result<(AdminResponse, AdminCmdResult)> {
+        info!(
+            self.logger,
+            "split is deprecated, redirect to use batch split";
+            "region_id" => self.region_state().get_region().id,
```
BusyJay, line 58: Redundant fields.

components/raftstore-v2/src/operation/command/mod.rs
SpadeA-Tang: Before creating checkpoint, the write batch should be flushed. Extract this so that exec_batch_split can use it.

components/raftstore-v2/src/operation/command/admin/mod.rs
SpadeA-Tang: why?

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Done

components/raftstore-v2/src/operation/command/admin/mod.rs
```diff
@@ -72,7 +75,10 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
             self.propose_conf_change(ctx, req)
         } else {
             // propose other admin command.
-            unimplemented!()
+            match cmd_type {
+                AdminCmdType::Split | AdminCmdType::BatchSplit => self.propose_split(ctx, req),
```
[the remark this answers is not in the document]
BusyJay: Split keys should be validated before being proposed.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,530 @@
...
+            .create_at(&derived_path, None, 0)
+            .unwrap_or_else(|e| {
+                panic!(
+                    "{:?} fails to create checkpoint with path {:?}: {:?}",
+                    self.logger.list(),
+                    derived_path,
+                    e
+                )
+            });
+        // Here, we open the tablet without registering as it is the split version. We
```
BusyJay, line 175: It should be registered immediately.

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -106,6 +130,16 @@ impl<EK: KvEngine, R> Apply<EK, R> {
         &mut self.region_state
     }
 
+    #[inline]
+    pub fn set_region_state(&mut self, region_state: RegionLocalState) {
```
BusyJay, line 134: Seems not used.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -47,6 +47,7 @@ use crate::{
 
 /// A per-thread context shared by the [`StoreFsm`] and multiple [`PeerFsm`]s.
 pub struct StoreContext<EK: KvEngine, ER: RaftEngine, T> {
+    pub store_id: u64,
```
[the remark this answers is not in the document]
SpadeA-Tang: It's not needed now. I will remove it.

components/raftstore-v2/src/fsm/apply.rs
```diff
@@ -57,14 +58,24 @@ pub struct ApplyFsm<EK: KvEngine, R> {
 
 impl<EK: KvEngine, R> ApplyFsm<EK, R> {
     pub fn new(
+        store_id: u64,
```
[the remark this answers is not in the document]
SpadeA-Tang: ditto

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,530 @@
...
+                regions,
+                derived,
+                tablet_index: log_index,
+            }),
+        ))
+    }
+}
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    pub fn propose_split<T>(
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

components/raftstore-v2/src/operation/command/admin/mod.rs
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/raft/apply.rs
SpadeA-Tang: It's not needed now.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,624 @@
...
+        return Err(box_err!(
+            "cmd_type is BatchSplit but it doesn't have splits request, message maybe \
+             corrupted!"
+                .to_owned()
+        ));
+    }
+    let split_reqs: &[SplitRequest] = req.mut_splits().get_requests();
+    let count = split_reqs.len();
+    for i in 1..count {
+        if split_reqs[i - 1].get_split_key() >= split_reqs[i].get_split_key() {
```
tonyxuqqi, line 55: should check split is not empty?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,530 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
```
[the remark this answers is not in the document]
SpadeA-Tang: Done for this part.

components/raftstore-v2/src/raft/apply.rs
SpadeA-Tang: Yes,  it is not needed now. I removed it.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: `should check split_key is not empty?`
It is checked at the beginning.

SpadeA-Tang: > Also should check "req.get_new_peer_ids().len() == region_to_split.get_peers().len()"

This code is to temporarily do some works of the split CoprocessorHost which does not check for it. Anyway, it will be checked in exec_batch_split.

components/raftstore-v2/src/fsm/apply.rs
```diff
@@ -4,14 +4,15 @@ use std::{
     pin::Pin,
     sync::{
         atomic::{AtomicBool, Ordering},
-        Arc,
+        Arc, Mutex,
```
BusyJay, line 7: Useless import.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,673 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+//! This module contains batch split related processing logic.
```
BusyJay, line 2: Missing blank line.

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -1,8 +1,13 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::mem;
+use std::{
+    collections::VecDeque,
```
BusyJay, line 4: Useless imports.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -425,6 +441,25 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
                     return Err(box_err!("invalid admin command type"));
                 }
             };
+            // Check if the region epoch changes as expected
```
BusyJay, line 444: This should be a unit test instead of code for production.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,673 @@
...
+};
+
+#[derive(Debug)]
+pub struct SplitResult {
+    pub regions: Vec<Region>,
+    pub derived: Region,
+    pub tablet_index: u64,
+}
+
+pub fn validate_batch_split(
```
BusyJay, line 63: How about making `apply::extract_split_keys` reuse the function?

components/raftstore-v2/src/fsm/apply.rs
SpadeA-Tang: removed

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Added.

components/raftstore-v2/src/raft/apply.rs
SpadeA-Tang: Removed

components/raftstore-v2/src/operation/command/mod.rs
SpadeA-Tang: Removed

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good suggestion.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1899,40 +1899,52 @@ mod confchange_cmd_metric {
...
-                .back()
-                .map_or_else(|| region_to_split.get_start_key(), Vec::as_slice)
-        {
-            return Err(box_err!("invalid split request: {:?}", split_reqs));
-        }
-        if req.get_new_peer_ids().len() != region_to_split.get_peers().len() {
+
+    let split_reqs: &[SplitRequest] = req.get_splits().get_requests();
+    let count = split_reqs.len();
+    if split_reqs[0].get_new_peer_ids().len() != region.get_peers().len() {
```
BusyJay, line 1909: You can use a temporary variable like:
```
let mut last_key = region.get_start_key();
for req in split_reqs {
    if req.get_split_key() <= last_key {
        ...
    }
    last_key = req.get_split_key();
}
check_exclusive(last_key)
```

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,630 @@
...
+            let mut split_req = SplitRequest::default();
+            split_req.set_split_key(key);
+            split_req.set_new_peer_ids(vec![1]);
+            req.mut_splits().mut_requests().push(split_req);
+        }
+        req
+    }
+
+    #[test]
+    fn test_validate_batch_split() {
```
BusyJay, line 552: The unit test can also be moved to v1.

components/raftstore/src/store/fsm/apply.rs
SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1899,40 +1899,45 @@ mod confchange_cmd_metric {
...
 }
 
-// Validate the request and the split keys
-pub fn extract_split_keys(
-    split_reqs: &BatchSplitRequest,
-    region_to_split: &Region,
-) -> Result<VecDeque<Vec<u8>>> {
-    if split_reqs.get_requests().is_empty() {
+pub fn validate_batch_split(req: &AdminRequest, region: &Region) -> Result<()> {
+    if !req.has_splits() || req.get_splits().get_requests().is_empty() {
```
BusyJay, line 1903: Redundant check for `has_splits`.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -6693,4 +6699,95 @@ mod tests {
...
+        req = new_batch_split_request(vec![b"k06".to_vec(), b"k07".to_vec(), b"k11".to_vec()]);
+        assert!(
+            validate_batch_split(&req, &region)
+                .unwrap_err()
+                .to_string()
+                .contains("not in region")
+        );
+
+        req = new_batch_split_request(vec![b"k06".to_vec(), b"k07".to_vec(), b"k08".to_vec()]);
+        validate_batch_split(&req, &region).unwrap();
```
BusyJay, line 6791: What about out of order error and duplicated key error?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,542 @@
...
+        proposal_ctx.insert(ProposalContext::SYNC_LOG);
+        proposal_ctx.insert(ProposalContext::SPLIT);
+
+        let data = req.write_to_bytes().unwrap();
+        self.propose_with_proposal_ctx(store_ctx, data, proposal_ctx.to_vec())
+    }
+}
+
+impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
+    pub fn exec_batch_split(
```
BusyJay, line 81: Missing `split` implementation. Although it will not be proposed by new versions, but it may be proposed already by old version.

components/raftstore/src/store/fsm/apply.rs
SpadeA-Tang: Good catch.  get_splits() will return a default if not set.

components/raftstore/src/store/fsm/apply.rs
SpadeA-Tang: Added.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,563 @@
...
+        let split_reqs = req.get_splits();
+        let new_region_cnt = split_reqs.get_requests().len();
+        let new_version = derived.get_region_epoch().get_version() + new_region_cnt as u64;
+        derived.mut_region_epoch().set_version(new_version);
+
+        let right_derive = split_reqs.get_right_derive();
+        let mut regions = Vec::with_capacity(new_region_cnt + 1);
+        // Note that the split requests only contain ids for new regions, so we need
+        // to handle new regions and old region separately.
+        if right_derive {
```
BusyJay, line 129: This is a bit complicated. How about:
```
let mut boundaries: Vec<&[u8]> = Vec::with_capacity();
boundaries.push(self.region().get_start());
extract_split_keys(&mut boundaries);
boundaries.push(self.region().get_end());

let mut derived_req = SplitReq::default();
derived_req.new_region_id = self.region().id;
derived_req.new_peer_ids = self.region.get_peers().map(|p| p.get_id()).collect();

let reqs = if right_derive {
    split_reqs.iter().chain(&[derived_req])
} else {
    &[derived_req].chain(split_reqs)
};

let regions = boundaries.windows(2).zip(reqs).map(|(start_key, end_key, req)| ...).collect();
```

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -405,8 +417,9 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
             let admin_req = req.get_admin_request();
             let (admin_resp, admin_result) = match req.get_admin_request().get_cmd_type() {
                 AdminCmdType::CompactLog => unimplemented!(),
-                AdminCmdType::Split => unimplemented!(),
-                AdminCmdType::BatchSplit => unimplemented!(),
+                // The proposal the Split will be rejected.
```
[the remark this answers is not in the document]
SpadeA-Tang: good catch.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -405,8 +417,9 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+                AdminCmdType::Split => self.exec_split(admin_req, entry.index)?,
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: good suggestion.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,569 @@
...
+        );
+
+        let split_reqs = req.get_splits();
+        let new_region_cnt = split_reqs.get_requests().len();
+        let new_version = derived.get_region_epoch().get_version() + new_region_cnt as u64;
+        derived.mut_region_epoch().set_version(new_version);
+
+        let mut derived_req = SplitRequest::default();
+        derived_req.new_region_id = derived.id;
+        derived_req.new_peer_ids = derived.get_peers().iter().map(|p| p.get_id()).collect();
```
BusyJay, line 132: Or maybe just not assign anything to `new_peer_ids`.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,569 @@
...
+                    .zip(req.get_new_peer_ids())
+                {
+                    peer.set_id(*peer_id);
+                }
+                new_region
+            })
+            .collect();
+
+        // Amend the boundary of the derived region
+        if right_derive {
```
BusyJay, line 164: It should be either regions.first() or regions.last().

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,569 @@
...
+    fsm::ApplyResReporter,
+    operation::AdminCmdResult,
+    raft::{Apply, Peer},
+    router::ApplyRes,
+};
+
+#[derive(Debug)]
+pub struct SplitResult {
+    pub regions: Vec<Region>,
+    pub derived: Region,
```
BusyJay, line 62: Instead of returning the a duplicated region, I think an index to show which region in `regions` is derived should be sufficient.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,569 @@
...
+            });
+        let tablet = self
+            .tablet_factory()
+            .open_tablet(region_id, Some(log_index), OpenOptions::default())
+            .unwrap();
+        self.publish_tablet(tablet);
+
+        self.region_state_mut().set_region(derived.clone());
+        self.region_state_mut().set_tablet_index(log_index);
+        // self.metrics.size_diff_hint = 0;
```
BusyJay, line 232: Should be deleted.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,569 @@
...
+        right_derived: bool,
+        new_region_ids: Vec<u64>,
+        split_keys: Vec<Vec<u8>>,
+        children_peers: Vec<Vec<u64>>,
+        log_index: u64,
+    ) -> HashMap<u64, Region> {
+        let mut splits = BatchSplitRequest::default();
+        splits.set_right_derive(right_derived);
+        let mut region_boundries = split_keys.clone();
+        region_boundries.insert(0, region_to_split.get_start_key().to_vec());
```
BusyJay, line 324: Pointless to repeat what has done in apply. You should use table based test cases and specify boundaries explicitly.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,569 @@
...
+            self.logger,
+            "split region";
+            "region" => ?derived,
+            "boundaries" => %KeysInfoFormatter(boundaries.iter()),
+        );
+
+        let split_reqs = req.get_splits();
+        let new_region_cnt = split_reqs.get_requests().len();
+        let new_version = derived.get_region_epoch().get_version() + new_region_cnt as u64;
+        derived.mut_region_epoch().set_version(new_version);
```
BusyJay, line 128: I don't think `derived` needs to be created explicitly at all.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: I think it's not intuitive.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: At lease, it's convinient to update the region state at the end of the apply.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2403,67 +2408,72 @@ where
...
-            "keys" => %KeysInfoFormatter(keys.iter()),
+            "boundaries" => %KeysInfoFormatter(boundaries.iter()),
         );
 
+        let split_reqs = req.get_splits();
         let new_region_cnt = split_reqs.get_requests().len();
         let new_version = derived.get_region_epoch().get_version() + new_region_cnt as u64;
         derived.mut_region_epoch().set_version(new_version);
 
+        let mut derived_req = SplitRequest::default();
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Done

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,571 @@
...
+        req.set_splits(splits);
+
+        // Exec batch split
+        let (resp, apply_res) = apply.apply_batch_split(&req, log_index).unwrap();
+
+        let regions = resp.get_splits().get_regions();
+        assert!(regions.len() == region_boundries.len());
+
+        let mut epoch = region_to_split.get_region_epoch().clone();
+        epoch.version += children_peers.len() as u64;
```
BusyJay, line 345: new_version should also be specified explicitly.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,571 @@
...
+        }
+
+        regions
+            .iter()
+            .map(|region| (region.id, region.clone()))
+            .collect()
+    }
+
+    #[test]
+    fn test_split() {
```
BusyJay, line 379: Don't write `assert_split` again and again, should use table test instead.

And derived_index should also be tested.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -6693,4 +6690,111 @@ mod tests {
...
+        req.set_split(split_req);
+        // Split is deprecated
+        assert!(
+            validate_batch_split(&req, &region)
+                .unwrap_err()
+                .to_string()
+                .contains("missing split requests")
+        );
+
+        req = new_batch_split_request(vec![vec![]]);
```
BusyJay, line 6734: Should use table test instead.

BusyJay (commented):
Rest LGTM

components/raftstore-v2/src/operation/command/admin/split.rs
tonyxuqqi: > > should check split_key is not empty?
> 
> It is checked at the beginning.

Do you mean this?
```
 if split_reqs.get_requests().is_empty() {
        return Err(box_err!("missing split requests"));
    }
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -6693,4 +6690,97 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_validate_batch_split() {
+        let mut region = Region::default();
+        region.set_start_key(b"k05".to_vec());
+        region.set_end_key(b"k10".to_vec());
+        region.set_peers(vec![new_peer(1, 2)].into());
+
+        let fail_cases = vec![
```
tonyxuqqi, line 6713: 这里fail_cases看上去有三类，一个是split is deprecated，就是第一个case，第二个case是peer id缺失的情况，
其余是各种split key不对的情况。
现在是强行把这三种情况合在一起了，所以每个case里就必须实现所有的代码。
其实如果把这三类分开的话，只是在第三种情况用for 循环，这样那段代码可以简洁一些。另外出错的error可以用一个变量保存，免得以后有修改需要修改一片。
说到table test，一般这个table里是没有代码只有数据的，代码都是在for循环里统一的。

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -6693,4 +6690,97 @@ mod tests {
...
+            (
+                {
+                    let mut req = AdminRequest::default();
+                    req.set_cmd_type(AdminCmdType::Split);
+                    let mut split_req = SplitRequest::default();
+                    split_req.set_split_key(b"k06".to_vec());
+                    req.set_split(split_req);
+                    req
+                },
+                "missing split requests",
```
tonyxuqqi, line 6726: all error messages should use variable as they're mostly duplicated.

components/raftstore-v2/src/operation/command/mod.rs
```diff
@@ -183,16 +184,26 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
 
     #[inline]
     fn propose<T>(&mut self, ctx: &mut StoreContext<EK, ER, T>, data: Vec<u8>) -> Result<u64> {
-        ctx.raft_metrics.propose.normal.inc();
+        self.propose_with_proposal_ctx(ctx, data, vec![])
```
tonyxuqqi, line 187: the name propose_with_proposal_ctx can be renamed as propose_with_ctx

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,642 @@
...
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    pub fn propose_split<T>(
+        &mut self,
+        store_ctx: &mut StoreContext<EK, ER, T>,
+        mut req: RaftCmdRequest,
+    ) -> Result<u64> {
+        validate_batch_split(req.mut_admin_request(), self.region())?;
+        let mut proposal_ctx = ProposalContext::empty();
+        proposal_ctx.insert(ProposalContext::SYNC_LOG);
+        proposal_ctx.insert(ProposalContext::SPLIT);
```
tonyxuqqi, line 76: Question: Where's the code that use ProposalContext::SYNC_LOG and ProposalContext::SPLIT? They're used as arbitrary bytes context in rust lib, but I don't find any one really consume these flags.

components/raftstore/src/store/fsm/apply.rs
SpadeA-Tang: Done

components/raftstore/src/store/fsm/apply.rs
SpadeA-Tang: Okay，done

components/raftstore-v2/src/operation/command/mod.rs
SpadeA-Tang: done

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: I don't know where it is used either. It is this in v1.

SpadeA-Tang:
PTAL again @tonyxuqqi @BusyJay

tonyxuqqi (approved):

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,621 @@
...
+        );
+        let split = req.get_split().to_owned();
+        let mut admin_req = AdminRequest::default();
+        admin_req
+            .mut_splits()
+            .set_right_derive(split.get_right_derive());
+        admin_req.mut_splits().mut_requests().push(split);
+        // This method is executed only when there are unapplied entries after being
+        // restarted. So there will be no callback, it's OK to return a response
+        // that does not matched with its request.
```
tabokie, line 101: I don't understand this comment. Where is the callback placed if there's one? Why response doesn't match the request in batch_split mode?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,621 @@
...
+            })
+            .collect();
+
+        let derived_index = if right_derive { regions.len() - 1 } else { 0 };
+
+        // We will create checkpoint of the current tablet for both derived region and
+        // split regions. Before the creation, we should flush the writes and remove the
+        // write batch
+        self.flush();
+        self.write_batch_mut().take();
```
tabokie, line 176: Why is this necessary?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,621 @@
...
+
+        // todo(SpadeA): Here: we use a temporary solution that we use checkpoint API to
```
tabokie, line 178: More info on this: The flush behavior is dependent on the value of `log_size_for_flush` (alternatively it copies the WAL). After we removed WAL, the native API will never flush. It would then become **incorrect** to use checkpoint anymore.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -0,0 +1,621 @@
...
+            )
+        });
+
+        for new_region in &regions {
+            let new_region_id = new_region.id;
+            if new_region_id == region_id {
+                continue;
+            }
+
+            let split_temp_path = self.tablet_factory().tablet_path_with_prefix(
```
tabokie, line 197: Where is this tablet published?

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: The wb now may be belong to the old tablet, but we have new tablet now. If it is not taken, then the future writes are still written in the old tablet.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: In the next PR, after the initialization of the split peer.

components/raftstore-v2/src/operation/command/admin/split.rs
tabokie: Should add a new comment. Because the reasoning is different from "persist all data to disk".

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: It is copied from v1.

components/raftstore-v2/src/operation/command/admin/split.rs
tabokie: BTW, why need to replace the leftmost tablet at all? Couldn't we reuse the source tablet?

SpadeA-Tang: > BTW, why need to replace the leftmost tablet at all? Couldn't we reuse the source tablet?

tablet suffix will be changed.

SpadeA-Tang: > Should add a new comment. Because the reasoning is different from "persist all data to disk".

Ok, done.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: I will point it out.

components/raftstore-v2/src/operation/command/admin/split.rs
tabokie: That's not the point, technically you can still use the same tablet and insert it as a new suffix. For reads on the old tablet, we can make sure its snapshot timestamp never exceeds the current maximum sequence number.

SpadeA-Tang: > That's not the point, technically you can still use the same tablet and insert it as a new suffix. For reads on the old tablet, we can make sure its snapshot timestamp never exceeds the current maximum sequence number.

Split here is not done. If load it with the new suffix, and the split failed, there may be some problem. In this implementation, if that case happens, just remove the these tablets.

tabokie: Fair

tabokie (approved):

PR #13693, opened 2022-10-31
*: prepare for raftstore-v2's split

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

This PR add a new API `create_checkpoint` which uses `create_checkpoint` of the rocksdb.
It also refactors a small part of v1's split in order to reduce code duplication when implementing v2.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_rocks/src/checkpoint.rs
+++ b/components/engine_rocks/src/checkpoint.rs
@@ -22,12 +22,12 @@ pub struct RocksEngineCheckpointer(rocksdb::Checkpointer);
 impl Checkpointer for RocksEngineCheckpointer {
     fn create_at(
         &mut self,
-        basedb_out_dir: &Path,
+        db_out_dir: &Path,
         titan_out_dir: Option<&Path>,
         log_size_for_flush: u64,
     ) -> Result<()> {
         self.0
-            .create_at(basedb_out_dir, titan_out_dir, log_size_for_flush)
+            .create_at(db_out_dir, titan_out_dir, log_size_for_flush)
             .map_err(|e| r2e(e))
     }
 }
--- a/components/engine_rocks/src/checkpoint.rs
+++ b/components/engine_rocks/src/checkpoint.rs
@@ -1,0 +1,38 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+use engine_traits::{Checkpoint, Code, Error, Result, Status};
+use rocksdb::Checkpointer;
+
+use crate::RocksEngine;
+
+impl Checkpoint for RocksEngine {
+    type Checkpointer = Checkpointer;
+
+    fn new_checkpointer(&self) -> Result<Self::Checkpointer> {
+        self.as_inner()
+            .new_checkpointer()
+            .map_err(|e| Error::Engine(Status::with_error(Code::IoError, e)))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use engine_traits::{Checkpoint, Peekable, SyncMutable, ALL_CFS};
+    use tempfile::tempdir;
+
+    use crate::util::new_engine;
+
+    #[test]
+    fn test_checkpoint() {
+        let dir = tempdir().unwrap();
+        let path = dir.path().join("origin");
+        let engine = new_engine(path.as_path().to_str().unwrap(), ALL_CFS).unwrap();
+        engine.put(b"key", b"value").unwrap();
+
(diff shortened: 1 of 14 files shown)


components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1899,6 +1899,42 @@ mod confchange_cmd_metric {
     }
 }
 
+// Validate the request and the split keys
+pub fn validate_and_get_split_keys(
```
BusyJay, line 1903:
```suggestion
pub fn extract_split_keys(
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -1899,6 +1899,42 @@ mod confchange_cmd_metric {
...
+            return Err(box_err!(
+                "invalid new peer id count, need {:?}, but got {:?}",
+                region_to_split.get_peers(),
+                req.get_new_peer_ids()
+            ));
+        }
+        keys.push_back(split_key.to_vec());
+    }
+
+    util::check_key_in_region(keys.back().unwrap(), region_to_split)?;
```
BusyJay, line 1933: Seems `check_key_in_region_exclusive` makes more sense. Though both functions work the same here.

components/engine_traits/src/misc.rs
```diff
@@ -95,4 +97,6 @@ pub trait MiscExt: CfNamesExt + FlowControlFactorsExt {
     ) -> Result<Option<(u64, u64)>>;
 
     fn is_stalled_or_stopped(&self) -> bool;
+
+    fn create_checkpoint(&self, basedb_out_dir: &Path) -> result::Result<(), String>;
```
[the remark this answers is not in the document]
SpadeA-Tang: Good suggestion.

components/raftstore/src/store/fsm/apply.rs
SpadeA-Tang: done

components/engine_traits/src/checkpoint.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+use crate::Result;
+
+pub trait Checkpoint {
+    type Checkpointer;
```
BusyJay, line 6: How to use `Checkpointer`?

components/engine_traits/src/checkpoint.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+use crate::Result;
+
+pub trait Checkpoint {
```
BusyJay, line 5: It should be named `Checkpointable` just like `Writeable`.

SpadeA-Tang: Done

components/engine_traits/src/checkpoint.rs
```diff
@@ -0,0 +1,20 @@
...
+pub trait Checkpointable {
+    type Checkpointer: Checkpointer;
+
+    fn new_checkpointer(&self) -> Result<Self::Checkpointer>;
+}
+
+pub trait Checkpointer {
+    fn create_at(
+        &mut self,
+        basedb_out_dir: &Path,
```
BusyJay, line 16:
```suggestion
        db_out_dir: &Path,
```

BusyJay (approved):
Rest LGTM

components/engine_rocks/src/checkpoint.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::path::Path;
+
+use engine_traits::{Checkpointable, Checkpointer, Result};
+
+use crate::{r2e, RocksEngine};
+
+impl Checkpointable for RocksEngine {
```
tabokie, line 9: RocksDB interface is too cumbersome, the overhead of `new_checkpointer` is negligible. Let's just expose only one trait and one interface `create_checkpoint`.

SpadeA-Tang: How about your suggestions about this? @BusyJay

BusyJay: I prefer to keep align with rocksdb's API.

tabokie (approved):

PR #13696, opened 2022-10-31
pd-client: pd client version 2

### LINKED ISSUE #13673: Refine PD client
## Development Task

PD client (and PD worker) has been too complicated to be maintained in the past years. There are subtle bugs due to its complicated reconnect mechanism and lock usage. I propose to refine the implementations by following changes:

0. Introduce a struct similar to `CachedTablet` from raftstore v2 named `CachedClient` (perhaps we can share the implementation). `CachedClient` wraps around native `RpcClient` and shared between threads. 
1. Introduce `wait_for_ready` API for CachedClient. The API is similar to native `RpcClient` that it will return when the client is ready to make probably successful requests. The API should first check whether the cached client needs to be updated, then check if native RpcClient is ready.
2. All async/sync APIs should call `wait_for_ready` first, and then makes further query. In this case, we can remove the retry mechanism as the request should probably succeed.
3. For streaming APIs like heartbeats and buckets, they should return a channel and refresh client in an async block like following:
    ```
    pub fn hearbeat(&self) -> Sender {
        let (tx, rx) = tikv_util::futures::bounded();
        let client = self.cached_client.clone();
        self.rpc_client.spawn(async move {
            loop {
                client.latest().wait_for_ready().await;
                let (g_tx, g_rx) = client.heartbeat();
                select!(g_tx.send_all(rx).await, g_rx);
            }
        });
     }
    ```
4. Introducing a connectivity maintenance coroutine that will connect the RpcClient if it's not connected or the PD address changes. It means the connect step is done in the background when PD client is constructed.

Compared to the current implementation, the new implementation should be more readable and easier to understand. And there should be no deadlock problem.

It's OK to either implement the client as a new module or just mix with the existing module. Support to be used in v2 is the highest priority. We can integrate it with v1 once it's stable.

tabokie:

Issue Number: Close #13673 

What's Changed:

- Move reconnect logic to one single coroutine `reconnect_loop`
- Move streaming RPCs to individual long-running coroutines that are not destroyed even when connection is rebuilt
- All client methods take mutable reference now
- `PdConnector` and some other constants are shared with v1
- Old client tests are moved to xx_legacy.rs unchanged

```commit-message
None
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/pd_client/src/util.rs
```diff
@@ -598,18 +608,19 @@ impl PdConnector {
...
             Ok((
+                channel.clone(),
                 PdClientStub::new(channel.clone()),
                 GetMembersResponse::default(),
             ))
         });
-        let client = PdClientStub::new(channel);
-        let option = CallOption::default().timeout(Duration::from_secs(REQUEST_TIMEOUT));
+        let client = PdClientStub::new(channel.clone());
+        let option = CallOption::default().timeout(Duration::from_secs(REQUEST_TIMEOUT * 5));
```
tonyxuqqi, line 617: Is there any reason to change it to 5x of original timeout value?

tonyxuqqi (approved):
Overall LGTM. Only a few minor comments.

tonyxuqqi (commented):
It looks some PD related test failed.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1412 @@
...
+            recv_res = bk_rx.fuse() => {
+                warn!("region buckets stream existed: {:?}", recv_res);
+            }
+        }
+        let _ = bk_tx.close().await;
+        reconnect.send(client.cache_version()).await.unwrap();
+    }
+}
+
+async fn tso_loop(
```
BusyJay, line 1307: These are not part of the client, instead the client should provide a general interface to handle streaming calls.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1412 @@
...
+        Err(box_err!(
+            "Connection unavailable {:?}",
+            self.channel().check_connectivity_state(false)
+        ))
+    }
+
+    /// Increases global version only when a new connection is established.
+    async fn reconnect(&mut self, force: bool, callback: &CallbackHolder) -> Result<()> {
+        let mut latest = self.latest.lock().await;
+        if latest.maybe_reconnect(&self.context, force).await? {
```
BusyJay, line 188: If there is a single background corountine to manage the lifetime of client/channel, then it doesn't need to reconnect within lock.

components/pd_client/src/client_v2.rs
tabokie: Could you give me a better idea of the "interface"? `TsoRequestStream` for example is very different from heartbeat or bucket, it requires one-on-one response and uses batching.

components/pd_client/src/client_v2.rs
tonyxuqqi: When it's reconnect, it means the current version does not work anyway,  so it does no hurt to have a lock here?

BusyJay: Other API callers may be blocked for arbitrary duration.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+    mut client: CachedRawClient,
+    on_reconnect: CallbackHolder,
+    mut reconnect: mpsc::UnboundedReceiver<u64>,
+) {
+    let interval = Duration::from_secs(REQUEST_TIMEOUT);
+    loop {
+        while !client.channel().wait_for_connected(interval).await {
+            // wait_for_connected will attempt to connect internally. When it fails we
+            // should force reconnect.
+            if let Err(e) = client.reconnect(true, &on_reconnect).await {
```
BusyJay, line 1316: If leader is still the same and member is still the same, there is no reason to build a new client.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+                let force = client.channel().check_connectivity_state(true) == ConnectivityState::GRPC_CHANNEL_SHUTDOWN;
+                if let Err(e) = client.reconnect(force, &on_reconnect).await {
+                    warn!("failed to reconnect pd"; "err" => ?e);
+                }
+            },
+            v = reconnect.next() => {
+                if v.is_none() {
+                    break;
+                } else if v.unwrap() == client.cache_version() {
+                    if let Err(e) = client.reconnect(true, &on_reconnect).await {
```
BusyJay, line 1332: No need to force reconnect.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+        });
+
+        // -1 means the max.
+        let retries = match cfg.retry_max_count {
+            -1 => std::isize::MAX,
+            v => v.saturating_add(1),
+        };
+        let pd_connector = PdConnector::new(env.clone(), security_mgr.clone());
+        for i in 0..retries {
+            match pd_connector.validate_endpoints(cfg, false).await {
```
BusyJay, line 300: Initialize connection should also be put into singleton connection coroutine.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+        Ok(Self {
+            cluster_id,
+            raw_client,
+            feature_gate: Default::default(),
+            on_reconnect,
+            reconnect_tx,
+        })
+    }
+
+    pub fn handle_reconnect<F: Fn() + Sync + Send + 'static>(&self, f: F) {
```
BusyJay, line 347:
```suggestion
    pub fn set_on_reconnect<F: Fn() + Sync + Send + 'static>(&self, f: F) {
```

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+    }
+
+    /// Used in integration tests.
+    pub fn get_leader(&mut self) -> Member {
+        block_on(self.raw_client.wait_for_ready()).unwrap();
+        self.raw_client.leader()
+    }
+
+    /// Used in integration tests.
+    pub fn reconnect(&mut self) -> Result<()> {
```
BusyJay, line 370: How about guard it with testexport feature?

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+            fail::fail_point!("region_heartbeat_send_failed", |_| {
+                Err(grpcio::Error::RemoteStopped)
+            });
+            Ok((r, WriteFlags::default()))
+        });
+        self.raw_client.stub().spawn(async move {
+            loop {
+                if let Err(e) = raw_client.wait_for_ready().await {
+                    warn!("failed to acquire client for RegionHeartbeat stream"; "err" => ?e);
+                    let _ = GLOBAL_TIMER_HANDLE
```
BusyJay, line 514: I don't get it. What's the point of backoff? Isn't it waited in `wait_for_ready` already?

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1339 @@
...
+        atomic::{AtomicU64, Ordering},
+        Arc, Mutex as StdMutex,
+    },
+    time::Duration,
+    u64,
+};
+
+use fail::fail_point;
+use futures::{
+    channel::mpsc,
```
BusyJay, line 28: Why not use tikv_util's version? For example, how to achieve batch using futures' version?

tabokie: Okay, but for batching is it equivalent to futures' [ready_chunks](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.ready_chunks)?

components/pd_client/src/client_v2.rs
tabokie: I still don't understand why, the code isn't any fewer with it stuffed into the coroutine.

components/pd_client/src/client_v2.rs
tabokie: So you mean the `force=true` condition will never be used in v2? But why is it used in v1 (e.g. force reconnect when tso stream ended)?

components/pd_client/src/client_v2.rs
BusyJay: They are not the same. `tikv_util` will queue the items without notifying receiver, `ready_chunks` will always notify receiver, but receiver will try batch as much as possible. The former does the batch on sender side, the later does batch on receiver side. TSO can't do batching with the returned stream/sink unless batch capacity is passed as parameter. Batch on sender side is good for throughput technically, as it allows precise control on when a batch is finished.

components/pd_client/src/client_v2.rs
BusyJay: The goal is not to reduce lines of code, but to make the lifetime of a connection clear. Now it's split into different parts.

components/pd_client/src/client_v2.rs
BusyJay: Because in v1, we don't rely on the auto retry of grpc channel, instead, we control the retry by our own. In v2, we don't do retry any more if the target address is still the same.

components/pd_client/src/client_v2.rs
tabokie: Is there a way to spawn coroutine when the client doesn't exist yet? And are requests simply rejected if the client is not created yet? If so the behavior seems inferior.

BusyJay: > Is there a way to spawn coroutine when the client doesn't exist yet

I don't see how coroutine and client are related. The shared part should be RawClient.

The request should not be rejected. It should call `wait_for_ready` and if it's timeout, error is returned. If caller expects the client should be ready before serving, then it can call `wait_for_ready` directly.

tabokie: `reconnect_loop` is spawned into grpc pool, which is accessed via a grpc client. The question is how to access that pool without a client in the first place.

BusyJay: I see. Technically this can be solved by creating a client to arbitrary address. Though make `Env` be able to spawn a future is more natural solution.

tabokie: > It should call wait_for_ready and if it's timeout

How to wait for it exactly (without adding new synchronization primitive which IMO makes the code more complex with little benefit)? My early draft before uses the mutex to synchronize (i.e. lock the mutex and move it into initialization coroutine, so that consumer can wait on that mutex), which is too obscure so I went back with the blocking approach.

BusyJay: I added a lame channel to grpcio, it should help merging first step with the rest. https://github.com/tikv/grpc-rs/pull/597

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1305 @@
...
+            on_reconnect_notify,
+            cache: None,
+            cache_version: 0,
+        }
+    }
+
+    #[inline]
+    async fn refresh_cache(&mut self) -> bool {
+        if self.cache_version < self.version.load(Ordering::Acquire) {
+            let latest = self.latest.lock().await;
```
BusyJay, line 163: Seems use `std::sync::Mutex` is enough.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1305 @@
...
+            false
+        }
+    }
+
+    #[inline]
+    async fn publish_cache(&mut self) {
+        let latest_version = {
+            let mut latest = self.latest.lock().await;
+            *latest = self.cache.clone();
+            self.version.fetch_add(1, Ordering::AcqRel) + 1
```
BusyJay, line 177: Technically, `Relax` should be sufficient.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1305 @@
...
+}
+
+/// A shared [`RawClient`] with a local copy of cache.
+#[derive(Clone)]
+pub struct CachedRawClient {
+    context: Arc<ConnectContext>,
+
+    latest: Arc<Mutex<Option<RawClient>>>,
+    version: Arc<AtomicU64>,
+    on_reconnect_notify: Arc<Notify>,
```
BusyJay, line 142: Why use `Notify`?

```
    println!("sending notification");
    notify.notify_waiters();

    notify.notified().await;
    println!("received notification");
```

Will only print one line.

tabokie: You are right. Maybe a full-fledged condvar is needed after all. The broadcast channel you proposed earlier isn't very handy either, because its sender might be blocked by an idle receiver that doesn't attempt to consume the messages. A combination of tokio::watch and tokio::broadcast would be perfect here but it doesn't exist...

BusyJay: >  its sender might be blocked by an idle receiver 

Why? Isn't it just overwrite the oldest messages?

tabokie: I see.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1358 @@
...
+
+        info!("trying to update PD client done"; "spend" => ?start.saturating_elapsed());
+        Ok(true)
+    }
+}
+
+struct CachedRawClientCore {
+    context: ConnectContext,
+
+    latest: Arc<Mutex<Option<RawClient>>>,
```
BusyJay, line 138: Useless `Arc`.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1358 @@
...
+            )
+            .await
+            {
+                return Err(box_err!("Connection is not initialized in time"));
+            }
+            assert!(self.refresh_cache());
+            assert!(self.cache.is_some());
+        }
+        select! {
+            r = self.core.grab_a_channel().wait_for_connected(REQUEST_TIMEOUT).fuse() => {
```
BusyJay, line 257: As in most case, it should be connected, so you can create a short path by calling `check_connectivity_state`.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1358 @@
...
+            let state = ConnectivityState::GRPC_CHANNEL_READY;
+            // Checks for leader change periodically.
+            client
+                .channel()
+                .wait_for_state_change(state, cfg.update_interval.0)
+                .await;
+        }
+        match client.reconnect().await {
+            Ok(true) => {
+                info!("pd reconnected");
```
BusyJay, line 408: Seems it's already logged in `maybe_reconnect`.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1358 @@
...
+        let lame_client = PdClientStub::new(Channel::lame(env, "0.0.0.0:0"));
+        lame_client.spawn(reconnect_loop(raw_client.clone(), cfg.clone()));
+
+        Ok(Self {
+            raw_client,
+            feature_gate: Default::default(),
+        })
+    }
+
+    pub fn set_on_reconnect<F: Fn() + Sync + Send + 'static>(&self, f: F) {
```
BusyJay, line 444: Actually, what I mean is
```
pub fn subscribe_reconnect(&self) -> Receiver;
```

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1358 @@
...
+        wake_policy: mpsc::WakePolicy,
+    ) -> Result<(mpsc::Sender<TsoRequest>, Self::ResponseChannel<TsoResponse>)>;
+
+    fn load_global_config(&mut self, list: Vec<String>) -> PdFuture<HashMap<String, String>>;
+
+    fn watch_global_config(
+        &mut self,
+    ) -> Result<grpcio::ClientSStreamReceiver<pdpb::WatchGlobalConfigResponse>>;
+
+    fn get_cluster_id(&mut self) -> Result<u64>;
```
BusyJay, line 508: Why need to be a `Result`? And isn't it just the same as L476?

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1358 @@
...
+                    Poll::Pending => return Poll::Pending,
+                    // If it's None or there's error, we need to update receiver.
+                    _ => {}
+                }
+            }
+
+            let latest = self.latest.lock().unwrap().take();
+            self.cache = latest;
+            if self.cache.is_none() {
+                return Poll::Pending;
```
BusyJay, line 612: This is wrong. If it's None and return Pending, there will be no one to wake it up again. The latest should not be a lock, instead it should be a channel.

components/pd_client/src/client_v2.rs
tabokie: `wait_for_ready` can fail. The L476 version will simply panic. I'm removing the latter now.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+        })
+    }
+
+    #[inline]
+    pub fn subscribe_reconnect(&self) -> broadcast::Receiver<()> {
+        self.raw_client.clone().on_reconnect_rx
+    }
+
+    #[cfg(feature = "testexport")]
+    pub fn set_on_reconnect<F: Fn() + Sync + Send + 'static>(&self, f: F) {
```
BusyJay, line 485: Why not change the test to use `subscribe_reconnect`?

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+        wake_policy: mpsc::WakePolicy,
+    ) -> Result<(mpsc::Sender<TsoRequest>, Self::ResponseChannel<TsoResponse>)>;
+
+    fn load_global_config(&mut self, list: Vec<String>) -> PdFuture<HashMap<String, String>>;
+
+    fn watch_global_config(
+        &mut self,
+    ) -> Result<grpcio::ClientSStreamReceiver<pdpb::WatchGlobalConfigResponse>>;
+
+    fn get_cluster_id(&mut self) -> Result<u64>;
```
BusyJay, line 545:
```suggestion
    fn cluster_id(&mut self) -> Result<u64>;
```

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+
+    fn is_cluster_bootstrapped(&mut self) -> Result<bool>;
+
+    fn alloc_id(&mut self) -> Result<u64>;
+
+    fn is_recovering_marked(&mut self) -> Result<bool>;
+
+    fn put_store(&mut self, store: metapb::Store) -> Result<Option<ReplicationStatus>>;
+
+    fn get_store_and_stats(&mut self, store_id: u64)
```
BusyJay, line 561:
```suggestion
    fn store_and_stats(&mut self, store_id: u64)
```

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+}
+
+pub struct CachedDuplexResponse<T> {
+    latest: mpsc::Receiver<ClientDuplexReceiver<T>>,
+    cache: Option<ClientDuplexReceiver<T>>,
+}
+
+impl<T> CachedDuplexResponse<T> {
+    fn new() -> (mpsc::Sender<ClientDuplexReceiver<T>>, Self) {
+        let (tx, rx) = mpsc::unbounded(mpsc::WakePolicy::Immediately);
```
BusyJay, line 634: Actually it should be bounded. But it's OK to leave it as an todo for now.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+        check_resp_header(resp.get_header())?;
+
+        Ok(resp.take_cluster())
+    }
+
+    fn get_region_and_leader(
+        &mut self,
+        key: &[u8],
+    ) -> PdFuture<(metapb::Region, Option<metapb::Peer>)> {
+        let _timer = PD_REQUEST_HISTOGRAM_VEC
```
BusyJay, line 1030: _timer should be moved into async block.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+                    break;
+                }
+                let res = hb_tx.send_all(&mut requests).await;
+                if res.is_ok() {
+                    // requests are drained.
+                    break;
+                } else {
+                    warn!("region heartbeat stream exited"; "res" => ?res);
+                }
+                let _ = hb_tx.close().await;
```
BusyJay, line 709: It should reconnect on rpc error.

components/pd_client/src/client_v2.rs
tabokie: These (and `get_store_and_stats`) involves network call, I think get_xx better informs the overhead.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1418 @@
...
+            .await?
+        {
+            self.publish_cache();
+            return Ok(true);
+        }
+        Ok(false)
+    }
+
+    #[inline]
+    fn check_resp<T>(&mut self, resp: GrpcResult<T>) -> GrpcResult<T> {
```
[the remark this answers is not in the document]
tabokie: It makes it easy to unwrap the error. e.g.
```rust
let resp = check_resp(resp)?;
```
is simpler than
```
check_resp(&resp);
let resp = resp?;
```

components/pd_client/src/client_v2.rs
tabokie: It's not as convenient because each time the test has to create a runtime/thread itself.

BusyJay: > has to create a runtime/thread itself.

Why not just `futures::executor::block_on`?

components/pd_client/src/client_v2.rs
BusyJay: There is no difference. I don't think there are any projects or languages use `get_` prefix to hint there is a overhead.

tabokie: IMO it's implicitly hinted by the naming convention of simple setter/getter. Here in rust simple accessor is named as the variable name, which implicitly means {verb}_{object} is not a simple getter.

Also for this client interface, the priority naming rule is to align with RPC names. It means `get_store_and_stats` should be named as it is to align with `GetStore` rpc.

BusyJay: > which implicitly means {verb}_{object} is not a simple getter.

Or it's probably treated as bad naming by anyone else. If you wan't to hint it's not a simple getter, then you should name it something else instead of bring back bad style.

tabokie: Emmm, I'm not sure what's the basis of your argument. I believe setter/getter refers to accessing private member variable. Language can have preference over how to name these setter/getter. Then it becomes a convention that such names have near zero overhead as they are only aliases for local access.

There shouldn't be bias toward naming outside this scope. "get"/"put" are perfectly valid verbs to name a function in this context.

BusyJay: > convention that such names have near zero overhead as they are only aliases for local access

I don't think getters should be near zero overhead. However, even in this case, `cluster_id` is near zero cost in most cases. In you implementation, it's wrong to wait for ready for accessing cluster id. Cluster ID never changes after it's initialized.

Because getters in Rust is preferred to be named without `get_` prefix. When a method has `get_` prefix, it's considered as bad style. But now you bring a special meaning for naming `get_xxx`, then `get_xx` can be ambiguous. Whether it's a good name depends on the context rather than simply syntax.

tabokie: > In you implementation, it's wrong to wait for ready for accessing cluster id.

This is incorrect. The cluster_id is initialized as zero before the first connection is available. We could of course only wait when the cluster_id is null, but that doesn't change the fact that this function *might* involves network IO.

> I don't think getters should be near zero overhead.

It's not a matter of thinking. Setter/getter by definition is a way to protect access to member variables (https://en.wikipedia.org/wiki/Mutator_method). It should have the same overhead as directly reading the variable.

> When a method has get_ prefix, it's considered as bad style.

I don't think this is the right interpretation. Choosing {variable} over get_{variable} is just a preference, which is mostly likely introduced because of rust's pervasive use of references.

> But now you bring a special meaning for naming get_xxx, then get_xx can be ambiguous.

On the contrary, only in a language with the naming convention to use get_xxx as getter, will we have such ambiguity. Unlike e.g. java, here in Rust we don't assign special meaning to function names like get_xx (getXX for that matter). There's nothing to dis-ambiguate of.

BusyJay: > It should have the same overhead as directly reading the variable.

I don't see how this is concluded. For example, in some lazy initializing cases, first time getting a field will initialize it with some value. Encapsulation is the purpose, not the overhead.

> I don't think this is the right interpretation.

That's how it's interpreted in Rust.

> we don't assign special meaning to function names like get_xx

The special meaning is just what you have mentioned in the first place: "These (and get_store_and_stats) involves network call, I think get_xx better informs the overhead". The ambiguity is there is no way to know whether it's a bad name or a special form to hint overhead until we look into the implementation and context.

As mentioned above, if you try to convey a meaning of network access, you should use a different name like `query_cluster_id` instead of `get_cluster_id`.

tabokie: > That's how it's interpreted in Rust.

What's the source?

> you should use a different name like query_cluster_id instead of get_cluster_id.

How is get different than query? Especially when the RPC is named as `GetStoreRequest`.

BusyJay: > What's the source?

It's a common sense as `get_xxx` is bad style.

> How is get different than query? Especially when the RPC is named as GetStoreRequest.

I think I have explained enough above why we avoid `get_` prefix in method naming. This is not a trait that aims to map the RPC definition apparently.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1377 @@
...
+                    });
+                if resp_tx.send(hb_rx).is_err() {
+                    break;
+                }
+                let res = hb_tx.send_all(&mut requests).await;
+                if res.is_ok() {
+                    // requests are drained.
+                    break;
+                } else {
+                    let res = raw_client.check_resp(res);
```
BusyJay, line 661: If PD cancel the stream, it's possible to return `RemoteStopped`, which will not be treated as `RpcFailure`.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1377 @@
...
+    /// Might panic if `wait_for_ready` isn't called up-front.
+    #[inline]
+    fn call_option(&self) -> CallOption {
+        self.cache.target_info.call_option()
+    }
+
+    /// Might panic if `wait_for_ready` isn't called up-front.
+    #[inline]
+    fn cluster_id(&self) -> u64 {
+        self.cache.members.get_header().get_cluster_id()
```
BusyJay, line 364: Should assert `cluster_id` is not 0.

tests/integrations/pd/test_rpc_client.rs
```diff
@@ -1,26 +1,39 @@
...
 use security::{SecurityConfig, SecurityManager};
 use test_pd::{mocker::*, util::*, Server as MockServer};
-use tikv_util::config::ReadableDuration;
-use tokio::runtime::Builder;
+use tikv_util::{config::ReadableDuration, mpsc::future::WakePolicy};
+use tokio::runtime::{Builder, Runtime};
 use txn_types::TimeStamp;
 
+fn setup_runtime() -> Runtime {
+    Builder::new_multi_thread()
```
BusyJay, line 17: Should assign name otherwise logs can be failed to trace.

tests/integrations/pd/test_rpc_client.rs
```diff
@@ -318,66 +315,7 @@ fn test_validate_endpoints_retry() {
...
-
-    let client = new_client(eps, None);
-
-    for _ in 0..3 {
-        func(&client);
-    }
-}
-
-#[test]
-fn test_retry_async() {
```
BusyJay, line 339: Though v2 doesn't have retry, but it has `wait_for_ready`. So the v2 should test whether `wait_for_ready` is called in all methods. It should be very easy with failpoint.

BusyJay (commented):
Rest LGTM

components/pd_client/src/client_v2.rs
tabokie: > It's a common sense as get_xxx is bad style.

Well, apparently me and the author who wrote the v1 version doesn't agree with you.

> I think I have explained enough above why we avoid get_ prefix in method naming.

All your arguments are based on a very subjective view of get_<none> naming, which is still not agreed upon.

BusyJay: > the author who wrote the v1 version doesn't agree with you

Because back in the day we wrote v1, we don't take style seriously. Naming are quite arbitrary.

> a very subjective view of get_ naming, which is still not agreed upon.

No, it's a not subjective, https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter.

tabokie: Yes, I saw that doc long ago. My understanding is that `getter` used in this docs has a very specific meaning, which is getting something that's already there (of course, since there's no definition for it in this specific doc, I'm only speculating).

> Since path is a getter, it would not be correct to call it get_path or as_path.

This line specifically calls out that `sth is a getter` is the premise. It implies (between the lines): not every method that returns an object can be quantified as a getter.

Anyway, I think this discussion is already wasting too much time. I may just change the verb.

components/pd_client/src/client_v2.rs
```diff
@@ -334,7 +342,10 @@ impl CachedRawClient {
 
     #[inline]
     fn check_resp<T>(&mut self, resp: GrpcResult<T>) -> GrpcResult<T> {
-        if matches!(resp, Err(GrpcError::RpcFailure(_))) {
+        if matches!(
+            resp,
+            Err(GrpcError::RpcFailure(_) | GrpcError::RemoteStopped)
```
BusyJay, line 347: Should also match RpcFinish

components/pd_client/src/client_v2.rs
```diff
@@ -319,8 +320,15 @@ impl CachedRawClient {
     /// Increases global version only when a new connection is established.
     /// Might panic if `wait_for_ready` isn't called up-front.
     async fn reconnect(&mut self) -> Result<bool> {
-        let force = self.channel().check_connectivity_state(true)
+        #[allow(unused_mut)]
+        let mut force = self.channel().check_connectivity_state(true)
```
BusyJay, line 324: Why not
```
let force = (|| {
    fail_point!("pd_client_force_reconnect", |_| true);
    false
})();
```?

**Review by BusyJay — APPROVED** [review]

File: tests/integrations/pd/test_rpc_client.rs [hunk 33]
```diff
@@ -1,26 +1,40 @@
...
 use test_pd::{mocker::*, util::*, Server as MockServer};
-use tikv_util::config::ReadableDuration;
-use tokio::runtime::Builder;
+use tikv_util::{config::ReadableDuration, mpsc::future::WakePolicy};
+use tokio::runtime::{Builder, Runtime};
 use txn_types::TimeStamp;
 
+fn setup_runtime() -> Runtime {
+    Builder::new_multi_thread()
+        .thread_name("poller")
```
BusyJay, line 18: Name should be `thd_name!("poller")`.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1376 @@
...
+                v = should_reconnect.next().fuse() => {
+                    match v {
+                        None => break,
+                        Some(v) if v < client.cache_version => continue,
+                        _ => {}
+                    }
+                }
+            }
+        }
+        if let Err(e) = client.reconnect().await {
```
BusyJay, line 426: There may need to be backoff between retires.

components/pd_client/src/client_v2.rs
```diff
@@ -0,0 +1,1376 @@
...
+            select! {
+                // Checks for leader change periodically.
+                _ = client
+                    .channel()
+                    .wait_for_state_change(state, cfg.update_interval.0)
+                    .fuse() => {}
+                v = should_reconnect.next().fuse() => {
+                    match v {
+                        None => break,
+                        Some(v) if v < client.cache_version => continue,
```
BusyJay, line 420: The channe may lead to OOM if the client always fail to connect PD.

BusyJay:
Note now we rely on grpc retry, so we need to set a smaller retry interval.

https://github.com/tikv/tikv/blob/e52eb4d59c796a1f585c9d26660463d0e95a9d61/src/server/raft_client.rs#L716-L717

PR #13697, opened 2022-10-31
scheduler: Optimize CPU usage of waking up

### LINKED ISSUE #13692: commit a4dc37b causes 2%~4% performance regression in different workloads

master 

Linux

### Steps to reproduce

run the following workloads with the commit c74c8ca907632263a0cfcd84b8413e9edc5c19e5 and commit a4dc37b0c07ee1b5bc7c60a5d8360666500ecdb4:
- bank and sysbench oltp_write_only,oltp_update_index,oltp_delete,oltp_read_write 
 

### What did you expect?
No performance regression.

### What did happened?

bench_type | bench_sub_type | threads | qps of  commit c74c8ca907632263a0cfcd84b8413e9edc5c19e5 | qps of  commit a4dc37b0c07ee1b5bc7c60a5d8360666500ecdb4 | diff(%)
-- | -- | -- | -- | -- | --
bank | bank-50m | 200 | 70244 | 67615 | -3.74%
sysbench | oltp_write_only | 200 | 56593 | 55497 | -1.94%
sysbench | oltp_update_index  | 200 | 22008 | 21574 | -1.97%
sysbench | oltp_delete | 200 | 84739 | 82265 | -2.92%
sysbench | oltp_read_write | 200 | 83881 | 81762 | -2.53%

MyonKeminta:

Issue Number: Close #13692

What's Changed:

```commit-message
Optimize CPU usage of waking up.

1. When popped entries are empty in `on_release_locks`, do not do anything in the high priority pool. This seems to be the reason of the performance regression mentioned in #13692 .
2. Skip waking up if the `LockWaitQueues` is empty.
```

### Related changes

-

### Check List 

Tests 

- Unit test

Side effects

-

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/scheduler.rs
+++ b/src/storage/txn/scheduler.rs
@@ -48,6 +48,7 @@ use parking_lot::{Mutex, MutexGuard, RwLockWriteGuard};
 use pd_client::{Feature, FeatureGate};
 use raftstore::store::TxnExt;
 use resource_metering::{FutureExt, ResourceTagFactory};
+use smallvec::SmallVec;
 use tikv_kv::{Modify, Snapshot, SnapshotExt, WriteData};
 use tikv_util::{
     deadline::Deadline, quota_limiter::QuotaLimiter, time::Instant, timer::GLOBAL_TIMER_HANDLE,
@@ -790,8 +791,17 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
     }
 
     fn on_release_locks(&self, released_locks: ReleasedLocks) {
-        let mut legacy_wake_up_list = vec![];
-        let mut delay_wake_up_futures = vec![];
+        // This function is always called when holding the latch of the involved keys.
+        // So if we found the lock waiting queues are empty, there's no chance
+        // that other threads/commands adds new lock-wait entries to the keys
+        // concurrently. Therefore it's safe to skip waking up when we found the
+        // lock waiting queues are empty.
+        if self.inner.lock_wait_queues.is_empty() {
+            return;
+        }
+
+        let mut legacy_wake_up_list = SmallVec::<[_; 4]>::new();
+        let mut delay_wake_up_futures = SmallVec::<[_; 4]>::new();
         let wake_up_delay_duration_ms = self
             .inner
             .pessimistic_lock_wake_up_delay_duration_ms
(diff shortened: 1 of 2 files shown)


src/storage/txn/scheduler.rs
```diff
@@ -790,8 +791,17 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
 
     fn on_release_locks(&self, released_locks: ReleasedLocks) {
-        let mut legacy_wake_up_list = vec![];
-        let mut delay_wake_up_futures = vec![];
+        // This function is always called when holding the latch of the involved keys.
+        // So if we found the lock waiting queues are empty, there's no chance
+        // that other threads/commands adds new lock-wait entries to the keys
+        // concurrently. Therefore it's safe to skip waking up when we found the
+        // lock waiting queues are empty.
+        if self.inner.lock_wait_queues.is_empty() {
```
sticnarf, line 799: Can we skip waking up according to the length of `queue_map`?

cfzjywxk:
@MyonKeminta 
The benchmark test results could be pasted in the description field or in the issue comment.

MyonKeminta:
## Benchmark results comparing to issue #13692

bench_type | bench_sub_type | threads | qps of  commit c74c8ca907632263a0cfcd84b8413e9edc5c19e5 | qps of  commit a4dc37b0c07ee1b5bc7c60a5d8360666500ecdb4 | qps of this PR
-- | -- | -- | -- | -- | --
bank | bank-50m | 200 | 70244 | 67615 | 70368
sysbench | oltp_write_only | 200 | 56593 | 55497 | 56732
sysbench | oltp_update_index  | 200 | 22008 | 21574 | 22221
sysbench | oltp_delete | 200 | 84739 | 82265 | 82386 (rerunning test)
sysbench | oltp_read_write | 200 | 83881 | 81762 | 83894

src/storage/txn/scheduler.rs
sticnarf: As you said it is most important to avoid the high pool work. Then, can we just remove this empty check, or is it equally important?

MyonKeminta: It's not as important. I did this just because I noticed it seems to be able to be optimized while I'm investigating the regression problem. I think it's better if in most cases we can check only an atomic value instead of lock and check the concurrent hashmap every time. I don't have any idea if it can be removed since I didn't test it... so I suggest keep this optimization.

sticnarf: OK, I got it. My thought is that this optimization may not always be so effective. As long as there are some lock waiting, although they are irrelavant to this key, the optimization does not work.

Wrapping the atomic int into a prometheus observer can avoid the duplication, but it's a bit complex so I'm not advising this way.

sticnarf (approved):
LGTM. But I'm still a little bit skeptical about the necessity of the empty check. Maybe we can verify in the future whether it affects real world performance.

cfzjywxk (approved):

Issue #13694, opened 2022-10-31
Skip accumulated locks by a second get/seek


sticnarf:
## Motivation
Consider a use case: the user application always reads and locks a fixed key but never changes it:

```rust
txn = client.start_transaction()
txn.lock("k1")
v1 = txn.get("k1")
txn.put("k2", "v2")
txn.commit()
```

No matter it's an optimistic or pessimistic transaction, the transaction will write a new `LOCK` record on `k1` in the write CF. If we repeats executing this transaction, there can be a lot of `LOCK` records on `k1` that are newer than the effective `PUT` record, like:

```
k1@100  LOCK
k1@99   LOCK
k1@98   LOCK
...
k1@11   LOCK
k1@10   PUT   value = "v1"
```

It will be a disaster when we want to read the value of `k1`. If it's a point get, we can only seek to the latest version and waste a lot of effort calling next until we find an effective record. The performance of range scan is also degraded. Furthermore, the current MVCC GC cannot remove these `LOCK` records because they are newer than the latest `PUT` records. So, the performance will keep getting worse until the key is effectively modified.

## Brief design

One of the conditions of the problem is that we are seeking the latest version of the key. If we can know what is the version of the latest `PUT` or `DELETE` record, we can directly move the cursor to that key without skipping the huge amount of ineffective versions.

To achieve this, we will store two more fields in the `LOCK` records in the write CF:

```rust
pub struct Write {
    ...
    /// The commit TS of the latest PUT/DELETE record
    last_change_ts: TimeStamp,
    /// The number of versions that need skipping from this key
    /// to find the latest PUT/DELETE record
    versions_to_last_change: u64,
}
```

If we find that the latest version of a key is a LOCK, we may choose to get or seek again according to the `last_change_ts`.
Because get/seek is a much heavier operation than next, `versions_to_last_change` will tell us whether it is worth using get/seek.

Refer to https://pingcap.feishu.cn/docx/IwomdGtc1oIVepx9tIec2Aycn7c for more detailed design. There are also some discussions in the doc.

## Development tasks

- [x] https://github.com/pingcap/kvproto/pull/997
- [x] #13698
- [x] #13728
- [x] #13721
- [x] #13717
- [x] ~~#13749~~ (reverted by #13834)
- [x] #13759
- [x] #13784
- [x] #13819
- [x] #13820

## Further work

[...]

PR #13698, opened 2022-10-31
txn_types: add info about last change to Lock and Write

sticnarf:

Issue Number: Ref #13694

What's Changed:

This PR is the starting preparation of skipping accumulated locks.

It only adds the ability of parsing and serializing `last_change_ts` and `versions_to_last_change`. These two fields are not used at all at this time. Please refer to the issue and design doc for the use of these fields.

```commit-message
This commit adds support for serializing and parsing the infomation
about the last change stored in the lock and write CF.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


sticnarf:
PTAL @MyonKeminta @ekexium

youjiali1995 (approved):

components/txn_types/src/lock.rs
```diff
@@ -85,6 +86,12 @@ pub struct Lock {
     // while committing is relatively expensive. So the solution is putting the ts of the rollback
     // to the lock.
     pub rollback_ts: Vec<TimeStamp>,
+
+    /// The commit TS of the latest PUT/DELETE record
```
ekexium, line 90:
```suggestion
    /// The commit TS of the last PUT/DELETE record before this lock
```
It could be a bit ambiguous if the lock is not the latest record.

components/txn_types/src/lock.rs
```diff
@@ -85,6 +86,12 @@ pub struct Lock {
...
+    pub last_change_ts: TimeStamp,
+    /// The number of versions that need skipping from the latest version to
```
ekexium, line 92:
```suggestion
    /// The number of versions that need skipping from the current version to
```

components/txn_types/src/lock.rs
```diff
@@ -188,6 +210,11 @@ impl Lock {
                 b.encode_u64(ts.into_inner()).unwrap();
```
ekexium, line 210: I'm just curious why line 180
```
b.encode_var_u64(self.ts.into_inner()).unwrap();
```
uses `encode_var_u64`?

ekexium (approved):
Just nits

components/txn_types/src/lock.rs
sticnarf: A real timestamp large enough so var int doesn't save space. Using var int only slows down encoding and decoding.

I guess there wasn't a reason. The original author just didn't think much about it.

components/txn_types/src/lock.rs
sticnarf: I only apply this suggestion to `Write`. These two fields in the lock CF always count from the latest version. Does it make sense?

components/txn_types/src/lock.rs
sticnarf: Due to similar reason, I think this change is not necessary.

components/txn_types/src/lock.rs
ekexium: sounds good

PR #13709, opened 2022-11-01
raftstore-v2: support tablet split

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

```
Subsequent PR of #13689 which implements propose and execute tablet split

This PR implements on apply result as well as create/initialize split peers.
In v1, split create and register PeerFsm directly which causes some race condition problems as the 
PeerFsm can also be created by raft message.

Now in v2, the PeerFsm creation entrance has been unified which means split does not create it directly. 
Instread, it first sends an AcrossPeerMsg to the store where the store first creates an unintialized PeerFsm 
and then sends the msg to the relevant peer. It includes the information to intialize the peer.
```

Todo: 
- [ ] Implement FreezeAndClone to substitute the CreateCheckpoint API.
- [ ] Only when parent receives all the responses can the apply index be updated.
- [ ] Consider more when the split peers is already initialized.
- [ ] Use DeleteFilesInRange and compaction filter to delete data not in the region.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/src/router/message.rs
```diff
@@ -129,6 +130,8 @@ pub enum PeerMsg {
     SnapshotGenerated(Box<RaftSnapshot>),
     /// Start the FSM.
     Start,
+    /// Messages from peer to peer in the same store
+    AcrossPeerMsg(AcrossPeerMsg),
```
BusyJay, line 134: Meaningless name, it should be specific.

components/raftstore-v2/src/router/message.rs
```diff
@@ -187,6 +193,7 @@ impl fmt::Debug for PeerMsg {
 
 pub enum StoreMsg {
     RaftMessage(Box<RaftMessage>),
+    PeerCreation(PeerCreation),
```
BusyJay, line 196:
```suggestion
    CreatePeer(CreatePeer),
```

components/raftstore-v2/src/router/message.rs
```diff
@@ -195,8 +202,14 @@ impl fmt::Debug for StoreMsg {
...
         match *self {
             StoreMsg::RaftMessage(_) => write!(fmt, "Raft Message"),
+            StoreMsg::PeerCreation(_) => write!(fmt, "Peer Creation"),
             StoreMsg::Tick(tick) => write!(fmt, "StoreTick {:?}", tick),
             StoreMsg::Start => write!(fmt, "Start store"),
         }
     }
 }
+
+pub struct PeerCreation {
```
BusyJay, line 212: Should be defined in admin module. And there should be one `Box` instead of two.

components/raftstore-v2/src/router/message.rs
SpadeA-Tang: Done

components/raftstore-v2/src/router/message.rs
SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/fsm/store.rs
```diff
@@ -20,28 +28,54 @@ where
...
+    /// region_id -> `RegionReadProgress`
+    pub region_read_progress: RegionReadProgressRegistry,
 }
 
 impl<E> StoreMeta<E>
 where
     E: KvEngine,
 {
-    pub fn new() -> StoreMeta<E> {
+    pub fn new(vote_capacity: usize) -> StoreMeta<E> {
```
tonyxuqqi, line 47: what's vote_capacity for?

components/raftstore-v2/src/operation/command/admin/mod.rs
```diff
@@ -36,6 +36,13 @@ pub enum AdminCmdResult {
     ConfChange(ConfChangeResult),
 }
 
+pub struct CreatePeer {
```
tonyxuqqi, line 39: CreatePeer --> CreatePeerInfo

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+        validate_batch_split(req.get_admin_request(), self.region())?;
         let mut proposal_ctx = ProposalContext::empty();
         proposal_ctx.insert(ProposalContext::SYNC_LOG);
         proposal_ctx.insert(ProposalContext::SPLIT);
 
         let data = req.write_to_bytes().unwrap();
         self.propose_with_ctx(store_ctx, data, proposal_ctx.to_vec())
     }
+
+    pub fn on_ready_split_region<T>(
```
tonyxuqqi, line 120: it will be be better to add the original code's fail point now than later.  
such as 
```
 fail_point!("on_split", self.ctx.store_id() == 3, |_| {});
 fail_point!("on_split_invalidate_locks");
```

tonyxuqqi, line 120: Please add the comment to highlight the difference with v1.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+        &mut self,
+        store_ctx: &mut StoreContext<EK, ER, T>,
+        derived_index: usize,
+        tablet_index: u64,
+        regions: Vec<Region>,
+    ) {
+        let derived = &regions[derived_index];
+        let derived_id = derived.id;
```
tonyxuqqi, line 128: derived_id is same as region_id below.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+        let derived_epoch = derived.get_region_epoch().clone();
+        let region_id = derived.get_id();
+
+        // Group in-memory pessimistic locks in the original region into new regions.
+        // The locks of new regions will be put into the corresponding new regions
+        // later. And the locks belonging to the old region will stay in the original
+        // map.
+        let region_locks = {
+            let mut pessimistic_locks = self.txn_ext().pessimistic_locks.write();
+            info!(self.logger, "moving {} locks to new regions", pessimistic_locks.len(); "region_id"=> region_id);
```
tonyxuqqi, line 138: no needed region_id.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+            let new_region_id = new_region.get_id();
+
+            if new_region_id == region_id {
+                let not_exist = meta
+                    .region_ranges
+                    .insert(enc_end_key(&new_region), new_region_id)
+                    .is_none();
+                assert!(not_exist, "[region {}] should not exist", new_region_id);
+                continue;
+            }
```
tonyxuqqi, line 196: `if let Some(r) = meta.regions.get(&new_region_id) {
                // Suppose a new node is added by conf change and the snapshot comes slowly.
                // Then, the region splits and the first vote message comes to the new node
                // before the old snapshot, which will create an uninitialized peer on the
                // store. After that, the old snapshot comes, followed with the last split
                // proposal. After it's applied, the uninitialized peer will be met.
                // We can remove this uninitialized peer directly.
                if util::is_region_initialized(rhttps://github.com/tikv/tikv/commit/9d87f98de78a599d13b7d10879cbee6457115127 {
                    panic!(
                        "[region {}] duplicated region {:?} for split region {:?}",
                        new_region_id, r, new_region
                    );
                }
                self.ctx.router.close(new_region_idhttps://github.com/tikv/tikv/commit/9d87f98de78a599d13b7d10879cbee6457115127;
            }`
Is this code in v1 needed in v2?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+
+            let mut raft_message = self.prepare_raft_message();
+            raft_message.mut_from_peer().set_id(raft::INVALID_ID);
```
tonyxuqqi, line 199: Question: why need to set INVALID_ID

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+                .send_control(StoreMsg::CreatePeer(Box::new(CreatePeer {
+                    raft_message,
+                    split_region_info: init_info,
+                })));
+        }
+        drop(meta);
+
+        if is_leader {
+            self.on_split_region_check_tick();
+        }
```
tonyxuqqi, line 234: missing the failpoint " fail_point!("after_split", self.ctx.store_id() == 3, |_| {});"

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+        let replace = region.get_region_epoch().get_version()
+            > self
+                .storage()
+                .region_state()
+                .get_region()
+                .get_region_epoch()
+                .get_version();
+
+        if !self.storage().is_initialized() || replace {
+            let mut wb = store_ctx.engine.log_batch(5);
```
tonyxuqqi, line 266: use the const instead

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+            // If this region has only one peer and I am the one, campaign directly.
+            if region.get_peers().len() == 1 {
+                raft_group.campaign().unwrap();
+                self.set_has_ready();
+            }
+            self.set_raft_group(raft_group);
+
+            need_schedule_apply_fsm = true;
+        } else {
+            // todo: when reaching here, it is much complexer.
```
tonyxuqqi, line 327: what's the logic to do here?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+            unimplemented!();
+        }
+
+        let mut meta = store_ctx.store_meta.lock().unwrap();
+
+        info!(
+            self.logger,
+            "init split region";
+            "region_id" => region_id,
```
tonyxuqqi, line 336: region_id is not needed.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+                .storage()
+                .region_state()
+                .get_region()
+                .get_region_epoch()
+        );
+
+        let mut split_progress = self.split_progress_mut();
+        *split_progress.get_mut(&child_region_id).unwrap() = true;
+
+        if split_progress.values().all(|v| *v) {
```
tonyxuqqi, line 438: what if there's one new splitted region failed for some reason,  how to recover from that partial status?

components/raftstore-v2/src/fsm/store.rs
SpadeA-Tang: The field thta needs it has been removed. I will remove this.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,20 +74,387 @@ pub struct SplitResult {
...
+        if is_leader {
+            self.set_approximate_size(estimated_size);
+            self.set_approximate_keys(estimated_keys);
+            // todo: this may should be doing in handle_peer_split_response.
+            self.heartbeat_pd(store_ctx);
+
+            info!(
+                self.logger,
+                "notify pd with split";
+                "region_id" => self.region_id(),
```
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: This message is used to make store create an uninitialized peer. We can use it to distinguish from the msg sent from the leader in other stores.
image
image

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: It's the capacity of the log batch, rather than the log index.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: If the peer is initialized and moreover, new than the information sent from the parent.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/router/message.rs
```diff
@@ -187,6 +193,7 @@ impl fmt::Debug for PeerMsg {
 
 pub enum StoreMsg {
     RaftMessage(Box<RaftMessage>),
+    CreatePeer(Box<CreatePeer>),
```
BusyJay, line 196: `SplitInit` can be sent directly without using `CreatePeer`. The message can be constructed when needed.

components/raftstore-v2/tests/integrations/cluster.rs
```diff
@@ -145,6 +145,16 @@ impl TestRouter {
         req.mut_header().set_term(meta.raft_status.hard_state.term);
         req
     }
+
+    pub fn region_detail(&self, region_id: u64) -> metapb::Region {
```
BusyJay, line 149: Why not use `must_query_debug_info`?

SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/router/message.rs
SpadeA-Tang: Okay.

components/raftstore-v2/src/operation/query/local.rs
```diff
@@ -176,6 +176,10 @@ where
             }
         }
 
+        debug!(self.logger,
```
BusyJay, line 179: Useless log.

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -139,7 +139,7 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
     }
 
     /// Partially filled a raft message that will be sent to other peer.
-    fn prepare_raft_message(&mut self) -> RaftMessage {
+    pub fn prepare_raft_message(&mut self) -> RaftMessage {
```
BusyJay, line 142: It's not for split purpose.

components/raftstore-v2/src/operation/life.rs
```diff
@@ -89,6 +90,43 @@ impl DestroyProgress {
...
+        let mut raft_msg = RaftMessage::default();
+        raft_msg.set_region_id(region_id);
+        raft_msg.set_region_epoch(msg.as_ref().region.get_region_epoch().clone());
+        raft_msg.mut_from_peer().set_id(raft::INVALID_ID);
+        raft_msg.set_to_peer(
+            msg.as_ref()
+                .region
+                .get_peers()
+                .iter()
+                .find(|p| p.get_store_id() == ctx.store_id)
```
BusyJay, line 116: store_id can be queried by `self.store.id`.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +286,245 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
 
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    pub fn on_ready_split_region<T>(
+        &mut self,
+        store_ctx: &mut StoreContext<EK, ER, T>,
+        derived_index: usize,
+        tablet_index: u64,
+        regions: Vec<Region>,
+    ) {
+        fail_point!("on_split", store_ctx.store_id == 3, |_| {});
```
BusyJay, line 297: It can be queried by self.peer().get_store_id().

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,13 +76,38 @@ pub struct SplitResult {
     pub tablet_index: u64,
 }
 
+// In split, parent peer wraps data in `CreatePeer` to create and initalize
+// split peers.
+pub struct CreatePeer {
```
BusyJay, line 81:
```suggestion
pub struct SplitInit {
```

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,13 +76,38 @@ pub struct SplitResult {
...
+    /// Split region
+    pub region: metapb::Region,
+    pub parent_is_leader: bool,
+    /// In-memory pessimistic locks that should be inherited from parent region
+    pub locks: PeerPessimisticLocks,
+    pub last_split_region: bool,
+}
+
+#[derive(Debug)]
+pub struct SplitRegionInitResp {
```
BusyJay, line 93: Useless struct.

components/raftstore-v2/src/router/message.rs
BusyJay, line 196:
```suggestion
    SplitInit(Box<SplitInit>),
```

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,13 +76,38 @@ pub struct SplitResult {
...
+    pub parent_epoch: RegionEpoch,
+    pub child_region_id: u64,
+    // FIXME: when it is false
+    pub result: bool,
+}
+
+pub enum RegionSplitMsg {
```
BusyJay, line 100: Useless encapsulation.

components/raftstore-v2/src/operation/query/local.rs
SpadeA-Tang: removed

components/raftstore-v2/src/operation/life.rs
SpadeA-Tang: okay

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: removed

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: removed

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +270,220 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+            // If this region has only one peer and I am the one, campaign directly.
+            if region.get_peers().len() == 1 {
+                raft_group.campaign().unwrap();
+                self.set_has_ready();
+            }
+            self.set_raft_group(raft_group);
+
+            need_schedule_apply_fsm = true;
+        } else {
+            // todo: when reaching here (peer is initalized before and cannot be replaced),
```
BusyJay, line 422: Just return is OK.

components/raftstore-v2/src/operation/command/admin/split.rs
tonyxuqqi: > It's the capacity of the log batch, rather than the log index.

Then it might be better to add a small comment why it's 5.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,13 +72,26 @@ pub struct SplitResult {
     pub tablet_index: u64,
 }
 
+// In split, parent peer wraps data in `CreatePeer` to create and initalize
+// split peers.
+pub struct SplitInit {
+    pub parent_region_id: u64,
+    pub parent_epoch: RegionEpoch,
```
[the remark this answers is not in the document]
SpadeA-Tang: It's not needed now.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -64,13 +72,26 @@ pub struct SplitResult {
...
+    /// Split region
+    pub region: metapb::Region,
+    pub parent_is_leader: bool,
```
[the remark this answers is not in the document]
SpadeA-Tang: Done

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: As this PR does not consider persistency now, these codes have been removed.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +268,202 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+        self.post_split();
+
+        let last_region_id = regions.last().unwrap().get_id();
+        for (new_region, locks) in regions.into_iter().zip(region_locks) {
+            let new_region_id = new_region.get_id();
+            if new_region_id == region_id {
+                continue;
+            }
+
+            let peer_creation_msg = PeerMsg::SplitInit(Box::new(SplitInit {
```
BusyJay, line 317: bad name.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +268,202 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+        }
+        drop(meta);
+    }
+
+    pub fn on_split_init<T>(
+        &mut self,
+        store_ctx: &mut StoreContext<EK, ER, T>,
+        init_info: Box<SplitInit>,
+    ) {
+        let SplitInit {
```
BusyJay, line 346: Why not access the fields directly?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +268,202 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+        if !self.storage().is_initialized() || replace {
+            let split_temp_path = store_ctx.tablet_factory.tablet_path_with_prefix(
+                SPLIT_PREFIX,
+                region_id,
+                RAFT_INIT_LOG_INDEX,
+            );
+            let tablet_path = store_ctx
+                .tablet_factory
+                .tablet_path(region_id, RAFT_INIT_LOG_INDEX);
+            std::fs::rename(&split_temp_path, &tablet_path).unwrap_or_else(|e| {
```
BusyJay, line 374: Should use `load_tablet`.

components/raftstore-v2/src/operation/life.rs
```diff
@@ -89,6 +90,41 @@ impl DestroyProgress {
 }
 
 impl Store {
+    /// The method is called during split.
+    /// The creation process is:
+    /// 1. create an uninitialized peer if not existed before
+    /// 2. initialize the peer by the information sent from parent peer
+    #[inline]
+    pub fn on_peer_creation<EK, ER, T>(
```
BusyJay, line 98:
```suggestion
    pub fn on_split_init<EK, ER, T>(
```

components/raftstore-v2/src/operation/life.rs
```diff
@@ -89,6 +90,41 @@ impl DestroyProgress {
...
+            msg.as_ref()
+                .region
+                .get_peers()
+                .iter()
+                .find(|p| p.get_store_id() == self.store_id())
+                .unwrap()
+                .clone(),
+        );
+
+        // It will create the peer if it is not existed
```
BusyJay, line 121:
```suggestion
        // It will create the peer if it does not exist
```

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: The ownership of the lock is needed.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Modified

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: done

components/raftstore-v2/src/operation/command/admin/split.rs
BusyJay: What's the compilation error?

SpadeA-Tang: Okay, I see. I removed it.

tonyxuqqi (approved):

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -167,6 +187,57 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+            //
+            // The difference of `expire` and `expire_remote_lease` is that `expire` expires
+            // lease both in local reader and raftstore while
+            // `expire_remote_lease` only expires the lease in local reader.
+            //
+            // V1 calls `expire_remote_lease` while here v2 calls `expire`. This
+            // difference is due to v2 only performs read by local reader and
+            // raftstore is just a place to renew the lease. If the lease in raftstore is
+            // not expired, it may not renew lease.
+            self.leader_lease.expire();
```
BusyJay, line 214: No, v2 should be the same as v1. `expire` is just used to forbid on-going read. All local reader should use the new lease as the epoch change.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -63,14 +72,23 @@ pub struct SplitResult {
     pub derived_index: usize,
     pub tablet_index: u64,
 }
+pub struct SplitInit {
+    /// Split region
+    pub region: metapb::Region,
```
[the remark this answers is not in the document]
SpadeA-Tang: It's accessed in life.rs

components/raftstore-v2/src/operation/life.rs
```diff
@@ -89,6 +90,40 @@ impl DestroyProgress {
...
+    pub fn on_split_init<EK, ER, T>(
+        &mut self,
+        ctx: &mut StoreContext<EK, ER, T>,
+        msg: Box<SplitInit>,
+    ) where
+        EK: KvEngine,
+        ER: RaftEngine,
+    {
+        let region_id = msg.as_ref().region.id;
+        let mut raft_msg = RaftMessage::default();
```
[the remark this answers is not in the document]
BusyJay: Avoid copy.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -14,6 +14,7 @@ use crossbeam::channel::{Sender, TrySendError};
 use engine_traits::{Engines, KvEngine, RaftEngine, TabletFactory};
 use file_system::{set_io_type, IoType};
 use futures::{compat::Future01CompatExt, FutureExt};
+use keys::enc_end_key;
```
tabokie, line 17: I have noticed rustc doesn't complain about unused imports in raftstore-v2. Does anyone know why? cc @BusyJay @sticnarf

SpadeA-Tang: Because #![allow(unused)] is added in the lib.rs.

tabokie: Oh, apparently this:
https://github.com/tikv/tikv/blob/fe997db4db8a5a096f8a45c0db3eb3c2e5879262/components/raftstore-v2/src/lib.rs#L24

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +267,175 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+            // existed in which case we should redirect it to the store.
+            match store_ctx.router.force_send(new_region_id, split_init) {
+                Ok(_) => {}
+                Err(SendError(PeerMsg::SplitInit(msg))) => {
+                    store_ctx.router.send_control(StoreMsg::SplitInit(msg));
+                }
+                _ => unreachable!(),
+            }
+        }
+        drop(meta);
```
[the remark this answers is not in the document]
tabokie: Add comment to explain why hold lock until now.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +267,175 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+        // Roughly estimate the size and keys for new regions.
+        let new_region_count = regions.len() as u64;
+        let mut meta = store_ctx.store_meta.lock().unwrap();
+        meta.set_region(
+            derived.clone(),
+            self,
+            RegionChangeReason::Split,
+            tablet_index,
+        );
+        self.post_split();
```
tabokie, line 307: Must it be inside lock? If so add comment in `post_split` to require the lock.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +267,175 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+
+        let last_region_id = regions.last().unwrap().get_id();
+        for (new_region, locks) in regions.into_iter().zip(region_locks) {
+            let new_region_id = new_region.get_id();
+            if new_region_id == region_id {
```
tabokie, line 312: How can this happen?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -9,51 +9,60 @@
...
+//! - on_ready_split_region: Update the relevant in memory meta info of the
+//!   parent peer, then send to the store the relevant info needed to create and
+//!   initialize the split regions.
 //!
 //! Split peer creation and initlization:
-//! - todo
-//!
-//! Split finish:
-//! - todo
+//! - init_split_region: In normal cases, the uninitialized split region will be
```
tabokie, line 24:
```suggestion
//! - on_split_init: In normal cases, the uninitialized split region will be
```

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +267,175 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+                self.set_has_ready();
+            }
+            self.set_raft_group(raft_group);
+        } else {
+            // todo: when reaching here (peer is initalized before and cannot be replaced),
+            // it is much complexer.
+            return;
+        }
+
+        let mut meta = store_ctx.store_meta.lock().unwrap();
```
tabokie, line 398:
```rust
{
  let meta = ;
}
```

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -610,7 +797,9 @@ mod test {
         // Split will create checkpoint tablet, so if there are some writes before
         // split, they should be flushed immediately.
         apply.apply_put(CF_DEFAULT, b"k04", b"v4").unwrap();
-        assert!(!apply.write_batch_mut().as_ref().unwrap().is_empty());
+        assert!(!WriteBatch::is_empty(
```
tabokie, line 800: Why?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -249,6 +267,175 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
...
+                Err(SendError(PeerMsg::SplitInit(msg))) => {
+                    store_ctx.router.send_control(StoreMsg::SplitInit(msg));
+                }
+                _ => unreachable!(),
+            }
+        }
+        drop(meta);
+    }
+
+    pub fn on_split_init<T>(
```
tabokie, line 336: I didn't see where this is called.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -55,11 +69,31 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
...
     /// region buckets.
     region_buckets: Option<BucketStat>,
+    last_region_buckets: Option<BucketStat>,
+
     /// Transaction extensions related to this peer.
     txn_ext: Arc<TxnExt>,
     txn_extra_op: Arc<AtomicCell<TxnExtraOp>>,
 }
 
+pub fn raft_config(peer_id: u64, applied_index: u64, cfg: &Config) -> raft::Config {
```
tabokie, line 79: Could this be implemented for `Config`?

components/raftstore-v2/src/fsm/store.rs
```diff
@@ -35,8 +42,21 @@ where
             store_id: None,
             readers: HashMap::default(),
             tablet_caches: HashMap::default(),
+            region_read_progress: RegionReadProgressRegistry::new(),
         }
     }
+
+    #[inline]
+    pub fn set_region<ER: RaftEngine>(
```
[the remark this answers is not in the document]
SpadeA-Tang: It followers the pattern the v1. But after removeing some fields not used in v2, it seems to be unnecessary in this way.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: No, I will change this.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: regions include dervied as well.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Two traits imported include is_empty, so use this way to let compiler happy.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: image

components/raftstore-v2/src/batch/store.rs
```diff
@@ -277,8 +279,12 @@ impl<EK: KvEngine, ER: RaftEngine, T> StorePollerBuilder<EK, ER, T> {
                     Some(p) => p,
                     None => return Ok(()),
                 };
-                let pair = PeerFsm::new(&cfg, &*self.tablet_factory, storage)?;
-                let prev = regions.insert(region_id, pair);
+                let (sender, peer_fsm) = PeerFsm::new(&cfg, &*self.tablet_factory, storage)?;
+                let region = peer_fsm.as_ref().peer().region().clone();
```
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

components/raftstore-v2/src/fsm/store.rs
```diff
@@ -1,16 +1,21 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::time::SystemTime;
+use std::{collections::BTreeMap, time::SystemTime};
```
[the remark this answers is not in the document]
SpadeA-Tang: Good catch.

components/raftstore-v2/src/operation/command/admin/split.rs
SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/operation/command/admin/split.rs
tabokie: Ah okay, I hide this file.

components/raftstore-v2/src/operation/command/admin/split.rs
tabokie: What do you mean? `region_locks` doesn't include locks for `region_id`, are you implying `regions.len() == region_locks.len() + 1`?

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: Done

components/raftstore-v2/src/router/message.rs
```diff
@@ -195,6 +202,7 @@ impl fmt::Debug for StoreMsg {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         match *self {
             StoreMsg::RaftMessage(_) => write!(fmt, "Raft Message"),
+            StoreMsg::SplitInit(_) => write!(fmt, "Peer Creation"),
```
BusyJay, line 205: Wrong message.

components/raftstore-v2/src/router/message.rs
```diff
@@ -167,6 +170,9 @@ impl fmt::Debug for PeerMsg {
             },
             PeerMsg::ApplyRes(res) => write!(fmt, "ApplyRes {:?}", res),
             PeerMsg::Start => write!(fmt, "Startup"),
+            PeerMsg::SplitInit(_) => {
+                write!(fmt, "Across peer msg")
```
BusyJay, line 174: Wrong message.

components/raftstore-v2/src/operation/life.rs
```diff
@@ -89,6 +90,50 @@ impl DestroyProgress {
...
+        EK: KvEngine,
+        ER: RaftEngine,
+    {
+        let region_id = msg.region.id;
+        let mut raft_msg = Box::new(RaftMessage::default());
+        raft_msg.set_region_id(region_id);
+        raft_msg.set_region_epoch(msg.region.get_region_epoch().clone());
+        // Set the peer_id be INVALID_ID so that the message will not be sent to the
+        // peer after the creation.
+        raft_msg.mut_from_peer().set_id(raft::INVALID_ID);
```
BusyJay, line 112: Why not just `raft_msg.take_from_peer();`?

components/raftstore-v2/src/operation/life.rs
```diff
@@ -89,6 +90,50 @@ impl DestroyProgress {
...
+                .clone(),
+        );
+
+        // It will create the peer if it does not exist
+        self.on_raft_message(ctx, raft_msg);
+
+        ctx.router
+            .force_send(region_id, PeerMsg::SplitInit(msg))
+            .unwrap_or_else(|e| {
+                panic!(
```
BusyJay, line 128: It's OK to return error. It's possible the peer is destroyed right before force send.

components/raftstore-v2/tests/integrations/test_split.rs
```diff
@@ -0,0 +1,209 @@
...
+    thread::sleep(Duration::from_secs(1));
+}
+
+fn put(
+    store_id: u64,
+    router: &mut TestRouter,
+    region: &metapb::Region,
+    key: &[u8],
+) -> RaftCmdResponse {
+    let mut req = RaftCmdRequest::default();
```
BusyJay, line 53: Why not `router.new_request_for`?

components/raftstore-v2/tests/integrations/test_split.rs
```diff
@@ -0,0 +1,209 @@
...
+    assert!(block_on(sub.wait_committed()));
+    block_on(sub.result()).unwrap()
+}
+
+fn must_put_succeed(store_id: u64, router: &mut TestRouter, region: &metapb::Region, key: &[u8]) {
+    let resp = put(store_id, router, region, key);
+    assert!(!resp.get_header().has_error(), "{:?}", resp);
+}
+
+fn must_put_error(store_id: u64, router: &mut TestRouter, region: &metapb::Region, key: &[u8]) {
```
BusyJay, line 84: Better implement these inline, it's hard to read to jump back and forth.

BusyJay (commented):
Overall workflow LGTM

components/raftstore-v2/src/router/message.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/src/router/message.rs
SpadeA-Tang: Good catch.

components/raftstore-v2/tests/integrations/test_split.rs
SpadeA-Tang: Done

components/raftstore-v2/tests/integrations/test_split.rs
SpadeA-Tang: Done

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -173,6 +175,63 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
         self.region().get_id()
     }
 
+    /// Set the region of a peer.
+    ///
+    /// This will update the region of the peer, caller must ensure the region
+    /// has been preserved in a durable device.
+    pub fn set_region(
```
tabokie, line 182: Is this supposed to be put here? Seems only simple getter/setters are placed in this file.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -394,6 +458,38 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+            return false;
+        }
+
+        // If last peer is the leader of the region before split, it's intuitional for
+        // it to become the leader of new split region.
+        let _ = self.raft_group.campaign();
+        true
+    }
+
+    pub fn txn_ext(&self) -> &Arc<TxnExt> {
```
tabokie, line 485: inline

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: So, which file you think is better?

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: Done

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -394,6 +458,39 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
 
+    #[inline]
+    pub fn post_split(&mut self) {
+        self.reset_region_buckets();
+    }
+
+    pub fn reset_region_buckets(&mut self) {
+        if self.region_buckets.is_some() {
+            self.last_region_buckets = self.region_buckets.take();
+            self.region_buckets = None;
```
tabokie, line 469: unnecessary.

components/raftstore-v2/src/raft/peer.rs
tabokie: I don't know really, just a thought.

tabokie (approved):

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: Good catch.

PR #13717, opened 2022-11-02
storage: calculate last_change_ts in acquire_pessimistic_lock

### LINKED ISSUE #13694: Skip accumulated locks by a second get/seek

sticnarf:

Issue Number: Ref #13694

What's Changed:

```commit-message
Information about the last change is calculated and stored in the
pessimistic lock.
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/txn_types/src/write.rs
+++ b/components/txn_types/src/write.rs
@@ -259,6 +259,25 @@ impl Write {
             versions_to_last_change: self.versions_to_last_change,
         }
     }
+
+    /// Returns the new `last_change_ts` and `versions_to_last_change` according
+    /// to this write record.
+    pub fn next_last_change_info(&self, commit_ts: TimeStamp) -> (TimeStamp, u64) {
+        match self.write_type {
+            WriteType::Put | WriteType::Delete => (commit_ts, 1),
+            WriteType::Lock | WriteType::Rollback => {
+                // If `last_change_ts` is zero, do not set `last_change_ts` to indicate we don't
+                // know where is the last change.
+                // This should not happen if data is written in new version TiKV. If we hope to
+                // support data from old TiKV, consider iterating to the last change to find it.
+                if !self.last_change_ts.is_zero() {
+                    (self.last_change_ts, self.versions_to_last_change + 1)
+                } else {
+                    (TimeStamp::zero(), 0)
+                }
+            }
+        }
+    }
 }
 
 #[derive(PartialEq, Clone)]
--- a/components/txn_types/src/write.rs
+++ b/components/txn_types/src/write.rs
@@ -28,6 +28,7 @@ const FLAG_ROLLBACK: u8 = b'R';
 const FLAG_OVERLAPPED_ROLLBACK: u8 = b'R';
 
 const GC_FENCE_PREFIX: u8 = b'F';
+const LAST_CHANGE_PREFIX: u8 = b'l';
 
 /// The short value for rollback records which are protected from being
 /// collapsed.
@@ -150,6 +151,12 @@ pub struct Write {
     /// * `Some(ts)`: A commit record that has been rewritten due to overlapping
     ///   rollback, and it's next version's `commit_ts` is `ts`
     pub gc_fence: Option<TimeStamp>,
+
+    /// The commit TS of the latest PUT/DELETE record
+    pub last_change_ts: TimeStamp,
+    /// The number of versions that need skipping from this record
+    /// to find the latest PUT/DELETE record
+    pub versions_to_last_change: u64,
 }
 
 impl std::fmt::Debug for Write {
@@ -169,6 +176,8 @@ impl std::fmt::Debug for Write {
             )
             .field("has_overlapped_rollback", &self.has_overlapped_rollback)
             .field("gc_fence", &self.gc_fence)
+            .field("last_change_ts", &self.last_change_ts)
+            .field("versions_to_last_change", &self.versions_to_last_change)
             .finish()
     }
 }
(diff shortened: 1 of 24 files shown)


ekexium (approved):
This pr should not close the linked issue

sticnarf:
> This pr should not close the linked issue

Thanks for reminding. Changed to ref.

youjiali1995 (approved):

components/txn_types/src/write.rs
```diff
@@ -259,6 +259,25 @@ impl Write {
...
+            WriteType::Put | WriteType::Delete => (commit_ts, 1),
+            WriteType::Lock | WriteType::Rollback => {
+                // If `last_change_ts` is zero, do not set `last_change_ts` to indicate we don't
+                // know where is the last change.
+                // This should not happen if data is written in new version TiKV. If we hope to
+                // support data from old TiKV, consider iterating to the last change to find it.
+                if !self.last_change_ts.is_zero() {
+                    (self.last_change_ts, self.versions_to_last_change + 1)
+                } else {
+                    (TimeStamp::zero(), 0)
```
MyonKeminta, line 276: Is it possible that we just return `(commit_ts, 1)` in this case, and if we seek to the `last_change_ts` and find it's Lock or Rollback, do `next`s to find the actual version? I think in this way maybe it's possible that for those keys that are frequently locked but rarely updated, they can benefit from the optimization after upgrading from an old version.

By the way, GC may change the real number of versions to last change. Does that matter?

sticnarf: > Is it possible that we just return `(commit_ts, 1)` in this case, and if we seek to the `last_change_ts` and find it's Lock or Rollback when reading, do `next`s to find the actual data? I think in this way maybe it's possible that for those keys that are frequently locked but rarely updated, they can benefit from the optimization after upgrading from an old version.

Emm, actually I don't intend to solve the problem for upgraded data at the first step.

If we indeed want to solve the problem for upgraded version, my idea is to iterate until the last change in `acquire_pessimistic_lock` or `prewrite` to get the correct value (which is what the comment in L272 means). Then, only the first time a key is written will be slow. I think this way is a bit better.

sticnarf: > By the way, GC may change the real number of versions to last change (by deleting old `Lock`s and `Rollback`s). Does that matter?

Right... This also brings inaccuracy. I don't have a good idea now but I don't think it's a very serious problem. The performance just doesn't come back to optimism, and it is likely for such a key to pile up enough versions again.

Anyway, it's always pleasant to solve this problem if possible. Do you have any good idea...?

MyonKeminta: > > By the way, GC may change the real number of versions to last change (by deleting old `Lock`s and `Rollback`s). Does that matter?
> 
> Right... This also brings inaccuracy. I don't have a good idea now but I don't think it's a very serious problem. The performance just doesn't come back to optimism, and it is likely for such a key to pile up enough versions again.
> 
> Anyway, it's always pleasant to solve this problem if possible. Do you have any good idea...?

I have no other idea either. I think it's ok to ignore the inaccuracy too. (I'm just confirming that you already noticed this case

PR #13721, opened 2022-11-02
storage: calculate last_change_ts in prewrite

sticnarf:

Issue Number: Ref #13694

What's Changed:

```commit-message
This commit implements the different cases when last_change_ts
is calculated in prewrite:
1. Inherit from the pessimistic lock
2. Calculate it when checking the new version
3. Amend the pessimistic lock
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/actions/prewrite.rs
+++ b/src/storage/txn/actions/prewrite.rs
@@ -21,7 +21,10 @@ use crate::storage::{
         },
         Error, ErrorInner, Lock, LockType, MvccTxn, Result, SnapshotReader,
     },
-    txn::{actions::check_data_constraint::check_data_constraint, LockInfo},
+    txn::{
+        actions::check_data_constraint::check_data_constraint, sched_pool::tls_can_enable,
+        scheduler::LAST_CHANGE_TS, LockInfo,
+    },
     Snapshot,
 };
 
@@ -62,7 +65,7 @@ pub fn prewrite<S: Snapshot>(
     let lock_status = match reader.load_lock(&mutation.key)? {
         Some(lock) => mutation.check_lock(lock, pessimistic_action)?,
         None if matches!(pessimistic_action, DoPessimisticCheck) => {
-            amend_pessimistic_lock(&mutation, reader)?;
+            amend_pessimistic_lock(&mut mutation, reader)?;
             lock_amended = true;
             LockStatus::None
         }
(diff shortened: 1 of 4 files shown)


src/storage/txn/actions/prewrite.rs
```diff
@@ -727,6 +746,8 @@ fn amend_pessimistic_lock<S: Snapshot>(
             }
             .into());
         }
+        (mutation.last_change_ts, mutation.versions_to_last_change) =
+            write.next_last_change_info(*commit_ts);
```
MyonKeminta, line 750: Do we need to check the `write`'s type here?

sticnarf:
...I give up comparing `last_change_ts` in `test_mess`. It's tedious and not so meaningful...

MyonKeminta (approved):

sticnarf:
PTAL @ekexium

src/storage/txn/actions/prewrite.rs
```diff
@@ -440,6 +454,11 @@ impl<'a> PrewriteMutation<'a> {
             self.txn_props.txn_size,
             self.min_commit_ts,
         );
+        // Only Lock needs to record `last_change_ts` in its write record, write records
+        // of other types themselves are effective changes.
```
ekexium, line 458: Nit: ROLLBACK doesn't, though it doesn't matter at all

ekexium (approved):

src/storage/txn/actions/prewrite.rs
sticnarf: Ah, I mean the possible LockType here. We don't have a `ROLLBACK` lock.

TonsnakeLin:
I find it updates the `(last_change_ts, versions_to_last_change)` in the function `check_for_newer_version` which is  not called in so scenario such as that the kind of transaction is  TransactionKind::Optimistic(true). It will make some locks have   `(last_change_ts, versions_to_last_change)` and some locks have not on the same key. Did we notice this?

sticnarf:
> I find it updates the `(last_change_ts, versions_to_last_change)` in the function `check_for_newer_version` which is not called in so scenario such as that the kind of transaction is TransactionKind::Optimistic(true). It will make some locks have `(last_change_ts, versions_to_last_change)` and some locks have not on the same key. Did we notice this?

`skip_constraint_check` is legacy. The client never sets this configuration (and will not set it in the foreseeable future).

Issue #13665, opened 2022-10-24
Reduce memory allocation in check leader to save CPU


overvenus:
## Development Task

Maps key-value pairs can be reused in leadership resolver, it can save CPU usage when there are many regions.

Issue #13656, opened 2022-10-24
Reuse map and set in check leader to save CPU


overvenus:
## Development Task

Reuse map and set in check leader to save CPU, and helps to reduce TiCDC  latency.

PR #13729, opened 2022-11-03
config: set default cdc.min-ts-interval to 500ms

overvenus:

Issue Number: ref #13656 ref #13665

What's Changed:

```commit-message
config: set default cdc.min-ts-interval to 500ms 
```

After above two PRs, tso thread CPU usage is reduced about 50%, it's ok to reduce `min-ts-interval` and reduce TiCDC latency.

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
config: set default cdc.min-ts-interval to 500ms  to reduce TiCDC latency
```


overvenus:
Closed in favor of #12839

PR #13640, opened 2022-10-20
raftstore-v2: support send/recv snapshot 

bufferflies:
### What is changed and how it works?
Issue Number: Ref https://github.com/tikv/tikv/issues/12842

What's Changed:
```
raftstore-v2: support send/recv snapshot
This PR implementation send/recv snapshot relative changes.
- add snap worker for multi-rocksdb 
- add tests and let the test framework supports multi-nodes.
Next part will implement apply snapshot
```
wait pr : #13439
### Related changes

Base on https://github.com/tikv/tikv/pull/13388

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
- Breaking backward compatibility

### Release note 

```release-note
None
```

PR #13736, opened 2022-11-03
raftstore-v2: make generate snapshot works with checkpoint

nolouch:

What's Changed:

```commit-message
raftstore-v2: make generate snapshot works with checkpoint
```
Part of https://github.com/tikv/tikv/pull/13439
### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_panic/src/misc.rs
+++ b/components/engine_panic/src/misc.rs
@@ -1,7 +1,5 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::path::PathBuf;
-
 use engine_traits::{DeleteStrategy, MiscExt, Range, Result};
 
 use crate::engine::PanicEngine;
@@ -74,10 +72,6 @@ impl MiscExt for PanicEngine {
     }
 
     fn is_stalled_or_stopped(&self) -> bool {
-        panic!()
-    }
-
-    fn checkpoint_to(&self, path: &[PathBuf], size_to_flush: u64) -> Result<()> {
         panic!()
     }
 }
--- a/components/engine_panic/src/misc.rs
+++ b/components/engine_panic/src/misc.rs
@@ -1,5 +1,7 @@
 // Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
 
+use std::path::PathBuf;
+
 use engine_traits::{DeleteStrategy, MiscExt, Range, Result};
 
 use crate::engine::PanicEngine;
@@ -72,6 +74,10 @@ impl MiscExt for PanicEngine {
     }
 
     fn is_stalled_or_stopped(&self) -> bool {
+        panic!()
+    }
+
+    fn checkpoint_to(&self, path: &[PathBuf], size_to_flush: u64) -> Result<()> {
         panic!()
     }
 }
--- a/components/raftstore-v2/src/batch/store.rs
+++ b/components/raftstore-v2/src/batch/store.rs
@@ -22,7 +22,7 @@ use kvproto::{
 use raft::INVALID_ID;
 use raftstore::store::{
     fsm::store::PeerTickBatch, local_metrics::RaftMetrics, Config, ReadRunner, ReadTask,
-    SnapManagerBuilder, StoreWriters, Transport, WriteSenders,
+    StoreWriters, TabletSnapManager, Transport, WriteSenders,
 };
 use slog::Logger;
 use tikv_util::{
@@ -375,7 +375,7 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
 
-        let snap_mgr = SnapManagerBuilder::default().build(
+        let snap_mgr = TabletSnapManager::new(
             tablet_factory
                 .tablets_path()
                 .join(Path::new("snap"))
--- a/components/raftstore-v2/src/batch/store.rs
+++ b/components/raftstore-v2/src/batch/store.rs
@@ -366,6 +366,7 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
         trans: T,
         router: &StoreRouter<EK, ER>,
         store_meta: Arc<Mutex<StoreMeta<EK>>>,
+        snap_mgr: TabletSnapManager,
     ) -> Result<()>
     where
         T: Transport + 'static,
@@ -375,15 +376,6 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
 
-        let snap_mgr = TabletSnapManager::new(
-            tablet_factory
-                .tablets_path()
-                .join(Path::new("snap"))
-                .to_str()
-                .unwrap()
-                .to_owned(),
-        );
-        snap_mgr.init()?;
         let mut read_runner = ReadRunner::new(router.clone(), raft_engine.clone());
         read_runner.set_snap_mgr(snap_mgr);
         let read_scheduler = workers
--- a/components/raftstore-v2/src/batch/store.rs
+++ b/components/raftstore-v2/src/batch/store.rs
@@ -2,6 +2,7 @@
 
 use std::{
     ops::{Deref, DerefMut},
+    path::Path,
     sync::{Arc, Mutex},
     time::Duration,
 };
@@ -21,7 +22,7 @@ use kvproto::{
 use raft::INVALID_ID;
 use raftstore::store::{
     fsm::store::PeerTickBatch, local_metrics::RaftMetrics, Config, ReadRunner, ReadTask,
-    StoreWriters, Transport, WriteSenders,
+    SnapManagerBuilder, StoreWriters, Transport, WriteSenders,
 };
 use slog::Logger;
 use tikv_util::{
@@ -373,10 +374,21 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
         workers
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
-        let read_scheduler = workers.async_read_worker.start(
-            "async-read-worker",
-            ReadRunner::new(router.clone(), raft_engine.clone()),
+
+        let snap_mgr = SnapManagerBuilder::default().build(
+            tablet_factory
+                .tablets_path()
+                .join(Path::new("snap"))
+                .to_str()
+                .unwrap()
+                .to_owned(),
         );
+        snap_mgr.init()?;
+        let mut read_runner = ReadRunner::new(router.clone(), raft_engine.clone());
+        read_runner.set_snap_mgr(snap_mgr);
+        let read_scheduler = workers
+            .async_read_worker
+            .start("async-read-worker", read_runner);
 
         let mut builder = StorePollerBuilder::new(
             cfg.clone(),
--- a/components/raftstore-v2/src/operation/ready/snapshot.rs
+++ b/components/raftstore-v2/src/operation/ready/snapshot.rs
@@ -67,6 +67,8 @@ impl PartialEq for SnapState {
 
 pub struct GenSnapTask {
     region_id: u64,
+    // The snapshot will be sent to the peer.
+    to_peer: u64,
     // Fill it when you are going to generate the snapshot.
     // index used to check if the gen task should be canceled.
     index: Arc<AtomicU64>,
@@ -77,9 +79,15 @@ pub struct GenSnapTask {
 }
 
 impl GenSnapTask {
-    pub fn new(region_id: u64, index: Arc<AtomicU64>, canceled: Arc<AtomicBool>) -> GenSnapTask {
+    pub fn new(
+        region_id: u64,
+        to_peer: u64,
+        index: Arc<AtomicU64>,
+        canceled: Arc<AtomicBool>,
+    ) -> GenSnapTask {
         GenSnapTask {
             region_id,
+            to_peer,
             index,
             canceled,
             for_balance: false,
@@ -134,6 +142,7 @@ impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
         snap_task.index.store(last_applied_index, Ordering::SeqCst);
         let gen_tablet_sanp_task = ReadTask::GenTabletSnapshot {
             region_id: snap_task.region_id,
+            to_peer: snap_task.to_peer,
             tablet: self.tablet().clone(),
             region_state: self.region_state().clone(),
             last_applied_term,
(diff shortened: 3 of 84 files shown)


components/engine_panic/src/misc.rs
```diff
@@ -74,4 +76,8 @@ impl MiscExt for PanicEngine {
     fn is_stalled_or_stopped(&self) -> bool {
         panic!()
     }
+
+    fn checkpoint_to(&self, path: &[PathBuf], size_to_flush: u64) -> Result<()> {
```
BusyJay, line 80: I think the API was added.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -267,14 +275,25 @@ impl<EK: KvEngine, ER: RaftEngine> Storage<EK, ER> {
...
+        if !res.success {
+            self.cancel_generating_snap(None);
+            return false;
+        }
+        let snap = res.snapshot;
         let mut snap_state = self.snap_state_mut();
         let SnapState::Generating {
             ref canceled,
             ref index,
          } = *snap_state else { return false };
```
bufferflies, line 288: this condition sometime is mistake.maybe use `!= ` is better.

components/raftstore/src/store/worker/region.rs
```diff
@@ -310,11 +309,6 @@ where
         }
 
         let start = Instant::now();
-        let _io_type_guard = WithIoType::new(if for_balance {
```
BusyJay, line 313: Why remove this?

components/raftstore/src/store/worker/mod.rs
```diff
@@ -25,7 +25,7 @@ pub use self::{
     cleanup_sst::{Runner as CleanupSstRunner, Task as CleanupSstTask},
     compact::{Runner as CompactRunner, Task as CompactTask},
     consistency_check::{Runner as ConsistencyCheckRunner, Task as ConsistencyCheckTask},
-    metrics::TLS_LOCAL_READ_METRICS,
+    metrics::{SNAP_COUNTER, SNAP_HISTOGRAM, TLS_LOCAL_READ_METRICS},
```
BusyJay, line 28: We can make the `metrics` module public directly.

components/raftstore/src/store/worker/region.rs
nolouch: misunderstand it useless.

components/raftstore-v2/src/operation/ready/snapshot.rs
nolouch: here is else.

components/raftstore/src/store/snap.rs
```diff
@@ -1492,6 +1492,11 @@ impl SnapManager {
         path.to_str().unwrap().to_string()
     }
 
+    pub fn get_path_for_tablet_checkpointer(&self, key: &SnapKey) -> PathBuf {
```
tonyxuqqi, line 1495: minor: checkpointer ---> checkpoint
Maybe the fn can be renamed as get_tablet_checkpoint_path(..)

components/raftstore/src/store/async_io/read.rs
```diff
@@ -141,11 +168,92 @@ where
...
+                        success,
+                    },
+                );
             }
         }
     }
 }
+
+/// Used for create checkpointer. return true if success.
+fn create_checkpointer_for_snap<EK: KvEngine>(
```
tonyxuqqi, line 233: minor: checkpointer--->checkpoint

components/raftstore/src/store/async_io/read.rs
```diff
@@ -141,11 +168,92 @@ where
...
+                self.snap_mgr().register(key.clone(), SnapEntry::Generating);
+                defer!(self.snap_mgr().deregister(&key, &SnapEntry::Generating));
+                let mut snapshot = Snapshot::default();
+                // Set snapshot metadata.
+                snapshot.mut_metadata().set_index(key.idx);
+                snapshot.mut_metadata().set_term(key.term);
+                let conf_state = util::conf_state_from_region(region_state.get_region());
+                snapshot.mut_metadata().set_conf_state(conf_state);
+                // Set snapshot data.
+                let mut snap_data = RaftSnapshotData::default();
```
tonyxuqqi, line 205: should we increase the SNAPSHOT_VERSION since the snapshot data's format is changed.

components/raftstore/src/store/async_io/read.rs
```diff
@@ -141,11 +168,92 @@ where
                     },
                 );
             }
-            ReadTask::GenTabletSnapshot { region_id, .. } => {
-                // TODO: implement generate tablet snapshot for raftstore v2
-                self.notifier
-                    .notify_snapshot_generated(region_id, Box::new(RaftSnapshot::default()));
+
+            ReadTask::GenTabletSnapshot {
```
tonyxuqqi, line 172: In v1,  the similar logic is in worker/region.rs.  And async_io/read.rs is shared between v1 and v2 too.
I'm wondering if it's better to have this logic in region.rs too.

components/raftstore/src/store/async_io/read.rs
bufferflies: In this PR #13640, I have mark this version is SNAPSHOP_VERSION_V2(3) to decide how to send/recv snapshot directory?

components/raftstore-v2/src/batch/store.rs
```diff
@@ -373,10 +374,21 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
         workers
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
-        let read_scheduler = workers.async_read_worker.start(
-            "async-read-worker",
-            ReadRunner::new(router.clone(), raft_engine.clone()),
+
+        let snap_mgr = SnapManagerBuilder::default().build(
```
BusyJay, line 378: Why do we need a manager?

components/raftstore/src/store/worker/region.rs
```diff
@@ -315,7 +315,7 @@ where
         } else {
             IoType::Replication
         });
-
+
```
BusyJay, line 318: Useless blank

components/raftstore/src/store/snap.rs
nolouch: I follower `checkpointer`  with https://github.com/tikv/tikv/pull/13693.

components/raftstore-v2/src/batch/store.rs
nolouch: yes, here still use snap manager to register the `snap_entry`, and do some statistics or state check same as old way.

components/raftstore/src/store/async_io/read.rs
nolouch: it changed a lot. we discussed that unify snapshot handle to async workers in v2. see https://github.com/tikv/tikv/pull/13555#issuecomment-1296563953.

components/raftstore/src/store/async_io/read.rs
nolouch: I follower `checkpointer` with https://github.com/tikv/tikv/pull/13693.

nolouch:
PTAL @BusyJay

components/raftstore-v2/src/batch/store.rs
BusyJay: All these can be done without a manager. We don't even need to register it.

components/raftstore/src/store/async_io/read.rs
```diff
@@ -66,10 +75,16 @@ pub struct FetchedLogs {
     pub logs: Box<RaftlogFetchResult>,
 }
 
+#[derive(Debug)]
+pub struct GenSnapRes {
```
BusyJay, line 79: Isn't it just `Option<Box<Snapshot>>` or `Result<Box<Snapshot>>`?

components/raftstore/src/store/async_io/read.rs
BusyJay, line 233: Why not inline the function?

components/raftstore/src/store/async_io/read.rs
```diff
@@ -141,11 +168,92 @@ where
...
+    if checkpointer_path.as_path().exists() {
+        // Remove the old checkpoint directly.
+        let Ok (_) = std::fs::remove_dir_all(checkpointer_path.as_path()).map_err(|e| {
+            error!("failed to remove old checkpoint"; "path" => %checkpointer_path.as_path().display(), "err" => %e);
+            e
+        }) else { return false };
+    }
+    // Here not checkpoint to a temporary directory first, the temporary directory
+    // logic already implemented in rocksdb.
+    let Ok(mut checkpointer) = tablet.new_checkpointer().map_err(|e| {
```
BusyJay, line 248: Perhaps return a `Result` is more straightforward.

components/raftstore/src/store/async_io/read.rs
tonyxuqqi: > In this PR #13640, I have mark this version is SNAPSHOP_VERSION_V2(3) to decide how to send/recv snapshot directory?

Got it.

components/raftstore/src/store/async_io/read.rs
tonyxuqqi: > it changed a lot. we discussed that unify snapshot handler to async workers in v2. see [#13555 (comment)](https://github.com/tikv/tikv/pull/13555#issuecomment-1296563953).

I share the same worry as Bokang that mixing them together could impact raft log fetch or vice versa. For now I'm OK with that. May need to revisit it if performance is impacted by this later.

components/raftstore-v2/src/batch/store.rs
tonyxuqqi: > All these can be done without a manager. We don't even need to register it.

What's the benefit of doing it without a manager? Is there any key difference of snapshot management between v1 and v2? @BusyJay

tonyxuqqi (approved):

components/raftstore-v2/src/batch/store.rs
```diff
@@ -375,7 +375,7 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
 
-        let snap_mgr = SnapManagerBuilder::default().build(
+        let snap_mgr = TabletSnapManager::new(
             tablet_factory
                 .tablets_path()
                 .join(Path::new("snap"))
```
bufferflies, line 381: In snapshot-v1，the snapshot path is: data_dir/snap. 
And in past , the snapshot root path is : data_dir/tablets/snap.
I don't know this is belongs to the expect.

components/raftstore-v2/src/batch/store.rs
nolouch: PTAL @BusyJay

BusyJay: We don't need GC anymore, just delete the files when unused. Generating a snapshot is not as expansive as before.

components/raftstore/src/store/snap.rs
```diff
@@ -1886,6 +1891,45 @@ impl SnapManagerBuilder {
...
+        if !path.is_dir() {
+            return Err(io::Error::new(
+                ErrorKind::Other,
+                format!("{} should be a directory", path.display()),
+            ));
+        }
+        Ok(())
+    }
+
+    pub fn get_tablet_checkpointer_path(&self, key: &SnapKey) -> PathBuf {
```
BusyJay, line 1927:
```suggestion
    pub fn tablet_checkpointer_path(&self, key: &SnapKey) -> PathBuf {
```

components/raftstore/src/store/snap.rs
```diff
@@ -1886,6 +1891,45 @@ impl SnapManagerBuilder {
...
+        let prefix = format!("{}_{}", SNAP_GEN_PREFIX, key);
```
BusyJay, line 1928: I suggest to add the target peer id to the path name, otherwise, it may be conflict between peers.

components/raftstore/src/store/snap.rs
```diff
@@ -1886,6 +1891,45 @@ impl SnapManagerBuilder {
...
+    // directory to store snapfile.
+    base: String,
+}
+
+impl TabletSnapManager {
+    pub fn new<T: Into<String>>(path: T) -> Self {
+        Self { base: path.into() }
+    }
+
+    pub fn init(&self) -> io::Result<()> {
```
BusyJay, line 1911: Why not do this immediately in `new`?

nolouch: will add more logic such as clean the stale `checkpointer` in here，same as v1.

nolouch:
PTAL @BusyJay

components/raftstore/src/store/async_io/read.rs
```diff
@@ -88,9 +103,38 @@ impl<EK: KvEngine, ER: RaftEngine, N: AsyncReadNotifier> ReadRunner<EK, ER, N> {
...
+    pub fn set_snap_mgr(&mut self, mgr: TabletSnapManager) {
+        self.sanp_mgr = Some(mgr);
+    }
+
+    #[inline]
+    fn snap_mgr(&self) -> &TabletSnapManager {
+        self.sanp_mgr.as_ref().unwrap()
+    }
+
+    fn create_checkpointer_for_snap(
```
BusyJay, line 121:
```suggestion
    fn generate_snap(
```

BusyJay (commented):
Rest LGTM

components/raftstore-v2/src/batch/store.rs
```diff
@@ -373,10 +374,21 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
...
             .store_writers
             .spawn(store_id, raft_engine.clone(), None, router, &trans, &cfg)?;
-        let read_scheduler = workers.async_read_worker.start(
-            "async-read-worker",
-            ReadRunner::new(router.clone(), raft_engine.clone()),
+
+        let snap_mgr = TabletSnapManager::new(
+            tablet_factory
+                .tablets_path()
+                .join(Path::new("snap"))
```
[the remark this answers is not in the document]
nolouch: add a `tablets_path` in here. PTAL @BusyJay .

components/engine_traits/src/engine.rs
```diff
@@ -343,6 +346,10 @@ where
         PathBuf::from(&self.root_path)
     }
 
+    fn tablets_snap_path(&self) -> PathBuf {
+        PathBuf::from(&self.root_path).join("tablets_snap")
```
BusyJay, line 350: I don't think this is something `TabletFactory` should care. Why not get the value from configuration directly?

nolouch: here config is `raftstore` config, do not contains KV storage data path. I think it should be initialize out of raftstore, I change it  with pass the param.

PR #13737, opened 2022-11-04
raftstore-v2: add proposal conflict checker

### LINKED ISSUE #12842: Support physical isolation between regions

BusyJay:
Issue Number: Close #12842 

What's Changed:
```commit-message
The implementation is a bit simplified and adjusted for v2, including:
- Not require term for every methods, because when leader's term is changed, it must become follower. In this case, the checker won't be accessed until it becomes leader again. So pass term in every methods is useless.
- Use the checker to detect if it's splitting and merging.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


components/raftstore-v2/src/operation/command/control.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    /// Check if the proposal is conflict with proposed admin commands in
+    /// current term.
+    ///
+    /// Returns None if passing the epoch check, otherwise returns the last
+    /// conflict conflict proposal meta.
+    pub fn check_conflict(
+        &mut self,
+        cmd_type: Option<AdminCmdType>,
```
tonyxuqqi, line 128: Could you please add a comment for None type cmd_type?  It's not straightforward without looking into the check_conflict implementation

components/raftstore-v2/src/operation/command/write/simple_write.rs
```diff
@@ -21,12 +22,14 @@ pub struct SimpleWriteEncoder {
     buf: Vec<u8>,
     channels: Vec<CmdResChannel>,
     size_limit: usize,
+    likely_proposed: bool,
```
tonyxuqqi, line 25: This parameter seems is same as "call_proposed_on_success"---its usage basically is call notify_proposed . I think changing the name to "call_proposed_on_success" is more clear.

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -395,6 +411,71 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
     pub fn on_wait_flush(&mut self, ch: crate::router::FlushChannel) {
         self.async_writer.subscirbe_flush(ch);
     }
+
+    pub fn on_role_changed<T>(&mut self, ctx: &mut StoreContext<EK, ER, T>, ready: &Ready) {
```
tonyxuqqi, line 415: Could you add a comment if there's any behavior change between v1 and v2.

components/raftstore-v2/src/operation/ready/mod.rs
```diff
@@ -395,6 +411,71 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+            }
+            self.proposal_control_mut().maybe_update_term(term);
+        }
+    }
+
+    /// If leader commits new admin commands, it may break lease assumption. So
+    /// we need to cancel lease whenever necessary.
+    ///
+    /// Note this method should be called before sending out any messages.
+    fn on_leader_commit_index_changed(&mut self, commit_index: u64) {
```
tonyxuqqi, line 463: Question: should split be handled as merge too ?

tonyxuqqi (approved):
Overall LGTM. Just a few minor comments.

components/raftstore-v2/src/operation/ready/mod.rs
BusyJay: No, the reason why merge is handled this way is because it will destroy the peer, so the lease needs to be destroyed.

components/raftstore-v2/src/operation/ready/mod.rs
BusyJay: There is no difference. Missing part is to be implemented.

BusyJay:
PTAL

components/raftstore-v2/src/operation/command/control.rs
```diff
@@ -0,0 +1,428 @@
...
+            self.pending_merge_index = 0;
+        }
+    }
+
+    /// Check if there is an on-going split command on current term.
+    ///
+    /// The answer is reliable only when the peer is leader.
+    #[inline]
+    pub fn is_splitting(&self) -> bool {
+        if self.proposed_admin_cmd.is_empty() {
```
Connor1996, line 228: it's covered below

components/raftstore-v2/src/operation/command/write/mod.rs
```diff
@@ -65,35 +73,38 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+                if let Err(mut e) = res {
+                    // TODO: query sibling regions.
+                    ctx.raft_metrics.invalid_proposal.epoch_not_match.inc();
+                    encoder.encode().1.report_error(cmd_resp::new_error(e));
+                    return;
+                }
+                // Only when it applies to current term, the epoch check can be reliable.
+                self.applied_to_current_term()
+            };
             let (data, chs) = encoder.encode();
```
Connor1996, line 105: Can we move `not leader` check before here? Then we can save time of encoding.

Connor1996: and do lazy encode until here?

components/raftstore-v2/src/operation/command/control.rs
BusyJay: It's just a shortcut for most cases. Iterator will take more instructions to find out it's empty.

components/raftstore-v2/src/operation/command/write/mod.rs
BusyJay: I don't understand. If it's not leader, then the proposal is rejected in the very beginning. For most cases, if it passes all the check, it's likely to be proposed, lazy encoding won't have much benefits but only adds more memory consumption and make code complicated.

PR #13742, opened 2022-11-06
debugger: fix compacting raftdb with tikv-ctl

### LINKED ISSUE #13515: tikv-ctl: compact raft will failed but the documents say it was allowed.
according to the tikv documents, `tikv-ctl compact` canbe apply on db "kv" and "raft", but it failed on raft because of the following codes:
https://github.com/tikv/tikv/blob/42d3222d3a5ae2f1e04e25077a33f0896725ffca/src/server/debug.rs#L172

guoxiangCN:

Issue Number: Close #13515

What's Changed:
Moves the function `get_db_from_type` in `Debugger` to a trait, and the tikv instance using Rocksdb as the raftlog backend is specially implemented, so that the compact raft db no longer reports errors.

Maybe I need to explain the following background: In the previous version, according to the description of tikv ctl tool in the official TIDB document, we can manually compact raftdb through the following instructions：
`./tikv-ctl  --host localhost:port compact -d raft`

However, after a version of 5. x, it can no longer be used ,even if we use Rocksdb instead of RaftEngine.
`DebugClient::compact: RpcFailure: 2-UNKNOWN "[src/server/debug.rs:169]: Get raft db is not allowed"`

The reason is that in order to support RaftEngine as an optional raftlog storage engine, the relevant code has a refactor, making this function unavailable at all.
However, the official document always says that it is supported, and we really need to compact raftdb in our actual application scenarios, so I propose this PR for repair.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Please add a release note.

```

### DIFF
--- a/src/server/debug.rs
+++ b/src/server/debug.rs
@@ -120,11 +120,35 @@ impl From<BottommostLevelCompaction> for debugpb::BottommostLevelCompaction {
     }
 }
 
+trait InnerRocksEngineExtractor {
+    fn get_db_from_type(&self, db: DbType) -> Result<&RocksEngine>;
+}
+
 #[derive(Clone)]
 pub struct Debugger<ER: RaftEngine> {
     engines: Engines<RocksEngine, ER>,
     reset_to_version_manager: ResetToVersionManager,
     cfg_controller: ConfigController,
+}
+
+impl<ER: RaftEngine> InnerRocksEngineExtractor for Debugger<ER> {
+    default fn get_db_from_type(&self, db: DbType) -> Result<&RocksEngine> {
+        match db {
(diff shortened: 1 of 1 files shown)


tabokie (commented):
Please add a unit test.

guoxiangCN:
> Please add a unit test.

done, also tested by a just-built binary of tikv

tabokie (approved):

Connor1996 (approved):
LGTM

PR #13744, opened 2022-11-07
storage: fix the flashback dead loop caused by a deleted key

### LINKED ISSUE #13743: flashback ddl is hang when some key's last mvcc record is delete
Release Version:   6.4.0
Edition:           Community
Git Commit Hash:   220a263c665b42a43cec3e38f2d9b46ffe074ca9
Git Commit Branch: heads/refs/tags/v6.4.0
UTC Build Time:    2022-11-04 07:20:37
Rust Version:      rustc 1.64.0-nightly (0f4bcadb4 2022-07-30)
Enable Features:   pprof-fp jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
[root@localhost bin.old.v6.4.0]# 

### Steps to reproduce
if the latest mvcc record is 'delete', we will skip this key during flashback, but for this reason, this key will not has  flashback commit ts write record, so this key will be processed as not flashback key. so will cause deadlock

### What did you expect?
flashback successfully
### What did happened?
flashback is hang

JmPotato:

Issue Number: Close #13743.

What's Changed:

```commit-message
The original thought is that when a key's latest MVCC record type is DELETE and the corresponding flashback operation is also DELETE, we skip it to avoid duplicated writing.
However, this will cause the flashback to fall into a dead loop since the key doesn't have the written record with the flashback `commit_ts` and the flashback will always try to write it forever.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -4978,6 +4978,87 @@ mod tests {
     }
 
     #[test]
+    fn test_flashback_to_version_deleted_key() {
+        let storage = TestStorageBuilderApiV1::new(MockLockManager::new())
+            .build()
+            .unwrap();
+        let (tx, rx) = channel();
+        let mut ts = TimeStamp::zero();
+        let (k, v) = (Key::from_raw(b"k"), b"v".to_vec());
+        // Write a key.
+        storage
+            .sched_txn_command(
+                commands::Prewrite::with_defaults(
+                    vec![Mutation::make_put(k.clone(), v.clone())],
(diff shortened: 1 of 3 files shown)


JmPotato:
/cc @Connor1996 @sticnarf 

Please help review this PR, thx.

Connor1996 (approved):

sticnarf (approved):

PR #13753, opened 2022-11-08
tikv_utils: make retry returns a `Send` future even the result type is `!Sync`

### LINKED ISSUE #13811: tikv_util: make `retrying` a `!Send` function can be `Send`
## Development Task
For now, if we are going to retry over a `!Send` future(for example, which's result is `!Sync`), the generated future by `tikv_util::stream::retry` isn't `Send`, which means is cannot be spawned into background.
However in fact, given we only need to fetch the last success or failure result, in fact we don't need to send the result into other threads, we can fix it.

YuJuncen:
Signed-off-by: hillium <yujuncen@pingcap.com>

Issue Number: Close #13811

What's Changed:

```commit-message
This PR have modified the `retry_ext` function in the `tikv_utils::stream` package, making the `final_result` not live across await points any more. So even the result type is `!Sync`, `retry` returns a `Send` future now. 
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


joccau (approved):
LGTM

kennytm (approved):

3pointer (approved):

YuJuncen:
@kennytm Please help to /merge this, thanks~

PR #13759, opened 2022-11-09
txn: save last change info in write records

sticnarf:

Issue Number: Ref #13694

What's Changed:

Only a one-line change to the code :)

```commit-message
In this commit, the commit action will save the last_change_ts and
versions_to_last_change in the lock to the write record.

It is unncessary to check the write type because it is checked during prewrite.
So, among the committable locks, only those of Lock type will have
last_change_ts and versions_to_last_change.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/actions/commit.rs
+++ b/src/storage/txn/actions/commit.rs
@@ -91,7 +91,8 @@ pub fn commit<S: Snapshot>(
         WriteType::from_lock_type(lock.lock_type).unwrap(),
         reader.start_ts,
         lock.short_value.take(),
-    );
+    )
+    .set_last_change(lock.last_change_ts, lock.versions_to_last_change);
 
     for ts in &lock.rollback_ts {
         if *ts == commit_ts {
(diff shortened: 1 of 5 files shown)


MyonKeminta (approved):

src/storage/txn/actions/commit.rs
```diff
@@ -320,4 +321,41 @@ pub mod tests {
...
     }
+
+    #[test]
+    fn test_inherit_last_change_info_from_lock() {
+        use pd_client::FeatureGate;
+
+        use crate::storage::txn::sched_pool::set_tls_feature_gate;
+
+        let feature_gate = FeatureGate::default();
+        feature_gate.set_version("6.5.0").unwrap();
```
TonsnakeLin, line 332: You have removed the `set version to 6.5.0` in case `test_mess()`, why does this case need to be set 6.5.0? TiKV version hasn't reached 6.5 currently?

sticnarf: In https://github.com/tikv/tikv/pull/13721, I removed the checks about `last_change_ts` in `test_mess`. So, it doesn't depend on this feature now.

When the PR was created, the bumping version PR was not merged yet.

TonsnakeLin (approved):

sticnarf:
PTAL @ekexium

ekexium (approved):

PR #13766, opened 2022-11-10
Increase read pool limit

### LINKED ISSUE #13690: the max size of read pool should be higher than the number of cpu
the max size of read pool is limited by number of cpu
https://github.com/tikv/tikv/blob/cbc302cd851a5c5dd7a4cdd9a1efa24eccfe08c6/src/read_pool.rs#L278
https://github.com/tikv/tikv/blob/a4dc37b0c07ee1b5bc7c60a5d8360666500ecdb4/src/config.rs#L1880
there is no such limit in v5.4.0

the upper limit of the read pool should be higher than number of cpu, i.e. 10x like other configurations, because the tasks in the pool do both cpu and io operations

v01dstar:

Issue Number: close #13690

What's Changed:

```commit-message
* Read pool size limit is removed. It was previously hard-coded to be
  the total number of cores. Limiting the number of threads to be less
  than the number of cores reduces the chance of context switch, but
  also makes TiKV prone to starvation problem (since read pool is
  currently FIFO), even if all the queries require no IO or IO is async
  (imagining there are number_of_cores clients sending big queries, and
  another numbers of clients sending small queries concurrently to a
  single TiKV server, and all of them require no IO). Thread starvation
  causes high tail latency which is even worse than context switching.
  According to the feature requester, increasing the number of threads
  significantly improved the tail latency in their environment. Thus, we
  should remove the limit, and leave it to the users.
```

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`: N/A
- Need to cherry-pick to the release branch: N/A

### Check List 

Tests 

- Unit test

Side effects

- Performance regression (in certain cases, iff the config is manually and mistakenly changed)
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Increase read pool limit to be 10X CPU quota.
```


BusyJay:
Is there any performance number?

v01dstar:
> Is there any performance number?

I invited the feature requester to leave a [comment](https://github.com/tikv/tikv/issues/13690#issuecomment-1316283666).

v01dstar:
Per jinpeng's suggestion offline, cap the number of threads used by read pool with a higher limit, instead of lifting it. Also, moves the `auto adjust` optimization into another PR.

src/config.rs
```diff
@@ -1964,15 +1961,6 @@ mod unified_read_pool_tests {
             ..cfg
         };
         invalid_cfg.validate().unwrap_err();
-        let invalid_cfg = UnifiedReadPoolConfig {
```
BusyJay, line 1967: The test case should be kept.

src/read_pool.rs
```diff
@@ -276,8 +276,11 @@ pub fn build_yatp_read_pool<E: Engine, R: FlowStatsReporter>(
             config.min_thread_count,
             config.max_thread_count,
             std::cmp::max(
-                UNIFIED_READPOOL_MIN_CONCURRENCY,
-                SysQuota::cpu_cores_quota() as usize,
+                std::cmp::max(
```
BusyJay, line 279: Why not follow the same style as in config?

src/config.rs
v01dstar: Reverted, but also changed the expectation accordingly.

src/read_pool.rs
v01dstar: Which "style" are you referring? If you mean this parameter should be set to  cpu_quota * 10 as in the config.rs, then it is a different things. I believe that this parameter was introduced for the "auto adjustment" feature, the original author's intention was
`config.min_thread_count < config.max_thread_count < this parameter (the real upper limit)`, in case the config.max_thread_count is set to a small value. I realized that the naming and logic are confusing, and was trying to change this, but other team members were conservative about changing it since it might change the user behavior. So, I kept it untouched.

zhangjinpeng1987 (approved):
LGTM

src/config/mod.rs
```diff
@@ -1808,10 +1808,7 @@ impl UnifiedReadPoolConfig {
                     .into(),
             );
         }
-        let limit = cmp::max(
```
BusyJay, line 1811: Why remove the check?

src/config/mod.rs
```diff
@@ -1900,11 +1897,7 @@ mod unified_read_pool_tests {
         };
         invalid_cfg.validate().unwrap_err();
         let invalid_cfg = UnifiedReadPoolConfig {
-            min_thread_count: 1,
```
BusyJay, line 1903: Why remove this?

src/config/mod.rs
v01dstar: If you mean why do I remove the comparsion. The reason is `cpu_cores_quota() * 10` is guaranteed to be bigger than `UNIFIED_READPOOL_MIN_CONCURRENCY`, besides, `UNIFIED_READPOOL_MIN_CONCURRENCY` is not tunable, so there is no need to have this comparison.

BusyJay (approved):

Issue #13760, opened 2022-11-09
Provide a way to get the regions info within a given key range


JmPotato:
## Development Task

Sometimes we may need to know which regions are located within a certain key range. Though this could be finished by querying the PD API, in some special cases such as region range hole, we MUST get it according to the TiKV data itself. Currently, `tikv-ctl` only supports dumping the regions info with given IDs. It would be nice if we can provide a way to get the regions info within a given key range.

PR #13768, opened 2022-11-10
cmd: support tikv-ctl to get the regions info within a given key range

HuSharp:

Issue Number: ref #13760

What's Changed:

```commit-message
Provide a way to get the regions info within a given key range just like which in [pd-ctl](https://docs.pingcap.com/zh/tidb/dev/pd-control#region-key---formatrawencodehex-key)

Also support limit for tikv-ctl raft region
```

### Check List 

Tests 

```
test in local tiup playground
```
`./tikv-ctl --data-dir ~/.tiup/data/XXX/tikv-0/data raft region --start XXXX --end YYYY --limit N`
image

`./tikv-ctl --data-dir ~/.tiup/data/XXX/tikv-0/data raft region --start XXXX --limit N`
9IUVAp2yyT
`./tikv-ctl --data-dir ~/.tiup/data/XXX/tikv-0/data raft region --end XXXX`
image

`./tikv-ctl --data-dir ~/.tiup/data/AAA/tikv-0/data raft region --limit N`
image

### Release note 

```release-note
None
```


cmd/tikv-ctl/src/executor.rs
```diff
@@ -151,17 +152,52 @@ pub trait DebugExecutor {
...
                 let region_state = r.region_local_state.as_ref();
                 if region_state.map_or(false, |s| s.get_state() == PeerState::Tombstone) {
                     return;
                 }
             }
+            // Check if the region is in the specified range
+            fn contains(r: &RegionInfo, start_key: &Vec<u8>, end_key: &Option<Vec<u8>>) -> bool {
+                let region = r
+                    .region_local_state
+                    .clone()
```
JmPotato, line 183: Why clone here? Can't we use `as_ref`?

cmd/tikv-ctl/src/executor.rs
```diff
@@ -151,17 +152,52 @@ pub trait DebugExecutor {
...
+            }
             let r = self.get_region_info(region_id);
             if skip_tombstone {
                 let region_state = r.region_local_state.as_ref();
                 if region_state.map_or(false, |s| s.get_state() == PeerState::Tombstone) {
                     return;
                 }
             }
+            // Check if the region is in the specified range
+            fn contains(r: &RegionInfo, start_key: &Vec<u8>, end_key: &Option<Vec<u8>>) -> bool {
```
JmPotato, line 180: Why create the function inside a loop?

cmd/tikv-ctl/src/executor.rs
```diff
@@ -151,17 +152,52 @@ pub trait DebugExecutor {
...
-    fn dump_region_info(&self, region_ids: Option<Vec<u64>>, skip_tombstone: bool) {
+    fn dump_region_info(
+        &self,
+        region_ids: Option<Vec<u64>>,
+        start_key: Option<Vec<u8>>,
+        end_key: Option<Vec<u8>>,
+        mut limit: usize,
+        skip_tombstone: bool,
+    ) {
+        if limit > DEFAULT_REGION_INFO_LIMIT {
```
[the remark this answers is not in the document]
HuSharp: Thx, get it!

cmd/tikv-ctl/src/executor.rs
```diff
@@ -151,17 +152,52 @@ pub trait DebugExecutor {
...
         let mut region_objects = serde_json::map::Map::new();
         for region_id in region_ids {
+            if limit > 0 && region_objects.len() >= limit {
+                break;
+            }
             let r = self.get_region_info(region_id);
             if skip_tombstone {
                 let region_state = r.region_local_state.as_ref();
                 if region_state.map_or(false, |s| s.get_state() == PeerState::Tombstone) {
                     return;
```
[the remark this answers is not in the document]
HuSharp: I think `limit` just for all `region_objects` thus there will be `break`.
Do u mean that `limit` is for single `region`'s object len?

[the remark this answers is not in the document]
JmPotato: I mean line 176, not 170.

HuSharp:
address all comments, plz take a look thx! @JmPotato

cmd/tikv-ctl/src/cmd.rs
```diff
@@ -604,10 +603,22 @@ pub enum RaftCmd {
...
 
+        #[structopt(long, default_value = "")]
+        /// hex start key
+        start: String,
+
+        #[structopt(long, default_value = "")]
+        /// hex end key
+        end: String,
+
+        #[structopt(long, default_value = "0", max_values = 10240)]
```
JmPotato, line 618: What about setting the max value to 10000? 10240 sounds weird 😂.

cmd/tikv-ctl/src/util.rs
```diff
@@ -62,6 +62,19 @@ pub fn perror_and_exit<E: Error>(prefix: &str, e: E) -> ! {
     tikv_util::logger::exit_process_gracefully(-1);
 }
 
+// Check if the region is in the specified range
+pub fn included_region_in_range(region_range: (&[u8], &[u8]), check_range: (&[u8], &[u8])) -> bool {
```
JmPotato, line 66: I suggest not using tuples to avoid later `xxx.0` and `xxx.1` which makes the code hard to read. Maybe it could look like this:

```rust
pub fn check_range_included(key_range: KeyRange, key_range_limit: KeyRange) -> bool {
```

`KeyRange` is an existing struct you can find here: https://github.com/tikv/tikv/blob/master/components/raftstore/src/store/util.rs#L392

cmd/tikv-ctl/src/cmd.rs
HuSharp: sUavxRWr1a
is to synchronize the pd-ctl

JmPotato (approved):

JmPotato (approved):

cmd/tikv-ctl/src/util.rs
```diff
@@ -62,8 +63,39 @@ pub fn perror_and_exit<E: Error>(prefix: &str, e: E) -> ! {
     tikv_util::logger::exit_process_gracefully(-1);
 }
 
+// Check if the part of region's `key_range` is in the specified
+// `key_range_limit`.
+pub fn check_range_included(key_range: &KeyRange, key_range_limit: &KeyRange) -> bool {
```
tabokie, line 68: You mean "intersecetd"?

HuSharp: Yes, you've made it clearer. I'll change relevant annotations.

cmd/tikv-ctl/src/util.rs
```diff
@@ -62,8 +63,39 @@ pub fn perror_and_exit<E: Error>(prefix: &str, e: E) -> ! {
...
+    // be considered as included.
+    if key_range.get_start_key() <= key_range_limit.get_start_key()
+        && (key_range.get_end_key().is_empty() /* when region end_key is infinite */
+            || (!key_range_limit.get_end_key().is_empty() /* when limit end_key is not infinite */
+                    && key_range.get_end_key() >= key_range_limit.get_end_key()))
+    {
+        return true;
+    }
+    false
+}
```
tabokie, line 93:
```rust
if !a.get_end_key().is_empty() && !b.get_start_key().is_empty() && a.get_end_key() < b.get_start_key() {
  false
}
if !b.get_end_key().is_empty() && !a.get_start_key().is_empty() && b.get_end_key() < a.get_start_key() {
  false
}
true
```

HuSharp: Thx for your help, the not contained case is indeed simpler than the contained case.
BTW, since it is a left-closed right-open interval, I added the = judgment:)
Plz take a look, thx!!! @tabokie @JmPotato

cmd/tikv-ctl/src/cmd.rs
```diff
@@ -604,10 +604,22 @@ pub enum RaftCmd {
...
 
+        #[structopt(long, default_value = "")]
+        /// hex start key
+        start: String,
+
+        #[structopt(long, default_value = "")]
+        /// hex end key
+        end: String,
+
+        #[structopt(long, default_value = "0", max_values = MAX_REGION_LIMIT)]
```
tabokie, line 619: Why have an upper bound when the default behavior is infinite (limit=0)?

tabokie (approved):

cmd/tikv-ctl/src/cmd.rs
HuSharp: Sry, to sync with pd-ctl, I'll set this to 16.

tabokie: I mean, if you allow user to print infinite amount of regions (limit = 0), then it doesn't make sense to limit the specific value of this field. I think it would be better to use 16 as default, and assign no restriction for max value.

HuSharp: Get it, already modified.

PR #13770, opened 2022-11-10
copr: fix _ pattern in `like` behavior for old collation

### LINKED ISSUE #13769: When the new collation is not enabled, `_` pattern in like misbehaved
The TiKV always uses the charset of the collation to decode and compare strings for `like` function. However, if the new collation is not enabled, the collation information is not pushed down from TiDB to TiKV, and TiKV cannot get the charset information from it.

The simple solution is getting the charset information from the original arguments.

Related issue: https://github.com/pingcap/tidb/issues/32323

YangKeao:

Issue Number: Close #13769

This function uses the charset information from the arguments to decode character, but compare them with the collator pushed down from TiDB.

### Related changes

Not sure. I'm not clear about the cherry-pick strategy of the bugfix :thinking:. I'll read some related documents tomorrow.

### Check List 

Tests 

- [x] Unit test

### Release note 

```release-note
```

### DIFF
--- a/components/tidb_query_expr/src/impl_like.rs
+++ b/components/tidb_query_expr/src/impl_like.rs
@@ -6,17 +6,21 @@ use tidb_query_datatype::codec::{collation::*, data_type::*};
 
 #[rpn_fn]
 #[inline]
-pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> Result<Option<i64>> {
+pub fn like<C: Collator, TargetCharset: Charset, PatternCharset: Charset>(
+    target: BytesRef,
+    pattern: BytesRef,
+    escape: &i64,
+) -> Result<Option<i64>> {
     let escape = *escape as u32;
     // current search positions in pattern and target.
     let (mut px, mut tx) = (0, 0);
     // positions for backtrace.
     let (mut next_px, mut next_tx) = (0, 0);
     while px < pattern.len() || tx < target.len() {
-        if let Some((c, mut poff)) = C::Charset::decode_one(&pattern[px..]) {
+        if let Some((c, mut poff)) = PatternCharset::decode_one(&pattern[px..]) {
             let code: u32 = c.into();
             if code == '_' as u32 {
-                if let Some((_, toff)) = C::Charset::decode_one(&target[tx..]) {
+                if let Some((_, toff)) = TargetCharset::decode_one(&target[tx..]) {
                     px += poff;
                     tx += toff;
                     continue;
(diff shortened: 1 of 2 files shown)


components/tidb_query_expr/src/lib.rs
```diff
@@ -91,10 +92,35 @@ fn map_compare_in_string_sig(ret_field_type: &FieldType) -> Result<RpnFnMeta> {
...
 
-fn map_like_sig(ret_field_type: &FieldType) -> Result<RpnFnMeta> {
+fn map_like_sig(ret_field_type: &FieldType, children: &[Expr]) -> Result<RpnFnMeta> {
+    // Must use the charset from args, but not the collator from ret_field_type.
+    // Because the TiDB doesn't push down the collation information when the new
+    // collation framework is not enabled, and always use the binary collation.
+    // However, the `_` pattern considers not only the order of strings, but
+    // also the number of characters. Some characters more than 1 bytes cannot
+    // be matched by `_` if the new collation framework is not enabled.
+    let target_collation = children[0]
```
YangKeao, line 102: I'm not sure whether the arguments have been validated in this function, and whether I could get the `children[0]` and `children[1]` safely without validating the `children.len() >= 2` :thinking:. Could anyone confirm for me?

YangKeao:
@wjhuang2016 PTAL

components/tidb_query_expr/src/impl_like.rs
```diff
@@ -35,13 +39,13 @@ pub fn like<C: Collator>(target: BytesRef, pattern: BytesRef, escape: &i64) -> R
...
+                    poff = if let Some((_, off)) = PatternCharset::decode_one(&pattern[px..]) {
                         off
                     } else {
                         break;
                     }
                 }
-                if let Some((_, toff)) = C::Charset::decode_one(&target[tx..]) {
+                if let Some((_, toff)) = TargetCharset::decode_one(&target[tx..]) {
                     if let Ok(std::cmp::Ordering::Equal) =
                         C::sort_compare(&target[tx..tx + toff], &pattern[px..px + poff])
```
sticnarf, line 50: If pattern and target have different charsets, does this comparison still make sense :thinking: ?

YangKeao: It's technically fine for TiKV in this situation, but not actually correct...

TiKV only have two charset format (for storage): binary or utf8mb4, and we only need to know whether they are equal, then using any specified collator to compare them is fine :thinking:.

If TiKV supports more charset in the storage (e.g. GBK, ...), this function need to be reworked (anyway, the original implementation also needs, because you cannot use a single charset to decode two binaries with different charsets).

YangKeao: I'll add some comments to describe the limitation.

YangKeao:
This PR is actually incorrect for this test case:

```
create table t (a varchar(255) character set utf8mb4 collate utf8mb4_bin, b varchar(255) character set binary collate binary);
insert into t values ("测试", "测%");
select * from t where a like b;
```

I'll close and consider other ways to solve this problem.

PR #13773, opened 2022-11-11
raftstore-v2: pd worker

tabokie:

What's Changed:

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore-v2/src/operation/pd.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        stats.set_sending_snap_count(0);
+        stats.set_receiving_snap_count(0);
+
+        stats.set_start_time(self.start_time().unwrap() as u32);
+
+        stats.set_bytes_written(0);
+        stats.set_keys_written(0);
+        stats.set_is_busy(false);
```
tonyxuqqi, line 53: Could these store level statistics be added in another PR?

tonyxuqqi (approved):

Connor1996 (commented):
Could you give a description of how much is changed compared to v1 pd worker

tabokie:
> Could you give a description of how much is changed compared to v1 pd worker

Nothing changed (except for the file structure). Many non-critical parts are still missing and commented. @Connor1996

components/raftstore-v2/src/worker/pd/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+mod split;
+mod store_heartbeat;
+mod update_max_timestamp;
+
+pub use heartbeat::HeartbeatTask;
+
+use crate::{batch::StoreRouter, router::PeerMsg};
+
+pub enum Task {
+    Heartbeat(HeartbeatTask),
```
Connor1996, line 32: call it `RegionHeartbeat`?

Connor1996: same for the file name

Connor1996 (commented):
rest LGTM

Connor1996 (approved):
LGTM

PR #13776, opened 2022-11-12
raftstore-v2: support send/recv tablet snapshot

bufferflies:

What's Changed:
```
raftstore-v2: support send/recv snapshot
This PR implementation send/recv snapshot relative changes.
-  new tablet runner to handler send/recv snapshot
-  add Limiter to restrict the io bandwidth 
Next part will implement apply snapshot
```
### Related changes

Base on https://github.com/tikv/tikv/pull/13388

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
None
```


src/server/raft_client.rs
```diff
@@ -428,6 +432,7 @@ struct AsyncRaftSender<R, M, B, E> {
     buffer: B,
     router: R,
     snap_scheduler: Scheduler<SnapTask>,
+    tablet_scheduler: Scheduler<SnapTask>,
```
nolouch, line 435: I think we can only use one scheduler，only need exist one.

bufferflies: yes, the two scheduler should not work in one time.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+                let task = async move {
+                    let result = recv_snap(stream, sink, snap_mgr, raft_router, limit).await;
+                    recving_count.fetch_sub(1, Ordering::SeqCst);
+                    if let Err(e) = result {
+                        error!("failed to recv snapshot"; "err" => %e);
+                    }
+                };
+                self.pool.spawn(task);
+            }
+            Task::Send { addr, msg, cb } => {
```
tonyxuqqi, line 403: Where's the code to call Task::Send

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+            let mut chunk = SnapshotChunk::default();
+            chunk.set_data(buffer);
+            sender
+                .feed((chunk, WriteFlags::default().buffer_hint(true)))
+                .await?;
+            if new_len < SNAP_CHUNK_LEN {
+                break;
+            }
+            buffer = Vec::with_capacity(SNAP_CHUNK_LEN);
+            off = 0
```
tonyxuqqi, line 148: the second SnapshotChunk will not have the file name?  But the receiver side will not know it and it will try to parse the file name anyway which will fail. I think here the buf should always include the file name

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+        };
+        // the format of chunk:
+        // |--name_len--|--name--|--content--|
+        let len = chunk[0] as usize;
+        let file_name = box_try!(std::str::from_utf8(&chunk[1..len + 1]));
+        let p = path.join(file_name);
+        let mut f = File::create(&p)?;
+        let mut size = chunk.len() - len - 1;
+        f.write_all(&chunk[len + 1..])?;
+        while chunk.len() >= SNAP_CHUNK_LEN {
```
tonyxuqqi, line 236: > while chunk.len() >= SNAP_CHUNK_LEN
It's not reliable to use the chunk.len() >=SNAP_CHUNK_LEN as the indicator that the previous file is ended or not. 
It's possible that the file's last part is exactly same as SNAP_CHUNK_LEN and then you'll merge the two files into one.

src/server/tablet_snap.rs
bufferflies: It will be called by `RaftClient` if the sending message type is the snapshot type , it will call `send_snapshot_sock`.

src/server/tablet_snap.rs
bufferflies: yes，the second snapshot chunk should not contain the file name, the file name only exists in the first snapshot chunk. We can use the snapshot chunk's data len to check the file has sent all . if the length < 1M , it means that the file has sent all. BTW, if the data's len is 1M exactly, it will send one snapshot chunk, but not carry any data .

tonyxuqqi (approved):

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+/// It will first send the normal raft snapshot message and then send the
+/// snapshot file.
+pub fn send_snap(
+    env: Arc<Environment>,
+    mgr: TabletSnapManager,
+    security_mgr: Arc<SecurityManager>,
+    cfg: &Config,
+    addr: &str,
+    msg: RaftMessage,
+    limiter: Limiter,
```
nolouch, line 167: Limiter can combine into `TabletSnapManager`?

components/raftstore/src/store/snap.rs
```diff
@@ -1959,6 +1960,28 @@ impl TabletSnapManager {
...
+        let prefix = format!("{}_{}", SNAP_REV_PREFIX, key);
+        PathBuf::from(&self.base).join(&prefix)
+    }
+    pub fn get_tmp_name_for_recv(&self, key: &TabletSnapKey) -> PathBuf {
+        let prefix = format!("{}_{}{}", SNAP_REV_PREFIX, key, TMP_FILE_SUFFIX);
+        PathBuf::from(&self.base).join(&prefix)
+    }
+
+    pub fn delete_snapshot(&self, key: &TabletSnapKey) -> bool {
+        let path = self.get_tablet_checkpointer_path(key);
```
tabokie, line 1974: Let's fix some name usage while you are at it. The checkpointer is a stateless helper struct that can be used to create checkpoints at many given checkpoint_path-s. Here we should use `get_tablet_checkpoint_path`.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+    key: TabletSnapKey,
+    raft_msg: RaftMessage,
+    io_type: IoType,
+    start: Instant,
+}
+
+impl RecvTabletSnapContext {
+    fn new(head_chunk: Option<SnapshotChunk>) -> Result<Self> {
+        // head_chunk is None means the stream is empty.
+        let mut head = head_chunk.ok_or_else(|| Error::Other("empty gRPC stream".into()))?;
```
tabokie, line 62: Move this part outside this function.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+        let data = snapshot.get_data();
+        let mut snapshot_data = RaftSnapshotData::default();
+        snapshot_data.merge_from_bytes(data)?;
+        let snapshot_meta = snapshot_data.take_meta();
+        let io_type = if snapshot_meta.get_for_balance() {
+            IoType::LoadBalance
+        } else {
+            IoType::Replication
+        };
+        let _with_io_type = WithIoType::new(io_type);
```
tabokie, line 84: This is a RAII style guard (i.e. the type is erased if `_with_io_type` goes out of scope). You need to save it in the context.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+    mgr: &TabletSnapManager,
+    mut sender: impl Sink<(SnapshotChunk, WriteFlags), Error = Error> + Unpin,
+    msg: RaftMessage,
+    key: TabletSnapKey,
+    limiter: Limiter,
+) -> Result<u64> {
+    let path = mgr.get_tablet_checkpointer_path(&key);
+    info!("begin to send snapshot file";"snap_key" => %key);
+    let files = fs::read_dir(&path)?
+        .map(|d| Ok(d?.path()))
```
tabokie, line 115:
```suggestion
        .map(|f| Ok(f?.path()))
```

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+        .collect::<Result<Vec<_>>>()?;
+    let mut total_sent = msg.compute_size() as u64;
+    let mut chunk = SnapshotChunk::default();
+    chunk.set_message(msg);
+    sender
+        .feed((chunk, WriteFlags::default().buffer_hint(true)))
+        .await?;
+    for path in files {
+        let name = path.file_name().unwrap().to_str().unwrap();
```
tabokie, line 124: Check if it's a directory.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+            total_sent += new_len as u64;
+            unsafe {
+                buffer.set_len(new_len);
+            }
+            let mut chunk = SnapshotChunk::default();
+            chunk.set_data(buffer);
+            sender
+                .feed((chunk, WriteFlags::default().buffer_hint(true)))
+                .await?;
+            if new_len < SNAP_CHUNK_LEN {
```
tabokie, line 144: According to std doc, only `readed==0` guarantees EOF. The read may be interrupted.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,526 @@
...
+                        ),
+                    );
+                    self.pool.spawn(sink.fail(status));
+                    return;
+                }
+                SNAP_TASK_COUNTER_STATIC.recv.inc();
+
+                let snap_mgr = self.snap_mgr.clone();
+                let raft_router = self.raft_router.clone();
+                let recving_count = Arc::clone(&self.recving_count);
```
tabokie, line 391:
```suggestion
                let recving_count = self.recving_count.clone();
```

src/server/tablet_snap.rs
bufferflies: Unlike V1， there are no place to using limit in generate snapshot.

src/server/errors.rs
```diff
@@ -66,6 +66,8 @@ pub enum Error {
 
     #[error("{0:?}")]
     OpenSsl(#[from] OpenSslError),
+    #[error("failed to send")]
```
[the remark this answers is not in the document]
bufferflies: fixed

src/server/tablet_snap.rs
bufferflies: fixed

src/server/tablet_snap.rs
bufferflies: fixed

src/server/tablet_snap.rs
bufferflies: fixed by filter the directory

src/server/tablet_snap.rs
bufferflies: fixed

src/server/tablet_snap.rs
bufferflies: fixed

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,537 @@
...
+            total_sent += new_len as u64;
+            unsafe {
+                buffer.set_len(new_len);
+            }
+            let mut chunk = SnapshotChunk::default();
+            chunk.set_data(buffer);
+            sender
+                .feed((chunk, WriteFlags::default().buffer_hint(true)))
+                .await?;
+            if readed == 0 {
```
tabokie, line 141: Wasteful to always send one empty chunk. Should loop for eof:
```rust
let mut readed = 0;
loop {
  let r = f.read(&mut buffer[off..])?;
  if r == 0 {
    break;
  } else {
    readed += r;
  }
}
```

tonyxuqqi: There's a problem without sending the empty chunk. In fact it's used as a separator when the file can be just sent with full chunk(s). Otherwise the receiver cannot know when to start a new file. @tabokie

bufferflies: how about carry with the file size in the first chunk size?

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,552 @@
...
+            key,
+            raft_msg: meta,
+            io_type,
+            start: Instant::now(),
+            chunk_size,
+        })
+    }
+
+    fn finish<R: RaftStoreRouter<impl KvEngine>>(self, raft_router: R) -> Result<()> {
+        let _with_io_type = WithIoType::new(self.io_type);
```
tabokie, line 96: No need, no IO in this function.

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,552 @@
...
+            if readed == 0 && new_len < SNAP_CHUNK_LEN {
+                break;
+            }
+            limiter.consume(readed);
+            new_len = readed + off;
+            total_sent += new_len as u64;
+            unsafe {
+                buffer.set_len(new_len);
+            }
+            let mut chunk = SnapshotChunk::default();
```
tabokie, line 150: This is still incorrect. When `new_len < SNAP_CHUNK_LEN` but file is not ended yet, the receiver will mistakenly treat this chunk as EOF.

Here's the full version of what I mentioned before:

```rust
for path in files {
  let mut current_chunk_len = buffer.len();
  loop {
    buffer.set_len(chunk_size);
    while current_chunk_len < chunk_size {
      let readed = f.read(&mut buffer)?;
      if readed == 0 { break; }
      total_len += readed;
    }
    /*send chunk*/
    if current_chunk_len < chunk_size {
      break;
    }
    buffer = Vec::new();
    current_chunk_len = 0;
  }
}
```

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,552 @@
...
+    };
+
+    use futures::{
+        channel::mpsc::{self},
+        executor::block_on,
+        sink::SinkExt,
+    };
+    use futures_util::StreamExt;
+    use grpcio::WriteFlags;
+    // use futures_util::SinkExt;
```
tabokie, line 492: rm

tabokie (commented):
Rest LG

src/server/errors.rs
```diff
@@ -66,6 +66,8 @@ pub enum Error {
 
     #[error("{0:?}")]
     OpenSsl(#[from] OpenSslError),
+    #[error("{0:?}")]
```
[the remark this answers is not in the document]
bufferflies: fix

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,552 @@
...
+    }
+}
+
+async fn send_snap_files(
+    mgr: &TabletSnapManager,
+    mut sender: impl Sink<(SnapshotChunk, WriteFlags), Error = Error> + Unpin,
+    msg: RaftMessage,
+    key: TabletSnapKey,
+    limiter: Limiter,
+) -> Result<u64> {
```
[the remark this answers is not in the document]
bufferflies: fixed add func `get_io_type_from_raft_message`

components/raftstore/src/store/snap.rs
```diff
@@ -1955,10 +1956,32 @@ impl TabletSnapManager {
...
+    }
+    pub fn get_tmp_path_for_recv(&self, key: &TabletSnapKey) -> PathBuf {
+        let prefix = format!("{}_{}{}", SNAP_REV_PREFIX, key, TMP_FILE_SUFFIX);
+        PathBuf::from(&self.base).join(prefix)
+    }
+
+    pub fn delete_snapshot(&self, key: &TabletSnapKey) -> bool {
+        let path = self.get_final_path_for_gen(key);
+        if path.exists() && fs::remove_dir_all(path.as_path()).is_err() {
+            warn!(
```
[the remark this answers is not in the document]
bufferflies: fixed

src/server/tablet_snap.rs
bufferflies: fix

src/server/tablet_snap.rs
bufferflies: fixed

components/raftstore/src/store/snap.rs
```diff
@@ -1955,10 +1956,33 @@ impl TabletSnapManager {
...
+        PathBuf::from(&self.base).join(prefix)
+    }
+    pub fn get_tmp_path_for_recv(&self, key: &TabletSnapKey) -> PathBuf {
+        let prefix = format!("{}_{}{}", SNAP_REV_PREFIX, key, TMP_FILE_SUFFIX);
+        PathBuf::from(&self.base).join(prefix)
+    }
+
+    pub fn delete_snapshot(&self, key: &TabletSnapKey) -> bool {
+        let path = self.get_final_path_for_gen(key);
+        if path.exists() && let Err(e)=fs::remove_dir_all(path.as_path()) {
```
tabokie, line 1975:
```suggestion
        if path.exists() && let Err(e) = fs::remove_dir_all(path.as_path()) {
```

bufferflies: fixed

tabokie (approved):

src/server/tablet_snap.rs
```diff
@@ -0,0 +1,557 @@
...
+    let files = fs::read_dir(&path)?
+        .map(|f| Ok(f?.path()))
+        .filter(|f| f.is_ok() && f.as_ref().unwrap().is_file())
+        .collect::<Result<Vec<_>>>()?;
+    let io_type = io_type_from_raft_message(&msg)?;
+    let _with_io_type = WithIoType::new(io_type);
+    let mut total_sent = msg.compute_size() as u64;
+    let mut chunk = SnapshotChunk::default();
+    chunk.set_message(msg);
+    chunk.set_data(usize::to_ne_bytes(SNAP_CHUNK_LEN).to_vec());
```
BusyJay, line 126: It's confusing to set both fields at the same time.

bufferflies: The snapchunk size can be changed in later version. It can keep compatibility.

PR #13784, opened 2022-11-14
mvcc: skip accumulated locks by a second get

sticnarf:

Issue Number: Ref #13694

What's Changed:

```commit-message
Prewrite and acquire_pessimistic_lock use get_write to get the latest PUT or
DELETE record. Point get is implemented by PointGetter. In these cases, with
last_change_ts and versions_to_last_change, now we use an additional get
operation to find the record directly if versions_to_last_change reaches
SEEK_BOUND.

I think no additional metrics are needed because the count of get operation
of the write CF represents represents it.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


sticnarf:
@MyonKeminta Whether do you think `SEEK_BOUND` is a good threshold for skipping? I find you did some research about this bound years ago.

MyonKeminta:
> @MyonKeminta Whether do you think `SEEK_BOUND` is a good threshold for skipping? I find you did some research about this bound years ago.

It makes sense to me to do the `get_cf` directly (which seems to include a `seek` operation (?))  if the accumulated locks/rollbacks is more than `SEEK_BOUND`. `SEEK_BOUND` is expected to stand for how many `next`s' overhead is as large as a `seek`'s. But since that research were done years ago, I'm actually not sure if its value is still suitable nowadays as TiKV and rocksdb are both evolving.

sticnarf:
@MyonKeminta @ekexium PTAL

src/storage/mvcc/reader/reader.rs
```diff
@@ -382,7 +382,31 @@ impl<S: EngineSnapshot> MvccReader<S> {
...
+                            if write.versions_to_last_change < SEEK_BOUND
+                                || write.last_change_ts.is_zero()
+                            {
+                                ts = commit_ts.prev();
+                            } else {
+                                let commit_ts = write.last_change_ts;
+                                let key_with_ts = key.clone().append_ts(commit_ts);
+                                let Some(value) = self
+                                    .snapshot
+                                    .get_cf(CF_WRITE, &key_with_ts)? else {
```
ekexium, line 395: Wow

ekexium (approved):

src/storage/mvcc/reader/reader.rs
MyonKeminta: Wow

MyonKeminta (approved):

PR #13785, opened 2022-11-14
copr: fix _ pattern in like behavior for old collation

YangKeao:

Issue Number: Close https://github.com/tikv/tikv/issues/13769

This PR is a rework for https://github.com/tikv/tikv/pull/13770. As I have force-pushed the branch, github doesn't allow me to reopen the old one :cry: .

What's Changed:

If the two charset of arguments are equal, use the charset from arguments to decode the bytes.

I have considered the following cases:

| Ret Collation  | First Arg Charset | Second Arg Charset | Description |
| ------------- | ------------- |------------- |------------- |
| binary  | binary  | binary | It works normally |
| binary  | utf8mb4  | binary | It happened when TiDB didn't push down the collation information. Just decode and compare the string byte by byte. A `_` in pattern will only match one byte in the first arg |
| binary  | binary  | utf8mb4 | It happened when TiDB didn't push down the collation information. Just decode and compare the string byte by byte. A `_` in pattern will only match one byte in the first arg |
| binary  | utf8mb4  | utf8mb4 | It happened when TiDB didn't push down the collation information. It'll decode and compare the string character by character. A `_` in pattern will match one unicode character in the first arg. This PR is actually fixing this case |
| utf8mb4_*  | utf8mb4  | binary | will not happen |
| utf8mb4_*  | utf8mb4  | utf8mb4 | It works normally |

### Check List 

Tests 

- [x] Unit test
- Integration test
- No code

### Release note 

```release-note
```

### DIFF
--- a/components/tidb_query_expr/src/lib.rs
+++ b/components/tidb_query_expr/src/lib.rs
@@ -44,9 +44,12 @@ pub mod impl_time;
 
 use tidb_query_common::Result;
 use tidb_query_datatype::{
-    codec::{collation::Collator, data_type::*},
-    match_template_charset, match_template_collator, Charset, Collation, FieldTypeAccessor,
-    FieldTypeFlag,
+    codec::{
+        collation::{Charset as _, Collator},
+        data_type::*,
+    },
+    match_template_charset, match_template_collator, match_template_multiple_collators, Charset,
+    Collation, FieldTypeAccessor, FieldTypeFlag,
 };
 use tipb::{Expr, FieldType, ScalarFuncSig};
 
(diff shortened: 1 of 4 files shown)


components/tidb_query_expr/src/lib.rs
```diff
@@ -91,10 +92,51 @@ fn map_compare_in_string_sig(ret_field_type: &FieldType) -> Result<RpnFnMeta> {
...
+    let pattern_collation = children[1]
+        .get_field_type()
+        .as_accessor()
+        .collation()
+        .map_err(tidb_query_datatype::codec::Error::from)?;
+
     Ok(match_template_collator! {
         TT, match ret_field_type.as_accessor().collation().map_err(tidb_query_datatype::codec::Error::from)? {
-            Collation::TT => like_fn_meta::<TT>()
+            Collation::TT =>  match_template_collator!(
```
wjhuang2016, line 109: Can we make it more readable?

YangKeao: I add a new macro to make it into only one layer :thinking: 

(Though I don't know whether it's a good habit to hide the complexity into macro...)

YangKeao: See https://github.com/tikv/tikv/pull/13785/files#diff-2e5b75759b522e863db26feb08e72847c515f0e04e69d5105294931a1d25751aR45

wjhuang2016 (approved):

sticnarf (approved):
LGTM

YangKeao:
/cherry-pick https://github.com/pingcap/tidb/tree/release-6.1

PR #13793, opened 2022-11-15
Raftstore-v2: transfer leader

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

```
Transfer leader is similar to v1.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note
None
```


components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -127,6 +129,18 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER,
         Self { fsm, store_ctx }
     }
 
+    #[inline]
+    fn schedule_pending_ticks(&mut self) {
+        let pending_ticks = self.fsm.peer.take_pending_ticks();
+        for tick in pending_ticks {
+            if tick == PeerTick::ReactivateMemoryLock {
+                self.register_reactivate_memory_lock_tick()
```
tonyxuqqi, line 137: what's the difference of register_reactivate_memory_lock_tick and self.schedule_tick(ReactivateMemoryLock)? if it's same, why bother to differentiate it here?

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -245,3 +262,37 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER,
...
+        {
+            pessimistic_locks.status = LocksStatus::Normal;
+            self.fsm.reactivate_memory_lock_ticks = 0;
+        } else {
+            drop(pessimistic_locks);
+            self.register_reactivate_memory_lock_tick();
+        }
+    }
+
+    pub fn register_reactivate_memory_lock_tick(&mut self) {
```
tonyxuqqi, line 295: no need to have a method if it's just schedule_tick.

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
```diff
@@ -0,0 +1,419 @@
...
+                        // Set this flag to propose this command like a normal proposal.
+                        cmd.mut_header()
+                            .set_flags(WriteBatchFlags::TRANSFER_LEADER_PROPOSAL.bits());
+                        cmd.mut_admin_request()
+                            .set_cmd_type(AdminCmdType::TransferLeader);
+                        cmd.mut_admin_request().mut_transfer_leader().set_peer(from);
+                        if let (PeerMsg::RaftCommand(req), sub) = PeerMsg::raft_command(cmd) {
+                            self.on_admin_command(ctx, req.request, req.ch);
+                        } else {
+                            unreachable!();
```
tonyxuqqi, line 203: i don't think we need unreachable!() here when on_admin_command failed. Just log error instead.

components/raftstore-v2/tests/integrations/test_transfer_leader.rs
```diff
@@ -0,0 +1,146 @@
...
+    metapb,
+    raft_cmdpb::{AdminCmdType, CmdType, Request, TransferLeaderRequest},
+};
+use raft::prelude::ConfChangeType;
+use raftstore_v2::router::PeerMsg;
+use tikv_util::store::new_peer;
+
+use crate::cluster::Cluster;
+
+fn put_data(
```
tonyxuqqi, line 17: do you have test case to verify the lock transfer logic?

components/raftstore-v2/src/fsm/peer.rs
SpadeA-Tang: Good suggestion.

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
SpadeA-Tang: unreachable here means PeerMsg::raft_command should only return PeerMsg::RaftCommand type.

components/raftstore-v2/tests/integrations/test_transfer_leader.rs
SpadeA-Tang: V1's integration test has tests of it. Once we have v2's integration test, we can reuse them.

tonyxuqqi (approved):

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -84,6 +84,7 @@ pub struct Peer<EK: KvEngine, ER: RaftEngine> {
     /// Transaction extensions related to this peer.
     txn_ext: Arc<TxnExt>,
     txn_extra_op: Arc<AtomicCell<TxnExtraOp>>,
+    pending_ticks: Vec<PeerTick>,
```
sticnarf, line 87: Is it more appropriate not to put it under the "transaction extentions" section? Most ticks don't seem to be related to transactions.

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -249,3 +262,33 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER,
         self.fsm.peer.propose_pending_writes(self.store_ctx);
     }
 }
+
+impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER, T> {
```
sticnarf, line 266: Why put it in a separate `impl` block?

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
```diff
@@ -0,0 +1,419 @@
...
+    let req = msg.get_admin_request();
+    if !req.has_transfer_leader() {
+        return None;
+    }
+
+    Some(req.get_transfer_leader())
+}
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    /// Return true to if the transfer leader request is accepted.
```
sticnarf, line 50: The "to" reads weird and should be removed. (I see it's copied from the original comment.)

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
```diff
@@ -0,0 +1,419 @@
...
+                "peer" => ?peer,
+            );
+            return false;
+        }
+
+        // Broadcast heartbeat to make sure followers commit the entries immediately.
+        // It's only necessary to ping the target peer, but ping all for simplicity.
+        self.raft_group_mut().ping();
+
+        let mut msg = eraftpb::Message::new();
```
sticnarf, line 138: Here, the index is missing compared to v1. I think it is worth a TODO comment because the index is important for reducing jitter.

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
```diff
@@ -0,0 +1,419 @@
...
+        // have marked some locks as deleted will fail because raft rejects more
+        // proposals.
+        // It is OK to return true here if it's in other states like MergingRegion or
+        // NotLeader. In those cases, the locks will fail to propose and nothing will
+        // happen.
+        if !pessimistic_locks.is_writable() {
+            return true;
+        }
+        pessimistic_locks.status = LocksStatus::TransferringLeader;
+        self.add_pending_tick(PeerTick::ReactivateMemoryLock);
```
sticnarf, line 338: `reactivate_memory_lock_ticks` is not reset.
And, what's the purpose of adding a pending tick instead of scheduling tick directly?

components/raftstore-v2/src/fsm/peer.rs
SpadeA-Tang: I think the impl above relates to the general logic. Is it better to put them together?

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
SpadeA-Tang: Ok, I see.

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
SpadeA-Tang: Yes, that's the warm up optimization which should be implemented later. I will add todo.

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
SpadeA-Tang: In v1, this method is implemented under struct PeerFsmDelegate, but it is under Peer in v2. It cannot access reactivate_memory_lock_ticks and others directly.

SpadeA-Tang: > reactivate_memory_lock_ticks is not reset.

good catch.

components/raftstore-v2/src/fsm/peer.rs
sticnarf: I don't have any preference. I was just a bit curious about it.

You can add a bit of comments to the impl block.

components/raftstore-v2/src/operation/command/admin/leader_transfer.rs
SpadeA-Tang: Yes, I added it in the `schedule_pending_ticks`

sticnarf (approved):
LGTM

PR #13810, opened 2022-11-16
log-backup: implement subscribe flushing

### LINKED ISSUE #13824: log-backup: implementation of subscribing flushes
## Development Task
We are going to implement the pull model for checkpoints. This need TiKV to support the interface `SubscribeFlushEvents`.

YuJuncen:

Issue Number: Close #13824

What's Changed:
This PR:

```commit-message
Implements the new gRPC interface "SubscribeFlush".
```

Added a new structure `SubscriptionManager` in the checkpoint manager module, this actor receives subscribe requests and flush events. It manages the event routing and the subscription life time management.

### Related changes

### Check List 

Tests 

- Integration test

### Release note 

```release-note
None
```


joccau (approved):

3pointer (approved):

components/backup-stream/src/checkpoint_manager.rs
```diff
@@ -76,8 +159,81 @@ impl CheckpointManager {
...
+        if let Some(mgr) = self.manager_handle.as_mut() {
+            let r = items
+                .map(|(r, ts)| {
+                    let mut f = FlushEvent::new();
+                    f.set_checkpoint(ts.into_inner());
+                    f.set_start_key(r.start_key);
+                    f.set_end_key(r.end_key);
+                    f
+                })
+                .collect::<Box<[_]>>();
```
overvenus, line 225: Why not `collect::<Vec<_>>`?

YuJuncen: Because the slice will never be appended, so we collect it as boxed slice to get rid of the `capacity` field in `Vec`.

(Also note that the message `Emit(Box<[FlushEvent]>)` uses boxed slice too.)

components/backup-stream/src/endpoint.rs
```diff
@@ -165,6 +165,8 @@ where
             ((config.num_threads + 1) / 2).max(1),
         );
         pool.spawn(op_loop);
+        let mut checkpoint_mgr = CheckpointManager::default();
+        pool.spawn(checkpoint_mgr.spawn_subscription_mgr());
```
Leavrth, line 169: is it necessary to wait for `checkpoint_mgr.manager_handle` to be ready, before return the `Endpoint`?

Leavrth (approved):

components/backup-stream/src/endpoint.rs
YuJuncen: The handle would be set upon `spawn_subscription_mgr` has been invoked, and subsequent messages can be buffered in the channel.🤔

components/error_code/src/backup_stream.rs
```diff
@@ -41,12 +41,18 @@ define_error_codes! {
...
         "Error happened when sending raft command.",
-        "This is an internal error, please ask the community for help."
+        "This is an internal error, most of them are happen while initial scanning and can be simply retried."
     ),
     RAFTSTORE => ("RaftStore",
         "Error happened reported from raft store.",
         "This is an internal error, please ask the community for help."
     ),
+    GRPC => (
+        "gRPC",
```
Leavrth, line 51:
```
GRPC => ("gRPC",
```

keep the format the same as others 😆

components/backup-stream/src/endpoint.rs
Leavrth: I see the `checkpoint_mgr.manager_handle` will be set by `channel::tx`, while `checkpoint_mgr.manager_handle` will be got by `as_ref` or `as_mut` in some other places.
I wonder whether `as_ref` and `as_mut` are atomic, like
```
thread1              |    thread2
*i = 1                    |   if *i == 1 {}
```

YuJuncen: Don't worry about that, the `&mut` requirement would make sure there won't be two threads concurrently access this field without lock as long as we are writing safe rust.

Leavrth: get

overvenus (approved):

Issue #13779, opened 2022-11-14
FD: Bi-Directional Replication for ticdc


xiongjiwei:
## Development Task
- [x] a new field `txn_source` in `write_cf` https://github.com/tikv/tikv/pull/13777
- [x] event feed handles the mark `txn_source` https://github.com/tikv/tikv/pull/13796

PR #13817, opened 2022-11-17
txn: make `txn_source` be u64 type

xiongjiwei:

Issue Number: ref https://github.com/tikv/tikv/issues/13779

What's Changed:

we use `u64` to reserve more space for future use. For now, the upper application is limited to setting this value under `0x80`, so there will no more cost to change it to `u64`

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -270,7 +270,7 @@ impl Lock {
             size += 1 + size_of::<u64>() + MAX_VAR_U64_LEN;
         }
         if self.txn_source != 0 {
-            size += 1 + size_of::<u64>() + MAX_VAR_U64_LEN;
+            size += 1 + MAX_VAR_U64_LEN;
         }
         size
     }
--- a/components/txn_types/src/lock.rs
+++ b/components/txn_types/src/lock.rs
@@ -96,7 +96,7 @@ pub struct Lock {
     /// The source of this txn. It is used by ticdc, if the value is 0 ticdc
     /// will sync the kv change event to downstream, if it is not 0, ticdc
     /// may ignore this change event.
-    pub txn_source: u8,
+    pub txn_source: u64,
 }
 
 impl std::fmt::Debug for Lock {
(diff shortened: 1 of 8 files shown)


sticnarf (approved):

xiongjiwei:
ref https://github.com/tikv/client-go/pull/618

src/storage/txn/actions/commit.rs
```diff
@@ -359,7 +359,7 @@ pub mod tests {
 
         let k = b"k";
         // WriteType is Put
-        must_prewrite_put_with_txn_soucre(&mut engine, k, b"v2", k, 25, 1);
+        must_prewrite_put_with_txn_soucre(&mut engine, k, b"v2", k, 25, 0x85);
         let lock = must_locked(&mut engine, k, 25);
         assert_eq!(lock.txn_source, 1);
```
sticnarf, line 364: This line should be changed

xiongjiwei: add a test with two `txn_source` one is less than 0x80 and one is greater

sticnarf:
> /clippy-darwin-arm64

You can just ignore it. It is broken from time to time...

components/txn_types/src/lock.rs
```diff
@@ -266,7 +270,7 @@ impl Lock {
             size += 1 + size_of::<u64>() + MAX_VAR_U64_LEN;
         }
         if self.txn_source != 0 {
-            size += 2;
+            size += 1 + size_of::<u64>() + MAX_VAR_U64_LEN;
```
MyonKeminta, line 273: Why is it sizeof u64 + MAX_VAR_U64_LEN ? Is there a mistake in `last_change_ts`? 😕 @sticnarf

sticnarf: Ah, here it should be `1 + MAX_VAR_U64_LEN`

sticnarf: The calculation in `last_change_ts` is because there are two integers (a fixed-length u64 and a var-length u64) after the flag.

MyonKeminta (approved):

PR #13818, opened 2022-11-17
raftstore-v2: implement region merge

tabokie:

Issue Number: Ref #12842

What's Changed:

```commit-message
Implement region merge for v2.

Key differences with v1:
- Previously prepare merge write region state to kvdb during apply, and commit
merge validate this state during apply. Now the state is written async after
applying prepare merge, and commit merge doesn't validate it any more.
- Target apply fsm uses async channel to wait for source apply fsm to be ready.
Read the flow chart in commit_merge.rs for more details.
- Tablets are both flushed, and merged to a temporary directory, then renamed
to the final path. Currently the flush is a blocking call. Its overhead should
be the same with checkpoint. The reason of not using checkpoint is to avoid
opening multiple temporary DBs.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore-v2/src/raft/peer.rs
```diff
@@ -39,8 +45,8 @@ use crate::{
     operation::{AsyncWriter, DestroyProgress, ProposalControl, SimpleWriteEncoder},
     router::{CmdResChannel, QueryResChannel},
     tablet::CachedTablet,
-    worker::PdTask,
     Result,
+    StoreMeta,
```
tabokie, line 49:
```suggestion
    Result, StoreMeta,
```

components/raftstore-v2/src/operation/command/admin/commit_merge.rs
```diff
@@ -0,0 +1,927 @@
...
+        }
+
+        let source_tablet: EK = match source_ready.tablet.downcast() {
+            Ok(t) => *t,
+            Err(t) => unreachable!("tablet type should be the same: {:?}", t),
+        };
+        self.tablet().flush_cfs(&[], true).unwrap();
+        // TODO: check both are trimmed.
+        let reg = self.tablet_registry();
+        let tmp_path = reg.tablet_path(self.region_id(), index);
```
BusyJay, line 508: It's not a temp path.

components/raftstore-v2/src/operation/command/admin/commit_merge.rs
```diff
@@ -0,0 +1,927 @@
...
+        if tmp_path.exists() {
+            std::fs::remove_dir_all(&tmp_path).unwrap();
+        }
+        let mut ctx = TabletContext::new(&region, Some(index));
+        {
+            let tablet = reg
+                .tablet_factory()
+                .open_tablet(ctx.clone(), &tmp_path)
+                .unwrap();
+            tablet.merge(&[&source_tablet, self.tablet()]).unwrap();
```
BusyJay, line 518: Can merge be done in checkpoint? It may be a lot safer as only manifest needs to be changed.

components/raftstore-v2/src/operation/command/admin/commit_merge.rs
```diff
@@ -0,0 +1,927 @@
...
+            region.get_region_epoch().get_version(),
+        ) + 1;
+        region.mut_region_epoch().set_version(version);
+        if keys::enc_end_key(&region) == keys::enc_start_key(source_region) {
+            region.set_end_key(source_region.get_end_key().to_vec());
+        } else {
+            region.set_start_key(source_region.get_start_key().to_vec());
+        }
+
+        let source_tablet: EK = match source_ready.tablet.downcast() {
```
BusyJay, line 501: I was thinking about making source peer create a checkpoint instead of passing db handles around. And unlike v1, v2 can skip calling source peer if checkpoint exists. Because there is no race at all in v2.

If we are going to introduce multi-range raft later, this implementation can also be reused.

components/raftstore-v2/src/operation/command/admin/commit_merge.rs
tabokie: Checkpoint has limited functionality. It only copies file without altering the content (except for the CURRENT which is a plaintext marker file). It also cannot resolve conflict of file number between instances.

BusyJay: What I mean is create a checkpoint and then modify it.

components/raftstore-v2/src/operation/command/admin/commit_merge.rs
tabokie: The purpose is to reduce number of db opening. Right now, it only opens once (after rename to final path), to introduce a checkpoint, it will open twice. I don't know how long it takes for split to open a checkpoint right now. With that information we can decide if it's worth optimizing.

BusyJay: Why it needs to be twice? For example, we can merge the two manifest from both checkpoints, rename it and then open.

tabokie: Manipulating manifest without opening a DB is unthinkable. The code to recover files and modify manifest are all coupled with DB. A better choice would be to avoid loading table footer for intermediate DBs.

components/raftstore-v2/src/operation/command/admin/rollback_merge.rs
```diff
@@ -0,0 +1,213 @@
...
+                    );
+                }
+            }
+        }
+    }
+}
+
+impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
+    // Match v1::exec_rollback_merge.
+    pub fn apply_rollback_merge(
```
BusyJay, line 110: Better leave rollback to other PR.

tabokie:
Tablet trim check is too complex. I moved it to a separate branch: [tikv](https://github.com/tabokie/tikv/commit/2d352758e23b687b0271358c2ec91981883db124), [kvproto](https://github.com/tabokie/kvproto/commit/64142180c96bbb32fbbaf4fd2499663cd40d0a24)

BusyJay:
I suggest you to split the PR into three PRs:
1. prepare merge, including prerequisites check, proposing and applying the command, and handle the result.
2. commit merge, including scheduling, proposing and applying the command, and handle the result.
3. rollback merge, including proposing and applying the command, and handle the result.

PR #13820, opened 2022-11-17
storage: skip accumulated locks on non-existing keys

sticnarf:

Issue Number: Ref #13694

What's Changed:

```commit-message
For non-existing keys, we will record last_change_ts as 0 and set a non-zero
versions_to_last_change. So, if we encounter such a write record when reading,
we can quickly know this key does not exist.
```

### Check List 

Tests 

- TODO

### Release note 

```release-note
None
```


ekexium (approved):

sticnarf:
PTAL @MyonKeminta

MyonKeminta (approved):

Issue #13787, opened 2022-11-14
Make Flashback compatible with CDC


JmPotato:
Ref #13303.

Since the flashback feature bypasses the standard transaction writing progress, so it basically has two incompatible critical points with the CDC.

1. The Percolator transaction rules are not strictly followed, so CDC will have problems capturing data changes generated by the flashback process.
2. The key cleaning work makes the Resolved TS advance unexpectedly while the flashback data changes are still pushed to the CDC.

To make flashback compatible with CDC, there are several places we need to modify.

- Flashback transaction modification results are basically the same as a 1PC request. So we can make it become or look like a 1PC transaction to make CDC handle it naturally without any other development.
- Same as the 1PC, we need to maintain the memory locks in `ConcurrencyManager` correctly to make sure the Resolved TS won't advance during the flashback progress.

By fixing these two problems, the compatibility issue between Flashback and CDC should be addressed without too many changes.

HuSharp:
if the memory lock is destroyed, consideration must be given to preventing resolved ts.

PR #13823, opened 2022-11-18
server, storage: make flashback compatible with resolved_ts

JmPotato:

Issue Number: ref #13787.

What's Changed:

```commit-message
- Write a lock into `lock_cf` to prevent `resolved_ts` from advancing during the flashback process.
- Move `FlashbackToVersionState::ScanWrite` to the front of `FlashbackToVersionState::ScanLock`.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```


JmPotato:
/cc @overvenus @Connor1996 @sticnarf 

Please help review this PR, thx a lot.

src/server/service/kv.rs
```diff
@@ -1749,7 +1750,23 @@ fn future_prepare_flashback_to_version<
             AdminCmdType::PrepareFlashback,
         )
         .await?;
-        Ok(PrepareFlashbackToVersionResponse::default())
+        // After the region is put into the flashback state, we need to write a lock to
+        // prevent `resolved_ts` from advancing.
+        let (cb, f) = paired_future_callback();
+        let res = storage_clone.sched_txn_command(req.clone().into(), cb);
```
sticnarf, line 1756: Cloning `req` is unnecessary

src/storage/txn/actions/check_txn_status.rs
```diff
@@ -153,10 +160,13 @@ pub fn rollback_lock(
     lock: &Lock,
     is_pessimistic_txn: bool,
     collapse_rollback: bool,
+    for_flashback: bool,
 ) -> Result<Option<ReleasedLock>> {
     let overlapped_write = match reader.get_txn_commit_record(&key)? {
         TxnCommitRecord::None { overlapped_write } => overlapped_write,
-        TxnCommitRecord::SingleRecord { write, .. } if write.write_type != WriteType::Rollback => {
+        TxnCommitRecord::SingleRecord { write, .. }
+            if write.write_type != WriteType::Rollback && !for_flashback =>
```
sticnarf, line 168: Why is it possible for flashback to roll back some lock with a committed record?

JmPotato: Here is the case:

- Lock `self.start_key` with flashback's `start_ts`.
- Flashback `self.start_key` to write a new record with flashback's `commit_ts`.
- Rolling back the lock `self.start_key` will see the newly written flashback committed record.

sticnarf: It breaks how transaction normally works so it makes me feel a bit worried.

Can you ensure `start_key` is not a real key, so there can't be a new record written to the key?

JmPotato: Okay, let me have a try on this way.

sticnarf: Or, I think you can make the `start_key` a bit special. Other keys are flashbacked in a 1PC way, except for this key with a lock, which is committed like a normal 2PC. It's committed until the very end of the process.

JmPotato: Done. PTAL.

src/storage/txn/commands/mod.rs
```diff
@@ -356,6 +356,16 @@ impl From<MvccGetByStartTsRequest> for TypedCommand<Option<(Key, MvccInfo)>> {
...
+            Key::from_raw(req.get_start_key()),
+            req.get_start_ts().into(),
+            req.take_context(),
+        )
+    }
+}
+
 impl From<FlashbackToVersionRequest> for TypedCommand<()> {
     fn from(mut req: FlashbackToVersionRequest) -> Self {
         new_flashback_to_version_read_phase_cmd(
```
HuSharp, line 371: How about changing this name？

src/storage/txn/commands/flashback_to_version.rs
```diff
@@ -72,19 +95,29 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for FlashbackToVersion {
...
-                if let Some(new_next_lock_key) =
-                    flashback_to_version_lock(&mut txn, &mut reader, mem::take(key_locks))?
-                {
-                    *next_lock_key = new_next_lock_key;
-                }
+            FlashbackToVersionState::Prewrite { ref key_to_lock } => {
+                txn.put_lock(
+                    key_to_lock.clone(),
+                    &Lock::new(
+                        LockType::Put,
```
HuSharp, line 109: Need to match this `start_key` flashbacked way.

src/server/service/kv.rs
```diff
@@ -1793,12 +1811,7 @@ fn future_flashback_to_version<
             // in `RegionLocalState` and region's meta, and when that
             // admin cmd is applied, will update the memory
             // state of the flashback
-            send_flashback_msg::<T, E>(
-                &raft_router,
-                req.get_context(),
-                AdminCmdType::FinishFlashback,
-            )
-            .await?;
+            send_flashback_msg::<T, E>(&raft_router, &ctx, AdminCmdType::FinishFlashback).await?;
```
HuSharp, line 1814: If tikv is down after the 2pc lock is released, but the completion flashback has not been sent yet.
At this point tidb will retry the flashback, but there is a window period when it will advance the CDC.

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -76,80 +82,86 @@ impl CommandExt for FlashbackToVersionReadPhase {
...
+///       `resolved_ts` from advancing.
+///   2. Read-and-flashback writes phase:
+///     - Scan all the latest writes and their corresponding values at
+///       `self.version`.
+///     - Write the old MVCC version writes again for all these keys with
+///       `self.commit_ts` excluding the `self.start_key` .
+///   3. Read-and-rollback locks phase:
+///     - Scan all locks.
+///     - Rollback all these locks excluding the `self.start_key` lock we write
+///       at the first phase.
```
sticnarf, line 97: I notice you swap the order of FlashbackLock and FlashbackWrite. Is there a reason for that?

Is it possible that some lock exists while you are writing the new flashbacked records? It looks a bit weird... And I just start wondering what will happen if there is a lock on the `start_key` before flashback begins...

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -134,11 +137,82 @@ pub fn flashback_to_version_write(
...
+
+pub fn commit_flashback_key(
+    txn: &mut MvccTxn,
+    reader: &mut SnapshotReader<impl Snapshot>,
+    key_to_commit: &Key,
+    start_ts: TimeStamp,
+    commit_ts: TimeStamp,
+) -> TxnResult<()> {
+    // If the key is not locked, it means that the key has been committed before and
+    // we are in a retry.
```
HuSharp, line 196: We can place the `check lock` at the beginning of the read phase to avoid redundant RW.

src/storage/txn/commands/flashback_to_version_read_phase.rs
JmPotato: > And I just start wondering what will happen if there is a lock on the start_key before flashback begins...

Since all the other read/write transactions will be blocked and fail during the flashback, I think this kind of lock doesn't have any side effects. 🤔

sticnarf: I'm just not sure how CDC will react if a lock is rewritten with a different start ts...

src/storage/txn/commands/flashback_to_version.rs
```diff
@@ -72,19 +99,23 @@ impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for FlashbackToVersion {
...
-            FlashbackToVersionState::ScanLock {
-                ref mut next_lock_key,
-                ref mut key_locks,
-            } => {
-                if let Some(new_next_lock_key) =
-                    flashback_to_version_lock(&mut txn, &mut reader, mem::take(key_locks))?
-                {
-                    *next_lock_key = new_next_lock_key;
-                }
+            FlashbackToVersionState::Prewrite { ref key_to_lock } => {
```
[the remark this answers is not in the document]
JmPotato: 😂Actually, this is added by the formatter automatically.

sticnarf (commented):
LGTM

JmPotato:
@overvenus @Connor1996 PTAL also, thx.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -134,11 +135,80 @@ pub fn flashback_to_version_write(
...
+        txn.put_write(
+            key_to_commit.clone(),
+            commit_ts,
+            Write::new(
+                WriteType::from_lock_type(lock.lock_type).unwrap(),
+                start_ts,
+                lock.short_value.take(),
+            )
+            .set_last_change(lock.last_change_ts, lock.versions_to_last_change)
+            .set_txn_source(lock.txn_source)
```
overvenus, line 203: Maybe we don't need to the `txn_source` in the sense of TiCDC bidirectional replication. What do you think? @xiongjiwei

overvenus (commented):
Please ignore

src/storage/txn/actions/flashback_to_version.rs
xiongjiwei: No, we do not need it. and actually `txn_source` is `0` here

components/test_raftstore/src/util.rs
```diff
@@ -1210,6 +1210,10 @@ pub fn must_flashback_to_version(
 ) {
     let mut prepare_req = PrepareFlashbackToVersionRequest::default();
     prepare_req.set_context(ctx.clone());
+    prepare_req.set_start_ts(start_ts);
+    prepare_req.version = version;
```
Connor1996, line 1214: mixed style

overvenus (approved):

Connor1996 (approved):
LGTM

src/storage/mvcc/reader/reader.rs
```diff
@@ -95,6 +95,11 @@ impl<S: EngineSnapshot> SnapshotReader<S> {
         self.reader.seek_write(key, ts)
     }
 
+    #[inline(always)]
+    pub fn get_value(&mut self, key: &Key, start_ts: TimeStamp) -> Result<Option<Value>> {
```
sticnarf, line 99: `SnapshotReader` typically uses its own `start_ts` for something. If it doesn't need the `start_ts` of `SnapshotReader` itself, I suggest to just use the inner `MvccReader` directly.

JmPotato: Removed. PTAL.

BTW, your suggestion made me realize that there is a mix of `SnapshotReader` and `MvccReader` in the flashback commands, and I think we can open another separate PR to improve this situation, after this PR is merged, of course.

HuSharp: address this comment at https://github.com/tikv/tikv/pull/13860 @sticnarf @JmPotato

Issue #9113, opened 2020-11-25
More fine-grained control on callback for raftstore


BusyJay:
## Bug Report

We currently rely on raft to handle callback of requests: a callback will be invoked when the corresponding entry is applied/committed or overwritten.  However, raft doesn't guarantee when will an entry be committed or overwritten, so the time when a callback is invoked is unpredictable. If a leader loses its leadership without rejoining the group or it becomes leader but takes a long time to commit its first log, any accepted requests will stuck. For the transaction layer, pending requests will occupy hash slots and lead to ServerIsBusy eventually due to conflict collisions.

I propose to add timeout check on callbacks. If a command is timeout, it should be invoked early using a `Timeout` error. Client should retry in this situation. When raftstore decides to return `Timeout` error, it should not serve any Read requests until it has applied to the `last_index`.

As a simple and quick implementation that just solves the problem of slow election, if a peer becomes follower from candidate, it can clear all the pending commands. Change from candidate to follower means it fails to campaign. If it was leader, its uncommitted logs may be overwritten by new leader.

youjiali1995:
@MyonKeminta @sticnarf PTAL. It adds another undetermined error to transaction commit.

cfzjywxk:
It's reasonable to return `timeout` or `undetermined ` error. Maybe processing of different comman types for `timeout` error should be considered on the client side.

sticnarf:
> If a leader loses its leadership without rejoining the group or it becomes leader but takes a long time to commit its first log, any accepted requests will stuck. For the transaction layer, pending requests will occupy hash slots and lead to ServerIsBusy eventually due to either real conflict or hash collisions.

What could be the cause of " lose its leadership without rejoining the group or it becomes leader but takes a long time to commit its first log"? If it's due to a persistent network error, it's unlikely that it could receive more requests until `ServerIsBusy`.

BusyJay:
Yes. Or a replica is removed without acknowledgement, or one follower is down and another follower is waiting for snapshot.

MyonKeminta:
I have a question. Currently we now release latch in scheduler by letting raftstore call a callback on apply finish. Consider this case:

1. Request A tries to write k1, acquired the latch, and waiting for apply finish
2. Request A timed out.
3. Request B acquires tries to write k1 again

What should we do in this case?

2.a Request A responds timeout error to the client, but still waiting apply finishing. This seems didn't solve the problem in this issue.

2.b Request A abandoned the callback and released the latch.
[...]

PR #13826, opened 2022-11-21
scheduler: Support resuming after meeting lock for AcquirePessimisticLock requests

MyonKeminta:

Issue Number: Ref #13298

Requires https://github.com/tikv/tikv/pull/13833

What's Changed:

```commit-message
Update scheduler's logic to support resuming a woken-up AcquirePessimisticLock request. It also includes changes to Latches to support deriving latches after a command released some locks (instead of releasing the latches and acquire them again).
```

### Related changes

-

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU

### Release note 

```release-note
None.
```


MyonKeminta:
Most of the changes in this PR is tests. It takes too much loc to write tests in the storage layer...
Should I move the test code to a separated PR?

src/storage/txn/scheduler.rs
```diff
@@ -880,9 +1031,15 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
             .unwrap();
     }
 
+    fn is_undetermined_error(_e: &tikv_kv::Error) -> bool {
+        // TODO: If there's some cases that `engine.async_write` returns error but it's
```
sticnarf, line 1035: If this happened, it should be a bug? Because it will affect correctness for other requests.

MyonKeminta: I'm afraid this problem already exists, and there's possible to be more such cases in the future.

* https://github.com/pingcap/tidb/issues/21355
* https://github.com/tikv/tikv/issues/9113
* https://github.com/tikv/tikv/pull/12369#discussion_r852017654

MyonKeminta: I just left an empty function here since I think we may need to handle these possibilities in the future.

MyonKeminta:
@sticnarf @cfzjywxk PTAL when you have time, thanks!

src/storage/txn/scheduler.rs
```diff
@@ -1076,32 +1233,59 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
+                // immediately.
+                if lock_info.parameters.wait_timeout.is_some() {
+                    assert_eq!(to_be_write.size(), 0);
+                    pr = Some(ProcessResult::Res);
+
+                    scheduler.on_wait_for_lock(&ctx, cid, lock_info, tracker);
+                } else {
+                    // For requests with `allow_lock_with_conflict`, key errors are set key-wise.
+                    // TODO: It's better to return this error from
+                    // `commands::AcquirePessimisticLocks::process_write`.
```
sticnarf, line 1251: So.., if we are setting locks in `pr` here now,  what's the meaning of setting a `PessimisticLockKeyResult::Failed` to `pr` in acquire_pessimistic_locks?

sticnarf (approved):
It's definitely not pleasant to go through the PR due to the complexity....

So, the final missing part is fixing the deadlock detection?

src/storage/txn/scheduler.rs
MyonKeminta: Though we support only single-key requests for now, the protocol is at first designed to support processing multiple keys in one request. In multiple keys cases, it's unavoidable that we may fail to lock some of the keys while the other keys is successfully locked. To allow patially failing, this type is added so that we can see in the resp.results that which key is failed.

The remaining missing parts are:

* deadlock detection
* handling the case that timeout when the request is resumed and being executing, but still meet lock (in current implementation, waiter manager will simply drop the waiter since the entry can't be found in the lock waiting queue; if it's added to the queue again, it will then never timeout).

cfzjywxk (approved):
We may need quick verification of the performance impact on the normal path.

src/storage/txn/scheduler.rs
cfzjywxk: Processing a single key is already difficult enough for us.. :sweat_smile:

Issue #13827, opened 2022-11-21
make tikv server not interact with raftstore directly


BusyJay:
## Development Task

We have hidden different storage by `tikv_kv::Engine`, but some of code in server module still access raft router directly, which makes it quite hard to develop a new `Engine` like raftstore v2. We should removed all direct reference of raft router from server module. Tasks include following:

- [ ] kv service
    - [ ] flashback
    - [ ] read index
    - [ ] raft and batch raft
    - [ ] snapshot
    - [ ] green gc
- [ ] debug service
- [ ] gc worker
- [ ] status server
- [ ] resolver
- [ ] sst importer

This is part of #12842.

PR #13828, opened 2022-11-21
*: make flashback not aware of raftstore router

BusyJay:
Issue Number: Ref #13827

What's Changed:

```commit-message
flashback is a transaction concept, better make it only interact with
the storage layer instead of raftstore directly.

This PR also converts raftstore errors to region errors for flashback,
so it will can make client retry more reliable.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/src/server/raftkv.rs
+++ b/src/server/raftkv.rs
@@ -15,10 +15,14 @@ use std::{
 use collections::{HashMap, HashSet};
 use concurrency_manager::ConcurrencyManager;
 use engine_traits::{CfName, KvEngine, MvccProperties, Snapshot};
+use futures::future::BoxFuture;
 use kvproto::{
     errorpb,
     kvrpcpb::{Context, IsolationLevel},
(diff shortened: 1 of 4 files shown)


src/server/raftkv.rs
```diff
@@ -149,6 +153,49 @@ where
...
+
+fn exec_admin<E: KvEngine, S: RaftStoreRouter<E>>(
+    router: &S,
+    req: RaftCmdRequest,
+) -> BoxFuture<'static, kv::Result<()>> {
+    let (cb, f) = paired_future_callback();
+    let res = router.send_command(
+        req,
+        raftstore::store::Callback::write(cb),
+        RaftCmdExtraOpts::default(),
```
BusyJay, line 189: I make it not pass disk full by default. Flashback is also a write, it's quite dangerous to allow it execute when disk is almost full.

Connor1996 (approved):
LGTM

sticnarf (approved):
LGTM

PR #13831, opened 2022-11-22
engine: adapt engine metrics for multi-rocks

tabokie:

Issue Number: Ref #12842 

What's Changed:

- Share one statistics for different tablets
- Iterate all tablets when flushing metrics
- Put private metrics into a histogram

```commit-message
```

### Related changes

tikv/rust-rocksdb#725

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


tabokie:
For single-rocksdb:
single
屏幕截图 2022-12-12 115628

For multi-rocksdb:
multi-7
image

components/server/src/server.rs
```diff
@@ -1953,25 +1954,39 @@ impl<T: fmt::Display + Send + 'static> Stop for LazyWorker<T> {
...
-        self.engines.raft.flush_metrics("raft");
-        if now.saturating_duration_since(self.last_reset) >= DEFAULT_ENGINE_METRICS_RESET_INTERVAL {
-            KvEngine::reset_statistics(&self.engines.kv);
-            self.engines.raft.reset_statistics();
+        let should_reset =
+            now.saturating_duration_since(self.last_reset) >= DEFAULT_ENGINE_METRICS_RESET_INTERVAL;
+        let mut is_first_instance = true;
+        self.tablet_registry
+            .for_each_opened_tablet(&mut |_, tablet: &mut CachedTablet<EK>| {
+                if let Some(tablet) = tablet.latest() {
```
BusyJay, line 1978: If you can make the method defined on statistics, then you don't need to care whether it's shared, just `registry.statistics.flush(); for_each(|t| t.flush_other());`.

src/config/mod.rs
```diff
@@ -1452,7 +1452,6 @@ impl RaftDbConfig {
         opts.set_max_manifest_file_size(self.max_manifest_file_size.0);
         opts.create_if_missing(self.create_if_missing);
         opts.set_max_open_files(self.max_open_files);
-        opts.enable_statistics(self.enable_statistics);
```
BusyJay, line 1455: There are some place don't use `engine_factory` at all, and those cases will miss statistics.

src/server/engine_factory_v2.rs
```diff
@@ -0,0 +1,516 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: Why create the file again?

metrics/grafana/tikv_trouble_shooting.json
```diff
@@ -7222,7 +7222,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "tikv_engine_pending_compaction_bytes{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", db=\"$db\"}",
+              "expr": "10 * sum(rate(tikv_engine_pending_compaction_bytes_sum{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", db=\"$db\"}[1m]))",
```
BusyJay, line 7225: Isn't it a guage? Why sum up with an interval?

tabokie: It's a histogram now.

components/engine_rocks/src/rocks_metrics.rs
```diff
@@ -1115,11 +1115,13 @@ pub fn flush_engine_properties(engine: &DB, name: &str) {
 // For property metrics
 #[rustfmt::skip]
 lazy_static! {
-    pub static ref STORE_ENGINE_SIZE_GAUGE_VEC: IntGaugeVec = register_int_gauge_vec!(
+    pub static ref STORE_ENGINE_SIZE_HISTOGRAM_VEC: HistogramVec = register_histogram_vec!(
         "tikv_engine_size_bytes",
         "Sizes of each column families",
-        &["db", "type"]
+        &["db", "type"],
+        exponential_buckets(1048576.0, 2.0, 20).unwrap()
```
BusyJay, line 1122: So if there are 1024 DBs, and the base size will be 1048576 * 1024 * 3 = 3G?

metrics/grafana/tikv_trouble_shooting.json
BusyJay: Why the change? If it's reported twice within 1m or lose some of the metrics within 1m, the metrics will become either too large or too small? Is it even a compatible change?

metrics/grafana/tikv_details.json
```diff
@@ -30316,7 +30340,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "tikv_engine_num_snapshots{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", db=\"$db\"}",
+              "expr": "10 * rate(tikv_engine_num_snapshots_sum{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", db=\"$db\"}[1m])",
```
Connor1996, line 30343: why plus 10

metrics/grafana/tikv_trouble_shooting.json
tabokie: Because it's the easiest way to report the statistics with some distribution data. Yes the actual reporting will be affected by the reporting frequency (e.g. 10s per reporting so it's needed to scale the counter by 10).

components/engine_rocks/src/rocks_metrics.rs
tabokie: I don't understand the calculation. Why is histogram bucket size related to sample number?

metrics/grafana/tikv_details.json
tabokie: Because we report the statistics each 10s. So the rate needs to be adjusted x10.

components/engine_rocks/src/rocks_metrics.rs
BusyJay: Never mind, I thought a tablet was a db.

metrics/grafana/tikv_trouble_shooting.json
BusyJay: But it's inaccurate. The correct way to do this is collect all the metrics in a temporary struct and publish them at once. So no definitions need to be adjusted.

```
let mut stats = CfStats::default();
flush_engine_properties(engine, name, &mut stats);
publish(engine, stats);

fn flush_engine_properties(engine, name, stats) {
    for (i, cf) in engine.cf_names().enumerate() {
        if let Some(v) = engine.get_property(cf, prop_name) {
            stats[i] += v;
        }
    }
}

fn publish(engine, name, stats) {
    for (i, cf) in engine.cf_names().enumerate() {
        REAL_METRICS.with_label_value(&name, cf).set(stats[i])
    }
}
```

PR #13832, opened 2022-11-22
*: remove legacy code

### LINKED ISSUE #13827: make tikv server not interact with raftstore directly

BusyJay:
Issue Number: Close #13827

What's Changed:

```commit-message
green gc is developed for gc without waking hibernated peers. But
since dynamic regions, it's not compatible with physical isolation.
And it's never used in production due to correctness concern.

Read index RPC is used by tiflash in the past. But now tiflash is
using internal message forward instead of explicit RPC. And I'm not
aware of any other project is utilizing the API.
```

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM
    Existing APIs will return unimplemented status code. But since they are not used, it should be OK.

### Release note 

```release-note
None
```


CalvinNeo (approved):
lgtm

MyonKeminta (approved):
LGTM. We also need to disable the related code from TiDB to avoid misusing. Maybe it should be done before merging this PR.

BusyJay:
> We also need to disable the related code from TiDB to avoid misusing.

@MyonKeminta can you help me with that part?

If these APIs are accessed by TiDB, gRPC unimplemented status code will be returned.

src/server/service/kv.rs
```diff
@@ -976,103 +812,6 @@ impl<T: RaftStoreRouter<E::Local> + 'static, E: Engine, L: LockManager, F: KvFor
         ctx.spawn(task);
     }
 
-    fn read_index(
```
MyonKeminta, line 979: Ah, wait, won't this be called by other TiKV nodes?

BusyJay: It's also explained in the PR description. It's not used as far as I know.

MyonKeminta:
> > We also need to disable the related code from TiDB to avoid misusing.
> 
> @MyonKeminta can you help me with that part?
> 
> If these APIs are accessed by TiDB, gRPC unimplemented status code will be returned.

Ok, but I may have time to do that later this week.

BusyJay:
> Ok, but I may have time to do that later this week.

No hurry, before next code frozen day is OK.

sticnarf (approved):

PR #13833, opened 2022-11-22
scheduler: Support keeping some latches for another command when releasing latches.

MyonKeminta:

Issue Number: Ref #13298

Part of #13826

What's Changed:

```commit-message
Makes latches support keeping some of the latch slots when releasing, and these kept latch slots can be derived by another scheduler command. 
This is necessary for supporting resuming pessimistic lock requests when releasing pessimistic locks. It provides a mechanism to avoid the latch from being acquired by other later-arriving requests at the time between the lock-releasing's end and the resumed pessimistic lock command's beginning.
```

### Related changes

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
  - None

### Release note 

```release-note
None
```


sticnarf (approved):
LGTM

cfzjywxk (approved):

PR #13834, opened 2022-11-22
txn: revert "calculate last_change_ts in rollback"

sticnarf:

Issue Number: Ref https://github.com/tikv/tikv/issues/13694

What's Changed:

```commit-message
This commit reverts #13749 (calculate last_change_ts in rollback).

If we calculate last_change_ts, consider the following case:
Key k has a write record with commit_ts = 5.
1. Prewrite k, start_ts = 10
2. Rollback k, start_ts = 30, last_commit_ts = 5.
3. Commit k, start_ts = 10, commit_ts = 20

Then, read with ts = 40, it will get an incorrect last_commit_ts from
the rollback record.

There is no easy way to handle the rollback case. I choose to give up
calculating it.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/mvcc/reader/reader.rs
+++ b/src/storage/mvcc/reader/reader.rs
@@ -2541,7 +2541,6 @@ pub mod tests {
         engine.commit(k, 1, 2);
 
         // Write enough ROLLBACK/LOCK recrods
-        engine.rollback(k, 5);
         for start_ts in (6..30).into_iter().step_by(2) {
             engine.lock(k, start_ts, start_ts + 1);
         }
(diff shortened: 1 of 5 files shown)


MyonKeminta (approved):
LGTM. Sorry for not noticing this case when reviewing...
Btw, is this problem found in any tests?

sticnarf:
> Sorry for not noticing this case when reviewing...
> Btw, is this problem found in any tests?

I should have considered this case...

It's reported by some other unit test after adding the optimization in #13820...

ekexium (approved):

PR #13836, opened 2022-11-22
tikv_kv: make async_snapshot return future

BusyJay:
Issue Number: Ref #13827.

What's Changed:
```commit-message
so we can reduce allocation and keep compatible with both v1 and v2.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/tikv_kv/src/lib.rs
+++ b/components/tikv_kv/src/lib.rs
@@ -6,6 +6,7 @@
 //! [`RocksEngine`](RocksEngine) are used for testing only.
 
 #![feature(min_specialization)]
+#![feature(type_alias_impl_trait)]
 
 #[macro_use(fail_point)]
 extern crate fail;
@@ -35,7 +36,7 @@ use engine_traits::{
     CF_DEFAULT, CF_LOCK,
 };
 use error_code::{self, ErrorCode, ErrorCodeExt};
-use futures::{future::BoxFuture, prelude::*};
+use futures::{compat::Future01CompatExt, future::BoxFuture, prelude::*};
 use into_other::IntoOther;
 use kvproto::{
     errorpb::Error as ErrorHeader,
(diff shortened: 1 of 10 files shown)


BusyJay:
Microbench shows that when both using `tikv_kv::snapshot`, the worst case (always return error) time cost is reduced from 960ns to 920ns. But if using `raftkv::async_snapshot` directly, the worst case time cost is increased from 920ns to 818ns. The difference due to the new implementation always use `MustCall` as its only callback, which is more costy than naive callback.

components/tikv_kv/src/lib.rs
```diff
@@ -302,17 +304,21 @@ pub trait Engine: Send + Clone + 'static {
...
-        wait_op!(|cb| self.async_snapshot(ctx, cb), timeout)
-            .unwrap_or_else(|| Err(Error::from(ErrorInner::Timeout(timeout))))
+        let deadline = Instant::now() + DEFAULT_TIMEOUT;
+        let timeout = GLOBAL_TIMER_HANDLE.delay(deadline).compat();
+        futures::executor::block_on(async move {
+            futures::select! {
+                res = self.async_snapshot(ctx).fuse() => res,
+                _ = timeout.fuse() => Err(Error::from(ErrorInner::Timeout(DEFAULT_TIMEOUT))),
+            }
+        })
```
sticnarf, line 321: Can we use `GLOBAL_TIMER_HANDLE.timeout()`? That looks simpler.

BusyJay: `timeout` is v1, so the future will be converted twice.

sticnarf: After creating the timeout, we can just call `wait` on the v1 future.

src/server/raftkv.rs
```diff
@@ -481,10 +458,46 @@ where
         }
         ASYNC_REQUESTS_COUNTER_VEC.snapshot.all.inc();
         let begin_instant = Instant::now_coarse();
-        self.exec_snapshot(
-            ctx,
-            req,
-            Box::new(move |res| match res {
+        let (cb, f) = paired_must_called_future_callback(drop_snapshot_callback);
```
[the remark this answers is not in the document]
BusyJay: I prefer to keep the v1 code as it is.

Connor1996 (approved):
LGTM

components/tikv_kv/src/lib.rs
BusyJay: Then it will require `SnapshotRes` to be `unpin`.

sticnarf (approved):

Issue #13800, opened 2022-11-15
Improve the flashback performance


JmPotato:
Ref #13303.

As we implemented the flashback feature by scanning the old writes and writing them again with the latest `commit_ts`, there is still a need to improve its performance by all kinds of means. So that we can achieve an acceptable performance under any software and hardware conditions.

PR #13842, opened 2022-11-24
storage: introduce hint_min_ts to speed up the flashback progress

JmPotato:

Issue Number: ref #13800.

What's Changed:

```commit-message
Introduce `hint_min_ts` during the flashback progress to only flashback those keys that have version changed as much as possible.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/src/storage/mvcc/reader/reader.rs
+++ b/src/storage/mvcc/reader/reader.rs
@@ -566,29 +566,23 @@ impl<S: EngineSnapshot> MvccReader<S> {
         Ok((locks, has_remain))
     }
 
-    /// Scan the writes to get all the latest keys with their corresponding
-    /// PUT/DELETE write records at the given version, if the version is not
-    /// specified, it will scan the latest version for each key, if the key
-    /// does not exist or is not visible at that point, an `Option::None` will
-    /// be placed. The return type is:
-    /// * `(Vec<(key, Option<write>)>, has_remain)`.
-    ///   - `key` is the encoded key without commit ts.
-    ///   - `write` is the PUT/DELETE write record at the given version.
-    ///   - `has_remain` indicates whether there MAY be remaining writes that
+    /// Scan the writes to get all the latest user keys. The return type is:
+    /// * `(Vec<key>, has_remain)`.
+    ///   - `key` is the encoded user key without `commit_ts`.
+    ///   - `has_remain` indicates whether there MAY be remaining user keys that
     ///     can be scanned.
     ///
     /// This function is mainly used by
     /// `txn::commands::FlashbackToVersionReadPhase`
     /// and `txn::commands::FlashbackToVersion` to achieve the MVCC
     /// overwriting.
-    pub fn scan_writes<F>(
+    pub fn scan_latest_user_keys<F>(
         &mut self,
         start: Option<&Key>,
         end: Option<&Key>,
-        version: Option<TimeStamp>,
         filter: F,
         limit: usize,
-    ) -> Result<(Vec<(Key, Option<Write>)>, bool)>
+    ) -> Result<(Vec<Key>, bool)>
     where
         F: Fn(&Key /* user key */, TimeStamp /* latest `commit_ts` */) -> bool,
     {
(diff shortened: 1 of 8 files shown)


src/storage/mvcc/reader/reader.rs
```diff
@@ -470,6 +476,8 @@ impl<S: EngineSnapshot> MvccReader<S> {
                 .prefix_seek(self.scan_mode.is_none())
                 .scan_mode(self.get_scan_mode(true))
                 .range(self.lower_bound.clone(), self.upper_bound.clone())
+                // `hint_min_ts` is only meaningful when it's the `commit_ts`.
```
sticnarf, line 479: What does this comment mean? I cannot understand it...

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -114,27 +123,32 @@ pub fn flashback_to_version_write(
         if txn.write_size() >= MAX_TXN_WRITE_SIZE {
             return Ok(Some(key.clone()));
         }
+        let old_write = reader.get_write(&key, flashback_version)?;
```
sticnarf, line 126: So, now you scan out the keys that need flashback, and use `get_write` to get the old write and rewrite. Could it cause a lot more `seek`? I'm not sure about the overall performance...

src/storage/mvcc/reader/reader.rs
JmPotato: I saw this line of comment: https://github.com/tikv/tikv/blob/master/src/storage/mvcc/reader/scanner/mod.rs#L323

Is this still correct or did I misunderstand it?

sticnarf: Ah, yeah. Write CF records commit ts range in its properties. I thought it's commented above the `hint_min_ts` field in the iterator and you were expressing something different.

You can say `hint_min_ts` filters data by commit_ts. "only meaningful when it's the `commit_ts`" is confusing.

src/storage/txn/actions/flashback_to_version.rs
JmPotato: Actual, the old way is also `scan` and `seek`: https://github.com/tikv/tikv/blob/dc347f98ae70e8f829a55d4df1f002b4ad17e72d/src/storage/mvcc/reader/reader.rs#L624-L630

🤔 This PR should reduce the overhead since it will reduce the key number we need to scan and seek. And furthermore, maybe we should rewrite the `scan_writes` to make it not seek the old version if we don't need it to read the old value now.

JmPotato: According to the current [implementation](https://github.com/tikv/tikv/blob/dc347f98ae70e8f829a55d4df1f002b4ad17e72d/src/storage/mvcc/reader/reader.rs#L625) of `scan_writes`, every key that needs to be flashbacked will also have a `seek` to find its latest visible write record before the given `version`, now since we don't need to get the record in `scan_writes` anymore, maybe we can rewrite it to only return the latest record without seeking. By this, the total number of `seek` should be kept as same as before.

JmPotato: Or we can just make `scan_writes` become `scan_latest_keys`.

sticnarf: > maybe we can rewrite it to only return the latest record without seeking

I doubt whether it's possible. In scan_write, after you get the latest version of one key, then how to move the cursor to the next key? near_seek is still the solution. In the current implementation, all the keys are scanned but each key is one near_seek. Now, all keys scanned need two near_seek but SSTs are filtered by hint.

Anyway, intuitionally this PR should win in real cases...

JmPotato: I rewrote `scan_writes` to `scan_latest_user_keys`, and now it will only return all the latest user keys, since we no longer need the write record. PTAL.

BTW, a real-world performance test is in progress by @fengou1, the results will be updated later.

sticnarf (approved):
LGTM

fengou1:
20 TB total data/20 TiKVs. each TiKV has 1 TB data with EBS volume configured 7000 IOPS/400MB/s:
without this PR, flashback takes: 
`I1124 03:24:54.378529       9 restore.go:164] [2022/11/24 03:24:54.378 +00:00] [INFO] [range_task.go:243] ["range task finished"] [name=br-flashback-runner] [startKey=] [endKey=] ["cost time"=3h57m50.972269424s] ["completed regions"=203653]`

with this PR, flashback takes:
`I1125 09:38:04.719140       9 restore.go:164] [2022/11/25 09:38:04.719 +00:00] [INFO] [range_task.go:243] ["range task finished"] [name=br-flashback-runner] [startKey=] [endKey=] ["cost time"=26m26.859229465s] ["completed regions"=203653]`

fengou1 (approved):

glorv (approved):
LGTM

JmPotato:
@sticnarf Could you please help merge this PR? Thx a lot.

PR #13845, opened 2022-11-24
reader: fix the invisible write record flashback bug

### LINKED ISSUE #13844: Flashback may overwrite wrong write record
## Bug Report

When `scan_writes` meets `WriteType::Lock | WriteType::Rollback`, it should immediately set `write` to `None`, otherwise if the current key is the last key, it will return the wrong write record to the flashback to perform the later overwriting.

https://github.com/tikv/tikv/blob/dc347f98ae70e8f829a55d4df1f002b4ad17e72d/src/storage/mvcc/reader/reader.rs#L640-L650

JmPotato:

Issue Number: Close #13844.

What's Changed:

```commit-message
Fix the bug that when a key's last write record is `WriteType::Lock` or `WriteType::Rollback`,
the `scan_writes` will return the invisible write record as the result, which will affect the flashback correctness.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None.
```

### DIFF
--- a/src/storage/mvcc/reader/reader.rs
+++ b/src/storage/mvcc/reader/reader.rs
@@ -638,13 +638,18 @@ impl<S: EngineSnapshot> MvccReader<S> {
                             break;
                         }
                         WriteType::Lock | WriteType::Rollback => {
-                            // We should find the latest visible version after it.
+                            // Only return the PUT/DELETE write record.
+                            write = None;
+                            // Reach the end.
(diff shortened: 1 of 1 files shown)


sticnarf (approved):

AndreMouche (approved):
LGTM

sticnarf (approved):

PR #13846, opened 2022-11-25
raftstore-v2: compact and gc raft logs

tabokie:

Issue Number: Ref #12842 

What's Changed:

```commit-message
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


components/raftstore-v2/src/worker/raft_log_gc.rs
```diff
@@ -0,0 +1,189 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: Can v1 version be reused?

tabokie: v1 interacts with kv engine.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,144 @@
...
+            return Err(box_err!(
+                "command format is outdated, please upgrade leader"
+            ));
+        }
+
+        let data = req.write_to_bytes().unwrap();
+        self.propose_with_ctx(store_ctx, data, vec![])
+    }
+
+    pub fn on_ready_compact_log<T>(
```
tabokie, line 53: I have no idea if it's correct to move the v1 check in `exec_compact_log` to this v2 function.

BusyJay:
Logs can't be gced until its data are all applied and flushed. So log gc depends on removing WAL. I suggest to continue the work after removing wal is supported.

components/raftstore-v2/src/operation/raft_log_gc.rs
```diff
@@ -0,0 +1,154 @@
...
+        admin.set_cmd_type(AdminCmdType::CompactLog);
+        admin.mut_compact_log().set_compact_index(compact_idx);
+        admin.mut_compact_log().set_compact_term(term);
+        req.set_admin_request(admin);
+
+        let (msg, _) = PeerMsg::raft_command(req);
+        if let Err(e) = store_ctx.router.send(self.region_id(), msg) {
+            error!(self.logger, "send compact log request failed");
+        }
+        // self.propose_command(store_ctx, req);
```
tonyxuqqi, line 150: `// self.propose_command(store_ctx, req);`
is it needed? If not, the comment can be deleted

components/raftstore-v2/src/operation/raft_log_gc.rs
```diff
@@ -0,0 +1,154 @@
...
+                .evict_entry_cache(true);
+            if !self.fsm.peer().entry_storage().is_entry_cache_empty() {
+                self.schedule_tick(PeerTick::EntryCacheEvict);
+            }
+        }
+    }
+}
+
+impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
+    fn raft_log_gc_imp<T>(&mut self, store_ctx: &mut StoreContext<EK, ER, T>) {
```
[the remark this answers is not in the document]
tabokie: No difference.

components/raftstore-v2/src/operation/raft_log_gc.rs
tabokie: Because I'm hesitating on the implementation. It's just very weird to use router to send message to the current peer. Alternatively I can directly call `on_admin_command` here, but I'm not sure if it's the intended usage of it. cc @BusyJay

components/raftstore-v2/src/operation/command/admin/compact_log.rs
tabokie: Because I seem to find access to apply_state in `Apply`, I'm not sure if it's supposed to be added.

components/raftstore-v2/src/operation/raft_log_gc.rs
BusyJay: Yes, you can call on_admin_command directly.

components/raftstore-v2/src/operation/raft_log_gc.rs
```diff
@@ -0,0 +1,146 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: The content of the files should be in operation/command/admin/compact_log.rs

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,144 @@
+// Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: Let's unify the file name and related struct/methods to raft log gc.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,144 @@
...
+        &mut self,
+        store_ctx: &mut StoreContext<EK, ER, T>,
+        res: CompactLogResult,
+    ) {
+        let first_index = self.entry_storage().first_index();
+        if res.compact_index <= first_index {
+            debug!(
+                self.logger,
+                "compact index <= first index, no need to compact";
+                "region_id" => self.region_id(),
```
BusyJay, line 113: Redundant fields as they are included in `self.logger`.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,144 @@
...
+        // TODO: check is_merging
+        // compact failure is safe to be omitted, no need to assert.
+        if res.compact_index <= self.entry_storage().truncated_index()
+            || res.compact_index > self.entry_storage().applied_index()
+        {
+            return;
+        }
+
+        // TODO: check entry_cache_warmup_state
+        self.schedule_raft_log_gc(store_ctx, res.compact_index);
```
BusyJay, line 129: It should not schedule raft log gc until all logs are applied and flushed.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
tabokie: What do you mean? compact_log -> raft_log_gc?

BusyJay: Yes.

components/raftstore-v2/src/batch/store.rs
```diff
@@ -432,6 +439,11 @@ impl<EK: KvEngine, ER: RaftEngine> StoreSystem<EK, ER> {
             ),
         );
 
+        let engine_gc_scheduler = workers.engine_gc_worker.start_with_timer(
```
BusyJay, line 442: Confusing name.

components/raftstore-v2/src/worker/engine_gc.rs
```diff
@@ -0,0 +1,204 @@
...
+pub enum Task {
+    RaftLogGc {
+        region_id: u64,
+        start_idx: u64,
+        end_idx: u64,
+    },
+}
+
+impl Task {
+    pub fn raft_log_gc(region_id: u64, start_idx: u64, end_idx: u64) -> Self {
```
BusyJay, line 27: Because we always use `RaftLogEngine` for v2, so it seems we don't need a worker at all. The logs can be just deleted in async writer a long side with other metas.

components/raftstore-v2/src/operation/command/admin/raft_log_gc.rs
```diff
@@ -0,0 +1,279 @@
...
+                "first_index" => first_index,
+            );
+            return;
+        }
+        // TODO: check is_merging
+        // compact failure is safe to be omitted, no need to assert.
+        if res.compact_index <= self.entry_storage().truncated_index() {
+            return;
+        }
+        res.compact_index = std::cmp::min(
```
BusyJay, line 230: What if those writes are flushed latter? I think you need a queue like `split_trace` to trace the compacted index and gc them after applied index is advanced.

components/raftstore-v2/src/batch/store.rs
tabokie: I was planning to put all engine related works into it, like tablet gc, log gc, purge etc.

components/raftstore-v2/src/operation/command/admin/raft_log_gc.rs
BusyJay: Rethink about it, the simplest way to do this is to do raft log gc whenever there is data flushed or raft log gc applied. The min(truncated_index, admin.flushed) is chosen so there doesn't need to be queue.

tabokie: How about trigger gc in peer::on_data_flushed, and do the actual gc in store async writer. I think avoiding sync write in apply pool is better in the long run.

BusyJay: > trigger gc in peer::on_data_flushed, and do the actual gc in store async writer.

That's amost what I mean. But it can't be done in `on_data_flushed`, it can only delete old logs when `persisted_applied` is advanced.

components/raftstore-v2/src/operation/command/admin/raft_log_gc.rs
```diff
@@ -0,0 +1,271 @@
...
+
+//! This module contains processing logic of the following:
+//!
+//! # `RaftLogGc` and `EntryCacheEvict` ticks
+//!
+//! On region leader, periodically compacts useless Raft logs from the
+//! underlying log engine, and evicts logs from entry cache if it reaches memory
+//! limit.
+//!
+//! # `CompactLog` command
```
BusyJay, line 11: Maybe we should rename the command to `GcLog`. I think keeping two names are confusing. Better always use gc log.

tabokie: But isn't that a name from kvproto? Might as well change log_gc to log_compact..

components/raftstore-v2/src/operation/command/admin/raft_log_gc.rs
tabokie: So in the latest version I check for persisted_applied in `maybe_compact_log_from_engine`.

components/raftstore-v2/src/operation/command/admin/raft_log_gc.rs
BusyJay: Yes. To minimize the impact for now, we can just change everything within TiKV.

tabokie: Then I think we should unify to compact instead, it involves fewer changes and is consistent with raft paper.

BusyJay: That works too.

components/raft_log_engine/src/engine.rs
```diff
@@ -658,34 +658,53 @@ impl RaftEngine for RaftLogEngine {
...
+                raft_group_id,
+                Some(REGION_STATE_KEY),
+                None,
+                false,
+                |key, _| {
+                    match &key[..key_len] {
+                        REGION_STATE_KEY | APPLY_STATE_KEY | FLUSH_STATE_KEY
+                            if NumberCodec::decode_u64(&key[key_len..]) < apply_index =>
+                        {
+                            batch.0.delete(raft_group_id, key.to_vec());
```
BusyJay, line 699: There should be at least one record before apply index, otherwise there is not recovery point.

tabokie: This function is supposed to be called with persisted_apply, I believe there will always be at least one record >= persisted_apply?

BusyJay: Not necessary. For example, if all CFs are flushed, then the apply indexes should be advanced to the latest. There is no states larger than persisted apply apparently.

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -115,9 +96,12 @@ impl<EK: KvEngine, ER: RaftEngine> Runner<EK, ER> {
             panic!("failed to sync kv_engine in raft_log_gc: {:?}", e);
         });
         RAFT_LOG_GC_KV_SYNC_DURATION_HISTOGRAM.observe(start.saturating_elapsed_secs());
+
+        fail::fail_point!("worker_gc_raft_log");
```
BusyJay, line 100: I don't think it's the same as before if configured with "return". In the past, only `batch_gc` is skipped. But now all the following processing are skipped.

components/raft_log_engine/src/engine.rs
```diff
@@ -658,34 +658,53 @@ impl RaftEngine for RaftLogEngine {
...
+        debug_assert!(REGION_STATE_KEY.len() == APPLY_STATE_KEY.len());
+        debug_assert!(REGION_STATE_KEY.len() == FLUSH_STATE_KEY.len());
+
+        let key_len = REGION_STATE_KEY.len();
+        self.0
+            .scan_raw_messages(
+                raft_group_id,
+                Some(REGION_STATE_KEY),
+                None,
+                false,
```
BusyJay, line 693: I'm not sure if the performance is the same, but if scan reverse, the logic can be a lot simplified to keep one version before apply index. Or we can also reverse encode the index if forward scan is more performant.

components/raft_log_engine/src/engine.rs
```diff
@@ -658,34 +658,83 @@ impl RaftEngine for RaftLogEngine {
...
+                            if NumberCodec::decode_u64(&key[key_len..]) < apply_index =>
+                        {
+                            if found_apply_state {
+                                batch.0.delete(raft_group_id, key.to_vec());
+                            } else {
+                                found_apply_state = true;
+                            }
+                        }
+                        FLUSH_STATE_KEY
+                            if NumberCodec::decode_u64(&key[key_len..]) < apply_index =>
```
BusyJay, line 723: Need to check cf id.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl<EK: KvEngine, R: ApplyResReporter> Apply<EK, R> {
+    pub fn apply_compact_log(
+        &mut self,
+        req: &AdminRequest,
+        _log_index: u64,
+    ) -> Result<(AdminResponse, AdminCmdResult)> {
+        Ok((
+            AdminResponse::default(),
+            AdminCmdResult::CompactLog(CompactLogResult {
```
BusyJay, line 202: Apply state should be changed, and you need log index to correctly store apply state.

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -137,28 +108,40 @@ impl<EK: KvEngine, ER: RaftEngine> Runner<EK, ER> {
...
-            cb()
+        fail::fail_point!("worker_gc_raft_log_finished");
+        cbs
+    }
+
+    fn flush(&mut self) {
+        if self.tasks.is_empty() {
+            return;
+        }
+        fail::fail_point!("worker_gc_raft_log_flush");
```
BusyJay, line 134: I don't see how this makes the behavior the same as past. As this is not key path, how about leave this file as it is to avoid introduce potential risk to v1?

tabokie: The interface is changed, now we don't now how many entries are compacted, so we don't report_collected anymore. Note that `report_collected` is only used in unit tests below.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
tabokie: Which state? apply state in peer is updated in `on_apply_res_compact_log`.

BusyJay: `RaftApplyState`. `on_apply_res_compact_log` only updates the memory version, not the disk version.

components/raft_log_engine/src/engine.rs
```diff
@@ -380,6 +381,7 @@ const REGION_STATE_KEY: &[u8] = &[0x03];
 const APPLY_STATE_KEY: &[u8] = &[0x04];
 const RECOVER_STATE_KEY: &[u8] = &[0x05];
 const FLUSH_STATE_KEY: &[u8] = &[0x06];
+const KEY_PREFIX_LEN: usize = 1;
```
BusyJay, line 384: Better add a comment to explain why it's 1.

components/raft_log_engine/src/engine.rs
```diff
@@ -658,34 +660,80 @@ impl RaftEngine for RaftLogEngine {
...
+                        {
+                            if found_apply_state {
+                                batch.0.delete(raft_group_id, key.to_vec());
+                            } else {
+                                found_apply_state = true;
+                            }
+                        }
+                        FLUSH_STATE_KEY => {
+                            let cf_id =
+                                NumberCodec::decode_u64(&key[KEY_PREFIX_LEN..KEY_PREFIX_LEN + 1]);
```
BusyJay, line 721: `decode_u64` consumes 8 bytes and returns u64. Apparently here it just need one byte. It's not varint codec.

BusyJay: Reconsider it a bit, the cleanup for flushed key is wrong. Flush key is encoded as `{flush key prefix}{cf_id}{tablet_index} -> {apply index}`. So you should not delete it until `tablet_index` is gced.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    pub fn maybe_compact_log_from_engine<T>(&mut self, store_ctx: &mut StoreContext<EK, ER, T>) {
+        let compact_index = std::cmp::min(
+            self.entry_storage().truncated_index(),
+            self.storage().apply_trace().persisted_apply_index(),
+        );
+        if compact_index > self.last_engine_compact_log_index() {
+            let region_id = self.region_id();
+            let lb = self.state_changes_mut();
```
BusyJay, line 257: After `state_changes` is mutated, should call `set_has_extra_writes`.

components/raftstore-v2/src/fsm/peer.rs
```diff
@@ -279,6 +281,9 @@ impl<'a, EK: KvEngine, ER: RaftEngine, T: Transport> PeerFsmDelegate<'a, EK, ER,
                     self.fsm
                         .peer_mut()
                         .on_data_flushed(cf, tablet_index, flushed_index);
+                    self.fsm
+                        .peer_mut()
+                        .maybe_compact_log_from_engine(self.store_ctx);
```
BusyJay, line 286: Should be called where `record_apply_trace` is called.

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -137,28 +109,39 @@ impl<EK: KvEngine, ER: RaftEngine> Runner<EK, ER> {
...
-                self.report_collected(0);
                 RAFT_LOG_GC_FAILED.inc();
             }
-            Ok(n) => {
-                debug!("gc log entries";  "entry_count" => n);
-                self.report_collected(n);
-                RAFT_LOG_GC_DELETED_KEYS_HISTOGRAM.observe(n as f64);
-            }
+        }
+        if let Err(e) = self.engines.raft.consume(&mut batch, false) {
```
BusyJay, line 121: Maybe you can keep the most of original flush by:
```
fn gc_raft_log() {
   ...
   consume(batch)
   ...
}

fn flush() {
    ...
    for _ in _ {
        engine.gc(batch);
    }
    ...
}
```

So most code is unchanged and easier to understand.

components/raftstore-v2/src/operation/command/admin/compact_log.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        self.maybe_compact_log_from_engine(store_ctx);
+    }
+
+    pub fn maybe_compact_log_from_engine<T>(&mut self, store_ctx: &mut StoreContext<EK, ER, T>) {
+        let compact_index = std::cmp::min(
+            self.entry_storage().truncated_index(),
+            self.storage().apply_trace().persisted_apply_index(),
+        );
+        if compact_index > self.last_engine_compact_log_index() {
```
BusyJay, line 255: `last_engine_compact_log_index` may not be necessary. For `on_ready_res`, it should compact log if `truncated_index < persisted_applied` before applying the updates, For `record_apply_trace`, it should compact log if `persisted_applied < truncated_index`.

components/raftstore/src/store/worker/raftlog_gc.rs
tabokie: If it's the failpoint bothering you, the old code is simply wrong. It skips writing log batch but doesn't skip consuming callbacks. It is not the intended behavior as per https://github.com/tikv/tikv/blob/8864e9bac814554bbe4acaddd01282f5fdae3dae/tests/failpoints/cases/test_snap.rs#L701-L702

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -5423,12 +5423,9 @@ where
         fail_point!("on_entry_cache_evict_tick", |_| {});
         if needs_evict_entry_cache(self.ctx.cfg.evict_cache_on_memory_ratio) {
             self.fsm.peer.mut_store().evict_entry_cache(true);
-        }
-        let mut _usage = 0;
-        if memory_usage_reaches_high_water(&mut _usage)
```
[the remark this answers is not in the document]
tabokie: In the next tick it will call `needs_evict_entry_cache` again, which calls `memory_usage_reaches_high_water` inside. I think checking later again makes more sense than checking right after a compaction.

Also, this further hides the function `memory_usage_reaches_high_water` and keeps the pattern consistent with https://github.com/tikv/tikv/blob/8864e9bac814554bbe4acaddd01282f5fdae3dae/components/raftstore/src/store/fsm/peer.rs#L5350-L5355

components/raftstore-v2/src/operation/command/admin/compact_log.rs
tabokie: You are right. But it makes the code a little more complex, and it introduces an extra dependency between caller and callee.

tabokie: On a second thought, I'll just add a new function `on_persisted_apply_advanced`, which is probably useful for tablet gc too.

BusyJay: Why? For example, `truncated_index` is advanced from 3 to 10 in `on_ready_res`, while `persisted_applied is 5. Because 3 < 5, so it should gc logs using `cmp::min(10, 5) = 5`.

tabokie: I see. It is still very complex.

BusyJay: It's just two if check to get rid of a field (and two if check).

components/raftstore/src/store/worker/raftlog_gc.rs
tabokie: Regardless of the behavior it is not what the test comment says, because tasks are lost forever. For RocksDB GC it means some entries will never be compacted again. So we will change the failpoint comment if that's what you prefer.

BusyJay: The return action is to simulate the case that the node is stopped before completing GC. It's more appropriate to use "pause", but "pause" will stop the case from restarting node.

On the other hand, writing raft engine can fail. If that error happens, we just ignore it for now. I think they are two different things.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -450,6 +456,18 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+    /// deadline time.
+    #[inline]
+    pub fn peer_heartbeat_is_fresh(&self, peer_id: u64, deadline: &Instant) -> bool {
+        if let Some(last_heartbeat) = self.peer_heartbeats.get(&peer_id)
+            && *last_heartbeat >= *deadline
+        {
+            return true;
+        }
+        false
+    }
```
BusyJay, line 469: You mean `matches!(self.peer_heartbeats.get(&peer_id), Some(last_heartbat) if *last_heartbeat >= *deadline)`?

components/raftstore/src/store/worker/raftlog_gc.rs
```diff
@@ -137,29 +127,21 @@ impl<EK: KvEngine, ER: RaftEngine> Runner<EK, ER> {
...
                 RAFT_LOG_GC_FAILED.inc();
             }
-            Ok(n) => {
-                debug!("gc log entries";  "entry_count" => n);
-                self.report_collected(n);
-                RAFT_LOG_GC_DELETED_KEYS_HISTOGRAM.observe(n as f64);
-            }
         }
-        RAFT_LOG_GC_WRITE_DURATION_HISTOGRAM.observe(start.saturating_elapsed_secs());
-        for cb in cbs {
```
BusyJay, line 160: Callback are not collected and invoked.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

Connor1996 (approved):
LGTM

tabokie:
Latest commit needs review. @bufferflies @BusyJay

BusyJay (approved):

PR #13850, opened 2022-11-28
cop: support batch coprocessor processing in tikv

### LINKED ISSUE #13849: cop: support batch coprocessor task processing in tikv
## Development Task

Ref: https://github.com/pingcap/tidb/issues/39361
Ref(RFC): https://github.com/pingcap/tidb/pull/39362 

Support processing batch coprocessor tasks in tikv to reduce the task queuing and RPC overhead.

cfzjywxk:

Issue Number: Close #13849 

What's Changed:
The proto change is https://github.com/pingcap/kvproto/pull/1018.

```commit-message
Support batch coprocessor task processing in tikv. All the tasks would be passed to tikv in a single RPC
request, they would be executed concurrently and all the results would be sent back in one response.
More tests and investigations are needed together  with the coprocessor client changes.
```

### Check List 

Tests 

- Unit test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```

### DIFF
--- a/src/coprocessor/endpoint.rs
+++ b/src/coprocessor/endpoint.rs
@@ -566,13 +566,14 @@ impl<E: Engine> Endpoint<E> {
             .collect();
         let mut index = 0;
         for cur_req in batch_reqs.into_iter() {
-            let cur_tracker = GLOBAL_TRACKERS.insert(::tracker::Tracker::new(RequestInfo::new(
+            let request_info = RequestInfo::new(
                 cur_req.get_context(),
                 RequestType::Unknown,
                 cur_req.start_ts,
-            )));
+            );
             match self.parse_request_and_check_memory_locks(cur_req, peer.clone(), false) {
                 Ok((handler_builder, req_ctx)) => {
+                    let cur_tracker = GLOBAL_TRACKERS.insert(::tracker::Tracker::new(request_info));
                     set_tls_tracker_token(cur_tracker);
                     batch_futs.push(self.handle_unary_request(req_ctx, handler_builder));
                     batch_extra.push((index, cur_tracker));
--- a/src/coprocessor/endpoint.rs
+++ b/src/coprocessor/endpoint.rs
@@ -1,6 +1,8 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::{borrow::Cow, future::Future, marker::PhantomData, sync::Arc, time::Duration};
+use std::{
+    borrow::Cow, future::Future, iter::FromIterator, marker::PhantomData, sync::Arc, time::Duration,
+};
 
 use ::tracker::{
     set_tls_tracker_token, with_tls_tracker, RequestInfo, RequestType, GLOBAL_TRACKERS,
@@ -9,7 +11,6 @@ use async_stream::try_stream;
 use concurrency_manager::ConcurrencyManager;
 use engine_traits::PerfLevel;
 use futures::{channel::mpsc, prelude::*};
-use futures_util::future::join_all;
 use kvproto::{coprocessor as coppb, coprocessor::StoreBatchTaskResponse, errorpb, kvrpcpb};
 use protobuf::{CodedInputStream, Message};
 use resource_metering::{FutureExt, ResourceTagFactory, StreamExt};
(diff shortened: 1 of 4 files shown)


src/coprocessor/endpoint.rs
```diff
@@ -485,33 +486,107 @@ impl<E: Engine> Endpoint<E> {
...
+            .take_tasks()
+            .iter_mut()
+            .map(|task| {
+                let mut new_req = req.clone();
+                new_req.ranges = task.take_ranges();
+                let mut new_context = new_req.take_context();
+                new_context.set_region_id(task.get_region_id());
+                new_context.set_region_epoch(task.take_region_epoch());
+                new_context.set_peer(task.take_peer());
+                new_req.set_context(new_context);
```
sticnarf, line 511:
```suggestion
                let new_context = new_req.mut_context();
                new_context.set_region_id(task.get_region_id());
                new_context.set_region_epoch(task.take_region_epoch());
                new_context.set_peer(task.take_peer());
```

src/coprocessor/endpoint.rs
```diff
@@ -485,33 +486,107 @@ impl<E: Engine> Endpoint<E> {
...
+            match self.parse_request_and_check_memory_locks(cur_req, peer.clone(), false) {
+                Ok((handler_builder, req_ctx)) => {
+                    batch_fut_results.push(self.handle_unary_request(req_ctx, handler_builder));
+                    batch_extra_info.push((index, cur_tracker));
+                }
+                Err(e) => {
+                    make_error_batch_response(&mut batch_responses[index], e);
+                }
+            }
+            batch_responses[index].task_id = index as u64;
```
sticnarf, line 534: Should we set the `task_id` in the request to the one in the response? (Then, it should be set in L512)

src/coprocessor/endpoint.rs
```diff
@@ -485,33 +486,107 @@ impl<E: Engine> Endpoint<E> {
...
+        mut req: coppb::Request,
         peer: Option<String>,
     ) -> impl Future<Output = MemoryTraceGuard<coppb::Response>> {
         let tracker = GLOBAL_TRACKERS.insert(::tracker::Tracker::new(RequestInfo::new(
             req.get_context(),
             RequestType::Unknown,
             req.start_ts,
         )));
+        let batch_task_num = req.tasks.len();
+        let mut batch_fut_results = Vec::with_capacity(batch_task_num + 1);
```
sticnarf, line 498: I suggest to extract all the batch part to a new function. And I think it's not necessary to combine the batch tasks with the original one. 

Then, we don't need the extra allocation if there is no batch tasks. The future that joins all batch tasks also process the results into `batch_responses`. I think it will be clearer.

cfzjywxk: My initial thought is that combining all the futures of original and batch tasks and waiting for them together so the latency would be `max(req_latency)`, but the logic also becomes complex. If the batched tasks are processed and waited separately the latency would be increased, especially when the original task is executing slowly. Any suggestion about this?

sticnarf: It's still possible to execute them concurrently.

One is the original task `fut1` and the other is the batched tasks `fut2` created by `join_all`. Then, use the `futures::join` macro to get both the results.

sticnarf (commented):
LGTM

src/coprocessor/endpoint.rs
```diff
@@ -485,38 +487,113 @@ impl<E: Engine> Endpoint<E> {
...
+            .iter_mut()
+            .map(|task| {
+                let mut new_req = req.clone();
+                new_req.ranges = task.take_ranges();
+                let new_context = new_req.mut_context();
+                new_context.set_region_id(task.get_region_id());
+                new_context.set_region_epoch(task.take_region_epoch());
+                new_context.set_peer(task.take_peer());
+                let mut response = coppb::StoreBatchTaskResponse::new();
+                response.set_task_id(task.get_task_id());
```
MyonKeminta, line 547: Is this used?

src/coprocessor/endpoint.rs
```diff
@@ -654,6 +731,31 @@ impl<E: Engine> Endpoint<E> {
     }
 }
 
+fn make_error_batch_response(batch_resp: &mut coppb::StoreBatchTaskResponse, e: Error) {
+    match e {
```
MyonKeminta, line 735: Do we need a warn log and update the `COPR_REQ_ERROR` metrics like `make_error_response` here?

MyonKeminta (commented):
Rest LGTM

src/coprocessor/endpoint.rs
cfzjywxk: I forget to remove these lines, thansk for the reminding.

src/coprocessor/endpoint.rs
cfzjywxk: I think so, related information is added.

MyonKeminta (approved):

src/coprocessor/endpoint.rs
```diff
@@ -485,38 +487,113 @@ impl<E: Engine> Endpoint<E> {
...
+            .take_tasks()
+            .iter_mut()
+            .map(|task| {
+                let mut new_req = req.clone();
+                new_req.ranges = task.take_ranges();
+                let new_context = new_req.mut_context();
+                new_context.set_region_id(task.get_region_id());
+                new_context.set_region_epoch(task.take_region_epoch());
+                new_context.set_peer(task.take_peer());
+                let mut response = coppb::StoreBatchTaskResponse::new();
```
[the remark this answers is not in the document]
ekexium: OK already removed

sticnarf (approved):

cfzjywxk:
@sticnarf @you06 @ekexium 
PTAL need another approval, thx

src/coprocessor/endpoint.rs
```diff
@@ -485,38 +487,111 @@ impl<E: Engine> Endpoint<E> {
...
+                Ok((handler_builder, req_ctx)) => {
+                    let cur_tracker = GLOBAL_TRACKERS.insert(::tracker::Tracker::new(request_info));
+                    set_tls_tracker_token(cur_tracker);
+                    let fut = self.handle_unary_request(req_ctx, handler_builder);
+                    let fut = async move {
+                        let res = fut.await;
+                        match res {
+                            Ok(mut resp) => {
+                                response.set_data(resp.take_data());
+                                response.set_region_error(resp.take_region_error());
```
you06, line 567: `resp.take_region_error` will create empty error(not `None`) when there is no region error.

PR #13854, opened 2022-11-28
tikv_kv: make async_write return stream

BusyJay:
Issue Number: Ref #13827

What's Changed:
```commit-message
This PR abstracts write interface with Stream trait so that we can keep compatible with
both v1 and v2.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/src/server/raftkv.rs
+++ b/src/server/raftkv.rs
@@ -50,7 +50,6 @@ use tikv_kv::{write_modifies, OnAppliedCb, WriteEvent};
 use tikv_util::{
     codec::number::NumberEncoder,
     future::{paired_future_callback, paired_must_called_future_callback},
-    mpsc::future::WakePolicy,
     time::Instant,
 };
 use txn_types::{Key, TimeStamp, TxnExtra, TxnExtraScheduler, WriteBatchFlags};
@@ -235,7 +234,7 @@ impl Stream for WriteResSub {
             WriteEvent::EVENT_COMMITTED => Poll::Ready(Some(WriteEvent::Committed)),
             u8::MAX => {
                 let result = unsafe { (*s.core.result.get()).take().unwrap() };
-                return Poll::Ready(Some(WriteEvent::Finished(result)));
+                Poll::Ready(Some(WriteEvent::Finished(result)))
             }
             e => panic!("unexpected event {}", e),
         }
@@ -281,7 +280,7 @@ impl WriteResFeed {
 
     fn notify(&self, result: kv::Result<()>) {
         unsafe {
-            (*self.core.result.get()).insert(result);
+            (*self.core.result.get()) = Some(result);
         }
         self.core.ev.store(u8::MAX, Ordering::Release);
         self.core.wake.wake();
--- a/src/server/raftkv.rs
+++ b/src/server/raftkv.rs
@@ -3,19 +3,25 @@
 // #[PerformanceCriticalPath]
 use std::{
     borrow::Cow,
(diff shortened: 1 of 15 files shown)


BusyJay:
Microbench shows that the new API takes about 700ns in worst case (always drop the requests) while the old API takes about 670ns.

BusyJay:
benchbot shows no obvious difference.

Bench Type | threads | master | pr | diff
----------|--|--|--|--
oltp insert | 200 | 44416 | 44269 | -0.33%
oltp insert | 100 | 33916 | 34437 | 1.54%
oltp read_write | 200 | 68916 | 67530 | -2.01%
oltp read_write | 100 | 54957 | 57442 | 4.52%

Connor1996 (approved):
LGTM

src/storage/txn/scheduler.rs
```diff
@@ -1498,23 +1491,14 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
...
-                    SCHED_STAGE_COUNTER_VEC.get(tag).async_write_err.inc();
-
-                    info!("engine async_write failed"; "cid" => cid, "err" => ?e);
-                    scheduler.finish_with_err(cid, e);
+                    return;
                 }
-            })
+            }
         }
+        // If it's not finished while the channel is closed, it means the write
```
sticnarf, line 1498: Should we log an error here to indicate TiKV falls into an abnormal state?

src/server/raftkv.rs
```diff
@@ -208,6 +198,95 @@ pub fn drop_snapshot_callback<T>() -> kv::Result<T> {
...
+        let mut cur_ev = s.core.ev.load(Ordering::Acquire);
+        if cur_ev == s.notified_ev {
+            s.core.wake.register(cx.waker());
+            cur_ev = s.core.ev.load(Ordering::Acquire);
+            if cur_ev == s.notified_ev {
+                return Poll::Pending;
+            }
+        }
+        s.notified_ev = cur_ev;
+        match cur_ev {
```
sticnarf, line 232: This means a later event will overwrite the previous one even if the event is not processed?

Then, if a request subscribes `Proposed` but `core.ev` already becomes `Committed` when it polls, it will  miss the `Proposed` event.

src/storage/txn/scheduler.rs
BusyJay: If TiKV is shutting down, the sender is dropped silently.

src/server/raftkv.rs
BusyJay: Yes, it's possible generally. But it won't happen for any known usage. Scheduler will either subscribe proposed or committed not both. If the event is overwritten, it must be written by final result.

tests/benches/hierarchy/mvcc/mod.rs
```diff
@@ -60,7 +60,7 @@ where
         .unwrap();
     }
     let write_data = WriteData::from_modifies(txn.into_modifies());
-    let _ = engine.async_write(&ctx, write_data, Box::new(move |_| {}));
+    let _ = tikv_kv::write(engine, &ctx, write_data, None);
```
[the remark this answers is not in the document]
BusyJay: I guess it relies on the messages handling order to speed up processing. Snapshot and write are handled one by one in other thread, so when the snapshot is taken, the writes must be persisted.

sticnarf (approved):

PR #13860, opened 2022-11-30
storage: Unify the flashback reader and fix the start_key bug

### LINKED ISSUE #13861: Flashback may prewrite and commit an illegal key

c1aceb003b9da06b75b70a1e545d52b994ab67dc

Doesn't matter.

### Steps to reproduce

Flashback after https://github.com/tikv/tikv/pull/13823 will prewrite and commit the `start_key` to block the `resolved_ts`, but since the `start_key` passed in by TiDB could be an illegal key, e.g, without the `row_id`, which will make the TiFlash panic as https://github.com/pingcap/tiflash/issues/6379#issuecomment-1331758294 said.

HuSharp:

Issue Number: Close #13861  Close https://github.com/pingcap/tiflash/issues/6379 Ref #13800 

What's Changed:

```commit-message
SnapshotReader typically uses its own start_ts for something. Since it doesn't need the start_ts of SnapshotReader itself,  
we can unify the reader into MvccReader.
And the start key from the client is actually a range, which is used to limit the upper bound of this flashback when scanning data, so it may not be a real key.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/txn/actions/flashback_to_version.rs
+++ b/src/storage/txn/actions/flashback_to_version.rs
@@ -7,7 +7,7 @@ use txn_types::{Key, Lock, LockType, TimeStamp, Write, WriteType};
 use crate::storage::{
     mvcc::{MvccReader, MvccTxn, SnapshotReader, MAX_TXN_WRITE_SIZE},
     txn::{actions::check_txn_status::rollback_lock, Result as TxnResult},
-    Snapshot, Statistics,
+    Snapshot,
 };
 
 pub const FLASHBACK_BATCH_SIZE: usize = 256 + 1 /* To store the next key for multiple batches */;
@@ -16,7 +16,6 @@ pub fn flashback_to_version_read_lock(
     reader: &mut MvccReader<impl Snapshot>,
     next_lock_key: Key,
     end_key: &Key,
-    statistics: &mut Statistics,
 ) -> TxnResult<Vec<(Key, Lock)>> {
     let result = reader.scan_locks(
         Some(&next_lock_key),
@@ -24,7 +23,6 @@ pub fn flashback_to_version_read_lock(
         |_| true,
         FLASHBACK_BATCH_SIZE,
     );
-    statistics.add(&reader.statistics);
     let (key_locks, _) = result?;
     Ok(key_locks)
 }
@@ -36,7 +34,6 @@ pub fn flashback_to_version_read_write(
     end_key: &Key,
     flashback_version: TimeStamp,
     flashback_commit_ts: TimeStamp,
-    statistics: &mut Statistics,
 ) -> TxnResult<Vec<Key>> {
     // Filter out the SST that does not have a newer version than
     // `flashback_version` in `CF_WRITE`, i.e, whose latest `commit_ts` <=
@@ -62,7 +59,6 @@ pub fn flashback_to_version_read_write(
         },
         FLASHBACK_BATCH_SIZE,
     );
-    statistics.add(&reader.statistics);
     let (keys, _) = keys_result?;
     Ok(keys)
 }
@@ -74,14 +70,16 @@ pub fn rollback_locks(
     snapshot: impl Snapshot,
     key_locks: Vec<(Key, Lock)>,
 ) -> TxnResult<Option<Key>> {
+    let mut reader = SnapshotReader::new(txn.start_ts, snapshot, false);
     for (key, lock) in key_locks {
         if txn.write_size() >= MAX_TXN_WRITE_SIZE {
             return Ok(Some(key));
         }
         // To guarantee rollback with start ts of the locks
+        reader.start_ts = lock.ts;
         rollback_lock(
             txn,
-            &mut SnapshotReader::new(lock.ts, snapshot.clone(), false),
+            &mut reader,
             key.clone(),
             &lock,
             lock.is_pessimistic_txn(),
@@ -152,31 +150,29 @@ pub fn flashback_to_version_write(
 pub fn prewrite_flashback_key(
     txn: &mut MvccTxn,
     reader: &mut MvccReader<impl Snapshot>,
-    key_to_lock: &Key,
+    start_key: &Key,
     end_key: &Key,
     flashback_version: TimeStamp,
     flashback_start_ts: TimeStamp,
 ) -> TxnResult<()> {
-    let first_key = get_first_user_key(reader, key_to_lock, end_key);
-    let key_to_lock = if let Some(ref key_to_lock) = first_key {
-        key_to_lock
+    let key_to_lock = if let Some(first_key) = get_first_user_key(reader, start_key, end_key)? {
+        first_key
     } else {
         return Ok(());
     };
-
-    let old_write = reader.get_write(key_to_lock, flashback_version, None)?;
+    let old_write = reader.get_write(&key_to_lock, flashback_version, None)?;
     // Flashback the value in `CF_DEFAULT` as well if the old write is a
     // `WriteType::Put` without the short value.
     if let Some(old_write) = old_write.as_ref() {
         if old_write.write_type == WriteType::Put
             && old_write.short_value.is_none()
             // If the value with `flashback_start_ts` already exists, we don't need to write again.
-            && reader.get_value(key_to_lock, flashback_start_ts)?.is_none()
+            && reader.get_value(&key_to_lock, flashback_start_ts)?.is_none()
         {
             txn.put_value(
                 key_to_lock.clone(),
                 flashback_start_ts,
-                reader.load_data(key_to_lock, old_write.clone())?,
+                reader.load_data(&key_to_lock, old_write.clone())?,
             );
         }
     }
@@ -236,19 +232,12 @@ pub fn get_first_user_key(
     reader: &mut MvccReader<impl Snapshot>,
     start_key: &Key,
     end_key: &Key,
-) -> Option<Key> {
-    // The start key from TiDB is actually a range, which is used to limit the
-    // upper bound of this flashback when scanning data, so is not a real key.
-    let (keys, _) = reader
-        .scan_latest_user_keys(Some(start_key), Some(end_key), |_, _| true, 1)
-        .unwrap();
-    let key_to_lock = if !keys.is_empty() {
-        keys[0].clone()
-    } else {
-        // If there is no key in user keys, we should return directly.
-        return None;
-    };
-    Some(key_to_lock)
+) -> TxnResult<Option<Key>> {
+    // The start key from the client is actually a range, which is used to limit the
+    // upper bound of this flashback when scanning data, so it may not be a real key.
+    let (mut keys_result, _) =
+        reader.scan_latest_user_keys(Some(start_key), Some(end_key), |_, _| true, 1)?;
+    Ok(keys_result.pop())
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


src/storage/txn/commands/flashback_to_version.rs
```diff
@@ -71,23 +71,20 @@ impl CommandExt for FlashbackToVersion {
 
 impl<S: Snapshot, L: LockManager> WriteCommand<S, L> for FlashbackToVersion {
     fn process_write(mut self, snapshot: S, context: WriteContext<'_, L>) -> Result<WriteResult> {
-        let mut reader = ReaderWithStats::new(
```
JmPotato, line 74: One of the main reasons we use `SnapshotReader` here is that `ReaderWithStats::new` returns a `SnapshotReader` wrapped with `context.statistics`. If we want to use `MvccReader`, the statics info needs to be handled manually by ourselves.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -71,18 +71,17 @@ pub fn flashback_to_version_read_write(
...
     for (key, lock) in key_locks {
         if txn.write_size() >= MAX_TXN_WRITE_SIZE {
             return Ok(Some(key));
         }
         // To guarantee rollback with start ts of the locks
-        reader.start_ts = lock.ts;
         rollback_lock(
             txn,
-            reader,
+            &mut SnapshotReader::new(lock.ts, snapshot.clone(), false),
```
JmPotato, line 84: It's not necessary to create a `SnapshotReader` for each key, in fact, you can just create one outside the loop.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -152,19 +151,27 @@ pub fn flashback_to_version_write(
 // transaction.
 pub fn prewrite_flashback_key(
     txn: &mut MvccTxn,
-    reader: &mut SnapshotReader<impl Snapshot>,
+    reader: &mut MvccReader<impl Snapshot>,
     key_to_lock: &Key,
```
JmPotato, line 155: It's not the key to lock now, rename it back to `start_key`.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -152,19 +151,27 @@ pub fn flashback_to_version_write(
...
     flashback_start_ts: TimeStamp,
 ) -> TxnResult<()> {
-    let old_write = reader.get_write(key_to_lock, flashback_version)?;
+    let first_key = get_first_user_key(reader, key_to_lock, end_key);
+    let key_to_lock = if let Some(ref key_to_lock) = first_key {
+        key_to_lock
+    } else {
+        return Ok(());
+    };
+
```
JmPotato, line 166:
```suggestion
    let key_to_lock = get_first_user_key(reader, key_to_lock, end_key);
    if key_to_lock.is_none() {
        return Ok(());
    };
```

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -225,6 +232,25 @@ pub fn commit_flashback_key(
     Ok(())
 }
 
+pub fn get_first_user_key(
+    reader: &mut MvccReader<impl Snapshot>,
+    start_key: &Key,
+    end_key: &Key,
+) -> Option<Key> {
+    // The start key from TiDB is actually a range, which is used to limit the
+    // upper bound of this flashback when scanning data, so is not a real key.
```
JmPotato, line 241: Try not directly mention the term `TiDB` inside TiKV as much as possible.

```suggestion
    // The start key from the client is actually a range, which is used to limit the
    // upper bound of this flashback when scanning data, so it may not be a real key.
```

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -225,6 +232,25 @@ pub fn commit_flashback_key(
...
+    let (keys, _) = reader
+        .scan_latest_user_keys(Some(start_key), Some(end_key), |_, _| true, 1)
+        .unwrap();
+    let key_to_lock = if !keys.is_empty() {
+        keys[0].clone()
+    } else {
+        // If there is no key in user keys, we should return directly.
+        return None;
+    };
+    Some(key_to_lock)
```
JmPotato, line 251:
```suggestion
    keys.pop()
```

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -133,7 +135,7 @@ impl<S: Snapshot> ReadCommand<S> for FlashbackToVersionReadPhase {
                 if key_locks.is_empty() {
                     // No more locks to rollback, continue to the prewrite phase.
                     FlashbackToVersionState::Prewrite {
-                        key_to_lock: self.start_key.clone(),
+                        key_to_lock: start_key.clone(),
```
JmPotato, line 138: This is not the exact key to lock anymore, maybe should rename it to `start_key`.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -225,6 +232,25 @@ pub fn commit_flashback_key(
...
+pub fn get_first_user_key(
+    reader: &mut MvccReader<impl Snapshot>,
+    start_key: &Key,
+    end_key: &Key,
+) -> Option<Key> {
+    // The start key from TiDB is actually a range, which is used to limit the
+    // upper bound of this flashback when scanning data, so is not a real key.
+    let (keys, _) = reader
+        .scan_latest_user_keys(Some(start_key), Some(end_key), |_, _| true, 1)
+        .unwrap();
```
JmPotato, line 244: Make `get_first_user_key` return `TxnResult<Option<Key>>` and use `?` here. Do not unwrap directly.

HuSharp:
Address all comments, PTAL!!! @JmPotato @sticnarf

src/storage/txn/commands/flashback_to_version.rs
HuSharp: Put them both in the reader to solve this problem.
for `process_write` it will be written in drop.
for `process_read`it will be handled manually.

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -490,34 +514,71 @@ pub mod tests {
...
+        let mut engine = TestEngineBuilder::new().build().unwrap();
+        let mut ts = TimeStamp::zero();
+        let (k, v1, v2) = (b"k", b"v1", b"v2");
+        must_prewrite_put(&mut engine, k, v1, k, *ts.incr());
+        must_commit(&mut engine, k, ts, *ts.incr());
+        must_prewrite_put(&mut engine, k, v2, k, *ts.incr());
+        must_commit(&mut engine, k, ts, *ts.incr());
+        must_get(&mut engine, k, ts, v2);
+
+        let prewrite_key = b"flashback_2pc";
```
JmPotato, line 552: I suggest using a more intuitive key like "j" to indicate it's before the "k".

```suggestion
        let start_key = b"j";
```

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -490,34 +514,71 @@ pub mod tests {
...
+        let (flashback_start_ts, flashback_commit_ts) = (*ts.incr(), *ts.incr());
+        // Rollback nothing.
+        assert_eq!(must_rollback_lock(&mut engine, k, flashback_start_ts), 0);
+        // Mock TiDB put a special key which is before b"k".
```
JmPotato, line 556:
```suggestion
        // Prewrite "k" not "start_key".
```

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -122,18 +123,16 @@ impl<S: Snapshot> ReadCommand<S> for FlashbackToVersionReadPhase {
     fn process_read(self, snapshot: S, statistics: &mut Statistics) -> Result<ProcessResult> {
         let tag = self.tag().get_str();
         let mut reader = MvccReader::new_with_ctx(snapshot, Some(ScanMode::Forward), &self.ctx);
+        reader.statistics = statistics.clone();
```
JmPotato, line 126: This is unnecessary.

src/storage/txn/commands/flashback_to_version.rs
```diff
@@ -71,27 +71,25 @@ impl CommandExt for FlashbackToVersion {
...
-            SnapshotReader::new_with_ctx(self.version, snapshot, &self.ctx),
-            context.statistics,
-        );
+        let mut reader =
+            MvccReader::new_with_ctx(snapshot.clone(), Some(ScanMode::Forward), &self.ctx);
```
JmPotato, line 75: Where is 'statistics.add(&reader.statistics)'?

src/storage/txn/commands/flashback_to_version_read_phase.rs
```diff
@@ -147,31 +146,45 @@ impl<S: Snapshot> ReadCommand<S> for FlashbackToVersionReadPhase {
...
+                    let key_to_lock = if let Some(first_key) =
+                        get_first_user_key(&mut reader, &self.start_key, &self.end_key)?
+                    {
+                        first_key
+                    } else {
+                        statistics.add(&reader.statistics);
+                        return Ok(ProcessResult::Res);
+                    };
+                    next_write_key = key_to_lock.clone();
+                    start_key = key_to_lock;
```
JmPotato, line 169: This could eliminate the `clone`.

```suggestion
                    let next_write_key = if let Some(first_key) =
                        get_first_user_key(&mut reader, &self.start_key, &self.end_key)?
                    {
                        first_key
                    } else {
                        statistics.add(&reader.statistics);
                        return Ok(ProcessResult::Res);
                    };
                    start_key = key_to_lock;
```

HuSharp:
Thx for your advice, I learned a lot from it! Had address all comments, PTAL! @JmPotato

src/storage/txn/actions/flashback_to_version.rs
```diff
@@ -490,34 +515,101 @@ pub mod tests {
...
+        assert_eq!(
+            must_commit_flashback_key(
+                &mut engine,
+                start_key,
+                flashback_start_ts,
+                flashback_commit_ts
+            ),
+            2
+        );
+        must_get(&mut engine, k, ts, v1);
```
JmPotato, line 602: Check the prewrite key also.

JmPotato (approved):

src/storage/txn/commands/flashback_to_version.rs
```diff
@@ -48,7 +48,7 @@ impl CommandExt for FlashbackToVersion {
             FlashbackToVersionState::RollbackLock { key_locks, .. } => {
                 latch::Lock::new(key_locks.iter().map(|(key, _)| key))
             }
-            FlashbackToVersionState::Prewrite { key_to_lock } => latch::Lock::new([key_to_lock]),
+            FlashbackToVersionState::Prewrite { start_key } => latch::Lock::new([start_key]),
```
sticnarf, line 51: Emm, the key locked in the latch is different from the actual written key...

As there cannot be any other valid write requests, it may be safe... But is it possible to make it correct by reading the key to lock in the read phase? I will be less worried :)

HuSharp: Thx for ur idea! Had to address it, PTAL!

sticnarf (approved):
LGTM

PR #13869, opened 2022-12-01
modify raft gc log impl for witness

ethercflow:
ref tikv/tikv#12876

Issue Number: Ref #12876 

What's Changed:

```
modify raft gc log impl for witness
```

### Check List 

Tests 

- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


components/raftstore/src/store/peer.rs
```diff
@@ -939,6 +939,10 @@ where
     /// The index of last compacted raft log. It is used for the next compact
     /// log task.
     pub last_compacted_idx: u64,
+    /// Record the time of the last compaction, the witness should query the
```
Connor1996, line 942: `compaction` is confusing, how about `raft log compact`

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2976,6 +3035,30 @@ where
             ));
         }
 
+        let voter_replicated_index = req.get_compact_log().get_voter_replicated_index();
+        if self.peer.is_witness && voter_replicated_index < compact_index {
```
Connor1996, line 3039: add comment here

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3942,6 +4034,27 @@ where
         cb.invoke_read(resp);
     }
 
+    fn handle_compact_raft_log(&mut self, ctx: &mut ApplyContext<EK>, voter_replicated_index: u64) {
```
Connor1996, line 4037: check_pending_compact_log

Connor1996 (commented):
rest LGTM

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2665,6 +2672,42 @@ where
...
+        );
+    }
+
+    fn on_voter_replicated_index_response(&mut self, msg: &ExtraMessage) {
+        if self.fsm.peer.is_leader() || !self.fsm.peer.is_witness() {
+            return;
+        }
+        self.ctx.apply_router.schedule_task(
+            self.region_id(),
+            ApplyTask::CheckCompact(self.region_id(), msg.voter_replicated_index),
```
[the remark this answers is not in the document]
ethercflow: It is not processed in peer, mainly because there is `PendingCmdQueue` in `ApplyDelegate`, so I think it is more uniform to put `compact cmd` in it. And the cycle triggered by this message is that the `compact cmd` has not been triggered for a long time, so I think it won't affect performance.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -190,6 +188,28 @@ impl<C> PendingCmdQueue<C> {
...
+
+    fn push_compact(&mut self, cmd: PendingCmd<C>) {
+        self.compacts.push_front(cmd);
+    }
+
+    fn pop_compact(&mut self, index: u64) -> Option<PendingCmd<C>> {
+        let mut pos: usize = 0;
+        let mut found = false;
+
+        for (i, c) in self.compacts.iter().enumerate() {
```
BusyJay, line 200: You mean `self.compacts.iter().position(|c| c.index < index)`?

components/raftstore/src/store/fsm/store.rs
```diff
@@ -561,6 +561,9 @@ where
             self.cfg.check_long_uncommitted_interval.0;
         self.tick_batch[PeerTick::CheckPeersAvailability as usize].wait_duration =
             self.cfg.check_peers_availability_interval.0;
+        // TODO: make it reasonable
+        self.tick_batch[PeerTick::RequestVoterReplicatedIndex as usize].wait_duration =
```
BusyJay, line 565: The index only makes sense when there are pending log gc. The tick can be register on demand.

components/raftstore/src/store/fsm/apply.rs
ethercflow: Fixed, thanks!

components/raftstore/src/store/fsm/store.rs
ethercflow: BTW,  should the Pending Compact cmd need to be persisted? eg: After several cmds were pending, a reboot occurred.

BusyJay: Technically it should. But it can be avoided by utilizing `RequestVoterReplicatedIndex`. Leader may return its first index as well as replicated index, so witness can gc logs even after being restarted.

Connor1996 (approved):
LGTM

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -190,6 +188,24 @@ impl<C> PendingCmdQueue<C> {
     fn set_conf_change(&mut self, cmd: PendingCmd<C>) {
         self.conf_change = Some(cmd);
     }
+
+    fn push_compact(&mut self, cmd: PendingCmd<C>) {
+        self.compacts.push_front(cmd);
```
BusyJay, line 193: We usually use `push_back` and `pop_front`.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -190,6 +188,24 @@ impl<C> PendingCmdQueue<C> {
...
+    }
+
+    fn pop_compact(&mut self, index: u64) -> Option<PendingCmd<C>> {
+        match self.compacts.iter().position(|c| c.index < index) {
```
BusyJay, line 197: You mean
```
let mut front = None;
while self.compacts.front().map_or(false, |c| c.index < index) {
    front = self.compacts.pop_front();
}
front
```

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -111,12 +111,8 @@ pub struct PendingCmd<C> {
 }
 
 impl<C> PendingCmd<C> {
-    fn new(index: u64, term: u64, cb: C) -> PendingCmd<C> {
-        PendingCmd {
-            index,
-            term,
-            cb: Some(cb),
-        }
+    fn new(index: u64, term: u64, cb: Option<C>) -> PendingCmd<C> {
```
BusyJay, line 114: Callback itself can be `Callback::None`, no need to make it an option.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2937,13 +2955,53 @@ where
...
+            return Ok(None);
+        }
+
+        match self.pending_cmds.pop_compact(voter_replicated_index) {
+            Some(cmd) => {
+                // compact failure is safe to be omitted, no need to assert.
+                compact_raft_log(&self.tag, &mut self.apply_state, cmd.index, cmd.term)?;
+                PEER_ADMIN_CMD_COUNTER.compact.success.inc();
+                Ok(Some(TaskRes::Compact {
+                    state: self.apply_state.get_truncated_state().clone(),
```
BusyJay, line 2981: When will `apply_state` be persisted?

components/raftstore/src/store/fsm/apply.rs
ethercflow: It may panic in `drop` as `Callback::None` is an `enum`, so I made it an `option`.

``` rust
impl<C> Drop for PendingCmd<C> {
    fn drop(&mut self) {
        if self.cb.is_some() {
            safe_panic!(
                "callback of pending command at [index: {}, term: {}] is leak",
                self.index,
                self.term
            );
        }
    }
}
```

components/raftstore/src/store/fsm/apply.rs
ethercflow: missed :(                 I plan to persist before notifying the store thread

components/raftstore/src/store/fsm/apply.rs
ethercflow: Yes, thanks!

components/raftstore/src/store/fsm/apply.rs
BusyJay: That's because it's not cleared just like other commands in `PendingCmd` before being dropped.

ethercflow: Oh, I see, should drain in delegate.destroy.

ethercflow: @BusyJay Addressed, PTAL again, thanks!

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2981,6 +3045,36 @@ where
...
+        let voter_replicated_index = req.get_compact_log().get_voter_replicated_index();
+        // If there is any voter lagging behind, the log truncation of the witness
+        // shouldn't be triggered even if it's force mode(raft log size/count exceeds
+        // the threshold or raft engine purge), otherwise the witness can't help the
+        // lagging voter catch up logs when leader is down. In this situation Compact
+        // index should be queued. If witness receives a voter_replicated_index
+        // that is larger than the pending compact index, logs can be deleted.
+        if self.peer.is_witness && voter_replicated_index < compact_index {
+            match self.pending_cmds.pop_compact(voter_replicated_index) {
+                Some(cmd) => {
```
BusyJay, line 3057: The current compact task should be pushed in queue even there are pending compacts. A case seems to be missing.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -3947,6 +4054,29 @@ where
...
+    fn check_pending_compact_log(
+        &mut self,
+        ctx: &mut ApplyContext<EK>,
+        voter_replicated_index: u64,
+    ) {
+        let res = self.delegate.try_compact_log(voter_replicated_index);
+        match res {
+            Ok(res) => {
+                if let Some(res) = res {
+                    self.delegate.write_apply_state(ctx.kv_wb_mut());
```
BusyJay, line 4066: Should reuse `prepare_for` and `commit_opt`.

components/raftstore/src/store/peer.rs
```diff
@@ -939,6 +939,15 @@ where
     /// The index of last compacted raft log. It is used for the next compact
     /// log task.
     pub last_compacted_idx: u64,
+    /// Record the time of the last raft log compact, the witness should query
+    /// the leader periodically whether `voter_replicated_index` is updated
+    /// if CompactLog admin command isn't triggered for a while.
+    pub last_compacted_time: Instant,
+    /// When the peer is witness, and there is any voter lagging behind, the
+    /// log truncation of the witness shouldn't be triggered even if it's
+    /// force mode, and this item will be set to `true`, after all pending
```
tonyxuqqi, line 948: I'm wondering in case a TiKV is down for 30 minutes, how big is the raft log accumulated?

components/raftstore/src/store/fsm/peer.rs
```diff
@@ -2665,6 +2674,42 @@ where
...
```
tonyxuqqi, line 2709: Question: Why do we need to explicitly call CheckCompact other than leverage raft log gc? In raft log gc we cannot delete some raft log if there's lag peer. But here if the lag peer's has made some progress, then we still can gc some logs. So I don't understand here why we need to call CheckCompact.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2981,6 +3045,36 @@ where
             ));
         }
 
+        let voter_replicated_index = req.get_compact_log().get_voter_replicated_index();
+        // If there is any voter lagging behind, the log truncation of the witness
+        // shouldn't be triggered even if it's force mode(raft log size/count exceeds
+        // the threshold or raft engine purge), otherwise the witness can't help the
+        // lagging voter catch up logs when leader is down. In this situation Compact
+        // index should be queued. If witness receives a voter_replicated_index
+        // that is larger than the pending compact index, logs can be deleted.
```
tonyxuqqi, line 3054: Question: why don't we change the logic in on_raft_gc_log_tick to make sure the correct compact_index is used under witness & lag peer case. Other than patching the logic here.

components/raftstore/src/store/fsm/peer.rs
ethercflow: `CheckCompact` is only used by witness. The leader still sends the compact command to all nodes as before, non-witnesses will be normal gc, only witness will judge whether to delay gc according to voter_replicated_index.

components/raftstore/src/store/fsm/apply.rs
ethercflow: Adjusting compact_index may cause some problems, such as not being able to GC in time and causing OOM, etc.

components/raftstore/src/store/peer.rs
ethercflow: I don't know yet, I need to actually test it.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -2981,6 +3045,36 @@ where
...
+        if self.peer.is_witness && voter_replicated_index < compact_index {
```
BusyJay, line 3055: What if `voter_replicated_index >= compact_index && self.pending_cmds.has_compact()`?

ethercflow: should drain

BusyJay (commented):
If witness is restarted, how to know there is pending raft log?

Rest LGTM

tonyxuqqi (approved):

ethercflow:
> If witness is restarted, how to know there is pending raft log?

Register the pull `voter_replicated_index` method in `Peer's start` method, when the apply thread detects that `has_compact` returns `false`, it is considered a reboot scenario, will do compaction using `voter_replicated_index` (leader's `first index`) as `compact_idx`.

components/raftstore/src/store/fsm/apply.rs
ethercflow: > That's because it's not cleared just like other commands in `PendingCmd` before being dropped.

@BusyJay  I found It is still possible to panic, such as calling `stop_node` in test code, it'll call `batch_system.shutdown` without call `delegate.destroy`.

components/raftstore/src/store/fsm/apply.rs
```diff
@@ -120,17 +120,17 @@ impl<C> PendingCmd<C> {
...
-        if self.cb.is_some() {
-            safe_panic!(
-                "callback of pending command at [index: {}, term: {}] is leak",
-                self.index,
-                self.term
-            );
-        }
-    }
-}
+// impl<C> Drop for PendingCmd<C> {
```
Connor1996, line 123: clean it?

ethercflow: Fixed, thanks! （It was not cleaned up in the `ApplyFsm drop` before, which caused a panic when `stop_node` occurred, so I commented it out first.）

PR #13881, opened 2022-12-05
storage: Resolve EBS flashback unlimit range

### LINKED ISSUE #13879: Flashback may miss the last region

https://github.com/tikv/tikv/commit/4b4fc4390d90e72eb87e49a223e756e6ba1e2688

### What happened?

`end_key` in a flashback request will be encoded directly without checking whether it's empty.

https://github.com/tikv/tikv/blob/4b4fc4390d90e72eb87e49a223e756e6ba1e2688/src/storage/txn/commands/mod.rs#L369-L379

Inside the flashback progress, if the `end_key` is empty, it will be less than all other keys, which will cause `scan_latest_user_keys` to return immediately without flashbacking any keys.

https://github.com/tikv/tikv/blob/4b4fc4390d90e72eb87e49a223e756e6ba1e2688/src/storage/mvcc/reader/reader.rs#L611-L616

Luckily, this won't be a problem for the Flashback SQL since it will use the table prefix key as the `end_key`.

https://github.com/pingcap/tidb/blob/0b1096eac5a500f8c624f08f384d0194da5386f4/ddl/cluster.go#L255-L258

HuSharp:

Issue Number: Close #13879 

What's Changed:

```commit-message
Resolve EBS flashback unlimit range.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -4875,7 +4875,7 @@ mod tests {
                 commit_ts,
                 version,
                 key.clone(),
-                Key::from_raw_maybe_unbounded(b"z"),
+                Some(Key::from_raw(b"z")),
             );
             if let Mutation::Put(..) = write.0 {
                 expect_value(
@@ -4997,7 +4997,7 @@ mod tests {
             commit_ts,
             2.into(),
             Key::from_raw(b"k"),
-            Key::from_raw_maybe_unbounded(b"z"),
+            Some(Key::from_raw(b"z")),
         );
         expect_value(
             b"v@1".to_vec(),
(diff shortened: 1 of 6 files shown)


src/storage/mod.rs
```diff
@@ -4875,7 +4875,7 @@ mod tests {
                 commit_ts,
                 version,
                 key.clone(),
-                Key::from_raw(b"z"),
+                Key::from_raw_maybe_unbounded(b"z"),
```
JmPotato, line 4878: What about passing `Some(Key::from_raw(b"z"))` in tests since we can make sure that it's not empty?

JmPotato (commented):
The rest LGTM.

src/storage/mod.rs
HuSharp: Okay, get it.

sticnarf (approved):

JmPotato (approved):

HuSharp:
@sticnarf can u merge this pr? Thx!!!

PR #13885, opened 2022-12-05
log-backup: make the safepoint lifetime 2hours

YuJuncen:
Signed-off-by: hillium <yu745514916@live.com>

Issue Number: Close https://github.com/pingcap/tidb/issues/39603

What's Changed:
This PR make the lifetime of service points that created by PITR 2 hours, which is longer than `FLUSH_INTERVAL * FLUSH_MAX_RETRY_TIME`

```commit-message
```

### Related changes

### Check List 

Tests 

- (Almost) No code

### Release note 

```release-note
None
```

### DIFF
--- a/components/backup-stream/src/checkpoint_manager.rs
+++ b/components/backup-stream/src/checkpoint_manager.rs
@@ -356,9 +356,13 @@ impl<PD: PdClient + 'static> FlushObserver for BasicFlushObserver<PD> {
             .update_service_safe_point(
                 format!("backup-stream-{}-{}", task, self.store_id),
                 TimeStamp::new(rts.saturating_sub(1)),
-                // Add a service safe point for 30 mins (6x the default flush interval).
-                // It would probably be safe.
-                Duration::from_secs(1800),
+                // Add a service safe point for 2 hours (40x the default flush interval).
+                // In some extreme scenario (for example, the external storage is unavailable),
+                // we may keep fail for flushing, we should make sure the safepoint present before
+                // flush errors expires retry count.
+                // TODO: We'd better make the coordinator, who really calculates the checkpoint to
+                // register service safepoint.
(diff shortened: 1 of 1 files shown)


Leavrth (approved):

components/backup-stream/src/checkpoint_manager.rs
```diff
@@ -356,9 +356,13 @@ impl<PD: PdClient + 'static> FlushObserver for BasicFlushObserver<PD> {
...
-                // Add a service safe point for 30 mins (6x the default flush interval).
-                // It would probably be safe.
-                Duration::from_secs(1800),
+                // Add a service safe point for 2 hours (40x the default flush interval).
+                // In some extreme scenario (for example, the external storage is unavailable),
+                // we may keep fail for flushing, we should make sure the safepoint present before
+                // flush errors expires retry count.
+                // TODO: We'd better make the coordinator, who really calculates the checkpoint to
+                // register service safepoint.
+                Duration::from_secs(60 * 60 * 2),
```
Leavrth, line 365: if need to large than `FLUSH_INTERVAL * FLUSH_MAX_RETRY_TIME`, it's better to use some expression contain `FLUSH_INTERVAL * FLUSH_MAX_RETRY_TIME` instead of `60*60*2`.

YuJuncen: We cannot trivially get `flush_interval` in this context. If we want to do so, we must pass a `config` into here (which would make things even more complex.) As a temporary solution, I think hard-coding here would be fine.

Leavrth: get

3pointer (approved):

overvenus (commented):
Why not add to config？

overvenus (approved):

PR #13886, opened 2022-12-05
raftstore-v2: use snapshot to initialize split

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
Create a new storage introduces unnecessary complexity and corner cases.
As split is an initialization just like snapshot, this PR reuses snapshot to make
the process a lot simpler and more robust.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/raftstore-v2/src/operation/command/admin/split.rs
+++ b/components/raftstore-v2/src/operation/command/admin/split.rs
@@ -27,6 +27,7 @@
 
 use std::collections::VecDeque;
 
+use collections::HashSet;
 use crossbeam::channel::{SendError, TrySendError};
 use engine_traits::{
     Checkpointer, DeleteStrategy, KvEngine, OpenOptions, RaftEngine, RaftLogBatch, Range,
@@ -37,17 +38,18 @@ use keys::enc_end_key;
 use kvproto::{
     metapb::{self, Region, RegionEpoch},
     raft_cmdpb::{AdminRequest, AdminResponse, RaftCmdRequest, SplitRequest},
-    raft_serverpb::RegionLocalState,
+    raft_serverpb::{RaftMessage, RaftSnapshotData, RegionLocalState},
 };
 use protobuf::Message;
-use raft::RawNode;
+use raft::{prelude::Snapshot, RawNode, INVALID_ID};
 use raftstore::{
     coprocessor::RegionChangeReason,
     store::{
         fsm::apply::validate_batch_split,
         metrics::PEER_ADMIN_CMD_COUNTER,
+        snap::TABLET_SNAPSHOT_VERSION,
         util::{self, KeysInfoFormatter},
-        PeerPessimisticLocks, PeerStat, ProposalContext, RAFT_INIT_LOG_INDEX,
+        PeerPessimisticLocks, PeerStat, ProposalContext, RAFT_INIT_LOG_INDEX, RAFT_INIT_LOG_TERM,
     },
     Result,
 };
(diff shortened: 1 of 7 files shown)


BusyJay:
PTAL

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -381,8 +396,20 @@ impl<EK: KvEngine, ER: RaftEngine> Storage<EK, ER> {
         self.entry_storage_mut().set_truncated_term(last_term);
         self.entry_storage_mut().set_last_term(last_term);
 
-        let key = TabletSnapKey::new(region_id, peer_id, last_term, last_index);
-        let mut path = snap_mgr.final_recv_path(&key);
+        let (path, clean_split) = match self.split_init_mut() {
```
tonyxuqqi, line 399: Q: is it possible both apply snapshot and split is on-going? Then here we use split first. The snapshot will be ignored and the disk space could be leaked forever.

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -353,108 +381,96 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+            if ids.remove(&region_id) {
+                found = true;
+                break;
+            }
+        }
+        assert!(found, "{:?} {}", self.logger.list(), region_id);
+        let split_trace = self.split_trace_mut();
+        if !split_trace.first().unwrap().1.is_empty() {
+            return;
+        }
```
tonyxuqqi, line 469: why we check the split_trace.first() instead of the tablet_index's one.

components/raftstore-v2/src/operation/ready/snapshot.rs
BusyJay: Yes it's possible. The snapshot is not leaked here, but the very moment in `step`. The cleanup should be done there.

components/raftstore-v2/src/operation/command/admin/split.rs
BusyJay: Because apply index should be advanced in order. In the final version, the index of split trace will be persisted one by one.

components/raftstore-v2/src/operation/ready/snapshot.rs
tabokie: Where exactly? I thought currently the tablet created in `recv_snap` is assumed to be always loaded.

BusyJay: https://github.com/tikv/tikv/blob/280d53b75c1d3e821435c440eb6493af536edabb/components/raftstore-v2/src/operation/ready/mod.rs#L66-L128

There is no such guarantee, a snapshot can be discarded for various reasons, conflict with split is just one of them.

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -132,9 +132,24 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+                meta.tablet_caches.insert(region_id, self.tablet().clone());
+                meta.readers
+                    .insert(region_id, self.generate_read_delegate());
+                meta.region_read_progress
+                    .insert(region_id, self.read_progress().clone());
+            }
             self.read_progress_mut()
                 .update_applied_core(persisted_index);
-            info!(self.logger, "apply tablet snapshot completely");
+            match self.storage_mut().split_init_mut().take() {
```
tabokie, line 145: No need to check init.scheduled?

components/raftstore-v2/src/operation/command/admin/split.rs
```diff
@@ -353,108 +381,104 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
         }
+        let _ = store_ctx
+            .router
+            .force_send(split_init.source_id, PeerMsg::SplitInitFinish(region_id));
+    }
 
-        self.schedule_apply_fsm(store_ctx);
+    pub fn on_split_init_finish<T>(&mut self, ctx: &mut StoreContext<EK, ER, T>, region_id: u64) {
+        let mut found = false;
+        for (tablet_index, ids) in self.split_trace_mut() {
```
tabokie, line 459: But `HashSet` is not ordered.

components/raftstore-v2/src/operation/ready/snapshot.rs
tabokie: I mean where is it physically deleted? If it should be done by tablet GC then why clean up the split tablet here during apply?

BusyJay: It's not deleted yet. It's one of the known leak problem and should be fixed in another PR.

components/raftstore-v2/src/operation/ready/snapshot.rs
BusyJay: No, `scheduled` just mean the message is sent to peer. No matter whether it's scheduled or not, `Some` means split has happened and need to be acked.

components/raftstore-v2/src/operation/command/admin/split.rs
BusyJay: `split_trace` is `Vec`, `ids` is `HashSet`.

components/raftstore-v2/src/operation/ready/snapshot.rs
BusyJay: The assertion should be removed. It's OK the snapshot is not created by split. As long as the peer is initialized, it can report to the leader split is finished.

tabokie (approved):

tonyxuqqi (approved):

PR #13890, opened 2022-12-05
raftstore-v2: correct store sizing for v2

tabokie:

Issue Number: Ref #12842 

What's Changed:

Identify hardlinks and apply deduplication before summing the store size.

```commit-message
```

### Check List 

Tests
- Unit test

### Release note 

```release-note
None
```


components/file_system/src/lib.rs
```diff
@@ -335,6 +338,78 @@ pub fn sync_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
...
+    }
+
+    fn size_imp(&mut self, path: &Path) -> io::Result<usize> {
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::MetadataExt;
+            let mut size = 0;
+            let dev = path.metadata()?.dev();
+            walk_dir(path.read_dir()?, &mut |f| {
+                if let Ok(meta) = f.metadata() {
```
BusyJay, line 394: I'm afraid this may be too expansive as pd heartbeat is quite frequent.

components/raftstore-v2/src/worker/pd/store_heartbeat.rs
```diff
@@ -259,7 +260,7 @@ where
     }
 
     /// Returns (capacity, used, available).
-    fn collect_engine_size(&self) -> Option<(u64, u64, u64)> {
+    fn collect_engine_size(&mut self) -> Option<(u64, u64, u64)> {
```
tonyxuqqi, line 263: needs to consider the capacity config as v1 did.

components/file_system/src/lib.rs
tonyxuqqi: it's tested in demo version and seems fine. @BusyJay 
However, assuming we delete the obsolete tablet in-time, do we still need this? @tabokie

BusyJay: The reason why hardlinks need to be excluded is there may be many files are essential the same. But if we can do cleanup in time, then the there should be not much same files, hence no need to solve hardlink.

Issue #13889, opened 2022-12-05
log-backup: make service GC safepoints live 24 hours


YuJuncen:
## Development Task
For now, "normal" service GC safepoints only live for 30 mins, which is too short for troubleshooting. We'd better make it longer, even that may block GC when task stop or scale-in.

PR #13893, opened 2022-12-06
log-backup: remove GC safepoint after task stops

YuJuncen:

Issue Number: Ref #13889 

What's Changed:

This PR removes GC safepoint after log task stops, so after the task stopped, the GC won't be stuck.

```commit-message
```

### Related changes

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


overvenus (commented):
What if a TiKV is tombstoned? Will it block GC for 24 hours?

PR #13900, opened 2022-12-07
raft_engine: remove redundant methods

BusyJay:

Issue Number: Ref #12842 

What's Changed:
```commit-message
These methods are defined in log batch already, redefine them again
introduce unnecessary maintenance. And these methods are also
confusing as they don't sync by default, which is very easy to make
mistake.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


tabokie (approved):

ethercflow (approved):
LGTM

PR #13901, opened 2022-12-07
raftstore-v2: implement a simplified version of CoprocessorHost

SpadeA-Tang:

Issue Number: Ref #12842

What's Changed:

```
A simplified version of CoprocessorHost in v1, it conveys information to LockManager.
```

### Related changes

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


tonyxuqqi (approved):

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -641,6 +660,34 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
             .advance_apply(apply_index, term, region);
     }
 
+    /// Register self to apply_scheduler so that the peer is then usable.
+    /// Also trigger `RegionChangeEvent::Create` here.
+    pub fn activate<T>(&mut self, ctx: &mut StoreContext<EK, ER, T>) {
```
tabokie, line 665: Why create a function for this, I see it's only called once.

components/raftstore-v2/src/raft/peer.rs
```diff
@@ -641,6 +660,34 @@ impl<EK: KvEngine, ER: RaftEngine> Peer<EK, ER> {
...
+    }
+
+    #[inline]
+    pub fn set_lead_transferee(&mut self, lead_transferee: u64) {
+        self.lead_transferee = lead_transferee;
+    }
+
+    /// Update states of the peer which can be changed in the previous raft
+    /// tick.
+    pub fn post_raft_group_tick(&mut self) {
```
tabokie, line 687: These two can be merged into `refresh_lead_transferee()`

src/server/lock_manager/mod.rs
```diff
@@ -205,6 +205,11 @@ impl LockManager {
         role_change_notifier.register(host);
     }
 
+    /// Creates a `RoleChangeNotifier` of the deadlock detector worker
+    pub fn generate_notifier(&self) -> RoleChangeNotifier {
```
tabokie, line 209:
```suggestion
    pub fn new_notifier(&self) -> RoleChangeNotifier {
```

src/server/raftkv2/node.rs
```diff
@@ -85,6 +85,7 @@ where
         snap_mgr: TabletSnapManager,
         concurrency_manager: ConcurrencyManager,
         causal_ts_provider: Option<Arc<CausalTsProviderImpl>>, // used for rawkv apiv2
+        lock_manager_observer: Arc<dyn LockManagerNotifier>,
```
tabokie, line 88: Use Option.

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: After the change of split init, yes, it is.

src/server/raftkv2/node.rs
SpadeA-Tang: Why, we can use a dummy notifier if it is not needed (in test for example).

components/raftstore-v2/src/raft/peer.rs
SpadeA-Tang: Done

src/server/raftkv2/node.rs
tabokie: Clearly None is better than dummy?

tabokie (approved):

src/server/raftkv2/node.rs
tabokie: Anyway, if previously coprocessor host is not an option, I'm okay with dummy too.

PR #13902, opened 2022-12-07
storage: Update lock wait info after acquiring locks

MyonKeminta:

Issue Number: Ref #13298

What's Changed:

```commit-message
Update lock wait info after acquiring locks, so that in case there is resumable pessimistic lock requests that's waiting in TiKV, the latest waiting relationship can be used to provide the diagnostic information and do deadlock detection.
```

### Related changes

\-

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
None
```


MyonKeminta:
Wait. There's still regression in bank-50m workload after rerunning. I'll investigate it.

MyonKeminta:
@sticnarf @cfzjywxk PTAL, thanks

src/storage/lock_manager/lock_waiting_queue.rs
```diff
@@ -1205,4 +1235,44 @@ mod tests {
...
+        lock_info.set_lock_for_update_ts(10);
+        let lock_info = vec![lock_info];
+
+        b.iter(|| {
+            queues.update_lock_wait(lock_info.clone());
+        });
+    }
+
+    #[bench]
+    fn bench_update_lock_wait_queue_len_512(b: &mut test::Bencher) {
```
cfzjywxk, line 1258: Any performance difference for `lockWaitQueues.update_lock_wait` if the queue size is large?

src/storage/txn/scheduler.rs
```diff
@@ -978,6 +979,28 @@ impl<E: Engine, L: LockManager> Scheduler<E, L> {
         resumable_wake_up_list
     }
 
+    fn on_acquired_locks_finished(&self, new_acquired_locks: Vec<kvrpcpb::LockInfo>) {
```
cfzjywxk, line 982: As it increases execution costs compared with the previous implementation, it might still impact performance if the server load is high. Maybe we could run such verifications in the benchbot cases.

src/storage/lock_manager/lock_waiting_queue.rs
MyonKeminta: The result was pasted in the above comment. 

```
test storage::lock_manager::lock_waiting_queue::tests::bench_update_lock_wait_empty         ... bench:         216 ns/iter (+/- 7)
test storage::lock_manager::lock_waiting_queue::tests::bench_update_lock_wait_queue_len_512 ... bench:     133,810 ns/iter (+/- 5,278)
```

src/storage/txn/actions/prewrite.rs
```diff
@@ -153,7 +153,10 @@ pub fn prewrite<S: Snapshot>(
         OldValue::Unspecified
     };
 
-    let final_min_commit_ts = mutation.write_lock(lock_status, txn)?;
+    let is_new_lock = (!matches!(pessimistic_action, DoPessimisticCheck) || lock_amended)
+        && !mutation.try_one_pc();
```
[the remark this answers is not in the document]
MyonKeminta: Ah, It seems that the condition `!mutation.try_one_pc()` should be removed. The `write_lock` function won't call `put_lock_for_1pc` in 1pc cases, where `is_new_lock` flag won't be used.

sticnarf:
Is it really necessary that all requests acquiring new locks update lock wait info? What if only the resumed acquiring pessimistic locks update the info? 🤔

MyonKeminta:
> Is it really necessary that all requests acquiring new locks update lock wait info? What if only the resumed acquiring pessimistic locks update the info? 🤔

Considering the case that a key has both resumable and non-resumable waiters and they are mixed in a single queue, there are some complicated possibilities. After waking up a pessimistic lock request that's in legacy mode, the key will be in an unlocked state even if there are requests still waiting in the queue. It's possible that the key is then locked by any unexpected transaction, which changes the waiting relationship unpredictably.

In the above situation, if the new lock is an optimistic lock (written by a prewrite request), it actually won't form a deadlock, but there's still other problem: if then some waiters exceed their timeout and reports `KeyIsLocked`<sup>[*]</sup> to the client with stale lock information, the client may then perform a useless resolve_lock. So, it would be better if they can have the up-to-date lock info.
 
<sup>\*: When waking up a legacy request, it tries to wake up other legacy requests in queue after `wake-up-delay-duration`, but 1) resumable requests and 2) legacy requests that is ordered behind a resumable request in the queue won't be affect, and they may continue waiting until timeout as normal. </sup>

To simplify it, I just try to let all lock-acquiring operations update the lock state to the lock waiting queue and see if the performance overhead is acceptable. This is actually not necessary if all waiting requests are in legacy mode, because once the lock is released, all remaining waiters will be woken up (returning `WriteConflict`) after waking up the first waiter, so that deadlock detection will be done when those transactions retry the statement and try to acquire the lock again, and there won't be the stale-lock-info issue. This can be done by adding a counter in `KeyLockWaitState` to count how many resumable waiters in the queue, but I didn't do it in this PR.

cfzjywxk:
@MyonKeminta 
Is the lock-view-related wait information or diagnose information guaranteed to be valid if all requests acquiring new locks update lock wait info?

MyonKeminta:
> @MyonKeminta Is the lock-view-related wait information or diagnose information guaranteed to be valid if all requests acquiring new locks update lock wait info?

Now it can reflect the change of the waiting relationship of a single pessimistic lock request. It would be more up-to-date than old version, but I think that information is still not perfect, since users may care more about the waiting relationship in TiDB's level. The absence of a waiter in TiKVs doesn't mean the transaction is not waiting for something (for example, a pessimistic lock request may timeout and the client will then send a retry request, which causes a short absence of the waiter in TiKV side).

cfzjywxk (approved):
LGTM

PR #13903, opened 2022-12-07
*: always share block cache

### LINKED ISSUE #12936: Remove the suppport for specifying block cache for cf manually
## Feature Request

For historical reason, we support specifying block cache for each cf for either raft db or kv db. But since we support sharing cache between CFs and dbs (3 years ago), these settings are not practical anymore. They are hard to use and easy to make mistake. We better remove those settings and always enabling share cache for simplicity.

BusyJay:
Issue Number: Close #12936

What's Changed:

There has been 5 months and nobody objects it so let's always enable share block cache.

```commit-message
```

### Related changes

    https://github.com/pingcap/docs-cn/pull/12241

### Check List 

Tests 

- Unit test
- Integration test

Side effects

    Use separate cache for each CF is not supported anymore.

### Release note 

```release-note
remove the support of specifying separate cache for each CF
```

### DIFF
--- a/src/storage/config.rs
+++ b/src/storage/config.rs
@@ -193,8 +193,6 @@ impl Default for FlowControlConfig {
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct BlockCacheConfig {
-    #[online_config(skip)]
-    pub shared: bool,
     pub capacity: Option<ReadableSize>,
     #[online_config(skip)]
     pub num_shard_bits: i32,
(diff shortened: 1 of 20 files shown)


src/storage/config.rs
```diff
@@ -193,8 +193,6 @@ impl Default for FlowControlConfig {
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct BlockCacheConfig {
-    #[online_config(skip)]
-    pub shared: bool,
```
tabokie, line 197: Option<bool> for compatibility?

zhangjinpeng1987 (approved):
LGTM

tabokie (approved):

PR #13904, opened 2022-12-07
gc_worker: disable gc if have negative ratio

### LINKED ISSUE #13909: gc_worker: Support disable GC temporarily

In some internal scenario, we might break the txn model temporarily, for example, during PITR restore, for better concurrency(and avoiding repeating sorting), we are going to randomly write stale records into TiKV. (It cloud be safe by some external constraints.)
If GC enabled during this time, it may corrupt the state and make the rocksDB eventually inconsistent. 

We need to be able disable GC fully via some config.
For simplicity, we can reuse the `gc.ratio-threshold` config, when it is negative or infinity(a better choice, however TOML doesn't support serializing it :( ), disable GC temporarily.

Perhaps we can make a RPC endpoint to allow us disable GC with TTL. However that is more complex.

### Teachability, Documentation, Adoption, Migration Strategy:
Disabling GC while `ratio_threshold` is negative is trivial. Just always return `false` once it is negative.
The flag `gc.ratio-threshold` is undocumented, hence I would prefer treating it as an internal switch, and add no more document here.
This is compatible with before, the only exception is when someone has already used negative `ratio_threshold` (seldomly, right?).
This flag can be changed online via `tikv-ctl`.

YuJuncen:
Signed-off-by: hillium <yu745514916@live.com>

Issue Number: Close #13909

What's Changed:
This PR allows GC to be skipped once the `ratio_threshold` is negative or infinity.

Disabling GC is useful in some abnormal scenarios where the transaction model would be break (e.g. writes with higher commit TS would be written BEFORE writes with lower commit TS, or write data with TS lower than current GC safe point).

Mainly, this is for fixing the data inconsistent found at https://github.com/pingcap/tidb/issues/39602: because PITR would apply unordered writes concurrently, once there are some delete records, GC might ripe them and make the write records (which haven't been restored yet) lives.

```commit-message
GC would be skipped once the `ratio_threshold` is negative or infinity.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
Now, if the config `gc.ratio_threshold` is less than `0`, GC would be disabled.
```

### DIFF
--- a/src/server/gc_worker/compaction_filter.rs
+++ b/src/server/gc_worker/compaction_filter.rs
@@ -918,8 +918,6 @@ pub mod test_utils {
 
 #[cfg(test)]
 pub mod tests {
-    use std::mem::MaybeUninit;
-
     use engine_traits::{DeleteStrategy, MiscExt, Peekable, Range, SyncMutable, CF_WRITE};
 
     use super::{test_utils::*, *};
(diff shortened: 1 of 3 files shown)


joccau (approved):
LGTM

3pointer (approved):

hicqu (approved):

PR #13912, opened 2022-12-08
*: clean up tablet factory

BusyJay:
Issue Number: Ref #12842 

What's Changed:

Notetable changes:
- TabletFactory is simplified to only provide 3 methods.
- Move CachedTablet to engine_trait.
- Provide `ConfigurableDb` trait to unify single db and tablets for configuration
- Add TabletRegistry to provide common access.

```commit-message
In the past, there are 5 types of tablet factory:
- Dummy factory
- Test Factory for v1 and v2
- Production factory for v1 and v2.

Tablet factory also requires TabletAccessor trait, so there is 10 implementations.
If there are bugs, we need to change at least 5 places, and 10 places in worst
cases. This PR simplifies the code by limiting the scope of engine factory, so it
only needs to provide create, destroy and exist check three functions. All other
functions are moved to type `TabletRegistry`. Also `TabletFactory` is only for v2
usage, so there will be only 3 different tablet factories and no accessor trait.
It should be a lot easier to adjust the behavior of tablet factory.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


src/config/configurable.rs
```diff
@@ -0,0 +1,141 @@
...
+            "Total count {}. Sample errors: {}",
+            error_count,
+            std::str::from_utf8(&error_samples).unwrap()
+        );
+    }
+    res
+}
+
+impl ConfigurableDb for TabletRegistry<RocksEngine> {
+    fn set_db_config(&self, opts: &[(&str, &str)]) -> ConfigRes {
```
tonyxuqqi, line 95: I think it will better to try to set all dbs even when there's one failed---as this error is not fatal and we still hope TiKV continue to serve. 
It will be easier to debug in the future. The failed db should be logged and db not in the log should mean success----of course we may still limit logging the max log count.

BusyJay: > it will better to try to set all dbs even when there's one failed

That's exact what it does now. The error behavior is exact the same as before, the only difference is now it's implemented using a single function instead of a struct.

src/server/engine_factory.rs
```diff
@@ -129,172 +105,125 @@ impl KvEngineFactory {
...
+impl TabletFactory<RocksEngine> for KvEngineFactory {
+    fn open_tablet(&self, id: u64, suffix: Option<u64>, path: &Path) -> Result<RocksEngine> {
+        let mut db_opts = self.db_opts();
+        let cf_opts = self.cf_opts();
+        if let Some(listener) = &self.inner.flow_listener && let Some(suffix) = suffix {
+            db_opts.add_event_listener(listener.clone_with(id, suffix));
         }
+        let kv_engine =
+            engine_rocks::util::new_engine_opt(path.to_str().unwrap(), db_opts, cf_opts);
+        if let Err(e) = &kv_engine {
```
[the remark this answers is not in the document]
BusyJay: I thought about it before. I think it's a nice-to-have. Creating a tablet does not race with current usage. To make things safe, not only creation should use rename, destroy should also use rename.

src/server/engine_factory.rs
```diff
@@ -129,172 +105,125 @@ impl KvEngineFactory {
...
+            error!("failed to create tablet"; "id" => id, "suffix" => ?suffix, "path" => %path.display(), "err" => ?e);
+        } else if let Some(listener) = &self.inner.flow_listener && let Some(suffix) = suffix {
+            listener.clone_with(id, suffix).on_created();
```
tonyxuqqi, line 161: on_created should be put in TabletRegistry. Otherwise the consumer (TabletFlowController) will try to get it from TabletRegistry, but failed because it may not be added to registry yet.

src/config/configurable.rs
tonyxuqqi: OK, I see. Thanks.

components/engine_traits/src/tablet.rs
```diff
@@ -0,0 +1,398 @@
...
+        let path = self.tablet_path(id, suffix);
+        if !create && !self.tablets.factory.exists(&path) {
+            return Err(Error::Other(box_err!(
+                "tablet ({}, {:?}) doesn't exist",
+                id,
+                suffix
+            )));
+        }
+        let tablet = self.tablets.factory.open_tablet(id, Some(suffix), &path)?;
+        let mut cached = self.get_or_default(id);
```
tabokie, line 221: What if `cached` already contains (id, suffix)?

src/storage/txn/flow_controller/tablet_flow_controller.rs
```diff
@@ -117,13 +117,7 @@ impl FlowInfoDispatcher {
...
 
                     let insert_limiter_and_checker = |region_id, suffix| -> FlowChecker<E> {
-                        let engine = tablet_factory
-                            .open_tablet(
-                                region_id,
-                                Some(suffix),
-                                OpenOptions::default().set_cache_only(true),
-                            )
-                            .unwrap();
+                        let engine = registry.get(region_id).unwrap().latest().unwrap().clone();
```
tabokie, line 120: Weak reference would be handy here. Maybe leave a todo.

components/engine_traits/src/tablet.rs
```diff
@@ -0,0 +1,398 @@
...
+            .entry(id)
+            .or_insert_with(|| CachedTablet::new(None))
+            .clone()
+    }
+
+    pub fn tablet_factory(&self) -> &dyn TabletFactory<EK> {
+        self.tablets.factory.as_ref()
+    }
+
+    pub fn remove(&self, id: u64) {
```
tabokie, line 201: Why not handle the whole deletion at this level? (instead of calling `remove && destroy_tablet`). In fact, exposing the factory in multi-rocks code looks bug-prone to me.

components/engine_traits/src/tablet.rs
BusyJay: Then it will return error at L220 as the tablet is locked.

components/engine_traits/src/tablet.rs
BusyJay: Because tablet can't be destroyed immediately. It should wait till no one is referencing the tablet. In complex case like split/merge, it should also wait till new tablet is persistent.

src/storage/txn/flow_controller/tablet_flow_controller.rs
BusyJay: I think this part needs to be refactored. Flow control should be bound to the lifetime of a peer instead of tablets. The refactor will be done later.

tonyxuqqi (approved):

tabokie (approved):

PR #13915, opened 2022-12-08
fix the bug that send mistake peer snapshot

bufferflies:
Signed-off-by: bufferflies <1045931706@qq.com>

Issue Number: Ref #12842

What's Changed:
It need to check the to peer id if the snapshot stat is generated

```commit-message
```

### Check List 

Tests 

- Unit test

Side effects

### Release note 

```release-note
None.

```


components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -217,9 +217,18 @@ impl<EK: KvEngine, ER: RaftEngine> Storage<EK, ER> {
                 }
             }
             SnapState::Generated(ref s) => {
+                if s.1 != to {
```
nolouch, line 220: maybe deconstruct first，like：
```
let (snap, to_peer) = s
```

nolouch (commented):
need tests.

BusyJay:
Need test case.

bufferflies:
> Need test case.

the unit test in storage.rs:
```
assert_eq!(s.snapshot(0, 8).unwrap_err(), unavailable);
let (snap, to_peer_id) = match *s.snap_state.borrow() {
```

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -217,9 +217,18 @@ impl<EK: KvEngine, ER: RaftEngine> Storage<EK, ER> {
...
+                    info!(self.logger(),
+                        "need previous peer to call the snapshot";
```
BusyJay, line 222: What if the previous peer doesn't need snapshot anymore?

bufferflies: It will cancel the snapshot if the previous peer doesn't the snapshot  anymore, so the `SnapState` will become Relax.

BusyJay: At which step?

components/raftstore-v2/src/operation/ready/snapshot.rs
```diff
@@ -318,28 +314,31 @@ impl<EK: KvEngine, ER: RaftEngine> Storage<EK, ER> {
...
+                    "compact_to" => compact_to,
+                );
+                states.remove(&id);
             }
+        } else {
+            self.snap_states.borrow_mut().clear();
         }
-        canceled.store(true, Ordering::SeqCst);
-        *snap_state = SnapState::Relax;
         self.gen_snap_task_mut().take();
```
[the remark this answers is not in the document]
bufferflies: I think the `gen_snap_task`  doesn't specify the `to_peer_id`, so it will be clear if the cancel has been called. It will generate one task in next snapshot.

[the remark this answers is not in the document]
bufferflies: The task will be taken no matter matched the to_peer_id, it will generate new task in next snapshot. The `gen_snap_task ` shouldn't be associated with one `to_peer_id`.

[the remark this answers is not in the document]
bufferflies: It will be canceled , I will verify the peer_id before cancel.

BusyJay (approved):

nolouch (approved):
lgtm

iosmanthus (approved):

PR #13917, opened 2022-12-08
raftstore-v2: add DATA_PREFIX

BusyJay:
Issue Number: Ref #12842 

What's Changed:
```commit-message
Because v2 doesn't share rocksdb, so it's possible to not write a prefix when
writing keys. However, because rocksdb doesn't support specifying infinite
upper bound in various APIs like properties, so we should solve those issues
before landing prefix-less write.
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```


tonyxuqqi (approved):

components/raftstore-v2/src/raft/apply.rs
```diff
@@ -169,4 +167,10 @@ impl<EK: KvEngine, R> Apply<EK, R> {
     pub fn take_admin_result(&mut self) -> Vec<AdminCmdResult> {
         mem::take(&mut self.admin_cmd_result)
     }
+
+    #[inline]
+    pub fn release_memory(&mut self) {
+        mem::take(&mut self.key_buffer);
+        self.write_batch.take();
```
tabokie, line 174: assert it's empty.

tabokie (approved):

PR #13923, opened 2022-12-09
copr: fix _ pattern in like behavior for old collation (#13785)

YangKeao:

Issue Number: Close https://github.com/tikv/tikv/issues/13769

This PR is a rework for https://github.com/tikv/tikv/pull/13770. As I have force-pushed the branch, github doesn't allow me to reopen the old one :cry: .

What's Changed:

If the two charset of arguments are equal, use the charset from arguments to decode the bytes.

I have considered the following cases:

| Ret Collation  | First Arg Charset | Second Arg Charset | Description |
| ------------- | ------------- |------------- |------------- |
| binary  | binary  | binary | It works normally |
| binary  | utf8mb4  | binary | It happened when TiDB didn't push down the collation information. Just decode and compare the string byte by byte. A `_` in pattern will only match one byte in the first arg |
| binary  | binary  | utf8mb4 | It happened when TiDB didn't push down the collation information. Just decode and compare the string byte by byte. A `_` in pattern will only match one byte in the first arg |
| binary  | utf8mb4  | utf8mb4 | It happened when TiDB didn't push down the collation information. It'll decode and compare the string character by character. A `_` in pattern will match one unicode character in the first arg. This PR is actually fixing this case |
| utf8mb4_*  | utf8mb4  | binary | will not happen |
| utf8mb4_*  | utf8mb4  | utf8mb4 | It works normally |

### Check List 

Tests 

- [x] Unit test
- Integration test
- No code

### Release note 

```release-note
```

### DIFF
--- a/components/tidb_query_datatype/src/codec/collation/charset.rs
+++ b/components/tidb_query_datatype/src/codec/collation/charset.rs
@@ -4,9 +4,15 @@ use std::str;
 
 use super::*;
 
+#[derive(PartialEq)]
+pub enum Charset {
+    Binary,
+    Utf8Mb4,
+}
+
 pub struct CharsetBinary;
 
-impl Charset for CharsetBinary {
+impl super::Charset for CharsetBinary {
     type Char = u8;
 
     #[inline]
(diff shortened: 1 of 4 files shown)


sticnarf (approved):

YangKeao:
> Some import paths have to be fixed.

Yep. I found something (like `Charset`) is missing. I may still need sometime to read the code / rewrite the macro :crying_cat_face: .

sticnarf (approved):
LGTM

wjhuang2016 (approved):

PR #13924, opened 2022-12-09
*: add raftstore v2 only tablet optimization

BusyJay:
Issue Number: Ref #12842 

What's Changed:

```commit-message
- No WAL should be written
- raftcf is dropped
- No concurrent write
- No multi batch write
- Use smaller bloomfilter ratio to reduce memory footprint
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/components/engine_traits/src/tablet.rs
+++ b/components/engine_traits/src/tablet.rs
@@ -1,6 +1,7 @@
 // Copyright 2022 TiKV Project Authors. Licensed under Apache-2.0.
 
 use std::{
+    fmt::{self, Debug, Formatter},
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicU64, Ordering},
@@ -9,6 +10,7 @@ use std::{
 };
 
 use collections::HashMap;
+use kvproto::metapb::Region;
 use tikv_util::box_err;
 
 use crate::{Error, Result};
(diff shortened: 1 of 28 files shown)


components/raftstore-v2/tests/integrations/test_basic_write.rs
```diff
@@ -147,4 +147,8 @@ fn test_put_delete() {
     assert!(!resp.get_header().has_error(), "{:?}", resp);
     let snap = router.stale_snapshot(2);
     assert_matches!(snap.get_value(b"key"), Ok(None));
+
+    // Check if WAL is skipped for basic writes.
+    let mut cached = cluster.node(0).tablet_registry().get(2).unwrap();
+    check_skip_wal(cached.latest().unwrap().as_inner().path());
```
tabokie, line 153: How does this pass? WAL will be created no matter how according to https://github.com/tikv/rocksdb/blob/8899a360623988adeec1d865bef9d24568ddd32c/db/db_impl/db_impl_open.cc#L1671

src/storage/config.rs
```diff
@@ -31,12 +31,21 @@ const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
...
+}
+
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, OnlineConfig)]
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct Config {
     #[online_config(skip)]
     pub data_dir: String,
+    #[online_config(skip)]
+    pub engine: EngineType,
```
tabokie, line 48: Maybe kv_engine_type, "engine" is confusing in certain context.

src/storage/config.rs
```diff
@@ -31,12 +31,21 @@ const DEFAULT_SCHED_PENDING_WRITE_MB: u64 = 100;
 const DEFAULT_RESERVED_SPACE_GB: u64 = 5;
 const DEFAULT_RESERVED_RAFT_SPACE_GB: u64 = 1;
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub enum EngineType {
+    RaftKv,
```
tabokie, line 37: A little confusing too. Maybe "shared" and "dedicated"?

components/raftstore-v2/tests/integrations/test_basic_write.rs
BusyJay: I check whether the file size is 0 instead of whether the file exists.

src/storage/config.rs
BusyJay: kv_engine is a concept related to raft_engine. Here, `EngineType` means `tikv_kv::Engine`, it can be `BtreeEngine`, `RocksEngine` and `RaftKv`.

BusyJay:
PTAL

tabokie (approved):

components/engine_traits/src/tablet.rs
```diff
@@ -69,20 +71,67 @@ impl<EK: Clone> CachedTablet<EK> {
...
+    pub suffix: Option<u64>,
+    /// The expected start key of the tablet. The key should be in the format
+    /// tablet is actually stored, for example should have `z` prefix.
+    ///
+    /// Any key that is smaller than this key can be considered obsolete.
+    pub start_key: Box<[u8]>,
+    /// The expected end key of the tablet. The key should be in the format
+    /// tablet is actually stored, for example should have `z` prefix.
+    ///
+    /// Any key that is larger than or equal to this key can be considered
```
tonyxuqqi, line 91: I think start_key and end_key can be optional instead. In some cases these are not needed (such as destroy_tablet)

components/engine_traits/src/tablet.rs
```diff
@@ -69,20 +71,67 @@ impl<EK: Clone> CachedTablet<EK> {
...
 // rename engine to tablet, so always use tablet for new traits/types.
 pub trait TabletFactory<EK>: Send + Sync {
     /// Open the tablet in `path`.
-    ///
-    /// `id` and `suffix` is used to mark the identity of tablet. The id is
-    /// likely the region Id, the suffix could be the current raft log
-    /// index. The reason to have suffix is that we can keep more than one
-    /// tablet for a region.
-    fn open_tablet(&self, id: u64, suffix: Option<u64>, path: &Path) -> Result<EK>;
+    fn open_tablet(&self, ctx: TabletContext, path: &Path) -> Result<EK>;
```
tonyxuqqi, line 131: Does the callee need to own the ctx? If not, maybe &TabletContext is enough? (in case the caller needs to use it somewhere else)

components/engine_traits/src/tablet.rs
BusyJay: It depends. For the moment it doesn't need it. But if we try to cleanup the data using rocksdb's API (DB::destroy), we will need options, which may need start key and end key.

components/engine_traits/src/tablet.rs
BusyJay: It needs to construct a compaction filter.

tonyxuqqi (approved):

Issue #11312, opened 2021-11-10
`make docker` failed on Linux


tabokie:
### Steps to reproduce
`make docker`
### What did you expect?

Build succeeds.

### What did happened?
```
Step 23/40 : RUN source /opt/rh/devtoolset-8/enable && make build_dist_release

...

error: failed to parse manifest at `/tikv/components/external_storage/export/Cargo.toml`

Caused by:
  can't find `tikv-cloud-storage` bin, specify bin.path
```

Issue #13867, opened 2022-12-01
log-backup: tracking issue for unify the infrastructure with other components 


YuJuncen:
## Development Task
We are using the `etcdv3/etcd-client` library for taking with etcd in PD, because there isn't a mature etcd client implemented with `grpc-rs`. However (sadly!) this requirement brings another ecosystem of infrastructures into the codebase of TiKV:
- [x] `rustls` vs `openssl`
- [ ] `tonic` vs `grpc-rs`
- [ ] `prost!` vs `protobuf-rs`

The most visible problem is we need to compile two different ecosystems: till now not so bad, because `external_storage` has already bring a subtree of `tokio` ecosystem into TiKV.

Things went even worse when we were going to talk about compatibility problems here: for example, a certificate provided for TiKV now must be generic enough to pass the checking both in `openssl` and `rustls`: the set of available certs now shrinks to the intersection of `available_certs_of(RUSTLS)` and `available_certs_of(OPENSSL)`, which is inconsistent and annoying.

We need try to migrate into the existing ecosystem in TiKV for long term.

PR #12430, opened 2022-04-25
backup-stream: Merge the feature branch `3pointer/br-stream` into master

YuJuncen:
### What is changed and how it works?

Issue Number: close https://github.com/tikv/tikv/issues/12429
Also ref https://github.com/pingcap/tidb/issues/29501.

What's Changed:

```commit-message
Added a new component `backup-stream` for TiKV, which allows TiKV listen the change from raftstore and collect them to external storage.
```

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
(TBD)

### Check List 

Tests 

- [x] Unit test 
- [x] Integration test
- [x] Manual test (add detailed scripts or steps below)
- No code

Side effects

- Performance regression
    - Consumes more CPU
    - Consumes more MEM

### Release note 

```release-note
Added the support for log backup.
```

PR #13937, opened 2022-12-14
fix docker build

tabokie:

Issue Number: Ref #11312

Fix a series of errors during `make docker`:
- failed to parse manifest
- Could not find `protoc` installation
- the linking modifiers `+bundle` and `+whole-archive` are not compatible with each other when generating rlibs
- error while loading shared libraries: libstdc++.so.6

```commit-message
Fix `make docker`.
```

### Check List 

Tests 

- Unit test

### Release note 

```release-note
None
```


tonyxuqqi (approved):

kolbe:
I checked out this PR and "make docker" still failed:

```
   Compiling server v0.0.1 (/tikv/components/server)
error: the linking modifiers `+bundle` and `+whole-archive` are not compatible with each other when generating rlibs

error: could not compile `tikv-ctl` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `tikv-server` due to previous error
+ r=101
+ set +x
make[1]: *** [Makefile:426: x-build-dist] Error 101
make[1]: Leaving directory '/tikv'
make: *** [Makefile:257: build_dist_release] Error 2
The command '/bin/sh -c source /opt/rh/devtoolset-8/enable && make build_dist_release' returned a non-zero code: 2
make: *** [Makefile:392: docker] Error 2
```

Dockerfile
```diff
@@ -50,6 +50,11 @@ RUN ln -s /usr/bin/cmake3 /usr/bin/cmake
 ENV LIBRARY_PATH /usr/local/lib:$LIBRARY_PATH
 ENV LD_LIBRARY_PATH /usr/local/lib:$LD_LIBRARY_PATH
 
+# Install protoc
+RUN curl -LO "https://github.com/protocolbuffers/protobuf/releases/download/v3.15.8/protoc-3.15.8-linux-x86_64.zip"
```
BusyJay, line 54: Why need this?

tabokie: I don't know, apparently it's needed after https://github.com/tikv/tikv/pull/12430.

BusyJay: /cc @YuJuncen

Dockerfile
```diff
@@ -72,8 +77,7 @@ RUN mkdir -p ./cmd/tikv-ctl/src ./cmd/tikv-server/src && \
     echo 'fn main() {}' > ./cmd/tikv-ctl/src/main.rs && \
     echo 'fn main() {}' > ./cmd/tikv-server/src/main.rs && \
     for cargotoml in $(find . -type f -name "Cargo.toml"); do \
-        sed -i '/fuzz/d' ${cargotoml} && \
-        sed -i '/profiler/d' ${cargotoml} ; \
+        sed -i '/fuzz/d' ${cargotoml} ; \
     done
```
tonyxuqqi, line 81: Do we consider to install libstdc++.so.6 in the run-time container (alpine) otherwise it will complain cannot find the file.

Connor1996 (commented):
Should create a new issue, don't know what's the error to be solved.

Dockerfile
YuJuncen: Because #12430 imports `tonic` which uses `prost` as default (it seems we cannot trivially replace `prost` with `protobuf`), and `prost` doesn't bundle `protoc` any more after some version.

Perhaps we'd better get rid of `prost` in the future or port `protobuf-rust` to `tonic`.  The progress could be tracked in https://github.com/tikv/tikv/issues/13867.

Connor1996 (approved):

